Issue #14, opened 2013-04-29
Hook into the mapping process by calling @BeforeMapping / @AfterMapping methods


agudian:
... e.g. to call a certain method that might flush the entity manager before starting to map ;).

Perhaps by:
* making the Freemarker template file name configurable in the AP
* or (perhaps better): allow to specify a super class for the mapping impl and call before / after methods. The super class might define methods for that, annotated with @BeforeMapping or something similar.

Cheers,
Andreas

agudian:
So guys, time to find out what we can do with this one in order to make all of us as happy as we are with the mapping config inheritance... :)

This is comment going to be a bit lengthy, so let's start with summarizing what problem I personally wanted to solve with this.

# The Problem

We are mapping between DTOs and JPA Entities and use optimistic locking to detect concurrent modifications. That means: entities carry a ```version``` attribute, which needs to be unchanged compared to the current database value when a DTO comes back from the client with the modified data. Hibernate increases the version automatically during updates and verifies the expected value in the database.
However, Hibernate increases the version of an entity instance only during the *flush*. So when mapping an entity to a DTO, we need to ensure that any modifications to the entity have been flushed already.
To emphasise this: for us, MapStruct would bring in only half the value if we were not able to solve this problem within the generated code, but had to go back to AOP proxies again (loosing some of the performance gains and requiring a noticable amount of additional memory).

# The Requirements

Whenever any entity (for us, all of them are based on an abstract base class) is the source of a mapping and the target of that mapping is a DTO (again, we have an abstract base class mirroring the properties of the base entity), we want to call a specific method somewhere with the source entity as argument (for us: ```flushIfRequired( entity )```).
Similar to what I outlined in the config-inheritance issue, we have so many mapper classes and entity / DTO pairs, that we don't want to specifically need to add an annotation / configuration / decorator method for all those mapping methods. We need some automatic approach that works on matching the parameter types of the mapping methods. If someone forgot to configure that for one method, it could very well be that it is discovered only when using the application.

# Solution in the current PR

The problem is solved for us by the current PR, and it has been part of our production code at work for a long long time now. That's one of the reasons why I regularly rebased it... :wink:

This is, what it does:
It introduces two new annotations: ```@BeforeMapping``` and ```@AfterMapping```, to be applied to methods that are to be *called* from generated mapping methods. In the code, I used the term "lifecycle method" for such annotated methods.
[...]

PR #84, opened 2014-01-04
#14 Add basic support for @AfterMapping / @BeforeMapping methods

agudian:
As outlined in issue #14.

Based on PR #67, as I reused some methods I created there... ;)

Oh, one thing that is not yet working is assigning potential return values from the callback method to any other objects. I'd like to deal with that in an extra issue.

Something that I implemented along the way (to get my test running), was making the MethodRetrievalProcessor also pick up methods declared in implemented interfaces and super-types of the type in question.

### DIFF
--- a/core-common/src/main/java/org/mapstruct/AfterMapping.java
+++ b/core-common/src/main/java/org/mapstruct/AfterMapping.java
@@ -26,12 +26,77 @@ import java.lang.annotation.Target;
 import org.mapstruct.util.Experimental;
 
 /**
- * Marks a method to be invoked at the end of a generated the mapping method.
+ * Marks a method to be invoked at the end of a generated mapping method, right before the last {@code return} statement
+ * of the mapping method. The method can be implemented in an abstract mapper class or be declared in a type (class or
+ * interface) referenced in {@link Mapper#uses()} in order to be used in a mapping method.
  * <p>
- * If the method has parameters, the method invocation is only generated if all parameters can be assigned by the source
- * or target parameters of the mapping method.
+ * Only methods with return type {@code void} may be annotated with this annotation.
+ * <p>
+ * If the method has parameters, the method invocation is only generated if all parameters can be <em>assigned</em> by
+ * the source or target parameters of the mapping method:
+ * <ul>
+ * <li>A parameter annotated with {@code @}{@link MappingTarget} is populated with the target instance of the mapping.
+ * </li>
+ * <li>A parameter annotated with {@code @}{@link TargetType} is populated with the target type of the mapping.</li>
+ * <li>Any other parameter is populated with a source parameter of the mapping, whereas each source parameter is used
+ * once at most.</li>
+ * </ul>
+ * <p>
+ * All <em>after-mapping</em> methods that can be applied to a mapping method will be used. Their order is determined by
+ * their location of definition:
+ * <ul>
+ * <li>The order of methods within one type can not be guaranteed, as it depends on the compiler and the processing
+ * environment implementation.</li>
+ * <li>Methods declared in one type are used after methods declared in their super-type.</li>
+ * <li>Methods implemented in the mapper itself are used before methods from types referenced in {@link Mapper#uses()}.
+ * </li>
+ * <li>Types referenced in {@link Mapper#uses()} are searched for <em>after-mapping</em> methods in the order specified
+ * in the annotation.</li>
+ * </ul>
+ * <p>
+ * Example:
+ *
+ * <pre>
+ * <code>
+ * &#64;AfterMapping
+ * public void calledWithoutArgs() {
+ *     // ...
+ * }
+ *
+ * &#64;AfterMapping
+ * public void calledWithSourceAndTargetType(SourceEntity anySource, &#64;TargetType Class&lt;?&gt; targetType) {
+ *     // ...
+ * }
+ *
+ * &#64;AfterMapping
+ * public void calledWithSourceAndTarget(Object anySource, &#64;MappingTarget TargetDto target) {
+ *     // ...
+ * }
+ *
+ * public abstract TargetDto toTargetDto(SourceEntity source);
+ *
+ * // generates:
+ *
+ * public TargetDto toTargetDto(SourceEntity source) {
+ *     if ( source == null ) {
+ *         return null;
+ *     }
+ *
+ *     TargetDto targetDto = new TargetDto();
+ *
+ *     // actual mapping code
+ *
+ *     calledWithoutArgs();
+ *     calledWithSourceAndTargetType( source, TargetDto.class );
+ *     calledWithSourceAndTarget( source, targetDto );
+ *
+ *     return targetDto;
+ * }
+ * </code>
+ * </pre>
  *
  * @author Andreas Gudian
+ * @see BeforeMapping
  */
 @Experimental
 @Target(ElementType.METHOD)
--- a/core-common/src/main/java/org/mapstruct/AfterMapping.java
+++ b/core-common/src/main/java/org/mapstruct/AfterMapping.java
@@ -1,0 +1,41 @@
+/**
+ *  Copyright 2012-2015 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.mapstruct.util.Experimental;
+
+/**
+ * Marks a method to be invoked at the end of a generated the mapping method.
+ * <p>
+ * If the method has parameters, the method invocation is only generated if all parameters can be assigned by the source
+ * or target parameters of the mapping method.
+ *
+ * @author Andreas Gudian
+ */
+@Experimental
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.CLASS)
+public @interface AfterMapping {
+
+}
--- a/core-common/src/main/java/org/mapstruct/BeforeMapping.java
+++ b/core-common/src/main/java/org/mapstruct/BeforeMapping.java
@@ -1,0 +1,41 @@
+/**
+ *  Copyright 2012-2015 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.mapstruct.util.Experimental;
+
+/**
+ * Marks a method to be invoked at the beginning of a generated mapping method.
+ * <p>
+ * If the method has parameters, the method invocation is only generated if all parameters can be assigned by the source
+ * or target parameters of the mapping method.
+ *
+ * @author Andreas Gudian
+ */
+@Experimental
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.CLASS)
+public @interface BeforeMapping {
+
+}
--- a/core-common/src/main/java/org/mapstruct/BeforeMapping.java
+++ b/core-common/src/main/java/org/mapstruct/BeforeMapping.java
@@ -26,12 +26,81 @@ import java.lang.annotation.Target;
 import org.mapstruct.util.Experimental;
 
 /**
- * Marks a method to be invoked at the beginning of a generated mapping method.
+ * Marks a method to be invoked at the beginning of a generated mapping method. The method can be implemented in an
+ * abstract mapper class or be declared in a type (class or interface) referenced in {@link Mapper#uses()} in order to
+ * be used in a mapping method.
  * <p>
- * If the method has parameters, the method invocation is only generated if all parameters can be assigned by the source
- * or target parameters of the mapping method.
+ * Only methods with return type {@code void} may be annotated with this annotation.
+ * <p>
+ * If the method has parameters, the method invocation is only generated if all parameters can be <em>assigned</em> by
+ * the source or target parameters of the mapping method:
+ * <ul>
+ * <li>A parameter annotated with {@code @}{@link MappingTarget} is populated with the target instance of the mapping.
+ * </li>
+ * <li>A parameter annotated with {@code @}{@link TargetType} is populated with the target type of the mapping.</li>
+ * <li>Any other parameter is populated with a source parameter of the mapping, whereas each source parameter is used
+ * once at most.</li>
+ * </ul>
+ * If a <em>before-mapping</em> method does not contain a {@code @}{@link MappingTarget} parameter, it is invoked
+ * directly at the beginning of the applicable mapping method. If it contains a {@code @}{@link MappingTarget}
+ * parameter, the method is invoked after the target parameter has been initialized in the mapping method.
+ * <p>
+ * All <em>before-mapping</em> methods that can be applied to a mapping method will be used. Their order is determined
+ * by their location of definition:
+ * <ul>
+ * <li>The order of methods within one type can not be guaranteed, as it depends on the compiler and the processing
+ * environment implementation.</li>
+ * <li>Methods declared in one type are used after methods declared in their super-type.</li>
+ * <li>Methods implemented in the mapper itself are used before methods from types referenced in {@link Mapper#uses()}.
+ * </li>
+ * <li>Types referenced in {@link Mapper#uses()} are searched for <em>after-mapping</em> methods in the order specified
+ * in the annotation.</li>
+ * </ul>
+ * <p>
+ * Example:
+ *
+ * <pre>
+ * <code>
+ * &#64;BeforeMapping
+ * public void calledWithoutArgs() {
+ *     // ...
+ * }
+ *
+ * &#64;BeforeMapping
+ * public void calledWithSourceAndTargetType(SourceEntity anySource, &#64;TargetType Class&lt;?&gt; targetType) {
+ *     // ...
+ * }
+ *
+ * &#64;BeforeMapping
+ * public void calledWithSourceAndTarget(Object anySource, &#64;MappingTarget TargetDto target) {
+ *     // ...
+ * }
+ *
+ * public abstract TargetDto toTargetDto(SourceEntity source);
+ *
+ * // generates:
+ *
+ * public TargetDto toTargetDto(SourceEntity source) {
+ *     calledWithoutArgs();
+ *     calledWithSourceAndTargetType( source, TargetDto.class );
+ *
+ *     if ( source == null ) {
+ *         return null;
+ *     }
+ *
+ *     TargetDto targetDto = new TargetDto();
+ *
+ *     calledWithSourceAndTarget( source, targetDto );
+ *
+ *     // actual mapping code
+ *
+ *     return targetDto;
+ * }
+ * </code>
+ * </pre>
  *
  * @author Andreas Gudian
+ * @see AfterMapping
  */
 @Experimental
 @Target(ElementType.METHOD)
--- a/core-common/src/main/java/org/mapstruct/MapMapping.java
+++ b/core-common/src/main/java/org/mapstruct/MapMapping.java
@@ -67,9 +67,10 @@ public @interface MapMapping {
 
 
     /**
-     * A value qualifier can be specified to aid the selection process of a suitable mapper. This is useful in case
-     * multiple mappers (hand written of internal) qualify and result in an 'Ambiguous mapping methods found' error.
-     *
+     * A value qualifier can be specified to aid the selection process of a suitable mapper for the values in the map.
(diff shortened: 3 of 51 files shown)


agudian:
Another rebase (and squash, which makes the rebasing easier for me)....

Just in time for this PR's first Birthday tomorrow! Everyone who's not merging this one has at least to sing now... "Happy Birthday... " :musical_note:  :birthday: :balloon: 

Soon enough this PR will be old enough to merge itself.

sjaakd:
> Soon enough this PR will be old enough to merge itself.

:smile:

agudian:
@gunnarmorling / @sjaakd: two days ago, I've pushed some rebased fresh commits to this PR, with added support for qualifier annotations on top of it.

Pls feel free to do a review. The one thing that I'm aware of that's missing is the support for these lifecycle methods in enum mapping methods. I'll add that with an additional commit / PR later.

sjaakd:
@agudian , I'll try to take a look later this week.  

I tried to give motivation for my reasoning back when I introcuded on the qualifier approach (yesterday). Can you take a look at that?

agudian:
> I tried to give motivation for my reasoning back when I introcuded on the qualifier approach (yesterday). Can you take a look at that?

I'm not really sure what you refer to. What did I miss?

sjaakd:
> I'm not really sure what you refer to. What did I miss?

I think we covered it now off-line..

agudian:
Rebased, fixed a bug in the first commit, removed the changes to the QualifierSelector as discussed, and added support for EnumMappings.

I think it's ready to go now... just waiting for your reviews. Most of it are tests, the rest is in the largest parts quite trivial... :wink:

gunnarmorling:
Cool, thanks, @agudian! I plan to look into it on the weekend.

sjaakd:
@agudian , I reviewed the code briefly. I think that I've done this one before, so I've got no additional comments except the 2 above. I'm ok with your approach.

agudian:
> @agudian , I reviewed the code briefly. I think that I've done this one before, so I've got no additional comments except the 2 above. I'm ok with your approach.

Yes, you reviewed it quite a while ago already and I had you comments worked in. Thanks for taking a look again - with all those rebases over the time, some stuff has changed since then (for instance on the MethodMatcher, which I had to retrofit for multiple source arguments again :wink:) - so I'm glad you took another look.

core-common/src/main/java/org/mapstruct/AfterMapping.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks a method to be invoked at the end of a generated the mapping method.
+ * <p>
+ * If the method has parameters, the method invocation is only generated if all parameters can be assigned by the source
```
gunnarmorling: Maybe it'd be a good idea to add an example to the JavaDoc (I think there is an issue for this in general, but we may do it right away for new APIs)?

core-common/src/main/java/org/mapstruct/MapMapping.java
```diff
@@ -55,6 +55,19 @@
     String valueDateFormat() default "";
 
     /**
+     * A qualifier can be specified to aid the selection process of a suitable mapper. This is useful in case multiple
+     * mappers (hand written of internal) qualify and result in an 'Ambiguous mapping methods found' error. A qualifier
+     * is a custom annotation and can be placed on either a hand written mapper class or a method.
+     * <p>
+     * Qualifiers listed here are used for the selection of method-level aspects, and in case nothing is specified for
```
gunnarmorling: What are "method-level aspects"? Why is this method needed in addition to `keyQualifiedBy()`/ `valueQualifiedBy()`?

processor/src/main/java/org/mapstruct/ap/model/source/MethodMatcher.java
```diff
@@ -170,36 +170,59 @@ private boolean matchResultType(Type resultType,
...
             }
         }
         return true;
     }
 
+    /**
+     * @param type the type
+     * @return {@code true}, if the type does NOT represent java.lang.Object
+     */
+    private boolean isNotObjectClass(TypeMirror type) {
```
gunnarmorling: I'd avoid the negated name: `isJavaLangObject(TypeMirror type)`.

processor/src/main/java/org/mapstruct/ap/model/source/MethodMatcher.java
```diff
@@ -170,36 +170,59 @@ private boolean matchResultType(Type resultType,
...
+        return !( type.getKind() == TypeKind.DECLARED
+                        && ( (TypeElement) ( (DeclaredType) type ).asElement() ).getQualifiedName().contentEquals(
+                            Object.class.getName() ) );
+    }
+
+    private static <E> boolean hasNonNullElements(List<E> elements) {
```
gunnarmorling: Could go to `Collections`?

processor/src/main/java/org/mapstruct/ap/model/source/SourceMethod.java
```diff
@@ -516,4 +518,28 @@ public MapperConfiguration getMapperConfiguration() {
...
+
+    public boolean isAfterMappingMethod() {
+        return isAfterMappingMethod( getExecutable() );
+    }
+
+    public boolean isBeforeMappingMethod() {
+        return isBeforeMappingMethod( getExecutable() );
+    }
+
+    public static boolean isLifecycleCallbackMethod(ExecutableElement executableElement) {
```
gunnarmorling: Should these three go to `Executables`?

core-common/src/main/java/org/mapstruct/AfterMapping.java
gunnarmorling: Could you also add a note describing what happens if there are several matching lifecycle methods (ordering)?

processor/src/main/resources/org.mapstruct.ap.model.IterableMappingMethod.ftl
```diff
@@ -50,6 +56,21 @@
         <#if !existingInstanceMapping>
             <@includeModel object=resultElementType/>[] ${resultName} = new <@includeModel object=resultElementType/>[<@iterableSize/>];
         </#if>
+    <#else>
+        <#if existingInstanceMapping>
```
gunnarmorling: How comes this stuff has changed?

gunnarmorling:
@agudian Did a first round and added some minor comments. I'll continue tomorrow. Thanks a lot for this piece of work!

agudian:
> > Is it the static nature of the invocations list that disturbs you? I could certainly change that and also add a getter for the referenced mapper. Shall I?
> 
> That would be a pretty useful extension.. Perhaps a separate Issue / PR.

Ah, right - now I'm back on track. I had in mind just changing my testcase, but you're right, this would require a separate issue.

core-common/src/main/java/org/mapstruct/AfterMapping.java
agudian: Good idea, I'll add the code examples there and go over the wording in general.

core-common/src/main/java/org/mapstruct/MapMapping.java
agudian: I forgot to change some of that back again. Originally, I had implemented that in case no key/value qualifiers were given, then qualifiers from this property are taken for them as well. I removed that part again and forgot to adjust the Javadoc.
The qualifiedBy method itself was added to have a place for qualifiers that are taken into account for selecting before/after methods.

processor/src/main/java/org/mapstruct/ap/model/source/MethodMatcher.java
agudian: done

processor/src/main/java/org/mapstruct/ap/model/source/MethodMatcher.java
agudian: done

processor/src/main/java/org/mapstruct/ap/model/source/SourceMethod.java
agudian: I moved them, looks good.

processor/src/main/resources/org.mapstruct.ap.model.IterableMappingMethod.ftl
agudian: I've moved those parts together so that there is only one block of result initialization so that I can follow with one single block for the ```beforeMappingReferencesWithMappingTarget``` references (that are ```@BeforeMapping``` methods with an ```@MappingTarget``` parameter and so they need an initialized target value.

agudian:
OK, pushed a commit addressing the comments of Gunnar. Waiting for the next review round... ;-)

processor/src/test/java/org/mapstruct/ap/test/callbacks/ClassContainingCallbacks.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    @BeforeMapping
+    public void withSourceBeforeMapping(Source source) {
+        INVOCATIONS.add( new Invocation( "withSourceBeforeMapping", source ) );
+    }
+
+    @BeforeMapping
+    @Qualified
+    public void withSourceBeforeMappingQualified(Source source) {
```
gunnarmorling: @agudian What is the influence of qualifiers wrt. to before/after mapping methods?

core-common/src/main/java/org/mapstruct/MapMapping.java
```diff
@@ -55,10 +55,21 @@
...
-     * error.
+     * A qualifier can be specified to aid the selection process of suitable {@link BeforeMapping} /
+     * {@link AfterMapping} methods.
+     * <p>
+     * Use {@link #keyQualifiedBy()} / {@link #valueQualifiedBy()} to specify qualifiers to be used for selecting key /
+     * value mapping methods.
      *
+     * @return the qualifiers
+     */
+    Class<? extends Annotation>[] qualifiedBy() default { };
```
gunnarmorling: Hum, I think that's quite confusing, given that there are `keyQualifiedBy()` and `valueQualifiedBy()` already. Is the application of qualifiers actually needed for the lifecycle methods? Or is this something which we could leave out, at least for now?

processor/src/main/java/org/mapstruct/ap/model/source/MethodMatcher.java
```diff
@@ -170,36 +170,59 @@ private boolean matchResultType(Type resultType,
...
+     * @param type the type
+     * @return {@code true}, if the type represents java.lang.Object
+     */
+    private boolean isJavaLangObject(TypeMirror type) {
+        return type.getKind() == TypeKind.DECLARED
+            && ( (TypeElement) ( (DeclaredType) type ).asElement() ).getQualifiedName().contentEquals(
+                Object.class.getName() );
+    }
+
+    private static <E> boolean hasNonNullElements(Iterable<E> elements) {
```
gunnarmorling: Could be moved to `Collections`?

core-common/src/main/java/org/mapstruct/AfterMapping.java
```diff
@@ -0,0 +1,103 @@
...
+ * <p>
+ * Only methods with return type {@code void} may be annotated with this annotation.
+ * <p>
+ * If the method has parameters, the method invocation is only generated if all parameters can be <em>assigned</em> by
+ * the source or target parameters of the mapping method:
+ * <ul>
+ * <li>A parameter annotated with {@code @}{@link MappingTarget} is populated with the target instance of the mapping.
+ * </li>
+ * <li>A parameter annotated with {@code @}{@link TargetType} is populated with the target type of the mapping.</li>
+ * <li>Any other parameter is populated with a source parameter of the mapping, whereas each source parameter is used
```
gunnarmorling: The "at most once" par makes me curious. If there e.g. is a method with two source parameters, and a before-mapping method exists, which has a single parameter - of one of the other method's source parameter types will it be invoked? Or will a before-mapping method only be invoked if it has exactly matching source parameter types?

core-common/src/main/java/org/mapstruct/BeforeMapping.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p>
+ * Only methods with return type {@code void} may be annotated with this annotation.
+ * <p>
+ * If the method has parameters, the method invocation is only generated if all parameters can be <em>assigned</em> by
+ * the source or target parameters of the mapping method:
+ * <ul>
+ * <li>A parameter annotated with {@code @}{@link MappingTarget} is populated with the target instance of the mapping.
+ * </li>
+ * <li>A parameter annotated with {@code @}{@link TargetType} is populated with the target type of the mapping.</li>
+ * <li>Any other parameter is populated with a source parameter of the mapping, whereas each source parameter is used
```
gunnarmorling: The "at most once" par makes me curious. If there e.g. is a method with two source parameters, and a before-mapping method exists, which has a single parameter - of one of the other method's source parameter types will it be invoked? Or will a before-mapping method only be invoked if it has exactly matching source parameter types?

core-common/src/main/java/org/mapstruct/BeforeMapping.java
```diff
@@ -0,0 +1,108 @@
...
+ * once at most.</li>
+ * </ul>
+ * If a <em>before-mapping</em> method does not contain a {@code @}{@link MappingTarget} parameter, it is invoked
```
gunnarmorling: What is the use case for supporting `@MappingTarget` here?

core-common/src/main/java/org/mapstruct/BeforeMapping.java
```diff
@@ -0,0 +1,108 @@
...
+ * directly at the beginning of the applicable mapping method. If it contains a {@code @}{@link MappingTarget}
+ * parameter, the method is invoked after the target parameter has been initialized in the mapping method.
```
gunnarmorling: Isn't the mapping target already passed in through the parameter?

core-common/src/main/java/org/mapstruct/BeforeMapping.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p>
+ * All <em>before-mapping</em> methods that can be applied to a mapping method will be used. Their order is determined
+ * by their location of definition:
+ * <ul>
+ * <li>Methods within one type are used in the order of their definition (declarations from compiled .class files may
```
gunnarmorling: I feel we should not guarantee the order within declaring classes. As you say, it's not specified  for referenced classes anyways, and I feel it's overly specific. Better let's not promise too much.

core-common/src/main/java/org/mapstruct/BeforeMapping.java
```diff
@@ -0,0 +1,108 @@
...
+ * differ in order, depending on the compiler used for their compilation and the compiler used for generating the
+ * mapper).</li>
+ * <li>Methods declared in one type are used before methods declared in their super-type.</li>
```
gunnarmorling: What is the rational for this order? E.g. `@Before` methods in JUnit are invoked from super-types first.

core-common/src/main/java/org/mapstruct/AfterMapping.java
```diff
@@ -0,0 +1,103 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks a method to be invoked at the end of a generated mapping method, right before the last {@code return} statement
+ * of the mapping method. The method can be implemented in an abstract mapper class or be declared in a type (class or
+ * interface) referenced in {@link Mapper#uses()} in order to be used in a mapping method.
+ * <p>
+ * Only methods with return type {@code void} may be annotated with this annotation.
```
gunnarmorling: Why are source parameters passed to `@AfterMapping` methods? I would only have expected a single parameter being supported which takes the result of the decorated mapping method?

core-common/src/main/java/org/mapstruct/AfterMapping.java
```diff
@@ -0,0 +1,103 @@
...
+ * <li>Any other parameter is populated with a source parameter of the mapping, whereas each source parameter is used
+ * once at most.</li>
+ * </ul>
+ * All <em>after-mapping</em> methods that can be applied to a mapping method will be used. Their order is determined by
+ * their location of definition:
+ * <ul>
+ * <li>Methods within one type are used in the order of their definition (declarations from compiled .class files may
+ * differ in order, depending on the compiler used for their compilation and the compiler used for generating the
+ * mapper).</li>
+ * <li>Methods declared in one type are used before methods declared in their super-type.</li>
```
gunnarmorling: JUnit does it the other way around. Should we do the same?

gunnarmorling:
@agudian Added some more comments. Any feedback welcome.

I am wondering whether we should mark that feature as "experimental" in the 1.0 Final by some means (e.g. in the docs), giving us some freedom to alter specific semantics in case needed. We could then either adapt in 1.1 or leave as is and "un-experimentalize". WDYT?

processor/src/test/java/org/mapstruct/ap/test/callbacks/ClassContainingCallbacks.java
agudian: Same as with the selection of property mapping methods or factory methods: using qualifiers, you limit the candidates to only those methods that carry the qualifiers. So for this particular method to be invoked, the mapping method needs `qualifiedBy=Qualifier.class` in `@BeanMapping`, `@IterableMapping` or `@MapMapping`.

core-common/src/main/java/org/mapstruct/MapMapping.java
agudian: Yea, I know what you mean. I added this here for completeness, as you wouldn't have any other way to use qualifiers for lifecycle methods together with map-mapping methods. There is no other application for this `qualifiedBy`, yet. 
One could think of selecting specific factory methods for initializing the map type to return (e.g. `@TinyMap`, or `@HugeMap`, or `@SynchronizedMap`, ...). If you think it's too much for now, I'll just drop it out again.

processor/src/main/java/org/mapstruct/ap/model/source/MethodMatcher.java
agudian: Right, we have that now. Will do... :+1:

core-common/src/main/java/org/mapstruct/BeforeMapping.java
agudian: If you have a mapping method with two source parameters `Adress` and `Person`, and `@BefroreMapping` methods without any parameters, with only an `Adress`, with only a `Person` and with both `Adress` and `Person`, then all four methods will be called.

core-common/src/main/java/org/mapstruct/BeforeMapping.java
agudian: Not necessarily - the `@BeforeMapping` method with the parameter list `(Source source, @MappingTarget target)` would be applicable to both `Target toTarget(Source s)` and `void intoTarget(Source s, @MappingTarget t)`. Because for the lifecycle method, it is only important to know - hey, there is going to be a bean mapping from Source to Target being done. `@MappingTarget` just denotes the instance to be returned by the mapping method.

To combine this with the previous comment, a use case could be some logging (for whatever reason), or if you'd want to perform a flush if you start mapping from an entity to a DTO, but not if you map from an entity into a JAXB type.

core-common/src/main/java/org/mapstruct/BeforeMapping.java
agudian: Right. And as I observed over the time, when working in Eclpse with the incremental generation, it is even more of a gamble to say what compilation units are pulled in as binaries and which are present in their source form.

core-common/src/main/java/org/mapstruct/BeforeMapping.java
agudian: It's arbitrary, but fixed - no other rationale behind it... I can modify the traversal order, if you find that to be more intuitive. 

(When I write up the docs, I should add a box containing a suggestion to not design lifecycle methods to be co-dependent / rely on specific ordering... that could be considered a code-smell).

core-common/src/main/java/org/mapstruct/AfterMapping.java
agudian: `@BeforeMapping` and `@AfterMapping` are both treated the same when it comes to parameter matching. On the one hand, the parameters of the lifecycle methods (in combination with `@MappingTarget` and `@TargetType`) are used to have control over the situations in which these methods are pulled in and applied. On the other hand, you are free to use these values to perform what ever you intended.
For example, an `@AfterMapping` method could be used to add some manual post-processing to the mapping target, using values form the mapping source (we have such a case at work, with abstract source and target types).

core-common/src/main/java/org/mapstruct/AfterMapping.java
agudian: Sure, why not... It makes sense for JUnit, so we could go along with the same arguments they have.

agudian:
> am wondering whether we should mark that feature as "experimental" in the 1.0 Final by some means (e.g. in the docs), giving us some freedom to alter specific semantics in case needed. We could then either adapt in 1.1 or leave as is and "un-experimentalize". WDYT?

Yes, the method selection might have some corner cases that could require some adjustments when faced with more real-world applications. So that sounds reasonable to me.

core-common/src/main/java/org/mapstruct/AfterMapping.java
gunnarmorling: I see. If it's that specific to parameter/return types, couldn't you use a decorator in this case, though?

agudian: Right, for specific types you could. At work we have a case where we need to do such a post-processing for all mappings where source and target type inherit from a specific pair of abstract base types. The processing that needs to be done works on the abstract types. The specific incarnations (seven or so) lie in different components all accross the applicaiton. So that already helps.
Not to mention that we can't make any good use of our decorator-feature with Spring, yet.

agudian:
@gunnarmorling added another commit to entertain your review comments :wink:

agudian:
Gunnar, do you have any further remarks that need to be handled in this PR? I'd like to have it merged before I start with the doc updates (our planned release weekend is getting closer :wink:)...

core-common/src/main/java/org/mapstruct/MapMapping.java
gunnarmorling: > If you think it's too much for now, I'll just drop it out again.

Yes, if you don't need it, then let's drop it.

gunnarmorling:
Can you add "#14" to the last two commit messages?

gunnarmorling:
@agudian, wrt. to marking this as experimenal, I would do it via a new meta-annotation `@org.mapstruct.util.Experimenal` which is given at the `@BeforeMapping` and `@AfterMapping` type annotations. Basically like what we do in OGM: https://github.com/hibernate/hibernate-ogm/blob/master/core/src/main/java/org/hibernate/ogm/util/Experimental.java

gunnarmorling:
@agudian, done with the review. If you could remove `@MapMapping#qualifier()`, add the issues keys to the commit messages and add the `@Experimental` annotation it's good to go from my side.

I am still not 100% convinced it's strictly needed, but time (and users) will tell. Either way, thanks for keeping that ball rolling and driving this thing home; The "MapStruct achievement award for PR rebasing endurance" is yours :)

agudian:
Awesome!

> Can you add "#14" to the last two commit messages?

My plan was to squash those commits when rebasing. :)

agudian:
> The "MapStruct achievement award for PR rebasing endurance" is yours :)

Exactly one week short of 1.5 years... Let's see if I have to pass on that trophy at some point... :trophy:

Thanks again for the continued reviews over the time!

sjaakd:
Congratulations!!!

Issue #168, opened 2014-03-08
Add ability to globally configure @Mapping for properties in reused super classes


agudian:
With #72 being resolved now, we need some means to allow ignoring certain properties from being mapped / reported as unmapped on a global scope.

I could think of something along the following lines (with much room for improvement for the naming):

```java
@GlobalMapping(
   sourceType=BaseEntity.class, 
    targetType=BaseDto.class, 
    mappings={ 
      @Mapping(target="uuid", ignore=true),
      @Mapping(source="id", target="technicalId") 
   }
)
```
That annotation (or a multiple of it) would be allowed at mapper level (e.g. in ```@Mapper```) and in the ```@MapperConfig``` annotation.

```sourceType``` and ```targetType``` would be optional, and the specified ```@Mapping``` configuration would be applied for each mapping method that has source / target types that fit (as in "assignable from").

gunnarmorling:
@agudian, I've been thinking a bit more about this. There is basically two things I'm concerned about with the proposed approach:

* The irregularity of global definitions: source and target types need to be specified within the annotations rather then method parameters / return types as with the existing means of mapping definitions; It also doesn't work for parameterized types (I don't think that's supported yet in general, but any solution we come up with should leave the door open)
* When looking at a mapping method definition, not all its property mappings are easily visible; You need to take global definitions possibly defined somewhere else into account

So what would you think about re-using `@InheritConfiguration` alternatively:

    @Mapping(source = "id", target = "primaryKey")
    BaseEntity dtoToEntity(BaseDto dto);

    @InheritConfiguration(name="dtoToEntity")
    CarEntity toCarEntity(CarDto carDto);

This would address the concerns above:

* "Global" (or basic) mappings would just be defined as any other mappings, it works for parameterized types.
* When looking at a specific method such as `toCarEntity()` there is a link to the method with the basic mappings, making things much easier to grasp

We'd also re-use an existing construct rather than defining a new one. Now some downsides:

* It doesn't allow for the definition of mappings where source or target is "any type". Personally I don't think it's a huge issue.
* It doesn't work if the target base type (e.g. `BaseDto`) is an abstract class as we cannot generate an implementation of the base mapping method then. That's a problem indeed. One solution I can think of is to define the base method on another mapper and have a flag in `@Mapper` which marks that mapper as not to be implemented, i.e. it's only supposed to serve as base mapping source. That'd require the inherit configuration mechanism to work across mappers. Maybe you have another, better idea for that?

[...]

PR #260, opened 2014-07-04
#168 allow to globally configure mappings using @GlobalMapping

agudian:
I had to move ```@Mapper``` and ```@MapperConfig``` to ```core``` and ```core-jdk8```, as they now require ```@GlobalMapping```, which in turn requires ```@Mapping```...

### DIFF
--- a/core-common/src/main/java/org/mapstruct/factory/Mappers.java
+++ b/core-common/src/main/java/org/mapstruct/factory/Mappers.java
@@ -18,11 +18,9 @@
  */
 package org.mapstruct.factory;
 
-import org.mapstruct.Mapper;
-
 /**
  * Factory for obtaining mapper instances if no explicit component model such as CDI is configured via
- * {@link Mapper#componentModel()}.
+ * {@link org.mapstruct.Mapper#componentModel()}.
  * <p>
  * Mapper implementation types are expected to have the same fully qualified name as their interface type, with the
  * suffix {@code Impl} appended. When using this factory, mapper types - and any mappers they use - are instantiated by
--- a/core-jdk8/src/main/java/org/mapstruct/GlobalMapping.java
+++ b/core-jdk8/src/main/java/org/mapstruct/GlobalMapping.java
@@ -1,0 +1,70 @@
+/**
+ *  Copyright 2012-2015 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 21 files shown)


agudian:
Rebased to current master... I'll start with the promised more elaborate example soon-ish.

agudian:
@gunnarmorling, @sjaakd: Here comes the example project:

In https://github.com/agudian/mapstruct-playground, I have created a self-containing, non-working scribble of how our project at work kind of looks like. We have one base entity, and one base DTO, plus some other base entity hiearchies e.g. like that AuditableEntity.

Most of our components have one single mapper, but some more complex components and two of the "other" applications that we are doing there use single mappers for each entity or small group of entites.

As of now the main application we have has more than 480 entity classes and the same number of DTO classes. Mainly because of these UUIDs in the DTOs and these audit-properties in the AuditableEntities, we use MapStruct with ```unmappedTargetPolicy=WARN``` - but that possibly hides actual unmappable properties.

With now more than 50 developers (I guess) in three countries and four projects that are based on our framework and that use MapStruct, we just have to make using it pretty simple:
* a centralized base configuration that takes care of those mandatory settings, preferrably rule based so that it can be controlled from that central location
* a fool-proof way to apply that centralized configuration

Besides the DTO/Entity mapping, we also use MapStruct for mappings to JAXB classes (from DTO or Entities) - so there might be an additional set of centralized configuration for those mappers.

I get the argument with the comprehensability - why is the code generated in the way it is, and how many hops do I need to take to find that piece of configuration. My take on this is that in such large projects as ours, it is not necessary that everyone must understand every notion of every framework involved. In fact, with the speed they need to get things done, they can't. Usually, they check how some specific thing is done in that other place and do it likewise. 

If only 10-20% of the developers in larger project understand why a certain property doesn't need to be explicitly ignored, or what that base-configuration does, then that's totally ok, IMHO. The odds are probably different in small projects - but then again, if one of the developers ask himself that question, then I can give an answer to that with two lines of text via instant messanger. For everyone else, it should still be simple enough to follow that Config-Class and check the Javadoc of those ```@GlobalMapping``` annotations. It's not that we hide anything - it is, as I think, still very much consistent to how we do many other things.

OK, enough the rambling. Too late now anyway. Your turn :wink:

sjaakd:
> OK, enough the rambling. Too late now anyway. Your turn

You don't have to convince me.. Its similar to my case. My case: [gml](http://schemas.opengis.net/gml/3.2.1/feature.xsd). All of our 20 - or so - datatypes that are going to be made the next couple of years are based on gml and related opengis application schemas. They inherit from `AbstractFeatureType` or even more rich `...FeatureTypes` in the opengis inheritance hierarchy. 

Most of the attributes should be populated default or even ignored in the mapping. Building these classes leads to java base classes for which I want to use a common base library (and hence default mappings). 

Note: it should therefor be possible to package MapStruct configuration packages in used libraries. I'm wondering whether we have the same problem there as described in #384 .

agudian:
Gunnar and I talked about this PR on the phone and he wondered why there was this class ```Any``` in ```@GlobalMapping``` is used as default / type wildcard instead of just using ```Object.class```. I didn't remember why I did that and after checking again, there's no techical reason. 

So I've now removed that Any type, rebased, updated the license headers, etc...

agudian:
Closed... replaced by #463

Issue #385, opened 2014-12-11
Ambiguous factory methods.


sjaakd:
When an `ObjectFactory`, e.g. one generated by JAXB contains 2 methods:

```java

public A createA() {
...
}

public B createB() {
...
}
```
And:

```java
public class A extends B {
}
```

MapStruct generates an Ambiguous factory method. MapStruct should be smart enough to select the 'tightest' match based on return type.

gunnarmorling:
Resolved by introducing new `@BeanMapping` method. I'll leave it open for as a reminder that the doc update still is pending.

sjaakd:
Documentation handled. Solution merged.

PR #391, opened 2014-12-17
#385 Inheritance selection on best fit source- and target type 

sjaakd:
* General improvement, factories and mapping method selection make use of the same (selector) mechanism.

* For the source-type, the fit was not applied in two directions. For that purpose the absolute (type) distance is used iso distance.

* There was no target-type inheritance implemented.

### DIFF
--- a/core-common/src/main/java/org/mapstruct/BeanMapping.java
+++ b/core-common/src/main/java/org/mapstruct/BeanMapping.java
@@ -1,0 +1,43 @@
+/**
+ *  Copyright 2012-2015 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Configures the mapping between two bean types
+ *
+ * @author Sjaak Derksen
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.SOURCE)
+public @interface BeanMapping {
+
+
+    /**
+     * Specifies the result type of the method to select in case ambiguous mappings / factory methods.
+     *
+     *
+     * @return the resultType to select
+     */
+    Class resultType() default void.class;
+}
--- a/processor/src/main/java/org/mapstruct/ap/model/AssignmentFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/AssignmentFactory.java
@@ -21,15 +21,12 @@ package org.mapstruct.ap.model;
 import java.util.List;
 import java.util.Set;
 
-import javax.tools.Diagnostic;
 
 import org.mapstruct.ap.model.assignment.Assignment;
 import org.mapstruct.ap.model.common.ConversionContext;
 import org.mapstruct.ap.model.common.Type;
 import org.mapstruct.ap.model.source.Method;
-import org.mapstruct.ap.model.source.SourceMethod;
 import org.mapstruct.ap.model.source.builtin.BuiltInMethod;
-import org.mapstruct.ap.model.source.selector.MethodSelectors;
 
 /**
  * Factory class for creating all types of assignments
@@ -56,51 +53,6 @@ public class AssignmentFactory {
 
     public static Direct createDirect(String sourceRef) {
         return new Direct( sourceRef );
-    }
-
-    public static MethodReference createFactoryMethod( Type returnType, MappingBuilderContext ctx ) {
-        MethodReference result = null;
-        for ( SourceMethod method : ctx.getSourceModel() ) {
-            if ( !method.overridesMethod() && !method.isIterableMapping() && !method.isMapMapping()
-                    && method.getSourceParameters().isEmpty() ) {
-
-                List<Type> parameterTypes = MethodSelectors.getParameterTypes(
-                        ctx.getTypeFactory(),
-                        method.getParameters(),
-                        null,
-                        returnType
-                );
-
-                if ( method.matches( parameterTypes, returnType ) ) {
-                    if ( result == null ) {
-                        MapperReference mapperReference = findMapperReference( ctx.getMapperReferences(), method );
-                        result = new MethodReference( method, mapperReference, null );
-                    }
-                    else {
-                        ctx.getMessager().printMessage(
-                                Diagnostic.Kind.ERROR,
-                                String.format(
-                                        "Ambiguous factory methods: \"%s\" conflicts with \"%s\".",
-                                        result,
-                                        method
-                                ),
-                                method.getExecutable()
-                        );
-                    }
-                }
-            }
-        }
-        return result;
-    }
-
-    private static MapperReference findMapperReference( List<MapperReference> mapperReferences, SourceMethod method ) {
-        for ( MapperReference ref : mapperReferences ) {
-            if ( ref.getType().equals( method.getDeclaringMapper() ) ) {
-                ref.setUsed( !method.isStatic() );
-                return ref;
-            }
-        }
-        return null;
     }
 }
 
--- a/processor/src/main/java/org/mapstruct/ap/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/BeanMappingMethod.java
@@ -103,7 +103,7 @@ public class BeanMappingMethod extends MappingMethod {
             boolean mapNullToDefault =
                 MapperConfig.getInstanceOn( ctx.getMapperTypeElement() ).isMapToDefault( prism );
 
-            MethodReference factoryMethod = AssignmentFactory.createFactoryMethod( method.getReturnType(), ctx );
+            MethodReference factoryMethod = ctx.getMappingResolver().getFactoryMethod( method, method.getResultType() );
             return new BeanMappingMethod( method, propertyMappings, factoryMethod, mapNullToDefault );
         }
 
--- a/processor/src/main/java/org/mapstruct/ap/model/IterableMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/IterableMappingMethod.java
@@ -124,8 +124,7 @@ public class IterableMappingMethod extends MappingMethod {
             boolean mapNullToDefault
                 = MapperConfig.getInstanceOn( ctx.getMapperTypeElement() ).isMapToDefault( prism );
 
-            MethodReference factoryMethod = AssignmentFactory.createFactoryMethod( method.getReturnType(), ctx );
-
+            MethodReference factoryMethod = ctx.getMappingResolver().getFactoryMethod( method, method.getResultType() );
             return new IterableMappingMethod(
                     method,
                     assignment,
--- a/processor/src/main/java/org/mapstruct/ap/model/MapMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/MapMappingMethod.java
@@ -141,7 +141,7 @@ public class MapMappingMethod extends MappingMethod {
             boolean mapNullToDefault =
                 MapperConfig.getInstanceOn( ctx.getMapperTypeElement() ).isMapToDefault( prism );
 
-            MethodReference factoryMethod = AssignmentFactory.createFactoryMethod( method.getReturnType(), ctx );
+            MethodReference factoryMethod = ctx.getMappingResolver().getFactoryMethod( method, method.getResultType() );
 
             keyAssignment = new LocalVarWrapper( keyAssignment, method.getThrownTypes() );
             valueAssignment = new LocalVarWrapper( valueAssignment, method.getThrownTypes() );
--- a/processor/src/main/java/org/mapstruct/ap/model/MappingBuilderContext.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/MappingBuilderContext.java
@@ -94,6 +94,17 @@ public class MappingBuilderContext {
                                        String targetPropertyName, String dateFormat, List<TypeMirror> qualifiers,
                                        String sourceReference);
 
+        /**
+         * returns a no arg factory method
+         *
+         * @param mappingMethod target mapping method
+         * @param targetType return type to match
+         *
+         * @return a method reference to the factory method, or null if no suitable, or ambiguous method found
+         *
+         */
+        MethodReference getFactoryMethod(Method mappingMethod, Type targetType);
+
         Set<VirtualMappingMethod> getUsedVirtualMappings();
     }
 
@@ -190,4 +201,5 @@ public class MappingBuilderContext {
     public Set<VirtualMappingMethod> getUsedVirtualMappings() {
         return mappingResolver.getUsedVirtualMappings();
     }
+
 }
--- a/processor/src/main/java/org/mapstruct/ap/model/source/selector/InheritanceSelector.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/source/selector/InheritanceSelector.java
@@ -20,7 +20,6 @@ package org.mapstruct.ap.model.source.selector;
 
 import java.util.ArrayList;
 import java.util.List;
-import javax.lang.model.type.TypeMirror;
 
 import org.mapstruct.ap.model.common.Parameter;
 import org.mapstruct.ap.model.common.Type;
@@ -37,11 +36,14 @@ public class InheritanceSelector implements MethodSelector {
     public <T extends Method> List<T> getMatchingMethods(
         Method mappingMethod,
         List<T> methods,
-        Type parameterType,
-        Type returnType,
-        List<TypeMirror> qualifiers,
-        String targetPropertyName
+        Type sourceType,
+        Type targetType,
+        SelectionCriteria criteria
     ) {
+
+        if ( sourceType == null ) {
+            return methods;
+        }
 
         List<T> candidatesWithBestMatchingSourceType = new ArrayList<T>();
         int bestMatchingSourceTypeDistance = Integer.MAX_VALUE;
@@ -50,7 +52,7 @@ public class InheritanceSelector implements MethodSelector {
         for ( T method : methods ) {
             Parameter singleSourceParam = method.getSourceParameters().iterator().next();
 
-            int sourceTypeDistance = parameterType.distanceTo( singleSourceParam.getType() );
+            int sourceTypeDistance = sourceType.distanceTo( singleSourceParam.getType() );
             bestMatchingSourceTypeDistance =
                 addToCandidateListIfMinimal(
                     candidatesWithBestMatchingSourceType,
--- a/processor/src/main/java/org/mapstruct/ap/model/source/selector/MethodSelector.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/source/selector/MethodSelector.java
@@ -19,7 +19,6 @@
 package org.mapstruct.ap.model.source.selector;
 
 import java.util.List;
-import javax.lang.model.type.TypeMirror;
 
 import org.mapstruct.ap.model.common.Type;
 import org.mapstruct.ap.model.source.Method;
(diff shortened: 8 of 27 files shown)


sjaakd:
> I don't understand the two-directional source type matching. Only types that are assignable should be checked at this point, so the result of distanceTo should always be >=0. According to the Javadoc, the method returns -1, if (and only if) the type is not assignable.

Ok. Agreed. Flipping result and target around fixes that issue: like this `method.getResultType().distanceTo( targetType )`


> Back when I introduced the inheritance- and best-matching stuff for the property mapping methods, Gunnar and I discussed whether or not to check for distances on the return type as well. Although my first instinct was "sure, we need that", I found some examples for myself were it wouldn't be clear in which direction the "best" match would try to optimize. Exact match only? The most special type? Or the type that is actually the closest? Relying only on the source parameters made the most sense for me, reporting an ambiguity-error otherwise. And now we have the qualifiers as well.

I don't quite understand this argument. Why would different rules apply for source iso target? 

And: what's the difference between 'closest' and most 'special'? Isn't that the same? 

I admit, it gets a bit more hairy when both source and target distance are considered. But that's probably explainable in the documentation: 'Source distance first, target distance next'. 

For myself, I need this for factory methods. After all, returning the type with the closest distance actually makes sense here. So I could check if there's a source type available.

Qualifiers are currently only in place for mapping methods, not for factory methods. Thats a gap that can be filled pretty soon now.

By the way: did you check the unit test (`org.mapstruct.ap.test.selection.inheritance`)?

There's no check on source type. But I could disable the `InheritanceTargetSelector` for resolving non-factory methods by checking the existence of a source type.

agudian:
> And: what's the difference between 'closest' and most 'special'? Isn't that the same?

With "closest", I mean the variant with the shortest hierarchy towards the other type, with "most special", I mean the variant with the longest distance. E.g. for MotorizedVehicle, the class Car (extends MotorizedVehicle) would be closer, but the class Sedan (extends Car) would be more special.

> I admit, it gets a bit more hairy when both source and target distance are considered. But that's probably explainable in the documentation: 'Source distance first, target distance next'.

Imagine there is an additional class MotorBike (extends MotorizedVehicle) and there are two mapping methods defined: ```toSedan(VehicleDto):Sedan``` and ```toBike(VehicleDto):MotorBike```. I think letting MapStruct choose either one of them for a mapping from VehicleDto to MotorizedVehicle would be wrong - instead, we should throw an ambiguous mapping error as we do it now. My point is: I think there are many examples were choosing property mapping methods based on the return type would lead to wrong or undesired mappings, and so we shouldn't make use of the target types at all. Although I must admit: I don't know if that would really happen in real usage scenarios.

> For myself, I need this for factory methods. After all, returning the type with the closest distance actually makes sense here.

Right. I'm with you on that. This, and the addition of qualifier support is what I meant with "pimp up the selection of the factory methods" :+1:

> There's no check on source type. But I could disable the InheritanceTargetSelector for resolving non-factory methods by checking the existence of a source type.

Perhaps we should add that information to SourceMethod - the MethodRetrievalProcessor already makes that check if a method is a mapping method, a factory method, or (on the #84 branch) if it is  a LifecycleMethod (After/Before mapping). We could then reuse that information in the method selectors. Perhaps in form of an enum, so that it's easier for me to add the qualifier support for After/Before mapping methods also... in case they ever come... :wink:....

sjaakd:
> With "closest", I mean the variant with the shortest hierarchy towards the other type, with "most special", I mean the variant with the longest distance. E.g. for MotorizedVehicle, the class Car (extends MotorizedVehicle) would be closer, but the class Sedan (extends Car) would be more special.

Ah.. got your point now.

> With "closest", I mean the variant with the shortest hierarchy towards the other type, with "most special", I mean the variant with the longest distance. E.g. for MotorizedVehicle, the class Car (extends 
> Imagine there is an additional class MotorBike (extends MotorizedVehicle) and there are two mapping methods defined: toSedan(VehicleDto):Sedan and toBike(VehicleDto):MotorBike. I think letting MapStruct choose either one of them for a mapping from VehicleDto to MotorizedVehicle would be wrong - instead, we should throw an ambiguous mapping error as we do it now. My point is: I think there are many examples were choosing property mapping methods based on the return type would lead to wrong or undesired mappings, and so we shouldn't make use of the target types at all. Although I must admit: I don't know if that would really happen in real usage scenarios.

Hmm. your totally right. I considered a single inheritance hierarchy up to now. But forks - as is the case with Motorbike and Sedan - I did not consider. However, we should be able to detect such forks in the selection process and disqualify the `InheritanceTargetSelector` results (still leading to an ambiguous mapping method result). WDYT?

So, where does this leave us on the factory method discussion? After considering your argument, does it make a difference whether or not we have a source parameter? So if we need to construct a MotorizedVehicle, the Sedan and the MotorBike both equally qualify, while the MotorBike is selected based on distance, right? So perhaps we do need this fork-detection anyway... Or, alternatively only prefer spot-on matches.


> Perhaps we should add that information to SourceMethod - the MethodRetrievalProcessor already makes that check if a method is a mapping method, a factory method, or (on the #84 branch) if it is a LifecycleMethod (After/Before mapping). We

Agreed.. Shall we take that one in a follow on PR (just to keep stuff clean?)

agudian:
> So, where does this leave us on the factory method discussion? After considering your argument, does it make a difference whether or not we have a source parameter? So if we need to construct a MotorizedVehicle, the Sedan and the MotorBike both equally qualify, while the MotorBike is selected based on distance, right? So perhaps we do need this fork-detection anyway... Or, alternatively only prefer spot-on matches.

Hmm, I'm not really sure. The idea with the fork detection is good, and I think _if_ we compare type distances between different return types, then we should do it only if all available types belong to the same hierarchy. As you say, anything else wouldn't make any sense.

Question is: should we do it? Or keep it simple and say that either there must be only one matching method, or a qualifier needs to be used (which would be a separate PR, I guess)?

@gunnarmorling: any opinion?

> > Perhaps we should add that information to SourceMethod - the MethodRetrievalProcessor already makes that check if a method is a mapping method, a factory method, or (on the #84 branch) if it is a LifecycleMethod (After/Before mapping).

> Agreed.. Shall we take that one in a follow on PR (just to keep stuff clean?)

The one that introduces the qualifier support for factory methods would be a good place... :wink:

gunnarmorling:
> @gunnarmorling: any opinion?

Not yet; Still need to wrap my head around the current state of discussion :)

gunnarmorling:
Could you give an example type hierarchy which currently raises an ambiguity error but would be handled unambigously with the proposed fix?

sjaakd:
> Could you give an example type hierarchy which currently raises an ambiguity error but would be handled unambigously with the proposed fix?

See #385. I'm dealing with generated JAXB object factories. I can't change the XSD that they are created from, they are given

sjaakd:
As an addition: I'll change the order of the filters. I think that Qualifiers and selection on JAXB annotations should have their chance first :smile: . The consequence will be that regardless of the source / and target inheritance Qualifiers / selection on JAXB will take precedence and pre-select methods.

agudian:
> As an addition: I'll change the order of the filters. I think that Qualifiers and selection on JAXB annotations should have their chance first :smile: . The consequence will be that regardless of the source / and target inheritance Qualifiers / selection on JAXB will take precedence and pre-select methods.

Great, I was about to suggest the same after we finish the current discussion :+1:

sjaakd:
@agudian : There's one more thing I'm considering: selecting a mapping / factory method regardless forks if the match is spot-on (exact). WDYT?

agudian:
> There's one more thing I'm considering: selecting a mapping / factory method regardless forks if the match is spot-on (exact). WDYT?

Right, a distance of 0 should win, no matter if there are other methods with distance > 0 and possibly different ancestors. Sounds very reasonable to me and would otherwise create undesired errors (you would want to have factory methods for MotorizedVehicle, MotorBike and Sedan in case you have properties with all those types MotorizedVehicle, MotorBike and Sedan)...

agudian:
(btw, my suggestion would be that we use the assignee-field for pull-requests only to make clear that someone wants to do the review. WDYT?)

sjaakd:
> (btw, my suggestion would be that we use the assignee-field for pull-requests only to make clear that someone wants to do the review. WDYT?)

Ok.. Good idea..

gunnarmorling:
> (btw, my suggestion would be that we use the assignee-field for pull-requests only to make clear that someone wants to do the review. WDYT?)

+1

gunnarmorling:
@sjaakd, Could you provide an example with classes and how the method selection for these is affected by this change? Or just point me to the right classes in the tests? Thx!

sjaakd:
checkout: `org.mapstruct.ap.test.selection.inheritance.FruitMapper`, the inheritance relation between `Fruit`, `Apple` and `GoldenDelicious` and the generated mapper implementation: selection of the proper factory methods.

Note: mapping method and factory method selection is harmonized (the same) by this PR.

gunnarmorling:
Ok, I haven't fully caught up with the complete discussion, but I understand you propose to take the target type into account when selecting a mapping (or factory) method as well now, right? We decided purposefully against it back then, but maybe it's time to reconsider. The idea was to emulate Java's method dispatch as closely as possible, and that's based on parameters only.

sjaakd:
> Op 7 jan. 2015 om 23:24 heeft Gunnar Morling <notifications@github.com> het volgende geschreven:
> 
> Ok, I haven't fully caught up with the complete discussion, but I understand you propose to take the target type into account when selecting a mapping (or factory) method as well now, right?
> 

Right! 

I use JAXB generated object factories. These generate factory methods with inheritance relations on the return types. Can't change those and neither apply qualifiers. 
> We decided purposefully against it back then,
> 

I know 😄
> but maybe it's time to reconsider. The idea was to emulate Java's method dispatch as closely as possible, and that's based on parameters only.
> 
I realize that, but in regular Java coding you have the methodname as well as discriminating factor. The methodselection mechanism doesn't .

gunnarmorling:
http://yuml.me/diagram/scruffy/class/%5BFruit%5D%5E-%5BApple%5D,%20%5BFruit%5D%5E-%5BBanana%5D,%20%5BApple%5D%5E-%5BGoldenDelicious%5D

sjaakd:
Right !

Verstuurd vanaf mijn iPhone

gunnarmorling:
Musing about this for a while now, I come to realize that the factory method ambiguity question is only one part of a larger issue. The same question arises if there are no factory methods but a generated mapping method is declared to return a super-type with one or more sub-types. Which of one should be instantiated in this case?

That being said, I'm not sure the proposed strategy for selecting the type to be instantiated (or obtained from the factory) is the way to go. It's not necessarily true that the "closest" type is the "right one", I don't think we can make that decision ourselves; Really it depends on the specific requirements and semantics of specific use cases. Rather my inclination is that the user should specify a hint what the implementation type should be.

Here is what I propose: Add an optional annotation `@BeanMapping` (consistent with `IterableMapping` and `MapMapping`) which allows to give that "hint" either as class object or qualifier:

    @BeanMapping(targetType=Apple.class)
    // OR like so:
    @BeanMapping(qualifiedBy=TropicalFruits.class)
    Fruit map(FruitDto source);

Note that this also allows the user to resolve the ambiguity in `ErroneousFruitMapper` (It's IMO a legit mapper definition, so we must provide a way to make this case work).

If we have this mechanism in place (which we IMO need either way to support `ErroneousFruitMapper`) we may think in a second step of alternative resolution strategies a user can "opt-in" (e.g. instantiating the declared return type). But I'd like us to give the explicit configuration a shot and see how it flies on real models.

gunnarmorling:
>  The same question arises if there are no factory methods but a generated mapping method is declared to return a super-type with one or more sub-types.

Hum, no, I think I wasn't right there. If there are no factory methods, we wouldn't even know all the sub-types (one cannot search for all sub-classes), so I think in this case we should instantiate the declared return type.

sjaakd:
> That being said, I'm not sure the proposed strategy for selecting the type to be instantiated (or obtained from the factory) is the way to go. It's not necessarily true that the "closest" type is the "right one", I don't think we can make that decision ourselves; Really it depends on the specific requirements and semantics of specific use cases. Rather my inclination is that the user should specify a hint what the implementation type should be.

Agreed. 

But (as always :smile:): I would propose to implement such mechanism on top of what is proposed here. The idea is that MapStruct works in the majority of the cases out-of-the box. 

Implementation wise: if you look at the order of filters, the  `@BeanMapping(targetType=Apple.class)` would take precedence over the `InheritanceTargetSelector `.

> Here is what I propose: Add an optional annotation @BeanMapping (consistent with IterableMapping and MapMapping) which allows to give that "hint" either as class object or qualifier:
> 
> @BeanMapping(targetType=Apple.class)
> // OR like so:
> @BeanMapping(qualifiedBy=TropicalFruits.class)
> Fruit map(FruitDto source);
> 
> Note that this also allows the user to resolve the ambiguity in ErroneousFruitMapper (It's IMO a legit mapper definition, so we must provide a way to make this case work).

My preference: `@BeanMapping(targetType=Apple.class)`.  The other option (qualifiedBy) would only confuse the user, since that mechanism works in an entirerly different way.

sjaakd:
One more addition: 

If we introduce `@BeanMapping(targetType=Apple.class)`  to select the proper object to return for the entire bean mapping methods,

And we keep the common selection mechanism (for factory methods and for mapping methods), then:

1. We might want to add `qualifiedBy`  to the proposed bean mapping` like this: @BeanMapping(targetType=Apple.class, qualifiedBy=<qualifier>)` to select the proper factory method

2. We might want to add `targetType` to the standard mapping like this: `@Mapping(targetType=Apple.class, qualifiedBy=<qualifier>)`  to select the proper mapping method.

And establish some kind of useful symmetry. WDYT?

sjaakd:
@gunnarmorling, I introduced the proposed `@BeanMapping(targetType=Apple.class)` mapping method. There are some things todo (like my WDYT above). I would propose to handle those in a separate PR.

Also: if there's no factory method, we create the return type in the bean mapping signature, so the new annotation is purely used for factory selection.

Can you have a look?

gunnarmorling:
@sjaakd, let my try and sum up where we are. If I got anything wrong or miss something, please let me know.

> I introduced the proposed @BeanMapping(targetType=Apple.class)

Just to be sure, what's the current approach now, does the *declared return type* of a mapping method play a role for factory mapping method selection or not? I don't think it should, for the reasons mentioned before. My preference is to only consider the parameters for mapping (or factory) method selection, and in doubt apply the additional filter information from `@BeanMapping`.

>  The other option (qualifiedBy) would only confuse the user, since that mechanism works in an entirerly different way.

> We might want to add qualifiedBy to the proposed bean mapping

So you actually came to like `@BeanMapping#qualifiedBy()` I do like it as it nicely fits into the existing ways of mapping method selection: If no method cannot be found unambigiously by means of its parameters, qualifier(s) are applied as an additional means of filtering the candidate list. ` Maybe `@BeanMapping#factoryMethodQualifiedBy()` makes it easier to understand?

In that light, `targetType()` seems a bit more irregular to me. But then there probably is the case where on is not in control of the factory methods (JAXB?) so one cannot put qualifiers to their definitions. For this case `targetType()` would provide an alternative approach. So we'd probably need both. I would rename `targetType()` into `resultType()`.

> We might want to add targetType to the standard mapping like this

Yes, that likely make sense. But let's do it separately, that PR is already huge by itself :) Could you open a new issue for that one?

> Also: if there's no factory method, we create the return type in the bean mapping signature, so the new annotation is purely used for factory selection.

Thinking about it, couldn't we also instantiate the given type (if present) in the non-factory case? It's not strictly required but I can see it may be useful in some cases and I don't see a reason for not doing it.

gunnarmorling:
@sjaakd:

> For the source-type, the fit was not applied in two directions. For that purpose the absolute (type) distance is used iso distance.

I'm not sure what this really means. Could you give an example? To move forward, can we split it up into two PRs, one handling the things we discussed around selection based on the return type (provided we came to an agreement there) and another one for this issue? It seems not strictly related, but then it may be correlated in terms of code changes / commits?

Btw. let's always strive for creating small and independent PRs (also doing refactorings separately, e.g. a refactoring PR laying grounds for subsequent "actual" changes). This makes it much easier to review and helps to keep the discussions manageable. Thx!

processor/src/main/java/org/mapstruct/ap/model/source/selector/InheritanceTargetSelector.java
```diff
@@ -0,0 +1,107 @@
...
+
+import org.mapstruct.ap.model.common.Type;
+import org.mapstruct.ap.model.source.Method;
+
+/**
+ * Selects on inheritance distance, e.g. the amount of inheritance steps from the target type.
+ *
+ * @author Sjaak Derksen
+ */
+public class InheritanceTargetSelector implements MethodSelector {
```
gunnarmorling: Ah, so apparently the declared return type still is used for factory mapping method selection? IMO we cannot really do that as we don't know which one is the "right" implementation type. Thus we should only apply the additional filtering information given via `@BeanMapping`.

gunnarmorling: Btw. I checked with CDI, and it's the same for "producer methods" there, if there are several producers with an assignable return type, there is no form of "clever" resolution logic but one needs to resolve the ambiguity via qualifiers. Which makes me think we should do the same.

sjaakd:
> Just to be sure, what's the current approach now, does the declared return type of a mapping method play a role for factory mapping method selection or not?

It does as last resort. And the choice can be overridden by whatever the user wants by means of the `@BeanMapping(targetType=Apple.class)` 

I believe the power of MapStruct is that the user does not need to specify something when its not needed. As you pointed out: ' It's not necessarily true that the "closest" type is the "right one"'. But if it covers most of the cases and the targetType construction can be used to override its choice...


> So you actually came to like @BeanMapping#qualifiedBy()?

I'm not sure if you follow me on this one :smile: . 

1. With `@BeanMapping` in essence you determine criteria for selecting factory methods. 

2. With `@Mapping`  you determine the criteria for selecting mapping methods (so with arg). 

I consider target type selections as rather different than qualifiers. So if we introduce the target type selection ( `@BeanMapping#targetType()` ) we could do likewise for property mapping ( `@Mapping#targetType()` ). And we could do the opposite thing as well: introduce `@BeanMapping#qualifiedBy()`.  Note: it comes for (almost) free, since we use the same selector mechanism and gives a consistent story for the user.

> But then there probably is the case where on is not in control of the factory methods (JAXB?) so one cannot put qualifiers to their definitions.

Spot on.. Its exactly that.. 

> So we'd probably need both. I would rename targetType() into resultType().

I'm not sure. When later we apply it to ( `@Mapping#targetType()` ) it fits in nicely with ( `@Mapping#target()` ). WDYT?

> Yes, that likely make sense. But let's do it separately, that PR is already huge by itself :) Could you open a new issue for that one?

Of course.. But lets first finalize the discussions around this one.

> Thinking about it, couldn't we also instantiate the given type (if present) in the non-factory case? It's not strictly required but I can see it may be useful in some cases and I don't see a reason for not doing it.

Thought about that too.. Perhaps also stuff for a different PR. Need to think a bit more on the consequences.


> For the source-type, the fit was not applied in two directions. For that purpose the absolute (type) distance is used iso distance.

Forget bout this one. @agudian also commented on it and it was a misunderstanding from my side. It is no longer part of the code. 

> Btw. let's always strive for creating small and independent PRs (also doing refactorings separately, e.g. a refactoring PR laying grounds for subsequent "actual" changes). This makes it much easier to review and helps to keep the discussions manageable. Thx!

Ok.. I did use several commits this time :smile: .

processor/src/main/java/org/mapstruct/ap/model/source/selector/InheritanceTargetSelector.java
sjaakd: You're right.. See my motivation below.  

We can take the `InheritanceTargetSelector` filter out again if you really think its the better choice.

sjaakd:
@gunnarmorling: on the workings of the selector mechanism:

* Selectors are cascaded
* In a cascade, the output of the first selector is the input of the second. The output of the second selector in the input of the third. And so on ( output selector N is input selector N+1 ).
* Input the first selector (`TypeSelector`) are all refered mapping methods and factory methods.
* The output of the `TypeSelector` are all methods that match to the desired signature (source type and result type).

Now asume (we do not consider result type for the simpicity of the example).
* we have specified a `qualifiedBy` that selects a method with desired source type  `Banana`
* all refered mapping methods have also a mapping method with source type 'Fruit'
* the order of the selectors is 1:`TypeSelector` 2:`InheritanceSelector` 3:`QualifierSelector`

Result
* The `TypeSelector` will offer both the mapping methods, the one with `Banana` and the one with `Fruit` to the `InheritanceSelector`
* The `InheritanceSelector` will disqualify the method with `Banana` based on source type distance (after all, the distance of `Banana`is 1 compared to `Fruit` 0 )
* The `QualifierSelector` only gets offered a method that does not carry a Qualifer anymore.

Result == no match.

I reckon the user intented a match here. So I flipped the order of the selectors according the following rules:
1. `TypeSelector` first
2. User defined choices (Qualifier, ResultType) first
3. Configured mechanisms (`XmlElementDesc`) next
4. Then automated mechanisms ( on inheritance distance ).

sjaakd:
@gunnarmorling I processed our discussion of last evening. Please have a look.

gunnarmorling:
@sjaakd Finally got to that one and merged it. Could you send a doc PR for the new `@BeanMapping` annotation as well? Many thanks :clap: !

sjaakd:
I'll do so! In coming weeks I'll allign 'BeanMapping' with 'Mapping'. I'll also push the 'InheritanceTargettypeSelector' solution including unit test on a separate branch for whenever a case occurs were we need it. Thanks for reviewing / merging.

PR #396, opened 2014-12-23
#224 checks for given dateFormat in DefaultConversionContext

timoe:
DefaultConversionContext now verifies the dateFormat if given. Based on source/target type the formatter is derived and the dateFormat is check against that.


agudian:
Saw it just now: please use rebase and fast-forward merges the next time. We usually want to stay clear of merge-commits... :wink:

PR #401, opened 2014-12-29
fix nested property mappings with Eclipse JDT compiler

### LINKED ISSUE #400: nested source generating invalid code
//Mapper
        PersonMapper INSTANCE = Mappers.getMapper(PersonMapper.class);
	@Mappings({@Mapping(source="child.id",target="name")}
	)
	abstract People personToPeople(Person person);

    //Generated code
    private String personChildId(Person person) {

        String string = null;
        if ( person != null ) {
            Person child = person.public hestwood.mapstruct.Person getChild() ;
            if ( child != null ) {
                String id = child.public java.lang.String getId() ;
                if ( id != null ) {
                    string = id;
                }
            }
        }
        return string;
    }

agudian:
extended the integration test accordingly.

Fixes #400.

### DIFF
--- a/processor/src/main/resources/org.mapstruct.ap.model.NestedPropertyMappingMethod.ftl
+++ b/processor/src/main/resources/org.mapstruct.ap.model.NestedPropertyMappingMethod.ftl
@@ -24,7 +24,7 @@
         return null;
     }
     <#list propertyEntries as entry>
-    <@includeModel object=entry.type/> ${entry.name} = <#if entry_index == 0>${sourceParameter.name}.${entry.accessor}<#else>${propertyEntries[entry_index-1].name}.${entry.accessor}</#if>;
+    <@includeModel object=entry.type/> ${entry.name} = <#if entry_index == 0>${sourceParameter.name}.${entry.accessorName}()<#else>${propertyEntries[entry_index-1].name}.${entry.accessorName}()</#if>;
     if ( ${entry.name} == null ) {
         return null;
     }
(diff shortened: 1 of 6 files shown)


processor/src/main/resources/org.mapstruct.ap.model.NestedPropertyMappingMethod.ftl
```diff
@@ -24,7 +24,7 @@
         return null;
     }
     <#list propertyEntries as entry>
-    <@includeModel object=entry.type/> ${entry.name} = <#if entry_index == 0>${sourceParameter.name}.${entry.accessor}<#else>${propertyEntries[entry_index-1].name}.${entry.accessor}</#if>;
+    <@includeModel object=entry.type/> ${entry.name} = <#if entry_index == 0>${sourceParameter.name}.${entry.accessorName}()<#else>${propertyEntries[entry_index-1].name}.${entry.accessorName}()</#if>;
```
gunnarmorling: Interesting, I'm wondering why this was not always a problem? I guess I still must be missing something.

gunnarmorling:
@agudian So was this a general issue, or only with Eclipse? Looks good to me. Could you rebase it and apply? Thx!

agudian:
> @agudian So was this a general issue, or only with Eclipse?

I'd say it's a general issue that was unnoticed in Oracle JDK due to the way toString() is implemented in the com.sun.*-Implementation of ```ExecutableElement``` ("getStringProperty()" on Oracle JDK vs "public java.lang.String getStringProperty()" in Eclipse JDT). The format is not specified by the API.

Rebased and pushed. Thanks for the review!

Issue #271, opened 2014-07-15
Improve error message in case an invalid expression is given


gunnarmorling:
When giving an invalid expression via `@Mapper#expression()` (i.e. one which is not in the form "java(&lt;EXPRESSION&gt;)"), the following error is reported:

> No property named "" exists in source parameter(s).]]

Instead we should clearly indicate the problem. Also the JavaDocs of `expression()` should explicitly state, that "java(&lt;EXPRESSION&gt;)" is the only form currently supported.

gunnarmorling:
Messages are good already, just improved the JavaDocs a bit.

PR #402, opened 2014-12-29
#271 example for expression with some errors. looks good so far

timoe:
Hey @gunnarmorling. Can  you please take a look at the tests I did. Issue #271 states that we need better error messages but those I played around with look good from my point of view - so, can you please explain what you meant by this issue? Or did this improve over the time, so that this is not longer an issue?


gunnarmorling:
Hi @timoe, sorry for the late reply :/

I just checked with your tests and you are right, the error messages look good now. I vaguely remember there was some clean-up in this area. I just pushed one commit to make the JavaDocs of `expression()` a bit more meaningful.

Going to close this PR therefore. Thanks nevertheless, if there is anything else you'd like to hack on, just assign to yourself. There is also an Eclipse plug-in now which is work in progress (https://github.com/mapstruct/mapstruct-eclipse), just in case that's your kind of thing :)

PR #403, opened 2014-12-30
allow autoboxing for primitive return types of property mapping methods to non-primitive properties

### LINKED ISSUE #398: For Boolean property, aToString(A):String is favoured instead of aToBool(A):boolean as property mapping method
if  there is method "String aToString(A a)"  , MapStruct would use it to convert A to boolean even if there is a more specific method "boolean a2bool(boolean)" .

 I think this is not reasonable.

agudian:
Fixes #398.


sjaakd:
@agudian Thanks!

Issue #407, opened 2015-01-03
NestedProperty: Primitives do not work


shasait:
Mapstruct produces the following code for the adapted integration test:

    private int sourceForNestedIntValue(Source source) {

        if ( source == null ) {
            return null;   <<< what to return here? 0 or some customizable value?
        }
        SomeIntType forNestedInt = source.getForNestedInt();
        if ( forNestedInt == null ) {
            return null;   <<< what to return here? 0 or some customizable value?
        }
        int value = forNestedInt.getValue();
        if ( value == null ) {   <<< null check must be skipped
            return null;
        }
        return value;
    }

shasait:
I will work on a fix (PR)...

agudian:
@shasait: Merged your PR. Thanks alot! :+1:

PR #408, opened 2015-01-03
#407 Fix primitive handling for nested properties

shasait:


agudian:
Your commit contains changes from #400 which are not really necessary for fixing this issue, right? Could you drop them out?

And using or extending a Unit-Test from within the processor module would be favourable. The integration tests are more bulky and harder to execute from within the IDE (you always need to mvn install on the command line first, etc) - so when ever possible, we try to create reproducer tests and tests for new features as Unit-Tests.

But I like the idea how you solved the Problem itself! :+1:

shasait:
Hmmm - wanted to squash both commits into one so the PR stays small - but it doesn't look correct

shasait:
ahh - because I reverted 400 it now looks this way... should be correct then.

shasait:
Yes - I saw this in the documentation, but my problem was, that EGit on squash opens a dialog were lines starting with "#" are comments therefore I was littlebit lost - but next time I know better and will simple change the message afterwards...

Issue #405, opened 2014-12-31
object factory generate invalid code inside explicitly declared list mapping function


hestwood:
If define below list mapping method in mapper:
abstract List<A> bListToAList(List<B> b)

Then use entityFactory:

<T> T createEntity(Class<T> entityClass){ .. }

Generated code inside bListToAList would be:

factory.createEntity( .class );

hestwood:
Below is the code to reproduce the problem , please kindly check.

```Java
public class People {
	private String name;
	private List<People> children;
	
	public List<People> getChildren() {
		return children;
	}

	public void setChildren(List<People> children) {
		this.children = children;
	}

	public String getName() {
		return name;
	}

	public void setName(String name) {
		this.name = name;
	}

}

public class Person {
	private String id;
	private List<Person> children;
	
	public String getId() {
		return id;
	}
	
	public void setId(String id) {
		this.id = id;
	}

	public List<Person> getChildren() {
		return children;
	}

	public void setChildren(List<Person> children) {
		this.children = children;
	}
}

public class EntityFactory {
	public static <T> T createEntity(@TargetType Class<T> entityClass ){
		return null;
	}
}

@Mapper(uses={EntityFactory.class})
public abstract class PersonMapper {
	
	PersonMapper INSTANCE = Mappers.getMapper(PersonMapper.class);
	@Mappings({@Mapping(source="id",target="name")}
	)
	abstract People personToPeople(Person person);
	
	abstract List<People> personListToPeopleList(List<Person> children);
	
}

//generated code
public class PersonMapperImpl extends PersonMapper {

    private final EntityFactory entityFactory = new EntityFactory();
    @Override
    List<People> personListToPeopleList(List<Person> children) {
        if ( children == null ) {
            return null;
        }

        List<People> list = entityFactory.createEntity( .class );

        for ( Person person : children ) {
            list.add( personToPeople( person ) );
        }

        return list;
    }

    @Override
    People personToPeople(Person person) {
        if ( person == null ) {
            return null;
        }

        People people = entityFactory.createEntity( People.class );

        people.setName( person.getId() );
        people.setChildren( personListToPeopleList( person.getChildren() ) );


        return people;
    }

}
```

hestwood:
I understand that it is not necessary to declare "personListToPeopleList" but for my case this method has to be declared so it could be referenced in other manual mapper method int he mapper class.

sjaakd:
@hestwood Thanks.

[...]

PR #409, opened 2015-01-05
#405 Fixing factory methods, not using raw type for @TargetType and supe...

sjaakd:
...rfluous implementation classes imports

1. The model for methods was not always applied correctly to factory methods. Hence, the result type was not passed along.
2. Non Raw types were passed as @TargetType, compilation error.
3. Superflous imports for implementation types when using factory methods in iterable and map mappings. Responsibility of importing impllementation type has been made dependent on whether or not a 'new' is required (no factory method)

### DIFF
--- a/processor/src/main/resources/org.mapstruct.ap.model.IterableMappingMethod.ftl
+++ b/processor/src/main/resources/org.mapstruct.ap.model.IterableMappingMethod.ftl
@@ -107,7 +107,7 @@
 <#macro iterableCreation>
     <@compress single_line=true>
         <#if factoryMethod??>
-            <@includeModel object=factoryMethod/>
+            <@includeModel object=factoryMethod targetType=resultType raw=true/>
         <#else>
             new
             <#if resultType.implementationType??>
--- a/processor/src/main/java/org/mapstruct/ap/model/GeneratedType.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/GeneratedType.java
@@ -164,8 +164,6 @@ public abstract class GeneratedType extends ModelElement {
             }
         }
 
-        addWithDependents( collection, typeToAdd.getImplementationType() );
-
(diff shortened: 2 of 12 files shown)


processor/src/main/resources/org.mapstruct.ap.model.IterableMappingMethod.ftl
```diff
@@ -107,7 +107,7 @@
 <#macro iterableCreation>
     <@compress single_line=true>
         <#if factoryMethod??>
-            <@includeModel object=factoryMethod/>
+            <@includeModel object=factoryMethod targetType=resultType raw=true/>
```
agudian: Adding ```raw=true``` is not necessary here and has no effect. Same for the MapMappingMethod. The only single place were we need to set ```raw=true``` is when adding ```<@includeModel ..>``` for a type to render. Actually, I have some clean-ups on that in PR #84.... If I had only known... :wink:

agudian:
I have only that remark on passing ```raw=true``` where it has no effect. If you change that and rebase, then please go ahead and push it... :+1:

sjaakd:
> I have only that remark on passing raw=true where it has no effect. If you change that and rebase, then please go ahead and push it...

I thought about not doing this.. But did you check `MethodReference.ftl`? 2 statements below the modified statement is `ext.raw`.. So I think it is possibly used. Anyway..

sjaakd:
merged

agudian:
> I thought about not doing this.. But did you check MethodReference.ftl? 2 statements below the modified statement is ext.raw.. So I think it is possibly used. Anyway..

I did see that as well, and it's not doing anything meaningful there as well... :wink:

sjaakd:
> I did see that as well, and it's not doing anything meaningful there as well...

I'll remove it in a next commit :smile:

Issue #417, opened 2015-01-11
qualified method is not used in mapping collections


natash:
I have noticed interesting befaviour that I don't know if it is intended or not. I will try to explain it with example:

    //first mapper
    @Mappings({ @Mapping(source = "game.players", target = "players", qualifiedBy = QualifyPlayerMapper.class) })
    TargetGame mapGame(Game game);
   
    //second mapper
    @Mappings({ @Mapping(source = "player.street", target = "addressStreet")})
    TargetPlayer mapPlayer(Player player);
       
    //some mapping method that I wanna use and it is qualiffied
    @QualifyPlayerMapper
    TargetPlayer mapPlayerMyWay(Player player);
   
    so in the generated class, in the playerCollectionToPlayerCollection method, is used  mapPlayer.
    In my opinion it should use the second method - mapPlayerMyWay
   
   
    Regards,
    Natalia

sjaakd:
@natash, can you include your definition of @QualifyPlayerMapper as well?

natash:
@Qualifier
@Target(ElementType.METHOD)
@Retention(RetentionPolicy.SOURCE)
public @interface QualifyPlayerMapper

sjaakd:
@natash , can you add a mapping (and see if it works):

```java
IterableMapping(qualifiedBy = QualifyPlayerMapper.class)
List<TargetPlayer> mapTargetPlayer(List<Player> Player);
```

Assumption: MapStruct does not delegate the `Qualifier` to the generated interable method. By making it explicit, you can add the qualifiedBy at the right level.

You can remove the complete `@Mappers` annotation from `TargetGame mapGame(Game game)`, since matching should take place on name (players==players).

gunnarmorling:
That's an interesting case; @sjaakd, I assume that when `@Mapping#qualifiedBy()` is given for a collection-typed property, that value is used to select the matching collection mapping method, right? 

Now in the case where the iterable mapping method is generated as a private helper method, should we apply the given qualifier to select the right element mapping method (I think that's what is suggested here).

I feel we should not so not to change the semantics of `qualifiedBy()` depending on whether the collection method is generated or not). As you say, one can achieve the required behavior by explicitly declaring the collection mapping method. This seems the more consistent appraoch to me, i.e. leave it as is. WDYT?

sjaakd:
[...]

PR #421, opened 2015-01-18
#417 Making dateFormat in @IterableMapping optional, adding checks for @...

sjaakd:
...IterableMapping and @MapMapping

* `@IterableMapping#dateFormat` is not mandatory anymore in `@IterableMapping`due to the introduction of `@IterableMapping#qualifiedBy`
* one of the parameters  `@IterableMapping#dateFormat` or  `@IterableMapping#qualifiedBy `needs to be specified. 
* the same is valid for `@MapMapping`

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/model/source/IterableMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/source/IterableMapping.java
@@ -19,9 +19,12 @@
 package org.mapstruct.ap.model.source;
 
 import java.util.List;
+import javax.annotation.processing.Messager;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
+import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.type.TypeMirror;
+import javax.tools.Diagnostic;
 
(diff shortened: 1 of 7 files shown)


processor/src/main/java/org/mapstruct/ap/model/source/IterableMapping.java
```diff
@@ -37,11 +40,21 @@
...
+    public static IterableMapping fromPrism(IterableMappingPrism iterableMapping, ExecutableElement method,
+                                            Messager messager) {
         if ( iterableMapping == null ) {
             return null;
         }
 
+        if ( iterableMapping.dateFormat().isEmpty() && iterableMapping.qualifiedBy().isEmpty() ) {
+            messager.printMessage(
+                Diagnostic.Kind.ERROR,
+                "'dateformat' and 'qualifiedBy' are both undefined in @IterableMapping, either define a "
```
agudian: Suggestion: let the message end with ", define at least one of them." - similar to your message for MapMapping. That way it's one spot less to modify in case we add new attributes to that annotation, and it's more consistent.

agudian:
I only have one suggestion, but otherwise it's all fine :+1:

sjaakd:
merged..

Issue #390, opened 2014-12-17
@Generated should list the maven version of the processor


sjaakd:
I'm dealing with a problem that I'm not sure what compiler plugin is used by Maven. Somehow I believe it deviates from the one used by my IDE.

To make sure, it would be very handy if the compiler plugin lists its version in the @Generated tag.

gunnarmorling:
+1 for adding the generator version to the generated output. An option for this seems reasonable.

The suggested "verbose" switch is something different IMO (I believe there is even an issue for it). But rather than adding comments to the generated code, we should create annotation processor diagnostics using the level NOTE, stating things such as candidate mapping methods and so on. It requires some testing though, e.g. I'm not sure how Maven deals with NOTE-level diagnostics.

sjaakd:
Done!

PR #422, opened 2015-01-18
#390 included maven coordinates as comments in generated code

sjaakd:
I've realized this feature by adding a manifest entry in the MANIFEST.MF file by means of the maven-jar-plugin. This is read by the annotation processor and included in the generated code.

Simple and effective :smile:

The full maven coordinates are added (groupId:artifactId:version).

One more remark: the `@Generated#comments` does not show up in the unit test (because the maven-jar-plugin is active at the end of the build).

### DIFF
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -45,8 +45,8 @@
         </dependency>
 
         <!-- Compile-only; Using "provided" scope as there is no such scope in Maven;
-             these dependencies are not required at runtime, only for prism generation
-             and tests -->
+        these dependencies are not required at runtime, only for prism generation
+        and tests -->
         <dependency>
             <groupId>com.jolira</groupId>
             <artifactId>hickory</artifactId>
@@ -90,6 +90,17 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                        </manifest>
+                    </archive>
+                </configuration>
+            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
--- a/processor/src/main/java/org/mapstruct/ap/model/GeneratedType.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/GeneratedType.java
@@ -18,11 +18,15 @@
  */
 package org.mapstruct.ap.model;
 
+import java.io.IOException;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Enumeration;
 import java.util.List;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.jar.Manifest;
 import javax.annotation.Generated;
 
 import org.mapstruct.ap.model.common.Accessibility;
@@ -36,6 +40,9 @@ import org.mapstruct.ap.model.common.TypeFactory;
  * @author Gunnar Morling
  */
 public abstract class GeneratedType extends ModelElement {
+
+    private static final String COMMENT_TAG = "Implementation-Version";
+    private static final String COMMENTS = initComment();
 
     private final String packageName;
     private final String name;
@@ -118,6 +125,10 @@ public abstract class GeneratedType extends ModelElement {
         return accessibility;
     }
 
+    public String getComments() {
+        return COMMENTS;
+    }
+
     @Override
     public SortedSet<Type> getImportTypes() {
         SortedSet<Type> importedTypes = new TreeSet<Type>();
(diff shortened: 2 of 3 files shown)


processor/src/main/java/org/mapstruct/ap/model/GeneratedType.java
```diff
@@ -118,6 +124,24 @@ public Accessibility getAccessibility() {
         return accessibility;
     }
 
+    public String getComments() {
```
agudian: I would suggest to statically hold the comment and initialize it only once (```private static final String COMMENT = initComment();```) - to save some CPU cycles... :wink:

agudian:
I was fiddling around with finding out some information about the compiler being used, as, for trouble-shooting, is equally interesting.
WDYT, should such information also find its way to the comment-Tag? Or into some explicitly enabled comment for debug-tracing?

processor/src/main/java/org/mapstruct/ap/model/GeneratedType.java
sjaakd: Done!

sjaakd:
> I was fiddling around with finding out some information about the compiler being used, as, for trouble-shooting, I find that equally interesting.
> WDYT, should such information also find its way to the comment-Tag? Or into some explicitly enabled comment for debug-tracing?

I think so.. :+1:  We've seen many problems in the compiler type (but I reckon I don't have to tell you :smile: ), so that would surely help in troubleshooting. Can you pick up this PR and add that somehow?

Funny thing: `@Generated` has a `String` array `value`, while `Comment` and `TimeStamp` are single values. Any idea why?

sjaakd:
> I was fiddling around with finding out some information about the compiler being used, as, for trouble-shooting, I find that equally interesting.

Ok.. I think I missed your point with 'fiddling'. I thought you meant that you got compiler info available in the processor somehow. What you mean is useing the `Build-Jdk` in the `MANIFEST.MF`, right? 

I can add that. No problem. But does that convey enough information for all possible JDK's? For `javac` for instance, it only adds a version: `1.7.0_55`

processor/pom.xml
```diff
@@ -92,6 +92,17 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <MapStruct-Generated-Comment>${project.groupId}:${project.artifactId}:${project.version}</MapStruct-Generated-Comment>
```
gunnarmorling: I don't think we need group/artifact id; when seeing the version number in `@Generated` that info can be derive. I'd just use the standard version number manifest header which should already be in there.

sjaakd: > which should already be in there.

I checked this out when I started with this PR. Our current manifest is pretty empty:
```java
Manifest-Version: 1.0
Archiver-Version: Plexus Archiver
Built-By: gunnar
Created-By: Apache Maven 3.0.4
Build-Jdk: 1.8.0_25
```
We could add: `<addDefaultImplementationEntries>true</addDefaultImplementationEntries>`. In that case the `MANIFEST.MF` will be extended with:
```java
Implementation-Title: ${project.name}
Implementation-Version: ${project.version}
Implementation-Vendor-Id: ${project.groupId}
Implementation-Vendor: ${project.organization.name}
Implementation-URL: ${project.url}
```
So, we could then use: `Implementation-Version`, extended with the `Build-Jdk` as proposed above in response to a remark of @agudian. WDYT?

gunnarmorling: > So, we could then use: Implementation-Version

+1

> extended with the Build-Jdk

But that's the JDK the given MapStruct version itself was built with, right? I don't think that helps (similar to group/artifact id, we can find out that info just by examining the JAR). It would be very useful though to add the name + version of the compiler being used to compile the project MapStruct is applied to. I don't know though how that can be determined.

sjaakd: So: the `@Generated#comments` will look like:
'Implementation-Version: 1.0.0-Beta3, Build-Jdk: 1.8.0_25'

sjaakd: > But that's the JDK the given MapStruct version itself was built with, right?

Grrrrrr. @$%@$..  you're right :smile:. @agudian , ideas?

sjaakd: Perhaps somehow via the environment properties?

agudian: That's what I meant with fiddling. I have some test code in place to examine where the APT-implementation comes from, and it would probably burn down to a combination of system-properties and jar file name (e.g. in case of the Eclipse implementation, I want to know which one as well)... Still need to figure out how to condense the information and not create too large strings. I already did some tests with JDKs 1.6 .. 1.9 and the Eclipse compiler bundled by felix, but not yet with the plain-vanilla Eclipse...

But anyway, Build-JDK does not help - as you found out yourself just now... :wink:

I would suggest you finish this PR and leave the issue open, I'll just add my stuff along side with it and create a new PR then.

sjaakd: > would suggest you finish this PR and leave the issue open, I'll just add my stuff along side with it and create a new PR then.

Ok.. 'Implementation-Version: 1.0.0-Beta3' it'll be ! I'll update and merge tonight.

Thanks for your input.

gunnarmorling: Cool, thx! Let's open a separate issue for adding the compiler info.

Issue #410, opened 2015-01-05
Static methods calls should not use a mapper instance


sjaakd:
See #405. Static methods currently use an instance of a mapper, but call the methods on the class.... which leads to problems for utility type of helpers which have typically a private constructor (e.g. due to check style, PMD or other demands).

So: 
1. MapStruct should call static methods on the class
2. If there's no reason to instantiate a mapper, e.g. no instance methods are called, the mapper should not be instantiated as private member of the mapper under construction.

gunnarmorling:
Interesting, never crossed my mind but you are right.

sjaakd:
Done..

PR #423, opened 2015-01-19
#410 refer to static methods via their class and not via their instance

sjaakd:


processor/src/main/resources/org.mapstruct.ap.model.DefaultMapperReference.ftl
```diff
@@ -18,4 +18,4 @@
      limitations under the License.
 
 -->
-private final <@includeModel object=type/> ${variableName} = <#if annotatedMapper>Mappers.getMapper( <@includeModel object=type/>.class );<#else>new <@includeModel object=type/>();</#if>
\ No newline at end of file
+<#if used>private final <@includeModel object=type/> ${variableName} = <#if annotatedMapper>Mappers.getMapper( <@includeModel object=type/>.class );<#else>new <@includeModel object=type/>();</#if></#if>
```
agudian: You also need to add that to AnnotationMapperReference.ftl... :)

processor/src/test/java/org/mapstruct/ap/test/references/statics/StaticsTest.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @author Sjaak Derksen
+ */
+@IssueKey( "410" )
+@WithClasses( { Beer.class, BeerDto.class, BeerMapper.class, Category.class, CustomMapper.class } )
+@RunWith(AnnotationProcessorTestRunner.class)
+public class StaticsTest {
+
+    @Test
+    public void shouldUseStaticMethod() {
```
agudian: Your Test could also assert on the non-existing fields for unused mappers. See ConflictingTypesNamesTest for an example. You could even extend JavaFileAssert with "containsField(...)" or "containsNoField(...)" ... :wink:

sjaakd: Did you check the `CustomMapper.class`? It has a private constructor. That serves 2 purposes:
1. Making sure an instance variable cannot be instantiated without compilation errors
2. Keeping check style happy which keeps insisting that utility classes should have private constructors.


> You could even extend JavaFileAssert with "containsField(...)" or "containsNoField(...)" ...

Like that. But don't think its needed in this case.

sjaakd:
Ok.. Fixed the `AnnotationMapperReference`. WDYT, can I merge it?

agudian:
Yes, go ahead! :+1:

processor/src/test/java/org/mapstruct/ap/test/references/statics/Beer.java
```diff
@@ -0,0 +1,37 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.references.statics;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+public class Beer {
```
gunnarmorling: :-)

gunnarmorling:
Nice, one. Thx!

Issue #429, opened 2015-01-24
Using Builder Pattern for SourceMethod


sjaakd:
Unifying approach.

sjaakd:
Done!

PR #430, opened 2015-01-24
#429 Using Builder Pattern for SourceMethod

sjaakd:


agudian:
Alrighty, go ahead with it. :+1:

Once my checkstyle commot from my PR is merged, I'll go over all the // CHECKSTYLE:OFF comments in the code and replace them with @SuppressWarnings("...").

sjaakd:
Thanks!

gunnarmorling:
Cool, thx!

I kind of liked the fact that the previous methods clearly conveyed which params are needed for which case, but I agree the long param list became quite unwieldly.

Issue #432, opened 2015-01-24
Introduce qualifiedBy in BeanMapping


sjaakd:
Factory methods can be selected as well by means of a qualifier.

sjaakd:
Merged. Doc for mapstruct.org already in place.

PR #435, opened 2015-01-25
#432 Selection of factory method based on Qualifier

sjaakd:
* Aligning `@BeanMapping` with other mapping types.
* Small improvement on templates (removing empty lines from fields)

### DIFF
--- a/core-common/src/main/java/org/mapstruct/BeanMapping.java
+++ b/core-common/src/main/java/org/mapstruct/BeanMapping.java
@@ -27,6 +27,8 @@ import java.lang.annotation.Target;
 /**
  * Configures the mapping between two bean types.
  *
+ * <p>Note: either @BeanMapping#resultType or @BeanMapping#qualifiedBy must be specified</p>
+ *
  * @author Sjaak Derksen
  */
 @Target( ElementType.METHOD )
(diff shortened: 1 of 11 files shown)


processor/src/main/java/org/mapstruct/ap/model/source/BeanMapping.java
```diff
@@ -0,0 +1,76 @@
...
+    public static BeanMapping fromPrism( BeanMappingPrism beanMapping, ExecutableElement method, Messager messager ) {
+        if ( beanMapping == null ) {
+            return null;
+        }
+
+        boolean resultTypeIsDefined = !TypeKind.VOID.equals( beanMapping.resultType().getKind() );
+        if ( !resultTypeIsDefined && beanMapping.qualifiedBy().isEmpty() ) {
+            messager.printMessage(
+                Diagnostic.Kind.ERROR,
+                "'resultType' and 'qualifiedBy' are are are undefined in @BeanMapping, "
```
gunnarmorling: Some "are" too much here :)

core-common/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -38,4 +39,15 @@
...
+   /**
+     * A qualifier can be specified to aid the selection process of a suitable factory method. This is useful in
+     * case multiple factory method (hand written of internal) qualify and result in an 'Ambiguous factory methods'
+     * error.
+     *
+     * A qualifier is a custom annotation and can be placed on either a hand written mapper class or a method.
+     *
+     * @return the qualifiers
+     */
+    Class<? extends Annotation>[] qualifiedBy() default { };
```
gunnarmorling: Can we add a note that either this or the other attribute is to be specified. Similar to how it's done for `@Mapping`.

gunnarmorling:
Looks good to me.

sjaakd:
@gunnarmorling  thanks for reviewing. I processed your comments and will merge soon.

Issue #433, opened 2015-01-24
Introduce resultType in Mapping


sjaakd:
Alignment with `@BeanMapping#resultType`. Mapping methods can be selected as well based on result type.

sjaakd:
Documentation handled. Solution merged.

PR #436, opened 2015-01-26
#433 introduce resultType in @Mapping

sjaakd:
* Introduction of resultType in `@Mapping`
* Doubting if it makes sense to do the same for `@IterableMapping` and `@MapMapping`
* Renaming package `org.mapstruct.ap.test.selection.inheritance` with `org.mapstruct.ap.test.selection.resulttype`
* Bonus: `@SuppressWarnings( "checkstyle:parameternumber" )` in `SourceMethod`

### DIFF
--- a/core-common/src/main/java/org/mapstruct/IterableMapping.java
+++ b/core-common/src/main/java/org/mapstruct/IterableMapping.java
@@ -29,6 +29,9 @@ import java.util.Date;
 /**
  * Configures the mapping between two iterable types, e.g. {@code List<String>} and {@code List<Date>}.
  *
+ * <p>Note: either  @IterableMapping#dateFormat, @IterableMapping#resultType or @IterableMapping#qualifiedBy
+ * must be specified</p>
+ *
  * @author Gunnar Morling
  */
 @Target(ElementType.METHOD)
@@ -52,4 +55,12 @@ public @interface IterableMapping {
      * @return the qualifiers
      */
     Class<? extends Annotation>[] qualifiedBy() default { };
+
+    /**
+     * Specifies the type of the element to be used in the result of the mapping method in case multiple mapping
+     * methods qualify.
+     *
+     * @return the elementTargetType to select
+     */
+    Class<?> elementTargetType() default void.class;
 }
--- a/core-jdk8/src/main/java/org/mapstruct/Mapping.java
+++ b/core-jdk8/src/main/java/org/mapstruct/Mapping.java
@@ -136,4 +136,11 @@ public @interface Mapping {
      * @return the qualifiers
      */
     Class<? extends Annotation>[] qualifiedBy() default { };
+
+   /**
+     * Specifies the result type of the mapping method to be used in case multiple mapping methods qualify.
+     *
+     * @return the resultType to select
+     */
+    Class<?> resultType() default void.class;
 }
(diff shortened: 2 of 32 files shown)


core-jdk8/src/main/java/org/mapstruct/Mapping.java
```diff
@@ -136,4 +136,11 @@
      * @return the qualifiers
      */
     Class<? extends Annotation>[] qualifiedBy() default { };
+
+   /**
+     * Specifies the result type of the mapping method to be used in case mapping factory methods qualify.
```
gunnarmorling: How do factory methods fit into the picture here? Should it be

"Specifieds the result type of the property mapping method to be used in case several methods qualify"

?

processor/src/main/java/org/mapstruct/ap/model/IterableMappingMethod.java
```diff
@@ -100,6 +100,7 @@ public IterableMappingMethod build() {
                 null, // there is no targetPropertyName
                 dateFormat,
                 qualifiers,
+                null, // resulttype does not seem to make sense
```
gunnarmorling: Would it make sense if considered as "element type"?

core-jdk8/src/main/java/org/mapstruct/Mapping.java
sjaakd: They don't.. Copy paste error. I'll fix

processor/src/main/java/org/mapstruct/ap/model/IterableMappingMethod.java
sjaakd: I'll check if I can add some meaningful test cases for this.

sjaakd:
@gunnarmorling, processed your comments. Have a look.

sjaakd:
Guys, can I merge this one (and the next)..?

core-common/src/main/java/org/mapstruct/IterableMapping.java
```diff
@@ -52,4 +55,11 @@
      * @return the qualifiers
      */
     Class<? extends Annotation>[] qualifiedBy() default { };
+
+    /**
+     * Specifies the result type of the element mapping method to be used in case multiple mapping methods qualify.
+     *
+     * @return the resultType to select
+     */
+    Class<?> resultType() default void.class;
```
gunnarmorling: `resultType` seems confusing to me, as it's really the type of the elements. Maybe `elementTargetType`, `keyTargetType`, `valueTargetType` (for map mappings) and `targetType` for bean mappings? I liked result type for the latter, but it doesn't seem to scale for iterables/maps.

gunnarmorling:
@sjaakd, looks good to me apart from that one renaming suggestion. If you like, move forward and merge yourself. I guess a doc update is coming as well? Thanks a lot!

core-common/src/main/java/org/mapstruct/IterableMapping.java
sjaakd: Ok. I'll adapt. But now we have an inconsistency for `@BeanMapping#resultType` and `@Mapping#resultType` which do in essence the same.  That's also confusing for the user, wouldn't you agree?

Issue #434, opened 2015-01-24
Determine resultType based on #resultType in absence of a factory method


sjaakd:

sjaakd:
Documentation handled. Solution merged.

PR #438, opened 2015-01-28
#434 construct resultType based on @BeanMaping#resultType when no factory methods selected

sjaakd:
Based on PR for issue #433. Check that one first.


gunnarmorling:
Looks good. That's a nice improvement. Thanks, @sjaakd!

Issue #411, opened 2015-01-06
Add import if source or target type of a mapping is an inner class


gunnarmorling:
From the user group: https://groups.google.com/forum/?hl=en#!topic/mapstruct-users/Dc1q-lApC-g

No import is added for inner classes used as source or target type. I only think we can handle static inner classes, as for non-static inner classes we wouldn't know which instance of the outer class should be used.

gunnarmorling:
Done.

Issue #209, opened 2014-05-05
When using enums as inner class, import statement is missing 


carstencogos:
When mapping to a type with a enum, and the enum is declared inside the bean the MapperImpl is missing the import statement. 

i.e

public class BeanWithEnum {

 public enum MyEnum { 
A, B };

private MyEnum field1;

....getter/setter ...

}

gunnarmorling:
Done.

PR #439, opened 2015-01-31
412 Support mapping static inner classes

eforest:
Adds import declaration for static inner classes / inner enums to mapper
even if it is in the same package.

Issue 209 is kind of a clone and is also solved.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/model/GeneratedType.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/GeneratedType.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.jar.Manifest;
+
 import javax.annotation.Generated;
 
 import org.mapstruct.ap.model.common.Accessibility;
@@ -43,6 +44,7 @@ public abstract class GeneratedType extends ModelElement {
 
     private static final String COMMENT_TAG = "Implementation-Version";
     private static final String COMMENTS = initComment();
+    private static final String JAVA_LANG_PACKAGE = "java.lang";
 
     private final String packageName;
     private final String name;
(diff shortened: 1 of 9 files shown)


gunnarmorling:
> testing for static nested classes

You mean like an static inner class within another static inner class? Seems a bit esoteric, but yes, it should work ideally.

processor/src/main/java/org/mapstruct/ap/model/GeneratedType.java
```diff
@@ -180,6 +178,28 @@ private void addWithDependents(Collection<Type> collection, Type typeToAdd) {
         }
     }
 
+    private boolean needsImportDeclaration(Type typeToAdd) {
```
gunnarmorling: Nice to extract this method!

eforest:
Maybe I mixed up 'nested' / 'inner' too much. 
Naming proposals are welcome. 

For clarification : 
* static nested classes are supported
* inner enums (#209) (don't know if inner enum is the right term here) are supported 
* inner classes (#411) (non-static nested classes) are **not** supported

sjaakd:
I think [this](http://stackoverflow.com/questions/1353309/java-static-vs-non-static-inner-class) sums it up quite nicely. Nested classes come in to variants: static and non-static. Non-static nested classes are aka inner classes.

gunnarmorling:
Rebased and applied. Thanks, @eforest! I've added a test for static inner classes declared within other static classes (this sort of nesting is allowed), this works as expected as well. Note that I reworded your commit message slightly so the issue numbers are rendered as links here on GitHub.

@sjaakd: You are right. We cannot really support non-static inner classes as we wouldn't know which instance of the outer class to use during instantiation. The only thing we really can do is raise meaningful exception when declaring a mapping for such classes.

eforest:
Cool, never used a static nesting level deeper than one.

BTW, is there a roadmap for releasing Beta-4?

Issue #339, opened 2014-11-10
MapStruct error messages clarity


sjaakd:
* #321 points out an issue on clarity. 
* The introduction of nested properties has lead to unclear message on ambiguous property mapping.

These can be fixed as individual cases. However, its perhaps time a the messages we have introduced in a more general perspective and tackle them in one go. Here is (at least what I found in a quick scan) the set of messages MapStruct can generate: 

```
// enum
"A source constant must be specified for mappings of an enum mapping method."
"A target constant must be specified for mappings of an enum mapping method."
"One enum constant must not be mapped to more than one target constant, but constant %s is mapped to %s."
"The following constants from the source enum have no corresponding constant in the target enum and must be be mapped via @Mapping: %s"

// decorators
"Specified decorator type is no subtype of the annotated mapper type."
"Specified decorator type has no default constructor nor a constructor with a single parameter accepting the decorated mapper type."
"Specified decorator type is no subtype of the annotated mapper type.

// inverse
"Resolved inverse mapping method %s() should not carry the @InheritInverseConfiguration annotation itself."
"Several matching inverse methods exist: %s(). Specify a name explicitly."
"None of the candidates %s() matches given name: \"%s\"."
"Given name \"%s\" matches several candidate methods: %s()."
"Given name \"%s\" does not match the only candidate. Did you mean: \"%s\"."

// factory
"Ambiguous factory methods: \"%s\" conflicts with \"%s\"."

// beanmapping
"Several possible source properties for target property \""
"Found several matching getters for property \"%s\""
"Unmapped target {0,choice,1#property|1<properties}: \"{1}\""

// beanmapping, property
"Can't map %s to \"%s %s\"." 
"Target property \"%s\" must not be mapped more than once."
"Target must not be empty in @Mapping"
"Source and constant are both defined in @Mapping, either define a source or a constant"
"Source and expression are both defined in @Mapping, either define a source or an expression"
"Expression and constant are both defined in @Mapping, either define an expression or a constant"
"Ambiguous mapping methods found for mapping " + mappedElement + " from %s to %s: %s."
"Value must be given in the form \"java(<EXPRESSION>)\""
"Method has no parameter named \"%s\"."
"Can't map \"%s %s\" to \"%s %s\"."
"The type of parameter \"%s\" has no property named \"%s\"."
"No property named \"%s\" exists in source parameter(s)."

// iterable
"Can't create implementation of method %s. Found no method nor built-in conversion for mapping source element type into target element type."

[...]

PR #440, opened 2015-02-02
#339 introduce resource bundle

sjaakd:
I took the approach of introducing a resource bundle iso the constants as proposed in #339. Also fixed some small inconsistencies (interpunction, one tiny spelling error). 

It shows nicely how messages / terminology are used. We could be more consistent with quotes (sometimes we use them, sometimes not). But all-in-all, its not that bad :smile: 

`String.format` a re-occuring pattern, the solution pushes the `String.format` to the newly introduced `FormattingMessager`.


gunnarmorling:
Wow, that seems like a an awful lot of work :)

sjaakd:
Ha! Now the stats work I want to be on top :smile:. Without kidding, it's not so bad. Just a bit boring and tedious. Costed me 3 hours in the end.

gunnarmorling:
@sjaakd, I think the centralization of messages is a great idea. But maybe we can do better than using Strings for referencing messages.

E.g. we can use an enum as in the top answer in this [SO post](https://stackoverflow.com/questions/446663/best-way-to-define-error-codes-strings-in-java). We can have the messages within the enum type itself, as we don't want to I18N messages, so there is no strong benefit to using a resource bundle. Such approach will avoid typos when referencing messages and also makes it a bit more refactoring-safe.

WDYT?

gunnarmorling:
> Ha! Now the stats work I want to be on top

Ah, nice that it's working now. Indeed I need to keep an eye on you ;)

gunnarmorling:
Or maybe an approach like this: https://code.google.com/p/jlibs/wiki/Internationalization

I wouldn't want to add a dependency to an external library, but we could easily implement such an approach ourselves e.g. using a dynamic proxy (or via code generation using an annotation processor :-) ).

sjaakd:
> E.g. we can use an enum as in the top answer in this SO post. We can have the messages within the enum type itself, as we don't want to I18N messages, so there is no strong benefit to using a resource bundle. Such approach will avoid typos when referencing messages and also makes it a bit more refactoring-safe.
> 
> WDYT?

I think that's a good idea.. However, I would like to switch off the 120 characters limit then on the definition part of this particular enum. I think it offers quite nice insight to have 'unbroken' strings (at least it did for me)..

gunnarmorling:
> However, I would like to switch off the 120 characters limit

Yes, I guess that makes sense.

sjaakd:
Lets go for the enum approach then.. Keep it simple, plain and stupid :+1:

gunnarmorling:
Ok :)

sjaakd:
Ok.. Done!

processor/src/main/java/org/mapstruct/ap/model/source/IterableMapping.java
```diff
@@ -52,12 +53,7 @@ public static IterableMapping fromPrism(IterableMappingPrism iterableMapping, Ex
         if ( !elementTargetTypeIsDefined
             && iterableMapping.dateFormat().isEmpty()
             && iterableMapping.qualifiedBy().isEmpty() ) {
-            messager.printMessage(
-                Diagnostic.Kind.ERROR,
-                "'dateformat', 'qualifiedBy' and 'elementTargetType' are undefined in @IterableMapping, "
-                    + "define at least one of them.",
-                method
-            );
+            messager.printMessage( Diagnostic.Kind.ERROR, method, Message.iterablemapping_noelements );
```
gunnarmorling: Now with the enum approach in place, couldn't also the severity/kind be part of the enum definition? I don't think one message is reported with different kinds at different places, so we can move it there and simplify the calls to printMessage().

processor/src/main/java/org/mapstruct/ap/util/Message.java
```diff
@@ -0,0 +1,100 @@
...
+package org.mapstruct.ap.util;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+public enum Message {
+
+    // CHECKSTYLE:OFF
+    beanmapping_noelements( "'resultType' and 'qualifiedBy' are undefined in @BeanMapping, define at least one of them." ),
```
gunnarmorling: Any chance you convert the enum member names to upper-case? It's just more common and what you'd expect from it. Not sure though whether there is a good way to automate?

processor/src/test/java/org/mapstruct/ap/model/common/DefaultConversionContextTest.java
```diff
@@ -119,26 +120,26 @@ private Type typeWithFQN(String fullQualifiedName) {
...
     }
 
-    private static class StatefulMessagerMock implements Messager {
+    private static class StatefulMessagerMock implements FormattingMessager {
 
         private Diagnostic.Kind lastKindPrinted;
 
         @Override
-        public void printMessage(Diagnostic.Kind kind, CharSequence msg) {
+        public void printMessage( Diagnostic.Kind kind, Message msg, Object... arg) {
```
gunnarmorling: We don't do a whitespace after braces in method definitions usually.

sjaakd: Ok..

processor/src/main/java/org/mapstruct/ap/util/Message.java
sjaakd: Yep.. I'll just do the refactor mode.. A bit of effort..

processor/src/main/java/org/mapstruct/ap/model/source/IterableMapping.java
sjaakd: Good point..

gunnarmorling: Maybe we can add a CS check for it?

sjaakd:
@gunnarmorling , processed your comments into an additional commit. Please have a look.

gunnarmorling:
Merged; Many thanks, @sjaakd! It's much nicer with the enum now :+1:

Issue #194, opened 2014-04-05
Only generate field / imports for a referenced mapper if it is actually used


agudian:

sjaakd:
Partially done.. The import remains. The field is only included if a non-static method is actually called. 

I'm wondering here wether we should do anything more. The user has a certain responsibility as well. WDYT?

agudian:
Getting rid of the import as well would be great, but we can tackle that in a separate issue later on in case we see the need for it.
Generally, such unused referenced mappers could lurk in from general mapper-config classes where maybe not every "used" mapper is required in all the mappers. So it might become tideous for users to get rid of them by creating another mapper-config or something like that.

@sjaakd: it's in beta4, right?

sjaakd:
> @sjaakd: it's in beta4, right?

Yep..

What I can do is make a fix for imports as well. Its a quite easy effort.. Now we set a primitive `boolean` used depending if its a static method / or not.. We might decide to make this is a wrapper `Boolean`. Not initialized means: not used at all and can be taken from the imports. WDYT?

We can keep this one open and if we agree, I'll make a fix today.

Btw: I think its better to move the check in the templates to an outer level (so to GeneratedType). Now sometimes an empty line is inserted for instance variables when MapStruct decides its not used.

sjaakd:
Sorry. Could not leave it like this :smile: lets fix it completely.

sjaakd:
ok now really done..

gunnarmorling:
Cool, thanks!

PR #443, opened 2015-02-04
#194 Only generate field / imports for a referenced mapper if it is actu...

sjaakd:
...ally used

And as bonus, some small template updates.. (fixing empty lines in case of omitted fields).

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/model/DecoratorConstructor.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/DecoratorConstructor.java
@@ -20,8 +20,8 @@ package org.mapstruct.ap.model;
 
 import java.util.Collections;
 import java.util.Set;
-
 import org.mapstruct.ap.model.common.ModelElement;
+
 import org.mapstruct.ap.model.common.Type;
 
 /**
@@ -29,7 +29,7 @@ import org.mapstruct.ap.model.common.Type;
  *
  * @author Gunnar Morling
  */
-public class DecoratorConstructor extends ModelElement {
+public class DecoratorConstructor extends ModelElement implements Constructor {
 
     private final String name;
     private final String delegateName;
@@ -46,6 +46,7 @@ public class DecoratorConstructor extends ModelElement {
         return Collections.emptySet();
     }
 
+    @Override
     public String getName() {
         return name;
     }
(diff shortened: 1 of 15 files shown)


processor/src/main/java/org/mapstruct/ap/model/DecoratorConstructor.java
```diff
@@ -29,13 +28,14 @@
  *
  * @author Gunnar Morling
  */
-public class DecoratorConstructor extends ModelElement {
+public class DecoratorConstructor extends Field {
```
gunnarmorling: Constructor extending Field? That seems a bit odd.

processor/src/test/java/org/mapstruct/ap/test/references/statics/StaticsTest.java
```diff
@@ -44,4 +55,19 @@ public void shouldUseStaticMethod() {
         assertThat( result ).isNotNull();
         assertThat( result.getCategory() ).isEqualTo( Category.STRONG ); // why settle for less?
     }
+
+    @Test
+    @WithClasses( {  BeerMapperWithNonUsedMapper.class, NonUsedMapper.class } )
+    public void shouldNotImportNonUsed() {
+
+        Beer beer = new Beer(); // what the heck, open another one..
```
gunnarmorling: :-)

gunnarmorling:
Looks good overall, only I find the re-usage of fields for constructors a bit dubious. Should pass them separately to the `Decorator` constructor instead of sneaking them in throught the fields list?

processor/src/main/java/org/mapstruct/ap/model/DecoratorConstructor.java
sjaakd: It is.. However its confirming the current practice..(that's why I ran into problems)..  I'll change that.

sjaakd:
> Looks good overall, only I find the re-usage of fields for constructors a bit dubious. Should pass them separately to the Decorator constructor instead of sneaking them in throught the fields list?

Good idea. Of course. 

I'll fix.. Do you mind that I merge immediately after (or do you still want to see it)?

gunnarmorling:
Yes, sure, go ahead. Generally just move forward in such case.

Issue #431, opened 2015-01-24
Add JDK 9 to the integration tests


agudian:
... and remove processor_plugin_java_7, as it does exactly the same as the _java_8 counterpart and the oracle_java_8 variant (that is: using javac from the Java 8 JDK that is set as JAVA_HOME).

The oracle_jdk_9 tests needs an explicit off-switch, as our cloudbees build does not have a JDK 9, yet.

PR #448, opened 2015-02-06
#431 add JDK 9 to integration tests and remove processor_plugin_java_7...

agudian:
...tests (processor_plugin_java_7 did exactly the same as the _java_8 variant, for what we are concerned).

Use -DprocessorIntegrationTest.canUseJdk9=false in case no JDK9 is configured in the toolchains.xml file (I've already configured that for the cloudbees builds).

@sjaakd and @gunnarmorling, you may want to install the JDK 9 preview build and add it to your toolchains file (example below was extended)... https://jdk9.java.net/download/

Btw, building MapStruct with JDK 9 almost works - some Tests in the processor fail because the error messages are not what is expected.

### DIFF
--- a/etc/toolchains-example.xml
+++ b/etc/toolchains-example.xml
@@ -33,4 +33,15 @@
             <jdkHome>C:\Program Files\Java\jdk1.8.0_11</jdkHome>
         </configuration>
     </toolchain>
+    <toolchain>
+        <type>jdk</type>
+        <provides>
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Rebased an applied. Thanks, @agudian!

gunnarmorling:
Filed #453 for investigating the tests failing when buiding MapStruct with JDK 9.

Issue #449, opened 2015-02-06
For String -> XmlGregorianCalendar conversion interpret source value as XML lexical representation by default


sjaakd:
Only if an explicit date format is given, the value should be interpreted according to this format.

gunnarmorling:
Change is done; waiting for doc update.

sjaakd:
Done.. Added text to Beta4 branch.

PR #450, opened 2015-02-06
#449 Handle XmlGregoriancalendar from String with empty date format as x...

sjaakd:
...ml datatype

### DIFF
--- a/processor/src/main/resources/org.mapstruct.ap.model.builtin.StringToXmlGregorianCalendar.ftl
+++ b/processor/src/main/resources/org.mapstruct.ap.model.builtin.StringToXmlGregorianCalendar.ftl
@@ -24,10 +24,15 @@ private XMLGregorianCalendar ${name}( String date, String dateFormat ) {
     }
 
     try {
-        DateFormat df = dateFormat != null ? new SimpleDateFormat( dateFormat ) : SimpleDateFormat.getInstance();
-        GregorianCalendar c = new GregorianCalendar();
-        c.setTime( df.parse( date ) );
-        return DatatypeFactory.newInstance().newXMLGregorianCalendar( c );
+        if ( dateFormat != null ) {
+            DateFormat df = new SimpleDateFormat( dateFormat );
+            GregorianCalendar c = new GregorianCalendar();
+            c.setTime( df.parse( date ) );
+            return DatatypeFactory.newInstance().newXMLGregorianCalendar( c );
+        }
+        else {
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Ok, I understood it wrong a bit yesterday. I thought `XmlGregorianCalendar` would have its own means for parsing strings with another format specification.

But actually one cannot specify a format here, Strings directly parsed into `XmlGregorianCalendar` always need to adhere to the XML lexical representation. So `Mapping#dateFormat()` and friends still would always expect a formatting string understood by `DateFormat`.

sjaakd:
> But actually one cannot specify a format here, Strings directly parsed into XmlGregorianCalendar always need to adhere to the XML lexical representation

Exactly. You could say that our initial implementation missed the primary use case of `XmlGregorianCalendar`

> So Mapping#dateFormat() and friends still would always expect a formatting string understood by DateFormat.

Yep.. That still remains as is.

--------
It is however a break with the current implementation behavior and deviates from the other `Date` related formats. So I guess we need to spent some words on it in the documentation.

sjaakd:
Proposal for text in documentation, additional bullet to end of chapter 5.1:

"* When converting from a `java.lang.String`, omitting `Mapping#dateFormat` results in using the default pattern and date format symbols for the default locale. An exception to this rule is `XmlGregorianCalendar` which results in parsing the `String` according to [XML Schema 1.0 Part 2, Section 3.2.[7-14].1, Lexical Representation](http://www.w3.org/TR/xmlschema-2/#dateTime-order)."

agudian:
> "* When converting from a java.lang.String, omitting Mapping#dateFormat results in using the default pattern and date format symbols for the default locale. An exception to this rule is XmlGregorianCalendar which results in parsing the String according to XML Schema 1.0 Part 2, Section 3.2.[7-14].1, Lexical Representation."

How about using just "String" iso "java.lang.String" (I think the chances are low that someone reads this to be the other string from package ```underwear``` :wink:). 

For the second sentence, I'd make it more plain to read: "An exception to this rule is XmlGregorianCalendar which cannot be configured with a date format, as the String value always needs to be in the format according to XML Schema 1.0 Part 2, Section 3.2.[7-14].1, Lexical Representation." -- unless I missunderstood the issue...

sjaakd:
It still can be configured with a date format . It then works similar as the String to date formats,

gunnarmorling:
> How about using just "String" iso "java.lang.String"

+1. Sounds good otherwise.

gunnarmorling:
Thanks, @sjaakd!

Issue #454, opened 2015-02-10
Processor-Classpath and Compile-Classpath must not be mixed


agudian:
While trying out the current code with a more elaborate "example" project, I came accross different issues.

First of which: the processor classpath can not be considered being the same as the compile-classpath. In NativeTypes#isTypeAvailable(String), we try to load classes from NativeTypes.class.getClassloader(). If you run Eclipse with JDK 8, but have a project using JDK 6, then the java.time.* types are present in the processor classpath, but not in the application - leading to an exception later on during processing.

I already have an idea what we could do to spot that in the integration tests properly. PR will follow tomorrow or so.

gunnarmorling:
Interesting. It's quite an obvious bug, but conditions for it are quite rare I guess (you need either Java 8 or Joda in the processor environment but not in the project under test). Good catch!

gunnarmorling:
Done.

PR #457, opened 2015-02-11
#454 Do not check the existence of types in the classloader of the processor

agudian:
... but on the classpath of the sources.

Unfortunately, I was not able to create a mvn build that could be used to reproduce the issue in an integration test. Perhaps you guys find a way for that... ?

Anyway, I tested that the issue is fixed by running the processor within the Eclipse build for a JDK 6 project, whereas Eclipse was started using JRE 8.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/conversion/Conversions.java
+++ b/processor/src/main/java/org/mapstruct/ap/conversion/Conversions.java
@@ -18,7 +18,6 @@
  */
 package org.mapstruct.ap.conversion;
 
-import javax.lang.model.util.Elements;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.Calendar;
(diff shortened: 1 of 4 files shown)


gunnarmorling:
As a test you could write an integration test which has Joda on the processor path but not on the project under compilation (assuming that's what triggering the exception)? But I don't think it's worth it. Your ad-hoc test in Eclipse is good enough IMO.

gunnarmorling:
Merged. Thanks, @agudian!

agudian:
> As a test you could write an integration test which has Joda on the processor path but not on the project under compilation (assuming that's what triggering the exception)?

That is exactly what I tried :smile: - but for the compiler plugin I wasn't able to find the processor when moving it out of the normal dependencies to the compiler-plugin-deps (even with setting forceJavacCompilerUse=true, which is the proposed workaround to get that working).  And the processor plugin mixed the classpaths in any case...

Issue #160, opened 2014-03-02
Enable using property mapping methods that contain @MappingTarget parameters


agudian:
Mapping methods that contain an ```@MappingTarget``` parameter should be usable as property mapping methods.

The generated code should cover the following cases:
* The ```@MappingTarget``` method could return void or the target type. In case the target type is returned, we should call the setter with that value.
* The target property may be null. I guess we should hand in a new instance of the target type (call constructor or a factory method).

Any thoughts?

sjaakd:
@agudian do you see time to work on this? Shall I give it a try?

agudian:
How could I stop you? You're on fire lately! :smile: so don't hold back!

sjaakd:
Thanks.. Doing some early-spring-cleaning :smile:

sjaakd:
Ok, lets start out with an example.

```java
public class Game {

    private Player player;

    public Player getPlayer() {
        return player;
    }

    public void setPlayer(Player player) {
        this.player = player;
    }
}

public class Player {

    private String name;

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    } 
}

public class TargetGame {

    private TargetPlayer player;

    public TargetPlayer getPlayer() {
        return player;
    }

    public void setPlayer(TargetPlayer player) {
        this.player = player;
    }

}

public class TargetPlayer {

    private String name;

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }
}

@Mapper
public interface ExampleMapper {

    void mapGame( Game game, @MappingTarget TargetGame targetGame );

    void mapPlayer( Player player, @MappingTarget TargetPlayer targetPlayer );

    TargetPlayer mapPlayer( Player player );
}
```

So the result (I think) should be:

```java
public class ExampleMapperImpl1 implements ExampleMapper {

    @Override
    public void  mapGame(Game game, TargetGame targetGame)  {
        if ( game == null ) {
            return;
        }

        mapPlayer( game.getPlayer(), targetGame.getPlayer() );
    }

    @Override
    public void mapPlayer(Player player, TargetPlayer targetPlayer)  {
        if ( player == null ) {
            return;
        }

        targetPlayer.setName( player.getName() );
    }

    @Override
    public TargetPlayer mapPlayer(Player player)  {
        if ( player == null ) {
            return null;
        }

        TargetPlayer targetPlayer = new TargetPlayer();

        targetPlayer.setName( player.getName() );

[...]

PR #461, opened 2015-02-14
#160 allow selection of update methods by refactoring MethodMatcher but ...

sjaakd:
...stopping selection of update method at new filter

This is the first step for #160: making sure that update methods are actually selected.  I cut this one into smaller  steps to make it more manageable for reviewing.

There's also some refactoring in this PR. Idea being: `MethodMatcher` is responsible for all the method matching (there was some stuff, like `@TargetType`  outside this class) and there was the suggestion that we could make mappings on more than 1 source parameter. That has been fixed.

The following things are still on my list for #160:
1. Updating the templates / wrappers, (which turns out more difficult than I though).  
2. Limiting: I don't think we can support nested (2 step) mappings for update methods, so method(method) or conversion(method). 
3. Iterable / MapMappings.
4. Forged methods.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/model/MappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/MappingMethod.java
@@ -50,7 +50,7 @@ public abstract class MappingMethod extends ModelElement {
         this.name = method.getName();
         this.parameters = method.getParameters();
         this.returnType = method.getReturnType();
-        this.targetParameter = method.getTargetParameter();
+        this.targetParameter = method.getMappingTargetParameter();
         this.accessibility = method.getAccessibility();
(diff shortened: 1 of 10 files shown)


agudian:
Looks good!

> 2. Limiting: I don't think we can support nested (2 step) mappings for update methods, so method(method) or conversion(method).

I think that should be OK :)

sjaakd:
Merge ready? 😃

agudian:
Your github mail didn't notify you about my editing, I'd say... :smile:

sjaakd:
Ah! I check it out this evening! 😄

Verstuurd vanaf mijn iPhone

sjaakd:
> Looks good! Go ahead with the merge, if you like.

Thanks. Somehow missed that last part in my mail :smile:

Issue #456, opened 2015-02-10
Formatting in generated code needs some cleanup


agudian:
Before we ship Beta4, we should try to clean up some formatting issues that crept in.

For example, in this file created by AbstractClassTest, there are a lot of unnecessary line breaks that we should try to get rid of again:

```java
package org.mapstruct.ap.test.abstractclass;

import javax.annotation.Generated;

@Generated(
    value = "org.mapstruct.ap.MappingProcessor",
    date = "2015-02-10T22:56:30+0100",
    comments = "version: , compiler: javac, environment: Java 1.8.0_25 (Oracle Corporation)"
)
public class SourceTargetMapperImpl extends SourceTargetMapper {

    private final ReferencedMapper referencedMapper = new ReferencedMapper();



    @Override
    public Target sourceToTarget(Source source) {
        if ( source == null ) {
            return null;
        }

        Target target = new Target();


        target.setBirthday( calendarToString( source.getBirthday() ) );
        target.setNotAttractingEqualsMethod( referencedMapper.objectToBoolean( source.getNotAttractingEqualsMethod() ) );
        target.setId( source.getId() );
        target.setSize( (long) source.getSize()
        );
        target.setManuallyConverted( referencedMapper.holderToInt( source.getManuallyConverted() ) );



        return target;
    }

    @Override
    public Target sourceToTargetFromBaseMapperInterface(Source source) {
        if ( source == null ) {
            return null;
        }

        Target target = new Target();


        target.setBirthday( calendarToString( source.getBirthday() ) );
        target.setNotAttractingEqualsMethod( referencedMapper.objectToBoolean( source.getNotAttractingEqualsMethod() ) );
        target.setId( source.getId() );
        target.setSize( (long) source.getSize()
        );
        target.setManuallyConverted( referencedMapper.holderToInt( source.getManuallyConverted() ) );



        return target;
    }

}
```

sjaakd:
:+1: Good points.. I did fix some of them already (and possibly introduced some more). But somehow they keep popping up when you try to make readable free marker lines e.g. (if - then - else) stuff.

gunnarmorling:
Yes, +1 for fixing this.

Instead of addressing it within the templates, we may add logic to `IndentationCorrectingWriter` which reduces several subsequent new lines to a single one.

agudian:
Noooo, didn't see you self-assigning this one. I'll push a PR with a proposed fix anyway, so you can see if you come up with a better approach...

[...]

PR #462, opened 2015-02-15
#456 Avoiding to print multiple consecutive blank lines

gunnarmorling:


agudian:
The output looks very good on Windows, even the linebreaks are correct now... :+1: 

Do you rebase and merge yourself?

gunnarmorling:
> The output looks very good on Windows, even the linebreaks are correct now... :+1:

Cool, glad to hear. Yes, it bugged me that we'd always use the line breaks from the templates rather than from the current system :)

Ah, to be sure, could you take the JAR built on CI (once it's merged) and test that one as well? The templates there should be Linux-style, which should be translated to Windows now properly. Thx!

> Do you rebase and merge yourself?

Just done. Thx!

agudian:
> Ah, to be sure, could you take the JAR built on CI (once it's merged) and test that one as well? The templates there should be Linux-style, which should be translated to Windows now properly. Thx!

Locally, the templates are also with linux line endings, also after the build. The generated sources then all use windows line endings. So it works as you expected it to... :+1:

gunnarmorling:
Cool, thanks for trying! I'm quite happy now with the formatter class; The
onEntry()/onExit() hooks make it much easier to put new logic into the
right place.

Issue #263, opened 2014-07-06
 Allow config classes to refer to another config class 


gunnarmorling:
There should be a processor option for defining the default collection mapping strategy in a global manner.

Alternatively, we could get rid of `defaultComponentModel` and `unmappedTargetPolicy` and rely on at `@MapperConfig` exclusively for shared mapper configurations. I'd even think that makes actually more sense. @sjaakd, @agudian WDYT?

gunnarmorling:
@agudian, @sjaakd: I think this problem would be nicely solved by allowing config classes to refer to another config class themselves:

    @MapperConfig
    public class RootConfig {}

    @MapperConfig(config=RootConfig.class)
    public class MyConfig {}

    @Mapper(config=MyConfig.class)
    public interface MyMapper { ... }

That way one can have one centralized root config for all other mappers and mapper configs. We could remove the processor options then (only leaving `SUPPRESS_GENERATOR*` in place, which I think make more sense as an AP option). WDYT?

agudian:
:+1:

sjaakd:
:+1:

Issue #60, opened 2013-07-25
Automatically create (private) mapping methods to map nested properties


agudian:

agudian:
What I meant here were forged methods for bean properties.

gunnarmorling:
@agudian Is there actually anything left to be done here? we already create forged methods for mapping nesting props, right? If that's not what you have in mind, could you add an example?

agudian:
Example:
```java
class Car {
    private SteeringWheel steeringWheel;
    private List<Seat> seats;
...
}
// classes CarDto, SteeringWheel, SteeringWheelDto, Seat, SeatDto look equally dull and don't contain any property name differences
```

In my Mapper, I'd like to just write:
```java
@Mapper
interface CarMapper {
  CarDto toCarDto( Car car );
  Car toCar( CarDto dto );
}
```

Right now, we report errors on unmappable target properties, and suggest to declare mapping methods for SteeringWheel <-> SteeringWheelDto and Seat <-> SeatDto. 
The idea was to create  forged methods for those cases as well, so that I'd only have to declare them myself if I need to fine-tune the mapping or want to publicly expose those methods in my mapper interface.

We can't do that, yet, can we?

sjaakd:
> We can't do that, yet, can we?

Nope.. But I would really like such a feature. Especially  in combination with #168 and #292 :smile:

sjaakd:
@navpil thanks for the great work!

PR #463, opened 2015-02-15
#168 Allow to inherit mapping method configurations from the @MapperConfig-annotated type

agudian:
...either automatically when all method types match, or explicitly using @InheritConfiguration.

I did basically what we discussed, I hope. But I'm sure there's still something I didn't get right... :wink:

One thing I changed was to relax the restrictions on chained ```@Inherit*Configuration``` - it is now possible to inherit from a method that also inherits from somewhere. For inheriting from the mapper-config methods, chained inheritance is needed - and so it was just easier for me to remove the restriction rather than refine the check to differentiate between what's necessary and what should be disallowed. Cycles are detected, though.

There are some tests missing, but I'd like to continue there as after you guys had a closer look on this.


agudian:
Thanks for your review! I'll look into them in more detail tonight.

> I'm a bit worried on multiple arguments in an inherit configuration.. (Wondering whether this works currently)..

Currently, this only works if the parameters have the same name in the "inheritance source method" and the "inheritence target method" (that goes both for multiple source params and for single source params). 
This PR fixes this if the parameter types are different enought. The methods Mapping#forInheritanceTo(SourceMethod) and SourceReference.copyForInheritanceTo(SourceMethod) were introduced to do exactly that (I should probably align the method names here :wink:). But it still doesn't take into account that you could have the same type twice, as in your example. 
I wonder if we could handle this separately, as it might not be straight forward to automatically detect which one would be the right.

sjaakd:
> I wonder if we could handle this separately, as it might not be straight forward to automatically detect which one would be the right.

Good idea.. We should make a separate issue for this one.

agudian:
Oh great.. @sjaakd, I just worked in most of your comments and pushed the updated commit - but now all your comments are gone here... 
That usually doesn't happen if you click on the tab "Files changed" above and place your comments there - github will then preserve the comments and just mark them as "commented on an outdated diff".

One note that I didn't yet do is on adding more tests, e.g. some including ```@MappingTarget``` - that and at least one errornous test will follow as soon as @gunnarmorling gave this one a closer look as well... just to make sure I'm on the right track now.

agudian:
Oh, so the missing comments are here: https://github.com/mapstruct/mapstruct/commit/9d850bb422ec5ca50d3761302ac362fc618b957f

gunnarmorling:
> but now all your comments are gone here... 

This happens if you force-push to a pull request branch. Better to just push more commits addressing the review remarks and only squash them once all are happy and the PR is good to be merged.

I'll look into it asap, likely it will be the weekend only, though. Thanks for doing this work! Can we close the first PR for it btw with this one in place now?

core-common/src/main/java/org/mapstruct/MapperConfig.java
```diff
@@ -88,5 +88,16 @@
...
+
+    /**
+     * The strategy to use for applying {@code @Mapping} configurations of prototype methods in the interface annotated
+     * with this annotation. If no strategy is configured, {@link MappingInheritanceStrategy#WHEN_EXPLICITLY_REFERENCED}
+     * will be used as default.
+     *
+     * @return The strategy to use for applying {@code @Mapping} configurations of prototype methods in the interface
+     *         annotated with this annotation.
+     */
+    MappingInheritanceStrategy mappingInheritedFromConfig()
```
gunnarmorling: WDYT about naming it `mappingInheritanceStrategy()` to be consistent with the style of the other options?

agudian:
> This happens if you force-push to a pull request branch. Better to just push more commits addressing the review remarks and only squash them once all are happy and the PR is good to be merged.

See here for example: #443 - there was a comment on a change that was replaced by force-pushing a new commit, but you can still see it there. For some longer discussions, rebasing might also be necessary. So I'd generally prefer using comments on "Files changed" instead of on changes within "Commits". Coincidentally, the ProTip on the bottom of the page also says right now:

> *ProTip!* Add comments to specific lines under [Files changed](https://github.com/mapstruct/mapstruct/pull/463/files). 

:smile: 

But of course: it's generally nicer to see the changes for review comments in a new commit to simplify the review iterations. The one merging the PR should just not forget to squash them later on...

core-common/src/main/java/org/mapstruct/MapperConfig.java
```diff
@@ -80,7 +80,7 @@
      *
```
gunnarmorling: The class-level docs of `@MapperConfig` need an update. AFAIU, only interfaces and abstract classes are valid configuration sources now. Also the comment should mention that any methods declared on the annotated config class are considered prototypes which can be inherited automatically or via `@InheritConfiguration`.

core-common/src/main/java/org/mapstruct/MappingInheritanceStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+     * Apply the {@code @Mapping} configuration only if the prototype method is explicitly referenced using
+     * {@link InheritConfiguration}.
+     */
+    WHEN_EXPLICITLY_REFERENCED,
+
+    /**
+     * Apply the {@code @Mapping} configuration if source and target types of the prototype method are assignable from
+     * the types of a given mapping method.
+     */
+    AUTOMATICALLY_WHEN_ALL_TYPES_MATCH,
```
gunnarmorling: @agudian, what is intended to happen when using this strategy and several prototype methods are matching? On first thought I'd say that method should be marked as erroneous, requiring to explicitly specify a prototype method via `@InheritConfiguration. WDYT?

processor/src/main/java/org/mapstruct/ap/processor/MapperCreationProcessor.java
```diff
@@ -507,34 +558,6 @@ private void reportErrorWhenInheritForwardAlsoHasInheritReverseMapping(SourceMet
 
     }
 
-    private void reportErrorIfInverseMethodHasInheritAnnotation(SourceMethod candidate,
```
gunnarmorling: Where did this check go?

processor/src/main/java/org/mapstruct/ap/util/Message.java
```diff
@@ -86,18 +86,16 @@
...
     INHERITINVERSECONFIGURATION_NO_NAME_MATCH( "Given name \"%s\" does not match the only candidate. Did you mean: \"%s\"." ),
-    INHERITCONFIGURATION_REFERENCE_HAS_FORWARD( "Resolved mapping method %s() should not carry the @InheritConfiguration annotation itself." ),
-    INHERITCONFIGURATION_REFERENCE_HAS_INVERSE( "Resolved mapping method %s() should not carry the @InheritInverseConfiguration annotation." ),
     INHERITCONFIGURATION_DUPLICATES( "Several matching methods exist: %s(). Specify a name explicitly." ),
     INHERITCONFIGURATION_INVALIDNAME( "None of the candidates %s() matches given name: \"%s\"." ),
     INHERITCONFIGURATION_DUPLICATE_MATCHES( "Given name \"%s\" matches several candidate methods: %s()." ),
-    INHERITCONFIGURATION_NO_NAME_MATCH( "Given name \"%s\" does not match the only candidate. Did you mean: \"%s\"." );
+    INHERITCONFIGURATION_NO_NAME_MATCH( "Given name \"%s\" does not match the only candidate. Did you mean: \"%s\"." ),
+    INHERITCONFIGURATION_MULTIPLE_PROTOTYPE_METHODS_MATCH( "More than one configuration prototype method are applicable. Use @InheritConfiguration to select one of them explicitly: %s." ),
```
gunnarmorling: Ah, I guess that answers my previous question. I think it's the right thing to do.

processor/src/main/java/org/mapstruct/ap/util/Message.java
```diff
@@ -86,18 +86,16 @@
...
+    INHERITCONFIGURATION_CYCLE( "Cycle detected while evaluating inherited configurations. Inheritence path: %s" );
```
gunnarmorling: s/inheritence/inheritance/

gunnarmorling: So we support cascades of config classes as well now? Is this also part of this change?

processor/src/test/java/org/mapstruct/ap/test/template/SourceTargetMapperErroneouslyAnnotated1.java
```diff
@@ -1,51 +0,0 @@
...
-import org.mapstruct.Mapping;
-import org.mapstruct.MappingTarget;
-import org.mapstruct.Mappings;
-import org.mapstruct.factory.Mappers;
-
-/**
- * @author Sjaak Derksen
- */
-@Mapper
-public interface SourceTargetMapperErroneouslyAnnotated1 {
```
gunnarmorling: What happened to all these removed tests and test mappers? Are they not required any longer?

gunnarmorling: It would be nice to have some tests for erroneous situations. I guess you planned to add those anyways?

core-common/src/main/java/org/mapstruct/MappingInheritanceStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+ * {@link Mapper#config()}.
+ *
+ * @author Andreas Gudian
+ */
+public enum MappingInheritanceStrategy {
+    /**
+     * Apply the {@code @Mapping} configuration only if the prototype method is explicitly referenced using
+     * {@link InheritConfiguration}.
+     */
+    WHEN_EXPLICITLY_REFERENCED,
```
gunnarmorling: I find these enum value names a bit verbose. Maybe just EXPLICIT and AUTO? The JavaDocs precisely describe the semantics. Just a feeling.

processor/src/main/java/org/mapstruct/ap/model/source/SourceMethod.java
```diff
@@ -61,10 +64,14 @@
     private final Accessibility accessibility;
     private final List<Type> exceptionTypes;
     private final MapperConfig config;
+    private final MappingOptions mappingOptions;
+    private final List<SourceMethod> prototypeMethods;
```
gunnarmorling: What are the "prototype methods" here? All the methods declared on the config class(es)? Or just matching ones?

processor/src/main/java/org/mapstruct/ap/model/source/SourceMethod.java
```diff
@@ -289,84 +302,31 @@ public Accessibility getAccessibility() {
...
-        // now, if they match they should have the same source parameter types each
-        for ( Integer count : test.values() ) {
-            if ( count != 0 ) {
-                return false;
-            }
-        }
-
-        // finally check the return type.
-        return  equals( getResultType(), method.getResultType() );
+    public boolean canInheritFrom(SourceMethod method) {
```
gunnarmorling: Nice!

processor/src/main/java/org/mapstruct/ap/model/source/SourceMethod.java
```diff
@@ -552,17 +502,6 @@ public boolean isStatic() {
         return executable.getModifiers().contains( Modifier.STATIC );
     }
 
-    private void increase(Type key, Map<Type, Integer> test) {
```
gunnarmorling: Apparently I'm not working enough on the code base. You guys are removing stuff which I didn't even know it was in there ;)

processor/src/main/java/org/mapstruct/ap/model/source/SourceReference.java
```diff
@@ -287,4 +289,20 @@ public Type getType() {
         }
 
     }
+
+    public SourceReference copyForInheritanceTo(SourceMethod method) {
```
gunnarmorling: A comment on this method's purpose would be nice.

processor/src/main/java/org/mapstruct/ap/processor/MapperCreationProcessor.java
```diff
@@ -356,6 +344,61 @@ else if ( method.isEnumMapping() ) {
...
+
+        MappingOptions mappingOptions = method.getMappingOptions();
+        List<SourceMethod> applicablePrototypeMethods = method.getApplicablePrototypeMethods();
+
+        MappingOptions inverseMappingOptions =
+            getInverseMappingOptions( availableMethods, method, initializingMethods, mapperConfig );
+
+        MappingOptions templateMappingOptions =
+            getTemplateMappingOptions(
+                concat( availableMethods, applicablePrototypeMethods ),
```
gunnarmorling: I think `join` or `union` would be a better name.

processor/src/main/java/org/mapstruct/ap/processor/MethodRetrievalProcessor.java
```diff
@@ -80,19 +93,60 @@ public int getPriority() {
...
+
+        TypeElement typeElement = asTypeElement( typeMirror );
+        List<SourceMethod> methods = new ArrayList<SourceMethod>();
+        for ( ExecutableElement executable : getAllEnclosedExecutableElements( elementUtils, typeElement ) ) {
+
+            ExecutableType methodType = typeFactory.getMethodType( typeElement, executable );
+            List<Parameter> parameters = typeFactory.getParameters( methodType, executable );
+            boolean containsTargetTypeParameter = SourceMethod.containsTargetTypeParameter( parameters );
+
+            // prototype methods don't have prototypes themselves
```
gunnarmorling: I don't mind this restriction atm., but I don't think it's generally required, right? In a multi-level inheritance hierarchy it make sense to have it.

core-common/src/main/java/org/mapstruct/MappingInheritanceStrategy.java
gunnarmorling: To be sure, `AUTO` only allows to omit `@InheritConfiguration`, whereas a reverse method is always required to be linked explicitly via `@InheritInverseConfiguration`, right?

processor/src/test/java/org/mapstruct/ap/test/inheritfromconfig/CarMapperWithAutoInheritance.java
```diff
@@ -26,24 +25,24 @@
...
  */
-@Mapper
-public interface SourceTargetMapperErroneouslyAnnotated2 {
-
-    SourceTargetMapperErroneouslyAnnotated2 INSTANCE
-            = Mappers.getMapper( SourceTargetMapperErroneouslyAnnotated2.class );
+@Mapper(
+    config = AutoInheritedConfig.class
+)
+public interface CarMapperWithAutoInheritance {
```
gunnarmorling: Is there a test for auto-inheriting from a method declared on the mapper itself rather than on a config class?

gunnarmorling:
> But of course: it's generally nicer to see the changes for review comments in a new commit to simplify the review iterations. The one merging the PR should just not forget to squash them later on...

+1 That'd be my preference.

gunnarmorling:
@agudian Added some comments/questions. This looks really great overall!

By allowing to declare prototypes on config classes and opting-in for auto-inheritance I don't see any general road blocks with this issue any more. The discussion took us quite a bit, but I definitely think we got something now which is consistent and very usable. A big thank you for driving this home and not letting go hope :)

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -112,4 +112,15 @@
      * @return The strategy to be applied when {@code null} is passed as source value to the methods of this mapper.
      */
     NullValueMappingStrategy nullValueMappingStrategy() default NullValueMappingStrategy.DEFAULT;
+
+    /**
+     * The strategy to use for applying {@code @Mapping} configurations of prototype methods in the interface specified
```
gunnarmorling: I'm wondering, is it restricted to bean mapping methods actually or also iterable/map mapping methods? The usefulness is much higher for the former, but I'm wondering whether it might make sense to allow to inherit configurations for the other types, too, at some point?

core-common/src/main/java/org/mapstruct/MapperConfig.java
agudian: Yes, I'll update the javadoc. 
Although the restriction that only interfaces or abstract classes are valid config sources now is not something that is currently technically required or checked. Let's do this in a seperate issue. OK?

core-common/src/main/java/org/mapstruct/MappingInheritanceStrategy.java
agudian: > To be sure, AUTO only allows to omit @InheritConfiguration, whereas a reverse method is always required to be linked explicitly via @InheritInverseConfiguration, right?

Exactly. :)

processor/src/main/java/org/mapstruct/ap/util/Message.java
agudian: > So we support cascades of config classes as well now? Is this also part of this change?

No, not yet. Chains of inheritance are supported, but there can be not more than one config class currently. It may however inherit prototype methods from super classes / interfaces, though - not that I forced this, I merely re-used the method-detection we have in place already and that one finds methods also on super types.

processor/src/test/java/org/mapstruct/ap/test/template/SourceTargetMapperErroneouslyAnnotated1.java
agudian: > What happened to all these removed tests and test mappers? Are they not required any longer?

They are not required anymore, as the restriction they tried to break/test is gone.

> It would be nice to have some tests for erroneous situations. I guess you planned to add those anyways?

Right, they're on the list of missing tests... :wink:

core-common/src/main/java/org/mapstruct/MapperConfig.java
agudian: OK with me... :)

core-common/src/main/java/org/mapstruct/MappingInheritanceStrategy.java
agudian: Makes sense together with your suggested renaming of the annotation property. My aim was to use names that could be read fluently and then already explained what they did. So verbose was what I was looking for.
But your shorter suggestions are more in line with the exisiting properties and enums. I don't have any strong preferences, so I'll change them... :grin:

processor/src/main/java/org/mapstruct/ap/model/source/SourceMethod.java
agudian: ```prototypeMethods``` are all the available prototype methods in the current scope. If needed, the list ```applicablePrototypeMethods``` is initialized and contains only those that are applicable.

processor/src/main/java/org/mapstruct/ap/processor/MethodRetrievalProcessor.java
agudian: Allowing prototype-style inheritance on prototype methods could invite ambiguous matches in the AUTOMATIC mode. If we find some nice semantics for dealing with that (or living with it), we could surely relax that later on. Especially when we actually allow multi-level config-classes.

processor/src/test/java/org/mapstruct/ap/test/inheritfromconfig/CarMapperWithAutoInheritance.java
agudian: Methods from the same mapper would not be pulled in automatically, but only using ```@InheritConfiguration```.

core-common/src/main/java/org/mapstruct/Mapper.java
agudian: That comes from writing the javadoc before actually diving into the implementation - because as it is now, it should somehow work for iterables and maps as well. 
Tests for them are also on my list.

agudian:
Thank you guys for your reviews. I'll make the requested modifications and add some more of those promised tests tonight. Or so. :wink:

agudian:
OK guys, I've pushed another commit. Let me know if I missed something.

As soon as you're OK with it, I'll rebase and squash... :-)

processor/src/main/java/org/mapstruct/ap/processor/MethodRetrievalProcessor.java
gunnarmorling: Could you give an example for such an ambiguity?

IIUC, you allow multi-level config inheritance for methods declared on actual mapper interfaces (no matter whether those methods are defined in one interface or an interface and a super-interface). If there is an ambiguity, it needs to be explicitly resolved by specifying `@InheritConfiguration`. Is there any real difference to the case where the method to inherit the config from is not defined on the mapper but on a config class?

I think both cases should be handled equally. Defining methods on config classes should be a rare cases. It's only needed if a) the return type is an abstract class or b) you don't want that method be usable as mapping method. But in all other cases I'd expect methods to inherit config from on actual mappers.

processor/src/test/java/org/mapstruct/ap/test/inheritfromconfig/CarMapperWithAutoInheritance.java
gunnarmorling: Ah, I see. Is there a need for this restriction? I think there is no fundamental difference between the two cases: a) "prototype methods" declared on a config class or b) prototype methods declared on the mapper or it supertypes. So IMO we should handle them equally.

processor/src/main/java/org/mapstruct/ap/processor/MethodRetrievalProcessor.java
sjaakd: > Defining methods on config classes should be a rare cases.

Actually, I don't agree here. I think that this will be a prime use case for me.

> It's only needed if a) the return type is an abstract class or b) you don't want that method be usable as mapping method.

and c) In the very near future, The method to which it applies is a forged method: MapMapping and IterableMapping (as existing today) and PropertyMapping ( #60 ) as proposed by Andreas. :smile:. 

I also don't understand why the return type per-se should be abstract. A lot of JAXB generated code produces non-abstract classes that serve as extension base all over the place (and don't have the attribute 'abstract="true"'. Do we exclude those from this methodology?

gunnarmorling: > I think that this will be a prime use case for me.

Out of interest, why would you prefer to declare prototype methods (= methods to be used as configuration source) in config classes rather than actual mappers?

> I also don't understand why the return type per-se should be abstract.

No, it's not that prototype methods in config classes *have* to declare an abstract return type. What I mean is: I would declare prototype methods in actual mappers whenever it's possible and feasible. I would only declare them in config classes in two cases:

* The config I wish to make inheritable refers to an abstract return type. In this case I cannot declare that method on a mapper, as no implemention can be generated for this method. Thus I'd use a config class
* I don't want to make the prototype method usable for clients of the mapper class (e.g. if I want to achieve that only mappings for actual sub-classes in a hierarchy are invoked). In that case I'd put it into the config class to "hide" it from mapper users.

Otherwise I'd prefer methods on actual mappers for the sake of comprehensability.

> In the very near future, The method to which it applies is a forged method

I'm not sure what you mean. Could you give an example for this case? A user can declare (and configure) e.g. an element mapping method which then is used by an iterable mapping method. Why would they declare that method on a config class rather than a mapper?

sjaakd: > The config I wish to make inheritable refers to an abstract return type.

What I meant is that a return type does not have to be abstract to be in an inheritance relation. The only way to have inheritance relation between return types in mapping method is as the base return type is in a mapping method in the `MappingConfig` class AFAIK.


> I'm not sure what you mean. Could you give an example for this case? A user can declare (and configure) e.g. an element mapping method which then is used by an iterable mapping method. Why would they declare that method on a config class rather than a mapper?

Of course that's possible (to be honest, I realized that later as well). But, perhaps the similar reason as you state above. To render a direct mapping invisible to the user. The generated forged method could then be private. If you map a object hierarchy, its kind of strange that you have to define all branch and leaves in an interface while you only require a mapping method at root level.

gunnarmorling: > The only way to have inheritance relation between return types in mapping method is as the base return type is in a mapping method in the MappingConfig class AFAIK.

Why would that be? `@InheritConfiguration` can refer to methods declared on the same mapper (that was its purpose prior to this PR). I'm confused, maybe we are talking about different things?

> But, perhaps the similar reason as you state above. To render a direct mapping invisible to the user. The generated forged method could then be private.

Yes, ok. Although it's seems a very special scenario. Why would one expose an iterable mapping method but not the element mapping method? I don't mind if we do not support this, at least for now. There is nothing wrong with it per se, only I think it'll hardly be used in practice.

agudian: Now I'm confused as well: I didn't intend to pull in the automatic inheritance for mapping methods in the same mapper again that we had before we introduced ```@InheritConfiguration```. My gut-feeling (no examples available) is that by also "auto"-inheriting mapping-configuration from the current mapper's methods, we increase the likelyhood of ambiguities and, and thus render the AUTO flag useless. 

Oh, I have an example after all: imagine you specify mappings for base types (abstract or not, these are types that you wouldn't use on their own) in the MapperConfig interface and want to auto-inherit them. In your mapper method, you have ```Car toCar(CarDto s)``` and ```void intoCar(CarDto s, @MappingTarget Car t)``` - if the current logic of auto-inheritance was applied, but would include the methods in the current mapper as well, then each of the methods would find two candidates for the auto-inheritance: the one regarding the base types in the MapperConfig interface and the sibling method defined in the mapper.

> Defining methods on config classes should be a rare cases.

I also see it more like Sjaak: I would have one or two prototype methods covering the base-types in the mapper-config interface to be consumed with auto-inheritance. All those other mapper classes use that mapperConfig and could then only mind their own business. Our Dtos / Entities don't span out a class hierarchy where it would make sense to have mapping methods for types "in between".
So from my perspective, inheriting mappings from methods with super-types within the same mapper would be the rare case. Funny :smile: ...

Now what do we do here... From where I'm standing, this is a change that could use some more thoughts... E.g. an additional stragety to allow being configured? Or would it make sense to select the "nearest" match for the auto-inheritance and continue inheriting transively instead of complaining about ambiguities too early?

Would it be OK to discuss / change that within a separate issue?

agudian:
> Did you check my comment:
> 'I think that Gunnar is also planning to copy some of the settings as processor settings see #263 (as top level). So then they probably have to go back..` on mapstruct@9d850bb ?

Sorry, yes I saw that and forgot to answer: My thought was that these plans are still a bit shaky and we don't know yet in which direction we might go. I preferred having an API that is readable and understandable without having to read too much javadoc and thought, well, if we don't currently need the DEFAULT here, why let the user have to read and follow one more link here... As soon as we know what to do with #263, you're right, it would probably have to go back again. Who knows when that will be :wink:.

But I wouldn't mind to revert it again, if you think that the API being "stable" in that regard would be more beneficial.

sjaakd:
> But I wouldn't mind to revert it again, if you think that the API being "stable" in that regard would be more beneficial.

Stability is overrated :smile: . Please do as you like.

processor/src/main/java/org/mapstruct/ap/processor/MethodRetrievalProcessor.java
sjaakd: > Would it be OK to discuss / change that within a separate issue?

Reading all of the above, I'm not sure yet that we have an issue (ok. perhaps a communication problem :smiley: ).

1. the current solution was actually made to inherit the configuration for an 'update' method from a 'create' method (just as you use as an example above).

2. The current solution (I just checked, see method `SourceMethod#isSimilar`) has an `equals` method on the last line for checking result types. So no inheritance over there.

3. What the new solution introduces is defining an inheritance relation for the result type. My original concern: I would not like to limit that to abstract result types only (but I think that Gunnar agrees with that).

4. I agree with Andreas that the `AUTO` mechanism should not apply to mapping methods within one `Mapper` but only to methods in the `MappingConfig` class. The example (update / create method) is a nice example why it would fail and lead to ambiguity if you did.

5. Wrt to the remark:

> Yes, ok. Although it's seems a very special scenario.

Is it? I would say that an interface is normal behavior is to hide the implementation. So only have a 'create' and 'update' for the root level element and hiding all the other branch and leaf mappings would actually make sense. What happens now is that I have to consider MapsStruct logic (the `Mapper` interface) as internal-implementation logic, wrap it in some kind of class and use another interface to that class to control the access to it and have something that I can actually manage (lifecycle wise). So, take for instance  OSGi why couldn't only the `Mapper` be  exported and we keep the generated implementation and the `MapperConfig` internal?

gunnarmorling: > In your mapper method, you have Car toCar(CarDto s) and void intoCar(CarDto s, @MappingTarget Car t) - if the current logic of auto-inheritance was applied, but would include the methods in the current mapper as well...

True, that case also came to my mind as a problem.

Ok, so I understood the AUTO mode to be more general than you guys had it in mind apparently. Given that I'm not big fan of it altogether I don't mind to limit its scope :) I feel AUTO is indeed a too generic name in this case, though (sorry for the confusion, as said, I understood you'd mean to auto-inherit *all* matching mappings). Maybe AUTO_INHERIT_FROM_CONFIG? That'd leave the door open for other flavors of auto-configuration down the road if needed.

gunnarmorling: So to make sure we're all on the same page:

* The default setting is EXPLICIT
* If AUTO_INHERIT_FROM_CONFIG is enabled (either on the mapper itself or the config class), configurations from methods declared in the config class are automatically inherited to methods on the mapper with assignable parameter/return types
* If more than one method from the config class are suitable candidates to inherit mappings from, an ambiguity error is raised for the concerned mapping method(s)
* `@InheritConfiguration` / `@InheritInverseConfiguration` needs to be given explicitly to inherit config from other methods on the same mapper or its super-types

Is this what we all agree on? If so, is there anything left to decide on?

One minor issue I see is that one needs to declare a base mapping method which actually should be usable by mapper clients and whose configuration should be auto-inherited by other methods, twice: once on the config class (to make it auto-inheritable) and another time on the mapper (to make it invocable). I can live with that, I wouldn't expect such method to be intended as usable in the most cases anyways.

gunnarmorling: >  I would say that an interface is normal behavior is to hide the implementation. So only have a 'create' and 'update' for the root level element and hiding all the other branch and leaf mappings would actually make sense. 

Dunno, not sure whether I'd subscribe to that for different mapping methods for one model. I don't see an actual case where one really would need this sort of thing. Esp. if it comes to iterable mappings and their element mapping method.

Btw. you already can achieve that if really needed by declaring your "internal" mappings on another mapper which you don't expose and import via `uses()`.

> So, take for instance OSGi why couldn't only the Mapper be exported and we keep the generated implementation and the MapperConfig internal?

Ah, yes, that's something I'd definitely find usefulness :) We should actually provide a way for generating mapper implementations in a specific package to facilitate this case. Could you open an issue for that? Thx!

gunnarmorling:
> But I wouldn't mind to revert it again, if you think that the API being "stable" in that regard would be more beneficial.

I don't have an opinion on that one :) Is there any other case which we can take as an example? Then we should do it similarly.

agudian:
> Is this what we all agree on?

Yes

> If so, is there anything left to decide on?

Nope, unless:

> > But I wouldn't mind to revert it again, if you think that the API being "stable" in that regard would be more beneficial.
> 
> I don't have an opinion on that one :) Is there any other case which we can take as an example? Then we should do it similarly.

I don't have another case at hand and would therefore leave it as proposed in the PR.

So I'll rename the enum value once again. Shall I push that as one more commit, or can I rebase and merge directly?

agudian:
Alrighty, there's the single commit above and I have a rebased and squashed version of this PR locally, waiting for your go to push upstream... :)

gunnarmorling:
Cool. I'll push it tonight.

agudian:
> Cool. I'll push it tonight.

OK, if you wanna do it, then I push the rebased / squashed version to my branch. I'll start witch checking what needs to be updated for the website.

core-common/src/main/java/org/mapstruct/MapperConfig.java
gunnarmorling: Created #477 for that.

processor/src/main/java/org/mapstruct/ap/processor/MethodRetrievalProcessor.java
gunnarmorling: Created #472 for generating mapper implementations within another package than the declaration.

gunnarmorling:
Merged. Many thanks, @agudian! I've added some more commits to polish comments etc. I've also created some follow-up issues.

Very nice to have this feature in finally :)

agudian:
> Merged. Many thanks, @agudian! I've added some more commits to polish comments etc.

Great, thank you! I've added another commit to fix a tiny typo.

Issue #459, opened 2015-02-12
Issues found implementing #339


sjaakd:
I found that some arguments in `String.format` were missing in `MAPMAPPING_KEY_MAPPING_NOT_FOUND` `MAPMAPPING_VALUE_MAPPING_NOT_FOUND` and `ITERABLEMAPPING_MAPPING_NOT_FOUND`. I need to check more thoroughly if more gaps are present when implementing #339

PR #464, opened 2015-02-17
#459 error message parameter forgotten

sjaakd:
should speak for itself :smile:

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/util/Message.java
+++ b/processor/src/main/java/org/mapstruct/ap/util/Message.java
@@ -50,11 +50,11 @@ public enum Message {
 
     CONSTANTMAPPING_MAPPING_NOT_FOUND( "Can't map \"%s %s\" to \"%s %s\"." ),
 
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/model/IterableMappingMethod.java
```diff
@@ -111,7 +111,10 @@ public IterableMappingMethod build() {
             );
 
             if ( assignment == null ) {
-                ctx.getMessager().printMessage( method.getExecutable(), Message.ITERABLEMAPPING_MAPPING_NOT_FOUND );
+                ctx.getMessager().printMessage( method.getExecutable(),
```
gunnarmorling: Do you think we really need to add the method name? The error marker will be raised right at the concerned method, so I don't think that's necessary.

I'd just say "No implementation can be generated for this method. Found no method nor implicit conversion for...".

sjaakd: No.. I just restored what was there in the first place.. I'll remove / update it according your comment.

sjaakd:
@gunnarmorling : can I merge?

gunnarmorling:
Merged. Thanks, @sjaakd!

Issue #460, opened 2015-02-13
Eclipse APT Incremental: generation does not work in some cases


agudian:
I have tested around with using MapStruct in Eclipse during the incremental compilation. Although the clean build always works, the incremental generation sometimes leads to errors in the generator (e.g. unable to map from a to b...), as if the "used" mappers defined in the mapper-config interface were not present.

So I stared a debug session with my big original workspace and found some funny behaviour:

At some places we transform a TypeMirror to a TypeElement using ```(TypeElement) ( (DeclaredType) mirror ).asElement()```. That also works fine in Eclipse. But if you then call ```#getEnclosedElements()``` on that TypeElement, you could end up with an empty list. The reason is that in some cases the TypeElement may be backed in Eclipse by an UnresolvedTypeBinding instead of a SourceTypeBinding or a BinaryTypeBinding. And funny enough, Eclipse does not try to resolve the binding upon usage.

We have two options, I'd say:
1. we check every point where we use  ```(TypeElement) ( (DeclaredType) mirror ).asElement()``` and instead do: ```elementUtils.getTypeElement( ( (TypeElement) ( (DeclaredType) usedMapper ).asElement() ).getQualifiedName().toString() )``` (might have an impact on the performance, though)
2. we fix that in Eclipse... and wait for the fix to be applied... and wait for the release to come...

Any suggestions?

gunnarmorling:
I would say if it can be fixed properly in Eclipse that should be the way to go. Of course we could do 1. as a work-around for the time being. We might extract a helper for this which falls back to `elementUtils#getTypeElement()...` if we don't find any enclosed methods the other way.

agudian:
> We might extract a helper for this which falls back to elementUtils#getTypeElement()... if we don't find any enclosed methods the other way.

Good idea, did that. :+1: 

For the Eclipse patch, I'll try to do that at some point.

gunnarmorling:
@agudian, I guess that one can be closed, right?

agudian:
Right. Forgot that. :+1:

PR #468, opened 2015-02-21
#460 add workaround for TypeElements that occasionally seem "empty" during incremental compilation in Eclipse JDT

agudian:
... no tests included... but I tried it with my biggy workspace here... ;)

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/model/common/TypeFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/common/TypeFactory.java
@@ -58,6 +58,7 @@ import org.mapstruct.ap.util.AnnotationProcessingException;
 import org.mapstruct.ap.util.SpecificCompilerWorkarounds;
 
 import static org.mapstruct.ap.util.SpecificCompilerWorkarounds.erasure;
+import static org.mapstruct.ap.util.SpecificCompilerWorkarounds.replaceTypeElementIfNecessary;
 
 /**
  * Factory creating {@link Type} instances.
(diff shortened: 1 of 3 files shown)


agudian:
@sjaakd / @gunnarmorling: do you guys have any opinion on this one? In case you wonder: testing this automatically is next to impossible, and recreating the issue in Eclipse requires a bit of luck as well...

gunnarmorling:
@agudian I think the change itself is good, I got a question on the general desing of `SpecificCompilerWorkarounds`, though.

Shouldn't this be an interface, with different implementations (one for javac, one for Eclipse) and we use the right one depending on the environment? That way e.g. your fix would only be part of the Eclipse impl, whereas the javac-based impl would be a no-op. Would that make sense?

agudian:
> Shouldn't this be an interface, with different implementations (one for javac, one for Eclipse) and we use the right one depending on the environment? That way e.g. your fix would only be part of the Eclipse impl, whereas the javac-based impl would be a no-op. Would that make sense?

Yes, that would make sense. Up until now we created workarounds that worked for both, javac and Eclipse. To make your suggestion work, we'd need to make sure that our detection mechanism for Eclipse / Javac is rock-solid, though... I wouldn't trust the current implementation in ```DefaultVersionInformation``` that far... :wink: - The ramifications of a false-negative detection of the eclipse compiler are pretty severe, in that the processor would not work properly. That's something I wouldn't try to squeeze in for Beta4.

gunnarmorling:
Rebased and applied. Thx, @agudian!

gunnarmorling:
Created #484 for making this an interface as discussed above.

PR #471, opened 2015-02-23
#160 part 2, basic property mappings

sjaakd:
I limited the solution to basic property mappings, so:
1. No collection / array types / adders -> we need to discuss this more thoroughly what an update method means in this context.
2. No 2 step (method, method) (conversion, method) (method, conversion) selection for update methods.
3. Also `MapMappings` and `IterableMappings` are not considered (excluded from solution)..

In short: we probably need some follow on PR's to cover the complete scope.
@gunnarmorling : I think you're right were it comes to the wrapper mechanism. Its nearing its end of use and we should at one point see where we want to go with a more comprehensible solution. I don't believe in templates all the way, but perhaps the project you showed some time ago, or perhaps plain java (and good-old a StringBuffer) is more appropriate.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/PropertyMapping.java
@@ -61,7 +61,8 @@ public class PropertyMapping extends ModelElement {
 
     private final String name;
     private final String sourceBeanName;
-    private final String targetAccessorName;
+    private final String targetWriteAccessorName;
+    private final String targetReadAccessorName;
     private final Type targetType;
     private final Assignment assignment;
     private final List<String> dependsOn;
@@ -71,7 +72,8 @@ public class PropertyMapping extends ModelElement {
         // initial properties
         private MappingBuilderContext ctx;
         private SourceMethod method;
-        private ExecutableElement targetAccessor;
+        private ExecutableElement targetWriteAccessor;
+        private ExecutableElement targetReadAccessor;
         private String targetPropertyName;
         private String dateFormat;
         private List<TypeMirror> qualifiers;
@@ -90,8 +92,13 @@ public class PropertyMapping extends ModelElement {
             return this;
         }
 
-        public PropertyMappingBuilder targetAccessor(ExecutableElement targetAccessor) {
-            this.targetAccessor = targetAccessor;
+        public PropertyMappingBuilder targetReadAccessor(ExecutableElement targetReadAccessor) {
+            this.targetReadAccessor = targetReadAccessor;
+            return this;
+        }
+
+        public PropertyMappingBuilder targetWriteAccessor(ExecutableElement targetWriteAccessor) {
+            this.targetWriteAccessor = targetWriteAccessor;
             return this;
         }
 
@@ -130,8 +137,7 @@ public class PropertyMapping extends ModelElement {
             return this;
         }
 
-        private enum TargetAccessorType {
-
+        private enum TargetWriteAccessorType {
             GETTER,
             SETTER,
             ADDER
@@ -140,18 +146,18 @@ public class PropertyMapping extends ModelElement {
         public PropertyMapping build() {
 
             // handle target
-            TargetAccessorType targetAccessorType = getTargetAcccessorType();
+            TargetWriteAccessorType targetAccessorType = getTargetAcccessorType();
             Type targetType = getTargetType( targetAccessorType );
 
             // handle source
             String sourceElement = getSourceElement();
             Type sourceType = getSourceType();
             String sourceRefStr;
-            if ( targetAccessorType == TargetAccessorType.ADDER && sourceType.isCollectionType() ) {
+            if ( targetAccessorType == TargetWriteAccessorType.ADDER && sourceType.isCollectionType() ) {
                 // handle adder, if source is collection then use iterator element type as source type.
                 // sourceRef becomes a local variable in the itereation.
                 sourceType = sourceType.getTypeParameters().get( 0 );
-                sourceRefStr = Executables.getElementNameForAdder( targetAccessor );
+                sourceRefStr = Executables.getElementNameForAdder( targetWriteAccessor );
             }
             else {
                 sourceRefStr = getSourceRef();
@@ -208,19 +214,21 @@ public class PropertyMapping extends ModelElement {
             return new PropertyMapping(
                 targetPropertyName,
                 sourceReference.getParameter().getName(),
-                targetAccessor.getSimpleName().toString(),
+                targetWriteAccessor.getSimpleName().toString(),
+                targetReadAccessor != null ? targetReadAccessor.getSimpleName().toString() : null,
                 targetType,
                 assignment,
                 dependsOn
             );
         }
 
-        private Assignment assignObject(Type sourceType, Type targetType, TargetAccessorType targetAccessorType,
+        private Assignment assignObject(Type sourceType, Type targetType, TargetWriteAccessorType targetAccessorType,
                                         Assignment rhs) {
 
             Assignment result = rhs;
 
-            if ( targetAccessorType == TargetAccessorType.SETTER ) {
+            if ( targetAccessorType == TargetWriteAccessorType.SETTER ) {
+
                 result = new SetterWrapper( result, method.getThrownTypes() );
                 if ( !sourceType.isPrimitive()
                     && !sourceReference.getPropertyEntries().isEmpty() /* parameter null taken care of by beanmapper */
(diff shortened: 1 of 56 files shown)


sjaakd:
Hmm.. I see that #459 accidentally ended up here. I'll remove it tonight..

sjaakd:
I relaxed the conditions a little bit. I do allow 2 step mapping methods, and conversions, only when the inner one is a createMethod or conversion (and the outer an update method).

Also added more unit test. Is ready for review now.

gunnarmorling:
Hi @sjaakd, could you rebase it to the latest master now that Andreas' config inheritance stuff is in? Going to look at it asap.

Btw. that one commit touches many files and code lines. Any chance you can split it up? E.g. into one commit with refactorings/renamings you did and another one with the "actual" changes? For instance that renaming around accessors could be its own commit. It's basically independent, so one can quickly understand and go over it when reviewing. Then another commit with the actual change would be much smaller already and easier to comprehend. Also you can speficy the motivation for each part of the change if it has its own commit message.

Generally it's hard to digest and understand the "meat" of a change if the commits are that big. I cannot give you a fixed upper limit, but I've experienced that usually well-defined and focused commits are more in the hundreds of lines at most and also don't touch several dozes of files. Of course there always can be exceptions.

sjaakd:
Hmm. Seems that you have a copy of the text above available :smile: . Don't get me wrong on that, you're right. 

This PR came into being by some iterating hence-and-forth, hence it ended up in one commit and, if you re-factor a method name as part of your that you're bound to end up with a lot of impacted classes.

On the positive side: I did already 160-1 as part of #160 last week. So, I'm trying to meet that expectation.

I did the rebase. What'll do is handle the re factoring part (and bringing the read accessors to the templates) separately as one commit early this evening, as you propose and stack the 'meat' in a separate commit on top of that.

gunnarmorling:
> I did the rebase. What'll do is handle the...

Cool, much appreciated.

I know how it is, no worries. It also took me a bit to really take advantage of the powers of git. Nowadays I often extract commit after commit from a "working set" of changes in my work space. Also makes the work easier actually because it gets independent changes out of your sight. I also often do interactive rebases and add stuff to previous commits if I notice it makes more sense there. It's a bit of a different mode of working, but I got to like it quite a lot :smile:

sjaakd:
Ok.. Done.. 

> Nowadays I often extract commit after commit from a "working set" of changes in my work space.

You someday have to show me how you do that in an easy way... It takes me some fiddling, but I managed:
1. a new branch created 160-2a from the old one 160-2
2. removing the unwanted stuff to the base of the refactoring (original big commit).
3. committing that stuff into a new commit (changes on big commit).
4. squashing it (to the refactoring commit) 
5. cherry-picking the original big commit from the old branch
6. removing the old branch 160-2
7. moving the new branch to the old one 160-2a -> 160-2

gunnarmorling:
@sjaakd, is there support for collection-typed members already? I.e. if there is an update method for the element type of an iterable, will it be invoked?

sjaakd:
> @sjaakd, is there support for collection-typed members already? I.e. if there is an update method for the element type of an iterable, will it be invoked?

Nope.. I left it out for the time being. Checkout the lines:
```java
            boolean preferUpdateMethods;
            if ( targetType.isCollectionOrMapType() ||  targetType.isArrayType() ||
                targetAccessorType == TargetWriteAccessorType.ADDER ) {
                preferUpdateMethods = false;
            }
            else {
                preferUpdateMethods = method.getMappingTargetParameter() != null;
            }
```

Collections and Maps are tricky anyway. E.g. what are we actually updating. In our iterable- and mapmaking methods we clear the existing collection / map and then start to add.  Imagine if we would support update methods on elements / entries. We would somehow have to figure out which object in the source collection would match which target object in the target collection and then update. Not to speak of which ones to delete or create.

Hence I think we should do the following:
1. First this solution
2. Then allow referencing update methods in used mappers (and hand written ones). (I hope I can still do that as part of this beta4, should not be too difficult).
3. Then support selecting of iterable / map mapping methods (adders will not be supported).
4. Stop....  (as in, do not tackle element / entry update mappings).

Note: the 2 step matching mechanism also does not support every option.

sjaakd:
and rebased.....

sjaakd:
found one issue.. `GetterWrapperForCollectionsAndMaps` requires an input for target type.

sjaakd:
added the selection for external mappers in this PR. Just to safe some time..

gunnarmorling:
> We would somehow have to figure out which object in the source collection would match which target object in the target collection and then update. Not to speak of which ones to delete or create.

Ah, I see what you mean. Not sure indeed how this could be done. It might work for lists/arrays (assuming they have the same size in source and target) and probably maps, but it will be a problem for sets.

gunnarmorling:
> found one issue.. GetterWrapperForCollectionsAndMaps requires an input for target type.

Is this fixed since then?

sjaakd:
> Is this fixed since then?

Yep..

processor/src/main/java/org/mapstruct/ap/model/assignment/UpdateWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+package org.mapstruct.ap.model.assignment;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import org.mapstruct.ap.model.common.Type;
+
+/**
+ * Wraps the assignment in a target setter.
```
gunnarmorling: Maybe you could add a sample of the output it generates (maybe a good idea for all the assignments in general)? Just an idea, no biggie.

processor/src/main/java/org/mapstruct/ap/model/PropertyMapping.java
```diff
@@ -130,23 +138,33 @@ public PropertyMappingBuilder existingVariableNames(Collection<String> existingV
...
                 // sourceRef becomes a local variable in the itereation.
                 sourceType = sourceType.getTypeParameters().get( 0 );
-                sourceRefStr = Executables.getElementNameForAdder( targetAccessor );
+                sourceRefStr = Executables.getElementNameForAdder( targetWriteAccessor );
             }
             else {
                 sourceRefStr = getSourceRef();
             }
 
+            // all the tricky cases will be excluded for the time being.
```
gunnarmorling: Could you open a follow-up issue describing those? It's not that we need to implement all of them necessarily, but good to keep a record of what you consider to be missing.

gunnarmorling:
Many thanks, @sjaakd. I've added some more comments. It looks good overall.

To be sure I understand what's going on:

* It only works for mapping "simple" properties of bean mappings, i.e. no collection-typed properties etc.
* Non-update mapping methods prefer non-update methods for mapping properties, update mapping methods prefer update methods for mapping properties (in case both are present for a given property type)
* If no mapping method of the same category is available, e.g. a non-update method would invoke an update mapping method for one of its properties

Is this correct?

Finally, some questions:

* I noticed the invoking bean mapping method is instantiates a target property prior to invoking an update method in case that property is null. What's the motivation for this? I'm wondering whether this shouldn't stay null if the object graph to be updated does not have this property set?
* For this instantiation, will be factory methods invoked (if present)?

processor/src/main/java/org/mapstruct/ap/model/assignment/UpdateWrapper.java
sjaakd: yep. Good point..

sjaakd:
> It only works for mapping "simple" properties of bean mappings, i.e. no collection-typed properties etc.

Yep.

> Non-update mapping methods prefer non-update methods for mapping properties, update mapping methods prefer update methods for mapping properties (in case both are present for a given property type)

Yep. Non-update mapping methods are aka create mapping methods (waiting for better terminology).

> I noticed the invoking bean mapping method is instantiates a target property prior to invoking an update method in case that property is null. What's the motivation for this? I'm wondering whether this shouldn't stay null if the object graph to be updated does not have this property set?

It was discussed originally in #160  on jan 23rd . Basically the source tree is leading in determining what objects should be created (if they are absent in the target).  


> For this instantiation, will be factory methods invoked (if present)?

That's true.. But not tested.. I'll add a TC for that (tomorrow).

processor/src/main/java/org/mapstruct/ap/model/PropertyMapping.java
sjaakd: #487, #488, #489

sjaakd:
> If no mapping method of the same category is available, e.g. a non-update method would invoke an update mapping method for one of its properties

Yep.

sjaakd:
I'll process your comments. I'll await the merge of @agudian and do a rebase. Do you think its good enough to then?

gunnarmorling:
> Do you think its good enough to then?

Yes :) If you like merge it yourself once you're done. Thx!

agudian:
> I'll await the merge of @agudian and do a rebase. Do you think its good enough to then?

Don't exactly know what you mean, but I've merged the two other PRs by Gunnar. You can now safely rebase again and push it to master. :+1: 

PS: sorry that I didn't chime in on the discussion, I was out-of-order yesterday and, actually, also today. But I agree with all of your conclusions, including that we don't have to support everything (e.g. the merging of iterables/maps, which would usually require at least hashCode/equals to be based on some primary key values and not the whole content... at least when one wants to relate to dto/entity mappings)

gunnarmorling:
@sjaakd Can you do the rebase and merge? It's the last thing missing for Beta4 :)

processor/src/main/java/org/mapstruct/ap/model/assignment/UpdateWrapper.java
sjaakd: I did not process this comment yet. I'll do that in a separate PR also fixing the other wrappers. Possibly together (lets discuss) with some more clarifications improvements.

PR #479, opened 2015-02-25
#168 Change retention of method-level annotations to CLASS...

agudian:
... as they now need to be read from other classes as well.

As #168 is not yet closed and as it should have been part of that, I thought I'd re-use that issue number. Hope you don't mind.

I left out ```@NullValueMapping```, as a) it's not correctly supported for the inheritance currently, and b) will be removed by Sjaak shortly anyway (and then the inheritance will work with that as well).


gunnarmorling:
Merged, thx!

PR #483, opened 2015-03-01
289 adding extra null check before add/putAll of a target collection

sjaakd:
I refactored the template of `getterCollectionMapWrapperr` a bit. It used the `SetterWrapper` to do its business which was rather confusing. 

Then I added the null check and the world started to make sense once more :smile:

Also, I don't know if you guys like that approach, but the first commit added the reproducers. To avoid non compiling code, I added `@Ignore` to the test. Perhaps something we can do more often?


processor/src/main/resources/org.mapstruct.ap.model.assignment.GetterWrapperForCollectionsAndMaps.ftl
```diff
@@ -0,0 +1,53 @@
...
+        </#list>
+    </#if>
+}
+<#macro _assignmentLine>
+     <@includeModel object=ext.targetType/> <@_localVar/> = <@_assignment/>;
+    if ( <@_localVar/> != null ) {
+        ${ext.targetBeanName}.${ext.targetAccessorName}().<#if ext.targetType.collectionType>addAll<#else>putAll</#if>( <@_localVar/> );
+    }
+</#macro>
+<#macro _localVar>target<#if ext.targetType.collectionType>Collection<#else>Map</#if></#macro>
```
agudian: Won't we get into trouble again if the method already has a variable called targetMap / targetCollection defined?

processor/src/test/java/org/mapstruct/ap/test/bugs/_289/Issue289Mapper.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface Issue289Mapper {
+
+    Issue289Mapper INSTANCE = Mappers.getMapper( Issue289Mapper.class );
+
+    Target sourceTarget(Source source);
```
agudian: could you please add a collection in target that has both getter and setter, and add a mapping-method with ```@MappingTarget```? AFAIR, that's what the case in the bug report was (it wasn't mentioned in the report itself, but I know where it's coming from :wink:).

processor/src/test/java/org/mapstruct/ap/test/bugs/_289/Target.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.bugs._289;
+
+import com.sun.javafx.scene.control.skin.VirtualFlow;
```
agudian: whoops, don't we have an enforcer rule in place to disallow "restricted API" such as com.sun.*...? I guess we should at some point... :grin:

processor/src/main/resources/org.mapstruct.ap.model.assignment.GetterWrapperForCollectionsAndMaps.ftl
sjaakd: we should be safe.. this is the generated code:

```java
            Collection<TargetElement> targetCollection = sourceElementCollectionToTargetElementCollection( source.getCollection() );
            if ( targetCollection != null ) {
                target.getCollection().addAll( targetCollection );
            }
        }
```
But, if you want to be more safe, I can base it on the target property name. Using safe-variable solution is more tricky. Means bringing up a collection with used names to this level.

processor/src/test/java/org/mapstruct/ap/test/bugs/_289/Issue289Mapper.java
sjaakd: hmm. are you sure about the setter?.. the issue starts with:

"I created a mapping into a collection which is only accessible by its getter.
The generated mapper does a NULL check on the target read-only collection, but I would have expected a NULL check on the source collection too."

I'll add some additional test cases that you propose, but I reckon I wan tot keep this one as well.

processor/src/test/java/org/mapstruct/ap/test/bugs/_289/Target.java
sjaakd: wow.. didn't notice that.. You can't rely on auto completion always..

agudian:
Great work! Looks so simple when you do it... :+1: 

I only got two remarks for you above... the variable naming might require a little more effort than changing the illegal import, though... :wink: 

> Also, I don't know if you guys like that approach, but the first commit added the reproducers. To avoid non compiling code, I added @Ignore to the test. Perhaps something we can do more often?

I wouldn't mind seeing the fix and the test in one commit. But if you do split it up, I guess it would also be ok to leave it in a failing state, i.e. without the ```@Ignore``` - as you would probably never push the commits individually (which may lead to a failing build). Better than accidentally forgetting to remove the ```@Ignore``` later on :wink:

processor/src/test/java/org/mapstruct/ap/test/bugs/_289/Issue289Mapper.java
agudian: Duh, you're right. Sorry for the confusion. But the additional test could still be useful (although I guess both cases end up in the same code when generating the assignments?)

processor/src/main/resources/org.mapstruct.ap.model.assignment.GetterWrapperForCollectionsAndMaps.ftl
agudian: Don't we have all the info available when creating the assignments? The localVar name could be a property in GetterWrapperForCollectionsAndMaps that needs to be provided when creating it.

sjaakd: I tried that initially.. But it the collection maintaining the local names to be maintained on `BeanMapping` level. I reckon that's not a bad idea to do anyway..

sjaakd: Ok.. I've done that.. And for good reason.. My test initially started to fail on `org/mapstruct/ap/test/source/expressions/java/JavaExpressionTest#testGetterOnly` when I used targetPropertyName only (targetTypeName == targetPropertyName).

sjaakd:
@agudian, done. I processed your comments and added them into a new commit. (unfortunately with a wrong number, but I'll squash them anyway, if you agree with the solutions to your comments).

agudian:
Very nice! Ready to be merged :+1: 

And thanks for taking care of this issue :beers:

gunnarmorling:
I couldn't take a close look, but if you think it's good, move forward and merge yourself. Many thanks!

sjaakd:
No problem. Thanks for reviewing

Issue #485, opened 2015-03-03
ArrayCopyWrapper local var possibly not safe


sjaakd:
The local var in the `ArrayCopyWrapper` does not use all available used variables

gunnarmorling:
Done.

PR #486, opened 2015-03-03
#485 ArrayCopyWrapper local var possibly not safe

sjaakd:
Bring `existingVariableNames` to `ArrayCopyWrapper`. Problem / Solution is straightforward


gunnarmorling:
Merged. Thanks, @sjaakd!

Issue #470, opened 2015-02-23
Miss import when decorator is not in same package


AydenSekey:
I tried to use a decorator which is not in the same package that my mapper.
The import miss in the generated mapper (Impl which contain delegate, cause by extends decorator), even if i use `imports` in `@Mapper`.

agudian:
Fix was pushed to master.

PR #490, opened 2015-03-03
#470 Adding import in case decorator is in another package than mapper i...

gunnarmorling:
...nterface


agudian:
Reviewed, rebased, pushed to master. Thanks, Gunnar!

Issue #481, opened 2015-02-27
Android support


pardom:
Currently there are class dependencies (Java 8 Dates and JAXB) that cause ClassNotFound exceptions in the annotation processor when compiling for Android. Are there plans to support the Android build environment?

gunnarmorling:
Hi, we haven't specifically thought about this yet, but it should be doable.

Actually it should work already without the Java 8 date types (these should only be accessed if actually present, so MapStruct can be used on Java 6/7), so this may be a bug. Assuming we fix that, it should be possible to do similar for the JAXB classes.

Do you have stracktraces which show where the CNFEs exactly happen?

pardom:
I created an example project exhibiting the behavior I'm seeing: https://github.com/pardom/MapStructTest

When I build using `./gradlew clean assembleBeta --debug` I see this output:
```
09:46:50.125 [INFO] [org.gradle.api.internal.tasks.compile.JdkJavaCompiler] Compiling with JDK Java compiler API.
09:46:50.811 [ERROR] [system.err] error: Couldn't find type java.time.ZonedDateTime. Are you missing a dependency on your classpath?
09:46:50.854 [ERROR] [system.err] 1 error
```

agudian:
That particular error was fixed already. Could you try using our latest snapshot version?

pardom:
I just updated to the snapshot and tried again.

`./gradlew clean assembleDebug --debug --refresh-dependencies`

Produces:
```
10:32:37.704 [INFO] [org.gradle.api.internal.tasks.compile.JdkJavaCompiler] Compiling with JDK Java compiler API.
10:32:38.192 [ERROR] [system.err] error: Couldn't find type java.time.ZonedDateTime. Are you missing a dependency on your classpath?
10:32:38.233 [ERROR] [system.err] 1 error
```

agudian:
Thanks for your feedback - I saw just now that the snapshot we have on deployed is too old and did not yet contain the fix that I was thinking about. I'm deploying the current state of our master branch right now, so it would be great if you could try again (I don't have an Android SDK installed).

pardom:
Cool, that took care of the `ZonedDateTime` CNFE. Now it's just JAXB:
```
14:11:36.706 [INFO] [org.gradle.api.internal.tasks.compile.JdkJavaCompiler] Compiling with JDK Java compiler API.
[...]

PR #495, opened 2015-03-05
#481 Only add JaxbElement if the class exists

pardom:
Android doesn't JaxbElement, so we should only include it if we can create an instance of it.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/model/source/builtin/BuiltInMappingMethods.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/source/builtin/BuiltInMappingMethods.java
@@ -23,6 +23,7 @@ import java.util.List;
 import org.mapstruct.ap.model.common.TypeFactory;
 import org.mapstruct.ap.util.Collections;
 import org.mapstruct.ap.util.JavaTimeConstants;
+import org.mapstruct.ap.util.JaxbConstants;
 
 /**
  * Registry for all built-in methods.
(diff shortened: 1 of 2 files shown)


processor/src/main/java/org/mapstruct/ap/model/source/builtin/BuiltInMappingMethods.java
```diff
@@ -35,7 +36,6 @@
 
     public BuiltInMappingMethods(TypeFactory typeFactory) {
         builtInMethods = Collections.newArrayList(
-            new JaxbElemToValue( typeFactory ),
             new DateToXmlGregorianCalendar( typeFactory ),
             new XmlGregorianCalendarToDate( typeFactory ),
```
gunnarmorling: You didn't have issues with `XmlGregorianCalendar`? Is this part of the Android platform then?

pardom: Processing succeeds with it and it appears to be in the runtime.

http://developer.android.com/reference/javax/xml/datatype/XMLGregorianCalendar.html

gunnarmorling: I see. Thanks for checking!

gunnarmorling:
Rebased and applied. Thanks, @pardom!

PR #496, opened 2015-03-05
#497 Bug to illustrate problem with resolving ambigious methods

dilipkrish:
When a mapping provides a qualifier to be used to resolve a mapping method, the annotation processor
should attempt to resolve it using the provided qualifier

In the example below, there is an explicit qualifier used to denote the algorithm used to calculate multiplier, denoted by a combination of ```@SourceTypeQualifier``` and ```@DoubleMultipler``` and ```@RandomMultiplier```. Since the mapping has explicitly qualified the algorithm I would assume that it would be used to resolve the ambiguity.

Granted that there is a work around using expressions. But I'd rather have an option where we use expression as a last resort.


dilipkrish:
After some digging I realized that not having a type level annotation ```@SourceTypeQualifier``` on ```MultiplierMapper``` was the cause of the problem. Everything works as expected. Easy to miss that

Issue #272, opened 2014-07-15
Support decorators with the CDI component model


gunnarmorling:
MapStruct decorators should be mapped to actual CDI decorators when working with componentModel="cdi".

sjaakd:
@sonata82 , any progress on this? Anything I can do to help?

sonata82:
I've created a solution for CDI using [Decorator and Delegate annotations](http://docs.oracle.com/javaee/6/api/javax/decorator/package-summary.html) - or so I thought, because my integrationtest is still failing.

I don't have a good solution for Spring yet. Please feel free to have a look at [my code](https://github.com/sonata82/mapstruct/tree/cdi_decorator/) and tell me if I'm going into the wrong direction ^^

gunnarmorling:
Hi @sonata82, sorry for the late reply, I somehow missed your last comment. Could you rebase your branch to the current master and submit a pull request? Then we can see the exact change and can discuss it right in place. No worries about Spring, if we got a working solution for CDI, that'd already by a good step forwards. Thx!

gunnarmorling:
> I've created a solution for CDI using Decorator and Delegate annotations

Yes, that's the right way.

Btw. I just took a quick look at your branch, shouldn't `SourceTargetMapperDecorator` be annotated with the @Decorator annotation? And do we actually have to generate two classes in this case? I'd expect CDI to take care of marrying the mapper bean and the manually written decorator. The two classes we currently generate should only be required with the default component model where we need to take care of the decorator application ourselves.

You might start by writing all the required classes from hand so you have a working example as a basis. Then you could think about which of these should be generated by MapStruct.

gunnarmorling:
Done. Thanks a lot, @sonata82!

PR #503, opened 2015-03-15
#272 Proposed solution for support of decorators with CDI and Spring

sonata82:
This solution just throws away the decorator definition, if annotation based mappers are created. This gives support in CDI and Spring, with minimal code changes.

The solution for Spring is not very good (using `@Qualifier` with Impl_) but it works. Please feel free to comment.

Referenced issue: #272

### DIFF
--- a/integrationtest/src/test/resources/springTest/src/test/java/org/mapstruct/itest/spring/SpringBasedMapperTest.java
+++ b/integrationtest/src/test/resources/springTest/src/test/java/org/mapstruct/itest/spring/SpringBasedMapperTest.java
@@ -24,6 +24,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mapstruct.itest.spring.SpringBasedMapperTest.SpringTestConfig;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.test.context.ContextConfiguration;
@@ -45,6 +46,9 @@ public class SpringBasedMapperTest {
     @Autowired
     private SourceTargetMapper mapper;
 
+    @Autowired
+    @Qualifier( "sourceTargetMapperDecorator" )
+    private DecoratedSourceTargetMapper decoratedMapper;
 
     @Test
     public void shouldCreateSpringBasedMapper() {
@@ -56,4 +60,15 @@ public class SpringBasedMapperTest {
         assertThat( target.getFoo() ).isEqualTo( Long.valueOf( 42 ) );
         assertThat( target.getDate() ).isEqualTo( "1980" );
     }
+
+    @Test
+    public void shouldInjectDecorator() {
+        Source source = new Source();
+
+        Target target = decoratedMapper.sourceToTarget( source );
+
+        assertThat( target ).isNotNull();
+        assertThat( target.getFoo() ).isEqualTo( Long.valueOf( 43 ) );
+    }
+
 }
--- a/processor/src/main/java/org/mapstruct/ap/processor/AnnotationBasedComponentModelProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/processor/AnnotationBasedComponentModelProcessor.java
@@ -55,6 +55,10 @@ public abstract class AnnotationBasedComponentModelProcessor implements ModelEle
             return mapper;
         }
 
+        if ( shouldDecoratorBeRemoved() ) {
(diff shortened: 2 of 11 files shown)


gunnarmorling:
Thanks, @sonata82! Going to look at it asap.

gunnarmorling:
@sonata82 I've noticed that there are several merge commits on your branch. Could you rebase it to the latest master and do a force-push to this branch, so it only contains the commit(s) with the actual changes? We prefer to work with rebases rather than merges in order to keep the project history lean and digestable. Let me know if you need any help with that?

processor/src/main/java/org/mapstruct/ap/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -55,6 +55,7 @@ public Mapper process(ProcessorContext context, TypeElement mapperTypeElement, M
             return mapper;
         }
 
+        mapper.removeDecorator();
```
gunnarmorling: Should this not rather be done only in the CDI and JSR 330 implementations?

integrationtest/src/test/resources/springTest/src/test/java/org/mapstruct/itest/spring/SpringBasedMapperTest.java
```diff
@@ -24,6 +24,7 @@
 import org.junit.runner.RunWith;
```
gunnarmorling: This file (and some others) has its file mode changed. I don't think that's intentional?

processor/src/main/java/org/mapstruct/ap/processor/AnnotationBasedComponentModelProcessor.java
sonata82: Yes! Well, actually, only in the CDI and Spring implementations. 

I reworked the code a little bit and instead added a method `protected abstract boolean shouldDecoratorBeRemoved();` which all implementations return either `true` or `false` from.

integrationtest/src/test/resources/springTest/src/test/java/org/mapstruct/itest/spring/SpringBasedMapperTest.java
sonata82: I created a new commit without those annoying file mode changes.

sonata82:
I had trouble rebasing my branch, probably because I merged from `remote/master` in between. I created a completely new branch from the current HEAD and applied my changes manually. 

Looking forward to your input!

integrationtest/src/test/resources/springTest/src/test/java/org/mapstruct/itest/spring/SpringBasedMapperTest.java
```diff
@@ -45,6 +46,9 @@
     @Autowired
     private SourceTargetMapper mapper;
 
+    @Autowired
+    @Qualifier( "sourceTargetMapperDecorator" )
+    private DecoratedSourceTargetMapper decoratedMapper;
```
gunnarmorling: Is there a way to have the decorated mapper injected without a qualifier (and let the wrapped mapper be injected into the decorator using a qualifier)?

processor/src/main/java/org/mapstruct/ap/processor/CdiComponentProcessor.java
```diff
@@ -44,4 +44,9 @@ protected Annotation getTypeAnnotation() {
     protected Annotation getMapperReferenceAnnotation() {
         return new Annotation( getTypeFactory().getType( "javax.inject.Inject" ) );
     }
+
+    @Override
+    protected boolean shouldDecoratorBeRemoved() {
```
gunnarmorling: Thinking about it, should we raise a warning in case `@DecoratedWith` is used with the "cdi" component model, stating that this isn't necessary and `@DecoratedWith` will be ignored in this case?

gunnarmorling:
@sonata82, finally coming back to this. A big sorry for the long radio silence, I was involved with some other things :/ I think it looks good overall, added two more comments inline. Just let me know whether you want to handle them, otherwise I can take over and finish this up.

Thanks a lot for this contribution! It's great to see the decorator feature working for all the component models.

gunnarmorling:
Going to close this one, I have created #566 which supersedes this one. Thanks a lot for your work @sonata82! The major difference is that the decoratorated mapper can now be injected without any qualifier for Spring.

Btw. which name can I add for you to the list of contributors in copyright.txt, @sonata82?

sonata82:
Great, that looks very good now that the qualifier can be omitted!

I'm glad to help out on such a great project, you can put me down as: Remko Plantenga

Issue #517, opened 2015-03-26
Order of properties in generated code or error message is unstable


agudian:
Just stumbled upon it: the order of the generated setter calls and the order of target properties in "Unmapped target property" messages depends on HashMap iteration order. 

Unless specified otherwise with ```dependsOn```, the order should remain to be the one found in the target type.

sjaakd:
thanks @agudian

PR #518, opened 2015-03-26
#517 make order of target properties in generated code and in messages stable

agudian:


sjaakd:
no further questions, your honor :+1:

Issue #508, opened 2015-03-18
Occasionally empty generated .java file in multithreaded build


agudian:
Following up #504, which threw an NPE, I now have some empty .java files flying around. Seems like a flush is missing somewhere...? I'm investigating...

agudian:
Damn, I didn't find anything, yet. An additional flush() in IndentationCorrectingWriter.close() after currentState.onExit(..) didn't change anything, either.

@gunnarmorling: any idea what that could be?

gunnarmorling:
@agudian, do you have a specific mapper definition which triggers that
error?

agudian:
Nope, but it seems like if it happens for one maven module, then it happens for all the mappers there.

We are building with JDK 1.8.0_25 with -T4 and the annotation processors (MapStruct and Hibernate Metamodel Gen) are picked up from the classpath by javac (via maven-compiler-plugin).

Just tried several builds locally with -T8, but it always worked as expected.

agudian:
Tomorrow I'll rewrite the poms to perform the processing with the processor plugin, separated from the acutal compilation. Let's see if that changes anything...

gunnarmorling:
You could try and remote-debug your Maven process, maybe modifying the
IndentationCorrectingWriter class so you can add a break point if an empty
file gets written. Will probably require some fiddling...

agudian:
That was my plan as well, but I can't reproduce it deterministically. I tried it locally several times without any luck, and I can't do it remotely on the jenkins build node...

agudian:
Using the processor-plugin didn't change anything.

What I can see is that if I end up with an empty file, the write(..) method in IndentationCorrectingWriter is never called and the encapsulating BufferedWriter is empty as well. That could hint to a threading-issue with freemarker or the way we use it. Didn't find it yet, though.

agudian:
What I found out today: freemarker's template cache returned an "empty" template, i.e. a Template instance that contains an empty rootNode, as if the input stream was empty.
By default, the cache checkes every five seconds if the template source may have changed. For us it doesn't change, but that's something that the cache seems to recognize - shouldn't be a problem. But the default cache storage uses weak references, allowing cache entries to be GC'ed. I've reconfigured the cache to not do any checks for modifications and to use a cache storage that uses hard references (i.e. once cached, an entry is not discarded).
[...]

PR #521, opened 2015-03-27
#508 better caching of templates, disable caching in jar file connections

agudian:
... to avoid suddenly closed InputStreams


gunnarmorling:
Rebased and applied. Thanks, @agudian!

Issue #523, opened 2015-03-30
Wrong variable type in getter-only collection assignment leads to compile error


agudian:
I have a target collection of type List<String> (getter-only) and a source collection of type Set<String>. This is the generated code:

```java
        if ( source.getSet() != null ) {
            if ( target.getList() != null ) {
                List<String> targetList = source.getSet();  // <-- Compile Error
                if ( targetList != null ) {
                    target.getList().addAll( targetList );
                }
            }
        }
```

agudian:
Created a rather simple PR for this.

I also just noticed that there are now two null-checks for the source-collection. @sjaakd, do you have an idea where that is coming from now?

sjaakd:
> I also just noticed that there are now two null-checks for the source-collection. @sjaakd, do you have an idea where that is coming from now?

I checked.. This is due to the update scenario. Try moving the following lines up in the `PropertyMapping` (somewhere around line 318)
```java
                );
            }
            else {
                // target accessor is getter, so wrap the setter in getter map/ collection handling
                result = new GetterWrapperForCollectionsAndMaps(
                    result,
                    method.getThrownTypes(),
                    targetType,
                    existingVariableNames
                );
            }

            // For collections and maps include a null check, when the assignment type is DIRECT.
            // for mapping methods (builtin / custom), the mapping method is responsible for the
            // null check. Typeconversions do not apply to collections and maps.
            if ( result.getType() == DIRECT ) {
                result = new NullCheckWrapper( result );
            }
```
```java
               // For collections and maps include a null check, when the assignment type is DIRECT.
                // for mapping methods (builtin / custom), the mapping method is responsible for the
                // null check. Typeconversions do not apply to collections and maps.
                if ( result.getType() == DIRECT ) {
                    result = new NullCheckWrapper( result );
                }
            }
            else {
                // target accessor is getter, so wrap the setter in getter map/ collection handling
                result = new GetterWrapperForCollectionsAndMaps(
                    result,
                    method.getThrownTypes(),
                    targetType,
                    existingVariableNames
                );
            }
```

And you'll notice testcase: `org.mapstruct.ap.test.collection.defaultimplementation.NoSetterCollectionMappingTest` starts to fail. The reason: in update methods (`@MappingTarget`), if you include an null source collection / map, the code currently does not touch the target collection (no clear). I reckon this behavior is debatable.

agudian:
@sjaakd, thanks... I guess we can optimize that later on :wink:

[...]

PR #525, opened 2015-03-30
#523 fix type of assignment in GetterWrapperForCollectionsAndMaps

agudian:
This only fixes that one spot described in issue #523, but my guess is that we could have a similar problem in other assignments with local vars as well.


gunnarmorling:
Looks good. Could you merge it yourself? Thanks, @agudian!

Issue #527, opened 2015-04-03
MapStruct does not select mapping methods based on wildcard generic '?'


sjaakd:
MapStruct does not generate, and neither select the proper iterable mapping method.

Given a source 
```java
public class Source {
    private List<? extends Apple> apple
}
```
And a target
```java
public class Target{
    private List<MyApple> apple
}
```

And a mapping method:
```java
public List<MyApple> toMyAppleList( List<Apple> apples);
```
I would expect MapStruct to select this mapping method. 

Instead, I need to define the following method hand written:
```java
public List<MyApple> toMyAppleList( List<? extends Apple> apples);
```
for MapStruct is able to make the mapping.

MapStruct does also not create (forge) automatically the above method.

sjaakd:
The problem is actually a bit more tricky:

In case of a wild card source (extends bounded) or a wild card target (super bounded):

1. The wrong name is generated
2. Non compilable code is generated inside this method.

Then parameterized types can never be generated in the current setup. If we ever want to do this, MapStruct needs to do a whole lot more analyses (e.g is the parameter stuck on the method, or on the @Mapper class.  

For super bounded wildcard sources and extends bounded targets. We would not know what kind of element to use.

Last but not least (really a separate issue). Mappings wildcard mappings of the same type are not handled correctly. MapStruct tries to generate a forged mapping methods while normally these are handled by an implementation type constructor.

sjaakd:
Merged and fixed

PR #530, opened 2015-04-05
Solution for #527

sjaakd:
The solution introduces a fix for supporting wildcard extends bounds on the source side and wildcard super bounds on the target side.

Next to this, error handling is introduced for the reverse case and for all typevar cases.

Selection of wildcard methods needs to be revisited (see #527). I'll open a new issue for this as soon as this one is completed.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/common/Type.java
@@ -37,6 +37,8 @@ import javax.lang.model.element.TypeElement;
 import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
+import javax.lang.model.type.TypeVariable;
+import javax.lang.model.type.WildcardType;
 import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
 
@@ -89,6 +91,10 @@ public class Type extends ModelElement implements Comparable<Type> {
     private List<ExecutableElement> setters = null;
     private List<ExecutableElement> adders = null;
     private List<ExecutableElement> alternativeTargetAccessors = null;
+
+
+    private Type boundingBase = null;
+
 
     //CHECKSTYLE:OFF
     public Type(Types typeUtils, Elements elementUtils, TypeFactory typeFactory,
@@ -221,6 +227,29 @@ public class Type extends ModelElement implements Comparable<Type> {
     public boolean isArrayType() {
         return componentType != null;
     }
+
+    public boolean isTypeVar() {
+        return (typeMirror.getKind() == TypeKind.TYPEVAR);
+    }
+
+    public boolean isWildCardSuperBound() {
+        boolean result = false;
+        if ( typeMirror.getKind() == TypeKind.WILDCARD ) {
+            WildcardType wildcardType = (WildcardType) typeMirror;
+            result = wildcardType.getSuperBound() != null;
+        }
+        return result;
+    }
+
+    public boolean isWildCardExtendsBound() {
+        boolean result = false;
+        if ( typeMirror.getKind() == TypeKind.WILDCARD ) {
+            WildcardType wildcardType = (WildcardType) typeMirror;
+            result = wildcardType.getExtendsBound() != null;
+        }
+        return result;
+    }
+
 
     public String getFullyQualifiedName() {
         return qualifiedName;
@@ -608,7 +637,8 @@ public class Type extends ModelElement implements Comparable<Type> {
     }
 
     /**
-     * A valid Java expression most suitable for representing null - useful for dealing with primitives from FTL.
+     * @return A valid Java expression most suitable for representing null - useful for dealing with primitives
+     * from FTL.
      */
     public String getNull() {
         if ( !isPrimitive() || isArrayType() ) {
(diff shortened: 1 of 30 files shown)


processor/src/main/java/org/mapstruct/ap/model/common/Type.java
```diff
@@ -665,4 +690,57 @@ public int compareTo(Type o) {
...
+     * <li>{@code<? extends Number>}, returns Number</li>
+     * <li>{@code<? super Number>}, returns Number</li>
+     * <li>{@code<?>}, returns Object</li>
+     * <li>{@code<T extends Number>, returns Number</li>
+     * </ul>
+     * @return the bound for this parameter
+     */
+    public Type getTypeBound() {
+
+        Type boundingBase = this;
```
sjaakd: yes it would. :smile:. I'll fix that. I initially thought of adding this in the `TypeFactory`. Anyway, I now fixed the `IterableMapping` and `MapMapping` but perhaps it should be added in the `PropertyMapping` as well (separate PR, don't want to make this too big).

processor/src/test/java/org/mapstruct/ap/test/collection/wildcard/IterableTypeVarBoundMapperOnMethod.java
```diff
@@ -0,0 +1,32 @@
...
+import org.mapstruct.Mapper;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+@Mapper
+public interface IterableTypeVarBoundMapperOnMethod {
+
+    <T extends String> List<T> map(List<String> in);
```
agudian: Funny that this even compiles, as String is final already... I could imagine this to break in future comipler versions, though...

agudian:
OK, I took a closer look and I didn't find anything wrong with it... (I didn't wrap my head around it fully, so I can't say I completely understood everything of it). :+1: 

Just in case gunnar and you can spin a review on #84 this weekend, would it be possible to hold back the merge until after #84 lands on master? I'm not exactly sure if I might have to adapt something there as well, if this PR would be merged first (my guess: no).

processor/src/test/java/org/mapstruct/ap/test/collection/wildcard/IterableTypeVarBoundMapperOnMethod.java
sjaakd: Of course.. Needed a simple type but forgot that a string is final. I'll change that..

sjaakd:
> Just in case gunnar and you can spin a review on #84 this weekend, would it be possible to hold back the merge until after #84 lands on master? I'm not exactly sure if I might have to adapt something there as well, if this PR would be merged first (my guess: no).

Sure, no problem.


I've now fixed this for Itterable and Map mappings. But I need to look into Property mappings as well (at least add a test case).

gunnarmorling:
Wow, that's a huge one :) Could you maybe describe a bit on a high level what has changed? Many thanks!

processor/src/test/java/org/mapstruct/ap/test/collection/wildcard/GoodIdea.java
```diff
@@ -0,0 +1,49 @@
...
+package org.mapstruct.ap.test.collection.wildcard;
+
+import java.math.BigDecimal;
+import javax.xml.bind.JAXBElement;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+public class GoodIdea {
```
gunnarmorling: :)

processor/src/main/java/org/mapstruct/ap/util/Message.java
```diff
@@ -88,6 +88,12 @@
     RETRIEVAL_PRIMITIVE_RETURN( "Can't generate mapping method with primitive return type." ),
     RETRIEVAL_ENUM_TO_NON_ENUM( "Can't generate mapping method from enum type to non-enum type." ),
     RETRIEVAL_NON_ENUM_TO_ENUM( "Can't generate mapping method from non-enum type to enum type." ),
+    RETRIEVAL_TYPE_VAR_SOURCE( "Can't generate mapping method for a generic type variable source." ),
```
gunnarmorling: What is a "generic type variable source"? Is it an unbound wildcard?

gunnarmorling:
From a quick view I think it's fine, still a quick explanation of what's happening in this change might be helpful.

Btw. did you guys notice we crossed the 100 stargazers mark. Yeah :)

processor/src/main/java/org/mapstruct/ap/util/Message.java
sjaakd: generics come basically in 2 forms: parameterized (like <T extends Something> on method / class level) and wildcards (the question mark). A generic type variable is the first one.  Checkout http://docs.oracle.com/javase/7/docs/api/javax/lang/model/type/TypeKind.html

sjaakd:
> From a quick view I think it's fine, still a quick explanation of what's happening in this change might be helpful.

I'll add an explanation tomorrow.



> Btw. did you guys notice we crossed the 100 stargazers mark. Yeah :)

Cool! :smile:

sjaakd:
After the explanation, and after the merge of @agudian's piece-de-resistance :smile:, can I squash and merge this one?

sjaakd:
Rebased and squashed into one commit.

This PR introduces selection and forging of wildcard parameterized Iterable and Map mapping methods. For target elements, a super bound is supported. For source elements an extends bound.

For the first, the forged (and selected) method looks like this:
```java
    protected List<? super org.mapstruct.ap.test.collection.wildcard.Plan> ideaListToPlanList(List<Idea> list) {
        if ( list == null ) {
            return null;
        }

        List<? super org.mapstruct.ap.test.collection.wildcard.Plan> list_ = new ArrayList<Plan>();

        for ( Idea idea : list ) {
            list_.add( map( idea ) );
        }

        return list_;
    }
```
 for the latter the forged (and selected) method looks like this:
```java
   protected List<Plan> ideaListToPlanList(List<? extends org.mapstruct.ap.test.collection.wildcard.Idea> list) {
        if ( list == null ) {
            return null;
        }

        List<Plan> list_ = new ArrayList<Plan>();

        for ( Idea idea : list ) {
            list_.add( map( idea ) );
        }

        return list_;
    }
```
The same goes for map mapping methods.

Note: It seems that the used `TypeMirror` api uses fully qualified names when representing wild card types. This is visible in the generated code and can be improved in the future.

---------------------

Also some code and tests have been added to prevent selection (and forging) of typevar parameterized types. An appropriate error will be generated. Examples:


```java
@Mapper
public interface IterableTypeVarBoundMapperOnMapper<T extends BigDecimal> {

     List<BigDecimal> map(List<T> in);
}
```
and
```java
@Mapper
public interface IterableTypeVarBoundMapperOnMethod {

    <T extends BigDecimal> List<T> map(List<BigDecimal> in);
}
```

agudian:
I'm just seeing it now with your example output above... 

> Note: It seems that the used TypeMirror api uses fully qualified names when representing wild card types. This is visible in the generated code and can be improved in the future.

Did you try it out with a different annotation processor implementation, e.g. with Eclipse? A simple addition to one of the integration tests could do that for you, if you don't want to pull out your dusty Eclipse installation... ;-). Just to make sure that this still produces compilable code...

sjaakd:
> Did you try it out with a different annotation processor implementation, e.g. with Eclipse? A simple addition to one of the integration tests could do that for you, if you don't want to pull out your dusty Eclipse installation... ;-). Just to make sure that this still produces compilable code...

I'll add an integration test. Do you suspect that other annotations processor will handle this differently?

sjaakd:
> Meh, I don't know... just a feeling. We had something vaguely similar with the string representation of executable elements, remember?

Ok.. Added some integration test.. No problem.. Will proceed to merge :smile: . Thanks.

Issue #365, opened 2014-11-28
Mapping from class with non-JavaBean style getters


scottmore:
Original discussion on mapstruct-users: 
https://groups.google.com/forum/#!topic/mapstruct-users/bURGk6CS4fk

There are cases where I need to map from Java classes that don't follow the getFoo() method JavaBeans naming convention, instead consistently lobbing off the "get" part.  For example, I need to automatically map from the following class to a class with matching "setName(String)" setter without having to define mappings explicitly (i.e via configuration annotation or similar).

```
public class SourceClass {

    private String name;

    public SourceClass() {
    }

    public String name() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }
}
```

One idea is to introduce an interface to map from getter method to property name, then have an optional argument on the configure annotation to specify a class implementing that interface.  The mapper could map method names in source class to property names (JavaBeans: getName -> name, My example: name -> name).  Hopefully that would play nice with the @Mapper annotations and allows for a very similar SetterTransform to map from property name to any bizarre setter method convention in the target class.

something like this:

    public interface GetterTransform {
       
        /**
         * Transform method receives all method names in source class of
         * no-argument, non-void return type methods and returns property
         * name or null if method is not to be mapped
         *
         * @param methodName Name of potential getter method in source class
         * @return Expected property name in target class or null to skip
         */
        public String transform(String methodName);
    }

A JavaBeans-compliant class transform would look something like this:
   
    public class JavaBeansTransform implements GetterTransform {

        public String transform(String methodName) {
            if (!methodName.startsWith("get") || methodName.length() <= 3) {
                return null;
            }
           
            // Lob off get and lowercase next character
            return String.format("%s%s",
                    methodName.substring(3, 4).toLowerCase(),
                    methodName.substring(4));
        }
    }

   
For classes like the SourceClass example above the transform would look like this:

    public class CustomTransform implements GetterTransform {
       
        public String transform(String methodName) {
            return methodName;
        }
    }

Gunnar Morling's reply:

>One minor issue I see with the proposal of
```
    @Mapper(propertyNamer=MyTransformer.class)
```
>is that it requires a class which actually only would be needed for the generation (MyTransformer) to be visible to the application; javac has a separation of the compilation classpath and the "processor path" which contains the classes required by annotation processing. Not sure whether it's really a problem; One could work around it by e.g. specifying the FQN instead (we have a set of processor options already).

chschu:
[...]

PR #536, opened 2015-04-10
#365 make accessor naming configurable

chschu:
This is the PR for the simple variant of #365.

### DIFF
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -149,11 +149,6 @@
                         </goals>
                         <configuration>
                             <minimizeJar>true</minimizeJar>
-                            <artifactSet>
-                                <excludes>
-                                    <exclude>${project.groupId}:mapstruct-spi</exclude>
-                                </excludes>
-                            </artifactSet>
                             <filters>
                                 <filter>
                                     <artifact>org.freemarker:freemarker</artifact>
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -43,6 +43,10 @@
             <groupId>org.freemarker</groupId>
             <artifactId>freemarker</artifactId>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>mapstruct-spi</artifactId>
+        </dependency>
 
         <!-- Compile-only; Using "provided" scope as there is no such scope in Maven;
         these dependencies are not required at runtime, only for prism generation
@@ -145,6 +149,11 @@
                         </goals>
                         <configuration>
                             <minimizeJar>true</minimizeJar>
+                            <artifactSet>
+                                <excludes>
+                                    <exclude>${project.groupId}:mapstruct-spi</exclude>
+                                </excludes>
+                            </artifactSet>
                             <filters>
                                 <filter>
                                     <artifact>org.freemarker:freemarker</artifact>
--- a/processor/src/main/java/org/mapstruct/ap/services/DefaultAccessorNamingStrategy.java
+++ b/processor/src/main/java/org/mapstruct/ap/services/DefaultAccessorNamingStrategy.java
@@ -30,6 +30,10 @@ import org.mapstruct.spi.AccessorNamingStrategy;
 class DefaultAccessorNamingStrategy implements AccessorNamingStrategy {
 
     @Override
+    public void setDefaultAccessorNamingStrategy(AccessorNamingStrategy defaultAccessorNamingStrategy) {
+    }
+
+    @Override
     public boolean isNonBooleanGetterName(String methodName) {
         return methodName.startsWith( "get" ) && methodName.length() > 3;
     }
--- a/processor/src/main/java/org/mapstruct/ap/services/DefaultAccessorNamingStrategy.java
+++ b/processor/src/main/java/org/mapstruct/ap/services/DefaultAccessorNamingStrategy.java
@@ -1,0 +1,76 @@
+/**
+ *  Copyright 2012-2015 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.services;
+
+import java.beans.Introspector;
+
+import org.mapstruct.spi.AccessorNamingStrategy;
+
+/**
+ * The default JavaBeans-compliant implementation of the {@link AccessorNamingStrategy} service provider interface.
+ *
+ * @author Christian Schuster
+ */
+class DefaultAccessorNamingStrategy implements AccessorNamingStrategy {
+
+    @Override
+    public boolean isNonBooleanGetterName(String methodName) {
+        return methodName.startsWith( "get" ) && methodName.length() > 3;
+    }
+
+    @Override
+    public boolean isBooleanGetterName(String methodName) {
+        return methodName.startsWith( "is" ) && methodName.length() > 2;
+    }
+
+    @Override
+    public boolean isSetterName(String methodName) {
+        return methodName.startsWith( "set" ) && methodName.length() > 3;
+    }
+
+    @Override
+    public boolean isAdderName(String methodName) {
+        return methodName.startsWith( "add" ) && methodName.length() > 3;
+    }
+
+    @Override
+    public String getPropertyNameForNonBooleanGetterName(String methodName) {
+        return Introspector.decapitalize( methodName.substring( 3 ) );
+    }
+
+    @Override
+    public String getPropertyNameForBooleanGetterName(String methodName) {
+        return Introspector.decapitalize( methodName.substring( 2 ) );
+    }
+
+    @Override
+    public String getPropertyNameForSetterName(String methodName) {
+        return Introspector.decapitalize( methodName.substring( 3 ) );
+    }
+
+    @Override
+    public String getElementNameForAdderName(String methodName) {
+        return Introspector.decapitalize( methodName.substring( 3 ) );
+    }
+
+    @Override
+    public String getNonBooleanGetterNameForSetterName(String methodName) {
+        return "get" + methodName.substring( 3 );
+    }
+}
--- a/processor/src/main/java/org/mapstruct/ap/services/Services.java
+++ b/processor/src/main/java/org/mapstruct/ap/services/Services.java
@@ -18,9 +18,7 @@
  */
 package org.mapstruct.ap.services;
 
-import java.util.Map;
 import java.util.ServiceLoader;
-import java.util.WeakHashMap;
 
 import org.mapstruct.spi.AccessorNamingStrategy;
 
@@ -31,7 +29,7 @@ import org.mapstruct.spi.AccessorNamingStrategy;
  */
 public class Services {
 
-    private static final Map<Class<?>, Object> CACHE = new WeakHashMap<Class<?>, Object>();
+    private static final AccessorNamingStrategy ACCESSOR_NAMING_STRATEGY = findAccessorNamingStrategy();
 
     private Services() {
     }
@@ -46,18 +44,18 @@ public class Services {
      *             {@link ServiceLoader#load(Class, ClassLoader)}.
      */
     public static AccessorNamingStrategy getAccessorNamingStrategy() {
-        AccessorNamingStrategy impl = (AccessorNamingStrategy) CACHE.get( AccessorNamingStrategy.class );
+        return ACCESSOR_NAMING_STRATEGY;
+    }
+
+    private static AccessorNamingStrategy findAccessorNamingStrategy() {
+        AccessorNamingStrategy impl = find( AccessorNamingStrategy.class );
         if ( impl == null ) {
-            impl = get( AccessorNamingStrategy.class );
-            if ( impl == null ) {
-                impl = new DefaultAccessorNamingStrategy();
-            }
-            CACHE.put( AccessorNamingStrategy.class, impl );
+            impl = new DefaultAccessorNamingStrategy();
         }
         return impl;
     }
 
-    private static <T> T get(Class<T> spi) {
+    private static <T> T find(Class<T> spi) {
         T matchingImplementation = null;
 
         for ( T implementation : ServiceLoader.load( spi, spi.getClassLoader() ) ) {
--- a/processor/src/main/java/org/mapstruct/ap/services/Services.java
+++ b/processor/src/main/java/org/mapstruct/ap/services/Services.java
@@ -18,7 +18,9 @@
  */
 package org.mapstruct.ap.services;
 
+import java.util.Map;
 import java.util.ServiceLoader;
+import java.util.WeakHashMap;
 
 import org.mapstruct.spi.AccessorNamingStrategy;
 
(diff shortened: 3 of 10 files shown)


processor/src/main/java/org/mapstruct/ap/services/Services.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Obtain an implementation of {@link AccessorNamingStrategy}. If no specialized implementation is found using
+     * {@link ServiceLoader}, a JavaBeans-compliant default implementation is returned.
+     *
+     * @return The implementation of {@link AccessorNamingStrategy}.
+     * @throws IllegalStateException If more than one implementation is found by
+     *             {@link ServiceLoader#load(Class, ClassLoader)}.
+     */
+    public static AccessorNamingStrategy getAccessorNamingStrategy() {
+        AccessorNamingStrategy impl = get( AccessorNamingStrategy.class );
```
agudian: I think we should cache the implementation that was found statically - we access this strategy rather often... :wink:

processor/pom.xml
```diff
@@ -145,6 +149,11 @@
                         </goals>
                         <configuration>
                             <minimizeJar>true</minimizeJar>
+                            <artifactSet>
+                                <excludes>
+                                    <exclude>${project.groupId}:mapstruct-spi</exclude>
```
agudian: IMO, we should include the SPI parts into the fat-jar as well (without shading into a different package, of course), as they are always required at runtime of the processor.

When adding the processor manually to javac or in an IDE (e.g. Eclipse), you currently only need to add one mapstruct-processor.jar - I think that is pretty convinient and it would be great if we could keep it that way. 

WDYT?

agudian:
I had a quick look, but now I'm out of time - I still have some ideas on how to add an integration test for this - I'll add some comments on that later (I might need to try some stuff out first :wink:).

But overall, it looks very good! :+1:

processor/src/main/java/org/mapstruct/ap/services/Services.java
chschu: Good idea, I'll add that. At first I thought there might be issues with different classloaders. But the `Services` class always uses the SPI interface's classloader to find the implementation, so that shouldn't be a problem.

processor/pom.xml
chschu: There certainly has been a reason why I did that, but I don't remember what it was. Having the SPI parts in the fat-jar is indeed convenient.

For maven builds with a custom SPI implementation, the SPI interfaces would be on the plugin classpath twice:
```
maven-processor-plugin -> custom-spi-impl -> mapstruct-spi
maven-processor-plugin -> mapstruct-processor -> mapstruct-spi
```
I'll check if that causes any issues, and remove the exclude if it doesn't.

agudian:
For the integration tests, you can create a two-module project below intergrationtest/src/test/resources, with your custom implementation in one module and with the mappers to generate in the other module (and a simple test to assert their correct behaviour). 
Your integration test class then can declare to use one or more ProcessorTypes, and as long as it's not ```PROCESSOR_PLUGIN_JAVA_8```, it should work fine (as for all other processor type values, mapstruct-processor and your custom SPI implementation would be picked from the project's compile-classpath and not from the maven-processor-plugin's plugin classpath).

processor/src/main/java/org/mapstruct/ap/services/Services.java
```diff
@@ -0,0 +1,77 @@
...
+import org.mapstruct.spi.AccessorNamingStrategy;
+
+/**
+ * A simple locator for SPI implementations.
+ *
+ * @author Christian Schuster
+ */
+public class Services {
+
+    private static final Map<Class<?>, Object> CACHE = new WeakHashMap<Class<?>, Object>();
```
agudian: Ah, ok... I actually had something more simple in mind: ```private static final AccessorNamingStrategy ACCESSOR_NAMING_STRATEGY = findStrategy();```
The classloader that we live in is quite short-lived anyway (in Maven, it's only used for the compilation of one particular source directory, in Eclipse it's either one clean build for a project, or one incremental compilation for one project), so any static stuff is discarded together with the classes quite promptly.

spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @author Christian Schuster
+ */
+public interface AccessorNamingStrategy {
+
+    /**
+     * Determine if a method name defines a getter with a non-boolean return type.
+     *
+     * @param methodName The method name.
+     * @return <code>true</code> if the method name can be a non-boolean getter, <code>false</code> otherwise.
```
agudian: Gunnar will probably point that out as well: we prefer using ```{@code true}``` instead of ```<code>true</code>``` because with ```{@code ...}```, you can include HTML-special characters without having to use the entities. E.g. ```{@code x < y}``` would have to be written as ```<code>x &lt; y</code>``` otherwise.

processor/src/main/java/org/mapstruct/ap/services/DefaultAccessorNamingStrategy.java
```diff
@@ -0,0 +1,76 @@
...
+import java.beans.Introspector;
+
+import org.mapstruct.spi.AccessorNamingStrategy;
+
+/**
+ * The default JavaBeans-compliant implementation of the {@link AccessorNamingStrategy} service provider interface.
+ *
+ * @author Christian Schuster
+ */
+class DefaultAccessorNamingStrategy implements AccessorNamingStrategy {
```
agudian: I wonder if it would be easier for implementors of the SPI, if we included the default implementation in the SPI module itself (e.g. in a package org.mapstruct.spi.default or something like that). 

Pro: the simple implementations of the methods make it much easier to grasp what the default behaviour is and what the Javadoc on the interface actually means. Plus, if we also made the default implementation ```public``` instead of having a default visibility, people can just extend the default implementation and overwrite only selected methods. That might make future extensions or even changes in the default behaviour easier to adapt for SPI users.

Con: Future default implementations might require more elaborate classes and we might have to break that principle then.

WDYT?

processor/src/main/java/org/mapstruct/ap/services/Services.java
chschu: You're right. I was just going for a "normal" cache and didn't take the short-livedness (is that even a word?) of the classloader into account. I'll change it.

spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java
chschu: Fixed.

processor/src/main/java/org/mapstruct/ap/services/DefaultAccessorNamingStrategy.java
chschu: I also thought about that. Future default implementations might need classes from `mapstruct-processor`, which would then cause a circular maven dependency to/from `mapstruct-spi`. So I'd rather keep the implementation secret and document the default behavior in the interface's Javadoc.

I thought about some kind of `setDefaultAccessorNamingStrategy()` that is invoked by the `Services` class before returning the custom implementation.

chschu:
@agudian: Creating a multi-module integration test seems to be a bit more complicated, because there are some inherited build settings (e.g. the relative path for the license check) that cause problems in nested modules.

The `ProcessorSuiteRunner` currently runs the `test` goal, which is not sufficient for a multi-module test. Changing that to `install` will do, but that will install lots of test artifacts in the local maven repository.

processor/src/main/java/org/mapstruct/ap/services/DefaultAccessorNamingStrategy.java
agudian: > I thought about some kind of setDefaultAccessorNamingStrategy() that is invoked by the Services class before returning the custom implementation.

Yes, I think that's a good idea

agudian:
> The ProcessorSuiteRunner currently runs the test goal, which is not sufficient for a multi-module test. Changing that to install will do, but that will install lots of test artifacts in the local maven repository.

Oh, that's strange - why is the test goal not sufficient? With the processor being taken from the compile classpath, the SPI should be pulled in from there as well... do you have an example of the not-working projects?

chschu:
I tried to add the following module structure under `integrationtest/src/test/resources`:

```
spiTest
  spiTest-spi (contains the custom implementation and META-INF/services/org.mapstruct.spi.AccessorNamingStrategy)
  spiTest-test (modified simpleTest, adds spiTest-spi to the processor plugin classpath)
```

When `spiTest-test` is built, it can't find `spiTest-spi` because it was never installed to the local maven repository.

agudian:
OK, now I got it. If you configure the pom in src/test/resources as parent of spiTest/spiTest-test, and add spiTest-spi as dependency (compile-scope) to spiTest-test (not the processor plugin, just as a normal dependency), then you can run the integration test with:
```java
@ProcessorSuite( baseDir = "spiTest", processorTypes = ProcessorType.ORACLE_JAVA_8 )

File: spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java [hunk 6]
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Determine if a method name defines a getter with a non-boolean return type.
+     * <p>
+     * Default implementation: Method name starts with "get" and is longer than 3 characters.
+     *
+     * @param methodName The method name.
+     * @return {@code true} if the method name can be a non-boolean getter, {@code false} otherwise.
+     */
+    boolean isNonBooleanGetterName(String methodName);
```
gunnarmorling: Is the distinction between boolean and non-boolean getters actually needed? Wouldn't it suffice to have one method `isGetterName()`?

spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java
```diff
@@ -0,0 +1,134 @@
...
+     * Extract the property name from a method name for which {@link #isNonBooleanGetterName(String)} returned
+     * {@code true}.
+     * <p>
+     * Default implementation: Remove first 3 characters ("get") of the method name, and
+     * {@link java.beans.Introspector#decapitalize(String) decapitalize} the result.
+     *
+     * @param methodName The method name, guaranteed to be a non-boolean getter name.
+     * @return The property name corresponding to the method name.
+     */
+    String getPropertyNameForNonBooleanGetterName(String methodName);
```
gunnarmorling: Same here, I feel it'd be enough to have a single method `getPropertyName(String methodName)` instead of the three separate ones. An implementation could do the distinction internally itself.

spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Extract the non-boolean getter method name from the setter method name of the same property.
+     * <p>
+     * Default implementation: Replace the first 3 characters ("get") of the method name with "set".
+     *
+     * @param methodName The method name, guaranteed to be a setter name.
+     * @return The corresponding non-boolean getter method name.
+     */
+    String getNonBooleanGetterNameForSetterName(String methodName);
```
gunnarmorling: I think I'd pass the property name rather than the setter name and not restrict it to non-booleans: `getGetterName(String property)`.

spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Set the default {@link AccessorNamingStrategy} implementation. Custom implementations may use it to
+     * keep parts of the default behavior. This method is invoked before any other method of the interface.
+     * <p>
+     * Default implementation: Does nothing.
+     *
+     * @param defaultAccessorNamingStrategy The default implementation.
+     */
+    void setDefaultAccessorNamingStrategy(AccessorNamingStrategy defaultAccessorNamingStrategy);
```
gunnarmorling: I'm wondering what the use case for this is. When would you actually make use of that if you work with a custom strategy? I'm not strictly against it if there is a case for it, just trying to figure what actually is needed.

spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java
gunnarmorling: To sum up, WDYT about having just the following methods on this contract:

* `isPropertySetter(String method)`
* `isPropertyGetter(String method)`
* `isAdder(String method)`
* `getPropertyName(String getterOrSetterMethod)`
* `getElementName(String adderMethod)`
* `getGetterName(String property)`

?

This seems a bit simpler on implementors to me. You could even think about collapsing the first three into a single one `MethodType getType(String method)` returning an enum with values `SETTER`, `GETTER`, `ADDER` and `OTHER`?

Any thoughts?

processor/src/main/java/org/mapstruct/ap/services/DefaultAccessorNamingStrategy.java
gunnarmorling: > So I'd rather keep the implementation secret and document the default behavior in the interface's Javadoc.

+1

> before returning the custom implementation.

I'm wondering when an implementor would actually use that? The implementation is fairly trivial, so I don't see much value in supporting that delegation and would rather prefer to make the exposed contract as lean as possible. Also if one works with custom name patterns in the first place, would delegation to the default implementation be useful at all?

spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java
gunnarmorling: Thanks :)

gunnarmorling:
@chschu Many thanks for providing this feature! It looks good overall to me, added some remarks on the design of the contract itself. It'd be nice if we could slim it down a bit.

spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java
chschu: The `Executables` class does the method signature checks, and those are different for boolean and non-boolean getters. We could pass the whole `ExecutableElement` and do the signature check in the SPI.

spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java
chschu: You're right, for those three it should be perfectly fine. I'd like to keep the interface methods on the same level of abstraction, so if we pass one of them an `ExecutableElement` instead of a `String`, all of them should get an `ExecutableElement`. The `getPropertyName()` method might need to throw an exception if no property name can be determined - that currently happens in the `Executables` class.

spi/src/main/java/org/mapstruct/spi/AccessorNamingStrategy.java
chschu: You don't have to use it, but it's there if you need it. A possible use case is to only change the getter or setter naming scheme, not both. That makes more sense if the interface has many methods, or complex ones you don't want to reimplement.

gunnarmorling:
@chschu, sorry for the long silence, I was busy with some other things. I think I'll simplify the contract like this:

    public interface AccessorNamingStrategy {

        enum MethodType { GETTER, SETTER, ADDER, OTHER }

        MethodType getMethodType(ExecutableElement method);
        String getPropertyName(ExecutableElement getterOrSetterMethod);
        String getElementName(ExecutableElement adderMethod);
        String getCollectionGetterName(String property);
}

One question though: What was the motivation to have the SPI in a separate module? Couldn't it just be a package in the processor JAR?

gunnarmorling:
@agudian, @sjaakd: Do you have any preference/thought wrt. to having the SPI within the processor JAR itself or a separate module? I think I'd prefer it in the processor JAR for the sake of simplicity.

agudian:
What ever we choose to do, the classes should end up in the processor jar - as you say, it simplifies the usage (e.g. when configuring it manually in eclipse or javac).

We already have extension points available for which the API resides within the processor module. So we wouldn't make anything worse. 
On the other hand, we could create the SPI module (which would be a 'provided' dependency for users) to clearly indicate our extension points.

I don't have any strong feelings for either way.

sjaakd:
Hmm. Looking from an SPI user p.o.v. I would suggest to make it a separate module. Why pull in all kind of other processor packages while you only need a few interfaces?

gunnarmorling:
My concern is that the SPI packages may end up twice on the classpath if they are in a separate module. As @agudian is saying, we should include (shade?) them into procesor JAR for ease of use. I don't think we can change the package (as we do for FreeMarker) as I'd expect SPI implementations to not link otherwise. So we'd keep them in the same package, but then a user could accidentally add processor JAR and SPI JAR and have the classes twice.

I think I will move the SPI package to the processor JAR and move all the existing classes to `org.mapstruct.ap.internal` (with the exception of `MappingProcessor` which is the "public" entry point). That way it is clear that the internal classes are not meant for public usage.

sjaakd:
> I think I will move the SPI package to the processor JAR and move all the existing classes to org.mapstruct.ap.internal (with the exception of MappingProcessor which is the "public" entry point). That way it is clear that the internal classes are not meant for public usage.

Ok.. seems a good approach to me.

sjaakd:
About the `AccessorNamingStrategy`. 

Does the SPI interface enforce an interface type? If we use a class instead, the user does not have to override all methods but can select which ones to override and for which ones to evert to default.

chschu:
According to the ServiceLoader JavaDoc, it doesn't need to be an interface. It could also be an abstract or even a concrete class. Haven't tried this, though.

gunnarmorling:
Superseeded by #561. Going to close this one. Thanks, @chschu!

Issue #540, opened 2015-05-08
Use enum name instead of toString() for enum to String conversion


gunnarmorling:

ps-powa:
https://github.com/mapstruct/mapstruct/pull/541

PR #541, opened 2015-05-08
#540 use name() instead of toString() for enum to String conversion

ps-powa:
Reason for change:
name() is safer than toString() because it always returns the enum name exactly as declared, whereas toString() can be overridden and may return any value.


gunnarmorling:
Merged; Thanks, Paul!

PR #545, opened 2015-05-11
515 Negative array index in indentation writer

sjaakd:
In expressions (and presumably also in contstants) its possible to create Strings definitions in the generated code. Indentation should therefore not be increased / decreased when braces (curly or round) are encountered inside a String definition (so between " "). 

Decreasing due to an unbalanced round brace inside a String definition led to the problem above.

Likewise, escaped '"' should be ignored.

The enum-state pattern (one of my favourites :+1: ) has been updated with a couple of extra states to handle this.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/writer/IndentationCorrectingWriter.java
+++ b/processor/src/main/java/org/mapstruct/ap/writer/IndentationCorrectingWriter.java
@@ -29,6 +29,16 @@ import java.util.Arrays;
  * This writer discards any leading whitespace characters following to a line break character. When the first
  * non-whitespace character is written after a line break, the correct indentation characters are added, which is four
  * whitespace characters per indentation level.
+ *
+ * <p>
+ * The state pattern is line oriented. It starts by writing text. Indentation is increased if a brace '('or
+ * brace '{' is encountered in the code to be generated and written out in state: IN_TEXT_START_OF_LINE. Whenever
+ * a line end occurs (PC or Linux style) the amount of enters is checked and at max set to 2.
+ *
+ * Whenever a string definition is encountered in the code that should be generated, increasing the indentation is
+ * stopped (so `{` and '(' are ignored) until the end of the string is encountered ('"'). To avoid writing a new
+ * indentation, the state then returns to IN_TEXT.
+ *
  * <p>
  * This is a very basic implementation which does not take into account comments, escaping etc.
  *
@@ -43,7 +53,7 @@ class IndentationCorrectingWriter extends Writer {
     private static final String LINE_SEPARATOR = System.getProperty( "line.separator" );
     private static final boolean IS_WINDOWS = System.getProperty( "os.name" ).startsWith( "Windows" );
 
-    private State currentState = State.IN_TEXT;
+    private State currentState = State.START_OF_LINE;
     private final StateContext context;
 
     IndentationCorrectingWriter(Writer out) {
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/writer/IndentationCorrectingWriter.java
```diff
@@ -142,27 +162,113 @@ void onBufferFinished(StateContext context) throws IOException {
...
-                                    context.lastStateChange,
-                                    context.currentIndex
-                                )
-                            )
-                        );
-                    }
+        /**
+         * In a String definition, Between un-escaped quotes " "
+         */
+        IN_TEXT_IN_STRING {
```
gunnarmorling: Maybe just `IN_STRING`?

processor/src/main/java/org/mapstruct/ap/writer/IndentationCorrectingWriter.java
```diff
@@ -142,27 +162,113 @@ void onBufferFinished(StateContext context) throws IOException {
...
+             */
+            @Override
+            void onBufferFinished(StateContext context) throws IOException {
+                flush( context );
+            }
+
+        },
+
+        /**
+         * Within any text, but after a String (" ").
```
gunnarmorling: The "after" is confusing me. Couldn't it e.g. also be between several Strings on one line?

gunnarmorling:
Good catch and nice fix, @sjaakd. I am only wondering whether that one new state is actually needed? I'd have said `IN_TEXT` and `IN_STRING` (where we don't increment/decrement indentation) should suffice?

processor/src/main/java/org/mapstruct/ap/writer/IndentationCorrectingWriter.java
sjaakd: If it would be a classical state pattern (abstract base class, each state is a class), then IN_STRING would be a substate of IN_TEXT (inherits from).. I now tried to express that in the name.

processor/src/main/java/org/mapstruct/ap/writer/IndentationCorrectingWriter.java
sjaakd: that's indeed true.. However, only want the "onEntry" of IN_TEXT be executed once. So I cannot return to IN_TEXT. Did that, but it generates an indentation before the string-def-closing '"'. Suggestions are welcome for a better name.

processor/src/main/java/org/mapstruct/ap/writer/IndentationCorrectingWriter.java
```diff
@@ -142,27 +162,113 @@ void onBufferFinished(StateContext context) throws IOException {
...
+            void onBufferFinished(StateContext context) throws IOException {
+                flush( context );
+            }
+
+        },
+
+        /**
+         * In a String, character following an escape character (should be ignored, can also be '"' ).
+         */
+        IN_TEXT_IN_STRING_ESCAPE {
```
sjaakd: Ah.. but the escaped character could also be a double quote '"'. In that case you don't want to interpreted it as a closing '"'. The reproducer demonstrates that. Checkout the generated code.

sjaakd:
> I'd have said IN_TEXT and IN_STRING (where we don't increment/decrement indentation) should suffice?

Checkout my comments above. I thought that as well (initially).

sjaakd:
screen shot 2015-05-12 at 21 18 17

@gunnarmorling , I processed your comments. I renamed the initial state to START_OF_LINE, that writes the indentations. There's only an IN_TEXT and there's an IN_STRING. 

However, I still need the IN_STRING_ESCAPE_CHAR, more or less because of the same reason as you needed the IN_LINE_BREAK: you need to know what the previous character is in order to ignore (also ignore the ").

WDYT? Can is it ready to squash and merge?

gunnarmorling:
> WDYT? Can is it ready to squash and merge?

@sjaakd, yes, I would say so :) Can you do the merge? Great work, many thanks. Glad to hear btw. you like that pattern. I am also quite happy with it. Nice diagram, what did you use to draw it?

sjaakd:
> Nice diagram, what did you use to draw it?

[StarUml](http://staruml.io/). Its basically free.. (evaluation version, so it will complain now and then to switch to the payed-deal).

I used to use the enum-state-pattern a lot. Telecom has a lot of state and its a nice compact pattern which maps beautifully to UML2. I even thought about making some kind of generator in the past from XMI -> Java-Enum-Pattern. Never did that :smile:

sjaakd:
@gunnarmorling , thanks for reviewing :smile:

PR #546, opened 2015-05-14
Generate Javadocs during mvn install

agudian:
So we'll have feedback on errors / warnings during our local builds and on Jenkins.

If I only knew what I got into... I had to fix a whole bunch of errors as well. :wink:

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/model/MappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/MappingMethod.java
@@ -29,6 +29,7 @@ import org.mapstruct.ap.model.common.ModelElement;
 import org.mapstruct.ap.model.common.Parameter;
 import org.mapstruct.ap.model.common.Type;
 import org.mapstruct.ap.model.source.Method;
+
 import static org.mapstruct.ap.util.Strings.getSaveVariableName;
 import static org.mapstruct.ap.util.Strings.join;
 
@@ -49,11 +50,11 @@ public abstract class MappingMethod extends ModelElement {
     private final String resultName;
 
     /**
-     * constructor to be overloaded when local variable names are required prior to calling this constructor. (e.g.
-     * for property mappings). It is supposed to be initialized with at least the parameter names.
+     * constructor to be overloaded when local variable names are required prior to calling this constructor. (e.g. for
+     * property mappings). It is supposed to be initialized with at least the parameter names.
      *
-     * @param method
-     * @param existingVariableNames
+     * @param method method
+     * @param existingVariableNames existingVariableNames
      */
     protected MappingMethod(Method method, Collection<String> existingVariableNames ) {
         this.name = method.getName();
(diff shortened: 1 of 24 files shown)


processor/src/main/java/org/mapstruct/ap/model/assignment/Assignment.java
```diff
@@ -59,14 +60,14 @@
...
      /**
-      * An assignment in itself can wrap another assignment. E.g.:
-      * <ul>
-      * <li>a MethodReference can wrap a TypeConversion, another MethodReference and ultimately a Simple</li>
-      * <li>a TypeConversion can wrap a MethodReference, and ultimately a Simple</li>
-      * </ul>
-      *
-      * @param assignment
-      */
+     * An assignment in itself can wrap another assignment. E.g.:
```
gunnarmorling: The new indentation looks not correct?

processor/src/main/java/org/mapstruct/ap/model/MappingMethod.java
```diff
@@ -49,11 +50,11 @@
...
 
     /**
-     * constructor to be overloaded when local variable names are required prior to calling this constructor. (e.g.
-     * for property mappings). It is supposed to be initialized with at least the parameter names.
+     * constructor to be overloaded when local variable names are required prior to calling this constructor. (e.g. for
+     * property mappings). It is supposed to be initialized with at least the parameter names.
      *
-     * @param method
-     * @param existingVariableNames
+     * @param method method
```
gunnarmorling: Personally I would have ommitted the param tags altogether. These shouldn't cause any errors / build failures btw. The reporting of JavaDoc tooling is really bad; it fails the build as soon as there actually is one real "error" (such as broken links or actual syntax errors), but these warnings also show up. It's a bit cumbersome to identify the actual trouble makers.

Same for the return tags. They should be fine in the original form.

processor/src/main/java/org/mapstruct/ap/model/common/Type.java
```diff
@@ -619,7 +617,9 @@ private int distanceTo(TypeMirror base, TypeMirror targetType) {
     }
 
     /**
-     * Whether this type can access the given method declared on the given type.
+     * @param type the type declaring the method
```
gunnarmorling: I don't think the param tags add much value here. Let's don't go overboard just to make the JavaDoc tool happy ;) The original form should be fine (and actually I intentionally use that style if @param/@return don't add any value.

gunnarmorling:
> If I only knew what I got into... I had to fix a whole bunch of errors as well. 

Ahaha :) FWIW, most of the fixes should not have been necessary; The tool fails the build as soon is there one real error, but then also displays alls the warning-level issues which themselves would not fail the build.

gunnarmorling:
Merged; Many thanks, @agudian!

agudian:
> Ahaha :) FWIW, most of the fixes should not have been necessary; The tool fails the build as soon is there one real error, but then also displays alls the warning-level issues which themselves would not fail the build.

Yeah... All those warning messages clutter the output pretty much... would be great if we could just suppress that... :-/

Thanks for merging!

Issue #513, opened 2015-03-21
Forged Method does not re-throw exceptions from encapsulated mapping methods


sjaakd:

sjaakd:
Prepared reproducer.. working on it.

sjaakd:
fixed. Solution merged

PR #549, opened 2015-05-17
#513 forged methods should re-throw exceptions from used mapping methods

sjaakd:
This PR contains the following commits:
1. reproducer
2. fix
3. code consistency: exceptions in a method signature are thrown types.
4. template improvement (guaranteed space before throws keyword in generated method signature)

The fix checks whether an iterable- or mapmappingmethod is forged  is forged. If so, it adds the thrown types of the used mapping methods. For mapmappingmethods it could be that an exception is already added by the key mapping. Hence the value mapping checks whether a type is already present.

I'll squash the first two commits later on.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/model/IterableMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/IterableMappingMethod.java
@@ -123,7 +123,7 @@ public class IterableMappingMethod extends MappingMethod {
             else {
                 if ( method instanceof ForgedMethod ) {
                     ForgedMethod forgedMethod = (ForgedMethod) method;
-                    forgedMethod.addThrownTypes( assignment.getExceptionTypes() );
+                    forgedMethod.addThrownTypes( assignment.getThrownTypes() );
                 }
             }
             // target accessor is setter, so decorate assignment as setter
--- a/processor/src/main/java/org/mapstruct/ap/model/IterableMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/model/IterableMappingMethod.java
@@ -29,6 +29,7 @@ import org.mapstruct.ap.model.assignment.LocalVarWrapper;
 import org.mapstruct.ap.model.assignment.SetterWrapper;
 import org.mapstruct.ap.model.common.Parameter;
 import org.mapstruct.ap.model.common.Type;
+import org.mapstruct.ap.model.source.ForgedMethod;
 import org.mapstruct.ap.model.source.Method;
 import org.mapstruct.ap.prism.NullValueMappingStrategyPrism;
 import org.mapstruct.ap.util.Message;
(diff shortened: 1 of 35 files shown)


processor/src/main/java/org/mapstruct/ap/model/IterableMappingMethod.java
```diff
@@ -116,6 +117,10 @@ public IterableMappingMethod build() {
                 false
             );
 
+            if ( method instanceof ForgedMethod ) {
+                method.getThrownTypes().addAll( assignment.getThrownTypes() );
```
agudian: For most other Methods, `getThrownTypes()` returns an immutable list. How about making the different requirement in `ForgedMethod` explicit by adding a method `addThrownTypes(..)` to `ForgedMethod`? It could also contain the contains-check that was implemented below for the MapMappingMethod builder.
The contains-check could also be avoided by declaring `thrownTypes` as `Set` (using LinkedHashSet as implementation, which guarantees an iteration in the order of the inserts).

agudian:
Great, @sjaakd! 
I've added only one small comment. Feel free to merge it yourself :+1:

sjaakd:
@agudian, thanks for reviewing!

gunnarmorling:
Sorry for the late reply. Thanks, guys! One thing though: WDYT about using a "sneaky throws" approach similar to that one we use in OGM: https://github.com/hibernate/hibernate-ogm/blob/master/core/src/main/java/org/hibernate/ogm/exception/impl/Exceptions.java

This leverages type erasure of the compiler. By calling `Exceptions.<RuntimeException>sneakyThrows( myCheckedException )` we can re-throw any checked exception without the need for declaration. That avoid the wrapping runtime exceptions. I think it'd make sense to employ that pattern in our generated code. Any thoughts?

agudian:
> Sorry for the late reply.

Relax, it's been only one day. :grin:

> This leverages type erasure of the compiler. By calling Exceptions.<RuntimeException>sneakyThrows( myCheckedException ) we can re-throw any checked exception without the need for declaration. That avoid the wrapping runtime exceptions. I think it'd make sense to employ that pattern in our generated code. Any thoughts?

Wow, didn't know that such a thing is possible. Looks very hacky, but I see why that works and probably continues to work in the future.

Would your proposal be adding that as a built-in method if needed?

gunnarmorling:
> Would your proposal be adding that as a built-in method if needed?

Yes, exactly. That way we could avoid all the "throw new RuntimeException(
e )" we do, making resulting stack traces a bit nicer. But I guess we can
track that separately now that this PR is merged already.

sjaakd:
I'm not sure for the following reasons:
1. We always put 'readable' code first. I don't think this kind of hack helps understand  the code, since the reader has to understand type-erasure (which is a tricky mechanism anyway in java).
2. I use a dedicated `TransformException` for all transformations (hand-written & generated). So I want to make sure that the caller of any method on a `Mapper` needs to catch the `TransformException` by contract.
3. As @agudian  puts it: its a bit hacky.

Perhaps we should think about introducing a `MappingException` at one time ourselves, to solve non resolvable mapping conflicts in the future. Up to now there's no need for it.

Issue #551, opened 2015-05-17
Integration test SimpleTest#orcale_java_6 fails


agudian:
```java
java.lang.AssertionError: isSubtype 15
	at com.sun.tools.javac.code.Types$5.visitType(Types.java:355)
	at com.sun.tools.javac.code.Types$5.visitType(Types.java:336)
	at com.sun.tools.javac.code.Types$DefaultTypeVisitor.visitWildcardType(Types.java:3205)
	at com.sun.tools.javac.code.Type$WildcardType.accept(Type.java:416)
	at com.sun.tools.javac.code.Types$DefaultTypeVisitor.visit(Types.java:3203)
	at com.sun.tools.javac.code.Types.isSubtype(Types.java:332)
	at com.sun.tools.javac.code.Types.isSubtype(Types.java:316)
	at com.sun.tools.javac.code.Types.isSubtypeUnchecked(Types.java:287)
	at com.sun.tools.javac.code.Types.isConvertible(Types.java:256)
	at com.sun.tools.javac.code.Types.isAssignable(Types.java:1484)
	at com.sun.tools.javac.code.Types.isAssignable(Types.java:1445)
	at com.sun.tools.javac.model.JavacTypes.isAssignable(JavacTypes.java:86)
	at org.mapstruct.ap.processor.creation.MappingResolverImpl$ResolvingAttempt.collectionTypeHasCompatibleConstructor(MappingResolverImpl.java:556)
	at org.mapstruct.ap.processor.creation.MappingResolverImpl$ResolvingAttempt.isPropertyMappable(MappingResolverImpl.java:510)
	at org.mapstruct.ap.processor.creation.MappingResolverImpl$ResolvingAttempt.getTargetAssignment(MappingResolverImpl.java:195)
	at org.mapstruct.ap.processor.creation.MappingResolverImpl$ResolvingAttempt.access$100(MappingResolverImpl.java:146)
	at org.mapstruct.ap.processor.creation.MappingResolverImpl.getTargetAssignment(MappingResolverImpl.java:114)
	at org.mapstruct.ap.model.PropertyMapping$PropertyMappingBuilder.build(PropertyMapping.java:181)
	at org.mapstruct.ap.model.BeanMappingMethod$Builder.applyPropertyNameBasedMapping(BeanMappingMethod.java:419)
	at org.mapstruct.ap.model.BeanMappingMethod$Builder.build(BeanMappingMethod.java:127)
[...]

Issue #455, opened 2015-02-10
Reconsider error handling during processing


agudian:
That one did cost me some nerves today: if some unexpected stuff happens during processing, we tend to throw and finally catch an AnnotationProcessingException.

Although we report that error message to the Messager, it did not show up anywhere in Eclipse and I just ended up seeing no classes being produced. That's totally OK if the types we process contain compile errors.

But what we can't do, I'd say, is to throw the exception for other errors that do not show up by the compiler already or that do not end up as error located on a source element. For example, we throw the exception in two places in TypeFactory. Eclipse would just swallow them (as it did for me, leaving me wonder what the heck was going on). The only thing that helped me pinning down the issue was to throw a runtime exception in those cases, which then did show up in the eclipse error log - and led me to the discovering we had the problem described in issue #454.

So I'd say we go through all points where we throw that exception and check if we not better throw something that is not caught in the MappingProcessor.

PR #552, opened 2015-05-18
#455 Add a fallback error reporting in case of any uncaught exceptions occur…

agudian:
otherwise those errors are just swallowed and hard to find. Like with #551.

The problem is that error reporing (at least in javac), only prints out the first line, so I'm replacing newlines of the stacktrace with a simple space.

Not the most beautiful output, but it helps as a last resort for internal errors.


agudian:
I'll do that and then I'll merge. Thanks for the review!

PR #555, opened 2015-05-19
#551, fix for isAssignable problem in oracle_java_6

sjaakd:
Problem is similar to isSubtype problem. Fixed it in a similar way, and only on this call spot. There are more call spots. I don't think there's to much harm in most of the other places (mostly exception handling), but there could be issues in the future.

So, if we see this popping up again, replace the isAssignable to the `SpecificCompilerWorkarounds#isAssignable` version.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/processor/creation/MappingResolverImpl.java
+++ b/processor/src/main/java/org/mapstruct/ap/processor/creation/MappingResolverImpl.java
@@ -47,6 +47,7 @@ import org.mapstruct.ap.model.source.selector.MethodSelectors;
 import org.mapstruct.ap.model.source.selector.SelectionCriteria;
 import org.mapstruct.ap.util.FormattingMessager;
 import org.mapstruct.ap.util.Message;
+import org.mapstruct.ap.util.SpecificCompilerWorkarounds;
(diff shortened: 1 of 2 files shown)


agudian:
Sounds reasonable to me. Great that you had extended the IT in that other PR, so that we saw the error early on.

I'm okay with it to be merged. @gunnarmorling: do you have anything to add here?

sjaakd:
> Sounds reasonable to me. Great that you had extended the IT in that other PR, so that we saw the error early on.

@ your advice :smile: :+1:

agudian:
I honestly forgot - the self-praise was unintended.... :grin:

sjaakd:
I'll go ahead with merging. I'm a bit annoyed by the failing Jenkins build reports. @gunnarmorling : if you have additional remarks I'll write an add-on PR :smile:.

sjaakd:
> the self-praise was unintended.

But justified... :smile:

agudian:
> I'll go ahead with merging. I'm a bit annoyed by the failing Jenkins build reports. @gunnarmorling : if you have additional remarks I'll write an add-on PR :smile:.

Great, I was just thinking the same thing, as that mail popped in... :+1:

Issue #512, opened 2015-03-20
Error description does not point to the troublesome mappings


sjaakd:
![screencap](https://cloud.githubusercontent.com/assets/6173283/6752483/34da6562-cf0f-11e4-90b6-edd5793ae3b7.png)

PR #559, opened 2015-05-25
#512, more descriptive error message for forged collection and map ma…

sjaakd:
…ppings

The error handling for forging methods is now handled by the property mapping. At that spot was already a good message. For 'plain' Iterable and MapMappings the error stays the same, since the location hint is sufficient in that case


gunnarmorling:
Looks good. Just merged it, only altered the commit message to remove the "," after the issue number. If you don't mind I'd prefer that pattern. Thanks, @sjaakd!

PR #561, opened 2015-05-26
#365 Making property names customizable

gunnarmorling:
Based on the work of @chschu this provides the naming strategy contract. I have simplified the contract a bit, moved the SPI package to the processor module and added an integration test. The latter is in a separate module for now, it may be added to the existing integrationtest module, but I couldn't make that work for the time being as the test needs two modules: one with the strategy, one which is using it.


parent/pom.xml
```diff
@@ -207,6 +207,11 @@
                 <version>${project.version}</version>
             </dependency>
             <dependency>
+                <groupId>${project.groupId}</groupId>
+                <artifactId>mapstruct-spi</artifactId>
```
agudian: that's not used anymore, right?

processor/src/main/java/org/mapstruct/ap/naming/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+/**
+ * <p>
+ * SPI (service provider interface) contracts to be implemented for advanced use cases.
```
agudian: Is this the right doc for this package? As far as I can see, this package currently only holds the default implementation of this one SPI...?

processor/src/main/java/org/mapstruct/ap/util/Executables.java
```diff
@@ -44,49 +43,30 @@
  */
 public class Executables {
 
+    private static AccessorNamingStrategy accessorNamingStrategy = Services.get(
```
agudian: Couldn't this also be `final`?

agudian:
Looks good so far - I just have some small remarks.

If it's all the same for you, I would like to try to move the integration-test into the integration-test module (will do that later tonight)...

parent/pom.xml
gunnarmorling: True, that can be removed.

processor/src/main/java/org/mapstruct/ap/naming/package-info.java
gunnarmorling: Hum, yes, seems like I mixed those up.

processor/src/main/java/org/mapstruct/ap/util/Executables.java
gunnarmorling: +1

gunnarmorling:
> Looks good so far - I just have some small remarks.

Cool, thanks for the review.

> I would like to try to move the integration-test into the integration-test module

+1 That'd be my preference, I just struggled to do so :)

Would you like to apply those other changes you pointed out together with the move?

agudian:
> Would you like to apply those other changes you pointed out together with the move?

Sure, I'll go over it later tonight.

Perhaps that helps you finding some time to.... you know... that other pr... :wink:

gunnarmorling:
Yes, believe it or not, it's on my to-do list for tonight :) Thanks,
Andreas! Btw. there was quite some nice buzz on MapStruct on Twitter today
:)

agudian:
I've worked in the comments, rebased the commits, moved the integration-test to the integration-test module (plain and simple, supporting only the `ORACLE_JAVA_8` processor type) and pushed to master.
@chschu, @gunnarmorling: thanks for this cool feature!

> Btw. there was quite some nice buzz on MapStruct on Twitter today :)
Great! :+1:

chschu:
@gunnarmorling @agudian Thanks for your effort, this looks great now!

gunnarmorling:
@chschu Thank you, too! Glad to hear you like the final outcome. Great to have this feature in place now.

Issue #556, opened 2015-05-21
Add customization for generated class names/bean names


ggtools:
When using the Spring component model, the generated classes are annotated with `@Component`. With this default configuration Spring will use the class name as the *bean* name. While this ok most of the time it might lead to name collisions if your mapper has a very common name such as `ParameterMapper`

The solution could be either to make the implementation prefix customizable or have a way to add a qualifier to the `@Component` annotation

gunnarmorling:
Thanks for the report! You are the second one to bring up this request :)

Would it work for you if we made this customizable through `@Mapper`/ `@MapperConfig`?

    @Mapper(implSuffix="MySuffix")?
    public class MyMapper { ... }

ggtools:
Yes that'll be perfect.

gunnarmorling:
Cool, would you maybe be interested in preparing a pull request for this? We've scheduled the RC1 for next week, I won't be able to do it myself by then, not sure about the others :) It should be a good one to get into the code base, just follow up how other `@Mapper(Config)` properties are handled and pass the file name to the template writer.

ggtools:
I'll try to find some time to prepare a PR so it can be included in the next release.

gunnarmorling:
Awesome, you rock, @ggtools!

ggtools:
Still try to work on this although time is a rare resource

gunnarmorling:
Cool, thanks for the update, @ggtools. We've moved the CR1 to next week now. If it's not in there, it's no big deal as we can do follow-up releases very soon just as needed.

gunnarmorling:
I am moving this to 1.1 for now; But if have a PR for this some time soon, we can do a CR2 including that feature.

agudian:
Fixed. Thanks, @ggtools!

Hamzaaton:
Hi @gunnarmorling, the "Impl" suffix worked (and still works) fine.  
I am trying to obfuscate my code source, so all the class names are modified to (a, b, c, d..), and I get a conflict for the Mappers and other Components. 
Is it possible to add an option that allows to generated Component to hava a name with which we can call with the Qualifier annotation?

Z-100:
I don't know if I'm the only one, but for me the `implementationName = "Something"` only worked while using it in the parentheses after @Mapper. @MapperConfig didn't work.

filiphr:
@Z-100 this issue was closed 6 years ago. If something is not working then please create a new issue.

Issue #472, opened 2015-02-24
Allow to specify target package for generated mapper implementation


gunnarmorling:

0x3333:
I was going to report and found this issue.

My use case is:

Implementations goes to a sub package called "impl".

Ex.:

com.mycompany.project.some.MyMapper
com.mycompany.project.some.impl.MyMapperImpl

Thanks.

gunnarmorling:
Thanks for your feedback! Out of interest, which component model are you using? CDI, Spring? Or are you working with the `Mappers` class provided by MapStruct?

0x3333:
Hi! I'm using Guice as DI, and a custom made framework tied to Guice ecosystem. That's why we have "namespace" like the one I said. The Mapper is from MapStruct itself. Great tool by the way.

gunnarmorling:
Ok, I see.

I was asking because I'm wondering how we could make this work when the `Mappers` factory is used (which you do not use AFAIU). This factory currently simply appends "Impl" to the name of the requested mapper interface, this would not work when this implementation class lives in another package. 

@sjaakd, you suggested to generate the "binding" from mapper interfaces to implementations so that the factory can use it. But where would such binding class live? I think it only can be in the mapper package itself (otherwise we'd just move the problem by one level). Or do you have another idea?

I'm still a bit reluctant to actually access that information from the annotations themselves at runtime, as that'd mean we'd need to re-implement the logic in the factory class (which would be no problem when only taking the `@Mapper` class into account, but I think we should also make the implementation package configurable via `@MapperConfig`). But maybe it's not such a big deal after all? WDYT?

gunnarmorling:
> Great tool by the way.

Thanks, glad to hear it's helpful to you :)

0x3333:
In my case I bind the Interface with the generated mapper statically.

My two cents is that u can generate a Binding class, e.g. my.company.mapper.MyMapper$Binding.
The class should live in the same package the mapper interface(@Mapper). The generated Mapper(Impl) can live in another alien package. The $Binding class can be used as a provider, in DI context and by the Mappers too.

agudian:
I would find it odd to habe a class in the package of the mapper interface that references the .impl package - in many projects, there are code analysis rules that restrict such references or the resulting package cycles. 
How about we create files in Meta-inf for SPI-style retrieval in case of the default component model?

gunnarmorling:
[...]

PR #565, opened 2015-05-31
#556 make the implementation name configurable

ggtools:
Implementation of `ServiceLoader` instead of duplicating the logic in `Mappers`. Issue #472 should be able to use this new mechanism.

Details:

- Add `implSuffix` in `@Mapper` & `@MapperConfig`

- Generate a file in `META-INF/services` when a custom suffix has been
  configured

- `Mappers` use `ServiceLoader` first in order to get a mapper's
  implementation before falling back to the former method


sjaakd:
@ggtools : first of all, many thanks for your PR.  I've added some remarks. 

My biggest concern is changing the retention policy on the `@Mapper` and `@MappingConfig`. We did have some discussions on that one in the past. There is a suggestion on how we could avoid that. But perhaps @gunnarmorling can shed his thoughts on this as well.

gunnarmorling:
> @ggtools : first of all, many thanks for your PR

+1, Many thanks!

> My biggest concern is changing the retention policy on the @Mapper and @MappingConfig. We did have some discussions on that one in the past

Right, changing the retention to `RUNTIME` seems not right. See the discussion in #472, the idea is to generate file(s) which provide the binding from mapper interface to implementation type and make these mapping types discoverable using the Java service loader mechanism, i.e. by generating a file under `META-INF/services`.

E.g. the contract could look like that:

    public interface MapperFactory {
       <T> T getMapper(Class<T> mapperInterface);
    }

MapStruct would generate an implementation of this interface which knows of all mappers of a compiliation unit (i.e. JAR) and the services file. The existing `Mappers` class would use the `java.util.ServiceLoader` to locate all mapper factories present at runtime and delegate to them to find requested mappers.

ggtools:
I agree with the discussion in #472 and I have to re-implement the same logic during the generation and at runtime. I never used `java.util.ServiceLoader` but I might have a look at it.

ggtools:
I'm wondering how the *service* file will be generated. I guess it'll be along with mappers' implementation. That means we couldn't split in several jars classes that have been generated at the same time. I'm not really sure it'll be an actual use case but I just had this thought.

gunnarmorling:
> That means we couldn't split in several jars classes that have been generated at the same time

The service loader mechanism can deal with several service files for the same SPI contract, all implementations from all files will be found. So we could generate one file per compilation unit (e.g. a Maven module) and the `Mappers` factory would find all implementations at runtime (e.g. in a multi-module Maven project).

ggtools:
I'll try to work on this next week-end

gunnarmorling:
Cool; Thanks for the heads-up. We'll do the next release today or tomorrow, but we can do a CR2 in case your change is ready e.g. next week.

gunnarmorling:
@ggtools Any news on that one?

ggtools:
I had a look and I need to dive into the Mapstruct generation code it'll probably take me some time. Sorry

gunnarmorling:
>  it'll probably take me some time.

Ok, no problem. Let's schedule it for 1.1 then.

> In a nutshell when a mapper is configured with a specific prefix [...]

Yes, this sounds about right. Just let me know in case you run into any issues with that. Btw. related to this is the customization of the package where generated mappers should go (see #472), we will benefit from the same mechanism there. In fact it should be very quickly be done, once the isse at hand has been implemented.

gunnarmorling:
Can we close this one now that #602 is there?

gunnarmorling:
Can we close this one now that #602 is there?

ggtools:
Sure since it is included in #602.

ggtools:
Closing it, see in #602

PR #566, opened 2015-06-03
#272 Decorator support for Spring and CDI

gunnarmorling:
Supersedes #503. For Spring this allows to inject the decorated mapper without any qualifier, provided the decorator is annotated with `@Primary` and the delegate is injected into the decorator using the "delegate" qualifier. For CDI nothing really is to do on our side, we only may consider to raise a warning if `@DecoratedWith` is used, as it's not needed in this case.


gunnarmorling:
@agudian, @sjaakd: Quickly merged this myself to get the release going. If you find anything suspicious, let's create a follow-up issue.

Issue #574, opened 2015-06-25
Missing resolution of generic type-vars of property accessors


agudian:
I have came across a bean class structure with a generic abstract superclass:

```java
public abstract class AbstractIdHoldingTo<ID> {
    private ID id;
    public ID getId() {..}
    public void setId(ID id) {..}
}
public abstract class AbstractTo extends AbstractIdHoldingTo<Long> {
}
public class TargetTo extends AbstractTo {
}
```

When mapping from/to TargetTo, MapStruct doesn't get that the property type of `id` is `Long`.

agudian:
The good thing is that we have everything in place already (the trick in `TypeFactory.getMethodType()` that uses `asMemberOf(..)`), which made the fix fairly simple.

Pull-Request follows right away.

gunnarmorling:
Done.

PR #575, opened 2015-06-25
#574 determine property type from getter/setter methods only in combination with ExecutableType

agudian:
… i.e. using asMemberOf to determine the type with any generics being resolved.

One thing that might still be open is the handling of lower/upper bounds, in case the generic type variables are not resolved completely. But that' s an improvement that can wait for 1.1, I'd say...


gunnarmorling:
Rebased and applied. Many thanks, @agudian!

> But that' s an improvement that can wait for 1.1, I'd say...

+1. Can you open an issue for it, providing an example of that situation? Thanks!

agudian:
> Can you open an issue for it, providing an example of that situation? Thanks!

Done. Thanks for merging this! :+1:

Issue #577, opened 2015-06-26
IndexOutOfBoundsException when mapping 2 arrays


lukaszostaniewicz:
Hi, 
```
public class A {
    protected int[] a;
    protected int[] b;
(+ getters/setters)
}
```
```
public class B {
    protected List<Integer> a;
    protected List<Integer> b;
(+ getters/setters)
}
```
```
@Mapper
public interface AtoBMapper {
    public B map(A source);
}
```
<artifactId>mapstruct-processor</artifactId>
 <version>1.0.0.CR1</version>

When classes contain 1 array/list only the mapping generated looks fine. 
When classes contain 2 arrays/lists as in the example above then the mapping generation leads to

AtoBMapper.java:6: error: Internal error in the mapping processor: 

```
java.lang.IndexOutOfBoundsException: Index: 0
at java.util.Collections$EmptyList.get(Collections.java:3212)   at org.mapstruct.ap.internal.model.IterableMappingMethod.equals(IterableMappingMethod.java:305)        
 at java.util.ArrayList.indexOf(ArrayList.java:298)      at java.util.ArrayList.contains(Array
List.java:281)         
 at org.mapstruct.ap.internal.model.PropertyMapping$PropertyMappingBuilder.forgeMapOrIterableMapping(PropertyMapping.java:484)   
 at org.mapstruct.ap.internal.model.PropertyMapping$PropertyMappingBuilder.build(PropertyMapping.java:195)
at org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.applyPropertyNameBasedMapping(BeanMappingMethod.java:426)         
 at org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.build(BeanMappingMethod.java:127)          
 at org.mapstruct.ap.internal.
processor.MapperCreationProcessor.getMappingMethods(MapperCreationProcessor.java:349)   
at org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMapper(MapperCreationProcessor.java:143)      
at org.mapstruct.ap.internal.processor.MapperCreationProcesso
r.process(MapperCreationProcessor.java:114)     
at org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:73)         
at org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:220)         
at org.mapstruct.ap.M
appingProcessor.processMapperTypeElement(MappingProcessor.java:200)     
[...]

PR #578, opened 2015-06-26
Fix IndexOutOfBounds exception during mapping two arrays to collections.

vgt-tomek:
Hello,

I'm not sure if issue 577 reported recently should be fixed in IterableMappingMethod:equals() method. Maybe getTypeParameters() should never be empty and fix is needed elsewhere. But if it is indeed just missing check for empty collection, then I prepared fix for this issue.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
@@ -302,8 +302,10 @@ public class IterableMappingMethod extends MappingMethod {
         }
 
         for ( int i = 0; i < getSourceParameters().size(); i++ ) {
-            if ( !getSourceParameters().get( i ).getType().getTypeParameters().get( 0 )
-                .equals( other.getSourceParameters().get( i ).getType().getTypeParameters().get( 0 ) ) ) {
+            List<Type> thisTypeParameters = getSourceParameters().get( i ).getType().getTypeParameters();
+            List<Type> otherTypeParameters = other.getSourceParameters().get( i ).getType().getTypeParameters();
+
+            if ( !thisTypeParameters.equals( otherTypeParameters ) ) {
                 return false;
             }
         }
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_577/Issue577Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_577/Issue577Test.java
@@ -1,0 +1,42 @@
+/**
+ *  Copyright 2012-2015 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
```diff
@@ -302,8 +302,13 @@ public boolean equals(Object obj) {
...
         for ( int i = 0; i < getSourceParameters().size(); i++ ) {
-            if ( !getSourceParameters().get( i ).getType().getTypeParameters().get( 0 )
-                .equals( other.getSourceParameters().get( i ).getType().getTypeParameters().get( 0 ) ) ) {
+            List<Type> thisTypeParameters = getSourceParameters().get( i ).getType().getTypeParameters();
+            List<Type> otherTypeParameters = other.getSourceParameters().get( i ).getType().getTypeParameters();
+
+            if ( thisTypeParameters.isEmpty() && otherTypeParameters.isEmpty() ) {
+                continue;
+            }
+            if ( !thisTypeParameters.get( 0 ).equals( otherTypeParameters.get( 0 ) ) ) {
```
agudian: I think the case that `thisTypeParameters` is empty and `otherTypeParameters` is not empty (and vice versa) is missing. Perhaps you could check on `...size() != ...size()` (as above with the sourceParameters) and then iterate over them to check each parameter pair?

agudian:
First of all: thanks for the PR! :+1: 

I think it's the right approach... I just have that one remark on the correctness of the check...

vgt-tomek:
Thank you for response and suggestion. It was a bit weird for me, that this part of code was expecting type parameters collection to always contain exactly one element, but I'm not familiar with the rest of the code, so I decided to assume, that it's correct.
But since it's not, I've just pushed commit with changes based on your suggestion. At this point only thing missing would be null-checks on those collections, but they were not there before, so most likely not needed here?

agudian:
> At this point only thing missing would be null-checks on those collections, but they were not present before, so most likely not needed here?

Right, the lists are never null... damn, we really should include some `@Nonnull` annotations here and there...

Anyway, your solution looks right to me know.. Although, after reading it, the size-check plus the element-wise `equal(..)` in the loop looked just too familiar. The javadoc of `List.equals(..)` defines exactly that to be what needs to be implemented by any List implementation anyway. So, this would also do it: 

```java
if ( !thisTypeParameters.equals(otherTypeParameters) ) {
   return false;
}
```

Sorry for the back-and-forth... 

While you're at it, and as I think it's the last thing that needs changing: could you please squash the commits into one and add "#577 " as prefix of the commit message?

And finally, under which name can we add you to the list of contributors in the copyright.txt file? :smiley:

vgt-tomek:
> Sorry for the back-and-forth...

No problem, that what PR's are for. :)

By squashing to one commit you mean to rewrite history on this branch? It's my first PR on GitHub, so I'm not sure if this won't mess up something, since rewriting is not advisable, when branch is already on remote. In case it will mess up, I can create PR again.

I see, that you're using full names in your copyright.txt file, so we can follow that schema and full name from my profile can be used. :)

agudian:
> It's my first PR on GitHub, so I'm not sure if this won't mess up something, since rewriting is not advisable, when branch is already on remote.

GitHub is pretty smart with handling that sort of things... Especially at the end of a discussion, it's totally ok to rebase and squash all the incremental changes and rewrite the remote branch, so that we can have a clean fast-forward merge on master...

So, this one is merged. Thanks a lot for your quick help! :+1:

gunnarmorling:
Many thanks for handling this, guys!

Issue #576, opened 2015-06-25
Assignability-check for @MappingTarget in before-/after-mapping methods is wrong


agudian:
I stared at those lines of code so many times, and yet, it doesn't work as I wanted it to. For example:

```java
public abstract class BaseDto {..}
public abstract class CarDto extends BaseDto {..}

@Mapper
public abstract class CarMapper {
   public abstract CarDto toCarDto(Car entity);

   @AfterMapping
   public void registerMapping(@MappingTarget BaseDto anyBaseDto) {
       ...
   }
}
```

I would expect that for the mapping to `CarDto`, the invocation of the callback `registerMapping` is generated, as the result type `CarDto` is assignable to the `@MappingTarget` parameter. But that currently only works if the `@MappingTarget` parameter is `java.lang.Object` (as in the unit-tests), or if it is the same type as the mapping result type.

gunnarmorling:
> I would expect that for the mapping to CarDto, the invocation of the callback registerMapping is generated

+1, that'd be my expectation as well.

gunnarmorling:
Done.

PR #579, opened 2015-06-26
#576 Fix assignability check for @MappingTarget parameters in BeforeMapping/AfterMapping methods

agudian:
I tried out a different way to organize the required classes for the test case: they are all located in the test-mapper. Some of our tests come with a large number of classes around them, which makes locating the actual test class a bit harder.

@gunnarmorling, @sjaakd: what do you think of that?


gunnarmorling:
> @gunnarmorling, @sjaakd: what do you think of that?

Works for me, I don't have any strong preference I think :) I guess I'd use separate packages for source and target types.

gunnarmorling:
Looks good. Rebased and applied. Thanks!

Issue #580, opened 2015-06-29
No null check when converting from XMLGregorianCalendar to LocalDate


CyNoSureDev:
The generated mapping for mapping from XMLGregorianCalendar fields to LocalDate fail if the from field contains a null value.

Snippets (generated code):
```java
 obj.setDeliveryDate(LocalDate.parse(xmlGregorianCalendarToString(xmlGregorianCalendar, null))
```
```java
if (date != null) {
    dto.setDeliveryDate(stringToXmlGregorianCalendar(DateTimeFormatter.ISO_LOCAL_DATE.format(date), null));
}
```

agudian:
Yep, that's not cool. There should be a null-check on `xmlGregorianCalendar` in the first snippet, just as it is generated for the second one.

Thanks for the report!

gunnarmorling:
Done.

PR #582, opened 2015-06-30
#580 Add null-check for mapped -> type-converted property mapping.

agudian:
At least our built-in mapping methods return null for null input, which would break the type-conversion.

@sjaakd, I think this null-check was left out intentionally to give users the chance to manually map null-values to non-null values? Could you please give this a quick check?

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -47,6 +47,7 @@ import org.mapstruct.ap.internal.util.Message;
 import org.mapstruct.ap.internal.util.Strings;
 
 import static org.mapstruct.ap.internal.model.assignment.Assignment.AssignmentType.DIRECT;
+import static org.mapstruct.ap.internal.model.assignment.Assignment.AssignmentType.MAPPED_TYPE_CONVERTED;
 import static org.mapstruct.ap.internal.model.assignment.Assignment.AssignmentType.TYPE_CONVERTED;
 import static org.mapstruct.ap.internal.model.assignment.Assignment.AssignmentType.TYPE_CONVERTED_MAPPED;
 
@@ -262,7 +263,8 @@ public class PropertyMapping extends ModelElement {
                     && !sourceReference.getPropertyEntries().isEmpty() /* parameter null taken care of by beanmapper */
                     && ( result.getType() == TYPE_CONVERTED
                     || result.getType() == TYPE_CONVERTED_MAPPED
-                    || result.getType() == DIRECT && targetType.isPrimitive() ) ) {
+                    || result.getType() == MAPPED_TYPE_CONVERTED
+                    || ( result.getType() == DIRECT && targetType.isPrimitive() ) ) ) {
                     // for primitive types null check is not possible at all, but a conversion needs
                     // a null check.
                     result = new NullCheckWrapper( result );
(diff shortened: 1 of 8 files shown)


sjaakd:
> @sjaakd, I think this null-check was left out intentionally to give users the chance to manually map null-values to non-null values? Could you please give this a quick check?

In general: (builtin) methods are responsible in their own method body to do a null check. Type conversions should have a null check at the call site.

Background: when  calling a (builtin) method is called, a source null value could have meaning.

agudian:
> Background: when calling a (builtin) method is called, a source null value could have meaning.

That's what I remembered as well as the reason not to do a null-check.

The alternative would be to add a local variable that is assigned with the result of the property-mapping method and do the null-check on that variable before handing it to the conversion.

agudian:
Hmm, handling the general case would require some major hacking. If I have a mapping-method to map from X to Y and then use the type conversion from Y to Z, then we would need to generate something like this:

```java
Y prop = mappingMethodCall( source.getProp() );
if ( prop != null ) {
    target.setProp( LocalDate.parse( prop ) );
}
```

That doesn't really fit the way we create and stick together the `Assignment`s right now. I'm sure I would introduce more bugs with this than I solve... :wink: 
What I could do more easily is adding the null-check on the source-property as it is done in this PR, but only in case we selected a built-in mapping method. But that's somehow fishy as well, as our generated mapping methods also return `null` by default, so that would cause NPEs there as well.

So I'm tempted to leave the solution as suggested in this PR, i.e. always do the null-check on the source property for the case of mapping+conversion. Users who want to supply special values for `null` input would have to create mapping methods that match the target without the need for the additional conversion.

WDYT, @sjaakd + @gunnarmorling?

sjaakd:
I think you have a point ! Any interference expected with update methods?

agudian:
> Any interference expected with update methods?

Interesting. Can we call an update-method and put the result into a conversion? That's not something that makes sense, does it? My gut says no-ish, but I don't know for sure.

sjaakd:
AFAIK two step update methods are not yet supported! There should be an issue on that. But plain conversion or plain mapping is.  Why wouldn't it make sense?

agudian:
What I'm talking about is this: source-property has type X, target property has type Z, and there is an update-method that takes X and _updates_ an instance of Y. And there is a conversion from Y to Z.
I guess you're right, this can make sense - do we support that already?

agudian:
Oh, actually (I just tried to do it) - when an update-method with void return type is attempted to be checked in the `resolveViaMethodAndConversion`, we end up with a NullPointerException in the MethodMatcher... I'll create an issue for that.

gunnarmorling:
@agudian, can you give me the details on this tomorrow night? I am somehow struggling to grasp the issue ;)

agudian:
> @agudian, can you give me the details on this tomorrow night? I am somehow struggling to grasp the issue ;)

I guess we forgot. If you still want to, we could phone or google-hangout?

gunnarmorling:
@agudian, sorry, just seeing your last comment now. Yes, would still be interested in having a talk. I'll ring you later this weekend :)

agudian:
As both issues are quite similar (and partially touch the same lines in the code), I've added the fix for #596 to this PR. Hope that's ok.

gunnarmorling:
@agudian This seems good, I will merge it as is. You mention though that there may be a more "proper" solution in general. If you still think it's worth it, could you open a new issue for that?

gunnarmorling:
Rebased and applied. Many thanks, @agudian!

Issue #581, opened 2015-06-29
ambigious packages generated


lukaszostaniewicz:
```
@Mapper
public interface A2AMapper {
    result.A[] asResult(List<source.A> source);
    result.A asResult(source.A source);
}
```
leads to
```
import result.A;
import source.A;
```
and
```
A2AMapperImpl.java:7: error: result.A is already defined in a single-type import
import source.A;
```

vgt-tomek:
I was able to reproduce this issue in test case scenario and I did find out, why is this happening.
Method TypeFactory:isImported() is used to decide, if class should be imported in generated file, or fully qualified name should be used. It's doing it by collecting all imports in hashMap with class simple name as key. This bug is happening, when two different simple names leads to the same import: Car and Car[]
Easiest solution, that seems to work here is just removing suffix in this method, when array is provided as simple name, as can be seen here: https://github.com/vgt-tomek/mapstruct/commit/3fdfa13891f47d999ddba2dbffdd8cbb902a4a4d
@agudian, is something like that, an acceptable solution, or should it be handled differently?

agudian:
That's kind of what I would have done myself. Just make sure that you remove all array dimensions and not just the first one :wink:

gunnarmorling:
+1 Seems reasonable. Thanks for taking this over, @vgt-tomek!

vgt-tomek:
Ok, then I should be able to finalize it and create pull request later today.

gunnarmorling:
Done.

PR #586, opened 2015-07-01
#581 Prevent multiple imports of similar simple class name from different packages

vgt-tomek:
Hello,

As discussed in bug ticket, I've finalized fix for #581 issue.
Few topics you might want to review, since I don't know yet, if you accept few code conventions I usually follow.

1. I consider good practice to not modify method parameters, to consider it final. Therefore when I need to modify it, I usually introduce new local variable and use it from there. That's the reason for adding *String trimmedName* in *isImported()* and use it instead of method parameter.

2. I moved code responsible for handling array brackets in simple class names to separate utility class, so I could add unit tests for it.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
@@ -405,7 +405,8 @@ public class TypeFactory {
     }
 
     private boolean isImported(String name, String qualifiedName) {
-        String importedType = importedQualifiedTypesBySimpleName.get( name );
+        String trimmedName = TypeFactory.trimSimpleClassName( name );
+        String importedType = importedQualifiedTypesBySimpleName.get( trimmedName );
 
         boolean imported = false;
         if ( importedType != null ) {
@@ -414,7 +415,7 @@ public class TypeFactory {
             }
         }
         else {
-            importedQualifiedTypesBySimpleName.put( name, qualifiedName );
+            importedQualifiedTypesBySimpleName.put( trimmedName, qualifiedName );
             imported = true;
         }
         return imported;
@@ -464,6 +465,17 @@ public class TypeFactory {
         }
 
         return collectionOrMap;
+    }
+
+    static String trimSimpleClassName(String className) {
+        if ( className == null ) {
+            return null;
+        }
+        String trimmedClassName = className;
+        while ( trimmedClassName.endsWith( "[]" ) ) {
+            trimmedClassName = trimmedClassName.substring( 0, trimmedClassName.length() - 2 );
+        }
+        return trimmedClassName;
     }
 
 }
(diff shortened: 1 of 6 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactoryUtil.java
```diff
@@ -0,0 +1,36 @@
...
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model.common;
+
+final class TypeFactoryUtil {
+
+    private TypeFactoryUtil() {
+    }
+
+    static String trimSimpleClassName(String className) {
```
gunnarmorling: What's the reason for having this method in a separate class?

gunnarmorling:
> I consider good practice to not modify method parameters, to consider it final. 

Ok, I see. That's alright, I don't really have a strong preference either way.

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactoryUtil.java
vgt-tomek: Here's the thought flow:

1. Of course at first this method can be private method inside TypeFactory class.
2. However I want it unit tested, therefore I need to have access to it from tests, therefore I increase access from private to package protected.
3. At this point it feels better to have it as separate package protected utility class.

Of course we can stop at point 2 and still have testable code, going to point 3 is more preference, than necessity.

gunnarmorling: My preference would be to have this method on `TypeFactory` itself. So if you don't mind, could you rework the commit so it does 2.?

processor/src/test/java/org/mapstruct/ap/internal/model/common/TypeFactoryUtilTest.java
```diff
@@ -0,0 +1,63 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model.common;
+
+import static org.fest.assertions.Assertions.assertThat;
+
+import org.junit.Test;
+
+public class TypeFactoryUtilTest {
```
gunnarmorling: Nice test! Could you add the `@IssueKey` annotation?

gunnarmorling:
@vgt-tomek, looks very good overall! If you could do that one rework I suggested, it's good to merge. Thanks a lot!

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactoryUtil.java
vgt-tomek: Sure, no problem. Like I said, I don't know yet, what code conventions you use and prefer, so that's why this is happening now. :)

vgt-tomek:
Method moved to TypeFactory class, as suggested.

gunnarmorling:
Merged; Thanks, @vgt-tomek!

Issue #590, opened 2015-07-02
NPE when calling resolveViaMethodAndConversion with update an method


agudian:
I have properties that can't be mapped and only an unsuitable update-method (with return type void). Then I get this exception instead of the "Can't map property ..." message.

```
java.lang.NullPointerException
    at org.mapstruct.ap.internal.model.source.MethodMatcher.matchSourceType(MethodMatcher.java:152)
    at org.mapstruct.ap.internal.model.source.MethodMatcher.matches(MethodMatcher.java:121)
    at org.mapstruct.ap.internal.model.source.SourceMethod.matches(SourceMethod.java:482)
    at org.mapstruct.ap.internal.model.source.selector.TypeSelector.getMatchingMethods(TypeSelector.java:44)
    at org.mapstruct.ap.internal.model.source.selector.MethodSelectors.getMatchingMethods(MethodSelectors.java:61)
    at org.mapstruct.ap.internal.processor.creation.MappingResolverImpl$ResolvingAttempt.getBestMatch(MappingResolverImpl.java:437)
    at org.mapstruct.ap.internal.processor.creation.MappingResolverImpl$ResolvingAttempt.resolveViaMethod(MappingResolverImpl.java:263)
    at org.mapstruct.ap.internal.processor.creation.MappingResolverImpl$ResolvingAttempt.resolveViaMethodAndConversion(MappingResolverImpl.java:394)
    at org.mapstruct.ap.internal.processor.creation.MappingResolverImpl$ResolvingAttempt.getTargetAssignment(MappingResolverImpl.java:234)
    at org.mapstruct.ap.internal.processor.creation.MappingResolverImpl$ResolvingAttempt.access$100(MappingResolverImpl.java:147)
    at org.mapstruct.ap.internal.processor.creation.MappingResolverImpl.getTargetAssignment(MappingResolverImpl.java:115)
    at org.mapstruct.ap.internal.model.PropertyMapping$PropertyMappingBuilder.build(PropertyMapping.java:180)
    at org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.applyPropertyNameBasedMapping(BeanMappingMethod.java:426)
    at org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.build(BeanMappingMethod.java:127)
    at org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMappingMethods(MapperCreationProcessor.java:349)
[...]

Issue #488, opened 2015-03-03
Follow up of #160, support all 2-step selection mechanisms for update mappings


sjaakd:
Supported now:
- direct
- updateMethod
- conversion
- updateMethod(conversion(source))
- updateMethod(createMethod(source))

Support still needed
- updateMethod(updateMethod(source))
- conversion(updateMethod(source))
- createMethod(updateMethod(source))

sjaakd:
is ok for now.. we can reopen when needed

PR #591, opened 2015-07-02
#590 Update methods are not supported for "method-and-conversion" methods…

agudian:
… (can currently cause NullPointerExceptions)

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_590/Issue590Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_590/Issue590Test.java
@@ -39,8 +39,7 @@ public class Issue590Test {
     @ExpectedCompilationOutcome(value = CompilationResult.FAILED,
         diagnostics = { @Diagnostic(type = SourceTargetMapper.class,
             kind = Kind.ERROR,
-            line = 34,
-            messageRegExp = "Can't map property \"[^ ]+ [^\"]+\" to \"[^ ]+ [^\"]+\"") })
+            messageRegExp = "Can't map property \"java\\.lang\\.String prop\" to \"[^ ]+ prop\"") })
     public void showsCantMapPropertyError() {
 
     }
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_590/Issue590Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_590/Issue590Test.java
@@ -1,0 +1,47 @@
+/**
+ *  Copyright 2012-2015 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Is there anything else to do - more generally - in addition to fix that NPE?

processor/src/test/java/org/mapstruct/ap/test/bugs/_590/Issue590Test.java
```diff
@@ -0,0 +1,47 @@
...
+@RunWith(AnnotationProcessorTestRunner.class)
+@WithClasses(SourceTargetMapper.class)
+public class Issue590Test {
+
+    @Test
+    @ExpectedCompilationOutcome(value = CompilationResult.FAILED,
+        diagnostics = { @Diagnostic(type = SourceTargetMapper.class,
+            kind = Kind.ERROR,
+            line = 34,
+            messageRegExp = "Can't map property \"[^ ]+ [^\"]+\" to \"[^ ]+ [^\"]+\"") })
```
gunnarmorling: This test makes me think wether we actually *should* support String <> StringBuilder as a built-in conversion? It's a separate issue, though.

processor/src/test/java/org/mapstruct/ap/test/bugs/_590/SourceTargetMapper.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * @author Andreas Gudian
+ */
+@Mapper(unmappedTargetPolicy = ReportingPolicy.WARN)
+public abstract class SourceTargetMapper {
+    public static final SourceTargetMapper INSTANCE = Mappers.getMapper( SourceTargetMapper.class );
+
+    public abstract void sourceToTarget(@MappingTarget Target target, Source source);
+
+    public <T extends Target> T unused(String string, @TargetType Class<T> clazz) {
```
gunnarmorling: So why is it expected that this is not used? I am somehow missing a piece :/

agudian:
> Is there anything else to do - more generally - in addition to fix that NPE?

Yes, as Sjaak pulled out in the issue: #488 describes the feature that is missing here and that is just explicitly disabled with this PR.

processor/src/test/java/org/mapstruct/ap/test/bugs/_590/SourceTargetMapper.java
agudian: I can debug some more, but without that intermediate String-to-SomethingNotUsable step that is suggested by this method `unused`, I didn't get the error.

agudian: OK, I got it now. It's because of how MappingResolverImpl currently works: in `resolveViaMethodAndConversion(..)` (and other `resolve..Method..`-variants), the list of all available methods is looped over, and for each of them the `Assignment` instance is created using `resolveViaMethod

processor/src/test/java/org/mapstruct/ap/test/bugs/_590/Issue590Test.java
agudian: Created. And I'll update the test to use a different type... :+1:

gunnarmorling:
Rebased and applied. Thanks, @agudian!

PR #599, opened 2015-07-07
Spring/JSR 330 decorators

agudian:
The funny part is the one with the jsr330 model - and how the `@Named(..)` is used... hmm...

Once everyone's ok with this, I'll add some code-examples to the javadoc of `@DecoratedWith`

### DIFF
--- a/integrationtest/src/test/resources/jsr330Test/src/main/java/org/mapstruct/itest/jsr330/SourceTargetMapperDecorator.java
+++ b/integrationtest/src/test/resources/jsr330Test/src/main/java/org/mapstruct/itest/jsr330/SourceTargetMapperDecorator.java
@@ -16,16 +16,21 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  */
-package org.mapstruct.itest.cdi;
+package org.mapstruct.itest.jsr330;
 
-import org.mapstruct.Mapper;
-import org.mapstruct.DecoratedWith;
-import org.mapstruct.itest.cdi.other.DateMapper;
+import javax.inject.Inject;
+import javax.inject.Named;
 
-@Mapper( componentModel = "cdi", uses = DateMapper.class )
-public interface DecoratedSourceTargetMapper {
+public abstract class SourceTargetMapperDecorator implements DecoratedSourceTargetMapper {
 
-    Target sourceToTarget(Source source);
+    @Inject
+    @Named("org.mapstruct.itest.jsr330.DecoratedSourceTargetMapperImpl_")
+    private DecoratedSourceTargetMapper delegate;
 
-    Target undecoratedSourceToTarget(Source source);
+    @Override
+    public Target sourceToTarget(Source source) {
+        Target t = delegate.sourceToTarget( source );
+        t.setFoo( 43L );
+        return t;
+    }
 }
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -179,6 +179,12 @@
                 <version>${org.springframework.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>commons-logging</groupId>
+                <artifactId>commons-logging</artifactId>
+                <version>1.1.3</version>
+            </dependency>
+
             <!-- Joda-Time -->
             <dependency>
                 <groupId>joda-time</groupId>
--- a/processor/src/test/java/org/mapstruct/ap/test/decorator/spring/PersonMapperDecorator.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/decorator/spring/PersonMapperDecorator.java
@@ -16,16 +16,24 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  */
-package org.mapstruct.itest.cdi;
+package org.mapstruct.ap.test.decorator.spring;
 
-import org.mapstruct.Mapper;
-import org.mapstruct.DecoratedWith;
-import org.mapstruct.itest.cdi.other.DateMapper;
+import org.mapstruct.ap.test.decorator.Person;
+import org.mapstruct.ap.test.decorator.PersonDto;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 
-@Mapper( componentModel = "cdi", uses = DateMapper.class )
-public interface DecoratedSourceTargetMapper {
+public abstract class PersonMapperDecorator implements PersonMapper {
 
-    Target sourceToTarget(Source source);
+    @Autowired
+    @Qualifier("delegate")
+    private PersonMapper delegate;
 
-    Target undecoratedSourceToTarget(Source source);
+    @Override
+    public PersonDto personToPersonDto(Person person) {
+        PersonDto dto = delegate.personToPersonDto( person );
+        dto.setName( person.getFirstName() + " " + person.getLastName() );
+
+        return dto;
+    }
 }
--- a/processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
+++ b/processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
@@ -102,12 +102,7 @@ class CompilingStatement extends Statement {
         }
     };
 
-    private static final List<String> LIBRARIES = Arrays.asList(
-        "mapstruct.jar",
-        "guava.jar",
-        "javax.inject.jar",
-        "joda-time.jar"
-    );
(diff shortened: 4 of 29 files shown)


gunnarmorling:
@agudian Could you add a short description of the meat of this change? How will the generated code look different in the future with this change?

agudian:
> @agudian Could you add a short description of the meat of this change? How will the generated code look different in the future with this change?

Let's just take a look at the new tests in `org.mapstruct.ap.test.decorator.*`:

## Spring

```java
@DecoratedWith(PersonMapperDecorator.class)
public interface PersonMapper {

    @Mapping( target = "name", ignore = true )

    AddressDto addressToAddressDto(Address address); // not touched by the decorator
}

public abstract class PersonMapperDecorator implements PersonMapper {

    @Autowired
    @Qualifier("delegate")

    @Override
    public PersonDto personToPersonDto(Person person) {
        PersonDto dto = delegate.personToPersonDto( person );
        dto.setName( person.getFirstName() + " " + person.getLastName() );

        return dto;
    }
}
```
generates (run SpringDecoratorTest):
```java
@Component
@Qualifier("delegate")
public class PersonMapperImpl_ implements PersonMapper {

    @Override
    public PersonDto personToPersonDto(Person person) {
        ...
    }

    @Override
    public AddressDto addressToAddressDto(Address address) {
       ...
    }
}

@Component
@Primary
public class PersonMapperImpl extends PersonMapperDecorator implements PersonMapper {

    @Autowired
    @Qualifier("delegate")

    @Override
    public AddressDto addressToAddressDto(Address address)  {
        return delegate.addressToAddressDto( address );
    }

}
```

The user would have to autowire this (as it is done in the test class) to use the decorated mapper:
```java
    @Autowired
    private PersonMapper personMapper;
```

That looks pretty ok, I would say. The integration test verifies that this still works if I have two different mappers that are decorated (i.e. two mappers that end up having the same qualifier, but different types).

## JSR 330

```java
@Mapper(componentModel = "jsr330")
@DecoratedWith(PersonMapperDecorator.class)
public interface PersonMapper {

    @Mapping( target = "name", ignore = true )

    AddressDto addressToAddressDto(Address address);
}

public abstract class PersonMapperDecorator implements PersonMapper {

    @Inject
    @Named("org.mapstruct.ap.test.decorator.jsr330.PersonMapperImpl_") // this is the ugly part!

    @Override
    public PersonDto personToPersonDto(Person person) {
        PersonDto dto = delegate.personToPersonDto( person );
        dto.setName( person.getFirstName() + " " + person.getLastName() );

        return dto;
    }
}
```
generates (run Jsr330DecoratorTest):
```java
@Named("org.mapstruct.ap.test.decorator.jsr330.PersonMapperImpl_")
public class PersonMapperImpl_ implements PersonMapper {

    @Override
    public PersonDto personToPersonDto(Person person) {
        ..
    }

    @Override
    public AddressDto addressToAddressDto(Address address) {
          ...
    }
}

@Named
public class PersonMapperImpl extends PersonMapperDecorator implements PersonMapper {

    @Inject
    @Named("org.mapstruct.ap.test.decorator.jsr330.PersonMapperImpl_")

    @Override
    public AddressDto addressToAddressDto(Address address)  {
        return delegate.addressToAddressDto( address );
    }
}
```

The user would have to inject this (as it is done in the test class) to use the decorated mapper:
```java
    @Inject
    @Named
    private PersonMapper personMapper;
```

Because in JSR 330 there is no such thing as `@Primary`, we have to work with the empty `@Named` when injecting the decorated mapper and the `@Named("org.mapstruct.ap.test.decorator.jsr330.PersonMapperImpl_")` when injecting the undecorated mapper in the decorator implementation. Without that, it was not possible to have two (different) decorated mappers within the application (I've extended the integration-test for that).

Perhaps we can do that a bit nicer once we work on the issue that came up on the mailing list with the configurable bean name (for CDI, Spring and JSR330). Until then we could leave the JSR 330 support as "experimental"...?

processor/src/test/java/org/mapstruct/ap/test/decorator/spring/PersonMapperDecorator.java
```diff
@@ -0,0 +1,39 @@
...
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.decorator.spring;
+
+import org.mapstruct.ap.test.decorator.Person;
+import org.mapstruct.ap.test.decorator.PersonDto;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+
+public abstract class PersonMapperDecorator implements PersonMapper {
```
gunnarmorling: I suppose we should add a note to the docs that one should not add `@Primary`to the decorator anymore (or remove that hint if it's there).

integrationtest/src/test/resources/jsr330Test/src/main/java/org/mapstruct/itest/jsr330/SourceTargetMapperDecorator.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.mapstruct.itest.jsr330;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+public abstract class SourceTargetMapperDecorator implements DecoratedSourceTargetMapper {
+
+    @Inject
+    @Named("org.mapstruct.itest.jsr330.DecoratedSourceTargetMapperImpl_")
```
gunnarmorling: Hum, referring to the impl name really is ugly. But I guess there is no other way :(

parent/pom.xml
```diff
@@ -179,6 +179,12 @@
                 <version>${org.springframework.version}</version>
             </dependency>
 
+            <dependency>
```
gunnarmorling: What is that needed for? Can the slf4j bridge used in tests (assuming that's needed for Spring)?

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
```diff
@@ -102,12 +102,7 @@ protected CompilationCache initialValue() {
         }
     };
 
-    private static final List<String> LIBRARIES = Arrays.asList(
```
gunnarmorling: Nice to get rid of the copying below. But should we establish a filter / white-list of the deps to propagate we actually want to avoid exposing too much?

gunnarmorling: Btw. `AnnotationNotFoundTest` fails for me on this branch. Does it pass for you? I guess it is somehow related to this change here.

processor/src/test/java/org/mapstruct/ap/testutil/runner/ModifiableURLClassLoader.java
```diff
@@ -56,9 +56,6 @@ protected void addURL(URL url) {
         }
     }
 
-    /**
-     * @param basePath
```
gunnarmorling: :)

gunnarmorling:
@agudian Many thanks for walking me through this change! Added some comments, nothing special apart from that one test failure.

parent/pom.xml
agudian: That was a fix for a dependency-version convergence that came up after adding spring as test-dep to the processor. Something like that. I'll check again.

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
agudian: Hmm, works for me. In what way does the test fail for you? Did you mvn clean install first?

agudian: > But should we establish a filter / white-list of the deps to propagate we actually want to avoid exposing too much?

Good idea!

gunnarmorling: > Hmm, works for me. In what way does the test fail for you? Did you mvn clean install first?

It happens when running `mvn clean install`. The test expects a failed compilation due to the annotation type missing, but actually the compilation passes for me. I assume the annotation is made available to the test compilation by that changed approach for setting up the test class path.

agudian:
@gunnarmorling: didn't have time to work on the rest of the stuff I wanted to add, yet, so it's still not ready for merge. But could you please check the last commit and see if the test works now on your machine?

gunnarmorling:
@agudian Re-pulled, all tests passing now :)

agudian:
Great, then I'll work on the javadoc tonight and then squash and merge it.

gunnarmorling:
Thanks, @agudian!

PR #602, opened 2015-07-15
Make destination package and name configurable

### LINKED ISSUE #472: Allow to specify target package for generated mapper implementation

ggtools:
Fix #472 
- Add interfacePackage` to `GeneratedType`
- Add new `destPackage` to `Mapper` and `MapperConfig'

### DIFF
--- a/core-common/src/main/java/org/mapstruct/Mapper.java
+++ b/core-common/src/main/java/org/mapstruct/Mapper.java
@@ -85,6 +85,28 @@ public @interface Mapper {
     String componentModel() default "default";
 
     /**
+     * Specifies the name of the implementation class. A single {@code *} wildcard can be used and will be replaced
+     * by the interface's or abstract class' name.
+     * <p>
+     * Defaults to postfixing the name with {@code Impl}
+     *
+     * @return The implementation suffix.
+     * @see #implementationPackage()
+     */
+    String implementationName() default "default";
+
+    /**
+     * Specifies the target package for the generated implementation.  A single {@code *} wildcard can be used and will
+     * be replaced by the interface's or abstract class' package.
+     * <p>
+     * Defaults to using the same package as the mapper interface/abstract class
+     *
+     * @return the destination package.
+     * @see #implementationName()
+     */
+    String implementationPackage() default "default";
+
+    /**
      * A class annotated with {@link MapperConfig} which should be used as configuration template. Any settings given
      * via {@link Mapper} will take precedence over the settings from the referenced configuration source. The list of
      * referenced mappers will contain all mappers given via {@link Mapper#uses()} and {@link MapperConfig#uses()}.
--- a/core-common/src/main/java/org/mapstruct/Mapper.java
+++ b/core-common/src/main/java/org/mapstruct/Mapper.java
@@ -85,26 +85,26 @@ public @interface Mapper {
     String componentModel() default "default";
 
     /**
-     * Specifies the name of the implementation class. A single {@code *} wildcard can be used and will be replaced
-     * by the interface's or abstract class' name.
+     * Specifies the name of the implementation class. The {@code <CLASS_NAME>} will be replaced by the
+     * interface/abstract class name.
      * <p>
-     * Defaults to postfixing the name with {@code Impl}
+     * Defaults to postfixing the name with {@code Impl}: {@code <CLASS_NAME>Impl}
      *
-     * @return The implementation suffix.
+     * @return The implementation name.
      * @see #implementationPackage()
      */
-    String implementationName() default "default";
+    String implementationName() default "<CLASS_NAME>Impl";
 
     /**
-     * Specifies the target package for the generated implementation.  A single {@code *} wildcard can be used and will
-     * be replaced by the interface's or abstract class' package.
+     * Specifies the target package for the generated implementation. The {@code <CLASS_NAME>} will be replaced
+     * by the interface's or abstract class' package.
      * <p>
      * Defaults to using the same package as the mapper interface/abstract class
      *
-     * @return the destination package.
+     * @return the implementation package.
      * @see #implementationName()
      */
-    String implementationPackage() default "default";
+    String implementationPackage() default "<PACKAGE_NAME>";
 
     /**
      * A class annotated with {@link MapperConfig} which should be used as configuration template. Any settings given
--- a/core-common/src/main/java/org/mapstruct/MapperConfig.java
+++ b/core-common/src/main/java/org/mapstruct/MapperConfig.java
@@ -85,26 +85,26 @@ public @interface MapperConfig {
     String componentModel() default "default";
 
     /**
-     * Specifies the name of the implementation class. A single {@code *} wildcard can be used and will be replaced
-     * by the interface's or abstract class' name.
+     * Specifies the name of the implementation class. The {@code <CLASS_NAME>} will be replaced by the
+     * interface/abstract class name.
      * <p>
-     * Defaults to postfixing the name with {@code Impl}
+     * Defaults to postfixing the name with {@code Impl}: {@code <CLASS_NAME>Impl}
      *
-     * @return The implementation suffix.
+     * @return The implementation name.
      * @see #implementationPackage()
      */
-    String implementationName() default "default";
+    String implementationName() default "<CLASS_NAME>Impl";
 
     /**
-     * Specifies the target package for the generated implementation.  A single {@code *} wildcard can be used and will
-     * be replaced by the interface's or abstract class' package.
+     * Specifies the target package for the generated implementation. The {@code <CLASS_NAME>} will be replaced
+     * by the interface's or abstract class' package.
      * <p>
      * Defaults to using the same package as the mapper interface/abstract class
      *
-     * @return the destination package.
+     * @return the implementation package.
      * @see #implementationName()
      */
-    String implementationPackage() default "default";
+    String implementationPackage() default "<PACKAGE_NAME>";
 
     /**
      * The strategy to be applied when propagating the value of collection-typed properties. By default, only JavaBeans
--- a/core-common/src/main/java/org/mapstruct/MapperConfig.java
+++ b/core-common/src/main/java/org/mapstruct/MapperConfig.java
@@ -85,6 +85,28 @@ public @interface MapperConfig {
     String componentModel() default "default";
 
     /**
+     * Specifies the name of the implementation class. A single {@code *} wildcard can be used and will be replaced
+     * by the interface's or abstract class' name.
+     * <p>
+     * Defaults to postfixing the name with {@code Impl}
+     *
+     * @return The implementation suffix.
+     * @see #implementationPackage()
+     */
+    String implementationName() default "default";
+
+    /**
+     * Specifies the target package for the generated implementation.  A single {@code *} wildcard can be used and will
+     * be replaced by the interface's or abstract class' package.
+     * <p>
+     * Defaults to using the same package as the mapper interface/abstract class
+     *
+     * @return the destination package.
+     * @see #implementationName()
+     */
+    String implementationPackage() default "default";
+
+    /**
      * The strategy to be applied when propagating the value of collection-typed properties. By default, only JavaBeans
      * accessor methods (setters or getters) will be used, but it is also possible to invoke a corresponding adder
      * method for each element of the source collection (e.g. {@code orderDto.addOrderLine()}).
--- a/core-common/src/main/java/org/mapstruct/factory/Mappers.java
+++ b/core-common/src/main/java/org/mapstruct/factory/Mappers.java
@@ -75,20 +75,24 @@ public class Mappers {
                 classLoader = Mappers.class.getClassLoader();
             }
 
-            ServiceLoader<T> loader = ServiceLoader.load( clazz, classLoader );
+            try {
+                @SuppressWarnings("unchecked")
+                T mapper = (T) classLoader.loadClass( clazz.getName() + IMPLEMENTATION_SUFFIX ).newInstance();
+                return mapper;
+            }
+            catch (ClassNotFoundException e) {
+                ServiceLoader<T> loader = ServiceLoader.load( clazz, classLoader );
 
-            if ( loader != null ) {
-                for ( T mapper : loader ) {
-                    if ( mapper != null ) {
-                        return mapper;
+                if ( loader != null ) {
+                    for ( T mapper : loader ) {
+                        if ( mapper != null ) {
+                            return mapper;
+                        }
                     }
                 }
+
+                throw new ClassNotFoundException("Cannot find implementation for " + clazz.getName());
             }
-
-            @SuppressWarnings("unchecked")
-            T mapper = (T) classLoader.loadClass( clazz.getName() + IMPLEMENTATION_SUFFIX ).newInstance();
-
-            return mapper;
         }
         catch ( Exception e ) {
             throw new RuntimeException( e );
--- a/core-common/src/main/java/org/mapstruct/factory/Mappers.java
+++ b/core-common/src/main/java/org/mapstruct/factory/Mappers.java
@@ -20,6 +20,8 @@ package org.mapstruct.factory;
 
 import org.mapstruct.Mapper;
 
+import java.util.ServiceLoader;
+
 /**
  * Factory for obtaining mapper instances if no explicit component model such as CDI is configured via
  * {@link Mapper#componentModel()}.
@@ -71,6 +73,16 @@ public class Mappers {
 
             if ( classLoader == null ) {
                 classLoader = Mappers.class.getClassLoader();
+            }
+
+            ServiceLoader<T> loader = ServiceLoader.load( clazz, classLoader );
+
+            if ( loader != null ) {
+                for ( T mapper : loader ) {
+                    if ( mapper != null ) {
+                        return mapper;
+                    }
+                }
             }
 
             @SuppressWarnings("unchecked")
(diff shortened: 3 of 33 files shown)


ggtools:
PR #565 needs to be merged before this one

agudian:
@ggtools, thanks for this great Pull-Request!

I have some general questions after scrolling over the changes:

* Having to specify a full package name as String seems quite tedious to me. What if we supported something like `implPackage = "*.impl"`, which would expand `*` to the package name of the mapper itself (you wouldn't have to use the `*`, but you could). Likewise, we could change `implSuffix` to `implName = "*Impl"` to also allow something like `implName = "MapStructGenerated*Implementation"`. I know some projects where postfixes are used to designate the nature of the implementation (e.g. `JpaCarDao` as JPA-implementation of a `CarDao`).

* The generated service-registry files wouldn't be needed if a user configured the `componentModel` `cdi`, `spring` or `jsr330` - right? So I think I'd suppress the generation of them in those cases... Or, looking from the other direction - only create them in case of componentModel `default` (see `AnnotationBasedComponentModelProcessor` as an example where some finalization steps are done for the non-`default` component models). Or do you see a need for the service-registry files to be created in any case?

* The tests are located in the the `bugs` package and use the issue-numbers. We usually only use that naming pattern for actual bugs. For features, we use packages and test class names that somehow relate to the feature (trying to group as much as possible). So there could be a test like `..ap.test.destination.DestinationClassNameTest` with methods for the impl-name and the package-name configuration... or something similar... What you actually do test the existing tests is very nice though... :+1:

ggtools:
The wildcard is a brilliant idea I'm going to make the change as soon as I can find some time.

Regarding the removal of the service-registry if the component model is not default I'm not sure it's a good idea. I currently work on a Spring based application. However for complex mappers based on a abstract classes, I sometime do very basic unit tests without the Spring framework. In that case removing the service registry would make such tests a lot more complicated. It will also introduce a difference between the *plain* mappers with the default package & implementation name which will be retrievable through `Mappers` while customized mappers won't. Does this make sense?

ggtools:
I make some changes to implement the wildcard system as well as moving the test in the right package for both this PR and #565

agudian:
> However for complex mappers based on a abstract classes, I sometime do very basic unit tests without the Spring framework. In that case removing the service registry would make such tests a lot more complicated. It will also introduce a difference between the plain mappers with the default package & implementation name which will be retrievable through Mappers while customized mappers won't.

Ah, I didn't look at the changes close enough, yet. I thought we would now always fetch the mappers in the default-model using SPI. Which would also make sense, I guess - that way, our default-component model would essentially be SPI.
Regarding the testability, you're right for the simple cases. But as soon as there is a collaborator pulled in via `uses`, there's no way the mapper can be used without a Spring/CDI/JSR330 container. Same goes for decorated mappers (at least once one of my other PRs is merged :wink:). So what I'm getting at is that the mappers are probably not usable anyway and we should make that clear by not even suggesting that the Service-Locator could be used.

One thing that draws my eye in the latest changes is the intrdoduction of the constant in `@Mapper` - we should avoid that. We don't usually have a runtime-dependency to the annotation-jar in the processor, so we shouldn't use them there (we actually have a checkstyle rule for that, but I guess the `import static` outsmarts that check :wink:). You also had the values `"default"` in `@Mapper` previously to cleanly detect if a value was specified in `@MapperConfig` but overwritten in `@Mapper`. I think that should come back in... :)

The tests look great, as does all the rest. I think we'll hit the finish line soon... :grin:

processor/src/main/java/org/mapstruct/ap/internal/model/Mapper.java
```diff
@@ -31,6 +31,8 @@
 import org.mapstruct.ap.internal.option.Options;
 import org.mapstruct.ap.internal.version.VersionInformation;
 
+import static org.mapstruct.Mapper.DEFAULT_IMPL_NAME;
```
gunnarmorling: As @agudian is saying there should be no compile dependency to the annotation module.

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -34,6 +34,7 @@
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.CLASS)
 public @interface Mapper {
+    String DEFAULT_IMPL_NAME = "*Impl";
```
gunnarmorling: I like the idea of the wild card mechanism, but I find the "*" a bit unintuitive. I'd prefer to use a more descriptive placeholder, such "<MAPPER_NAME>".

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -85,6 +86,28 @@
...
     /**
+     * Specifies the name of the implementation class. A single {@code *} wildcard can be used and will be replaced
+     * by the interface's or abstract class' name.
+     * <p>
+     * Defaults to {@code *Impl} which means: <i>postfix the name with {@code Impl}</i>
+     *
+     * @return The implementation suffix.
+     * @see #implPackage()
+     */
+    String implName() default DEFAULT_IMPL_NAME;
```
gunnarmorling: Could we avoid the abbreviations in the public API, i.e. `implementationName` and `implementationPackage`. Reads a bit nicer.

ggtools:
No problem to remove the dependency. For `default` I initially though of using it but eventually I had to test against the *actual* default value. For instance not generate the SPI stuff if someone put `implName="*Impl"`. But for consistency sake I can use `default` in the annotations.

core-common/src/main/java/org/mapstruct/Mapper.java
ggtools: I agree that `*` is not super intuitive but what do you mean by `""`?

ggtools:
So my last update:

- Remove the runtime dependency to Mapper
- Replace `impl(Name|Package)` by `implementation$1`
- Use `default` instead of `*Impl` or `*` for default value in annotations

There still a couple of things to check

- Change the wildcard from `*` to something more intuitive
- Check if we want to disable the SPI generation when the component model is not default

core-common/src/main/java/org/mapstruct/Mapper.java
gunnarmorling: Doh, sorry, GitHub masked some characters. I meant something like "&lt;CLASS_NAME&gt;" and "&lt;PACKAGE_NAME&gt;".

gunnarmorling:
Thanks for the update, @ggtools! I'll will take a closer look at this PR tonight or tomorrow.

gunnarmorling:
> Change the wildcard from * to something more intuitive

+1, my preference would be `&lt;CLASS_NAME&gt;` / `&lt;PACKAGE_NAME&gt;`

> Check if we want to disable the SPI generation when the component model is not default

+1 I don't think it's needed when using Spring or CDI as component model. Also we could disable it, if neither class nor package name are customised.

core-common/src/main/java/org/mapstruct/factory/Mappers.java
```diff
@@ -73,6 +75,16 @@ private Mappers() {
                 classLoader = Mappers.class.getClassLoader();
             }
 
+            ServiceLoader<T> loader = ServiceLoader.load( clazz, classLoader );
```
gunnarmorling: It'd be better to do it as a fall-back only after the normal approach, as I suppose using the service loader is slower.

processor/src/main/java/org/mapstruct/ap/internal/model/Mapper.java
```diff
@@ -129,16 +135,34 @@ public Builder extraImports(SortedSet<Type> extraImportedTypes) {
             return this;
         }
 
+        public Builder implName(String implName) {
+            this.implName = "default".equals( implName ) ? "*Impl" : implName;
```
gunnarmorling: Could you extract constants for "*" (or whatever we end up using for instead)? Makes it easier to change :)

gunnarmorling:
@agudian, you said

> But as soon as there is a collaborator pulled in via uses, there's no way the mapper can be used without a Spring/CDI/JSR330 container.

Is that true? The generated code should use `Mappers#get()` to set up the field with the collaborator in that case, benefiting from the service loader mechanism, too.

core-common/src/main/java/org/mapstruct/factory/Mappers.java
ggtools: Sure I also though about implementing a cache to let Mappers keep track of what implementation needs to be used but I think we should do some benchmarks first

ggtools:
I'll update my PR soon. Regarding removing the SPI thingy when Spring/CDI/JSR330 is selected I still have a bad feeling about this as this'll break the upward compatibility and I don't think having the service file configuration still here will have much impact.

ggtools:
Ok so I made some changes. Regarding the generation of the SPI files I introduced a new processor properties to be able to force the generation of the SPI files if needed.

gunnarmorling:
> Regarding the generation of the SPI files I introduced a new processor properties to be able to force the generation of the SPI files if needed.

+1 That sounds very reasonable.

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -85,6 +85,28 @@
...
     /**
+     * Specifies the name of the implementation class. A single {@code *} wildcard can be used and will be replaced
+     * by the interface's or abstract class' name.
+     * <p>
+     * Defaults to postfixing the name with {@code Impl}
+     *
+     * @return The implementation suffix.
+     * @see #implementationPackage()
+     */
+    String implementationName() default "default";
```
gunnarmorling: I wish we could use "<CLASS_NAME>Impl" as the default name (because that's what will be used), but I don't think we can because otherwise we could not distinguish the case where a specific `@Mapper` annotation overrides the default from `@MapperConfig` from the case where the inherited default should be used.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -234,6 +241,11 @@ else if ( constructor.getParameters().size() == 1 ) {
             extraImports.add( type );
         }
 
+        // Add original package if a dest package has been set
+        if ( !"*".equals( mapperConfiguration.implementationPackage() ) ) {
```
gunnarmorling: That's a left-over from the previous name I think, right?

core-common/src/main/java/org/mapstruct/Mapper.java
ggtools: Yes that was why I changed this back to `default`

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
ggtools: Just pushed a commit to fix this

core-common/src/main/java/org/mapstruct/Mapper.java
agudian: > I wish we could use "<CLASS_NAME>Impl" as the annotation default value (because that's what will be used), but I don't think we can because otherwise we could not distinguish the case where a specific @Mapper annotation overrides the default from @MapperConfig from the case where the inherited default should be used.

I thought about this the other day while driving on a four hour trip.... The thing is, actually we could distinguish it - but as we're using hickory to generate the annotation-prisms, we're used to rely on those prisms to access the annotation values. The APT API would give us the information if a value was explicitly specified or not - the hickory-generated prims just return the default values in those cases, for convinience.
I'll create a speparate issue where we can discuss/find a better solution and with that we can then clean up all our annotation definitions.

ggtools:
I just rebase my branch from the current master

agudian:
@ggtools, I've just pushed a PR that handles the detection of specified / unspecified annotations more nicely, which allows us to avoid using "default" as default values in the annotations.

Check the code examples here: https://github.com/mapstruct/mapstruct/pull/623/files#diff-edd112063ebfe5afe8f83d4a0ad78377L81

core-common/src/main/java/org/mapstruct/MapperConfig.java
```diff
@@ -85,6 +85,28 @@
...
     /**
+     * Specifies the name of the implementation class. A single {@code *} wildcard can be used and will be replaced
+     * by the interface's or abstract class' name.
+     * <p>
+     * Defaults to postfixing the name with {@code Impl}
+     *
+     * @return The implementation suffix.
+     * @see #implementationPackage()
+     */
+    String implementationName() default "default";
```
gunnarmorling: Ok, with @agudian's change in place, the actual default value can be used now.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperServiceProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        if ( !context.isErroneous() && spiGenerationNeeded && mapper.hasCustomImplementation() ) {
+            writeToSourceFile( context.getFiler(), mapper );
+        }
+        return null;
+    }
+
+    @Override
+    public int getPriority() {
```
gunnarmorling: Could you change it to 10000 (and the `MapperRenderingProcessor` to 9999). That's the interval specified in the docs of `getPriority()`.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperServiceProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+import javax.annotation.processing.Filer;
+import javax.lang.model.element.TypeElement;
+import javax.tools.FileObject;
+import javax.tools.StandardLocation;
+import java.io.IOException;
+
+/**
+ * @author Christophe Labouisse on 12/07/2015.
+ */
+public class MapperServiceProcessor  implements ModelElementProcessor<Mapper, Void> {
```
gunnarmorling: A short class comment would be nice.

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
```diff
@@ -101,6 +101,7 @@
         "mapstruct.suppressGeneratorVersionInfoComment";
     protected static final String UNMAPPED_TARGET_POLICY = "mapstruct.unmappedTargetPolicy";
     protected static final String DEFAULT_COMPONENT_MODEL = "mapstruct.defaultComponentModel";
+    protected static final String ALWAYS_GENERATE_SPI = "mapstruct.alwaysGenerateSpi";
```
gunnarmorling: I'd name it `mapstruct.alwaysGenerateServicesFile`.

gunnarmorling:
@ggtools I've added some more minor comments. The major missing piece is to use the proper default values as it's possible now with the latest changes to master. Once these are addressed, it's good to go from my side. 

Many thanks for contributing this great feature and doing the review dance. Sorry for having been a bit slow with reacting to this PR, so a big thanks for your patience, it's much appreciated :)

processor/src/test/java/org/mapstruct/ap/test/destination/DestinationPackageNameTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * @author Christophe Labouisse on 27/05/2015.
+ */
+@IssueKey( "556" )
+@RunWith(AnnotationProcessorTestRunner.class)
+public class DestinationPackageNameTest {
+    @Test
+    @WithClasses({ DestinationPackageNameMapper.class })
+    public void shouldGeneratedInRightPackage() throws Exception {
```
gunnarmorling: a "d" too much in the method name.

processor/src/test/java/org/mapstruct/ap/test/destination/DestinationPackageNameTest.java
```diff
@@ -0,0 +1,80 @@
...
+        DestinationPackageNameMapper instance = DestinationPackageNameMapper.INSTANCE;
+        assertThat( instance.getClass().getName() )
+                .isEqualTo( "org.mapstruct.ap.test.destination.dest.DestinationPackageNameMapperImpl" );
+    }
+
+    @Test
+    @WithClasses({ DestinationPackageNameMapperWithSuffix.class })
+    public void shouldGeneratedInRightPackageWithSuffix() throws Exception {
```
gunnarmorling: Same here.

processor/src/test/java/org/mapstruct/ap/test/destination/DestinationClassNameTest.java
```diff
@@ -0,0 +1,88 @@
...
+    @Test
+    @WithClasses({ DestinationClassNameMapper.class })
+    public void shouldGenerateRightName() throws Exception {
+        DestinationClassNameMapper instance = DestinationClassNameMapper.INSTANCE;
+        assertThat( instance.getClass().getSimpleName() ).isEqualTo( "MyDestinationClassNameMapperCustomImpl" );
+    }
+
+    @Test
+    @WithClasses({ DestinationClassNameWithJsr330Mapper.class })
+    public void shouldNotGenerateSpi() throws Exception {
```
gunnarmorling: Is there a test for using this option with another component model than the default one?

agudian: this test uses JSR 330, at least... I'd say that is sufficient. When this is merged, I'll have to check if pr #599 still works with reconfigured class/package names (That PR also adds the posibility to fully initialize and inject JSR 330 and Spring mappers in our unit-tests - so I'll just extend the tests then).

ggtools:
OK everything should be good now

agudian:
@ggtools, I've squashed your commits into a single one and added one additional commit to fix a copy-paste error in the annotation's javadoc. It's all on the master branch now.

Thanks a lot for all this work! It's a great feature! :+1:

Issue #604, opened 2015-07-21
Missing import when using update-method as property-mapping


agudian:
There is an import missing in case an update-method is used for a property mapping:

```java
        if ( target.getProp() == null ) {
            target.setProp( new TargetProp() );
        }
        targetPropMapper.toTargetProp( source.getProp() , target.getProp() );
```

gunnarmorling:
Done.

PR #605, opened 2015-07-21
#604 Add import for property target type when using an update-method

agudian:
This PR builds on PR #582, as the same classes were touched (and I wanted to make sure that I don't break anything that I fixed in those two issues contained in #582).


gunnarmorling:
Rebased and applied. You know what comes next: Many thanks, @agudian :)

agudian:
Thanks for the merge-orgy, Gunnar! :+1:

Issue #600, opened 2015-07-11
Feature Request: Possibility to provide default value


luxmeter:
Hi guys,

First of all, I appreciate your work and I'm happy that MapStruct exists. Keep it up!

To my request:
I've encountered a situation where I want to define a default value in case the source attribute is null.
Something like that:
```
@Mapper(target="countryCode", source="country.code", default="DE")
```

I solved my problem through a workaround:
```
@Mapper(target="countryCode", expression="java((source.getCountry() != null) ? source.getCountry().getCode() : \"DE\";)")
```

Would be awesome if this code snippet could be generated automatically.

Thank you!

sjaakd:
Hi @luxmeter .. We have a `@NullMappingStrategy`. Please checkout [here](http://mapstruct.org/documentation/#section-14). Perhaps this can be extended with a 'constant' parameter somehow. That 'constant' parameter would work in line with our current constant @gunnarmorling WDYT?

Best regards,
Sjaak

luxmeter:
Am I right that the Strategy is applied to all cases where a null is encountered? I just want to specify default values for specific attributes.

Best regards

sjaakd:
> Am I right that the Strategy is applied to all cases where a null is encountered? I just want to specify default values for specific attributes.

That's indeed right.

gunnarmorling:
`NullValueMappingStrategy` applies to mapping methods themselves, but not single properties. Somehow re-using may lead to confusions IMO. Having a new attribute `@Mapping#defaultValue()` seems nicer to me.

luxmeter:
Hi,

I decided to contribute and try to add the feature by myself. @agudian gave me already a little introduction. Let's see how far I come. I will inform you about my progress but will take some time...

BR

gunnarmorling:
@luxmeter Thanks for giving this a try! Just let us know in case you hit any road blocks. So how would the behaviour around `defaultValue()` exactly be? I'd say it'd be ignored when using either `constant()` or `expression()` and only be used as default when accessing a source property (either implicitly or by `source()` being given, right?

luxmeter:
@gunnarmorling Hi, that's also what I would say. Will try to implement it today. Let's see how far I come.

luxmeter:
[...]

PR #616, opened 2015-08-08
#600 added defaultValue feature

luxmeter:
A quick overview:

* Added a `defaultValue` property for the Mapping annotation (also for java8)
* If the `defaultValue` is set and the source type is not a primitive, the defaultValueAssignment is built
  - This is done by using the `ConstantMappingBuilder`. I bet that this is a overkill but I wasn't sure what I need really to build the assignment
* Adapted the BeanMappingMethod.ftl, PropertyMapping.ftl, NullCheckWrapper.ftl and SetterWrapper.ftl to render the assignment in case the defaultValueAssignment is available

### DIFF
--- a/core-jdk8/src/main/java/org/mapstruct/Mapping.java
+++ b/core-jdk8/src/main/java/org/mapstruct/Mapping.java
@@ -154,4 +154,16 @@ public @interface Mapping {
      * @return the dependencies of the mapped property
      */
     String[] dependsOn() default { };
+
+    /**
+     * <p>
+     * In case the source property is null the provided default {@link String} value is set.
+     * If the designated target
+     * property is not of type {@code String}, the value will be converted by applying a matching conversion method or
+     * built-in conversion.
+     * </p>
+     * Either this attribute or {@link #constant()} or {@link #expression()} may be specified for a given mapping.
+     * @return Default value to set in case the source property is null.
+     */
+    String defaultValue() default "";
 }
--- a/core/src/main/java/org/mapstruct/Mapping.java
+++ b/core/src/main/java/org/mapstruct/Mapping.java
@@ -151,4 +151,16 @@ public @interface Mapping {
      * @return the dependencies of the mapped property
      */
     String[] dependsOn() default { };
+
+    /**
+     * <p>
+     * In case the source property is null the provided default {@link String} value is set.
+     * If the designated target
+     * property is not of type {@code String}, the value will be converted by applying a matching conversion method or
+     * built-in conversion.
+     * </p>
+     * Either this attribute or {@link #constant()} or {@link #expression()} may be specified for a given mapping.
+     * @return Default value to set in case the source property is null.
+     */
+    String defaultValue() default "";
 }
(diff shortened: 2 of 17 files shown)


gunnarmorling:
Many thanks, @luxmeter! We will take a look at this PR asap. In the meantime. could you rebase it to the latest master (I've merged some other PRs in parallel). Just force-push to this branch and I'll re-pull then. Thanks!

luxmeter:
done :)

core-jdk8/src/main/java/org/mapstruct/Mapping.java
```diff
@@ -154,4 +154,15 @@
...
      */
     String[] dependsOn() default { };
+
+    /**
+     * In case the source property is null the provided default {@link String} value is set.
+     * If the designated target
+     * property is not of type {@code String}, the value will be converted by applying a matching conversion method or
+     * built-in conversion.
+     *
+     * The default value is ignored when using either {@link #constant()} or {@link #expression()} for the mapping.
```
agudian: Instead of ignoring the default value, an error should pop up. Just like when specifiying both `constant` and `expression`. Check the usage of `Message.PROPERTYMAPPING_EXPRESSION_AND_CONSTANT_BOTH_DEFINED` for an example.

agudian:
I've played around a bit with it. It works great with the normal case, but I'm not sure if the generated code is really how it should look like, because IMO the null-check to see if the default value needs to be applied should always be done on the source property - not on the possible result of a source property mapping.

Example: I've added a method "interner(String):String" to the mapper in the test case, which calls `intern()` on the input value. So the target objects only gets populated with interned values... ;). Anyway, this is how the generated code looks:
```java
        countryDts.setCode( interner( country.getCode() ) );
        if ( interner( country.getCode() ) == null ) {
            countryDts.setCode( interner( "DE" ) );
        }
```

I think it should rather look like this:
```java
        if ( country.getCode() != null ) {
            countryDts.setCode( interner( country.getCode() ) );
        else {
            countryDts.setCode( interner( "DE" ) );
        }
```

WDYT, do you see a way to change it to that direction? Or is there a reason not to do it this way?

luxmeter:
Sounds reasonable to me. Fixed it and force-pushed again.

agudian:
Looks better already. One question: I see that especially in the tests, the code-formatting is a bit off (e.g. tabs instead of spaces). Is that coming with our mapstruct code-formatter settings? Or were those files not formatted correctly?

luxmeter:
Fixed. Regarding the formatting: I had to tweak a bit the settings since I imported the Eclipse settings into IntelliJ. I couldn't even choose the IntelliJ settings. Probably because I'm using the community edition currently. As I played with the settings, I accidentally changed the tab/space settings. Shouldn't happen again.

agudian:
Okay, I've rebased your commit, added another one to slightly improve the generated code and javadoc... and I've added you to the list of contributors... :+1: 

It's all on master now.

Thanks again for this! Hope to see more from you - just let us know if you're in the mood :wink:

gunnarmorling:
Nice, many thanks for this! One more thing, could you also send a pull request for the documentation (see https://github.com/mapstruct/mapstruct.org/)?

agudian:
> One more thing, could you also send a pull request for the documentation

I've just committed the doc-change for this to a new branch that we can use for the next release.

Issue #603, opened 2015-07-15
Java 8 - Cant use static method on interface with decorators.


martin0esteves:
When adding a static method to an interface (java 8) for using in an expression the decorator created will try to override it and calling it using the mapper interface instance given to the decorator.

It tries to build a decorator instance that overrides mapA() and calls it using the delegate inside that method, which gives the following two errores while compiling:

- "error: method does not override or implement a method from a supertype"
- "error: illegal static interface method call" <--- Cant call interface static method using instances

Instead, I belive it should ignore the static method.

Code example that fails:

    @Mapper
    @DecoratedWith(IMapperDecorator.class)
    public interface IMapper {
        IMapper INSTANCE = Mappers.getMapper(IMapper.class);
      
        @Mapping(target = "atributeA", expression = "java( IMapper.mapA() )" )
        AtributeAClass mapToAtributeAClass(ClassWithA classWithA);
    
        static String mapA() { return null; }
    }

gunnarmorling:
Hi @martin0esteves, thanks for the report. I am not surprised, we didn't not really take static (interface) methods into the consideration. Would you be interested in providing a fix maybe?

martin0esteves:
Hello Gunnar, thank you for the offer. I would like to implment this fix but I´m short of time at the moment. Anyways, I do not discart coming back to implement it sometime.

bedla:
Consider PR at https://github.com/mapstruct/mapstruct/pull/617 . Thx, Ivos

agudian:
Merged the fix by @bedla.

PR #617, opened 2015-08-09
Issue 603

bedla:
Hi, I have created Pull request for  issue #603 . First commit is about refactoring test-utils to seperated module and moving processor-tests into own module.
Second commit dependent on first commit and filters out statis and default methods from interface. Third commit is about refactoring classes from upstream to honour changes from first commit.
Hope you will like it.
Ivos

### DIFF
--- a/integrationtest/src/test/resources/java8Test/src/main/java/org/mapstruct/ap/test/bugs/_603/Source.java
+++ b/integrationtest/src/test/resources/java8Test/src/main/java/org/mapstruct/ap/test/bugs/_603/Source.java
@@ -1,0 +1,22 @@
+/**
+ *  Copyright 2012-2015 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.bugs._603;
+
+public class Source {
+}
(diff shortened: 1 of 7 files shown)


gunnarmorling:
Many thanks, Ivo! Going to look at it soon.

gunnarmorling:
Hey @bedla, what's the motivation for moving all the tests into a separate module? I think it's a good practice to have tests close to the code they test, i.e. in the same module.

agudian:
The "only" problem is source/target type 1.6 in the processor-module, right? Maven-wise, we could use different configurations for main and test compilation, but the IDEs would get in trouble then.

Still, I would also favour keeping the tests in the same module. Until we start doing any fancy stuff with default methods, we could just rely on the Java 8 integration test (see Java8Test and src/test/resources/java8Test/... in the integration-test module). Running it in the IDE requires you to `mvn install` the processor module first, but that should be ok for corner cases like this. (For hints on debugging the integration tests, see http://mapstruct.org/contributing/#section-04).

gunnarmorling:
> Maven-wise, we could use different configurations for main and test compilation, but the IDEs would get in trouble then.

Right, for Maven it should be no problem. What happens when importing that set-up into Eclipse? I am open for giving it a try. If it's not workable then +1 for amending the existing Java 8 test in the integration test module and leaving all others in the processor module as is.

bedla:
Hi, right it is because of Java8 specific syntax - and I agree that it is better to have test close enough to classes they test.
I can revert separation to modules and make test 1.8 compatible in pom.xml... I hope that methods with body are the only one difference between 1.7 and 1.8 syntax.
Ivos

bedla:
I have created test branch without module separation https://github.com/bedla/mapstruct/tree/ISSUE-603-no_module ... It looks ok from Maven perspective, but IntelliJ IDEA is not happy with it because you can set diferent language level per sources and test sources (you can vote for issue here https://youtrack.jetbrains.com/issue/IDEA-85478 ). I dont know how about Eclipse and other editors.

agudian:
> I dont know how about Eclipse and other editors.

Yeah, I don't know about any editor that can currently do it. But thanks for trying it out! If it worked with IntelliJ, then I would have added a tweak to switch to complete Java 8 in Eclipse "automatically". But we shouldn't go down that path then...

So I'd suggest extending the Java8Test project in the integrationtest-module, as discussed above... :-/ 

What do you think?

bedla:
Fixed typo in prev comment :)

Anyways I can extend Java8Test, it looks good for now. Would you like to create new Pull request? Or I can force push this/my remote branch... :]

Ivos

agudian:
Great, please just do the force-push to the existing remote. :+1:

bedla:
Branch updated :-)

agudian:
Looks great! 
If you could just remove the compiler-config in the processor/pom.xml again and squash all your commits into one, I'll merge it to master tonight. :-)

bedla:
Done :-)

agudian:
@bedla merged it and I've added you to the list of contributors. Again, many thanks for your work and for your patience with us! Feel free to send us more stuff any time you like :wink:.

gunnarmorling:
Thanks, guys!

bedla:
No problem. It is pleasure to work on project like this.
Ivos

Issue #622, opened 2015-08-13
Blank constants report missing mapping


johnhearn:
Hi,
When mapping a constant to a target string property, if the constant is an empty string such as:
```java
@Mapping(constant = "", target = "string") 
```
then a the mapping is ignored and an error *Missing target property: "string"* is reported. If the constant is changed to any non-blank value (for example, to a space character) the warning disappears.
This is not the behaviour I was expecting since an empty string is being explicitly set to the target property. I have been able to set a blank string using the workaround:

```java
@Mapping(expression = "java(\"\")", target = "string")
```
though this is obviously less pleasing.

agudian:
Once #618 is fixed, we should be able to solve this quite nicely.

johnhearn:
Sorry, just noticed that this is a duplicate of #588

agudian:
Right - closing it :)

Thanks anyway for the report! :+1:

gunnarmorling:
Merged the fix for #618 to master.

Issue #588, opened 2015-07-02
Empty constant String result in unmapped target property


paul999:
```
public interface CodeSpaceConstants {
    String CS_CONSTRUCTION_STANDARD = "";
    String CS_WELLHEAD_PROTECTOR = "";
}
```

```
    @Mappings( {
        @Mapping( target = "codeSpace", constant = CodeSpaceConstants.CS_CONSTRUCTION_STANDARD )
    })
    abstract ConstructionStandardType toConstructionStandardType( String value );
```
Results in
Unmapped target property "Codespace"

while expected
Mapped to empty String.

gunnarmorling:
Hum, that's a tough one. Because we cannot really distinguish constant from not being given at all vs. being given with the empty string. Same for expression() and source(), the empty string is the default value for all of these. One solution would be to come up with a "magic value" representing null as the default value. But it would have to be something which is not a legal, constant, expression or source name.

agudian:
I think that's not possible at all, as we treat the `constant` expression for Strings as literals - and there is no invalid String literal that we could use as "default" in the annotation definition.

As a workaround, this is possible:
```java
    String EXPR_EMTPY_STRING = "java( \"\" )";

    @Mapping( target = "codeSpace", expression = EXPR_EMPTY_STRING) 
    ConstructionStandardType toConstructionStandardType( String value );
```

gunnarmorling:
If we use something which is not a legal Java identifier - such as `%NULL%` - it should be alright. No one would ever want to use that as an actual value for `constant`, `expression` or `source`.

agudian:
Once #618 is fixed, we should be able to solve this quite nicely.

gunnarmorling:
Done. Turns out the JSR 269 API allows to distinguish the case where the value explicitly set to the default value of the attribute from the case where no value is given and thus the default applies. Thus setting `constant=""` can be supported now.

PR #623, opened 2015-08-13
Detect unspecified annotation attributes

### LINKED ISSUE #618: Better way to distinguish between defined and undefined annotation values
Currently we use the default value "default" for annotation attributes if we want to be able to check if a value was explicitly specified or not.
That's because the hickory-prisms don't allow us to check if an attribute was specified or not, but just returns the default in case it wasn't defined. The APT API can be used to get that information, though.

I see two possibilities:
1. We fork hickory and extend the generated source to have methods like `componentModelDefined():boolean` in addition to `componentModel():String`.
2. We create our own wrapper and a simple metamodel so we can still access the attributes in a generic and still type-safe way (with unit-tests that check that the metamodel fits the state of the annotations).

Option 2 might be the easiest, I guess.

Once we have that in place, we can go through our annotations and add the real "default" values that are also documented in the javadoc.

agudian:
This fixes #618 and #588 (and the duplicate #622).

#602 (configurable destination packages and implementations) should be aligned to this solution as well. See examples in MapperConfiguration: https://github.com/mapstruct/mapstruct/pull/623/files#diff-edd112063ebfe5afe8f83d4a0ad78377L81

### DIFF
--- a/core-common/src/main/java/org/mapstruct/BeanMapping.java
+++ b/core-common/src/main/java/org/mapstruct/BeanMapping.java
@@ -61,5 +61,5 @@ public @interface BeanMapping {
      *
      * @return The strategy to be applied when {@code null} is passed as source value to the methods of this mapping.
      */
-    NullValueMappingStrategy nullValueMappingStrategy() default NullValueMappingStrategy.DEFAULT;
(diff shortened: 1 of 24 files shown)


gunnarmorling:
Very nice change, @agudian! Thanks!

Issue #636, opened 2015-09-03
default methods (Java 8) do not work anymore


jdubois:
This was working with 1.0.0.CR1 and is now failing with 1.0.0.CR2.

For JHipster we use interface default methods in our mappings, see for example:

https://github.com/jhipster/jhipster-sample-app-dto/blob/2dbdea12a49f3adcc0c16bf33c808c66a7bb439c/src/main/java/com/mycompany/myapp/web/rest/mapper/BankAccountMapper.java#L22

And since 1.0.0.CR2 those do not work anymore.

This issue makes our sample project fail, see https://github.com/jhipster/generator-jhipster/issues/1971

Of course this can be solved by migrated to an abstract class (I tested and it works), but I liked the interface with default methods better. It's less code and looks better to me.

gunnarmorling:
Thanks for the report, will look into it asap. There has been one issue in CR2 which seems related: https://github.com/mapstruct/mapstruct/issues/603.

bedla:
Consider Pull request #638 
Ivos

sjaakd:
@jdubois : do you see a possibility to give the proposed solution in #638 a try and see if this fixes the issue?

sjaakd:
@bedla, @lamperi, I assume that PR #638 replaces #637.. Am I correct here?

lamperi:
@sjaakd Yes we both made a fix but it seems like @bedla is more complete as it seems to do something with static methods too. Mine was my first look at mapstruct's source code and only altered the previous commit which was rewritten by bedla anyway.

I will close mine since #638 probably gets the job done better.

sjaakd:
@lamperi .. No problem.. thanks a lot for your contribution.

sjaakd:
should be fixed now.. .Please have a try.

LuccaPrado:
I´m  having the same problem but with java 11 with 1.3.0.Final. There's some way to bypass this?

filiphr:
@LuccaPrado please open new issues if something is not working. Anyways default methods should work. I am using that in other projects and there are no problems. Can you provide an example to show the problem?

LuccaPrado:
> 
> 
> @LuccaPrado please open new issues if something is not working. Anyways default methods should work. I am using that in other projects and there are no problems. Can you provide an example to show the problem?

Sorry, I went further an saw that there's an Issue for 1.3.0-Final about builder with aftermapping and beforemapping!
So just ignore my comment here

PR #637, opened 2015-09-04
#636 Fix default methods

lamperi:
It seems like fixing of #603 triggered this bug by assuming that static
interface methods and default interface methods should be treated the
same way. Reverted the behaviour on default methods and added a test
case for it.


lamperi:
Pull request #638 is probably better fix.

Issue #631, opened 2015-08-19
NullPointerException when using generics in mapper interface


gunnarmorling:
From [SO](http://stackoverflow.com/questions/32096886/mapstruct-mapping-of-generics):

    @Mapper
    public interface MyMapper<X extends Base_1, Y extends Base_2>{
        public X mapBase_2ToBase_1(Y obj);
    }

This causes the following NPE:

    java.lang.NullPointerException
    at org.mapstruct.ap.util.SpecificCompilerWorkarounds.replaceTypeElementIfNecessary(SpecificCompilerWorkarounds.java:90)
    at org.mapstruct.ap.util.Executables.getAllEnclosedExecutableElements(Executables.java:189)
    at org.mapstruct.ap.model.common.Type.getAllExecutables(Type.java:395)
    at org.mapstruct.ap.model.common.Type.getSetters(Type.java:464)
    at org.mapstruct.ap.model.common.Type.getPropertyWriteAccessors(Type.java:353)
    at org.mapstruct.ap.model.BeanMappingMethod$Builder.souceMethod(BeanMappingMethod.java:93)
    at org.mapstruct.ap.processor.MapperCreationProcessor.getMappingMethods(MapperCreationProcessor.java:345)
    at org.mapstruct.ap.processor.MapperCreationProcessor.getMapper(MapperCreationProcessor.java:143)
    at org.mapstruct.ap.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:114)
    at org.mapstruct.ap.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:73)
    at org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:193)
    at org.mapstruct.ap.MappingProcessor.processMapperTypeElement(MappingProcessor.java:173)
    at org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:154)

sjaakd:
Hmmm. I added logic some while ago not to allow parameterized types like this. I reckon this one should have been flagged already with a warning in the `MethodRetrievalProcessor`

sjaakd:
Ah.. `org.mapstruct.ap.test.collection.wildcard.IterableTypeVarBoundMapperOnMapper` does take the case into consideration were a parameterized type is being used. However the more simple case of using a type-var directly is not covered.

sjaakd:
[...]

PR #639, opened 2015-09-07
#631 fix for parameterized source / target types

sjaakd:
Hi guys,

I'm not sure this is the way forward, so this one needs some more tests. What I've done is work with the type bound iso the type itself (in Parameter, MappingMethod). Its a bit crude. We should perhaps (if this is the way forward) these model elements in this way in stead of putting it on the getter. 

I also would like to add some test case in which the type bounds are coming from another package (make sure imports are handled correctly).

But first.. Is this the way forward? I'll extend the solution if we agree..

Cheers,
Sjaak

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
@@ -377,6 +377,18 @@ public class MethodRetrievalProcessor implements ModelElementProcessor<Void, Lis
             return false;
         }
 
+      for ( Parameter sourceParameter : sourceParameters ) {
+            if ( sourceParameter.getType().isTypeVar() ) {
+                messager.printMessage( method, Message.RETRIEVAL_TYPE_VAR_SOURCE );
+                return false;
+            }
+        }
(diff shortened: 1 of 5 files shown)


gunnarmorling:
Hey @sjaakd, thanks for giving this a look! I think we should not support this situation - at least for now - and instead raise a meaningful error. I am still not clear what the use case would be for using type parameters as input/output types of generated mapping methods.

Sure, we could only consider the upper bounds, but what would be the use of that? If a user would somehow instantiate the mapper with more specific sub-types bound, they'd still not get the attributes of these sub-types mapped. For the return type we also would not know which type to instantiate, e.g. the bound type could be an abstract type.

So all in all, I'd just flag that situation as erroneous.

sjaakd:
Ha.. That's why I was reluctant to continue.. I'll rewrite this one leading to a meaningful error message. Thanks.

gunnarmorling:
Cool, thanks. Yes, that's what makes most sense to me atm.

sjaakd:
@gunnarmorling : ok.. done.. can I merge this one?

gunnarmorling:
Done :) Thanks, @sjaakd!

PR #641, opened 2015-09-13
Pr/638

sjaakd:
@bedla, I processed my comments. Please have a look. If you agree, I'll merge it to the master branch


processor/src/main/resources/org/mapstruct/ap/internal/model/MethodReference.ftl
```diff
@@ -20,7 +20,16 @@
 -->
 <@compress single_line=true>
     <#-- method is either internal to the mapper class, or external (via uses) declaringMapper!=null -->
-    <#if declaringMapper??><#if static><@includeModel object=declaringMapper.type/><#else>${mapperVariableName}</#if>.</#if>${name}<#if (parameters?size > 0)>( <@arguments/> )<#else>()</#if>
+    <#if declaringMapper??><#if static><@includeModel object=declaringMapper.type/><#else>${mapperVariableName}</#if>.<@params/>
+    <#elseif static && mapperToImplement??><@includeModel object=mapperToImplement/>.<@params/>
```
gunnarmorling: Does this work with a multi-leveled inheritance chain, too? I.e. the mapper interface extends another interface which declares the static/default method?

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
```diff
@@ -511,6 +522,19 @@ public boolean isStatic() {
         return executable.getModifiers().contains( Modifier.STATIC );
     }
 
+    @Override
+    public boolean isDefault() {
+        boolean isDefault = false;
+        if ( mapperToImplement != null ) {
```
gunnarmorling: When can that be null?

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
```diff
@@ -198,16 +195,10 @@ private static void addNotYetOverridden(Elements elementUtils, List<ExecutableEl
         alreadyCollected.addAll( 0, safeToAdd );
     }
 
-    private static boolean isNotStaticMethodInInterface(ExecutableElement element, TypeElement parentType) {
-        return !( parentType.getKind().isInterface() &&
+    public static boolean isInterfaceDefaultMethod(ExecutableElement element, TypeElement parentType) {
```
gunnarmorling: Are there other default methods than "interface default methods"? I.e. is the check for the interface required at all?

processor/src/main/resources/org/mapstruct/ap/internal/model/MethodReference.ftl
sjaakd: Good question.. I'll add a test case for that.. 

Which brings me to another point.. I've been struggling with the integration test (finally managed, but handy is not the term I would use).. Isn't it time to add java 8 test to the standard unit test somehow (facilitating more easy debug, and less risk)..

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
sjaakd: AFAIK the `SourceMethod` from which these are taken can also be referenced methods, right? They don't require implementation (e.g. can be hand written).

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
sjaakd: nope.. will adapt

processor/src/main/resources/org/mapstruct/ap/internal/model/MethodReference.ftl
gunnarmorling: Yes, that'd be nice. We'd just have to make sure that the generated processor JAR still works with Java 7 ((some of) the integration tests should make that sure as they also run with JDK 7 when configured properly).

sjaakd: Ok.. I'll write an additional issue / PR for that one..

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
sjaakd: But I'm wrong.. I'll adapt..

gunnarmorling: But those could be default methods, too? I guess it's mood if you process the other comment and not consider the type at all.

processor/src/main/resources/org/mapstruct/ap/internal/model/MethodReference.ftl
sjaakd: > Does this work with a multi-leveled inheritance chain, too?

No.. It doesn't.. Added a test and it fails (for the reason it cannot find the static method).  Need to look in to that more carefully..  I'll finish it this evening..

bedla:
Hi, I am back to keyboard... 

Note for MyMapper class : I have added it to simulate more natural behaviour for generated mapper - I mean interferences between interface static/default methods and "normal" methods.

Note for removal of  MethodRetrievalProcessor.findStaticMethodFromInterfaceType : I looks like that I have added more logic than needed.

I like it as a whole :-)

Ivos

sjaakd:
@bedla .. Thanks for  your review. Your solution pretty much worked. Its some fine tuning. I also got some comments from @gunnarmorling which I'm processing now.

sjaakd:
@gunnarmorling .. you want to hear something funny? The eclipse can't either handle inherited default or static interface methods :smile:

ERROR] /Users/sjaak/Development/java/NetBeansProjects/mapstruct/mapstruct1/integrationtest/target/tmp/org.mapstruct.itest.tests.Java8Test/eclipse_jdt_java_8/src/main/java/org/mapstruct/ap/test/bugs/_636/SourceTargetMapper.java:[27] 
	public interface SourceTargetMapper extends SourceTargetBaseMapper {
	                 ^^^^^^^^^^^^^^^^^^
Internal error in the mapping processor: java.lang.IllegalArgumentException: element is not valid for the containing declared type  	at org.eclipse.jdt.internal.compiler.apt.model.TypesImpl.asMemberOf(TypesImpl.java:89)  	at org.mapstruct.ap.internal.model.common.TypeFactory.getMethodType(TypeFactory.java:264)  	at org.mapstruct.ap.internal.processor.MethodRetrievalProcessor.getMethod(MethodRetrievalProcessor.java:182)  	at org.mapstruct.ap.internal.processor.MethodRetrievalProcessor.retrieveMethods(MethodRetr

sjaakd:
The other compilers do.. BTW..

sjaakd:
@gunnarmorling .. if you agree, I squash the first three commits into one. The latter one I'll let in place. I think we have to discuss the above issue (eclipse compiler problem).

gunnarmorling:
Squash your's but not @bedla's; credit where credit is due :)

sjaakd:
> Squash your's but not @bedla's; credit where credit is due :)

Of course..

sjaakd:
Thanks @bedla (Ivo) for your contribution. Great work.

gunnarmorling:
Many thanks, guys!

bedla:
With pleasure and thanks for polishing :-)
Ivos

Issue #640, opened 2015-09-12
Allow NullValueMappingStrategy to be configured for forged collection mapping methods.


samwright:
Currently if there is a mapper for converting `A -> B`, and elsewhere there is a mapping that requires `List<A> -> List<B>`, MapStruct generates a private method for it without any extra declarations required in the interface. This is a very cool feature.

However, this method is configured to return `null` if the source `List<A>` field is `null`, regardless of the `@Mapper#nullValueMappingStrategy`. I would like to be able to configure these generated mapping methods to `RETURN_DEFAULT` (e.g. an empty list) when the source `List<A>` field is `null`.

Because I need this behaviour, I have to declare the iterable mapping methods in the mapper interface and use `@IterableMapping#nullValueMappingStrategy`, e.g.

```java
@IterableMapping(nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT)
List<B> aToBList(List<A> aList);
```

This doesn't really add value to the interface, and is required for each iterable return type I need (i.e. is bad for all the reasons that the forged methods are good).

Could this be added as a configuration to `@Mapping`? E.g. `@Mapping#forgedNullValueMappingStrategy`.

Alternatively, if `@Mapping#defaulValue` could take a special string to represent an empty collection/map (e.g. an empty string) then I could get the behaviour I need.

I'm happy to help implementing this, though I'm unfamiliar with the code-base and would need some pointers.

Thanks for the awesome tool!

sjaakd:
@samwright : did you know you could also define the `NullValueMappingStrategy` on the `Mapper` and `MapperConfig`. These form a hierarchical structure, were `MapperConfig` is at the top, which can be overridden by `Mapper` which, in its turn, can be overridden by individual mapping methods.

I've to checkout how they work on forged methods thought. But you can group methods requiring a strategy together in a used mapper and put it on the `Mapper`.

gunnarmorling:
Right, `nullValueMappingStrategy()` given via `@Mapper`/`@MapperConfig` should be applied to forged methods, too. If it doesn't, I'd consider that a bug.

samwright:
@gunnarmorling I expected that to happen, but as this example shows the forged methods always follow the `RETURN_NULL` strategy, regardless of the `@Mapper#nullValueMappingStrategy`:

```java
@Mapper(nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT)
public abstract class PersonMapper
{
[...]

PR #645, opened 2015-09-14
#640 - applying Mapper or Mapperconfig NullValueMappingStrategy to fo…

sjaakd:
…rged methods

Solution is straightforward. `MapperConfiguration` contains all enums on the `Mapper` and `MapperConfig`. The `MapperConfiguration` is copied from the calling `SourceMethod` which is in the same `Mapper` as the forged method. 

Added 4 test cases to test all scenarios. Validation of `MapperConfig` should not be needed, since thats already tested in the NullValueMapping test cases.


gunnarmorling:
Rebased and applied. Many thanks, @sjaakd!

agudian:
test this please

Issue #664, opened 2015-10-20
Add blueprint as a component model option.


samwright:
Blueprint xml files can be generated from JSR330 annotations using the blueprint-maven-plugin, but it requires that each mapper class is annotated with `@Singleton` (the `@Named` annotation is still valid, but only as an additional qualifier for selecting the bean to inject). Currently the `jsr330` component model only annotates them with `@Named`.

Possible solutions:
1. Annotate mapper classes with `@Singleton` (in addition to `@Named`) when `jsr330` is used as the component model.
2. Add a new component model option `blueprint` which acts like `jsr330` does currently, but also adds the `@Singleton` annotation.
3. Allow for the mapper class implementation to be annotated arbitrarily, for example by annotating the mapper interface with something like `@AnnotateWith(MyCustomAnnotation.class)`.

Solution 1 is the simplest, and I can't think of any problems arising from it, so that's my preference. If people see this as a good idea, I'd be happy to implement it.

agudian:
Well, I like options 1 and 3, with a tendency towards the latter.

Adding `@Singleton` is super simple and apperently works for your case, but to be honest I don't have that much experience with JSR 330 in the wild, so I can't really assess the impact of this.

Allowing to add a custom annotation to the generated types is something that would be useful for other stuff aswell, not only in JSR 330, but also for CDI and Spring. But going that way would also mean that we'd need an idea how to configure that the annotation is to be added instead of those we currently add or in addition to, or to add the annotations with arguments (I could imagine that'd be on the wishlist pretty soon).

In any case - let's keep an eye on this for our release after the upcoming 1.0.

samwright:
I've implemented option 1 (see pull request) and yes, it is super simple. From my limited understanding, adding `@Singleton` is the correct way to do JSR330, though some DI frameworks assume that a class annotated with `@Named` is a singleton. As such I'd see option 1 as a simple fix for an eclectic bug.

Option 3 would then be a new feature which I have no need for (once option 1 is implemented) so I wouldn't want to spend effort implementing it without a clear user story justifying it.

If you agree with all that, I'll write some tests for the commit with a view to merge it after 1.0 and after someone with more experience of JSR330 than I confirms that this approach is sensible.

samwright:
Updated pull request

agudian:
PR applied. Thanks, Sam!

PR #665, opened 2015-10-20
664 - annotate mapper implementations with @Singleton when using jsr330.

samwright:
First attempt at implementing #664. Simply put, some dependency injection frameworks (e.g. `blueprint-maven-plugin` which generates a blueprint xml file) that use JSR330 annotations require beans to be annotated with `@Singleton`, whereas mapstruct currently only applies the `@Named` annotation. 

This commit applies the `@Singleton` annotation when using JSR330 annotations to mapper classes in addition to the `@Named` annotation, which is used by `blueprint-maven-plugin` as a qualifier for bean injection.

I couldn't find any tests which verified that the `@Named` annotation was applied, so I didn't add any tests for verifying the application of `@Singleton`. Similarly the integration test doesn't require the `@Singleton` annotation to work, so currently the content of this commit is untested (except by hand) which isn't good.

Options for testing:
1.  Add a unit test class for `Jsr330ComponentProcessor` to verify that the `@Named` and `@Singleton` annotations are returned where expected.
2. Add an integration test that uses `blueprint-maven-plugin` to generate an xml file (which requires the `@Singleton` annotation) and verify that a mapper class was successfully configured, for example by using http://aries.apache.org/modules/blueprintnoosgi.html to check that the mapper class can be retrieved as a bean.

I'm happy to implement either or all of those options, but I'll wait for someone to say whether or not this is a good approach.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/Jsr330ComponentProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/Jsr330ComponentProcessor.java
@@ -42,16 +42,16 @@ public class Jsr330ComponentProcessor extends AnnotationBasedComponentModelProce
     @Override
     protected List<Annotation> getTypeAnnotations(Mapper mapper) {
         if ( mapper.getDecorator() == null ) {
-            return Collections.singletonList( named() );
+            return Arrays.asList( singleton(), named() );
         }
         else {
-            return Collections.singletonList( namedDelegate( mapper ) );
+            return Arrays.asList( singleton(), namedDelegate( mapper ) );
         }
     }
 
     @Override
     protected List<Annotation> getDecoratorAnnotations() {
-        return Collections.singletonList( named() );
+        return Arrays.asList( singleton(), named() );
     }
 
     @Override
@@ -67,6 +67,10 @@ public class Jsr330ComponentProcessor extends AnnotationBasedComponentModelProce
     @Override
     protected boolean requiresGenerationOfDecoratorClass() {
         return true;
+    }
+
+    private Annotation singleton() {
+        return new Annotation( getTypeFactory().getType( "javax.inject.Singleton" ) );
     }
 
     private Annotation named() {
--- a/core-common/src/main/java/org/mapstruct/DecoratedWith.java
+++ b/core-common/src/main/java/org/mapstruct/DecoratedWith.java
@@ -123,9 +123,9 @@ import java.lang.annotation.Target;
  * <p>
  * JSR 330 doesn't specify qualifiers and only allows to specifically name the beans. Hence, the generated
  * implementation of the original mapper is annotated with {@code @Named("fully-qualified-name-of-generated-impl")}
- * (please note that when using a decorator, the class name of the mapper implementation ends with an underscore). To
- * inject that bean in your decorator, add the same annotation to the delegate field (e.g. by copy/pasting it from the
- * generated class):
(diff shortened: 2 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/Jsr330ComponentProcessor.java
```diff
@@ -42,7 +42,7 @@ protected String getComponentModelIdentifier() {
     @Override
     protected List<Annotation> getTypeAnnotations(Mapper mapper) {
         if ( mapper.getDecorator() == null ) {
-            return Collections.singletonList( named() );
+            return Arrays.asList( singleton(), named() );
         }
         else {
             return Collections.singletonList( namedDelegate( mapper ) );
```
agudian: You need to add the annotation here as well, as otherwise the delegate (i.e. the actual mapper implementation) for a decorator does not get the `@Singleton` annotation.

agudian:
Thanks for the pull-request! I'm basically OK with it, but as discussed in #664, we should wait with including it until after 1.0.0.Final, I think.

On the tests. I think the methods are too trivial to actually "unit" test them. Having a new test in the integration-test module on the other hand is quite expensive, at least on my home-machine and on our Jenkins instance :wink:. We already have a JSR330 processor-test in the processor module for the decorator feature. You could extend that one to check for the presence of both `@Singleton` and `@Named` - I'm just not sure at the moment if the test also includes a non-decorated mapper.

Anyway, you could check the presence either using reflection, or using the `GeneratedSource` rule - see `VersionInfoTest` for an example how that rule is used to assert on the `.java` file content of the generated classes. I'm just seeing that it can't deal with decorators, yet, though... So you'd have to teach it the different file name patterns for those decorators... :wink:

samwright:
I've amended the line you commented as you suggested.

As for tests, unit-testing does seem a bit OTT since they are annotations that are configuring how beans are integrated, so an integration test would be more telling. The integration tests as they stand require these annotations, so they aren't really applicable. 

Ideally we'd add an integration test using the blueprint-noosgi module to see if the mappers are being injected properly, but as you say that will add to the time required for testing. Also, such an addition is probably beyond the scope of this ticket, and would probably prompt people to want similar integration tests for the other DI systems.

So, with that said, would you consider this change trivial-enough to not require testing, beyond the manual testing I've done?

gunnarmorling:
@samwright Thanks for the update. Going to look into it asap.

agudian:
@samwright, thanks for rebasing! The change itself looks good. Not testing it at all still feels weired, though.

I suggest the following: extend `Jsr330DecoratorTest` to check that the generated sources are annotated with the `@Singleton` annotation. You could perform the check using the `GeneratedSource` rule, which could be extended with a method `forDecoratedMapper(Class<*> mapper)`, which would return the `Impl_.java` file. Then just check if the annotation is present there and that should be enough, I guess. What do you think?

agudian:

gunnarmorling:
@agudian Where are we with this one? I'm unsure about what'd have to be done to drive it home.

agudian:
I can pick it up, add the unit test and merge it. I guess that'll be the easiest way... :)

gunnarmorling:
Cool, thanks, Andreas!

agudian:
Rebased the commit, added a test to verify the annotations and pushed it to master.

agudian:
@samwright Thanks again for taking this up! And sorry for all the delay here...

PR #666, opened 2015-10-22
Autodetect toolchain conf

### LINKED ISSUE #642: Build should ignore unavailable JDKs if toolchains.xml is present
If there is a _~/.m2/toolchains.xml_ file which contains some additional JDKs (e.g. Java 6) but not all possible values from `ProcessorSuite` (e.g. not JDK 6 and 9), the integration tests will fail with an exception complaining about the missing entries:

    Caused by: org.apache.maven.plugin.MojoFailureException: Cannot find matching toolchain definitions for the following toolchain types:
jdk [ vendor='oracle'  version='[1.6,1.7)' ]

Can we somehow make it that this does not fail the build? Personally I'd be willing to test against JDK 7 and 8 but not against 6 (hard to get in on Mac any longer) and 9 (had trouble with that because it was used by default for some applications, e.g. Eclipse).

agudian:
This fixes #642 and makes the handling of the toolchains-file a bit better. No need to activate the JDK 9 tests by using a VM-parameter anymore... :wink:

Also, I had to change one of the Joda-Tests, as it fails now on my machine (no idea if it's due to Windows 10 or JDK 1.8.0_60 or both...).

### DIFF
--- a/integrationtest/src/test/java/org/mapstruct/itest/testutil/runner/ProcessorSuite.java
+++ b/integrationtest/src/test/java/org/mapstruct/itest/testutil/runner/ProcessorSuite.java
@@ -52,12 +52,12 @@ public @interface ProcessorSuite {
         /**
          * Use an Oracle JDK 1.6 (or 1.6.x) via toolchain support to perform the processing
          */
-        ORACLE_JAVA_6( "oracle-[1.6,1.7)", "javac", "1.6" ),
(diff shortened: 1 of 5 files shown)


sjaakd:
Hi @agudian, looks good to me.. go ahead.

agudian:
Thanks, @sjaakd!

Issue #667, opened 2015-10-22
Use JDK and Eclipse compiler in AnnotationProcessorTestRunner


agudian:
The recurring problems that we have with subtle and not-so-subtle differences in the APT API implementation in JDK and in the Eclipse compiler show that our integration tests don't cover enough cases.

So I propose to just use our processor tests, which contain all sorts of edge cases and specific feature tests, and run the processor not only with the JDK compiler, but also with the Eclipse compiler. Kind of like the JUnit Parameterized runner, all test methods would be executed with the JDK variant _and_ the JDT variant.

gunnarmorling:
Done.

PR #668, opened 2015-10-22
#667 Run all processor tests with both the JDK compiler and the Eclipse compiler

agudian:
Implements the suggestion from #667.

All generated sources and compilation results are kept in a new directory structure, making manual inspection easier, and also simplifying parallel test-class execution (no need for that thread-local counter anymore).

I had to adapt some tests that only contained the first of two error messages for the same line (javac swallows all but the first, Eclipse not) and also fixed some typos in some error messages on my way.

When executed, some tests fail due to current issues we have with the Eclipse implementation.

One thing that I think could be useful is to configure if a test should use only JDK or only Eclipse, either permanently via annotation or for debugging purposes via VM-parameter (this currently requires commenting out a line in `AnnotationProcessorTestRunner`).

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
@@ -100,11 +100,11 @@ public enum Message {
     INHERITCONFIGURATION_BOTH( "Method cannot be annotated with both a @InheritConfiguration and @InheritInverseConfiguration." ),
     INHERITINVERSECONFIGURATION_DUPLICATES( "Several matching inverse methods exist: %s(). Specify a name explicitly." ),
     INHERITINVERSECONFIGURATION_INVALID_NAME( "None of the candidates %s() matches given name: \"%s\"." ),
-    INHERITINVERSECONFIGURATION_DUPLICATE_MATCHES( "Given name \"%s\" matches several candidate methods: %s()." ),
+    INHERITINVERSECONFIGURATION_DUPLICATE_MATCHES( "Given name \"%s\" matches several candidate methods: %s." ),
     INHERITINVERSECONFIGURATION_NO_NAME_MATCH( "Given name \"%s\" does not match the only candidate. Did you mean: \"%s\"." ),
     INHERITCONFIGURATION_DUPLICATES( "Several matching methods exist: %s(). Specify a name explicitly." ),
     INHERITCONFIGURATION_INVALIDNAME( "None of the candidates %s() matches given name: \"%s\"." ),
-    INHERITCONFIGURATION_DUPLICATE_MATCHES( "Given name \"%s\" matches several candidate methods: %s()." ),
+    INHERITCONFIGURATION_DUPLICATE_MATCHES( "Given name \"%s\" matches several candidate methods: %s." ),
     INHERITCONFIGURATION_NO_NAME_MATCH( "Given name \"%s\" does not match the only candidate. Did you mean: \"%s\"." ),
     INHERITCONFIGURATION_MULTIPLE_PROTOTYPE_METHODS_MATCH( "More than one configuration prototype method is applicable. Use @InheritConfiguration to select one of them explicitly: %s." ),
     INHERITCONFIGURATION_CYCLE( "Cycle detected while evaluating inherited configurations. Inheritance path: %s" );
@@ -113,12 +113,12 @@ public enum Message {
     private final String description;
     private final Diagnostic.Kind kind;
 
-    private Message(String description) {
+    Message(String description) {
         this.description = description;
         this.kind = Diagnostic.Kind.ERROR;
     }
 
-    private Message(String description, Diagnostic.Kind kind) {
+    Message(String description, Diagnostic.Kind kind) {
         this.description = description;
         this.kind = kind;
     }
--- a/processor/src/test/java/org/mapstruct/ap/test/defaultvalue/DefaultValueTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/defaultvalue/DefaultValueTest.java
@@ -18,6 +18,8 @@
  */
 package org.mapstruct.ap.test.defaultvalue;
 
+import java.text.ParseException;
+
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mapstruct.ap.testutil.IssueKey;
@@ -26,8 +28,6 @@ import org.mapstruct.ap.testutil.compilation.annotation.CompilationResult;
 import org.mapstruct.ap.testutil.compilation.annotation.Diagnostic;
 import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutcome;
 import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
-
-import java.text.ParseException;
 
 import static org.fest.assertions.Assertions.assertThat;
 
@@ -136,6 +136,10 @@ public class DefaultValueTest {
                             line = 33,
                             messageRegExp = "Constant and default value are both defined in @Mapping,"
                                     + " either define a defaultValue or a constant." ),
+                    @Diagnostic(type = ErroneousMapper.class,
+                            kind = javax.tools.Diagnostic.Kind.ERROR,
+                            line = 33,
+                            messageRegExp = "Can't map property \".*Region region\" to \".*String region\"\\. Consider")
             }
     )
     public void errorOnDefaultValueAndConstant() throws ParseException {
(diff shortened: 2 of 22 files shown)


processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -100,11 +100,11 @@
     INHERITCONFIGURATION_BOTH( "Method cannot be annotated with both a @InheritConfiguration and @InheritInverseConfiguration." ),
     INHERITINVERSECONFIGURATION_DUPLICATES( "Several matching inverse methods exist: %s(). Specify a name explicitly." ),
     INHERITINVERSECONFIGURATION_INVALID_NAME( "None of the candidates %s() matches given name: \"%s\"." ),
-    INHERITINVERSECONFIGURATION_DUPLICATE_MATCHES( "Given name \"%s\" matches several candidate methods: %s()." ),
+    INHERITINVERSECONFIGURATION_DUPLICATE_MATCHES( "Given name \"%s\" matches several candidate methods: %s." ),
     INHERITINVERSECONFIGURATION_NO_NAME_MATCH( "Given name \"%s\" does not match the only candidate. Did you mean: \"%s\"." ),
     INHERITCONFIGURATION_DUPLICATES( "Several matching methods exist: %s(). Specify a name explicitly." ),
     INHERITCONFIGURATION_INVALIDNAME( "None of the candidates %s() matches given name: \"%s\"." ),
-    INHERITCONFIGURATION_DUPLICATE_MATCHES( "Given name \"%s\" matches several candidate methods: %s()." ),
```
gunnarmorling: How odd.

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilationCache.java
```diff
@@ -0,0 +1,48 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.testutil.runner;
+
+import org.mapstruct.ap.testutil.compilation.model.CompilationOutcomeDescriptor;
+
+class CompilationCache {
```
gunnarmorling: Could you add a description? What is this needed for? Thx!

processor/src/test/java/org/mapstruct/ap/testutil/runner/EclipseCompilingStatement.java
```diff
@@ -0,0 +1,82 @@
...
+import org.codehaus.plexus.compiler.CompilerException;
+import org.codehaus.plexus.compiler.CompilerResult;
+import org.codehaus.plexus.logging.console.ConsoleLogger;
+import org.eclipse.tycho.compiler.jdt.JDTCompiler;
+import org.junit.runners.model.FrameworkMethod;
+import org.mapstruct.ap.MappingProcessor;
+import org.mapstruct.ap.testutil.compilation.model.CompilationOutcomeDescriptor;
+
+/**
+ * Statement that uses the JDK compiler to compile.
```
gunnarmorling: "JDK"="Eclipse"?

processor/src/test/java/org/mapstruct/ap/testutil/runner/GeneratedSource.java
```diff
@@ -39,11 +39,21 @@
  */
 public class GeneratedSource implements TestRule {
 
+    private static ThreadLocal<CompilingStatement> compilingStatement = new ThreadLocal<CompilingStatement>();;
```
gunnarmorling: How comes we need a thread-locale?

processor/src/test/java/org/mapstruct/ap/test/defaultvalue/DefaultValueTest.java
```diff
@@ -136,6 +136,10 @@ public void shouldHandleUpdateMethodsFromEntityToEntity() {
                             line = 33,
                             messageRegExp = "Constant and default value are both defined in @Mapping,"
                                     + " either define a defaultValue or a constant." ),
+                    @Diagnostic(type = ErroneousMapper.class,
```
gunnarmorling: Where does this additional diagnostic come from?

processor/src/test/java/org/mapstruct/ap/test/template/InheritConfigurationTest.java
```diff
@@ -53,7 +53,7 @@ public void shouldInheritConfigurationSingleCandidates() {
         assertThat( createdTarget ).isNotNull();
         assertThat( createdTarget.getStringPropY() ).isEqualTo( "1" );
         assertThat( createdTarget.getIntegerPropY() ).isEqualTo( 2 );
-        assertThat( createdTarget.getNestedResultProp() ).isEqualTo( "nested");
+        assertThat( createdTarget.getNestedResultProp() ).isEqualTo( "nested" );
```
gunnarmorling: Hum, I thought we had a CS rule which should have discovered that long time ago. But apparently we don't...

gunnarmorling:
> When executed, some tests fail due to current issues we have with the Eclipse implementation.

So should we add some means of ignoring specific tests on Eclipse only (e.g. `SkipByCompiler(ECLIPSE)`) in order to move forward with this change and address the reveiled issues one by one?

gunnarmorling:
Looks good to me overall, @agudian. See my remarks for some details. I think it's very helpful, thanks a lot! Even more value now to extract the AP testing infra into its own project after the 1.0 release.

processor/src/test/java/org/mapstruct/ap/testutil/runner/EclipseCompilingStatement.java
agudian: Yupp, Copy/paste... :)

processor/src/test/java/org/mapstruct/ap/testutil/runner/GeneratedSource.java
agudian: Before the change, there was a lot of ThreadLocal stuff in the CompilingStatement itself. I've removed all of that over there, but this part remains, as a `TestRule` is not by API limited to be used as a `@Rule`, but can also be used as `@ClassRule`. If it was used that way and if the test methods would be executed in parallel, then we could run into troubles here.
As I'm writing this and checking in the code, I just notice that the usage of the `CompilationCache` is not thread-safe for parallel method execution anymore... uhm...
I guess I'll need to think about that. I guess it makes sense to just ditch the support of parellel method execution altogether (would kill the cache effect anyway), and just support the parallel class execution.

processor/src/test/java/org/mapstruct/ap/test/defaultvalue/DefaultValueTest.java
agudian: The JDK compiler swallows all but the first diagnostic for the same type+line+kind. Eclipse hands out all of them. And we report as much as possible, which I think is good. There are other changes in this PR that also relate to this (e.g. the sorting of the diagnostics).

processor/src/test/java/org/mapstruct/ap/test/template/InheritConfigurationTest.java
agudian: Funny story. There's a whole bunch of stuff that the checkstyle version we currently use misses. I found that with the updated version that is used in the Eclipse checkstyle plugin (for the same rule set). I'll send a PR for updating checkstyle in the build and for fixing the thereby found issues as well. I have it prepared locally already.

agudian:
> So should we add some means of ignoring specific tests on Eclipse only (e.g. SkipByCompiler(ECLIPSE)) in order to move forward with this change and address the reveiled issues one by one?

Yes, I thought about that - but in the end they all fail for the very same reason currently. But I can add that feature anyway.

agudian:
@gunnarmorling, @sjaakd, I've updated the javadocs a bit (and clarified that test classes can run in parallel, but methods within one test not). I think I also fixed the problem with running only single test methods in your IDEs. I was able to reproduce it with a surefire-execution.

If it's OK with you, I would leave out the ignore-eclipse-option for now and instead rebase the asMemberOf PR accordingly.

agudian:
Ok, I'll revert the version back to 2.18.1, but you'll still not be able to run a single method then. If Netbeans would hand over a pattern that is documented and supported by surefire, then it would be fine. But we'll fix this in Surefire: https://issues.apache.org/jira/browse/SUREFIRE-1191 - so you'll have to wait until 2.19.1 :wink:. Is that OK for you?

agudian:
Oh... or you modify your mvn script to replace the test package name with the path (or just drop the package declaration)... Just to make it work in the meantime.

sjaakd:
> Is that OK for you

I'm fine with that.. 



> Oh... or you modify your mvn script to replace the test package name with the path (or just drop the package declaration)... Just to make it work in the meantime.

Need to dig into NB where this can be done..

agudian:
> Need to dig into NB where this can be done..

I would guess it just calls your systems `mvn` script - and that's what I would edit...

gunnarmorling:
Going to close this one as it's part of that other PR of yours, @agudian. Thanks a lot, testing has greatly improved with that!

PR #680, opened 2015-11-04
Agudian processor test runner with eclipse

sjaakd:
This solution should not be merged.. Its just a very press attempt to have some discussion on..


gunnarmorling:
Hey @sjaakd, I think we can close this one now with @agudian's alternative proposal?

sjaakd:
Yes, please.. It was there to reflect.. I like @agudian 's solution way better :smile:

Issue #644, opened 2015-09-14
Find alternative way to handle asMemberOf when using Eclipse 


sjaakd:
The following error occurs in integration test (Java8) when `fooFromId and `barFromId ` in:
```java
package org.mapstruct.ap.test.bugs._636;

import org.mapstruct.Mapper;
import org.mapstruct.Mapping;
import org.mapstruct.Mappings;
import org.mapstruct.factory.Mappers;

@Mapper
public interface SourceTargetMapper extends SourceTargetBaseMapper {
    SourceTargetMapper INSTANCE = Mappers.getMapper( SourceTargetMapper.class );

    @Mappings({
        @Mapping(source = "idFoo", target = "foo"),
        @Mapping(source = "idBar", target = "bar")
    })
    Target mapSourceToTarget(Source source);

    default Foo fooFromId(long id) {
        return new Foo(id);
    }

    static Bar barFromId(String id) {
        return new Bar(id);
    }

}
```
are brought to the (intended) base class:

```java
package org.mapstruct.ap.test.bugs._636;

import org.mapstruct.Mapper;

@Mapper
public interface SourceTargetBaseMapper {

    // TODO.. move default and static interface method here when problem in eclipse processor is fixed.

}
```

```
ERROR] /Users/sjaak/Development/java/NetBeansProjects/mapstruct/mapstruct1/integrationtest/target/tmp/org.mapstruct.itest.tests.Java8Test/eclipse_jdt_java_8/src/main/java/org/mapstruct/ap/test/bugs/_636/SourceTargetMapper.java:[27] 
public interface SourceTargetMapper extends SourceTargetBaseMapper {
^^^^^^^^^^^^^^^^^^
Internal error in the mapping processor: java.lang.IllegalArgumentException: element is not valid for the containing declared type at org.eclipse.jdt.internal.compiler.apt.model.TypesImpl.asMemberOf(TypesImpl.java:89) at org.mapstruct.ap.internal.model.common.TypeFactory.getMethodType(TypeFactory.java:264) at org.mapstruct.ap.internal.processor.MethodRetrievalProcessor.getMethod(MethodRetrievalProcessor.java:182) at org.mapstruct.ap.internal.processor.MethodRetrievalProcessor.retrieveMethods(MethodRetr
```

Ergo, it seems that the eclipse compiler plugin cannot handle the Java8 constructs `default` and `static method` in an `interface` base class.

gunnarmorling:
[...]

PR #683, opened 2015-11-06
#644 Workaround for Eclipse as member of

agudian:
This is still work-in-progress, but all tests are green and it works when used within a real, live Eclipse.

Several things should be done: 
* do we want to go over the Reflection stuff again and perhaps avoid some CPU cycles there? 
* is this funky try-catch-try-again-with-workaround too weird? Should we at least remember that we used the workaround once and then always use it for the same processor-round?
* it would be great if you try to break it, so that at least we know where our holes still are (and perhaps in preparation of a possibly more complete fix in Eclipse)

That's it for me today... ;)


gunnarmorling:
Cool, thanks a lot, @agudian! I will check it out later this weekend. One quick thing though: Could you create a separate issue for the JodaTime update? That way we'll have it explicitly in the change log. Thanks!

agudian:
> Could you create a separate issue for the JodaTime update? That way we'll have it explicitly in the change log.

Of course. I added that commit to make my tests work again, as I hadn't rebased at that time, yet. I'll also revert the changes to that Joda-test that I committed some days ago, as now I know that it's the issue with Java 1.8.0_60 together with Joda < 2.9... I'll clean that up.

agudian:
> The problem here is the package scoped constructor of TypeMirrorImpl. A bit dirty perhaps, but we could place our own util within a package org.eclipse.jdt.internal.compiler.apt.model.

That one won't work, as it would trigger a SecurityException... Classes in the same package need to be loaded by the same classloader and from the same jar for the package-visibility access to be granted.

agudian:
> Just a matter of preference, but I prefer an explicit comparator :smile: iso in-line..

Sure, I still want to add a test somewhere to check if that actually works before cleaning up the code... :+1:

sjaakd:
> That one won't work, as it would trigger a SecurityException... Classes in the same package need to be loaded by the same classloader and from the same jar for the package-visibility access to be granted.

Hmm. I always wondered about that..  Makes sense of course.. Another strategy would be calling the `asElement` and getting the binding from the element (that seems strangely enough public). I assume its the same binding.

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -261,7 +261,7 @@ else if ( type.isPrimitive() ) {
      */
     public ExecutableType getMethodType(TypeElement includingType, ExecutableElement method) {
         DeclaredType asType = (DeclaredType) replaceTypeElementIfNecessary( elementUtils, includingType ).asType();
-        TypeMirror asMemberOf = typeUtils.asMemberOf( asType, method );
+        TypeMirror asMemberOf = SpecificCompilerWorkarounds.asMemberOf( asType, method, typeUtils, elementUtils );
```
gunnarmorling: Should we add a CS rule to make sure that nowhere in our code base the original `isMemberOf()` is invoked? IIRC, they provide a way to fail the build if specific methods are invoked.

processor/src/main/java/org/mapstruct/ap/internal/util/SpecificCompilerWorkarounds.java
```diff
@@ -115,4 +135,161 @@ public static TypeElement replaceTypeElementIfNecessary(Elements elementUtils, T
...
+
+            if ( null != eclipseWorkaround ) {
+                return eclipseWorkaround;
+            }
+
+            throw e;
+        }
+    }
+
+    private static final class EclipseEnvironment {
```
gunnarmorling: Are you sure this class *only* is loaded in the Eclipse case? All of bit being in one class makes me a bit nervous. I know it'll be separate class files, but still...

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilationCache.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.mapstruct.ap.testutil.compilation.model.CompilationOutcomeDescriptor;
+
+/**
+ * Remembers the last {@link CompilationRequest} together with the outcome of that request. Consecutive tests that
+ * request the same compilation can reuse the results from this holder.
+ *
+ * @author Andreas Gudian
+ */
+class CompilationCache {
```
gunnarmorling: Is the cache still needed? I remember you saying it'd be obsolete now?

processor/src/main/java/org/mapstruct/ap/internal/util/SpecificCompilerWorkarounds.java
```diff
@@ -115,4 +135,161 @@ public static TypeElement replaceTypeElementIfNecessary(Elements elementUtils, T
...
+
+                    // if nothing was found, traverse the interfaces and collect all candidate methods that match
+                    List<MethodBinding> candidatesFromInterfaces = new ArrayList<MethodBinding>();
+                    collectFromInterfaces(
+                        methodBinding,
+                        referenceBinding,
+                        new HashSet<ReferenceBinding>(),
+                        candidatesFromInterfaces );
+
+                    java.util.Collections.sort( candidatesFromInterfaces, new Comparator<MethodBinding>() {
```
gunnarmorling: Why is that sorting part needed here?

processor/src/main/java/org/mapstruct/ap/internal/util/SpecificCompilerWorkarounds.java
```diff
@@ -115,4 +135,161 @@ public static TypeElement replaceTypeElementIfNecessary(Elements elementUtils, T
...
+            }
+
+            throw e;
+        }
+    }
+
+    private static final class EclipseEnvironment {
+        public static TypeMirror asMemberOf(DeclaredType containing, Element element, Types typeUtils,
+                                            Elements elementUtils) {
+            if ( typeUtils instanceof TypesImpl && containing instanceof DeclaredTypeImpl
```
gunnarmorling: Just a tiny style/test thing, but I thought I'd bring it up: I'd prefer to invert the condition and return null right here. That way all the remainder of the method can be indented one level less which I think is good for readability. Your call to change it or not :)

processor/src/main/java/org/mapstruct/ap/internal/util/SpecificCompilerWorkarounds.java
agudian: Well, it is only loaded if an IllegalArgumentException is thrown. But nevertheless, I'll improve the error handling anyway, so that in any case we get to know the `element` and `containingElement` that the method was called with. Plus the stacktrace, of course.

processor/src/main/java/org/mapstruct/ap/internal/util/SpecificCompilerWorkarounds.java
gunnarmorling: What I mean is, do we really need to collect all matching methods or couldn't we return after the first one was found? The compiler itself should complain if the same type parameter is bound several times in incompatible ways. But maybe I am missing something?

processor/src/main/java/org/mapstruct/ap/internal/util/SpecificCompilerWorkarounds.java
agudian: Oh and the classloading is tested within the integration-tests. Before I extracted the Eclipse-specific stuff, the integration-tests that only use (and see) the JDK compiler failed with a NoClassDefFoundError.

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
agudian: Hm, yes, I guess we could. But I also like Sjaaks idea to wrap the `Types` and `Elements` instances with delegators, then we can't miss anything anymore.

gunnarmorling:
@agudian Looks good to me, added some comments on details. I'll take another look later on at the previously failing tests to fully grasp everything.

processor/src/main/java/org/mapstruct/ap/internal/util/SpecificCompilerWorkarounds.java
agudian: In case you find the same method provided in two implemented interfaces. That's what I mumbled about in the Gitter-chat. I still need to create a test case to check if that implementation here even works.

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilationCache.java
agudian: The cache is still needed to avoid recompilation for consecutive test methods. That's a major speedup for our tests, and test classes where each method brings their own set of classes to compile are noticably slower.

processor/src/main/java/org/mapstruct/ap/internal/util/SpecificCompilerWorkarounds.java
agudian: > The compiler itself should complain if the same type parameter is bound several times in incompatible ways. And as we are traversing the hierarchy from bottom to top, we should first encounter the most-specific binding in an inheritance line if that is the concern.

The compiler complains if there are incompatible declarations within one type or it's entire hiearchy. And for one type, there is only one matching method. But if the type itself doesn't contain the method and the superclasses don't contain the method either (checking the superclass hierarchy is enough, no need to check the interfaces there), then there can still be different declarations in the interfaces directly implemented by the type (or their super-interfaces). In that case, the most specialized variant wins. For each interface type, the most specialized variant is automatically the one that was found first - you just need to compare the variants that are found in siblings of the hierarchy. Hence the different termination conditions for the recursion in `collectFromInterfaces`. The JLS probably has some better and more complete wording of this, but you can try it out rather quickly with the example classes from the chat. That's what I did... :wink:

processor/src/main/java/org/mapstruct/ap/internal/util/SpecificCompilerWorkarounds.java
agudian: Yes, sure. There's some other code-style stuff as well still lying around that I'll try to tackle after I added at least another test.

agudian:
> @agudian Looks good to me, added some comments on details. I'll take another look later on at the previously failing tests to fully grasp everything.

Great, thanks for all your comments, @gunnarmorling and @sjaakd!

processor/src/main/java/org/mapstruct/ap/internal/util/SpecificCompilerWorkarounds.java
agudian: Ok, I just tried again... and indeed you can't have multiple methods with the same parameter-erasure within one non-abstract class (when searched over the full hierarchy). I'm getting the name-clash compile error now that you probably expected. It is still possible for the return-type, though. So I guess the handling of the parameter types in the comparator could be removed...?

agudian:
Sorry for the many commits, but I did a rebase to get my branches in order... :wink:.

I didn't yet get to the beautifying of the code and the refactoring with the wrapped processing env, but at least I got to adding a proper test case - and I stumbled accross two other already existing bugs. I fixed them in two separate commits and I'll create some issues for them tomorrow for our book-keeping.

sjaakd:
I checked 0d5dcb0 to 7e4c1da.. Looks good.. .

sjaakd:
One more which sprang to my mind.. Should we handle (bounded) wildcards in relation to generic type variables as source / target bean? Do we expect any issues in there?

so:
```java
public interface TestBase<T>{

     public void setTest1(List<? extends T> list);

     public void setTest2(List<? super T> list);

}

public interface Test extends TestBase<Number> {

}

public class TestImpl implements Test{

    @Override
    public void setTest1(List<? extends Number> list) {
    }

    @Override
    public void setTest2(List<? super Number> list) {
    }

}
```

agudian:
> One more which sprang to my mind.. Should we handle (bounded) wildcards in relation to generic type variables as source / target bean? Do we expect any issues in there?

No idea, you can try it... but I'd keep my hands off trying to fix that for 1.0... :wink:

I've just added a commit with the refactoring that we talked about. Looks quite good, I think. 

Funny thing is, that I had to remove calls to `erasure` in `SpecificCompilerWorkarounds.isAssignable`, or otherwise any Iterable/Map related tests failed (there were parts in the code that called the original `Types.isAssignable` directly).

processor/pom.xml
```diff
@@ -57,7 +57,11 @@
             <artifactId>mapstruct</artifactId>
             <scope>provided</scope>
         </dependency>
-
+        <dependency>
```
gunnarmorling: This should better be marked `optional`. `provided` suggests it's needed at runtime but provided by the deployer/container.

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/TypesDecorator.java
```diff
@@ -0,0 +1,146 @@
...
+import javax.lang.model.type.WildcardType;
+import javax.lang.model.util.Types;
+
+/**
+ * Replaces the usage of {@link Types} within MapStruct by delegating to the original implementation or to our specific
+ * workarounds if necessary.
+ *
+ * @author Andreas Gudian
+ */
+public class TypesDecorator implements Types {
```
gunnarmorling: Nice idea with the decorator :)

processor/src/test/java/org/mapstruct/ap/test/abstractclass/generics/GenericsHierarchyTest.java
```diff
@@ -0,0 +1,99 @@
...
+    Target.class
+})
+public class GenericsHierarchyTest {
+
+    @Test
+    public void determinesItemCSourceGetter() {
+        AbstractClassExposingItemC source = new SourceWithItemC();
+
+        source.setItem( new ItemC() );
+        // make sure the jdk compiler resolves the same as we expect
```
gunnarmorling: I don't quite understand this comment. Could you elaborate a bit?

processor/src/test/java/org/mapstruct/ap/test/abstractclass/generics/GenericsHierarchyTest.java
```diff
@@ -0,0 +1,99 @@
...
+        source.getItem().setTouchedC( false );
+
+        Target target = GenericsHierarchyMapper.INSTANCE.toTarget( source );
+
+        assertThat( target.getItemC().isTouchedC() ).isTrue();
+        assertThat( target.getItemC().isTouchedB() ).isFalse();
+    }
+
+    @Test
```
gunnarmorling: It's a bit tough to make sense of these tests. Could you give the methods more meaningful names, such as `typeParameterIsResolvedToMostSpecificBound()` or something like that?

gunnarmorling:
@agudian added some more comments. Mainly I am having a very hard time to understand that new test :) I'd still prefer if we used "actual entities" instead of A, B, C (didn't we agree on that once?), I think it helps with keeping things digestable. Ah, could you tell me which tests of the previously existing ones were failing on ejc without your fix?

Thanks a lot for hacking on this!

processor/pom.xml
agudian: Totally... Chaning it to `<scope>provided</scope><optional>true</optional>`

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/TypesDecorator.java
agudian: I think so, too. praise => @sjaakd

processor/src/test/java/org/mapstruct/ap/test/abstractclass/generics/GenericsHierarchyTest.java
agudian: In certain cases the Eclipse compiler tends to let you do stuff that javac just resolves differently and vice versa - I wanted to make sure that what I write in Eclipse (including the assumptions that I make with the test class setup) is still valid when compiled with javac (in this case on the command line with maven). So if the statement `source.getItem().setTouchedC( false );` wouldn't compile in the maven build, then we'd know that the javac implementation of `asMemberOf` would not resolve `getItem()` in the scope of `AbstractClassExposingItemC` to return `ItemC`, but something less specific - and the test wouldn't fail because we did something wrong in our implementation, but because the compiler just sees it that way... :wink:

At least that was my train of thoughts at this point...

processor/src/test/java/org/mapstruct/ap/test/abstractclass/generics/GenericsHierarchyTest.java
agudian: Yes. Something like `typeParameterIsResolvedToItemC()` versus `typeParameterIsResolvedToItemB()`, as I still want to distinguish between the hierarchy level `ItemC` and `ItemB`.

agudian:
> Mainly I am having a very hard time to understand that new test :) I'd still prefer if we used "actual entities" instead of A, B, C (didn't we agree on that once?), I think it helps with keeping things digestable.

Yes, we did agree on that. Do you guys have any suggestions for naming the class hierarchy (`ItemC` extends `ItemB` extends `ItemA`) and the corresponding "similar" interfaces and abstract classes... Would it be easier to understand if I used `Car`, `Sedan`, `RedSedan`?

> Ah, could you tell me which tests of the previously existing ones were failing on ejc without your fix?

Not sure anymore, but I think it was `GenericsTest`, `InheritedMappingMethodTest` and `AbstractClassTest`? You could check-out the commit just before the fix to make sure... :wink:

agudian:
> My only worry is that the Types interface changes through java versions.. The warning: Compatibility Note: Methods may be added to this interface in future releases of the platform.) worries me a bit. Should we use java.lang.reflect.Proxy, disadvantage: reflection, advantage: future proof & compacter.

At least we should for now make sure that one of us tests with JDK 9 before we release... I didn't install a current build for a while now, will do it later. But at least since JDK 6, there were no additions (my integration tests run with JDK 6, 7, 8 at the moment).

> At the other hand: should we do Elements in a similar fashion?

I'd say yes, but not before we acutally need it... :)

agudian:
Speaking of the integration tests... `SimpleTest.oracle_java_6` fails because of an assertion-failure in isSubtype - so I guess the thing with the erasure is needed somewhere after all... Damn... I don't wanna... :sob:

sjaakd:
> Speaking of the integration tests... SimpleTest.oracle_java_6 fails because of an assertion-failure in isSubtype - so I guess the thing with the erasure is needed somewhere after all... Damn... I don't wanna...

Yep.. I vaguely remember this being a java6 problem.. Can we detect the version somehow (to make it conditional)? I think it should be nearby..

sjaakd:
> I'd say yes, but not before we acutally need it... :)

What about.. :
```java
    public static TypeElement replaceTypeElementIfNecessary(Elements elementUtils, TypeElement element) {
        if ( element.getEnclosedElements().isEmpty() ) {
            TypeElement resolvedByName = elementUtils.getTypeElement( element.getQualifiedName() );
            if ( resolvedByName != null && !resolvedByName.getEnclosedElements().isEmpty() ) {
                return resolvedByName;
            }
        }
        return element;
    }
```

agudian:
> Yep.. I vaguely remember this being a java6 problem.. Can we detect the version somehow (to make it conditional)? I think it should be nearby..

If we do that, I bet that the other tests that I fixed wouldn't work under Java 6 anymore (for which we might not have test coverage right now)... Anyone interested in picking this up? I can't do anything anymore today... too exhausted...

agudian:
> What about.. :

Awh... right. Let's get back to that another time after 1.0?

sjaakd:
> Awh... right. Let's get back to that another time after 1.0?

No Problem.

sjaakd:
> Anyone interested in picking this up? I can't do anything anymore today... too exhausted...

I'll have a quick look.. But not going to make it late (I'm tired as well, and started on my wednesday :beer: already)

sjaakd:
Stopping as well.. My work is on https://github.com/sjaakd/mapstruct/tree/agudian-eclipse-as-member-of

gunnarmorling:
> One more which sprang to my mind.. Should we handle (bounded) wildcards in relation to generic type variables as source / target bean? Do we expect any issues in there?

@sjaakd, could you open an issue (with an example) for post 1.1 for adding a test so we know that works? Thx!

gunnarmorling:
> Yes, we did agree on that. Do you guys have any suggestions for naming the class hierarchy

For inheritance hierarchies I find actual hierarchies such as `Mammal` -> `Elephant` -> `IndianElephant` good as they are instantly memorizable.

gunnarmorling:
> Not sure anymore, but I think it was ...

Thanks! I suppose I also should see it with your other PR.

gunnarmorling:
>> My only worry is that the Types interface changes through java versions

> At least we should for now make sure that one of us tests with JDK 9 before we release

+1 That'd be good enough for now. Whenever JDK 10 comes out, we can people expect to upgrade to a new MapStruct version :)

gunnarmorling:
Regarding the JDK 6 issue, in general I am fine with giving up JDK 6 support, it'd be a bit unfortunate though to do it only in the 1.0.Final.

But if we don't find a suitable fix, that's the way it is then. I wouldn't put too much effort into this, given that the (free) JDK 6 is end-of-life for 2.5 years now. We then should just document the cases which cause issues on JDK 6.

agudian:
> For inheritance hierarchies I find actual hierarchies such as Mammal -> Elephant -> IndianElephant good as they are instantly memorizable.

Any fitting suggestions for the interfaces/classes? :-)

agudian:
The latest JDK 9 builds works fine with our integration tests (which need a lot of work, btw - they're in bad shape regarding style and coverage).

sjaakd:
> Regarding the JDK 6 issue, in general I am fine with giving up JDK 6 support, it'd be a bit unfortunate though to do it only in the 1.0.Fina

I think you underestimate the number of users working with an old JDK 6. For myself, for instance, it will take time to migrate to java7 (updating a lot of application servers, combining it with new release EAP). 

So I want to get it to work on java6 as well before we push 1.0 out .

gunnarmorling:
> Any fitting suggestions for the interfaces/classes? :-)

I'd go for `Mammal` et al. for the classes, and something like `Nameable<T>`, `Identifiable<T>` for interfaces. And then `class LongIdentified implements Identifiable<Long> { ... }`, just as needed. Something like that :)

gunnarmorling:
> I think you underestimate the number of users working with an old JDK 6

I hope they are all on support contracts to get the extended-life versions :) I understand your point, but then I think we also must move forward, so e.g. I'd not like to delay the release for another month due to this issue.

To give another data point, most of our Hibernate libs require Java 7 nowadays, and I have not noticed any pushback from the community about this. Also it's not that MapStruct would not work at all on JDK 6, some specific cases would be affected (which we'd have to describe somehow).

But I agree, if we find a solution, that'd be even better of course. So if you manage to make it work, that'd be awesome :)

sjaakd:
> But I agree, if we find a solution, that'd be even better of course. So if you manage to make it work, that'd be awesome :)

I'm going to take another shot at this tonight.. Otherwise I cannot use the 1.0 version of MapStruct in my projects and that would be a pitty.

gunnarmorling:
Hum, yes, that'd be not cool indeed. I hope you'll be able to find a solution. Let me know in case you want to discuss any ideas or so :)

agudian:
> I'd go for Mammal et al. for the classes, and something like Nameable<T>, Identifiable<T> for interfaces. And then class LongIdentified implements Identifiable<Long> { ... }, just as needed. Something like that :)

Well, the `T` would have to be one of the mammals in the different variants of interfaces / abstract classes... Would it be OK to concentrate on that in the aftermath of this PR?

agudian:
> Also it's not that MapStruct would not work at all on JDK 6, some specific cases would be affected (which we'd have to describe somehow).

It's more like it wouldn't work in most cases, i.e. - whenever you have collections/maps in your beans. And only because of the refactoring with the `TypesDecorator` and not because of the asMemberOf-workaround.
I'll take another look tonight as well.

But I generally agree to phase out the JDK 6 support at some point... after 1.0... :wink:

agudian:
OK, I think I found a fix... you'll love this... :-/

agudian:
Commit added, please take a look. @sjaakd, I started with your approach, but didn't get it to work either... :confused:

sjaakd:
@agudian , I started to look at your solution.. Bit first a small remark on my solution:

```java
    private static final String V1_6 = "1.6";
    private static final String COMPILER = "javac";

    private final Types delegate;
    private final ProcessingEnvironment processingEnv;
    private final boolean isOracleJdkVersion6;

    public TypesDecorator(ProcessingEnvironment processingEnv, VersionInformation versionInformation) {
        this.delegate = processingEnv.getTypeUtils();
        this.processingEnv = processingEnv;
        this.isOracleJdkVersion6 = COMPILER.equals( versionInformation.getCompiler() ) &&
            versionInformation.getRuntimeVersion().startsWith( V1_6 );
    }
```
establishes correctly the right compiler version (my initial solution didn't). Useful in other scenarios perhaps, but not here and now :smile:

I started re-reading the generics stuff. Did you see where I used the getTypeBound() initially?

```
Usages of Type.getTypeBound [9 occurrences]
MapStruct Processor
MapMappingMethod.java
  117:  Type keySourceType = sourceTypeParams.get( 0 ).getTypeBound();
  118:  Type keyTargetType = resultTypeParams.get( 0 ).getTypeBound();
  145:  Type valueSourceType = sourceTypeParams.get( 1 ).getTypeBound();
  146:  Type valueTargetType = resultTypeParams.get( 1 ).getTypeBound();
IterableMappingMethod.java
  100:  : first( sourceParameterType.getTypeParameters() ).getTypeBound();
  102:  : first( resultType.getTypeParameters() ).getTypeBound();
  256:  return sourceParameterType.getTypeParameters().get( 0 ).getTypeBound();
Type.java
  713:  return getTypeBound().getName();
TypeFactory.java
  330:  typeParameters.add( getType( typeParameter ).getTypeBound() );
```

Apart from your solution, I think that after 1.0 we should consider to get rid of this getTypeBound. I'm not yet sure how wild-cards are handled, but type variables might as well be resolved by the asMemberOf to a concrete type. As to handle wild-cards: if I understand correctly from the excellent explanation [here] (http://www.angelikalanger.com/GenericsFAQ/FAQSections/TechnicalDetails.html#FAQ503) wildcards can only be replaced by themselves, so possibly the solution is wrong.

agudian:
> establishes correctly the right compiler version (my initial solution didn't). Useful in other scenarios perhaps, but not here and now :smile:

Yepp, that was the first thing that I tried as well... :)

> Apart from your solution, I think that after 1.0 we should consider to get rid of this getTypeBound. I'm not yet sure how wild-cards are handled, but type variables might as well be resolved by the asMemberOf to a concrete type. As to handle wild-cards: if I understand correctly from the excellent explanation here wildcards can only be replaced by themselves, so possibly the solution is wrong.

Yeah, I'm also quite sure that we're making it ourselves too easy here regarding the bounds. But if JDK (8 or other versions as well) returns in `asMemberOf` the constructor `new ArrayList(Collection<? extends ? super Type>)`, then they seem to do something wrong as well... It's really all quite confusing...

sjaakd:
@agudian good observations & nice work :smile: . I've some questions (wondering about why you moved the `getTypeBound`). But then again, I think we should get rid of it anyway.  I personally think that the concept of wild-cards is sometimes a bit hard to grasp now and then in Java, but after 1.0 we need to reconsider some of the stuff anyway and do it the correct way. 

Also, I now put limits on parameterization of methods because I was not aware of this useful `asMemberOf`. Now I think we can resolve more scenarios (even if you parameterize the mapper interface / abstract class itself). And of course, with a fully functional death star (asMemberOf).. 

I'm happy with the solution. So, go ahead as far as I'm concerned by getting this stuff in. @gunnarmorling : WDYT?

sjaakd:
> returns in asMemberOf the constructor new ArrayList(Collection<? extends ? super Type>), then they seem to do something wrong as well... It's really all quite confusing...

Ah.. But you try to attach meaning to: `Collection<? extends ? super Type>`. I don't think that that's something you should do. Its a capture. On the risk that I keep repeating myself, this [is](http://www.angelikalanger.com/GenericsFAQ/FAQSections/TechnicalDetails.html#FAQ501) what I found:
```
What is the capture of a wildcard?
An anonymous type variable that represents the particular unknown type that the wildcard stands for.  The compiler uses the capture internally for evaluation of expressions 
``` 
and this is followed by an example:
```
 Example (pseudo code showing the wildcard capture): 

    private static <T_?001> void reverse_?001(List<T_?001> list) {
      // ...  the implementation ...
    }
    public static void reverse(List<?> list) {
      reverse_?001(list);
    } 
```
In the context of the above: `Collection<? extends ? super Type>` might as well have been `T_?001` or `banana` :smile:

agudian:
Right. And when comparing banana with seagull in isAssignable, the JDK says 'sure, why not!'...

Btw, I think our death-star feels more like the one in Episode VI than the one in IV... So we need to be careful in any case.

agudian:
> There are more calls.. Aren't they troublesome? Oh.. wait a minute.. this is the only place were the `SpecificCompilerWorkArounds#isAssignable` was called, right? Brilliant :+1:

Hehe, I didn't even notice, but you're right. I was just tempted by Gunnars comments that he left in that place, suggesting that it's all just because of the failing `asMemberOf`...

I'll do some smaller cleanups again on that part, and try out a less intrusive variant (only removing `? extends ? super` if that is found, as it currently causes the trouble).

agudian:
> I'll do some smaller cleanups again on that part, and try out a less intrusive variant (only removing `? extends ? super` if that is found, as it currently causes the trouble).

Did the cleanup, but I left the logic in place at it is now. Seems to work and I don't want to go down into that rabbit hole now. I'll create an issue to remind us to revisit that whole thing again.

sjaakd:
> Right. And when comparing banana with seagull in isAssignable, the JDK says 'sure, why not!'...

On the differences between JDK 5, 6, 7 opposed to JDK 8.. There are differences. They relate to the context in which a type is inferred. JDK  5, 6, 7 only uses the assignment context. JDK 8 additionally uses method invocation context. However, I only can find references to type variables (not wild cards).

agudian:
Closing this one in favour of the rebased version in PR #692.

Issue #691, opened 2015-11-13
Overhaul test classes and mappers in the Integration Tests


agudian:
At least one of the tests should be retrofittet to contain much more meaningful class names and property names. They should also cover more of our features, such as some of the Generics handling, Collections/Maps, Built-Ins, Enums, Decorators (is already there, but I want to mention them), Before/After-Methods, Forged Methods, and other stuff that we could imagine to behave differently in different compilers.
One candidate that MUST be replaced is the SimpleTest. There could very well be many more bean classes and mappers within that simpleTest project, it doesn't have to be just one mapper... (the integration tests have a rather high startup-penalty for each compiler, so joining more mappers within there would benefit us...).

Another idea would be to create one integration test that basically compiles/processes the processor-test classes (filtering out mappers/classes that are located in paths `*/error/*`) with each of the supported compilers. If that worked, we'd have a maximum coverage of our feature-set tested across different compiler implementations.

agudian:
Merged.

Issue #690, opened 2015-11-13
Reconsider Type boundary removal in ResolvingAttempt.hasCompatibleCopyConstructor


agudian:
One of the fixes that were required to make PR #683 work was to remove the first level of wildcard boundaries of the argument type checked for assignability to from the source collection type.

The problem showed up in the test `WildCardTest.shouldGenerateSuperBoundTargetForgedIterableMethod()` where the argument type of the `ArrayList<T>(Collection<? extends T>)` with T ~> `? super Plan` resolved to `Collection<? extends ? super Plan>`. Both JDK and Eclipse showed that result.
When checking if `List<Idea>` is assignable to `Collection<? extends ? super Plan>` using `Types.isAssignable` (where `Idea` has no relation to `Plan`), Eclipse correctly returns `false`. But JDK returns `true`, causing a compile error when the generated code actually contains that constructor call.

Anyway, the current solution of just removing the `? extends` part from the argument type feels fishy and could already be wrong for some cases. We need to dig deeper there, perhaps checking more closely if we need to do a smarter workaround or a workaround only in case of the JDK. Plus, if we start using `asMemberOf` more consistently, we might face other spots where `isAssignable` would need to be "fixed". So it could be something that needs to be moved to the `TypesDecorator` or the compiler-workarounds class.

PR #692, opened 2015-11-13
#644 Workaround for Eclipse as member of (rebased)

agudian:
Rebased variant of #683 to not remove all the commit-comments there by force-pushing.

### DIFF
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -57,7 +57,12 @@
             <artifactId>mapstruct</artifactId>
             <scope>provided</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.eclipse.tycho</groupId>
+            <artifactId>tycho-compiler-jdt</artifactId>
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
         <!-- Test -->
         <dependency>
             <groupId>junit</groupId>
@@ -84,11 +89,7 @@
             <artifactId>javax.inject</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.eclipse.tycho</groupId>
-            <artifactId>tycho-compiler-jdt</artifactId>
-            <scope>test</scope>
-        </dependency>
+
         <!-- plexus-container-default is a runtime-dependency of the tyco-compiler -->
         <dependency>
             <groupId>org.codehaus.plexus</groupId>
@@ -237,6 +238,11 @@
                         <groupId>com.jolira</groupId>
                         <artifactId>hickory</artifactId>
                         <version>${com.jolira.hickory.version}</version>
+                    </dependency>
+                    <dependency>
+                        <groupId>org.eclipse.tycho</groupId>
+                        <artifactId>tycho-compiler-jdt</artifactId>
+                        <version>${org.eclipse.tyco.compiler-jdt.version}</version>
                     </dependency>
                 </dependencies>
             </plugin>
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -84,6 +84,17 @@
             <artifactId>javax.inject</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.tycho</groupId>
+            <artifactId>tycho-compiler-jdt</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <!-- plexus-container-default is a runtime-dependency of the tyco-compiler -->
+        <dependency>
+            <groupId>org.codehaus.plexus</groupId>
+            <artifactId>plexus-container-default</artifactId>
+            <scope>test</scope>
+        </dependency>
 
         <!-- Spring -->
         <dependency>
@@ -109,7 +120,6 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-
     <build>
         <plugins>
             <plugin>
@@ -126,11 +136,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
-                <configuration>
-                    <systemPropertyVariables>
-                        <mapper.test.output.dir>compilation-tests_fork-${surefire.forkNumber}</mapper.test.output.dir>
-                    </systemPropertyVariables>
-                </configuration>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
@@ -48,18 +48,14 @@ import javax.lang.model.type.PrimitiveType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Elements;
-import javax.lang.model.util.SimpleElementVisitor6;
-import javax.lang.model.util.SimpleTypeVisitor6;
 import javax.lang.model.util.Types;
 
 import org.mapstruct.ap.internal.prism.MappingTargetPrism;
 import org.mapstruct.ap.internal.prism.TargetTypePrism;
 import org.mapstruct.ap.internal.util.AnnotationProcessingException;
 import org.mapstruct.ap.internal.util.Collections;
-import org.mapstruct.ap.internal.util.SpecificCompilerWorkarounds;
 
-import static org.mapstruct.ap.internal.util.SpecificCompilerWorkarounds.erasure;
-import static org.mapstruct.ap.internal.util.SpecificCompilerWorkarounds.replaceTypeElementIfNecessary;
+import static org.mapstruct.ap.internal.util.workarounds.SpecificCompilerWorkarounds.replaceTypeElementIfNecessary;
 
 /**
  * Factory creating {@link Type} instances.
@@ -82,10 +78,10 @@ public class TypeFactory {
         this.elementUtils = elementUtils;
         this.typeUtils = typeUtils;
 
-        iterableType = erasure( typeUtils, elementUtils.getTypeElement( Iterable.class.getCanonicalName() ).asType() );
+        iterableType = typeUtils.erasure( elementUtils.getTypeElement( Iterable.class.getCanonicalName() ).asType() );
         collectionType =
-            erasure( typeUtils, elementUtils.getTypeElement( Collection.class.getCanonicalName() ).asType() );
-        mapType = erasure( typeUtils, elementUtils.getTypeElement( Map.class.getCanonicalName() ).asType() );
+            typeUtils.erasure( elementUtils.getTypeElement( Collection.class.getCanonicalName() ).asType() );
+        mapType = typeUtils.erasure( elementUtils.getTypeElement( Map.class.getCanonicalName() ).asType() );
 
         implementationTypes.put( Iterable.class.getName(), getType( ArrayList.class ) );
         implementationTypes.put( Collection.class.getName(), getType( ArrayList.class ) );
@@ -139,9 +135,9 @@ public class TypeFactory {
 
         Type implementationType = getImplementationType( mirror );
 
-        boolean isIterableType = SpecificCompilerWorkarounds.isSubType( typeUtils, mirror, iterableType );
-        boolean isCollectionType = SpecificCompilerWorkarounds.isSubType( typeUtils, mirror, collectionType );
-        boolean isMapType = SpecificCompilerWorkarounds.isSubType( typeUtils, mirror, mapType );
+        boolean isIterableType = typeUtils.isSubtype( mirror, iterableType );
+        boolean isCollectionType = typeUtils.isSubtype( mirror, collectionType );
+        boolean isMapType = typeUtils.isSubtype( mirror, mapType );
 
         boolean isEnumType;
         boolean isInterface;
@@ -159,7 +155,7 @@ public class TypeFactory {
             isInterface = declaredType.asElement().getKind() == ElementKind.INTERFACE;
             name = declaredType.asElement().getSimpleName().toString();
 
-            typeElement = declaredType.asElement().accept( new TypeElementRetrievalVisitor(), null );
+            typeElement = (TypeElement) declaredType.asElement();
 
             if ( typeElement != null ) {
                 packageName = elementUtils.getPackageOf( typeElement ).getQualifiedName().toString();
@@ -177,8 +173,7 @@ public class TypeFactory {
 
             if ( componentTypeMirror.getKind() == TypeKind.DECLARED ) {
                 DeclaredType declaredType = (DeclaredType) componentTypeMirror;
-                TypeElement componentTypeElement =
-                        declaredType.asElement().accept( new TypeElementRetrievalVisitor(), null );
+                TypeElement componentTypeElement = (TypeElement) declaredType.asElement();
 
                 name = componentTypeElement.getSimpleName().toString() + "[]";
                 packageName = elementUtils.getPackageOf( componentTypeElement ).getQualifiedName().toString();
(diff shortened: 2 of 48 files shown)


agudian:
This supersedes PRs #668 (all commits from there are included here) and #683 (this is a rebased version of it).

processor/pom.xml
```diff
@@ -57,7 +57,12 @@
             <artifactId>mapstruct</artifactId>
             <scope>provided</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.eclipse.tycho</groupId>
```
gunnarmorling: @agudian, one thing I am wondering: Did you actually test this in a running Eclipse? I.e. can the processor "see" the JDT classes in the OSGi environment?

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -467,6 +449,45 @@ else if ( collectionOrMap.isMapType()
...
+     * <li>{@code<? extends Number>}, returns Number</li>
+     * <li>{@code<? super Number>}, returns Number</li>
+     * <li>{@code<?>}, returns Object</li>
+     * <li>{@code<T extends Number>, returns Number}</li>
+     * </ol>
+     *
+     * @param typeMirror the type to return the bound for
+     * @return the bound for this parameter
+     */
+    public TypeMirror getTypeBound(TypeMirror typeMirror) {
```
gunnarmorling: Should it be named `getUpperBound()`?

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
```diff
@@ -541,67 +539,49 @@ private boolean isPropertyMappable(Type sourceType, Type targetType) {
...
+                ExecutableType typedConstructor = (ExecutableType) typeUtils.asMemberOf(
+                    (DeclaredType) targetType.getTypeMirror(),
+                    constructor );
+
+                TypeMirror parameterType = Collections.first( typedConstructor.getParameterTypes() );
+                if ( parameterType.getKind() == TypeKind.DECLARED ) {
+                    // replace any possible type bounds in the type parameters of the parameter types, as in JDK super
+                    // type bounds in the arguments are returned from asMemberOf with "? extends ? super XX"
+                    //
+                    // It might also be enough to just remove "? super" from type parameters of
```
gunnarmorling: Should we open an isue for this?

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/EclipseAsMemberOfWorkaround.java
```diff
@@ -0,0 +1,196 @@
...
+        MethodBinding inSuperclassHiearchy = findInSuperclassHierarchy( methodBinding, referenceBinding );
+
+        if ( inSuperclassHiearchy != null ) {
+            return env.getFactory().newTypeMirror( inSuperclassHiearchy );
+        }
+
+        // if nothing was found, traverse the interfaces and collect all candidate methods that match
+        List<MethodBinding> candidatesFromInterfaces = new ArrayList<MethodBinding>();
+
+        collectFromInterfaces(
```
gunnarmorling: Should we restrict this search to the case where an interface is the sub-type? Not sure whether it makes a difference, it may implicitly be guaranteed anyways.

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
```diff
@@ -115,4 +114,34 @@ public static TypeElement replaceTypeElementIfNecessary(Elements elementUtils, T
...
+     * @see <a href="https://bugs.eclipse.org/bugs/show_bug.cgi?id=481555">Eclipse Bug 481555</a>
+     */
+    static TypeMirror asMemberOf(Types typeUtils, ProcessingEnvironment env, DeclaredType containing, Element element) {
+        TypeMirror result = null;
+        Exception lastException = null;
+        try {
+            try {
+                result = typeUtils.asMemberOf( containing, element );
+            }
+            catch ( IllegalArgumentException e ) {
```
gunnarmorling: So this assumes the IAE only ever occurs with Eclipse, right? Probably it's good enough, though I'd personally rather check whether it's actually Eclipse and otherwise pass on the IAE so to not obfuscate it.

processor/src/test/java/org/mapstruct/ap/test/abstractclass/generics/AbstractClassExposingItemB.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.abstractclass.generics;
+
+public abstract class AbstractClassExposingItemB
```
gunnarmorling: Is there an issue already for renaming the test classes into something more easily to grasp=

processor/pom.xml
agudian: Yes, I tested that as well. No funny problems there...

processor/src/test/java/org/mapstruct/ap/testutil/runner/AnnotationProcessorTestRunner.java
```diff
@@ -57,72 +59,81 @@
...
-            return new FrameworkMethod(
-                klassToUse.getDeclaredMethod( m.getName(), m.getMethod().getParameterTypes() ) );
-        }
-        catch ( NoSuchMethodException e ) {
-            throw new RuntimeException( e );
+    /**
+     * Allows to only execute selected methods, even if the executing framework is not aware of parameterized tests
+     * (e.g. some versions of IntelliJ, Netbeans, Eclipse).
+     */
+    private static final class FilterDecorator extends Filter {
```
gunnarmorling: Cool, that works now. Thanks :)

processor/pom.xml
gunnarmorling: Ok, that's cool. It's not quite clear to me why it works, but maybe annotation processor JARs are treated "special" by Eclipse's OSGi environment in some way.

gunnarmorling:
@agudian Added some more comments. It looks good to me overall, many thanks. I don't fully understand why the sorting of candidates is needed, but I trust you it's the right thing to do. I hope it will be easier to understand after the test types have been renamed :)

@sjaakd, can you confirm all works for you, too, when running on JDK 6?

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
agudian: I only moved that logic out from `Types` and the corresponding method there was `getTypeBound` as well. I think as it also extracts a lower bound (from `? super X`)... so...
But as Sjaak wrote in the other PR, our handling of type-var bounds might be a bit rough anyway and it's something we need to dig in deeper.

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
agudian: I already have... #690

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/EclipseAsMemberOfWorkaround.java
agudian: Could you be more specific? I don't understand what you mean... :)

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
agudian: Right... I think I mentioned that point in one of the comments in the other PR... but with so many changes and comments it's hard to keep track... :confused:
Should we still change that? I'd at least do that in an additional PR - either for the coming release or for the next. Your call... :wink:

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/EclipseAsMemberOfWorkaround.java
gunnarmorling: I mean, should we check wether `containing` is an interface?

processor/src/test/java/org/mapstruct/ap/test/abstractclass/generics/AbstractClassExposingItemB.java
agudian: Now there is. #693

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
gunnarmorling: Ah, sorry for missing that from an earlier comment :/

I would prefer to do it for 1.0, hiding exceptions is no good practice. But it's ok to do it separately.

processor/src/test/java/org/mapstruct/ap/test/abstractclass/generics/AbstractClassExposingItemB.java
gunnarmorling: Thanks.

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/EclipseAsMemberOfWorkaround.java
agudian: It could also be that `containing` is an abstract class that we start searching from. Only if that abstract class already contains an implementation of the method we're looking for, we can stop searching (as it can't be any _less_ specific than what's declared in the interfaces).

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
agudian: I've added a commit to delegate to the Eclipse-specific implementation only if we really are running within Eclipse.

processor/pom.xml
sjaakd: should we test all java (6,7,8) versions icm with the JDT compiler?

agudian: I'm just starting to play around with #691, which would then test 1.7 and 1.8 language level... I don't remember why we're not testing language level 1.6 with JDT in the integration-tests, though.

sjaakd:
> @sjaakd, can you confirm all works for you, too, when running on JDK 6?

I need to take a look tomorrow.. I try to run unit test on our unit test cases. I commented java 8 stuff (like java.time) out. After cutting some stuff out I got 8 failures and 3 errors.

Reasons vary.

Failures:
1. org.mapstruct.ap.test.abstractclass.generics.GenericsHierarchyTest does not work (1 error, 3 failures).
2. org.mapstruct.ap.test.defaultvalue.DefaultValueTest, seems to be order and amount of error messages 
3.  org.mapstruct.ap.test.dependency.OrderingTest, seems to be order and amount of error messages
4. in org.mapstruct.ap.test.erroneous.annotationnotfound.AnnotationNotFoundTest, seems to order and amount of error messages

Errors:
5. org.mapstruct.ap.test.severalsources.SeveralSourceParametersTest, java.lang.annotation.Repeatable is not found.
6.  org.mapstruct.ap.test.unmappedtarget.UnmappedTargetTest,  java.lang.annotation.Repeatable is not found.
7. org.mapstruct.ap.test.versioninfo.VersionInfoTest,  java.lang.annotation.Repeatable is not found.

agudian:
Check #694 on what the different compilers have to say... :)

agudian:
To sum that up what the JDK 6 compilation of `GenericsHierarchyTest` shows: MapStruct generates what we would expect (exactly the same `GenericsHierarchyMapperImpl` as for all other compilers), but javac in JDK 6 can't yet compile it. I think that's totally OK for us.

sjaakd:
> but javac in JDK 6 can't yet compile it. I think that's totally OK for us.

Agreed.

gunnarmorling:
> but javac in JDK 6 can't yet compile it. I think that's totally OK for us.

Why is it ok? I am confused :) Is it using some JDK 7/8 only language feature?

I guess my main question is whether MapStruct built based on this PR is working for you on JDK 6, @sjaakd. I suppose it is, otherwise you would have raised your voice already :)

sjaakd:
> @sjaakd. I suppose it is, otherwise you would have raised your voice already :)

The one test case which is failing that is relevant is @agudian newly added test case. The other ones are explainable. 

Now, first of all my apologies for not testing this at my work yet. Its a very busy time (@work), and pulling in some late hours to get stuff out, so not much time remaining for MapStruct :cry:. 

Speeding up matters: I think I read it somewhere (in one of the issues), but can I get the 1.0.0-SNAPSHOT versions from somewhere, so that I don't have to build it myself @work?

agudian:
> Why is it ok? I am confused :) Is it using some JDK 7/8 only language feature?

No, but javac 1.6 can't resolve which method is meant in the generated code - I don't know if it's a bug in javac, or if that behaviour was later added to the JLS (javac 1.6 had a couple of problems with resolving funky generics, and many of them have been fixed in 1.7 and 1.8 - or ended up with a refinement of the spec). 
But still, the APT implementation in JDK 1.6 is able to deduct the right method type for us, as we wouldn't otherwise generate the mapper implementation in exactly that way (including the calls to `modifyItemC` and `modifyItemB`). So this is as good as we can get it, I guess.

agudian:
> Speeding up matters: I think I read it somewhere (in one of the issues), but can I get the 1.0.0-SNAPSHOT versions from somewhere, so that I don't have to build it myself @work?

You can get the latest snapshots of master from the oss-sonatype snapshots repo:
https://oss.sonatype.org/content/repositories/snapshots/

Right now I'm deploying the current state of this PR to that location... :+1:

gunnarmorling:
> Right now I'm deploying the current state of this PR to that location...

Ah, one step ahead of me. Just meant to do the same, thanks!

gunnarmorling:
Merged; Many thanks, @agudian!

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/EclipseAsMemberOfWorkaround.java
```diff
@@ -0,0 +1,196 @@
...
+            methodBinding,
+            referenceBinding,
+            new HashSet<ReferenceBinding>(),
+            candidatesFromInterfaces );
+
+        // there can be multiple matches for the same method name from adjacent interface hierarchies.
+        Collections.sort( candidatesFromInterfaces, MostSpecificMethodBindingComparator.INSTANCE );
```
gunnarmorling: @agudian Thinking a bit more about this, it seems that sorting step could be avoided if we'd sort the interfaces implemented/extended by a given type themselves already (i.e. where we call `superInterfaces()`). That sorting would be simpler (just checking if one is super-type of the other) than all the param business we do in the current sort. WDYT?

agudian: I don't think that it would be enough. You could still implement two unrelated interfaces that each contain a method 'getItem()' with (possibly generic) types that in the end are comparible. E.g one boils down to getItem():Number, the other to getItem():Long.

sjaakd:
> You can get the latest snapshots of master from the oss-sonatype snapshots repo:
> https://oss.sonatype.org/content/repositories/snapshots/
> 
> Right now I'm deploying the current state of this PR to that location...

Hi guys.. Took me some time to figure out that I needed to set my pluginRepository to sonatype as well (which I think is a bit strange, since MapStruct is a dependency of the BSC plugin, not a plugin itself).

But.. most importantly, I can confirm my current code still compiles / generates at 1.6 level  (at least on my current projects). So we did not 'break' anything here. Nice piece of work @agudian :+1:

PR #699, opened 2015-11-25
#698 Fix missing import for enums in case of default values.

### LINKED ISSUE #698: @Mapping enum to default doesn't add import for enum
I am trying to map an enum to a default value like this

@Mapping(target = "type", defaultValue = "A")

The target enum is in a different package than the mapper.  When the mapper generates I get a compile error in the generated class saying "cannot find symbol".  I went and looked at the generated code and it all looks correct except there is no import statement for the defaulted enum.

I can get around it if I add a mapping for the enum to itself like below, but I would rather not.

public ObjType toType(ObjType type);

Note, I am using version 1.0.0.CR2 of mapstruct.

agudian:
Fixes #698.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/defaultvalue/CountryMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/defaultvalue/CountryMapper.java
@@ -33,6 +33,7 @@ public abstract class CountryMapper {
             @Mapping( target = "id", defaultValue = "42" ),
             @Mapping( target = "zipcode", defaultValue = "1337" ),
             @Mapping( target = "region", defaultValue = "someRegion" ),
+            @Mapping( target = "continent", defaultValue = "EUROPE" )
     } )
     public abstract CountryDts mapToCountryDts(CountryEntity country);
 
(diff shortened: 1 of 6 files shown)


processor/src/test/java/org/mapstruct/ap/test/defaultvalue/CountryMapper.java
```diff
@@ -40,19 +41,19 @@
             @Mapping( target = "code", defaultValue = "DE" ),
             @Mapping( target = "id", defaultValue = "42" ),
             @Mapping( target = "zipcode", defaultValue = "1337" ),
-            @Mapping( target = "region", ignore = true )
-
+            @Mapping( target = "region", ignore = true ),
+            @Mapping( target = "continent", defaultValue = "EUROPE" )
```
sjaakd: earth?, solar system? galaxy? local cluster?.. We still have room for more patches :smile:

gunnarmorling: Ahahaha :)

gunnarmorling:
Merged; Thanks for the quick fix, @agudian!

sjaakd:
I had no more comments either :smile: .. thanks!

Issue #710, opened 2015-12-17
Properly work together with other processors creating bean super types


gunnarmorling:
We need to defer handling of mapper types which have source/target types whose type hierarchy is not fully existent yet, because super-types are to be generated by other annotation processors.

I'll send a PR for this shortly, it's a fall-out of my investigations regarding the MapStruct/Lombok integration.

gunnarmorling:
Done.

Issue #709, opened 2015-12-15
Investigate benefit of passing originating elements to Filer#createSourceFile()


gunnarmorling:
From the `Filer` docs:

> If there are no originating elements, none need to be passed.  This information may be used in an incremental environment to determine the need to rerun processors or remove generated files. Non-incremental environments may ignore the originating element information.

Passing all the types influencing a mapper implementation (mapper interface, source/target types, config types etc.) may cause automatic regeneration in the IDE if e.g. a source object is modified. Let's see whether Eclipse properly evaluates that info.

gunnarmorling:
Ok, so it seems to improve the situation at least in some cases. ERROR markers seem to be generated/refreshed, i.e. if I have a `source()` property name specified and change the corresponding setter name, an error marker will be created for that mapping.

But it wouldn't be Eclipse if it worked completely: for WARN markers it does not seem to work (e.g. unmapped target reports are not shown if the policy is WARN but they will be shown if the policy is ERROR).

Nevertheless I'd say it's better than nothing...

PR #711, opened 2015-12-17
#710 Play nice with other processors for generating supertypes of mapping source/target types

gunnarmorling:
Hey @agudian, @sjaakd Got something for 1.0.1 :)

This re-arranges a bit the pieces in `MappingProcessor`:

* First process any mappers deferred from previous rounds
* Then process mappers from this round

If in einer step a source/target type with an incomplete hierarchy is encountered, its processing is deferred to a next round.

We might defer handling in some other cases as well (e.g. a superclass for a mapper or config type could be generated?!), but as this seems very exotic, I'd say let's wait for the actual need. Should be simple to do with the general deferring mechanism in place.

Note that this does not address the Lombok issue, but it improves the situation when using MapStruct with other processors playing by the rules of JSR 269.


gunnarmorling:
Oops, just seeing I accidentally merged this already :/ I think it's good, but please let me know if you have any remarks.

agudian:
One thing that jumps to my attention is the tab-indentation in the integration test java files... :eyes: 

But on the stuff itself: did you test-drive it in Eclipse as well? 

And how about the idea with collecting and passing all the originatingElements during the mapper creation? That sounded like it could also help us getting things right(er).

agudian:
And one more thing. How about remembering only the classnames in `deferredMappers`, to avoid keeping references to the `TypeElement` instances. And when using the set later on, you do a fresh lookup of the new `TypeElement` using the class name anyway.

gunnarmorling:
> One thing that jumps to my attention is the tab-indentation in the integration test java files...

Agh... I suppose CS doesn't check these?

> But on the stuff itself: did you test-drive it in Eclipse as well? 

Hum, no, but good idea. I'll give it a try

> And how about the idea with collecting and passing all the originatingElements during the mapper creation?

Right, working on this as we speak as part of #709.

> How about remembering only the classnames in deferredMappers

Ah yes, should have done that to begin with. Thanks!

I'll send a PR with the adjustments.

PR #713, opened 2015-12-21
jaxb underscores in schema element names

grandmasterpixel:
Hi,
since I have only been working with jaxb, xml schemas and soap for about 2 weeks there is a lot I still don't know about the problem I "solved" in this PR.

I have a wsdl of a third party soap service. Some of the elements have names using underscores instead of camel case like this:
```xml
<element name="planet_name" type="string"/>
```
which is perfectly ok.

When generating code from that wsdl with the "cxf-codegen-plugin" I get Code mit mostly "javaified" names (camel case with no underscores)
In the "ObjectFactory" the factory methods have @XmlElementDecl annotations that use the underscore names to set their "name" properties. Which might look like this:
 ```java
    @XmlElementDecl(namespace = "http://foo.bar", name = "planet_name", scope = TypeName.class)
    public JAXBElement<FieldType> createTypeNamePlanetName(FieldType value) {
        return new JAXBElement<FieldType>(_TypeNamePlanetName_QNAME, FieldType.class, TypeName.class, value);
    }
```

Still kind of ok until I am trying to map my classes to the generated jaxb classes. For the cases where there is more than one property of the same type and one of them was generated from an underscore name in the schema I get a "Ambiguous mapping methods found for mapping property" compile error.

That seems to be because the XmlElementDeclSelector tries to exactly match the camel case name with the underscore name in the annotation. Failing that is falls back to only matching the scope, which result in too many candidates.

**What I changed** was add a third match on a method suffix. It seemed to me the generated code in ObjectFactory always followed the pattern of "TypeNamePropertyName" in the method names, so using the property name and match it as a suffix seemed like it might work.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
@@ -21,18 +21,25 @@ package org.mapstruct.ap.internal.model.source.selector;
 import java.util.ArrayList;
 import java.util.List;
 
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
 import javax.xml.bind.annotation.XmlElementDecl;
+import javax.xml.bind.annotation.XmlElementRef;
 
 import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.model.source.SourceMethod;
 import org.mapstruct.ap.internal.prism.XmlElementDeclPrism;
+import org.mapstruct.ap.internal.prism.XmlElementRefPrism;
 
 /**
- * Selects those methods with matching {@code name} and {@code scope} attributes of the {@link XmlElementDecl}
- * annotation, if that is present. Matching happens in the following order:
+ * Finds the {@link XmlElementRef} annotation on a field (of the mapping result type or its super types) matching the
+ * target property name. Then selects those methods with matching {@code name} and {@code scope} attributes of the
+ * {@link XmlElementDecl} annotation, if that is present. Matching happens in the following order:
  * <ol>
  * <li>Name and Scope matches</li>
  * <li>Scope matches</li>
@@ -46,9 +53,11 @@ import org.mapstruct.ap.internal.prism.XmlElementDeclPrism;
 public class XmlElementDeclSelector implements MethodSelector {
 
     private final Types typeUtils;
+    private final Elements elementUtils;
 
-    public XmlElementDeclSelector(Types typeUtils) {
+    public XmlElementDeclSelector( Types typeUtils, Elements elementUtils) {
         this.typeUtils = typeUtils;
+        this.elementUtils = elementUtils;
     }
 
     @Override
@@ -66,6 +75,8 @@ public class XmlElementDeclSelector implements MethodSelector {
         List<T> nameMatches = new ArrayList<T>();
         List<T> scopeMatches = new ArrayList<T>();
         List<T> nameAndScopeMatches = new ArrayList<T>();
+        XmlElementRefInfo xmlElementRefInfo =
+            findXmlElementRef( sourceMappingMethod.getResultType(), criteria.getTargetPropertyName() );
 
         for ( T candidate : methods ) {
             if ( !( candidate instanceof SourceMethod ) ) {
@@ -81,10 +92,10 @@ public class XmlElementDeclSelector implements MethodSelector {
 
             String name = xmlElememtDecl.name();
             TypeMirror scope = xmlElememtDecl.scope();
-            TypeMirror target = sourceMappingMethod.getExecutable().getReturnType();
 
-            boolean nameIsSetAndMatches = name != null && name.equals( criteria.getTargetPropertyName() );
-            boolean scopeIsSetAndMatches = scope != null && typeUtils.isSameType( scope, target );
+            boolean nameIsSetAndMatches = name != null && name.equals( xmlElementRefInfo.nameValue() );
+            boolean scopeIsSetAndMatches =
+                scope != null && typeUtils.isSameType( scope, xmlElementRefInfo.sourceType() );
 
             if ( nameIsSetAndMatches ) {
                 if ( scopeIsSetAndMatches ) {
@@ -110,6 +121,74 @@ public class XmlElementDeclSelector implements MethodSelector {
         }
         else {
             return methods;
+        }
+    }
+
+    /**
+     * Iterate through resultType and its super types to find a field named targetPropertyName and return information
+     * about:
+     * <ul>
+     * <li>what the value of the name property of the XmlElementRef annotation on that field was</li>
+     * <li>on which type the field was found</li>
+     * </ul>
+     *
+     * @param resultType starting point of the iteration
+     * @param targetPropertyName name of the field we are looking for
+     * @return an XmlElementRefInfo containing the information
+     */
+    private XmlElementRefInfo findXmlElementRef(Type resultType, String targetPropertyName) {
+        TypeMirror startingMirror = resultType.getTypeMirror();
+        XmlElementRefInfo defaultInfo = new XmlElementRefInfo( targetPropertyName, startingMirror );
+        if ( targetPropertyName == null ) {
+            /*
+             * sometimes MethodSelectors seem to be called with criteria.getTargetPropertyName() == null so we need to
+             * avoid NPEs for that case.
+             */
+            return defaultInfo;
+        }
+
+        TypeMirror currentMirror = startingMirror;
+        TypeElement currentElement = resultType.getTypeElement();
+
+        /*
+         * Outer loop for resultType and its super types. "currentElement" will be null once we reach Object and try to
+         * get a TypeElement for its super type.
+         */
+        while ( currentElement != null ) {
+            /*
+             * Inner loop tries to find a field with the targetPropertyName and assumes that where the XmlElementRef is
+             * set
+             */
+            for ( Element enclosed : currentElement.getEnclosedElements() ) {
+                if ( enclosed.getKind().equals( ElementKind.FIELD )
+                    && enclosed.getSimpleName().contentEquals( targetPropertyName ) ) {
+                    XmlElementRefPrism xmlElementRef = XmlElementRefPrism.getInstanceOn( enclosed );
+                    if ( xmlElementRef != null ) {
+                        return new XmlElementRefInfo( xmlElementRef.name(), currentMirror );
+                    }
+                }
+            }
+            currentMirror = currentElement.getSuperclass();
+            currentElement = elementUtils.getTypeElement( currentMirror.toString() );
+        }
+        return defaultInfo;
+    }
+
+    private static class XmlElementRefInfo {
+        private final String nameValue;
+        private final TypeMirror sourceType;
+
+        XmlElementRefInfo(String nameValue, TypeMirror sourceType) {
+            this.nameValue = nameValue;
+            this.sourceType = sourceType;
+        }
+
+        public String nameValue() {
+            return nameValue;
+        }
+
+        public TypeMirror sourceType() {
+            return sourceType;
         }
     }
 }
--- a/integrationtest/src/test/resources/jaxbTest/pom.xml
+++ b/integrationtest/src/test/resources/jaxbTest/pom.xml
@@ -50,6 +50,7 @@
                     <schemaDirectory>\${project.build.resources[0].directory}/schema/</schemaDirectory>
                     <schemaIncludes>
                         <include>**/test1.xsd</include>
+                        <include>**/underscores.xsd</include>
                     </schemaIncludes>
                     <bindingDirectory>\${project.build.resources[0].directory}/binding</bindingDirectory>
                     <bindingIncludes>
--- a/integrationtest/src/test/resources/jaxbTest/src/main/java/org/mapstruct/itest/jaxb/SourceTargetMapper.java
+++ b/integrationtest/src/test/resources/jaxbTest/src/main/java/org/mapstruct/itest/jaxb/SourceTargetMapper.java
@@ -24,6 +24,7 @@ import org.mapstruct.itest.jaxb.xsd.test1.OrderDetailsType;
 import org.mapstruct.itest.jaxb.xsd.test1.OrderType;
 import org.mapstruct.itest.jaxb.xsd.test2.OrderStatusType;
 import org.mapstruct.itest.jaxb.xsd.test2.ShippingAddressType;
+import org.mapstruct.itest.jaxb.xsd.underscores.SubType;
 
 
 /**
@@ -32,6 +33,7 @@ import org.mapstruct.itest.jaxb.xsd.test2.ShippingAddressType;
 @Mapper(uses = {
     org.mapstruct.itest.jaxb.xsd.test1.ObjectFactory.class,
     org.mapstruct.itest.jaxb.xsd.test2.ObjectFactory.class,
+    org.mapstruct.itest.jaxb.xsd.underscores.ObjectFactory.class,
     JaxbMapper.class
 })
 public interface SourceTargetMapper {
@@ -47,6 +49,8 @@ public interface SourceTargetMapper {
 
     ShippingAddressDto shippingAddressToDto(ShippingAddressType source);
 
+    SubTypeDto subTypeToDto(SubType source);
+
     // target 2 source methods
     OrderType targetToSource(OrderDto target);
 
@@ -55,4 +59,6 @@ public interface SourceTargetMapper {
     OrderStatusType dtoToStatus(OrderStatusDto target);
 
     ShippingAddressType dtoToShippingAddress(ShippingAddressDto source);
+
+    SubType dtoToSubType(SubTypeDto source);
 }
--- a/integrationtest/src/test/resources/jaxbTest/src/main/java/org/mapstruct/itest/jaxb/SubTypeDto.java
+++ b/integrationtest/src/test/resources/jaxbTest/src/main/java/org/mapstruct/itest/jaxb/SubTypeDto.java
@@ -1,0 +1,40 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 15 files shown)

### CODE REVIEW

**Comment by agudian** [comment]

**Comment by sjaakd** [comment]
Xjc also has an arg to decide how to handle dashes/underscores during code generation I believe. Would that be of any use?

**Comment by gunnarmorling** [comment]
@sjaakd, maybe the JAXB method matching strategy should be adapted to take
the "_" pattern in `name()` into account as well?

**Comment by grandmasterpixel** [comment]
failed to mention, I tried setting underscoreBinding="asCharInWord" but for some reason that only made it convert "planet_name" to "Planet_Name" which still does not match what was set in the annotation

**Comment by grandmasterpixel** [comment]
ok, the suffix idea was kind of stupid. It falls down as soon as are properties like "planetName" and "currentPlanetName" where one is a suffix of the other. then the code in my branch produces ambiguous matches. so nothing gained :)

A different idea I am exploring right now is to inspect the targetType and find the field annotated with XmlElementRef. These annotations also contain the name exactly how it would be spelled over in the XmlElementDecl annotation.

**Comment by sjaakd** [comment]
> @sjaakd, maybe the JAXB method matching strategy should be adapted to take
> the "_" pattern in `name()` into account as well?

Yep.. That would make sense.. Would that solve the problem above? E.g. construct a regex that matches both `_` and `-`.. or something along that line..

**Comment by gunnarmorling** [comment]
> Would that solve the problem above?

Yes, I think it would. So from a camel-cased property name such as "planetName" we'd also find an element named "planet_name" in XML. Though we'd have to think about avoiding ambiguities if the elements "planet_name" and "planetName" are there.

> A different idea I am exploring right now is to inspect the targetType and find the field annotated with XmlElementRef. These annotations also contain the name exactly how it would be spelled over in the XmlElementDecl annotation.

That sounds interesting. How would that look like? Thanks for working on this btw., @grandmasterpixel!

**Comment by grandmasterpixel** [comment]
So my idea was something like this:
Since I get code from xjc that looks like this
```java
@XmlAccessorType(XmlAccessType.FIELD)
@XmlType(name = "ShippingAddressType", propOrder = {
    "street",
    "houseNumber",
    "city",
    "country",
    "planetName"
})
public class ShippingAddressType {
//...
    @XmlElementRef(name = "planet_name", namespace = "http://www.mapstruct.org/itest/jaxb/xsd/test2", type = JAXBElement.class)
    protected JAXBElement<String> planetName;
//...
}
```
and this
```java
@XmlRegistry
public class ObjectFactory {
//...
    @XmlElementDecl(namespace = "http://www.mapstruct.org/itest/jaxb/xsd/test2", name = "houseNumber", scope = ShippingAddressType.class)
    public JAXBElement<String> createShippingAddressTypeHouseNumber(String value) {
        return new JAXBElement<String>(_ShippingAddressTypeHouseNumber_QNAME, String.class, ShippingAddressType.class, value);
    }

    @XmlElementDecl(namespace = "http://www.mapstruct.org/itest/jaxb/xsd/test2", name = "planet_name", scope = ShippingAddressType.class)
    public JAXBElement<String> createShippingAddressTypePlanetName(String value) {
        return new JAXBElement<String>(_ShippingAddressTypePlanetName_QNAME, String.class, ShippingAddressType.class, value);
    }

}
```
I can see what factory method to choose from the information on the annotations.

So when XmlElementDeclSelector gets called to find a factory method for "planetName" it would need to
* inspect the ShippingAddressType class and find the field for the property
* look up the name in the XmlElementRef annotation
* then use that to find the method annotated with XmlElementDecl having the right name and scope

the last point was already solved by you so is was trying my luck on the other parts.
I got caught up in a misunderstanding about the meaning of the sourceType and targetType parameters of the getMatchingMethods method. I mistook them for the source and target of the mapping method, but I think they are actually the types of the source and target properties. (in which case I think they should have been called sourcePropertyType and targetPropertyType ;))
anyway: a useable reference to the ShippingAddressType should be present the mappingMethod parameter so I try using that in my next attempt :)

grandmasterpixel:
what I last pushed still needs some code and branch cleanup but it is able to find the right factory for "planet_name" that I added to the jaxb integration test earlier.
I am using sourceMappingMethod.getResultType() over sourceMappingMethod.getReturnType() because it seemed to be able to have the actual type be overridden. Didn't look to hard though since I just wanted this idea out of my head for now and deal with the specifics later ;). I assumed it has something to do with cases where we want to map to interfaces or abstract classes (google auto values for example) but looking at the docs it seems not to be exactly that.

grandmasterpixel:
My first XmlElementRef lookup loop did not traverse up the inheritance chain. Now it does and a new version of the jaxb integration test tells me that that is the right thing to do :)
This should now fix the issue I had. Have not compiled our code with this yet though

grandmasterpixel:
I am back at work and my colleagues are already using a mapstruct version built from my branch so at least our problem is solved.
Also looking at the docs for [XmlElementRef](https://docs.oracle.com/javase/7/docs/api/javax/xml/bind/annotation/XmlElementRef.html) and [XmlElementDecl](https://docs.oracle.com/javase/7/docs/api/javax/xml/bind/annotation/XmlElementDecl.html) it looks like I am clearly taking the right approach here.
But this is still missing at least one thing:
The XmlElementRefs (plural) annotation as demonstrated in an example in the XmlElementDecl javadoc. That seems not easily solvable to me at the moment. I have not thought about it much though.

Since our main problem is solved I will only be working on this on my own time now. Probably mostly in the evening (Central European Time) or weekends. That is unless we run into XmlElementRefs at work :)

gunnarmorling:
Hey, thanks for reporting back! Back online again, too, and we've planned
to finally take a closer look at this change later this week.

grandmasterpixel:
Today I have been playing around a bit.
With a schema like this:
```xml
	<complexType name="pea">
		<choice maxOccurs="unbounded">
			<element name="foo" type="xmlelementrefs:FooType"/>
			<element name="bar" type="xmlelementrefs:BarType"/>
		</choice>
	</complexType>
	<complexType name="FooType">
		<sequence>
			<element name="theFoo" type="string"/>
			<element name="otherFoo" type="string"/>
		</sequence>
	</complexType>
	<complexType name="BarType">
		<sequence>
			<element name="theBar" type="string"/>
			<element name="otherBar" type="string"/>
		</sequence>
	</complexType>

	<element name="pea" type="xmlelementrefs:pea"/>
```
I get a Pea class with a property like this
```java
    @XmlElementRefs({
        @XmlElementRef(name = "bar", namespace = "http://www.mapstruct.org/itest/jaxb/xsd/xmlelementrefs", type = JAXBElement.class),
        @XmlElementRef(name = "foo", namespace = "http://www.mapstruct.org/itest/jaxb/xsd/xmlelementrefs", type = JAXBElement.class)
    })
    protected List<JAXBElement<?>> fooOrBar;
```
Which means this xml creates a heterogeneous List of a FooType and a BarType
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<pea xmlns="http://www.mapstruct.org/itest/jaxb/xsd/xmlelementrefs">
    <foo>
        <theFoo>0</theFoo>
        <otherFoo>0</otherFoo>
    </foo>
    <bar>
        <theBar>0</theBar>
        <otherBar>0</otherBar>
    </bar>
</pea>
```
So to create a mapping method for that mapstruct would need to generate some kind "if instanceof else if" construct, I think. Does mapstruct already do something like that in the nonjaxb case?

sjaakd:
Hi @grandmasterpixel ,

I was just reviewing your code :smile:

> So to create a mapping method for that mapstruct would need to generate some kind "if instanceof else if" construct, I think. Does mapstruct already do something like that in the nonjaxb case?

Did you look at the `@TargetType` annotation?  That was intended AFAIK to add a target type to the selection. Perhaps that's usable?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
```diff
@@ -112,4 +119,38 @@ else if ( nameMatches.size() > 0 ) {
             return methods;
         }
     }
+
+    private Pair getMatch(SourceMethod sourceMappingMethod, final String targetPropertyName) {
+        final TypeMirror startingMirror = sourceMappingMethod.getExecutable().getReturnType();
```
sjaakd: I reckon resultType is more safe.. (you do that a bit later as well)

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
```diff
@@ -112,4 +119,38 @@ else if ( nameMatches.size() > 0 ) {
             return methods;
         }
     }
+
+    private Pair getMatch(SourceMethod sourceMappingMethod, final String targetPropertyName) {
```
sjaakd: why is `targetPropertyName` final and `sourceMappingMethod` not? Although I agree, its better to use the final keyword to indicate a variable is not modified (and its slightly better performing as well I once understood), I don't think we do that in other places of the code. So perhaps we should stick to the conventions and remove the final keywords.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
```diff
@@ -112,4 +119,38 @@ else if ( nameMatches.size() > 0 ) {
...
+        final TypeMirror startingMirror = sourceMappingMethod.getExecutable().getReturnType();
+        TypeMirror currentMirror = startingMirror;
+        TypeElement currentElement = sourceMappingMethod.getResultType().getTypeElement();
+
+        while (currentElement != null) {
+            for (Element enclosed : currentElement.getEnclosedElements()) {
+                if (enclosed.getKind().equals( ElementKind.FIELD ) && targetPropertyName != null
```
sjaakd: are there situations in which `targetPropertyName == null`?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
```diff
@@ -112,4 +119,38 @@ else if ( nameMatches.size() > 0 ) {
...
+                        && enclosed.getSimpleName().contentEquals( targetPropertyName )) {
```
sjaakd: I'm having problems to understand this loop. Can you add some comments here what you are trying to achieve?

The problem is that the selectors are used for both selecting factory methods and mapping methods. This case is very much geared towards finding factory methods in an object factory.

What I understand: you are looking for fields in the target types of the candidate mapping methods? What you need is (I think) the target bean, or better, an `XmlElementRef` `AnnotationMirror`. Wouldn't it be better to add those to the selection criteria when you call the whole mapper chain? Thats what the `SelectionCriteria` are for.

You mentioned somewhere in the discussion that the `targetType` should be called `targetPropertyType`. I now understand your remark better :smile: . But on this level, we're just trying to find a mapping method for mapping a `sourceType` to a `targetType`. There is no understanding/knowledge of bean properties here. Selection context should come via the `SelectionCriteria`.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
```diff
@@ -112,4 +119,38 @@ else if ( nameMatches.size() > 0 ) {
...
+                    XmlElementRef xmlElementRef = enclosed.getAnnotation( XmlElementRef.class );
```
sjaakd: We use prism's for obtaining annotation mirrors. Have a look at `org.mapstruct.ap.internal.prism.PrismGenerator`.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
```diff
@@ -112,4 +119,38 @@ else if ( nameMatches.size() > 0 ) {
...
+                    }
+                }
+            }
+            currentMirror = currentElement.getSuperclass();
+            currentElement = elementUtils.getTypeElement( currentElement.getSuperclass().toString() );
+        }
+        return new Pair(targetPropertyName, startingMirror);
+    }
+
+    //CHECKSTYLE:OFF
```
sjaakd: why is `CHECKSTYLE:OFF` needed?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
```diff
@@ -112,4 +119,38 @@ else if ( nameMatches.size() > 0 ) {
...
+    private static class Pair {
```
sjaakd: `Pair` is not very descriptive.

sjaakd:
@grandmasterpixel . First of all, thanks for your contribution.

I ran through your code. Main thing is that I find the selector hard to understand. 

What I do understand is that you (somehow) try to match up the `XmlElementRef` and `XmlElementDecl` in the following way:
You obtain the `XmlElementRef` via the following steps:
1. Find the target property in the target bean (or iterable /map) of  the mapping method we try to implement.
2. Find the annotation `XmlElementRef`
3. See if you find a name match in the `XmlElementDecl` object factory method.

I think you're on the right track here.

Main issues: 
1. I think that this `XmlElementRef` (annotation of the target property) is not altogether different from the `targetPropertyName`. So perhaps it should be part of the `SelectionCriteria`. Its 'meta' information to make the selection if you catch my idea.
2. Use `Prisms`. Its a more safe way to obtain an annotation.
3. Use some more comments :smile:.

Next: I like the integration test a lot. However, I would like to see some (fixed) unit test as well. See for instance test package: `org.mapstruct.ap.test.selection.jaxb.test1`.

Last but not least, I think you should cover additional problems in a separate PR. In this way we can speed up your solution a bit :+1:

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
sjaakd: You could also consider using `AbstractMap.SimpleEntry `

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
gunnarmorling: Yes, but it's not needed for the JAXB types. We use the prisms for the MapStruct annotations to avoid having to provide them on the processor path. As JAXB is provided by the JDK, accessing the annotation is fine.

gunnarmorling:
Hey @grandmasterpixel, I am a bit confused by the changes in `XmlElementDeclSelector`, esp. this handling around super-types.

Wouldn't it be sufficient to alter the name check so it's based on a "normalized" value of `XmlElementDecl#name()` (by camel-casing and removing the underscore)?

E.g. we look for a method matching property name "planetName", and we have two candidate methods:

    * `XmlElementDecl@name("universe_name")` -> convert name to "universeName", compare to "planetName" -> no match
    * `XmlElementDecl@name("planet_name")` -> convert name to "planetName", compare to "planetName" -> match

Now if there was another method XmlElementDecl@name("planetName")`, we'd still have an ambiguity between the last two, but I think that's alright. Not much we can do about it, but it's an edge case and users can work around that by providing qualified wrapper methods if needed.

WDYT?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
grandmasterpixel: sometimes I make variables final because I am accessing them in an iteration. that is to show readers that these variables are to be understood as the input and everything else can be considered output of that iteration.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
grandmasterpixel: yes it may be null. If I remember correctly, I broke like 99% of the unittests at first because I did not expect SelectionCriteria.getTargetPropertyName() (which is passed into "targetPropertyName") to be null :)

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
grandmasterpixel: Since prims are already used for XmlElementDecl in XmlElementDeclSelector I think I switch to prisms just for consistency.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
grandmasterpixel: Because I am not using proper private scope and accessor methods in "Pair"

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
gunnarmorling: Sounds good. Didn't notice they are used for that already, but as that's the case +1 :)

grandmasterpixel:
If it just had been the case that I just needed to camelcase (planet_name->planetName) XmlElementDecl.name to get a valid match I would have probably gone with that.
But there is also the case of the underscoreBinding="asCharInWord" setting that can be used to customize the binding (http://docs.oracle.com/javaee/5/tutorial/doc/bnbbf.html) which would generate "planet_Name".
Which means without knowledge of how the user configured xjc I would have needed to do at least "planet_name->{planetName, planet_Name} compare both to current property". I did not know if there might be further configuration options so I figured using the annotations was the most reliable source of information of which ObjectFactory method to choose.

regarding jaxb collisions:
We had those too, for simple cases cxf has an "autoNameResolution" switch that we use.
Our collision was in the class names though, but I would still assume it does the following:
If an xml type had both planet_name and planetName elements, a class with properties planetName and planetName2 would be generated.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/XmlElementDeclSelector.java
grandmasterpixel: uh, thanks, using getExecutable().getReturnType() would have actually been a bug :)

grandmasterpixel:
just pushed something that I hope fixes all of the issues sjaakd mentioned

gunnarmorling:
@grandmasterpixel, cool, looks good to merge for me. Neat stuff with the `@XmlElementRef` path!

Could you only change the e-mail of the commits to the one associated with your GitHub account and squash everything into a single commit; That way they will be properly [linked to your user](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/). Thanks! Your name (for copyright.txt) is Vincent Alexander Beelte, right?

sjaakd:
@grandmasterpixel. Looks good at first glance. I'll have one more thorough look tomorrow. 

One thing I spotted though is the absence of unit test (only integration test). I do think its quite straightforward to add a unit test as well (there are already some for jaxb selection). In that way, we spot problems immediately and don't have to wait for the outcome of the integration test.

grandmasterpixel:
Forgot the unittest. Will be present in the squashed commit.
And yes, my name is Vincent Alexander Beelte

gunnarmorling:
Ah, could you also create an issue for it and refer it's number from the commit message ("#123 Bla bla"). Thx!

grandmasterpixel:
added unittest, rebased from origin/master fixed compilation, also I opened and referenced ticket #726
next I will be opening a ticket to request the checkstyle and licence checking be done in an earlier maven phase :)

agudian:
Writing this to trigger the Jenkins pull-request-builder:
ok to test.

gunnarmorling:
Merged; @grandmasterpixel, many thanks for implementing this and following up on all the review remarks. Much appreciated!

agudian:
or not, as gunnar was faster merging it... :wink:

PR #714, opened 2015-12-21
Feature/number format conversion

cliedeman:
First draft implementation for supporting custom number formatting during conversions

### DIFF
--- a/core-common/src/main/java/org/mapstruct/IterableMapping.java
+++ b/core-common/src/main/java/org/mapstruct/IterableMapping.java
@@ -47,6 +47,14 @@ public @interface IterableMapping {
     String dateFormat() default "";
 
     /**
+     * A format string as processable by {@link java.text.DecimalFormat} if the annotated method maps from a
+     *  {@link Number} to a {@link String} or vice-versa. Will be ignored for all other element types.
+     *
+     * @return A decimal format string as processable by {@link java.text.DecimalFormat}.
+     */
+    String numberFormat() default "";
+
+    /**
      * A qualifier can be specified to aid the selection process of a suitable mapper. This is useful in case multiple
      * mappers (hand written of internal) qualify and result in an 'Ambiguous mapping methods found' error.
      *
--- a/core-jdk8/src/main/java/org/mapstruct/Mapping.java
+++ b/core-jdk8/src/main/java/org/mapstruct/Mapping.java
@@ -81,6 +81,14 @@ public @interface Mapping {
     String dateFormat() default "";
 
     /**
+     * A format string as processable by {@link java.text.DecimalFormat} if the annotated method maps from a
+     *  {@link Number} to a {@link String} or vice-versa. Will be ignored for all other element types.
+     *
+     * @return A decimal format string as processable by {@link java.text.DecimalFormat}.
+     */
+    String numberFormat() default "";
+
+    /**
      * A constant {@link String} based on which the specified target property is to be set. If the designated target
      * property is not of type {@code String}, the value will be converted by applying a matching conversion method or
      * built-in conversion.
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/PossibleNumberToStringConversion.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/PossibleNumberToStringConversion.java
@@ -1,0 +1,168 @@
+/**
+ *  Copyright 2012-2015 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.conversion;
+
+import org.mapstruct.ap.internal.model.common.ConversionContext;
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.util.NativeTypes;
+
+
+import org.mapstruct.ap.internal.util.Strings;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.text.DecimalFormat;
+import java.text.ParseException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Conversion hooks for {@link Number} to {@link String}. The idea being that <em>if</em>
+ * we have a conversion to string and the source type is a {@code Number} and the mapping
+ * has a {@link org.mapstruct.Mapping#numberFormat()} we will replace the standard conversion
+ * with a {@link DecimalFormat} version
+ *
+ * @author Ciaran Liedeman
+ */
+public abstract class PossibleNumberToStringConversion extends SimpleConversion {
+
+    protected final Class<?> primitiveType;
+    protected final Class<?> wrapperType;
+    private final Class<?> sourceType;
+    private boolean sourceTypeNumberSubclass;
+
+    public PossibleNumberToStringConversion(final Class<?> sourceType) {
+        this.sourceType = sourceType;
+
+        if ( sourceType.isPrimitive() ) {
+            wrapperType = NativeTypes.getWrapperType( sourceType );
+            primitiveType = sourceType;
+        }
+        else {
+            wrapperType = sourceType;
+            primitiveType = NativeTypes.getPrimitiveType( sourceType );
+        }
+
+        sourceTypeNumberSubclass = NativeTypes.isNumber( primitiveType )
+                || NativeTypes.isNumber( sourceType );
+    }
+
+    @Override
+    protected String getToExpression(ConversionContext conversionContext) {
+        if ( isDecimalFormatConversion( conversionContext ) ) {
+            return getFormatterToConversionExpression( conversionContext );
+        }
+        else {
+            return getFallbackToExpression( conversionContext );
+        }
+    }
+
+    @Override
+    protected String getFromExpression(ConversionContext conversionContext) {
+        if ( isDecimalFormatConversion( conversionContext ) ) {
+            return getFormatterFromConversionExpression( conversionContext );
+        }
+        else {
+            return getFallbackFromExpression( conversionContext );
+        }
+    }
+
+    protected abstract String getFallbackToExpression(ConversionContext conversionContext);
+
+    protected abstract String getFallbackFromExpression(ConversionContext conversionContext);
+
+
+    public Set<Type> getToConversionImportTypes(ConversionContext conversionContext) {
+        if ( isDecimalFormatConversion( conversionContext ) ) {
+            return Collections.singleton( conversionContext.getTypeFactory().getType( DecimalFormat.class ) );
+        }
+        else {
+            return super.getToConversionImportTypes( conversionContext );
+        }
+    }
+
+    private boolean isDecimalFormatConversion(ConversionContext conversionContext) {
+        return sourceTypeNumberSubclass && conversionContext.getNumberFormat() != null;
+    }
+
+    protected Set<Type> getFromConversionImportTypes(ConversionContext conversionContext) {
+        if ( isDecimalFormatConversion( conversionContext ) ) {
+            return Collections.singleton( conversionContext.getTypeFactory().getType( DecimalFormat.class ) );
+        }
+        else {
+            return super.getFromConversionImportTypes( conversionContext );
+        }
+    }
+
+    protected List<Type> getFromConversionExceptionTypes(ConversionContext conversionContext) {
+        if ( isDecimalFormatConversion( conversionContext ) ) {
+            return Collections.singletonList( conversionContext.getTypeFactory().getType( ParseException.class ) );
+        }
+        else {
+            return super.getFromConversionExceptionTypes( conversionContext );
+        }
+    }
+
+    private String getFormatterToConversionExpression(ConversionContext conversionContext) {
+        StringBuilder sb = new StringBuilder();
+
+        appendDecimalFormatter( sb, conversionContext );
+        sb.append( ".format" );
+        sb.append( "( <SOURCE> )" );
+
+        return sb.toString();
+    }
+
+    private String getFormatterFromConversionExpression(ConversionContext conversionContext) {
+        StringBuilder sb = new StringBuilder();
+
+        appendDecimalFormatter( sb, conversionContext );
+        sb.append( ".parse" );
+        sb.append( "( <SOURCE> ).toString()" );
+
+        return getParseTypeExpression( sb.toString() );
+    }
+
+    private void appendDecimalFormatter(StringBuilder sb, ConversionContext conversionContext) {
+        sb.append( "new DecimalFormat( " );
+
+        if ( conversionContext.getNumberFormat() != null ) {
+            sb.append( "\"" );
+            sb.append( conversionContext.getNumberFormat() );
+            sb.append( "\"" );
+        }
+
+        sb.append( " )" );
+    }
+
+    protected String getParseTypeExpression(String input) {
+        if ( BigDecimal.class == sourceType ) {
+            // TODO this step should be avoided, we return a BigDecimal from the formatter
+            return "new BigDecimal(" + input + ")";
+        }
+        else if ( BigInteger.class == sourceType ) {
+            return "new BigInteger(" + input + ")";
+        }
+        else {
+            return wrapperType.getSimpleName() + ".parse" +
+                    Strings.capitalize( primitiveType.getSimpleName() ) + "( " + input + " )";
+        }
+    }
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/PrimitiveToStringConversion.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/PrimitiveToStringConversion.java
@@ -19,8 +19,6 @@
 package org.mapstruct.ap.internal.conversion;
 
 import org.mapstruct.ap.internal.model.common.ConversionContext;
-import org.mapstruct.ap.internal.util.NativeTypes;
-import org.mapstruct.ap.internal.util.Strings;
 
 /**
  * Conversion between primitive types such as {@code byte} or {@code long} and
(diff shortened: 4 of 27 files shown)


agudian:

gunnarmorling:
Thanks a lot for sending this PR, @cliedeman! We are going to look into it asap, but it may take some more days, X-Mas is coming up :)

build-config/src/main/resources/build-config/checkstyle.xml
```diff
@@ -122,7 +122,7 @@
         </module>
         <module name="MethodLength"/>
         <module name="ParameterNumber">
-            <property name="max" value="10"/>
```
sjaakd: We prefer to do this in the source code. For this purpose we have an annotation: `@SuppressWarnings("checkstyle:parameter number")`. See for instance `org.mapstruct.ap.internal.model.source.SourceMethod`

core-common/src/main/java/org/mapstruct/IterableMapping.java
```diff
@@ -46,6 +46,8 @@
      */
     String dateFormat() default "";
 
+    String numberFormat() default "";
```
sjaakd: Can you add some javadoc? It is a public API.

core-jdk8/src/main/java/org/mapstruct/Mapping.java
```diff
@@ -80,6 +80,8 @@
      */
     String dateFormat() default "";
 
+    String numberFormat() default "";
```
sjaakd: javadoc? (general)

processor/src/main/java/org/mapstruct/ap/internal/conversion/DecimalFormatterHelper.java
```diff
@@ -0,0 +1,100 @@
...
+            primitiveType = sourceType;
+        } else {
+            wrapperType = sourceType;
+            primitiveType = NativeTypes.getPrimitiveType( sourceType );
+        }
+    }
+
+
+    public Set<Type> getToConversionImportTypes(ConversionContext conversionContext) {
+        return java.util.Collections.singleton( conversionContext.getTypeFactory().getType( DecimalFormat.class ) );
```
sjaakd: import static

processor/src/main/java/org/mapstruct/ap/internal/conversion/DecimalFormatterHelper.java
```diff
@@ -0,0 +1,100 @@
...
+        appendDecimalFormatter( sb, conversionContext );
+        sb.append( ".parse" );
+        sb.append( "( <SOURCE> ).toString()" );
+
+        return getParseTypeExpression( sb.toString() );
+    }
+
+    private void checkIfWeCanFormat() {
+        if ( ! DecimalFormatConstants.NUMBER_TYPES.contains( sourceType )) {
+            throw new IllegalStateException( String.format(
```
sjaakd: Is it possible to bring the message up to this level? i'm not in favor of using an `IllegalStateException` for a user error. See for example
```java
ctx.getMessager().printMessage(
                            method.getExecutable(),
                            beanMappingPrism.mirror,
                            Message.BEANMAPPING_NOT_ASSIGNABLE, resultType, method.getResultType()
                        );
```

sjaakd: with a new (more to the point) message of course :smile:

sjaakd:
There is somewhere a locale problem involved.. I get:

```
  NumberFormatConversionTest.shouldApplyReverseStringConversions:86 expected:<345[346.456756]> but was:<345[.0]>
  NumberFormatConversionTest.shouldApplyReverseStringConversions:86 expected:<345[346.456756]> but was:<345[.0]>
  NumberFormatConversionTest.shouldApplyStringConversions:52 expected:<'1[.]00'> but was:<'1[,]00'>
  NumberFormatConversionTest.shouldApplyStringConversions:52 expected:<'1[.]00'> but was:<'1[,]00'>

```
when trying to build you solution. Have a look how the date formatter tests are carried out.

sjaakd:
@cliedeman : what about Rounding. Default `RoundingMode HALF_EVEN` is used.. But for myself, there are cases that I require truncate (`RoundingMode.DOWN`)

sjaakd:
@cliedeman : Good work. Thanks for contributing. I've got some comments though. I think @gunnarmorling should have a look as well.

It would be nice if formatting and rounding could also be applied to other conversions.. E.g. `BigDecimal` to `BigDecimal`. Any idea how your solution could be extended in this direction?

build-config/src/main/resources/build-config/checkstyle.xml
cliedeman: done

core-common/src/main/java/org/mapstruct/IterableMapping.java
cliedeman: done

core-jdk8/src/main/java/org/mapstruct/Mapping.java
cliedeman: done

processor/src/main/java/org/mapstruct/ap/internal/conversion/DecimalFormatterHelper.java
cliedeman: done

cliedeman:
@sjaakd Hi, thanks for the comments.

I have fixed the broken tests. The setDefaultLocale did the trick.

For rounding down I guess we would have to add another top level annotation option. E.g. numberFormatRoundingType? I can't think of any other way.

As for extending to BigDecimal and BigInteger it is quite easy. The DecimalFormatterHelper was created with that in mind, I just wanted to see what you guys think before implementing the changes everywhere because the decimal formatter helper feels out of place with the style of the rest of the converters.

I would extract the shared code from WrapperToStringConverter and PrimitiveToStringConverter that uses the decimal formatter helper and create class like PossibleNumberToStringConversion (working title :smile: ) and then change the BigDecimal and BigInteger converters to extend this.

processor/src/main/java/org/mapstruct/ap/internal/conversion/DecimalFormatterHelper.java
cliedeman: Can you elaborate on where this code should go? Is there an example of a similar error being raise if we have a dateFormat defined but are not mapping between string and date?

sjaakd:
@cliedeman wow.. Your fast :smile:.

processor/src/main/java/org/mapstruct/ap/internal/conversion/DecimalFormatterHelper.java
sjaakd: > Can you elaborate on where this code should go? Is there an example of a similar error being raise if we have a dateFormat defined but are not mapping between string and date?

the messager is the default mechanism to flag problems to the end user (it obeys also some rules that you can set: e.g. warning, error). Its not yet used in conversion, but you'll find examples in for example the `BeanMappingMethod.java`. Problem is to bring the context from which you can obtain the messager  to the point were you require it.

cliedeman:
@sjaakd I've implemented BigDecimal and BigInteger with formatting and restructured  the code

processor/src/main/java/org/mapstruct/ap/internal/conversion/PossibleNumberToStringConversion.java
```diff
@@ -0,0 +1,168 @@
...
+            sb.append( conversionContext.getNumberFormat() );
+            sb.append( "\"" );
+        }
+
+        sb.append( " )" );
+    }
+
+    protected String getParseTypeExpression(String input) {
+        if ( BigDecimal.class == sourceType ) {
+            // TODO this step should be avoided, we return a BigDecimal from the formatter
```
sjaakd: should this TODO still be fixed?

sjaakd:
> For rounding down I guess we would have to add another top level annotation option. E.g. numberFormatRoundingType? I can't think of any other way.

Thinking out loud.. Since the rounding mode is a (non fluent) setter on the NumberFormatter, for that to work, you probably need the number formatter as an instance variable of the  mapper, iso constructing a new one, each call. Right? Doing that would possibly incur thread safety issues.. Hmmm..

sjaakd:
@cliedeman looks good. Would be a very nice addition. Thanks. @gunnarmorling : can you have a look as well.

core-common/src/main/java/org/mapstruct/IterableMapping.java
```diff
@@ -47,6 +47,14 @@
     String dateFormat() default "";
 
     /**
+     * A format string as processable by {@link java.text.DecimalFormat} if the annotated method maps from a
+     *  {@link Number} to a {@link String} or vice-versa. Will be ignored for all other element types.
```
gunnarmorling: This includes primitive numeric types as well, right? I'd say it like that:

> ...maps from a numeric type (either primitive or a sub-type of {@link Number}) to a {@link String} or vice-versa...

core-common/src/main/java/org/mapstruct/IterableMapping.java
```diff
@@ -47,6 +47,14 @@
     String dateFormat() default "";
 
     /**
+     * A format string as processable by {@link java.text.DecimalFormat} if the annotated method maps from a
```
gunnarmorling: You can import this, our CS rules should allow for it. Makes for nicer readable docs.

processor/src/main/java/org/mapstruct/ap/internal/conversion/PrimitiveToStringConversion.java
```diff
@@ -28,28 +26,23 @@
...
-
-    private final Class<?> sourceType;
-    private final Class<?> wrapperType;
+public class PrimitiveToStringConversion extends PossibleNumberToStringConversion {
 
     public PrimitiveToStringConversion(Class<?> sourceType) {
+        super( sourceType );
         if ( !sourceType.isPrimitive() ) {
-            throw new IllegalArgumentException( sourceType + " is no primitive type." );
+            throw new IllegalArgumentException( sourceType + " is not primitive type." );
```
gunnarmorling: I think it should be the original one or "is not *a* primitive type".

processor/src/main/java/org/mapstruct/ap/internal/conversion/PossibleNumberToStringConversion.java
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Conversion hooks for {@link Number} to {@link String}. The idea being that <em>if</em>
+ * we have a conversion to string and the source type is a {@code Number} and the mapping
+ * has a {@link org.mapstruct.Mapping#numberFormat()} we will replace the standard conversion
+ * with a {@link DecimalFormat} version
+ *
+ * @author Ciaran Liedeman
+ */
+public abstract class PossibleNumberToStringConversion extends SimpleConversion {
```
gunnarmorling: So why is it a *possible* number-to-string conversion? In which case wouldn't be such one?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -302,6 +302,7 @@ else if ( mapping.getSourceName() != null ) {
                                     .qualifiers( mapping.getQualifiers() )
                                     .resultType( mapping.getResultType() )
                                     .dateFormat( mapping.getDateFormat() )
+                                    .numberFormat( mapping != null ? mapping.getNumberFormat() : null)
```
gunnarmorling: The null check isn't really needed here. If it was null, an NPE would have been raised before.

processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java
```diff
@@ -60,6 +65,19 @@ private NativeTypes() {
         tmp.put( char.class, Character.class );
 
         PRIMITIVE_TO_WRAPPER_TYPES = Collections.unmodifiableMap( tmp );
+
+        NUMBER_TYPES.add( short.class );
```
gunnarmorling: `byte` and `Byte` are missing.

processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java
```diff
@@ -77,4 +95,16 @@ private NativeTypes() {
...
+
+    public static boolean isNumber(Class<?> clazz) {
+        if ( clazz == null ) {
+            return false;
+        }
+        else if ( clazz.isPrimitive() ) {
+            return NUMBER_TYPES.contains( clazz );
+        }
+        else {
+            return Number.class.isAssignableFrom( clazz );
```
gunnarmorling: These seem to be covered by `NUMBER_TYPES` already?

processor/src/test/java/org/mapstruct/ap/test/conversion/numbers/SourceTargetMapper.java
```diff
@@ -0,0 +1,51 @@
...
+    @Mappings( {
+            @Mapping( target = "i", numberFormat = NUMBER_FORMAT ),
+            @Mapping( target = "ii", numberFormat = NUMBER_FORMAT ),
+            @Mapping( target = "d", numberFormat = NUMBER_FORMAT ),
+            @Mapping( target = "dd", numberFormat = NUMBER_FORMAT ),
+            @Mapping( target = "l", numberFormat = NUMBER_FORMAT ),
+            @Mapping( target = "ll", numberFormat = NUMBER_FORMAT ),
+            @Mapping( target = "complex1", numberFormat = "##0.##E0" ),
+            @Mapping( target = "complex2", numberFormat = "$#.00" ),
+            @Mapping( target = "bigDecimal1", numberFormat = "#0.#E0" ),
```
gunnarmorling: What happens if an invalid format expression is specified? Are they validated at some point? It would be very nice to flag these at build time rather than when applying the formatting at runtime.

gunnarmorling:
@cliedeman Finally got to checking this one out; Thanks a lot, it looks very good overall, just added some (minor) remarks here and there.

When addressing these, could you also rebase to latest master and squash everything into one commit. Good to go otherwise.

Thanks again, very nice to have this one in.

gunnarmorling:
@cliedeman, looking at the generated code, I for String -> int see:

    Integer.parseInt( new DecimalFormat( "##.00" ).parse( target.getI() ).toString() )

Is there a reason that it isn't the following instead:

    new DecimalFormat( "##.00" ).parse( target.getI() ).intValue()

?

processor/src/main/java/org/mapstruct/ap/internal/conversion/PossibleNumberToStringConversion.java
cliedeman: In the cases of:
Boolean -> String (WrapperToStringConversion)
boolean -> String (PrimitiveToStringConversion)

processor/src/main/java/org/mapstruct/ap/internal/conversion/PrimitiveToStringConversion.java
cliedeman: Ill change it back

processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java
cliedeman: Yes, the current code is too relaxed and will accept other subclasses of number

processor/src/test/java/org/mapstruct/ap/test/conversion/numbers/SourceTargetMapper.java
cliedeman: I could 'test' them at some point during the mapping. I am a fan of the fail early approach. Where would this be appropriate? BeanMappingMethod?

core-common/src/main/java/org/mapstruct/IterableMapping.java
cliedeman: The doc definitely needs to be improved, primitive is misleading though as it includes char

cliedeman:
@gunnarmorling Hi there, I took a peek inside the DecimalFormat code. Looks like it always returns either a double or long. This is a problem because we can lose precision. 

E.g: 
```System.out.println(f.parse( "0.1000000000000000055511151231257827021181583404541015625" )); // prints 0.1```

I need to enable DecialFormat#setParseBigDecimal. How would I go about creating a local decimal format and configuring? Or could I create a factory method that creates a decimal formatter and invokes parse after setting parse big decimal to true?

Still outstanding: 
error on bad numberFormats during compile
Update @mapping javadoc

core-common/src/main/java/org/mapstruct/IterableMapping.java
gunnarmorling: Yes, but together with "numeric" it's clear.

processor/src/main/java/org/mapstruct/ap/internal/conversion/PossibleNumberToStringConversion.java
gunnarmorling: I see. Maybe we should separate boolean from the numeric ones then? Not strictly needed, but a bit cleaner IMO.

processor/src/test/java/org/mapstruct/ap/test/conversion/numbers/SourceTargetMapper.java
gunnarmorling: `Mapping.fromMappingPrism(MappingPrism, ExecutableElement, FormattingMessager)` would be a good place, we have some other validations for annotation values there already.

gunnarmorling:
> I need to enable DecialFormat#setParseBigDecimal. How would I go about creating a local decimal format and configuring?

You could do that through a private method you generate and then invoke. We do that already in some other cases, check out `ForgedMethod` and `NestedPropertyMappingMethod`.

gunnarmorling:
Hey @cliedeman, replied to some questions in the outdated diffs.

agudian:
>> I need to enable DecialFormat#setParseBigDecimal. How would I go about creating a local decimal format and configuring?
> 
> You could do that through a private method you generate and then invoke. We do that already in some other cases, check out ForgedMethod and NestedPropertyMappingMethod.

Wouldn't this be more like a `BuiltInMethod`?

gunnarmorling:
Yes, true indeed :)

gunnarmorling:
Hey @cliedeman, I was wondering whether you still plan to work on this and finish it up?

gunnarmorling:
@cliedeman, do you think you'll find some time to finish this one? If not, that's perfectly fine, I can take over then. Just let me know, so I know whether I should jump on it or not. Thanks!

cliedeman:
@gunnarmorling

Hi sorry for the late reply, please do jump in I can't commit any time to this pull request like now but would definitely like to see it through. 

Last thing I was looking at was moving to the 'forge' methods. I think it requires quite a lot of rework from what I had originally done.

gunnarmorling:
@sjaakd will try to take over this one :-) Thanks a lot!

sjaakd:
> Wouldn't this be more like a BuiltInMethod?

I'm looking at this one and filled in the gaps. But... If we truly want to go ahead with this we have need to rewrite all conversions that involve primitive (number) and number types to builtin methods.. That's a whole lot of more work than I bargained for.. Alternatively I could include a builtin method as factory method of a `DecimalFormatter` if a number pattern is included and call that somehow from the conversion...

BTW: I then would also opt to include the 'RoundingMode' as parameter on `@Mapping` (and others)

sjaakd:
> Wouldn't this be more like a BuiltInMethod?

Its neither a BuiltInMethod or ForgeMethods. Both are mapping methods. Or, we must do the BigDecimalToString BitIntegerToString as separate (conditional: based on presence of number format) mapping methods altogether... Perhaps this is the right approach.

Creating a `DecimalFormatter` with the right initializers called from another mapping is more like a helper method which is called from a conversion.

sjaakd:
Is superseded by #811

Issue #701, opened 2015-11-25
Update checkstyle version used in build


agudian:
The current version misses some of the checks and we have plenty of false-negatives. I stumbled accross that with the newer version that is used in my Eclipse installation, showing me quite a lot style issues.

I'll prepare a PR to update checkstyle and fix the issues.

PR #716, opened 2016-01-02
#701 Update checkstyle version and fix several previously undetected issues

agudian:
Update checkstyle version and fix several previously undetected issues in our code (mostly missing whitespaces and redundant modifiers) and in the generated code (superflous whitespaces before commas)

Requrires update of guava and cdi-weld do to dependency convergence.


agudian:
Thanks for taking a look, Sjaak! :)

Issue #669, opened 2015-10-23
Add support for hasX methods to check if a source property X should be set


spoerri:
Please support mapping to and from instances of message classes generated by Google Protocol Buffers (version 2).

In addition to simple getters for primitive fields, these classes have methods that indicate whether the field is set on the message at all.

If the field is not set of the source object (protobuff), we should refrain from setting it on the destination object.

E.g. protobuf generated code:
public double getPrice();
public boolean hasPrice();

And on a separate builder object, just:
public void setPrice(double price);

We should generate:
if (carMessage.hasPrice()) {
 car.setPrice(carMessage.getPrice());
}

And:
if (car.hasPrice()) {
 carMessage.setPrice(car.getPrice());
}

matt-blanchette:
A similar use case is the Guava/JDK8 Optional types, with isPresent() boolean check before calling setter (and an extra .get() call for the value).

agudian:
Hi @spoerri,

interesting idea! Adding support for `hasX()` methods is certainly something that we can do.

Calling that "support for protocol buffers" might be a bit too far fetched, though. Looking at the Docs, it seems that they suggest using builder classes for the instance creation, and that might just be a whole different story - although we already support stuff like `addX(X x)` for collections, so mapping into builder instances would probably work.

Anyway, let's do this step by step. Would it be OK for you if we changed the title of this issue to something like "Add support for hasX methods to check if a source property X should be set"?

You'd then probably hit the same problem that the feature request #649 addresses.

agudian:
> A similar use case is the Guava/JDK8 Optional types, with isPresent() boolean check before calling setter (and an extra .get() call for the value).

I'd see `Optional` as a slightly different case (because of the get() and the creation semantics), but one that we definitely should support. I'm creating a new issue for that.

gunnarmorling:
@agudian, did you ever create that issue for `Optional`?

sjaakd:
fixed and implemented by means of SPI and annotation

Spaum145:
Is this hasX method still supported? Can't make it work in 1.2.0.Final.

filiphr:
@Spaum145 yes this should still be supported. As this is quite an old issue can you please create a new issue with your problem. Maybe a reproducal case would be ideal, as we have tests for this and it works

Spaum145:
[...]

Issue #649, opened 2015-09-29
Update property in MappingTarget only if source property is not null


cbornet:
I would like to update target attributes only if the source ones are not null (to implement some kind of partial update). Is there a way to do that ? If not, could it be implemented, for instance as an option of MappingTarget ?
What I get

``` java
 @Override
 public void updateFooFromDto(FooDto fooDto, Foo foo) {
        if ( fooDto== null ) {
            return;
        }

        site.setBar( fooDto.getBar() );
        site.setBaz( fooDto.getBaz() );
}
```

What I would like to generate:

``` java
 @Override
 public void updateFooFromNonNullAttributesOfDto(FooDto fooDto, Foo foo) {
        if ( fooDto== null ) {
            return;
        }
        if ( fooDto.getBar() != null ) {
            site.setBar( fooDto.getBar() );
        }
         if ( fooDto.getBaz() != null ) {
            site.setBar( fooDto.getBaz() );
        }
}
```

gunnarmorling:
Actually your suggested code is what you should get if `bar` and `baz` on the source side are nullable types. Are you working with primitive types there by any chance?

cbornet:
They are primitive's wrapper types such as Integer, String, Boolean, ...

agudian:
That's not supported, yet. A similar question was asked on the mailing-list today and also in the past already.
So it seems there is a need for this feature. :)

What we need is some idea on how a user would configure this.

PierrickPuimeanChieze:
There was some suggestions on the issue #369

For my part, I think that a new Null strategy (named for example IGNORE) would be the best.

matt-blanchette:
+1

matt-blanchette:
Agreed that a new NullValueMappingStrategy like IGNORE would work better for me as well.

The discussion in #369 mentions changing @Mapping, which could work,
but this involves adding @Mapping for those automatically mapped by name.

rsharifu:
I completely agree. I'm going to jump in and use this library in my project when IGNORE is released.

jannikweichert:
I'd highly appreciate a solution for this as well. A new NullValueMappingStrategy seems to be promising approach

bhdrk:
+1

sjaakd:
I added an [example](https://github.com/sjaakd/mapstruct-examples) PR. Please have a look at the current MapStruct behavior. I'm not sure I understand what is actually desired in this issue.

bhdrk:
[...]

PR #720, opened 2016-01-14
#669 Add support for hasX methods to check if a source property X sho…

seanjob:


agudian:

agudian:
Wow, that's a big pull-request, @seanjob. Awesome! Thanks a lot for your efforts!

We'll be looking into it in more detail later, but I'd like to start discussing the feature and the API changes first.

I especially wonder if we should make checking for `hasX` configurable - my first gut feeling would have been to just add the check if the method is present (and has no arguments and returns `boolean`). @gunnarmorling and @sjaakd probably have an opinion here as well :wink:... If we leave it configurable, then I would suggest that we think about the actual way again, i.e. the property name in the annotation and the type of the property (does it have to be a strategy enum / could it be just a boolean flag / could we combine it with nullValueMappingStrategy, as it relates to a similar problem-scope).

Another point to discuss would be the extension of the SPI API (which is required in this case). I wonder if "hasser" is the optimal term there. I've seen frameworks that use a format like `isXSet` or `isXPresent`. Perhaps we can come up with something that's more natural to read. "availability checker"? I don't know.

The implementation itself looks fine and we'll definitely go forward with this. However, we should take the time to consider the user-facing parts in a larger group here first.

seanjob:
Thanks for the quick response.

When @spoerri brought up this idea, we first tried to implement it by detecting the hasX method automatically rather than explicitly configuring. And then we realized, it is error prone - if the method name has been misspelled, there is no way we can catch it.

The new annotation is on both Mapper and Mapping, I need to define three states - just like Boolean, so we know whether we can override the state on Mapping by the one on Mapper (in some cases, we don't want to do hasX check for all properties). It seems like we don't support Boolean in annotation, so we use enum.

I am not familiar with nullValueMappingStrategy. I can try to combine them if you believe it is good idea.

I agree with you,  the "hasser" is not very clear. I use it but because I don't have a better name. :)

I see people using "hasX" or "isXPresent" in real life. Whatever they are using, they can define their own name in AccessorNamingStratege.

gunnarmorling:
Hi @seanjob, many thanks for coming up with this! I like the proposal in general, but I'd like us to think about the API a bit more.

Taking a step back, it's a more general issue how we check whether source bean properties "are set" or not, and it'd be great if we could have one option for controlling this. Currently we do the following:

* For "built-in conversions" (i.e. things happening in-place, such as calling `String#valueOf(int)`) we check != null first if the source type is a non-primitive one
* For conversions invoking another method we always pass in the value, letting that method itself decide how to deal with a null input; There was just recently a question on the mailing list asking for doing the null check without calling the method

Calling `hasXy()` basically is just another flavor of that. So maybe we can find one more generalized setting to cover all the aspects. Something like this maybe:

    @Mapper(valueSetCheckStrategy=IS_NULL)
    public class MyMapper { ... }

And `ValueSetCheckStrategy` (or however we end up naming it) with the following values:

    IS_NULL (always check != null, no matter whether it's an inline or method conversion)
    IS_NULL_INLINE (the current behavior, check != null for inline conversions)
    CUSTOM (invoke custom hasXy() method, name to be given through the accessor naming strategy)

Any thoughts?

seanjob:
Hi @gunnarmorling,

Thanks for your comment.

I think it is a good idea to have more general annotation, like ValueSetCheckStrategy. 

I will change my PR to use ValueSetCheckStrategy. However, I am not sure how this new annotation will affect existing implementation. Could you point out, currently where we are doing these null checks?

Is everybody fine with the new wrapper - HasCheckWrapper?

gunnarmorling:
> Could you point out, currently where we are doing these null checks?

I think `NullCheckWrapper` is the starting point. You could check out where it's referenced and only apply if the value set check strategy mandates it. Otherwise you'd do the has check.

> Is everybody fine with the new wrapper - HasCheckWrapper? 

Sounds good to me.

gunnarmorling:
Hey @seanjob, how is it going with this issue? Do you still plan to work on this? Any help needed?

seanjob:
Hi @gunnarmorling,

I made the changes you suggested on Jan 19. Do yo have chance to review it?

I had comment in my commit, for "IS_NULL" strategy, I didn't do anything. 

Please give me some hints if you believe "IS_NULL" has to be implemented and supported.

I will resolve the conflicts once you finish review.

Thanks,

seanjob:
Hi @gunnarmorling,

I will be free tomorrow and out of town for a couple of days.

If I can have some feedback by tomorrow, I probably can wrap up this pull request before I leave.

Thanks,

gunnarmorling:
Ok, cool. I'll try and look into it later tonight.

processor/src/test/java/org/mapstruct/ap/test/hascheck/HasCheckTest.java
```diff
@@ -0,0 +1,113 @@
...
+    CustomMapper.class,
+    MyLongWrapper.class,
+    Source.class,
+    Target.class,
+    SourceWtCheck.class,
+    TargetWtCheck.class
+})
+@RunWith(AnnotationProcessorTestRunner.class)
+public class HasCheckTest {
+    @Test //Can't test seeing exception if config is on, as there is compilation error
```
gunnarmorling: What compilation error is this? Generally we should prevent to generate non-compiling code (if that's what's happening) and rather raise a meaningful error from within the generator.

processor/src/test/java/org/mapstruct/ap/test/hascheck/HasCheckTest.java
```diff
@@ -0,0 +1,113 @@
...
+        source.setSomeList( list );
+
+        Target target = SourceTargetMapper.INSTANCE.sourceToTarget( source );
+
+        Assert.assertEquals( object, target.getSomeObject() );
+        Assert.assertEquals( 5.0, target.getSomeDouble(), 0.01 );
+        Assert.assertEquals( (Integer) 7, target.getSomeInteger() );
+        Assert.assertEquals( longWrapper.getMyLong(), target.getSomeLong().getMyLong() );
+
+        for (int i = 0; i < list.size(); i++) {
```
gunnarmorling: We have FEST Assert on the class path which allows for much nicer assertions:

    assertThat( target.getSomeList() ).containsExactly( "first", "second" );

processor/src/main/java/org/mapstruct/ap/spi/MethodType.java
```diff
@@ -43,5 +43,10 @@
     /**
      * Any method which is neither a JavaBeans getter, setter nor an adder method.
      */
-    OTHER;
+    OTHER,
+
+    /**
+     * A JavaBeans hasser method, e.g. {@code public String hasName()}.
+     */
+    HASSER;
```
gunnarmorling: I'm still no fan of "HASSER" :) How about "PRESENCE_CHECK" or "IS_VALUE_SET_CHECKER" or so?

Also better not refer to JavaBeans as there is no such concept in the spec.

gunnarmorling:
> I had comment in my commit, for "IS_NULL" strategy, I didn't do anything.
> Please give me some hints if you believe "IS_NULL" has to be implemented and supported.

Yes, I think this should be implemented. E.g. consider the following example where two properties are mapped:

    Target someGeneratedMappingMethod(Source source) {
        Target target = ...;

        // inline conversion of String to Integer
        if ( source.getString != null ) {
            target.setInteger( Integer.valueOf( source.getString() ) );
        }

        // conversion via other mapping method
        target.setId( otherMapper.idToString( source.getId() );
    }

I.e. for inline conversions the generated code will contain a null check (if the source type is non-primitive), whereas for conversions through another mapping method *no* null check is done, leaving the handling of null to that other method. That is mostly a good default, but sometimes it's desirable to not invoke the target setter at all (JAXB being an example, as recently [brought](https://groups.google.com/forum/#!topic/mapstruct-users/ndKl7K9wmDk) up on mapstruct-users).

So if the strategy is set to `IS_NULL`, the following code should be generated instead:

    if ( source.getId() != null ) {
        target.setId( otherMapper.idToString( source.getId() );
    }

Does that clarify things?

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -148,4 +148,12 @@
...
      */
     MappingInheritanceStrategy mappingInheritanceStrategy() default MappingInheritanceStrategy.EXPLICIT;
+
+    /**
+     * Decide whether we should check null or hasX method before mapping.
+     * It is global, can only override the one on {@link Mapping} if it is UNDEFINED.
+     *
+     * @return strategy about how to do null or has value check
+     */
+    ValueSetCheckStrategy valueSetCheckStrategy() default ValueSetCheckStrategy.IS_NULL_INLINE;
```
gunnarmorling: I am still trying to find a better name. How about `SourceValuePresenceCheckStrategy`? Not exactly the shortest, but maybe better to reason about?

core-common/src/main/java/org/mapstruct/MapperConfig.java
```diff
@@ -135,4 +135,12 @@
      */
     MappingInheritanceStrategy mappingInheritanceStrategy()
         default MappingInheritanceStrategy.EXPLICIT;
+
+    /**
+     * Decide whether we should check null or hasX method before mapping.
+     * It is global, can only override the one on {@link Mapping} if it is UNDEFINED.
```
gunnarmorling: I am a bit confused by this message; How is the order of precedence? I'd expect the value on `@Mapping` take precedence over `@Mapper` over `@MapperConfig`. Is that what happens?

core-common/src/main/java/org/mapstruct/ValueSetCheckStrategy.java
```diff
@@ -0,0 +1,52 @@
...
+     * name to be given through the accessor naming strategy
+     */
+    CUSTOM,
+
+    /**
+     * Undefined, default value in {@link Mapping}, so it can be overridden
+     * by the value defined in {@link Mapper}
+     *
+     */
+    UNDEFINED;
```
gunnarmorling: We don't need that one. We can find out whether a value is actually set on any of the three annotations and thake the value from the most specific one (i.e. we can distinguish whether the annotation value `IS_NULL_INLINE` is actually set in a given annotation instance or whether `IS_NULL_INLINE` is just retrieved as it's the annotation's default value. 

Just check out how it's done for the other annotation values from `@Mapper`/`@MapperConfig`, the prisms allow to make that distinction.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -462,6 +482,50 @@ else if ( propertyEntries.size() == 1 ) {
...
+
+            switch ( valueSetCheckStrategy ) {
+                case IS_NULL_INLINE:
+                    return createNullCheckWrapper( assignment, isUpdate );
+
+                case CUSTOM:
+                    String hasMethod = this.getSourceHasMethod();
+
+                    if ( hasMethod == null ) {
+                        throw new IllegalArgumentException( sourceReference.getParameter().getName()
```
gunnarmorling: Better not throw an exception from the AP environment. Instead we should create actual compilation errors which are gracefully handled by the processing infrastructure. That way the errors are also linked to the line number of the flawed mapping method.

You can find examples in this class, e.g. like so:

    ctx.getMessager().printMessage( method.getExecutable(),
        Message.SOME_KEY,
        targetPropertyName
    );

Instead of "hasX()",  the actual name of the value presence check method would be nice to read:

> Using custom source value presence checking strategy, but no method 'boolean hasMyProperty()` was found.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/HasCheckWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+
+import org.mapstruct.ap.internal.model.common.Type;
+
+/**
+ * Wraps the assignment in a target setter if CheckHas is set to true.
+ *
+ * @author Sean Huang
+ */
+public class HasCheckWrapper extends AssignmentWrapper {
```
gunnarmorling: Could you add to the JavaDoc the snippet of code it generates (or, if feasible, do it via the `toString()`) method; Esp. to make the difference to `UpdateHasCheckWrapper` more apparent.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/HasCheckWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+        this.sourceHasMethod = sourceHasMethod;
+    }
+
+    @Override
+    public List<Type> getThrownTypes() {
+        List<Type> parentThrownTypes = super.getThrownTypes();
+        List<Type> result = new ArrayList<Type>( parentThrownTypes );
+        for ( Type thrownTypeToExclude : thrownTypesToExclude ) {
+            for ( Type parentThrownType : parentThrownTypes ) {
+                if ( parentThrownType.isAssignableTo( thrownTypeToExclude ) ) {
```
gunnarmorling: Why are these excluded here?

processor/src/test/java/org/mapstruct/ap/test/hascheck/SourceDoesNotHave.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package org.mapstruct.ap.test.hascheck;
+
+
+/**
+ * @author Sean Huang
+ */
+public class SourceDoesNotHave extends Source {
+
+    private int noPrimitive;
```
gunnarmorling: Why "noPrimitive", it is primitive, no ;)

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -462,6 +482,50 @@ else if ( propertyEntries.size() == 1 ) {
...
+                            + " does not have 'hasX()' method." );
+                    }
+                    return isUpdate ? new UpdateHasCheckWrapper( assignment, method.getThrownTypes(), hasMethod )
+                        : new HasCheckWrapper( assignment, method.getThrownTypes(), hasMethod );
+
+                case IS_NULL:
```
gunnarmorling: I suppose you'd just have to handle `IS_NULL` like `IS_NULL_INLINE`?

gunnarmorling:
@seanjob Done with the review. Excellent work, looking forward to integrating this feature!

I've added some remarks, but nothing special, mostly cosmetics/naming. The only major thing missing is the handling for `IS_NULL`.

Btw. sorry for the long radio silence, didn't notice you pushed an update. Always better add a short comment when pushing more commits.

seanjob:
@gunnarmorling,

Thanks for reviewing.

Sorry, I have been distracted by some unexpected things these days. I will finish the changes once I  become free again.

I use Stash, didn't know Github won't send update automatically. Will add comment when pushing next time.

Sorry again.

gunnarmorling:
Hey @seanjob, no need to be sorry. Your contribution and help is much appreciated, take the time you need. Looking forward to integrating this change :)

processor/src/test/java/org/mapstruct/ap/test/hascheck/HasCheckTest.java
seanjob: @gunnarmorling 

I think the exception I see is a good one to "prevent to generate non-compiling code".

org.junit.ComparisonFailure: [Compilation failed. Diagnostics: [DiagnosticDescriptor: ERROR SourceTargetHasCheckMapper.java:31 Internal error in the mapping processor: java.lang.IllegalArgumentException: source does not have 'hasX()' method.  	at org.mapstruct.ap.internal.model.PropertyMapping$PropertyMappingBuilder.getOrcreateWrapperByValueCheckStrategy(PropertyMapping.java:499)  	at org.mapstruct.ap.internal.model.PropertyMapping$PropertyMappingBuilder.assignObject(PropertyMapping.java:326)  	at org.mapstruct.ap.internal.model.PropertyMapping$PropertyMappingBuilder.build(PropertyMapping.java:236)  	at org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.applyPropertyNameBasedMapping(BeanMappingMethod.java:430)  	at org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.build(BeanMappingMethod.java:129)  	at org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMappingMethods(MapperCreationProcessor.java:361)  	at org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMapper(MapperCreationProcessor.java:143)  	at org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:114)  	at org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:1)  	at org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:271)  	at org.mapstruct.ap.MappingProcessor.processMapperTypeElement(MappingProcessor.java:251)  	at org.mapstruct.ap.MappingProcessor.processMapperElements(MappingProcessor.java:217)  	at org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:156)  	at com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:794)  	at com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:705)  	at com.sun.tools.javac.processing.JavacProcessingEnvironment.access$1800(JavacProcessingEnvironment.java:91)  	at com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1035)  	at com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1176)  	at com.sun.tools.javac.main.JavaCompiler.processAnnotations(JavaCompiler.java:1170)  	at com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:856)  	at com.sun.tools.javac.main.Main.compile(Main.java:523)  	at com.sun.tools.javac.api.JavacTaskImpl.doCall(JavacTaskImpl.java:129)  	at com.sun.tools.javac.api.JavacTaskImpl.call(JavacTaskImpl.java:138)  	at org.mapstruct.ap.testutil.runner.JdkCompilingStatement.compileWithSpecificCompiler(JdkCompilingStatement.java:85)  	at org.mapstruct.ap.testutil.runner.CompilingStatement.compile(CompilingStatement.java:373)  	at org.mapstruct.ap.testutil.runner.CompilingStatement.generateMapperImplementation(CompilingStatement.java:159)  	at org.mapstruct.ap.testutil.runner.CompilingStatement.evaluate(CompilingStatement.java:93)  	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)  	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)  	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)  	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)  	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)  	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)  	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)  	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)  	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)  	at org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner.runChild(AnnotationProcessorTestRunner.java:80)  	at org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner.runChild(AnnotationProcessorTestRunner.java:52)  	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)  	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)  	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)  	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)  	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)  	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)  	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:50)  	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)  	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:459)  	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:675)  	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:382)  	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:192)  ]] expected:<[SUCCEED]ED> but was:<[FAIL]ED>
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:422)
	at org.fest.assertions.ConstructorInvoker.newInstance(ConstructorInvoker.java:36)
	at org.fest.assertions.ComparisonFailureFactory.newComparisonFailure(ComparisonFailureFactory.java:60)
	at org.fest.assertions.ComparisonFailureFactory.comparisonFailure(ComparisonFailureFactory.java:46)
	at org.fest.assertions.Fail.comparisonFailed(Fail.java:83)
	at org.fest.assertions.Fail.failIfNotEqual(Fail.java:71)
	at org.fest.assertions.GenericAssert.isEqualTo(GenericAssert.java:217)
	at org.mapstruct.ap.testutil.runner.CompilingStatement.generateMapperImplementation(CompilingStatement.java:169)
	at org.mapstruct.ap.testutil.runner.CompilingStatement.evaluate(CompilingStatement.java:93)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner.runChild(AnnotationProcessorTestRunner.java:80)
	at org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner.runChild(AnnotationProcessorTestRunner.java:52)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:50)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:459)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:675)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:382)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:192)

processor/src/main/java/org/mapstruct/ap/spi/MethodType.java
seanjob: How about PRESENCE_CHECKER?

core-common/src/main/java/org/mapstruct/MapperConfig.java
seanjob: yes

core-common/src/main/java/org/mapstruct/ValueSetCheckStrategy.java
seanjob: But what if we have CUSTOM on Mapper and defaultValue=IS_NULL_INLINE on Mapping, which one the prisms will pick?

Do we have same property on both Mapper/MapperConfig and Mapping before?

gunnarmorling:
@seanjob Any news on this front? Do you still feel like finishing up this issue?

seanjob:
@gunnarmorling, Just become free today, will work on it tomorrow. May have some questions regarding your comments.

seanjob:
Hi @gunnarmorling,

I just finished. Could you take a look?

Thanks,

gunnarmorling:
Thanks, @seanjob! I'll take a look and report back.

Walker555:
Can you please integrate this feature ASAP. Thank you very much.

gunnarmorling:
Yes, sorry about the delay, I was very busy with other things recently.
Going to look into asap.

spoerri:
Any news?

ydemartino:
I am also really looking forward to see this feature integrated.

seanjob:
@gunnarmorling 

Do you have any concern? Please let me know if we can make this pull request better.

Thanks,

bfg:
+1

gunnarmorling:
Hey @seanjob, finally getting back to this. Sorry again for the delay. The team was a bit distracted by other stuff recently. We'll finally review it in the next few days and hopefully get to merging than. Thanks for hanging in!

core-jdk8/src/main/java/org/mapstruct/Mapping.java
```diff
@@ -180,4 +182,12 @@
...
      */
     String defaultValue() default "";
+
+    /**
+     * Decide whether we should check null or hasX method before mapping.
+     * The value on {@link Mapper} can override this one.
+     *
+     * @return strategy about how to do null or has value check
+     */
+    SourceValuePresenceCheckStrategy sourceValuePresenceCheckStrategy() default IS_NULL;
```
sjaakd: is this in line with the current way of handling? I thought its is_null_in_line. I would not prefer changed default behavior.

core-common/src/main/java/org/mapstruct/SourceValuePresenceCheckStrategy.java
```diff
@@ -0,0 +1,45 @@
...
+     * Always check != null, no matter whether it's an inline or method conversion
+     *
+     */
+    IS_NULL,
+
+    /**
+     * Will invoke custom hasX() method, before mapping,
+     * name to be given through the accessor naming strategy
+     */
+    CUSTOM;
```
sjaakd: wat about custom_null_inline?

core-common/src/main/java/org/mapstruct/SourceValuePresenceCheckStrategy.java
```diff
@@ -0,0 +1,45 @@
...
+    IS_NULL_INLINE,
+
+    /**
+     * Always check != null, no matter whether it's an inline or method conversion
+     *
+     */
+    IS_NULL,
+
+    /**
+     * Will invoke custom hasX() method, before mapping,
```
sjaakd: what's exactly custom to hasX? can you configure the name (e.g. isSetX)?

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -501,6 +497,62 @@ else if ( propertyEntries.size() == 1 ) {
             }
         }
 
+        private Assignment getOrcreateWrapperByValueCheckStrategy(Assignment assignment, boolean isUpdate) {
```
sjaakd: Can we get a more meaningfull name?

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/Assignment.java
```diff
@@ -42,7 +42,20 @@
...
         MAPPED_TYPE_CONVERTED,
         /** assignment is first type converted, and then mapped (builtin/custom) */
-        TYPE_CONVERTED_MAPPED
+        TYPE_CONVERTED_MAPPED;
+
+        public boolean isDirect() {
+            return this == DIRECT;
+        }
+
+        public boolean isTypeRelated() {
```
sjaakd: What do this booleans signify? `isTypeRelated` and `isMapRelated` both concern mapping and conversion. Its just the order in which they are called that matters. So if you map a method first, you allow null (no npe check) because the implementation of the called mapping method should handle null (it could also have a meaning).

The two introduced methods obfuscate the original intention of these enum constant. Especially if you consider that `isMapRelated` could also be related to map's (they require special treatments as well in some places).

processor/src/test/java/org/mapstruct/ap/test/presencecheck/MyLongWrapper.java
```diff
@@ -0,0 +1,38 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.presencecheck;
+
+
+/**
+ * @author Sjaak Derksen
```
sjaakd: ? hmm. Can't remember this one :smile:

processor/src/test/java/org/mapstruct/ap/test/presencecheck/SourceTargetMapper.java
```diff
@@ -0,0 +1,82 @@
...
+     * expected:<[SUCCEED]ED> but was:<[FAIL]ED>
+     *
+    @Mappings( {
+        @Mapping(target = "someDouble", defaultValue = "111.1"),
+        @Mapping(target = "someInteger", defaultValue = "222"),
+        @Mapping(target = "someLong", defaultValue = "333"),
+    } )
+    TargetWtCheck sourceToTargetWithConfigOn(SourceWtCheck source);
+    */
+}
```
sjaakd: I reckon I miss some tests with the configuration hierarchy.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -384,26 +384,23 @@ private void applyPropertyNameBasedMapping() {
...
                             Mapping mapping = method.getSingleMappingByTargetPropertyName( targetProperty.getKey() );
                             DeclaredType sourceType = (DeclaredType) sourceParameter.getType().getTypeMirror();
 
                             SourceReference sourceRef = new SourceReference.BuilderFromProperty()
                                 .sourceParameter( sourceParameter )
-                                .type( ctx.getTypeFactory().getReturnType( sourceType, sourceAccessor ) )
-                                .accessor( sourceAccessor )
+                                .type( ctx.getTypeFactory().getReturnType( sourceType, sourceReadAccessor ) )
+                                .readAccessor( sourceReadAccessor )
+                                .presenceChecker( sourcePresenceCheckers )
```
sjaakd: sourcePresenceChecker's' (plural)?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -501,7 +525,8 @@ else if ( candidates.size() == 1 ) {
             }
             // can only be the case for Booleans: isFoo() and getFoo(); The latter is preferred then
             else if ( candidates.size() == 2 ) {
-                if ( candidates.get( 0 ).getSimpleName().toString().startsWith( "get" ) ) {
+                if ( candidates.get( 0 ).getSimpleName().toString().startsWith( "get" ) ||
+                    candidates.get( 0 ).getSimpleName().toString().startsWith( "has" )) {
```
sjaakd: hardCoded `has`?

sjaakd:
@seanjob : Thanks for the PR. I've provided some comments. My main concern is that the name of the 'hasser' is fixed. I reckon that should somehow be configurable (also from previous comments). Please have a look. I believe that also @gunnarmorling is still going to take a look.

sjaakd:
as an afterthought.. Perhaps its better to have a `String` here iso an `enum` for `SourceValuePresenceCheckStrategy`. It could have the values (constants predefined in the `MappingConstants`): `<IS_NULL_INLINE>`, `<IS_NULL>`, `<HASSER>=has`, `<IS_SET>=isSet`, and own custom values.. WDYT?

agudian:
Would the String-value actually be `"<HASSER>=has"`? Or would we find a form that would allow actual custom patterns, such as `has<PROPERTY>`, `is<PROPERTY>Set` or whatever, where `<PROPERTY>` would be replaced by the property name (capitalized if not beginning at the start of the String). That would be in line with our handling of the implementation name and package name.

sjaakd:
> Would the String-value actually be "<HASSER>=has"? Or would we find a form that would allow actual custom patterns, such as has<PROPERTY>, is<PROPERTY>Set or whatever, where <PROPERTY> would be replaced by the property name (capitalized if not beginning at the start of the String). That would be in line with our handling of the implementation name and package name.

👍 .. Indeed that would offer more flexibility.. Would you then 
1) opt for the solution were we have a `SourceValuePresenceCheckStrategy` (`<IS_NULL_INLINE>`, `<IS_NULL>`, default `<IS_NULL_INLINE>`) next to a `SourceValuePresenceCheckMethod` with `has<PROPERTY>, is<PROPERTY>Set (default empty)`. We could predefine some 'known' patterns in our constants.
2) try to fit it all in one string somehow..?

gunnarmorling:
Still need to load all the details back into memory :)

But IIRC, the idea was to have just one value for "hassers" in the source value presence strategy and expand `AccessorNamingStrategy` accordingly. I.e. one would have to implement that SPI based on the patterns one is using and e.g. return method type = HASSER if the method name is `hasXy()`. The right "hasser" method would be used for a given property if the source value presence strategy for it is "hasser".

Would that make sense, @sjaakd?

sjaakd:
> Would that make sense, @sjaakd?

Now I understand.. That indeed would make sense.

Remains the issue of the combinations: <IS_NULL_INLINE> with a <CUSTOM>

sjaakd:
And. a thing that crossed my mind last week: to get around all the issues with Lombok.. Could we add options that checks for the presence of a field but still generate code that calls the getter / setter that would be there at compilation time? So, to remain in line with the previous remark: `<LOMBOK_INLINE>`, `<LOMBOK>`

gunnarmorling:
WDYM by "<IS_NULL_INLINE> with a <CUSTOM>"?

Regarding

> Could we add options that checks for the presence of a field but still generate code that calls the getter / setter that would be there at compilation time?

Yes, I think by now that this is the best way forward in things Lombok. But does Lombok even support something like value presence methods? I suppose using plain getters would be enough there?

sjaakd:
> WDYM by " with a "?

Typo.. corrected my comment.. :smile:


> Yes, I think by now that this is the best way forward in things Lombok. But does Lombok even support something like value presence methods?

No.. but the thinking is: its a 'special' strategy. We don't check if there's a getter or setter. We just assume there is going to be one when Lombok has done its job. We just check for the presence of the field. For the remainder it will behave exactly the same as the normal `NULL` or `NULL_IN_LINE`

gunnarmorling:
> corrected my comment

Ah, yes, sorry, fell into the same trap with mine ;) Could you give an example of what you mean here?

agudian:
I think it would be great if we left the term 'lombok' out of the API. If we go down that road to infer the eventual presence of getter/setter merhods based on Lombok annotations, it should be transparent to the user - perhaps kicking in as a fallback if no accessor methods are there, yet (depends on the setup of the project and the build).

On the topic, my feeling is that we are creating a lot of detailed choices here were we used to automatically pick the best one before (e.g. inline vs not inline). We shouldn't need to make this turing-complete ;-). If we can make the API twice as intuitive by not supporting some corner cases, we should stick to the simpler approach.
So perhaps the SourceValuePresenceStrategy might only contain NULL_CHECK (default) and PRESENCE_CHECK_METHOD (can for now detect hasX methods, or other patterns implemented via SPI). We shouldn't need to let users explicitly switch on the SPI usage using a special strategy enum - we don't do that for the accessor naming SPI either.

But I must also confess that I don't understand why the distinction between null-check and null-check-inline was required here in the first place. I'm sure you didn't just make that up just for the fun of it. :wink:

sjaakd:
> I think it would be great if we left the term 'lombok' out of the API. If we go down that road to infer the eventual presence of getter/setter merhods based on Lombok annotations, it should be transparent to the user - perhaps kicking in as a fallback if no accessor methods are there, yet (depends on the setup of the project and the build).

Perhaps we should park the Lombok discussion. 

The idea from my side was that for the `NULL_CHECK` or the `NULL_CHECK_INLINE` we could have 2 new alternatives: `NULL_CHECK_LOMBOK` or the `NULL_CHECK_INLINE_LOMBOK`. The only difference with the `NULL_CHECK` being that we don't check for the presence of getters and setters but only on the field, assuming that Lombok generates them.  Wy ' confuse' that with `SourceValuePresenceCheckStrategy` you might think? To answer that question: the Because (as @gunnarmorling pointed out), 'HASSER's are not generated on Lombok. And should we go for such strategy: then I do favor a name with 'LOMBOK' in an enum constant, because of clarity to the user. After all: this is the only use case for Mapstruct calling setters/getters whilst they are not there during annotation processing.   


> But I must also confess that I don't understand why the distinction between null-check and null-check-inline was required here in the first place. I'm sure you didn't just make that up just for the fun of it. 😉

Agreed. I'm not sure why its there either. AFAIK we agreed some time ago that null checking is up to the implementer of  mapping methods (builtin or handwritten). Reason: null is just a value that could have meaning as well. So as far as I'm concerned: we could skip the `NULL_CHECK_INLINE`.

gunnarmorling:
Agreed on a path forward with @agudian and @sjaakd in a call:

* Enhance `AccessorNamingStrategy` so it can return a new `MethodType#VALUE_PRESENT_CHECKER` from `getMethodType()` and the name of the property for such methods from `getPropertyName()`
* Implement the default naming strategy so it detects `hasXy()` methods; For other patterns (e.g. `isXySet()`), the user would need to provide an alternative strategy implementation
* If a value present checker method is available for a given source property, invoke it; Otherwise fall back to the current behavior
* The value present method needs to be invoked *always*, i.e. not only for inline conversions as the case with is-null checks today; Otherwise we'd dereference potentially unset values when invoking conversion methods

gunnarmorling:
@sjaakd, @agudian, reading the thread again, I found the motivation for the `IS_NULL` strategy:

> [...] for inline conversions the generated code will contain a null check (if the source type is non-primitive), whereas for conversions through another mapping method no null check is done, leaving the handling of null to that other method. That is mostly a good default, but sometimes it's desirable to not invoke the target setter at all (JAXB being an example, as recently brought up on mapstruct-users).
>
> So if the strategy is set to IS_NULL, the following code should be generated instead:
>
> if ( source.getId() != null ) {
>    target.setId( otherMapper.idToString( source.getId() );
> }

Based on that, I think having this strategy one way or another would make sense. But I suppose we can address it separately from the usage of `hasXy()` methods? It would only affect the case where no such method is available, and we then could add - separately - an option to control wether to do `IS_NULL`or `IS_NULL_INLINE` (current behavior). WDYT? If you agree, I'll open a separate issue for that one, so we can finish the support for `hasXy()` methods.

sjaakd:
> But I suppose we can address it separately from the usage of hasXy() methods

That's in principle ok (from an issue perspective). But it then feels a bit a waste to 'throw' the current code (that basically already does this) away. WDYT?

gunnarmorling:
Ah, I wasn't sure whether the IS_NULL behavior would be there already. If so, yes, then I suppose it'd make sense to make use of it. The new option would then only apply to the is-null case, whereas the hasXy() methods would be invoked if present?

sjaakd:
> Ah, I wasn't sure whether the IS_NULL behavior would be there already. If so, yes, then I suppose it'd make sense to make use of it. The new option would then only apply to the is-null case, whereas the hasXy() methods would be invoked if present?

Exactly..

sjaakd:
> Agreed. But then let's find a more suitable naming than "is_null" and "is_null_inline". Perhaps something that sematically reads like "don't set if source is null" or something?

perhaps we should combine it with the name.. (now sourceValuePresenceCheckStrategy).. 

doNotSetSourceWhenValueEquals
 - null
 - null_unless_mapped

gunnarmorling:
How about this:

    enum NullCheckStrategy {
        ALWAYS,
        ON_IMPLICIT_CONVERSION
    }

?

sjaakd:
@gunnarmorling Ok. And what would be the name of the property?

sjaakd:
When checking.. I found that the `SourcePresenceValueStrategy` is implemented on all levels. MapperConfig, Mapper and Mapping. The latter one does not make sense when the options are reduced to 
```
    enum NullCheckStrategy {
        ALWAYS,
        ON_IMPLICIT_CONVERSION
    }
```
WDYT?

Note: the SPI takes precedence if there's a hasser method found (according the SPI), then its used anyway.

agudian:
You're right, that feels strange. Plus, as it also covers what's described in #649 (update property only if source is not null), we might want to take that into account for the wording as well...

sjaakd:
superseded by #821

gunnarmorling:
Closing this one as it's superseded by #821.

Issue #729, opened 2016-01-24
Mapping Set<Long> to int throws NPE


greuelpirat:
When the source class has an Attribute `Set<Long> attr1` and the target class has the same attributename with a simple type like `int attr1`, a NullPointerException will be thrown at 

`org.mapstruct.ap.internal.processor.creation.MappingResolverImpl$ResolvingAttempt.hasCompatibleCopyConstructor(MappingResolverImpl.java:547)`

    sourceType = {org.mapstruct.ap.internal.model.common.Type@3381} "java.util.Set<java.lang.Long>"
    targetType = {org.mapstruct.ap.internal.model.common.Type@3382} "int"

gunnarmorling:
How comes you are closing it again? The mapping isn't supported (seems not really sensible to map a set into a single int value), but this shouldn't trigger an NPE but rather a meaningful error.

greuelpirat:
I could solved this with a workaround by adding the following mapping-method:

    default Integer sizeOfCollection(Collection<?> collection) {
        return collection == null ? 0 : collection.size();
    }

Sorry, I was a little confused, I will reopen it again.

gunnarmorling:
Yes, that makes sense. Maybe you'd like to help out and send in a PR for having MapStruct raise a more meaningful error in this case?

greuelpirat:
I will look for it. may take some time.

gunnarmorling:
Cool; Looking forward to your contribution. Just let me know if you need any help.

gunnarmorling:
Done.

PR #731, opened 2016-02-01
#729 Fix trying to resolving constructor on primitive types

greuelpirat:
I have prepared a fix for the NullPointerException. The fix itself was not that hard. But for an dedicated unit test I need some help. The test-framework is quite complicated.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
@@ -544,8 +544,8 @@ public class MappingResolverImpl implements MappingResolver {
          */
         private boolean hasCompatibleCopyConstructor(Type sourceType, Type targetType) {
             if ( targetType.isPrimitive() ) {
-				return false;
-			}
+                return false;
+            }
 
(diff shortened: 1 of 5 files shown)


agudian:

gunnarmorling:
@greuelpirat, thanks for looking into this!

Regarding a test, check out `org.mapstruct.ap.test.collection.erroneous.ErroneousCollectionMappingTest`, I think that's where it should go. Actually, `shouldFailToGenerateImplementationBetweenCollectionAndNonCollection()` seems to cover that case on first sight, but apparently something still is different.

In general, we have a custom JUnit runner, `AnnotationProcessorTestRunner` which takes a set of test sources (mappers, source/target beans etc.) and compiles them. Then one can either invoke the generated mapper and test its functionality by asserting the target object of the mapping (e.g. see `CollectionMappingTest`). Or one can assert the presence of specific compilation errors in order to make sure that MapStruct rightfully rejects specific wrong mappings. That's what you see in `ErroneousCollectionMappingTest`.

greuelpirat:
Thank for you help. I have added an unittest. Please check it out.

gunnarmorling:
Merged; Thanks, @greuelpirat!

Issue #634, opened 2015-08-24
Generics method does not match


bedla:
Hi,
I found that when you have mapper defined as 
```java
@Mapper
public interface SourceTargetMapper {
    Bar mapFooToBar(Foo foo);
    List<Bar> mapFooListToBarList(List<Foo> fooList);
//    TargetBar mapSourceToTarget(SourceFoo source);
    Target<Bar> mapSourceToTarget(Source<Foo> source);
}
```
And source and target classes as
```java
public class Source<T> {
    private final List<T> items;
    public Source(List<T> items) { this.items = items; }
    public List<T> getContent() { return items; }
}
public class Target<T> {
    private List<T> content;
    public List<T> getContent() { return content; }
    public void setContent(List<T> content) { this.content = content; }
}
```
Matching algorithm is not working in this case because of `getContent()` from Source returns unspecified type `List<T>` and not `List<Foo>`. I found it while investigating issue #607 .
When I specify parameter on on Source/Target class it works as expected (see commented line on mapper).
```java
public class SourceFoo extends Source<Foo> {
    public SourceFoo(List<Foo> items) { super( items ); }
}
public class TargetBar extends Target<Bar>{ }
```
From source code it lookls like it is more deeper in javax.lang.model.* implementation / I guess.

Ivos

sjaakd:
@bedla : have you checkout out the `org.mapstruct.ap.test.selection.generics.GenericTypeMapper` and associated tests?

antslava:
Any update on this issue?

gunnarmorling:
Hey @antslava, we didn't yet get to fixing this. If you need it quickly, you might consider to send in a PR for fixing it?

gunnarmorling:
Ok, I think I got an angle on this. Will send a PR hopefully shortly. Essentially, we are using (Type)Element in several places where we actually should use DeclaredType instead which conveys the information about the type parameter.

antslava:
@gunnarmorling, Thanks for your help.

gunnarmorling:
Hey @antslava, I just sent in PR https://github.com/mapstruct/mapstruct/pull/738. Maybe you could build MapStruct from that branch (just run "mvn clean install") and see whether it fixes the issue for you, too?

antslava:
Hi @gunnarmorling, thanks. I will try it today and write you about results.

antslava:
Hi @gunnarmorling , your PR works. Thanks for your help.

[...]

Issue #737, opened 2016-02-03
Ensure types given for @Mapper#uses() and config() are valid


gunnarmorling:

Captain1653:
You may use  [this](https://mapstruct.org/documentation/stable/reference/html/#controlling-nested-bean-mappings)

filiphr:
I think that this has been resolved by #2056

Issue #389, opened 2014-12-12
Target properties should support nesting (as in #65) too.


sjaakd:
See example below, by @ascuderetti

```java
CarDto
{
   Long idColor;
   BigDecimal price;
}

CarEntity
{
  ColorEntity color;
  BigDecimal price;
}

ColorEntity
{
  Long id;
  String valule;
}
```

(ps. Entity is the object mapping the db ER schema usually managed by a framework like Hibernate, Dto is the object mapping client requirements)

In a READ context Entity is the "source" and Dto the "target", we will have this mapping:

```java
source="color.id" target="idColor"
source="price" target="price"
```
...and your choice is ok.

In a WRITE context (Dto is the "source" and Entity is the "target"), we should have this mapping:

```java
source="idColor" target="color.id"
source="price" target="price"
```
...in this context the nesting on the target side is useful...!

A typical write context case is, for example, when a user inserts a new car. He selects a color from a select.

The http data will be mapped on the CarDto and the CarDto on the Entity with a code like this:

```java
CarEntity car = new CarEntity();

ColorEntity color = new ColorEntity();
color.setId(dto.getIdColor());
//...
//...
car.setColor(color);
//....
```
In this way a framework like Hibernate, with the right configuration, will process the object color with the id setted, nested in the CarEntity, like a foreign key of CarTable to ColorTable.

In my software project this is a common pattern.

---------

Also `@InheritInverseConfiguration` should reverse the nested source properties.

---------

Solution direction & further information: see #65.

gunnarmorling:
@sjaakd, @ascuderetti: I still think it is a good idea that each mapping method only deals with one object on the target side. How would we e.g. deal with the unreported target properties warning if a method could also be in charge of populating sub-properties in the graph? Also the selection of specific factory methods for the sub-nodes wouldn't work.

The given example could be implemented like so, using a manually written mapper class:

    @Mapper(uses=ColorMapper.class)
    public interface CarMapper {

        // generated implementation looks like this:
        // carEntity.setColor( colorMapper.getObjectById( dto.getIdColor(), ColorEntity.class ) )
        @Mapping(target="color", source="idColor")
        CarEntity carDtoToCarEntity(CarDto dto);
    }

    public class ColorMapper {
        <T> T getObjectById(long id, @TargetType Class<T> entityType) {
            // could get it from the database or cache or just return entityType.newInstance( id );
        }
    }

Alternatively, a simple expression could be used:

[...]

PR #738, opened 2016-02-03
#634 Resolve type parameters when processing source/target bean accessors

gunnarmorling:
Hey @sjaakd and @agudian, have a nice fix for that issue with generic source target types. 

The trick was to use `DeclaredType` instead of `TypeElement` when examining the accessors of source and target beans, from there it worked as expected. I am kind of amazed about this actually, seems our type/resolution system isn't so bad after all when correctly using it :)

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/generics/container/Bar.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/generics/container/Bar.java
@@ -1,0 +1,39 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.generics.container;
+
+public class Bar {
+
+    private long value;
+
+    public Bar() {
+    }
+
+    public Bar(long value) {
+        this.value = value;
+    }
+
+    public long getValue() {
+        return value;
+    }
+
+    public void setValue(long value) {
+        this.value = value;
+    }
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -33,7 +33,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 
 import javax.lang.model.element.ExecutableElement;
-import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeMirror;
 import javax.tools.Diagnostic;
 
@@ -407,8 +407,7 @@ public class BeanMappingMethod extends MappingMethod {
                         ExecutableElement sourceAccessor = getSourceAccessor( targetProperty.getKey(), candidates );
                         if ( sourceAccessor != null ) {
                             Mapping mapping = method.getSingleMappingByTargetPropertyName( targetProperty.getKey() );
-
-                            TypeElement sourceType = sourceParameter.getType().getTypeElement();
+                            DeclaredType sourceType = (DeclaredType) sourceParameter.getType().getTypeMirror();
 
                             SourceReference sourceRef = new SourceReference.BuilderFromProperty()
                                 .sourceParameter( sourceParameter )
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -73,14 +73,36 @@ public class PropertyMapping extends ModelElement {
     private final List<String> dependsOn;
     private final Assignment defaultValueAssignment;
 
+    private enum TargetWriteAccessorType {
+        GETTER,
+        SETTER,
+        ADDER;
+
+        public static TargetWriteAccessorType of(ExecutableElement accessor) {
+            if ( Executables.isSetterMethod( accessor ) ) {
+                return TargetWriteAccessorType.SETTER;
+            }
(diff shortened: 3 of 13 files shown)


sjaakd:
> Hey @sjaakd and @agudian, have a nice fix for that issue with generic source target types.

Hey @gunnarmorling , Brilliant!. I'll have a look this evening.

processor/src/test/java/org/mapstruct/ap/test/generics/container/Bar.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.generics.container;
+
+public class Bar {
```
agudian: Well, I just have to write this: "Didn't we agree to use more speaking examples in the tests instead of Source, Target, Foo, Bar?"... :wink:

sjaakd: > Well, I just have to write this: "Didn't we agree to use more speaking examples in the tests instead of Source, Target, Foo, Bar?"...

:smiling_imp:

agudian:
I'm done with the review and it looks all good. What I wondered was if we can somehow avoid all those unchecked casts to `DeclaredType`...?

Anyway, I'm going to play around with now, see if it solves any other stuff as well... :)

agudian:
> I'm done with the review and it looks all good. What I wondered was if we can somehow avoid all those unchecked casts to `DeclaredType`...?

Looks like adding a method `getDeclaredType()` to `Type` would eliminate most of the casts. WDYT?

processor/src/test/java/org/mapstruct/ap/test/generics/container/Bar.java
gunnarmorling: Hm, yes, I have to give you that one :-)

To my defence, I took the example from the original bug report; See, I wanted to make sure that the reported case really works ;-)

But yes, I still believe we should use other examples in general. All the Foo, Bar, Baz make my head spin too much :-)

gunnarmorling:
>  What I wondered was if we can somehow avoid all those unchecked casts to DeclaredType...?

I looked into this, but it's not easy. I think in general we are safe as we *know* that we deal with a declared type. Specifically, we would have done the same cast before, just at a different place.

It's not safe against crazy usage, though. E.g. `@Mapper#config(int.class)`. That'd case a CCE (as it would have before). I've filed #737 to validate this eagerly and raise a meaningful error instead.

> Looks like adding a method getDeclaredType() to Type would eliminate most of the casts. WDYT?

Not each `Type` represents a declared type. E.g. `int`. So I prefer the explicit cast which makes you (hopefully) think about whether it's right or not, instead of having this method which suggests it's always available. Ideally we feed in `DeclaredType` from the "system boundaries" (as I did for the config()/uses()) stuff, i.e. having one validated cast at the beginning.

agudian:
> So I prefer the explicit cast which makes you (hopefully) think about whether it's right or not, instead of having this method which suggests it's always available.

Exactly that crossed my mind as well. :+1: 

> I've filed #737 to validate this eagerly and raise a meaningful error instead.

I already wondered what exactly you meant with that ticket... This PR explains it... 

Okay, I'm all :+1: for this. I didn't get to break it while testing around locally as well, so please go ahead and merge it whenever your're ready... Unless Sjaak finds something else, that is... :grinning:

agudian:
oh, seems like I closed it accidentally... strange...

sjaakd:
Me like :smile:

```java
@Mapper
public interface SourceTargetMapper1 extends SourceTargetMapperBase<Bar, Foo> {

    SourceTargetMapper1 INSTANCE = Mappers.getMapper( SourceTargetMapper1.class );

    Target<Bar> mapSourceToTarget(Source<Foo> source);
}
```
```java
public interface SourceTargetMapperBase<T, S>  {

    T mapFooToBar(S foo);

}
```
generated code
```java
public class SourceTargetMapperImpl implements SourceTargetMapper {

    @Override
    public Bar mapFooToBar(Foo foo) {
        if ( foo == null ) {
            return null;
        }

        Bar bar = new Bar();

        if ( foo.getValue() != null ) {
            bar.setValue( Long.parseLong( foo.getValue() ) );
        }

        return bar;
    }

    @Override
    public Target<Bar> mapSourceToTarget(Source<Foo> source) {
        if ( source == null ) {
            return null;
        }

        Target<Bar> target = new Target<Bar>();

        target.setContent( fooListToBarList( source.getContent() ) );

        return target;
    }

    protected List<Bar> fooListToBarList(List<Foo> list) {
        if ( list == null ) {
            return null;
        }

        List<Bar> list_ = new ArrayList<Bar>();
        for ( Foo foo : list ) {
            list_.add( mapFooToBar( foo ) );
        }

        return list_;
    }
}
```

sjaakd:
I'm going to try some more.. But I reckon the above can be added as testcase.

processor/src/test/java/org/mapstruct/ap/test/generics/container/Bar.java
sjaakd: Ok. But should we add the testcases to an issue package then? `generics/container` does not really ring a bell for me..

sjaakd:
@gunnarmorling . Nice :+1: .. 

2 comments:
1.  I think you're PR is a bit wider applicable (as I showed above). Perhaps that unit test should be added.
2. The test cases should be in a package with an issue number or in the generic package itself. The container does not ring a bell to me.

Besides that: please go ahead and merge.

gunnarmorling:
> 1. I think you're PR is a bit wider applicable (as I showed above). Perhaps that unit test should be added.
> 2. The test cases should be in a package with an issue number or in the generic package itself. The container does not ring a bell to me.

Thanks, will do. I had doubts about the bug-number named packages and looked for something more "semantical", but I can change that.

sjaakd:
> Thanks, will do. I had doubts about the bug-number named packages and looked for something more "semantical", but I can change that.

The advantage of these packages is that you can copy received reproducers and leave them as is for the time being (speed & demonstration purposes: problem solved). 

So what if we would do this up to the point when we make a major release. Then its nice to refactor the reproducers into the right package, more complying to our agreements such as meaningful names and remove the 'legacy' bug packages and start with a clean slate. .

gunnarmorling:
Hey, yes, that's an interesting idea. Seems quite some work to do
now though for all the existing bug packages.

sjaakd:
> Hey, yes, that's an interesting idea. Seems quite some work to do
> now though for all the existing bug packages.

Hmm. .my ideas are always larger than the time available :smile: . But there's also no hurry. I consider it more as a gap filler: keeping the brain busy, while not over stretching it on a really nasty problem (such as #389 :cry: .

gunnarmorling:
Rebased and applied; Thanks for your comments, @sjaakd and @agudian! I've renamed the test package and added the test case suggested by Sjaak.

Issue #509, opened 2015-03-19
Allow to map source enum constants which do not exist in target to null values


Fundu4ek:

sjaakd:
There is an ignore which you can specify on a mapping.. Would that do the trick?

Fundu4ek:
I have a case when some enum constants can be mapped directly but others which can not I want to be null. And I can't find some trick to map source constants to null values in target (except using custom Mapper)

Fundu4ek:
It would be nice to not trigger errors on such enum conversions, but achieve this result when using nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT for example

sjaakd:
ah.. got your point.

hyungsul:
Instead of implicitly mapping to null, can we add special value "null" and annotate the null mapping?
Since null is not a constant value in Java, it is not possible to use ``null`` directly in the annotation attribute value. But, it is possible to use String ``"null"`` and treat it as a special value in processor.
Enum values are in uppercase (by convention), it might be safe to introduce the special value in lowercase, especially for null. 

```java
@Mapper
public interface OrderMapper {

    OrderMapper INSTANCE = Mappers.getMapper( OrderMapper.class );

    OrderDto orderEntityToDto(OrderEntity order);

    @Mappings({
        @Mapping(source = "EXTRA", target = "null"),
        @Mapping(source = "STANDARD", target = "DEFAULT"),
        @Mapping(source = "NORMAL", target = "DEFAULT")
    })
    ExternalOrderType orderTypeToExternalOrderType(OrderType orderType);
}
```

How do you think? It could be done by small changes on [EnumMappingMethod.ftl](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/resources/org/mapstruct/ap/internal/model/EnumMappingMethod.ftl) and [EnumMappingMethod.java](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/EnumMappingMethod.java).

gunnarmorling:
That's an interesting idea; As `null` is no valid enum value this seems safe to do for enum mapping methods.

In hindsight I am regretting that we re-used `@Mapping` instead of introducing a separate annotation such as `@EnumMapping`. This would have helped to avoid potential confusions.

gunnarmorling:
@agudian @sjaakd That one should be addressed by now via value mappings, right?

[...]

PR #745, opened 2016-02-09
#509 Add null mapping for enum mapping methods

hyungsul:
I implemented null mapping for enum mapping methods.
The following is an example of this feature, where "EXTRA" enum value is mapped to null.

```java
@Mapper
public interface OrderMapper {

    OrderMapper INSTANCE = Mappers.getMapper( OrderMapper.class );

    OrderDto orderEntityToDto(OrderEntity order);

    @Mappings({
        @Mapping(source = "EXTRA", target = "null"),
        @Mapping(source = "STANDARD", target = "DEFAULT"),
        @Mapping(source = "NORMAL", target = "DEFAULT")
    })
    ExternalOrderType orderTypeToExternalOrderType(OrderType orderType);
}
```

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/EnumMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/EnumMappingMethod.java
@@ -149,7 +149,8 @@ public class EnumMappingMethod extends MappingMethod {
                         );
                         foundIncorrectMapping = true;
                     }
-                    else if ( !targetEnumConstants.contains( mappedConstant.getTargetName() ) ) {
+                    else if ( !targetEnumConstants.contains( mappedConstant.getTargetName() )
+                        && !mappedConstant.getTargetName().equals("null") ) {
                         ctx.getMessager().printMessage( method.getExecutable(),
                             mappedConstant.getMirror(),
                             mappedConstant.getTargetAnnotationValue(),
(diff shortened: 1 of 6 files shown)


agudian:

agudian:

processor/src/main/java/org/mapstruct/ap/internal/model/EnumMappingMethod.java
```diff
@@ -149,7 +149,8 @@ else if ( !sourceEnumConstants.contains( mappedConstant.getSourceName() ) ) {
                         );
                         foundIncorrectMapping = true;
                     }
-                    else if ( !targetEnumConstants.contains( mappedConstant.getTargetName() ) ) {
+                    else if ( !targetEnumConstants.contains( mappedConstant.getTargetName() )
```
gunnarmorling: Thinking about it, I'd say this should work the other way around, too. I.e. "null" on the source side should be mapped into a given value of the target side enum.

gunnarmorling:
Hey @hyungsul, thanks for this PR! Looks good overall, the only thing I see is that it should work the other way around, too.

Also could you rework the commit messages to begin with "#509 ..."? Thanks!

hyungsul:
I updated the commit message-

gunnarmorling:
Hey @hyungsul, sorry for the delay.

So in between @sjaakd has been working on a new approach for mapping enums, or more generically, any value mappings of sort: See PR #764. This will also take care of mapping null values. The reason being that we concluded it has been a bad idea to (ab-)use `@Mapping` for value mappings. We'll thus deprecate this usage, with the new `@ValueMapping` annotation as a replacement.

I'm hence going to close this PR; Many thanks for your contribution and making us aware of the issue nevertheless!

Issue #746, opened 2016-02-09
ZonedDateTime to XMLGregorianCalendar


JanuszProgramowania:
When I try to map a property type ZonedDateTime to property with type XMLGregorianCalendar in generated mapper implementation occurs compilation error.

`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project myProject: Compilation failure
[ERROR] /C:/Development/Projects/myProject/target/generated-sources/com/atg/ws/dto/CustomerOrderHeaderMapperImpl.java:[308,51] cannot find symbol
[ERROR] symbol:   variable TimeZone
[ERROR] location: class com.atg.ws.dto.CustomerOrderHeaderMapperImpl`

This is caused by missing import in generated file.
`import java.util.TimeZone;`

This TimeZone class is used is the following generated function:
```
    private Calendar zonedDateTimeToCalendar(ZonedDateTime dateTime) {
        if ( dateTime == null ) {
            return null;
        }
        Calendar instance = Calendar.getInstance( TimeZone.getTimeZone( dateTime.getZone() ) );
        instance.setTimeInMillis( dateTime.toInstant().toEpochMilli() );
        return instance;
    }
```

This looks like a bug to me

agudian:
Yupp, it's a bug. Thanks for reporting! 

I'll create a fix in a pull-request right away...

agudian:
pull-request merged.

PR #747, opened 2016-02-09
#746 Add missing import for TimeZone in ZonedDateTimeToCalendar

agudian:
Just a quick fix. I checked (again) the other built-ins for missing imports, but didn't find any.

The tests didn't discover this, as both directions of the conversion were tested in the same mapper, and the `CalenderToZonedDateTime` adds the import correctly.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/CalendarToZonedDateTime.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/CalendarToZonedDateTime.java
@@ -23,7 +23,6 @@ import static org.mapstruct.ap.internal.util.Collections.asSet;
 import java.time.ZonedDateTime;
 import java.util.Calendar;
 import java.util.Set;
-import java.util.TimeZone;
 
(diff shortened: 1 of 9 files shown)


agudian:
> @agudian : Should we add a unit test?

We might want to refactor the tests for the built-in methods to use only one built-in for each generated mapper. I'll create an issue for it, if it's OK with you...
On second thought, I'll just do that tomorrow or so and extend this PR... we have enough of those issues lying around already... :grin:

gunnarmorling:
Ok, I'll leave that one open then for the time being. If you don't get to soon-ish, let's merge this one and do the test re-org separately. Thanks for providing this quick fix btw :)

agudian:
Refactoring of the tests wasn't that much work after all, as the `BuiltInTest` was already using separate mappers for everything. Only Mappers for `ZonedDateTime` were missing.
I've reconfigured the test class to perform only two compilations for each compiler, which reduces the run-time significantly, at least for me :wink:.

agudian:
merged

Issue #754, opened 2016-02-13
Convert reference documentation to AsciiDoc


gunnarmorling:
Hey @sjaakd and @agudian, I spent some time and started to convert the reference documentation into the AsciiDoc format. This gives us many advantages:

* Simpler mark-up, no more mixture of HAML and Markdown
* Proper, automated numbering :)
* Scrollable menu, nothing cut off anymore on smaller screens
* Can be rendered to HTML and PDF
* We can include the reference guide to the distribution bundle

Please find the very first rough result attached. That's the PDF (GitHub doesn't allow to attach the HTML, which looks a bit nicer :/):

[index.pdf](https://github.com/mapstruct/mapstruct/files/129086/index.pdf)

I still need to do several changes, such as fixing references but you already get a good impression. The gitbook stuff looks nice to, but for now I prefer to have the docs in the project itself so we can include it into the dist as mentioned above.

Note that I just aim for the conversion with this as a first step. Subsequently we also should re-structure the guide to make it more logical, but I'll leave that as separate step.

Let me know what you think.

agudian:
Very nice! :+1:
I'll try to find some time tonight to look into it in more detail.

sjaakd:
Indeed. Very clear I was able to read it in the supermarket on my mobile while pushing the cart and doing shopping :smile: I'll look into it a bit further later this weekend. 

Ps I did see the url/references were not converted.

sjaakd:
> Ps I did see the url/references were not converted.

I meant the local references, like: `[Implicit type conversions](#section-datatype-conversion-01))`

gunnarmorling:
I've pushed my work-in-progress to https://github.com/gunnarmorling/mapstruct/tree/754, so if you like you can also build it yourself (mvn clean install -pl documentation) and take a look at the HTML under target/generated-docs.

Have fun shopping for now, @sjaakd :-)

gunnarmorling:
Done :)

PR #755, opened 2016-02-13
#754 Moving reference guide to core project, converting it to AsciiDoc

gunnarmorling:
Hey @sjaakd, @agudian, there you go :) I'm quite happy with the result, in the end it didn't take as long as I had feared.

I also re-organized some chapters, we are down to 11 top-level chapters now. It's not perfect yet and some sections need actual rewriting, but it's a bit better that way already.

Once this PR is merged, I will send another one for updating the website to make use of the new guide.


gunnarmorling:
Ah, I put the guide under the [Creative Commons Attribution-ShareAlike 4.0 International License](http://creativecommons.org/licenses/by-sa/4.0/). I hope that's ok with you ;) There were two commits from other authors to the original guide, but as this were just tiny fixes, that should be alright.

sjaakd:
> Ah, I put the guide under the Creative Commons Attribution-ShareAlike 4.0 International License. I hope that's ok with you ;) There were two commits from other authors to the original guide, but as this were just tiny fixes, that should be alright.

I'm ok :+1:

sjaakd:
@gunnarmorling Looks nice :+1: 

> Once this PR is merged, I will send another one for updating the website to make use of the new guide.

I assume that you are going to add a dependency to this  in our MapStruct.org build. So, what will be placed on Nexus when you compile this pom, and how will the web-site build make use of that?

gunnarmorling:
> So, what will be placed on Nexus when you compile this pom, and how will the web-site build make use of that? 

Good question. I thought about just copying the files manually as a first step, but indeed it'd be nice to automate it somehow. Ideas welcome :)

sjaakd:
> Good question. I thought about just copying the files manually as a first step, but indeed it'd be nice to automate it somehow. Ideas welcome :)

I reckon that you can use the maven-assembly-plugin to create an assembly (e.g. a zip) which contains the `mapstruct-reference-guide.asciidoc` file. You an unpack that in the mapstruct.org build. That's the way I move around Flex resources for my web projects.

agudian:
Or we move it to the mapsrtuct.org repo again and let it build there. But I guess one reason to move it here is to have the reference-doc bound to a specific (release) version, right? So we could also think about adding the documentation as git-submodules (one for each release) in the m.org repo, pointing to the respective tags in this repo...?

sjaakd:
> So we could also think about adding the documentation as git-submodules (one for each release) in the m.org repo, pointing to the respective tags in this repo...?

That's more or less the same approach as I proposed right? So: the reference-doc sub-module has version X in repo and we build org.m against that version X (dependency) using the ascii-doc file downloaded from version X in the repo. X being a stable version. As soon as we release, we only have to set the maven dep of org.m to the new release.

agudian:
Well, that would mean that we'd have to get the docs right at the first attempt... ;-). We should always be able to just fix up the docs of released versions easily, without doing another release.

sjaakd:
> Well, that would mean that we'd have to get the docs right at the first attempt... ;-).

Yes, I don't see a problem :+1: ?


> We should always be able to just fix up the docs of released versions easily, without doing another release.

You mean the on-line version? Because the one we are now going to package alongside the code can't e withdrawn once its released.. So, perhaps we should maintain some kind of addendum on the website. Or... alternatively.. we could publish a snapshot-version from a git maintenance branch.

gunnarmorling:
> But I guess one reason to move it here is to have the reference-doc bound to a specific (release) version, right?

Yes, exactly, the reference guide needs to be part of the distribution bundle. Also it must be clearly linked to specific releases of the project, hence I think this repo is the place where it should be.

Regarding sub-modules etc., I am open for everything, as long as it easy to use and maintain. Give it a try if you like.

For now I've just copied the rendered HTML/PDF and copied it to the mapstruct.org repo:

* HTML: http://mapstruct.org/documentation/1.0/reference/html/index.html
* PDF: http://mapstruct.org/documentation/1.0/reference/pdf/mapstruct-reference-guide.pdf
* Also the API JavaDoc is available online now: http://mapstruct.org/documentation/1.0/api/index.html

Check out the short blog post for the details: http://mapstruct.org/news/2016/02/14/reference-guide-brushed-up.html.

> We should always be able to just fix up the docs of released versions easily, without doing another release

As Sjaak is saying, we cannot alter the release bundle once published, but we can update the version on the website if really needed. I think in most cases doing it for the next version should be good enough.

Btw. the minor (1.0, soon 1.1) version is part of the URL scheme, so we can have multiple documentation versions online in the future. Once 1.1 is out, there should be a "stable" alias for the current stable release.

gunnarmorling:
Rebased and applied. Thanks for your feedback, @sjaakd and @agudian! Please let me know should you find any further issues.

Issue #750, opened 2016-02-12
Introducing `@Named` qualifier after CDI analogy 


sjaakd:
` @Named("somemapping")` annotation and adding a qualifiedByName("somemapping") attribute to `@Mapping`.

See #749 for more details.

sjaakd:
@gunnarmorling .. See my answers. Thans for reviewing.

PR #756, opened 2016-02-14
750

sjaakd:
Introducing `@Named` annotation. See #750 

There's something strange on the `MappingCreationProcessors` with the line endings. How did you guys configure the auto CRLF? I've tried several things. Somehow my standard mac git gets the line-endings mixed up. Anyway, I've got an additional commit (see below), but it does not show the CRLF changes. Strange....


gunnarmorling:
Regarding the line endings, did you read this one: https://help.github.com/articles/dealing-with-line-endings/

We have a .gitattributes in the repo which should take care of it, so I am surprised that it apparently doesn't work for you. Have you tried the explicit setting described in that link?

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/named/MovieMapper.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @author Sjaak Derksen
+ */
+@Mapper( uses = { Titles.class, SomeOtherMapper.class, KeyWordMapper.class, FactMapper.class } )
+public interface MovieMapper {
+
+    MovieMapper INSTANCE = Mappers.getMapper( MovieMapper.class );
+
+    @Mappings( {
+        @Mapping( target = "title", qualifiedByName = { "TitleTranslator", "EnglishToGerman" } )
```
gunnarmorling: Interesting that you allow for several values. Does it mean that *all* given qualifiers must be matched? It's a deviation from what we do for traditional qualifiers. Is there a strong need for this? For the sake of consistency with `qualifier()`, I'd prefer a single value to be given.

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/handwritten/Titles.java
```diff
@@ -22,15 +22,16 @@
...
 
 import com.google.common.collect.ImmutableMap;
+import org.mapstruct.Named;
 
 /**
  *
  * @author Sjaak Derksen
  */
 @TitleTranslator
+@Named( "TitleTranslator" )
```
gunnarmorling: What does it mean if the qualifier is given on the type level? Just seeing it for `@TitleTranslator`, too, btw. I always had it in mind for methods only. Is it the default for all methods declared by this type?

processor/src/main/java/org/mapstruct/ap/internal/model/source/SelectionParameters.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+import javax.lang.model.type.TypeMirror;
+
+/**
+ * Holding parameters common to the selection process, common to IterableMapping, BeanMapping, PropertyMapping and
+ * MapMapping
+ *
+ * @author Sjaak Derksen
+ */
+public class SelectionParameters {
```
gunnarmorling: It seems to resemble `SelectionCriteria` a bit. Can/should the two be unified?

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
```diff
@@ -58,6 +58,8 @@
      *
      * @param method method
      * @param existingVariableNames existingVariableNames
+     * @param beforeMappingReferences
+     * @param afterMappingReferences
      */
     protected MappingMethod(Method method, Collection<String> existingVariableNames,
```
gunnarmorling: I suppose I'd rather remove the entire comment. Doesn't provide any information really.

gunnarmorling:
Thanks a lot, @sjaakd. Looks good overall, see above for some inline comments.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SelectionParameters.java
sjaakd: I thought about that as well. However:
1. Selection Criteria contain more criteria, e.g. the ones that are derived during the process.
2. Package restrictions restrict us from global usage. It feels like a mixup of concerns to move SectionCriteria to the source package (away from the selection process).
However, bundling selection parameters is a logical thing to do. I spotted a fault, because in `Mapping` they should for instance not be part of the reverse mapping.

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
sjaakd: Ok.. I added those because my IDE whines about that :smile: I'll remove them.

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/handwritten/Titles.java
sjaakd: > Is it the default for all methods declared by this type?

Yes. And.. moreover it is refined by a qualifier on method level. So it's possible to say, for this mapping methods look for `Titles` and translate those from `EghlishToGerman`. Both qualifications need to be met.

Its actually behavior that we specified from the beginning (I did not invent this for the `@Named`.


> Just seeing it for @TitleTranslator, too, btw. I always had it in mind for methods only

:smile: we've got more functionality than you think..

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/named/MovieMapper.java
sjaakd: > Does it mean that all given qualifiers must be matched?

True.. However, this is functionality as we we designed it already for the regular `@Qualifier`. Did not change this for `@Named`. See comment elsewhere.

sjaakd:
> Thanks a lot, @sjaakd. Looks good overall, see above for some inline comments.

@gunnarmorling , thanks for reviewing :+1: I answered your questions. I'll squash the last two commits and remove the comment all-together. Let me know if you're happy with the answers and I'll move ahead with merging.

sjaakd:
> Regarding the line endings, did you read this one: https://help.github.com/articles/dealing-with-line-endings/

I did.. however, Git seems to be a bit confused.. Hope its finished now (it started some time ago with the MapStruct/examples).

agudian:
Hey guys, I didn't do a full review, but it would be great if the javadocs of the new annotation and the new attribute would contain some meaningful examples. We have that on the list for a while now, and should do it at least for all new annotations and elements. And it would need to be done for the doc-updates anyway (which could also be part of this PR :grin:)

agudian:
Oh great, now I see it's merged already... guess I'm too late now...

sjaakd:
> Hey guys, I didn't do a full review, but it would be great if the javadocs of the new annotation and the new attribute would contain some meaningful examples. We have that on the list for a while now, and should do it at least for all new annotations and elements. And it would need to be done for the doc-updates anyway (which could also be part of this PR )

Sorry @agudian .. I reckon I was a bit too fast closing this. I wrote a follow up issue #757 .. I'll take this up this weekend.

sjaakd:
Hmm. something went wrong in the end with commit https://github.com/sjaakd/mapstruct/commit/6a5e2a79e4a8f98a5c86d8f9bd15c891b6ecdcbd

The line-endings got checked in, but not the original commit. I just noticed it..

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/named/MovieMapper.java
gunnarmorling: Ok, no idea why but I assumed we'd only allow for a single `@Qualifier` type being given, but it's an array indeed. I must have dreamed :)

gunnarmorling:
@sjaakd, can you also send a PR with the doc updates? No excuses anymore, now that the docs are in the project itself and more easily written than ever :)

sjaakd:
> @sjaakd, can you also send a PR with the doc updates? No excuses anymore, now that the docs are in the project itself and more easily written than ever :)

Aye sir.. (see above) :smile:

Issue #751, opened 2016-02-12
Get access to mapper instance


sjaakd:
I keep on finding cases were I need access to a used mapper instance (abstract class).

Cases I came up with so far: 
1. Inserting sequence numbers.
2. Reuse of common base mapper (were you need an instance variable on the mapper distinguising between behaviors).

The only way to do this  (in the standard component model) is `ThreadLocals`. 

For scenario 1. It woudl be nice if the generated mapper generates standard getters on its instance variables when a getter is present in the `@Mapper` interface / abstract class with a used mapper as return type.

For scenario 2. It could help if a used mapper is constructed by factory as well (not only target objects).

See also http://g00glen00b.be/mapstruct/

afdia:
I also needed this feature some time ago where I had a state mapper which is used by several other mappers.
The state mapper maps state objects but also fills in missing fields based on a state-cache it holds (so if only iso3 is available it fills in name and vice versa)

The problem is that the state-cache is expensive to initialize and should be reused as a singleton, but the application doesnt use jsr330 or spring.

If I wouldn't use MapStruct I could pass arguments in the constructor and had full control over the objects it uses. e.g.:
```
StateCache stateCache // passed via constructor to this class and already initialized
OtherMapper otherMapper = new OtherMapper(new StateMapper(stateCache));
```

In MapStruct we cannot pass constructor arguments, so I tried to use a setter to set the cache after the initialization of OtherMapper, but I do not get the reference to StateMapper (because it's a private field within ConcreteMapperImpl)

If I understand your issue description correctly, after your change it should be possible to add the following method to OtherMapper
```
abstract StateMapper getStateMapper();
```
and it would create an implementation which returns the private StateMapper instance?

sjaakd:
@afdia , thanks for your response. The PR which is now out there is going to be replaced. My current ideas:
1. Generate constructors with overloaded arguments whenever you use an abstract class as `Mapper` definition. I've got a working solution for that.
2. Generate a factory implementation as well. The user would have to define the factory methods in an interface or abstract class which would be available via the `Mappers` : so `Mappers.getFactory(<factory-interface>) interface similar like a mapper would be. 
3. For the solution of nested mappers, I would like to propose to add the possibility to appoint a hand written factory (perhaps via the already existing 'uses' mechanism). 
4. The user could write such hand-written factory and delegate to the proper overloaded constructor via the user defined interface and obtained generated factory.

 I still need to study how this would work with the other component models. 

WDYT?

[...]

PR #763, opened 2016-02-21
#751, making used mappers available

sjaakd:
This PR makes the used mappers available for initialization or other state changes. The factory part of #751 is not addressed in this PR.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/accessingmappers/InventoryMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/accessingmappers/InventoryMapper.java
@@ -1,0 +1,49 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.accessingmappers;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+@Mapper(uses = { RadioMapper.class, TvMapper.class } )
+public abstract class InventoryMapper {
+
+    public static final InventoryMapper INSTANCE = Mappers.getMapper( InventoryMapper.class );
+
+    private final InventorySequence inventorySequence = new InventorySequence();
+
+    public  InventoryEntity mapInventory(InventoryDto source) {
+        getRadioMapper().init( inventorySequence );
+        getTVMapper().init( inventorySequence );
+        return handleMapInventory( source );
+    }
+
+    protected abstract InventoryEntity handleMapInventory(InventoryDto source);
+
+    protected abstract RadioMapper getRadioMapper();
+
+    protected abstract TvMapper getTVMapper();
+
+
+}
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -465,6 +465,11 @@ private CarMapper mapper;
 
 A mapper which uses other mapper classes (see <<invoking-other-mappers>>) will obtain these mappers using the configured component model. So if `CarMapper` from the previous example was using another mapper, this other mapper would have to be an injectable CDI bean as well.
 
+[TIP]
+====
+The used mapper can be made accessible (e.g. for initialization), by simply adding a getter to the mapper that returns the used mapper.
+====
+
 [[datatype-conversions]]
 == Data type conversions
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ReferenceAccessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ReferenceAccessor.java
@@ -1,0 +1,55 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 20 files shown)


sjaakd:
test this please

processor/src/test/java/org/mapstruct/ap/test/accessingmappers/InventoryMapper.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+@Mapper(uses = { RadioMapper.class, TvMapper.class } )
+public abstract class InventoryMapper {
+
+    public static final InventoryMapper INSTANCE = Mappers.getMapper( InventoryMapper.class );
+
+    private final InventorySequence inventorySequence = new InventorySequence();
+
+    public  InventoryEntity mapInventory(InventoryDto source) {
+        getRadioMapper().init( inventorySequence );
```
gunnarmorling: Why is it that you need these init() calls here? Couldn't radio mapper and tv mapper just initialize their sequence field themselves in their constructor?

sjaakd: They share the same sequence..

gunnarmorling: Ok, got it. The issue I see with the current proposal is that the other classes using the sequence can't assign the sequence passed via init() to a final field, thus these references may not be published safely to other threads.

I take it you don't use Spring or CDI?

sjaakd: > I take it you don't use Spring or CDI?

Nope..  The standard component model.


> The issue I see with the current proposal is that the other classes using the sequence can't assign the sequence passed via init() to a final field, thus these references may not be published safely to other threads.

I reckon thread safety is up to the programmer. If he/she want's to use this in a static context, or in a context that pools objects, he/she should be aware of this. Personally, I recreate mappers via the `Mapper.getMapper` whenever I need one..

gunnarmorling: The thing is that we kind of encourage static singletons via the `INSTANCE` pattern we often use. How about making your sequence a singleton, obtainable via a static method such as `Sequence#getInstance()`?

sjaakd: > The thing is that we kind of encourage static singletons via the INSTANCE pattern we often use.

Huh? the  `Mapper.getInstance` creates a new mapper every time.. Right?

sjaakd: > How about making your sequence a singleton, obtainable via a static method such as Sequence#getInstance()?

How would that help me? I need one sequence within one bean call. Anyway. this is just one example. Another is were I've got 2 quality regimes that require different mappers, deep down in an object graph.

gunnarmorling: > Huh? the Mapper.getInstance creates a new mapper every time.. Right?

Yes, but this pattern invokes it only once, then that same instance is re-used: `MyMapper.INSTANCE.mapXyz()`.

> I've got 2 quality regimes that require different mappers

WDYM by that :)

I am wondering whether we should ensure singletons during one top-level `getMapper()` call, i.e. if one dependency is requested several times, the same instance would be returned. Then, if you'd inject your sequence via `uses()`, this should do the trick for you?

sjaakd: The thing is that we kind of encourage static singletons via the INSTANCE pattern we often use.
  Ok.. I get it. You mean by assigning it to a final static we imply that Mappers are stateless. 
  
  To eleborate a bit more.. I've got a requirement that id's in a result document are generated unique within the scope of a soap request. Some of my mappers are shared between products as common utils. I don't want to revert to CDI only to produce mappers with state added at construction time. Running through the returned document the first entry I find get 1, the next 2, etc. ID's are required for common java / XSD classes. I now use a thread local to get the ID generater (basically a wrapped counter) to the right mapper (generated, but also hand written). But I don't like thread-locals and try to avoid them were possible. Using a classic singleton patterns makes the problem only worse, since then my sequence generated is shared between all threads running in my VM.
  Anyway. I don't think MapStruct should impose limitatons on how a user would like to configure a mapper (completely hand written or generated).
  A colleague of mine stumbled on another situation. The ones with QualityRegimes. He would like to configure the behaviour of a mapper depending based on the regime. Regime 'A' supports another set of string based value mappings then regime 'B'. Depending on the regime a different hash map is chosen when callign a qualifying method.
  Hope this makes the case a bit more clear.
  BTW: another option would be to support a factory for creation of Mappers. I reckon that impact will be quite large.
  
  
  
   Op 25 februari 2016 om 22:46 schreef Gunnar Morling <notifications@github.com>:
   
   
   In processor/src/test/java/org/mapstruct/ap/test/accessingmappers/InventoryMapper.java:
   > +import org.mapstruct.factory.Mappers;
> +
> +
> +/**
> + *
> + * @author Sjaak Derksen
> + */
> +@Mapper(uses = { RadioMapper.class, TvMapper.class } )
> +public abstract class InventoryMapper {
> +
> +    public static final InventoryMapper INSTANCE = Mappers.getMapper( InventoryMapper.class );
> +
> +    private final InventorySequence inventorySequence = new InventorySequence();
> +
> +    public  InventoryEntity mapInventory(InventoryDto source) {
> +        getRadioMapper().init( inventorySequence );

   The thing is that we kind of encourage static singletons via the INSTANCE pattern we often use. How about making your sequence a singleton, obtainable via a static method such as Sequence#getInstance()?

gunnarmorling: Yes, I understand where you coming from. I think it's a valid and common issue, but not one easily solved. Essentially, you need to share "state" during the runtime of one `getMapper()` call.

Just thinking out loud here: How about we provide an overloaded version of `getMapper()`, that takes a callback provided by the user which is in charge of instantiating required dependencies. That you could ensure identity of dependencies as needed:

    Sequence sequence = ...;
    InventoryMapper inventoryMapper = Mappers.getMapper(
        InventoryMapper.class,
        ( dependencyClazz, defaultFactory ) -> {
            if ( dependencyClazz = Sequence.class ) {
                return sequence; // return same sequence instance during this call 
            }
            else {
                return defaultFactory.getMapper( dependencyClazz );
            }
        }
    );

The implementation would have to keep that callback as thread-local so it can be accessed by all `getMapper()` calls in the course of one invocation, but that'd be encapsulated within MapStruct's code of the factory, so the user wouldn't have to deal with that.

sjaakd: What about splitting the problem in two. What I mean is that the construction of abstract classes with overloaded constructors can be done by generating a factory from an interface with methods mimicking the constructors you want to use. E.g via our own @Factorize annotation. The factory itself can be constructed in the same manner as mappers.

Second part: a similar mechanism for zero arg constructors in mappers. The mapper could look for zero-arg method marked with our own @produces annotation. No implementation will be generated. The user has to implement those and call the proper method on the generated factory. 

Naming could be better probably.. 😄

Verstuurd vanaf mijn iPhone

sjaakd:
Working on a different solution.

Issue #749, opened 2016-02-10
Move Enum mappings from @Mapping to an @ValueMapping 


sjaakd:
I'm often in the situation that I need to map
String -> String
Integer -> Sting
String -> Integer
etc. 

The current way is to specify a qualifer. So you and up with 10 to 20 qualifiers for 20 String to String methods + a method implementation for each one of them. In other words, more code then you write it by hand. Qualifiers / methods are simply to clunky.

Proposal:
Make it possilbe to directly refrence a (immutable) Map to be used for such mappings.

gunnarmorling:
Why is it that you need to specify a qualifier to begin with? Can you share
an example?

sjaakd:
Well, I've got several methods that require a mapping from String to String. Each making use of its own (static) `Map`. To distinguish between them, I now need a qualifier to select the proper mapping.

Example coming up :smile:

sjaakd:
Example:

```java

public class SourceClass {

    private String sourceMember;

    public String getSourceMember() {
        return sourceMember;
    }

    public void setSourceMember(String sourceMember) {
        this.sourceMember = sourceMember;
    }
   
}
```

``` java

public class TargetClass {

    private String targetMember;
    private String secondMember;
    
    public String getTargetMember() {
        return targetMember;
    }

    public void setTargetMember(String targetMember) {
        this.targetMember = targetMember;
    }

    public String getSecondMember() {
        return secondMember;
    }

    public void setSecondMember(String secondMember) {
        this.secondMember = secondMember;
    }
    
    
}

```

``` java
import java.util.HashMap;
import java.util.Map;

public class Source2TargetMapper {

    public static Map<String, String> secondMemberMapping;
    static {
        secondMemberMapping = new HashMap<String, String>();
        secondMemberMapping.put( "first", "1st" );
        secondMemberMapping.put( "second", "2nd" );
    }
    
    public TargetClass source2Target( SourceClass source ) {
        
        TargetClass target = new TargetClass();
        // Straight forward mapping: (could be done with @Mapping( target = "targetMember", source = "sourceMember" ) using mapstruct)
        target.setTargetMember( source.getSourceMember() );
        
        // can be mapped using mapstruct 'expression': 
        // @Mapping( target = "secondMember", expression = "java(Source2TargetMapper.secondMemberMapping.get(source.sourceMember))")
        // or by using a qualifier
        target.setSecondMember( secondMemberMapping.get( source.getSourceMember() ) );
        
        return target;
[...]

PR #764, opened 2016-02-22
#749 Introduction of ValueMappings (enum-to-enum-mapping only)

sjaakd:
This should solve the 'default' and 'null-source' (see #509) issues. It also introduces a way to deprecate the enum-mapping as being part of the  bean `@Mapping`.

* javadoc is still based on the original idea of `@ValueMapping`. I'll adapt that soon and will add some examples as well.
* doc will needs to be updated. I'll do so soon as well.


sjaakd:
I still need to consider reverse mappings. First idea here; only consider mappings of mapping type source. 

Note: Null targets could perhaps be mapped to mapping type null.

core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
```diff
@@ -0,0 +1,217 @@
...
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Configures the mapping of source constant value to target constant value.
```
gunnarmorling: Seeing these examples, I came across this very interesting doclet: https://github.com/jtulach/codesnippet4javadoc. It'd allow us to include code from actual source files into JavaDoc comments. Maybe you feel like giving it a try?

sjaakd: Yep... You came as being called :+1: .. I was just implementing the reverse cases when I saw that I did not follow the 
target - source convention ( but source -target ). I did the code and just thought.. hey.. need to do the javadoc as well.

sjaakd:
@gunnarmorling , I'll have a look at the auto doc generation later this week. Need to fix the javadoc still. For now, I just pushed a commit on the reverse mappings.

I'm not sure though about the null source / null target automation in the reverse mapping.  But have a look yourself.

core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
sjaakd: Wouldn't that give a cyclic dependency? Our API projects then start to depend on the UT of their implementation.

gunnarmorling: We could have the example snippets in the test dir of the API module. So there'd still be some duplication with the actual tests I suppose, but at least we would compile the snippets as part of the build and thus make sure they are not broken.

sjaakd:
The code snippet 4 javadoc works nice. The only problem is duplication, especially when you consider that we have a java8 and regular library to cope with the `@Repeatable` annotation.

gunnarmorling:
> The code snippet 4 javadoc works nice.

Thanks for trying it out. Seems like a good thing then :)

> The only problem is duplication, especially when you consider that we have a java8 and regular library to cope

Couldn't the snippets live in src/test/java of the common module?

core-jdk8/src/main/java/org/mapstruct/Mapping.java
```diff
@@ -37,6 +37,10 @@
  * In addition, the attributes {@link #dateFormat()} and {@link #qualifiedBy()} may be used to further define the
  * mapping.
  *
+ * <p>
+ * <b>IMPORTANT NOTE:</b> the enum mapping capability is deprecated and replaced by {@link ValueMapping} it
```
gunnarmorling: Are we raising a WARN note, too, when using `@Mapping` for enum-to-enum mapping?

gunnarmorling:
@sjaakd, what's the core.zip file :)

processor/src/test/java/org/mapstruct/ap/test/value/SpecialOrderMapper.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+@Mapper
+public interface SpecialOrderMapper {
+
+    SpecialOrderMapper INSTANCE = Mappers.getMapper( SpecialOrderMapper.class );
+
+    OrderDto orderEntityToDto(OrderEntity order);
+
+    @ValueMappings({
+        @ValueMapping( target = "DEFAULT", valueMappingType = ValueMappingType.NULL ),
```
gunnarmorling: So I have to admit I find the approach around `ValueMappingType` less comprehensible that what we used to have. It would be so much nicer to read (and write) like so:

    @ValueMapping( target = "DEFAULT", source="null" )
    @ValueMapping( target = "null", source = "STANDARD" )

I understand that the proposed mechanism is more versatile, but I feel we are adding complexity without gaining very much at this point. The only thing supported is enum mappings, but it's happening in a much more concealed way now by this generic approach.

My preference is to introduce an explicit `@EnumMapping`, because that's what we want out if this right now. We still can add `@ValueMapping` further down the road if we see the need for other value mappings.

processor/src/test/java/org/mapstruct/ap/test/value/DefaultOrderMapper.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+@Mapper
+public interface DefaultOrderMapper {
+
+    DefaultOrderMapper INSTANCE = Mappers.getMapper( DefaultOrderMapper.class );
+
+    OrderDto orderEntityToDto(OrderEntity order);
+
+
+    @ValueMapping( valueMappingType = ValueMappingType.DEFAULT_AFTER_APPLYING_NAME_BASED_MAPPINGS, target = "DEFAULT" )
```
gunnarmorling: So is this meant as a "catch all unmatched" mapping for all unmapped ones? If so, how about this, using the more explicit `@EnumMapping`:

    @EnumMapping( source = "*", target = "DEFAULT" )

Much easier to read, isn't it? That's the advantage of the explicit `@EnumMapping`: We can use special values such as "null" and "*" because these are no valid enum constants, hence there are no ambiguities.

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -28,6 +28,13 @@
 public enum Message {
 
     // CHECKSTYLE:OFF
+// CHECKSTYLE:OFF
```
gunnarmorling: That looks odd?

core/src/test/java/org/mapstruct/ap/test/value/OrderMapper.java
```diff
@@ -0,0 +1,43 @@
...
+
+    OrderMapper INSTANCE = Mappers.getMapper( OrderMapper.class );
+
+    @ValueMappings({
+        @ValueMapping(source = "EXTRA", target = "SPECIAL"),
+        @ValueMapping(source = "STANDARD", target = "DEFAULT"),
+        @ValueMapping(source = "NORMAL", target = "DEFAULT")
+    })
+    ExternalOrderType orderTypeToExternalOrderType(OrderType orderType);
+// FINISH: OrderMapper
```
gunnarmorling: I suppose the closing "}" should be part of the snippet?

core-jdk8/src/test/java/org/mapstruct/ap/test/value/OrderMapper.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.mapstruct.Mapper;
+import org.mapstruct.ValueMapping;
+import org.mapstruct.ValueMappings;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Gunnar Morling
+ */
+// BEGIN: OrderMapper
```
gunnarmorling: Couldn't those be referenced from the test suite in the processor module? Or can one not reference across module boundaries?

processor/src/test/java/org/mapstruct/ap/test/value/DefaultOrderMapper.java
gunnarmorling: So more specifically, we can allow "null" as valid value for `target()` and `source()` and "*" as valid value for `source()`.

gunnarmorling:
Hey @sjaakd, finally got back to this. Sorry for the delay :/ I've added some more comments inline; All in all I'd feel much better if we went for the more explicit `@EnumMapping` for the sake of explicitness and easier comprehensibility, also enum mappings is the only thing we need right now.

Note that this does not prevent us from adding `@ValueMapping` as a more generic vehicle in the future, should it be needed. This wouldn't render `@EnumMapping` obsolete, as there are advantages to it which justifies the existence of this more specialized vehicle in addition to the more generic mechanismm.

core-jdk8/src/test/java/org/mapstruct/ap/test/value/OrderMapper.java
sjaakd: How would that look dependency wise? The code needs to compile. Processor test is dependent on the API modules. The API modules require the snippets.. The same goes for the relation between API Common and API Java8 . We then would need the `@Mapping` and `@ValueMapping` in the API common module. Right?

processor/src/test/java/org/mapstruct/ap/test/value/SpecialOrderMapper.java
sjaakd: Ok.. I'll reconsider. However, if we followup on the `@ValueMapping`, we leave room open for enum-to-string and string-to-enum.  Are you planning a dedicated mapping for each from-to type in the future?

sjaakd: I'll need some time to think about your proposals.. :smile: . Perhaps we should have a small telco later this week. See if @agudian  can shed some light on this discussion as well.

core-jdk8/src/test/java/org/mapstruct/ap/test/value/OrderMapper.java
gunnarmorling: The snippets are taken from sources, not compiled classes. So you could reference it like so:

    <additionalparam>-snippetpath "${session.executionRootDirectory}/processor/src/test/java"</additionalparam>

This just did the trick for me. This leaves the case of `OrderMapperImpl`, but then having this in the docs maybe is a bit too much anyways?

processor/src/test/java/org/mapstruct/ap/test/value/SpecialOrderMapper.java
gunnarmorling: Yes, we can do a quick call next week. How about Monday?

> we leave room open for enum-to-string and string-to-enum

Yes, that's true. Actually, `@ValueMapping` is growing on me if we go for the more concise "null" / "\*" syntax. 

Of course these are valid strings, but this could be handled using escaping. So if you really wanted to map to the Strings "\*" or "null", you'd have to put some escaping in place: "\\\*" and "\null". If you wanted to map into "\\\*", the backslash would have to be escaped as well: "\\\\*" and so on. We also could use "\_null\_" instead of "null", decreasing the number of cases where escaping is needed.

Bottom line being, if we managed to get rid of `ValueMappingType` that way, I'd like it much better :)

How does this sound?

sjaakd: > How does this sound?

actually: quite good :+1: . Lets settle on some reusable syntax for general use on wildcards, null, default and the likes :). Backspacing is a natural way to do it I'd say.

sjaakd: > Yes, we can do a quick call next week. How about Monday?

Ok. sometime after 20:00. Perhaps we could discuss the constructor  thingy ( #751 )  as well. I've got some ideas on how to continue here as well.

processor/src/test/java/org/mapstruct/ap/test/value/DefaultOrderMapper.java
sjaakd: > "catch all unmatched" mapping for all unmapped ones?

Not sure what you mean. You have 2 situations:
1. Let all source methods, not mentioned in the mapping end up in default (ValueMappingType.DEFAULT)
2. Try to match source to target based on name, and end up in default for the remainder. (`ValueMappingType.DEFAULT_AFTER_APPLYING_NAME_BASED_MAPPINGS`)

So your proposal `source = "*"` does (I reckon) not distinguish between the two. Idea's?

processor/src/test/java/org/mapstruct/ap/test/value/SpecialOrderMapper.java
sjaakd: > Yes, that's true. Actually, @ValueMapping is growing on me if we go for the more concise "null" / "*" syntax.

When you repeat it enough, it just sounds like `@EnumMapping` :smile: :imp:

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
sjaakd: Hmm. you mean the fact that the check style is off or the changed ident..?

core/src/test/java/org/mapstruct/ap/test/value/OrderMapper.java
sjaakd: will adapt

core-jdk8/src/test/java/org/mapstruct/ap/test/value/OrderMapper.java
sjaakd: > but then having this in the docs maybe is a bit too much anyways?

Currently, its common practice to refer to the generated code. Are you suggesting to stop doing that (e.g. in favor of explanation)?

sjaakd: > This just did the trick for m

Ok.. I'll do the same then. Although that means you refer (resources / code) outside a specific module. You don't think that's a problem?

core-jdk8/src/main/java/org/mapstruct/Mapping.java
sjaakd: > Are we raising a WARN note, too, when using @Mapping for enum-to-enum mapping?

No..  My normal development practice is that I change from WARN non blocking to WARN blocking policy after I'm ready developing. Perhaps better would be an INFO mechanism.

processor/src/test/java/org/mapstruct/ap/test/value/SpecialOrderMapper.java
sjaakd: I'll remove the zip file.. This accidentally ended up in the source code.

processor/src/test/java/org/mapstruct/ap/test/value/DefaultOrderMapper.java
gunnarmorling: > not mentioned in the mapping end up in default
>  and end up in default for the remainder. 

Aren't these two the same?

That's what I suggest to cover by "*": Anything not mentioned in the mapping will be mapped to the value given for "*". Everything mentioned explicitly will be mapped to the given value.

Or am I missing something?

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
gunnarmorling: More the fact that this line is repeated five times :)

core-jdk8/src/test/java/org/mapstruct/ap/test/value/OrderMapper.java
gunnarmorling: > Are you suggesting to stop doing that (e.g. in favor of explanation)?

Good question; I wasn't even aware that we are doing that :) I think I'd prefer to not show implementations (not least to not create some sort of "specification" people will rely on) and rather give some more abstract explanations.

gunnarmorling: > Although that means you refer (resources / code) outside a specific module. You don't think that's a problem?

Admittedly, it's not "100% clean" (e.g. when cd-ing in the "core" module and building from there, the snippets won't be found), but I think it's the lesser evil compared to all the duplication. But it's your call really :)

sjaakd: > I think I'd prefer to not show implementations (not least to not create some sort of "specification" people will rely on)

Hmm. we do that all over our documentation in general. So, let's not do this in the javadoc then and settle for a more abstract notation.

> Admittedly, it's not "100% clean"

OK. I'll go ahead with your proposal.. (Was already doing that)..

gunnarmorling: Ah, you meant the ref guide. + 1, Yes, let's not show impls in Javadoc.
Those should be as concise as possible.

processor/src/test/java/org/mapstruct/ap/test/value/DefaultOrderMapper.java
sjaakd: > Or am I missing something?

I think so. Currently source and target are also matched based on name. If source name == target name MapStruct automatically adds a mapping. With the introduction of "*" "default" or whatever we call it, we suddenly have a change of behavior according to your statement above.  Should the name based mapping not be executed now?

Use case: a user has specified only the differences between source and target values and relies on MapStruct to complete those based on name. This works. At one point he/she realizes that a default value was forgotten. So the line with the "*" is added. Suddenly all ends up in default.

Now: my suggestion would be to have to completers.. "*" and "+" or something like that to signify the difference.

sjaakd: Or perhaps even better: 

```
@ValueMapping( source = "EXTRA", target = "SPECIAL" )
@ValueMapping( source = "#", target = "#" )
@ValueMapping( source = "*", target = "DEFAULT" )
```
(Better character than `#` could be `?`).. Nice thing.. This would also work with String to String mappings (of course with escaping there)..

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
sjaakd: duh.

sjaakd: At least we know for sure now its off..

processor/src/test/java/org/mapstruct/ap/test/value/DefaultOrderMapper.java
gunnarmorling: Ah, sorry, I see now that my comment was a bit misleading.

I meant to leave the name-based default as is: I.e. if source and target match, they would be mapped by default. `source="*"` would only refer to those that don't have a match with the same name on the target side nor are mapped explicitly by another mapping.

But I see where you are coming from; It might be that `source="*"` is too easy understood as "match anything not explicitly mentioned in other mappings".

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -107,7 +105,13 @@
     INHERITCONFIGURATION_DUPLICATE_MATCHES( "Given name \"%s\" matches several candidate methods: %s." ),
     INHERITCONFIGURATION_NO_NAME_MATCH( "Given name \"%s\" does not match the only candidate. Did you mean: \"%s\"." ),
     INHERITCONFIGURATION_MULTIPLE_PROTOTYPE_METHODS_MATCH( "More than one configuration prototype method is applicable. Use @InheritConfiguration to select one of them explicitly: %s." ),
-    INHERITCONFIGURATION_CYCLE( "Cycle detected while evaluating inherited configurations. Inheritance path: %s" );
+    INHERITCONFIGURATION_CYCLE( "Cycle detected while evaluating inherited configurations. Inheritance path: %s" ),
+
+    VALUEMAPPING_DUPLICATE_SOURCE( "Source value mapping: \"%s\" cannot be mapped more than once." ),
+    VALUEMAPPING_INCORRECT_NAME_BASE_CONTINUEATION( "Source == \"?\" can only be used in combination with target == \"?\"." ),
```
sjaakd: ok in the end I choose '$' need to fix.

sjaakd:
So.. I tried to incorporate our latest ideas..  I think we're almost there in our discussion.

screen shot 2016-03-07 at 00 00 05

WDYT?

parent/pom.xml
```diff
@@ -211,6 +212,13 @@
                 <version>3.0.20</version>
             </dependency>
 
+            <!-- Doclet for inclusion of code snippets -->
+            <dependency>
+                <groupId>org.apidesign.javadoc</groupId>
+                <artifactId>codesnippet-doclet</artifactId>
+                <version>${org.apidesign.javadoc.codesnippit.version}</version>
+            </dependency>
```
agudian: This is not required here, as the version needs to be managed at the javadoc plugin dependencies.

core/pom.xml
```diff
@@ -69,6 +69,9 @@
                         </goals>
                     </execution>
                 </executions>
+                <configuration>
+                    <excludes>**/test/**/*.java</excludes>
+                </configuration>
```
agudian: Can be removed again, right?

sjaakd: Yes.. Its relic from a past solution

core/pom.xml
```diff
@@ -140,6 +143,33 @@
...
+                <executions>
+                    <execution>
+                        <id>include-snippets</id>
+                        <phase>generate-test-resources</phase>
+                        <goals>
+                            <goal>javadoc</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${project.build.directory}/test-classes/</outputDirectory>
+                            <reportOutputDirectory>${project.build.directory}/test-classes/</reportOutputDirectory>
```
agudian: hmm? to test-classes? That was just for playing around, right?

parent/pom.xml
sjaakd: True! Plugin dependencies are not managed by dependency management, perhaps I should include it as default java doc plugin management .

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -107,7 +105,13 @@
     INHERITCONFIGURATION_DUPLICATE_MATCHES( "Given name \"%s\" matches several candidate methods: %s." ),
     INHERITCONFIGURATION_NO_NAME_MATCH( "Given name \"%s\" does not match the only candidate. Did you mean: \"%s\"." ),
     INHERITCONFIGURATION_MULTIPLE_PROTOTYPE_METHODS_MATCH( "More than one configuration prototype method is applicable. Use @InheritConfiguration to select one of them explicitly: %s." ),
-    INHERITCONFIGURATION_CYCLE( "Cycle detected while evaluating inherited configurations. Inheritance path: %s" );
+    INHERITCONFIGURATION_CYCLE( "Cycle detected while evaluating inherited configurations. Inheritance path: %s" ),
+
+    VALUEMAPPING_DUPLICATE_SOURCE( "Source value mapping: \"%s\" cannot be mapped more than once." ),
+    VALUEMAPPING_INCORRECT_NAME_BASE_CONTINUEATION( "Source == \"$\" can only be used in combination with target == \"$\"." ),
```
agudian: Use `source = "$"` instead of `==` as it is displayed at the location of the assignment.

core/pom.xml
sjaakd: yep..

sjaakd:
Ok... I processed the comments along the lines of the [gist](https://gist.github.com/gunnarmorling/114036775b0c40fa9142). Please have a look. 

I did not add the warning for enum mappings in the current `@Mapping` yet. Most likely the unit test breaks. We first need to decide what to do with that UT (scrap it and only test the warning?). I'll squash the commits into one later.

sjaakd:
Please checkout #776 .. This supersedes this PR (left open for the comments, but can be closed when ready)

Issue #770, opened 2016-03-04
Qualified String Mapper sometimes used for general String mapping


gbrehmer:
I did not have a testcase to reproduce this error yet.

Sometimes a String-To-String mapping method with special Qualifier will be selected for all String-to-String mapping cases, where Qualifier is not set.

A clean project eclipse build fixes the problems. In addition these problems never occur during command line build process (or builds on jenkins).

My Environment: MacOS, Java 1.8.0_31, JHipster 2.27.0 (mapstruct 1.0.0), STS 3.5.2 / eclipse 4.5.1, gradle, Gradle eclipse Plugin & self configured APT settings on project level.

agudian:
Does your qualifier annotation have retention=CLASS defined? If not, it would explain what you observe.

gbrehmer:
I'm using retention=Source like in the official documentation.

```@Qualifier
@Target(ElementType.METHOD)
@Retention(RetentionPolicy.SOURCE)
public @interface Password {

}

gunnarmorling:
Ok, could you use CLASS then and let us know whether this resolves the
issue for you?

Depending on the state of (partial) compilations in Eclipse, SOURCE-level
annotations cannot be read. We should update the docs to recommend CLASS by
default.

gunnarmorling:
Going to close this, as Andreas has fixed the examples/docs to use retention `CLASS`. Feel free to re-open if the issue remains also if you are using `CLASS`, too.

gbrehmer:
Yes, thanks, changing retention to `CLASS` works as expected

gunnarmorling:
Cool, glad to hear and thanks for reporting back!

PR #772, opened 2016-03-07
#770 Update retention-policy of custom qualifier annotations in docs and tests

agudian:
… to `CLASS`, as retention policy `SOURCE` is not enough when working with multi-module projects or with incremental compilation.


gunnarmorling:
Rebased and merged. Thanks, @agudian! Really nice to have the docs as part of the main project now :)

PR #776, opened 2016-03-08
#749 Introduction of ValueMappings (enum-to-enum-mapping only)  squashed

sjaakd:
@gunnarmorling , @agudian .. did not want to destroy the history in PR #764 .. So here's a clean one, squashed all the prior commits/rework and adapted to our telco.

Have a look at the introduction of the `MappingConstants` class. I first added the constants to the `@Mapper` itself, later to the `@ValueMappings`, however, javadoc does not like it (errors). So this leaves room for future additions of constants and hopefully harmonizes constant names.

I also added the deprecation warning.

### DIFF
--- a/core-common/src/main/java/org/mapstruct/MappingConstants.java
+++ b/core-common/src/main/java/org/mapstruct/MappingConstants.java
@@ -1,0 +1,46 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+/**
+ * Contains all constants defined in the mapping process.
+ *
+ * @author Sjaak Derksen
+ */
+public final class MappingConstants {
+
+    private MappingConstants() {
+    }
+
+    /**
+     * Represents a {@code null} source or target.
+     */
+    public static final String NULL = "<NULL>";
+
+    /**
+     * Represents any source that is not already mapped by either a defined mapping or by means of name based mapping.
+     */
+    public static final String ANY_REMAINING = "<ANY_REMAINING>";
+
+    /**
+     * Represents any source that is not already mapped by a defined mapping.
+     */
+    public static final String ANY_UNMAPPED = "<ANY_UNMAPPED>";
+
+}
--- a/core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
+++ b/core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
@@ -1,0 +1,123 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Configures the mapping of source constant value to target constant value.
+ * <p>
+ * Supported mappings are
+ * <ol>
+ * <li>Enumeration to Enumeration</li>
+ * </ol>
+ * <p>
+ * <B>Example 1:</B>
+ * <pre>
+ * <code>
+ * public enum OrderType { RETAIL, B2B, EXTRA, STANDARD, NORMAL }
+ *
+ * public enum ExternalOrderType { RETAIL, B2B, SPECIAL, DEFAULT }
+ *
+ * &#64;ValueMapping(source = "EXTRA", target = "SPECIAL"),
+ * &#64;ValueMapping(source = "STANDARD", target = "DEFAULT"),
+ * &#64;ValueMapping(source = "NORMAL", target = "DEFAULT")
+ * ExternalOrderType orderTypeToExternalOrderType(OrderType orderType);
+ * </code>
+ * Mapping result:
+ * ╔═════════════════════╦════════════════════════════╗
+ * ║ OrderType           ║ ExternalOrderType          ║
+ * ╠═════════════════════╬════════════════════════════╣
+ * ║ null                ║ null                       ║
+ * ║ OrderType.EXTRA     ║ ExternalOrderType.SPECIAL  ║
+ * ║ OrderType.STANDARD  ║ ExternalOrderType.DEFAULT  ║
+ * ║ OrderType.NORMAL    ║ ExternalOrderType.DEFAULT  ║
+ * ║ OrderType.RETAIL    ║ ExternalOrderType.RETAIL   ║
+ * ║ OrderType.B2B       ║ ExternalOrderType.B2B      ║
+ * ╚═════════════════════╩════════════════════════════╝
+ * </pre>
+ * MapStruct will <B>WARN</B> on incomplete mappings. However, if for some reason no match is found an
+ * {@link java.lang.IllegalStateException} will be thrown.
+ * <p>
+ * <B>Example 2:</B>
+ * <pre>
+ * <code>
+ * &#64;ValueMapping( source = "&lt;NULL&gt;", target = "DEFAULT" ),
+ * &#64;ValueMapping( source = "STANDARD", target = "&lt;NULL&gt;" ),
+ * &#64;ValueMapping( source = "&lt;ANY_REMAINING&gt;", target = "SPECIAL" )
+ * ExternalOrderType orderTypeToExternalOrderType(OrderType orderType);
+ * </code>
+ * Mapping result:
+ * ╔═════════════════════╦════════════════════════════╗
+ * ║ OrderType           ║ ExternalOrderType          ║
+ * ╠═════════════════════╬════════════════════════════╣
+ * ║ null                ║ ExternalOrderType.DEFAULT  ║
+ * ║ OrderType.STANDARD  ║ null                       ║
+ * ║ OrderType.RETAIL    ║ ExternalOrderType.RETAIL   ║
+ * ║ OrderType.B2B       ║ ExternalOrderType.B2B      ║
+ * ║ OrderType.NORMAL    ║ ExternalOrderType.SPECIAL  ║
+ * ║ OrderType.EXTRA     ║ ExternalOrderType.SPECIAL  ║
+ * ╚═════════════════════╩════════════════════════════╝
+ * </pre>
+ *
+ * @author Sjaak Derksen
+ */
+@Repeatable(ValueMappings.class)
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.METHOD)
+public @interface ValueMapping {
+    /**
+     * The source value constant to use for this mapping.
+     *
+     * <p>
+     * <b>Valid values:</b>
+     * <ol>
+     * <li>enum constant name</li>
+     * <li>{@link MappingConstants#NULL}</li>
+     * <li>{@link MappingConstants#ANY_REMAINING}</li>
+     * <li>{@link MappingConstants#ANY_UNMAPPED}</li>
+     * </ol>
+     * <p>
+     * <b>NOTE:</b>When using &lt;ANY_REMAINING&gt;, MapStruct will perform the normal name based mapping, in which
+     * source is mapped to target based on enum identifier equality. Using &lt;ANY_UNMAPPED&gt; will not apply name
+     * based mapping.
+     *
+     * @return The source value.
+     */
+    String source();
+
+    /**
+     * The target value constant to use for this mapping.
+     *
+     * <p>
+     * <b>Valid values:</b>
+     * <ol>
+     * <li>enum constant name</li>
+     * <li>{@link MappingConstants#NULL}</li>
+     * </ol>
+     *
+     * @return The target value.
+     */
+    String target();
+
+}
--- a/core/src/main/java/org/mapstruct/ValueMapping.java
+++ b/core/src/main/java/org/mapstruct/ValueMapping.java
@@ -1,0 +1,126 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Configures the mapping of source constant value to target constant value.
+ * <p>
+ * Supported mappings are
+ * <ol>
+ * <li>Enumeration to Enumeration</li>
+ * </ol>
+ * <p>
+ * <B>Example 1:</B>
+ * <pre>
+ * <code>
+ * public enum OrderType { RETAIL, B2B, EXTRA, STANDARD, NORMAL }
+ *
+ * public enum ExternalOrderType { RETAIL, B2B, SPECIAL, DEFAULT }
+ *
+ * &#64;ValueMappings({
+ *    &#64;ValueMapping(source = "EXTRA", target = "SPECIAL"),
+ *    &#64;ValueMapping(source = "STANDARD", target = "DEFAULT"),
+ *    &#64;ValueMapping(source = "NORMAL", target = "DEFAULT")
+ * })
+ * ExternalOrderType orderTypeToExternalOrderType(OrderType orderType);
+ * </code>
+ * Mapping result:
+ * ╔═════════════════════╦════════════════════════════╗
+ * ║ OrderType           ║ ExternalOrderType          ║
+ * ╠═════════════════════╬════════════════════════════╣
+ * ║ null                ║ null                       ║
+ * ║ OrderType.EXTRA     ║ ExternalOrderType.SPECIAL  ║
+ * ║ OrderType.STANDARD  ║ ExternalOrderType.DEFAULT  ║
+ * ║ OrderType.NORMAL    ║ ExternalOrderType.DEFAULT  ║
+ * ║ OrderType.RETAIL    ║ ExternalOrderType.RETAIL   ║
+ * ║ OrderType.B2B       ║ ExternalOrderType.B2B      ║
+ * ╚═════════════════════╩════════════════════════════╝
+ * </pre>
+ * MapStruct will <B>WARN</B> on incomplete mappings. However, if for some reason no match is found an
+ * {@link java.lang.IllegalStateException} will be thrown.
+ * <p>
+ * <B>Example 2:</B>
+ * <pre>
+ * <code>
+ * &#64;ValueMappings({
+ *    &#64;ValueMapping( source = "&lt;NULL&gt;", target = "DEFAULT" ),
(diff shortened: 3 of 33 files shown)


core-common/src/main/java/org/mapstruct/MappingConstants.java
```diff
@@ -0,0 +1,47 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+public class MappingConstants {
```
agudian: You could make this class `final`. And add some Javadoc as well? Then you could leave out the prefix `Mapping contant, ..` in the javadocs for the constants below - it's kinda obvious then, I'd say, :)

core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
```diff
@@ -0,0 +1,125 @@
...
+ *    &#64;ValueMapping(source = "STANDARD", target = "DEFAULT"),
+ *    &#64;ValueMapping(source = "NORMAL", target = "DEFAULT")
+ * })
+ * ExternalOrderType orderTypeToExternalOrderType(OrderType orderType);
+ * </code>
+ * Mapping result:
+ * ╔═════════════════════╦════════════════════════════╗
+ * ║ OrderType           ║ ExternalOrderType          ║
+ * ╠═════════════════════╬════════════════════════════╣
+ * ║ &lt;NULL&gt;              ║ &lt;NULL&gt;                     ║
```
agudian: use only `null` instead of `&lt;NULL&gt;`?

core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
```diff
@@ -0,0 +1,125 @@
...
+ * </ol>
+ * <p>
+ * <B>Example 1:</B>
+ * <pre>
+ * <code>
+ * public enum OrderType { RETAIL, B2B, EXTRA, STANDARD, NORMAL }
+ *
+ * public enum ExternalOrderType { RETAIL, B2B, SPECIAL, DEFAULT }
+ *
+ * &#64;ValueMappings({
```
agudian: As this is the jdk8 variant of the file, we should leave out the `@ValueMappings` in the examples here...

core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
```diff
@@ -0,0 +1,125 @@
...
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.METHOD)
+public @interface ValueMapping {
+    /**
+     * The source value constant to use for this mapping.
+     *
+     * <p>
+     * <b>Valid values:</b>
+     * <ol>
+     * <li>ENUM IDENTIFIER</li>
```
agudian: "enum constant name" instead of "ENUM IDENTIFIER"?

agudian: or "name of the enum constant"?

processor/src/test/java/org/mapstruct/ap/test/value/ErroneousOrderMapperMappingSameConstantTwice.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.mapstruct.ap.test.value;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.ValueMapping;
+import org.mapstruct.ValueMappings;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Gunnar Morling
```
agudian: Who's the author here? :grin:

processor/src/test/java/org/mapstruct/ap/test/value/EnumToEnumMappingTest.java
```diff
@@ -0,0 +1,264 @@
...
+ *
+ * @author Gunnar Morling
+ */
+@IssueKey("128")
+@WithClasses({ OrderEntity.class, OrderType.class, OrderDto.class, ExternalOrderType.class })
+@RunWith(AnnotationProcessorTestRunner.class)
+public class EnumToEnumMappingTest {
+
+    @Test
+    @WithClasses( OrderMapper.class )
```
agudian: Could you combine all the non-erroneous mappers into the class-level `@WithClasses` annotation? That avoids some re-compilation cycles during the test run, which is handy for those of use with slower machines... :eyes:

processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
```diff
@@ -0,0 +1,292 @@
...
+            return source;
+        }
+
+        public String getTarget() {
+            return target;
+        }
+
+
+
+
```
agudian: some emtpy lines...

processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
```diff
@@ -0,0 +1,292 @@
...
+import org.mapstruct.ap.internal.model.source.SourceMethod;
+import org.mapstruct.ap.internal.model.source.ValueMapping;
+import org.mapstruct.ap.internal.prism.BeanMappingPrism;
+import static org.mapstruct.ap.internal.util.Collections.first;
+import org.mapstruct.ap.internal.util.Message;
+import org.mapstruct.ap.internal.util.Strings;
+
+/**
+ * A {@link MappingMethod} which maps one enum type to another, optionally configured by one or more
+ * {@link EnumMapping}s.
```
agudian: You mean `@ValueMapping` here, right?

processor/src/main/java/org/mapstruct/ap/internal/model/source/ValueMapping.java
```diff
@@ -0,0 +1,165 @@
...
+import org.mapstruct.ap.internal.util.Message;
+
+/**
+ * Represents the mapping between one value constant and another.
+ *
+ * @author Sjaak Derksen
+ */
+public class ValueMapping {
+
+    public static final String ANY = "<ANY>";
```
agudian: Would you add a unit test to verify that the constants here have the same value as the fields in `@MappingConstants`?

agudian: Hmm, or additionally put the constants in a class next to our existing enum-prisms (e.g. `MappingConstantsPrism`), to emphasise that it reflects the content of that other class in the API.

agudian:
I'm done with the re-review and found some more things to nag about :wink:. Other than that, it's great. Although there is no update to the docs included, yet.. :grin:. And we could prepare the migration notes already so that we don't forget it during the release...

sjaakd:
> I'm done with the re-review and found some more things to nag about . Other than that, it's great. Although there is no update to the docs included, yet.. . And we could prepare the migration notes already so that we don't forget it during the release...

Hey @agudian .. thanks for reviewing.. I'll process your comments. Of course I'll add some documentation as well. Just want to be sure we all agree on this feature before I put in additional work :wink: .

processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
sjaakd: ay.. Slip of the tongue..

processor/src/test/java/org/mapstruct/ap/test/value/EnumToEnumMappingTest.java
sjaakd: I changed it because its a bit harder to debug (figuring out which mapper fails). This is why I comment out the Eclipse version as well when I'm designing the UT. However, I can put them back now..

core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
sjaakd: I actually got it from the java spec (see https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.9.1), where the actual name is referred to as 'the identifier'  (and with access modifier + constructor argument is called enum constant. However, it does not any clarity. i'll change..

core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
sjaakd: Ok. I'll do so.

core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
sjaakd: you mean in the javadoc? Or as constant value..? Wouldn't the latter be deviating from the convention we now seem to introduce for <ANY>?

core-common/src/main/java/org/mapstruct/MappingConstants.java
sjaakd: That's true.. But in my experience making it just `Constant` would lead to name clashes with almost everybody that uses the same name.. Just try to do an automatic import on Constant and see the size of the list that pops up in an arbitrary project :smile:.

sjaakd: I'll make the class final and add some more javadocs..

processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
sjaakd: check!

processor/src/main/java/org/mapstruct/ap/internal/model/source/ValueMapping.java
sjaakd: Ok.. good suggestion..

processor/src/test/java/org/mapstruct/ap/test/value/ErroneousOrderMapperMappingSameConstantTwice.java
sjaakd: its called creating buy-in .. But to be honest, the test cases mainly come from his hand in the good-old-enum mapping methods.. So credits to who'm own them..

core-common/src/main/java/org/mapstruct/MappingConstants.java
agudian: > That's true.. But in my experience making it just Constant would lead to name clashes with almost everybody that uses the same name.. Just try to do an automatic import on Constant and see the size of the list that pops up in an arbitrary project :smile:.

Totally true. What I meant was the start of the javadocs of the constants (NULL, ANY, ...). They all start with the prefix "Mapping contant, ", which I think is redundant in that scope.

core-jdk8/src/main/java/org/mapstruct/ValueMapping.java
agudian: Yes, I just mean the wording in the javadoc here. You want to describe what input creates what output in the generated mapping method, and `"<NULL>"` is not the input here, but `null` is. Or did I misunderstand the meaning of that table?

sjaakd: you're right.. I'll be a bit more exact here and refer to the 'real values' (use * as well in the table).

sjaakd:
@agudian : I accidentally amended the processed comments on the last commit. But I've processed your comments and added documentation.
@gunnarmorling : can I merge this?

agudian:
Awesome, I'm happy now! :+1: 

Wanna wait for @gunnarmorling to take a final look before squashing again?

sjaakd:
Yep. I'm not sure whether he already started reviewing. But I'll wait for his final verdict . Thanks for the review.!

core-common/src/main/java/org/mapstruct/MappingConstants.java
```diff
@@ -0,0 +1,45 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+public final class MappingConstants {
```
gunnarmorling: A class comment would be great. Or just move the constants to `@ValueMapping` itself?

core-common/src/main/java/org/mapstruct/MappingConstants.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public static final String NULL = "<NULL>";
+
+    /**
+     * Represents any source that is not already mapped by either a defined mapping or by means of name based mapping.
+     */
+    public static final String ANY = "<ANY>";
+
+    /**
+     * Represents any source that is not already mapped by either a defined mapping.
```
gunnarmorling: The "either" is too much.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -1113,17 +1115,124 @@ public interface OrderMapper {
...
+            break;
+            default: throw new IllegalArgumentException( "Unexpected enum constant: " + orderType );
+        }
+
+        return externalOrderType_;
+    }
+}
+----
+====
+By default an error will be raised by MapStruct in case a constant of the source enum type does not have a corresponding constant with the same name in the target type and also is not mapped to another constant via `@ValueMapping`. This ensures that all constants are mapped in a safe and predictable manner. An IllegalStateException is thrown when for some reason an unrecognized source value occurs.
```
gunnarmorling: No need to change this right now, but could you prefer shorter lines in the future? I usually break after each sentence or after a "," (i.e. each "unit of thought" goes on one line) which works quite good. The long lines make it very tough to diff changes down the road.

gunnarmorling: "IllegalStateException" is raised? Shouldn't we create a proper error instead? Actually, never should exceptions be raised to the user from our end.

core-common/src/main/java/org/mapstruct/MappingConstants.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Represents a {@code null} source or target.
+     */
+    public static final String NULL = "<NULL>";
+
+    /**
+     * Represents any source that is not already mapped by either a defined mapping or by means of name based mapping.
+     */
+    public static final String ANY = "<ANY>";
```
gunnarmorling: Should we name it "<ANY_REMAINING>" or so? I am having the same feeling as you did with the "*", i.e. it can be understood to cover *anything*? It's cool with me, if you think there is no issue.

processor/src/test/java/org/mapstruct/ap/test/value/ErroneousOrderMapperMappingSameConstantTwice.java
gunnarmorling: :-)

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -320,8 +321,21 @@ else if ( method.isMapMapping() ) {
...
+                ValueMappingMethod valueMappingMethod = new ValueMappingMethod.Builder()
+                    .mappingContext( mappingContext )
+                    .souceMethod( method )
+                    .valueMappings( mappingOptions.getValueMappings() )
+                    .build();
+                mappingMethods.add( valueMappingMethod );
+            }
             else if ( method.isEnumMapping() ) {
 
+                messager.printMessage(
```
gunnarmorling: Nice with the deprecation. Is there still a test in place to make sure it still works for the time being?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -1113,17 +1115,124 @@ public interface OrderMapper {
...
+}
+----
+====
+    
+*Note:* MapStruct would have refrained from mapping the `RETAIL` and `B2B` when `<ANY_UNMAPPED>` was used instead of `<ANY>`.  
+
+
+[WARNING]
+====
+The mapping of enum to enum via the `@Mapping` annotation is *DEPRECATED*. It will be removed from future versions of MapStruct.
```
gunnarmorling: Can you add: "Please adapt existing enum mapping methods to make use of `@ValueMapping` instead".

core/src/main/java/org/mapstruct/ValueMapping.java
```diff
@@ -0,0 +1,124 @@
...
+ * </code>
+ * Mapping result:
+ * ╔═════════════════════╦════════════════════════════╗
+ * ║ OrderType           ║ ExternalOrderType          ║
+ * ╠═════════════════════╬════════════════════════════╣
+ * ║ null                ║ ExternalOrderType.DEFAULT  ║
+ * ║ OrderType.STANDARD  ║ null                       ║
+ * ║ OrderType.RETAIL    ║ ExternalOrderType.RETAIL   ║
+ * ║ OrderType.B2B       ║ ExternalOrderType.B2B      ║
+ * ║ *                   ║ ExternalOrderType.SPECIAL  ║
```
gunnarmorling: I think spelling out the "*" to the actual values makes it easier to follow.

gunnarmorling:
Hey @sjaakd, added some more thoughts, nothing special. Really great work overall! The solution is nice, I think the discussions we had payed off :+1:.

If you like, merge it yourself once having processed my last set of comments?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
agudian: I don't think this was changed in this PR, but I would expect the generated code to throw an Exception if the source enum was expanded without regenerating the code. I don't think there is something else can we do there? :-/

gunnarmorling: > generated code to throw an Exception if the source enum was expanded

Aaah, that's what this is about. I assumed it to be an exception from
within the processor (which is something we should never do). But yes, that
makes sense. The wording should be improved then a bit ("The generated
mapping method will raise an exception if ...").

core-common/src/main/java/org/mapstruct/MappingConstants.java
sjaakd: To be clear, you mean an empty String as constant? so: `public static final String ANY = ""`

core-common/src/main/java/org/mapstruct/MappingConstants.java
sjaakd: Ok.. fixed

core-common/src/main/java/org/mapstruct/MappingConstants.java
sjaakd: > Or just move the constants to @ValueMapping itself?

I did that initially. But ran into problems with the javadoc.. It does not seem to refer correctly to constants defined in annotations.

core-common/src/main/java/org/mapstruct/MappingConstants.java
gunnarmorling: Agh, no, the rendering was mixed up :/ I've updated it, please check again :)

core-common/src/main/java/org/mapstruct/MappingConstants.java
sjaakd: > Or just move the constants to @ValueMapping itself?

You don't want to run into a situation that you have to define them again in `@Mapping` because at one time in the future you might need them there as well (harmonization argument).

sjaakd: > Or just move the constants to @ValueMapping itself?

Last but not least: duplication between the java8-core an core..

sjaakd: > A class comment would be great

fixed.

core/src/main/java/org/mapstruct/ValueMapping.java
sjaakd: changed to any..

core-common/src/main/java/org/mapstruct/MappingConstants.java
agudian: :see_no_evil: - no code completion possible there... :grin: 
Still, I'd say "any" is a good wording for something that is supposed to match "anything". Alternatively we could go for "<ELSE>" and "<UNLESS_MATCHING>" or something along that way... But I guess we should just go ahead with the one solution we have now and not over-think it... So I'm all for leaving it up to Sjaak now as well :)

agudian: > Agh, no, the rendering was mixed up :/ I've updated it, please check again :)

Funny, that used to work before, right? My comment was garbled as well...

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: > No need to change this right now, but could you prefer shorter lines in the future? I usually break after each sentence or after a "," (i.e. each "unit of thought" goes on one line) which works quite good. The long lines make it very tough to diff changes down the road.

The first sentence is yours :smile:. Anyway.. I limited myself to the essential on the next line (1165) now.

sjaakd: > Aaah, that's what this is about. I assumed it to be an exception from
> within the processor (which is something we should never do). But yes, that
> makes sense. The wording should be improved then a bit ("The generated
> mapping method will raise an exception if ...").

Done!

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
sjaakd: Did you checkout the UT in the old enum package? Each test case there has: 

```java
    @ExpectedCompilationOutcome(
        value = CompilationResult.SUCCEEDED,
        diagnostics = {
            @Diagnostic(type = OrderMapper.class,
                kind = Kind.WARNING,
                line = 41,
                messageRegExp = "Mapping of Enums via @Mapping is going to be removed in future versions of "
                    + "MapStruct\\. Please use @ValueMapping instead!")
        }
    )
```

So I reckon all old test cases test this implicitly.

core-common/src/main/java/org/mapstruct/MappingConstants.java
sjaakd: > Should we name it "<ANY_REMAINING>" or so? I am having the same feeling as you did with the "", i.e. it can be understood to cover *anything? It's cool with me, if you think there is no issue.

The problem I have is that <ANY_UNMATCHED> is actually a larger set than <ANY>.. That feels strange.. I opt for <ANY_REMAINING> ( :+1: @gunnarmorling ). If there are no further objections ( @agudian ), or forever hold your peace.. :Smile:

agudian: *He who commits shall have thy way.*

sjaakd:
Ok.. one more final QC from our automated Jenkins build. Then I'll squash..

core-common/src/main/java/org/mapstruct/MappingConstants.java
gunnarmorling: Ok.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
gunnarmorling: Ok, sounds great.

Issue #775, opened 2016-03-08
Mapping method matching not working for lower-bounded target type


ugrave:
Example Objects to Map:

```
public class SourceObject {
        private List<String> numbers = new ArrayList<>();

        public List<String> getNumbers() {
            return numbers;
        }
 }
```

```
public class TargetObject {
        private List<Integer> numbers = new ArrayList<>();

        public TargetObject setNumbers(Iterable<? extends Integer> numbersToAdd) {
            for (Integer n : numbersToAdd) {
                numbers.add(n);
            }
            return this;
        }
}
```
The TargetObject is generated automatically by the [immutables](http://immutables.github.io/) framework.

The mapper looks like:

```
public abstract class TestMapper {

    public abstract TargetObject map(SourceObject source);

    public abstract List<Integer> mapToList(Collection<String> source);

    public Integer map(String value) {
        return Integer.valueOf(value);
    }
}
```

On mapper generation mapstructs generates an error on map method:
`Can't map property "java.util.List<java.lang.String> numbers" to "java.lang.Iterable<? extends java.lang.Integer> numbers". Consider to declare/implement a mapping method: "java.lang.Iterable<? extends java.lang.Integer> map(java.util.List<java.lang.String> value)".`

If i'm adding the following mapping method, everything is working:
```
public List<? extends Integer> mapToIterable(List<String> source) {
   return mapToList(source);
}
```
The generated mapper uses the mapToIterable to set the property on the target object. (This is my current solution i used for solve the problem)

I also try to change the return value ofthe mapToList method to `List<? extends Integer>`. But then i get the error:

```
Can't generate mapping method for a wildcard extends bound result
```

I have also try to define the mapping method to use with the IterableMapping annotation. But that does not help.

agudian:
Fixed. Thanks for letting us know, Ulrich! :+1:

amineRS2i:
Please can you tell us your solution !!!

gunnarmorling:
@amineRS2i, this has been fixed in the 1.1.0.Beta1 release. Is this the version you are using?

amineRS2i:
yes i i use this version, but i have open a ticket for my problem : 
[(https://github.com/mapstruct/mapstruct/issues/800)]
can you help me please ?

PR #783, opened 2016-03-14
Fix for #775: Iterable forged methods and lower bounds

agudian:
#775 actually uncovers two problems, I tried to fix them in two separate commits. The creation of the forged method for Iterables is straight forward I guess, but the lower-bound matching is a bit voodooish... I didn't get it to break though while trying around a bit. We still might have similar issues with upper-bounds, in case anyone wants to try that out in detail... :smirk:

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_775/IterableContainer.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_775/IterableContainer.java
@@ -1,0 +1,34 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_775/IterableContainer.java
```diff
@@ -0,0 +1,34 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.bugs._775;
+
+/**
+ * @author Andreas Gudian
+ */
+public class IterableContainer {
+    private Iterable<? extends Integer> values;
```
gunnarmorling: LOL, `Integer` is final, but I guess for the sake of the test... ;)

gunnarmorling:
LGTM, I'll merge once the enum stuff is in.

gunnarmorling:
> We still might have similar issues with upper-bounds, in case anyone wants to try that out in detail...

Ok, I'd say let's take these as they are actually reported. It's a continuous process here :)

sjaakd:
> LGTM, I'll merge once the enum stuff is in.

Its in...

PR #794, opened 2016-04-06
#796 Added files via upload

maposchmi:


agudian:

agudian:
@maposchmi The pull-request isn't really in a state that works for us (for once, the classes ended up in a different directory).

But the feature that you describe is definitely worth following up. Would you be willing to further work on this?

In that case, I'd suggest you make yourself a little more familiar with git and github first (learning curve is not that steep and there are great articles directly here on github and on [git-scm.com](https://git-scm.com/documentation) that help you getting started with git. That way you'd be able to properly commit the changes you make, and then create a pull-request from that branch. If you're more into GUI clients (like me), I can recommend [SmartGit](http://www.syntevo.com/smartgit/download). Once you're a little more up to speed with git and github, we can further talk about the actual content of the proposed changes. For example where we'd expect some Javadocs and where we suggest adding tests... :smile:

sjaakd:
I reckon I can close this one. @maposchmi : as @agudian pointed out, let us know if you want  to continue on this one.

PR #808, opened 2016-05-03
#751 Statefull Mappers and Mapper Factories

sjaakd:
Most of the approach is discussed in the issue text of #751.

So what's missing:
1. Review.. Before I continue, is this the right approach?
2. TODO's left in code. Scenarios which are not covered. How far should I go in this PR?
3. Unit tests. My plan is to extend the UT in the current package and,
4. Include a new UT package with mapper factory testing.


sjaakd:
Hmm.. something seems to go wrong with the Prisms??

agudian:
The type `HandleAsMapper` seems to be missing... :)

sjaakd:
Hmm! That explains! Will fix tomorrow evening!

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -116,6 +116,13 @@
     Class<?> config() default void.class;
 
     /**
+     * Factory for constructing the generated mapper when mapper is an abstract class.
+     *
+     * @return
+     */
+    Class<?> factory() default void.class;
```
sjaakd: Hmm. this is a relic of the past. Must remove

agudian:
retest this, please

agudian:
restest this, please

agudian:
retest this, please

agudian:
retest this please (sorry for the noise, @sjaakd - I'm trying to fix the PRB job...)

agudian:
retest this, please

agudian:
retest this, please

sjaakd:
😄! No problem. Enjoying a nice 🍻 while camping! A tad cold though... 😉

Verstuurd vanaf mijn iPhone

agudian:
retest this, please

agudian:
retest this, please

gunnarmorling:
@sjaakd, so I've been thinking a bit more about this one. Couldn't you simply declare the dependencies to stateful _used mappers_ as fields in an abstract mapper contract and we just add means to signal MapStruct to re-use those references instead of adding fields to the generated implementation. That way you'd be in charge of setting up the references in the state you need them. Wouldn't that be much simpler? Or am I missing something?

sjaakd:
So, how would that look like? Can you give an example of the code and generated code?

How, would you call such state full mapper from another mapper and provide the state?

gunnarmorling:
The mapper contract would look like this:

    @Mapper
    public abstract class MyMapper {

        protected final MyStatefulMapper msm = new MyStatefulMapper( <some-state-parameter );

        public abstract OrderDto orderToDto(Order order);
    }

I.e. you'd be in charge yourself of passing along / initializing the required state when instantiating the field within the abstract mapper type.

We may either consider all methods available through such manually set up field as available mapping methods, or we may tag the fields specifically:

    @Mapper
     public abstract class MyMapper {

        @UsedMapper
        protected final MyStatefulMapper msm = new MyStatefulMapper( <some-state-parameter );

        public abstract OrderDto orderToDto(Order order);
    }

Would that work for your purposes, @sjaakd?

sjaakd:
So, where's `<some-state-parameter` coming from? 

What if MyStatefulMapper is not a hand-written mapper, but also a generated one. I don't have access to the constructor of the generated implementation. Would you then still have the current factory mechanism?

BTW: I remember :smile: (and I think you found it too). The reason I've this `@UsedMapper` is to distinguish fields not being a mapper from fields that should treated as mapper and be analyzed for finding matching mapping methods.

PR #741, opened 2016-02-04
#740 baseline update to version 2.19.1 of Surefire plugin

sjaakd:

PR #811, opened 2016-05-22
#741 number format for number to string conversions

sjaakd:


sjaakd:
* String to Number type and vice versa by means of number formatter
* String to Primitive type and vice verse by means of number formatter
* Mechanism of HelperMethods introduced (similar like BuiltInMethod and ForgedMethod).
* FormattingParameters grouped, just like SelectionParameters.

sjaakd:
please test this

agudian:
retest this, please.
(pull-request-builder is now fixed, I've just updated the plugin in Jenkins - there was a regression due to a security fix in one of the last Jenkins updates).

sjaakd:
ah.. nice.. See stuff coloring green now.. :smile:

agudian:
reviewed. Looks good.

sjaakd:
Ok. fixed commit history after some git magic.. Took me 2 hours, but on the upside that I do understand now a bit more on how git works :smile:

core-common/src/main/java/org/mapstruct/IterableMapping.java
```diff
@@ -47,6 +47,14 @@
     String dateFormat() default "";
 
     /**
+     * A format string as processable by {@link java.text.DecimalFormat} if the annotated method maps from a
```
gunnarmorling: Minor nit-picking: Can you import the type and use the short name here? That reads nicer in rendered docs.

processor/src/main/java/org/mapstruct/ap/internal/model/HelperMethod.java
```diff
@@ -0,0 +1,240 @@
...
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.model.source.Method;
+import org.mapstruct.ap.internal.util.MapperConfiguration;
+import org.mapstruct.ap.internal.util.Strings;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+public abstract class HelperMethod implements Method {
```
gunnarmorling: Could you add a comment of what it is?

processor/src/main/java/org/mapstruct/ap/internal/model/source/FormattingParameters.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.mapstruct.ap.internal.model.source;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+public class FormattingParameters {
+
+    private final String date;
```
gunnarmorling: Is there a need to keep the two separately internally? Couldn't we just pass a single string instead of this wrapper?

processor/src/main/java/org/mapstruct/ap/internal/model/source/MapMapping.java
```diff
@@ -78,36 +80,45 @@ public static MapMapping fromPrism(MapMappingPrism mapMapping, ExecutableElement
             mapMapping.valueQualifiedByName(),
             valueTargetTypeIsDefined ? mapMapping.valueTargetType() : null);
 
-        return new MapMapping(
+        FormattingParameters keyFormatting = new FormattingParameters(
             mapMapping.keyDateFormat(),
-            keySelection,
+            mapMapping.keyNumberFormat() );
```
gunnarmorling: Do we prevent somewhere that date *and* number format are given at the same time?

gunnarmorling:
@sjaakd looks good to me overall. Only the meaning of "helper methods" is not quite clear to me; Some commenting would help. Check out the inline comments, nothing fancy. Good to merge after that!

Thanks for driving this home!

core-common/src/main/java/org/mapstruct/IterableMapping.java
sjaakd: done

processor/src/main/java/org/mapstruct/ap/internal/model/source/FormattingParameters.java
sjaakd: Does it clarify the code more to transport an agnostic formatting String? Besides, I suspect that there will be more of these formatting parameters (just as selection parameters) in the future. And they are mutual exclusive now, but new ones might not.

processor/src/main/java/org/mapstruct/ap/internal/model/source/MapMapping.java
sjaakd: No. I reckon its (silently) ignored. Perhaps such checking should be in relation to the source type and not in relation to each-other.

core-common/src/main/java/org/mapstruct/IterableMapping.java
sjaakd: didn't even know that was possible (having imports just for the sake of javadoc and no compiler warnings) 👍

sjaakd:
> Thanks for driving this home!

No problem. I did not honor all your remarks, but answered the questions. I you agree I can squash and merge this one.

sjaakd:
@cliedeman : many thanks for your work on this PR.

PR #821, opened 2016-06-29
#669 Support for presence check, based on #720

sjaakd:
@agudian .. I'm a bit stuck. .Will continue tomorrow. Think that this is your expertise area. I'm trying to add a service via a new annotation `@WithService`. It will place a new entry in the META-INF/services/. This works. See presence check test case. Its placed in the `classes org.mapstruct.ap.test.presenceCheckWithCustom_jdk/classes/META-INF/services/org.mapstruct.ap.spi.AccessorNamingStrategy`. However, its not picked up by the test runner.

Not sure how to continue (yet).


sjaakd:
> Not sure how to continue (yet).

Hmm. I just realized, I need to add this to the compiler 'StandardLocation.SOURCE_PATH'   and not to 'StandardLocation.SOURCE_OUTPUT' or 'CLASS_OUTPUT'

Perhaps I should generate a resources directory for these and point the compiler towards this directory.

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
```diff
@@ -16,7 +16,7 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  */
-package org.mapstruct.ap.internal.naming;
+package org.mapstruct.ap.spi;
```
agudian: Hehe... This is what I suggested back then when we introduced the interface, but we decided to not expose the default implementation - I don't fully remember the reasons. But I'm fine with it as it is now.

agudian:
The compiler doesn't do anything with a resources directory, so just puttng it in the `CLASS_OUTPUT` would be fine.
However... We're currently not using a fresh classloader for each invocation of the compiler, so any MapStruct classes are loaded only once. In `Executables`, we hold the `AccessorNamingStrategy` in a `static final` field. And in `Services`, we also cache it (I wonder why...).
Anyway, to make this work, we'd have to jump some more hoops, e.g. using a new class-loader hierarchy in case services are used. I'd like to give it a try, but not today anymore... :).
In the meantime, this would need to be tested in an integration test. See `NamingStrategyTest`, which already does it.

sjaakd:
> The compiler doesn't do anything with a resources directory, so just puttng it in the CLASS_OUTPUT would be fine.

That didn't work. Your next remark, explains why :) . ' We're currently not using a fresh classloader' When add a CP to the `ModifiableURLClassLoader` I'm able get the `PresenceCheckAccessorNamingStrategy`, but only when I initialize `ServiceLoader` with the `Thread.currentThread.contextClassLoader()` iso the class loader of an arbitrary class in the Mapstruct processor.

BTW: Why don't we use the context class loader? Is there a risk in using this iso an arbitrary class loader?

> In Executables, we hold the AccessorNamingStrategy in a static final field. And in Services, we also cache it (I wonder why...).

Moreover, why cache at all? I've been digging deep into the ServiceLoader code. I think its cached there as well 😄 Feels like `Services` is redundant code.

Reckon this is sufficient:
```java
        Iterator<T> services = ServiceLoader.load( serviceType, Services.class.getClassLoader() ).iterator();

        if ( services.hasNext() ) {
            return services.next();
        }
        else {
            return defaultValue;
        }
```


> Anyway, to make this work, we'd have to jump some more hoops, e.g. using a new class-loader hierarchy in case services are used. I'd like to give it a try, but not today anymore... :).

I would like to be able to control this from the UT. I can understand that you don't want to try this today 😄 . Proposal: if you find time to do this, I can work and continue (temporary) on the context class loader fix, debugging and making a PR. We can remove this by a decent solution before merging. 

> In the meantime, this would need to be tested in an integration test. See NamingStrategyTest, which already does it.

I know. But I need some more control for debugging all kinds of cases. So my preference goes to adding services similarly as `WithClasses` (see in the WIP).

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
sjaakd: I think that Hibernate Validator does something similar for their SPI default implementations. Anyway, I put in the ' hasser' as a ready-made alternative. That helps for my own testing as well and serves as example at the same moment.

sjaakd:
more to follow (mainly test cases and activating the new strategy). Furthermore a fresh compiler is needed for a unit test case. Otherwise, I cannot place a service on the class path.

agudian:
@sjaakd I got the thing with `@WithService` working with the Jdk-Compiler, and I'd like to give it a shot with the Eclipse-Compiler as well. I'll continue with that tomorrow evening...

sjaakd:
Gr8!! Thanks. I need another day to finish the PR. Now looking into nesting.

Verstuurd vanaf mijn iPhone

sjaakd:
Hey guys.. This one is ready for review as well. With thanks to @agudian .

sjaakd:
One of the things I spotted: I reckon that `NestedPropertyMappingMethod` can be improved. It generates unneeded checks icm `SourcePresenceChecks`. However, fixing this would mean bringing up more information (such as the strategy) to `NestedPropertyMappingMethod`. I'm not sure whether that's desirable from code understandibillity perspective.

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
gunnarmorling: I think we didn't feel like exposing that default implementation to users as it will make it more difficult to change it in the future. But if you guys think it's ready for that, we can do it.

core-common/src/main/java/org/mapstruct/SourceValuePresenceCheckStrategy.java
```diff
@@ -0,0 +1,58 @@
...
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+/**
+ * Strategy to decide how to check null
+ *
+ * @author Sean Huang
+ */
+public enum SourceValuePresenceCheckStrategy {
```
gunnarmorling: As it's now only dealing with null checks, should it be named `NullCheckStrategy`?

core-common/src/main/java/org/mapstruct/MapperConfig.java
```diff
@@ -135,4 +137,13 @@
...
         default MappingInheritanceStrategy.EXPLICIT;
+
+    /**
+     * Determines when to include a null check on the source property value of a bean mapping.
+     *
+     * Can be overridden by the one on {@link Mapper} or {@link Mapping}.
+     *
+     * @return strategy how to do null checking
+     */
+    SourceValuePresenceCheckStrategy sourceValuePresenceCheckStrategy() default ON_IMPLICIT_CONVERSION;
```
gunnarmorling: Somewhere here we should mention that a value presence check method will take precedence if present, and that this setting will not apply in this case.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -296,63 +291,106 @@ private Assignment getDefaultValueAssignment() {
             return null;
         }
 
-        private Assignment assignObject(Type sourceType, Type targetType, TargetWriteAccessorType targetAccessorType,
-                                        Assignment rhs) {
+        /**
+         *  assigns
```
gunnarmorling: These JavaDoc tags can go away.

processor/src/main/java/org/mapstruct/ap/internal/services/Services.java
```diff
@@ -29,53 +28,18 @@
...
-    private static <T> T find(Class<T> spi) {
-        T matchingImplementation = null;
-
-        for ( T implementation : ServiceLoader.load( spi, spi.getClassLoader() ) ) {
-            if ( matchingImplementation == null ) {
-                matchingImplementation = implementation;
-            }
-            else {
-                throw new IllegalStateException(
-                    "Multiple implementations have been found for the service provider interface "
```
gunnarmorling: Hum, seems we won't raise an exception anymore if there are multiple service implementations? That check was quite meaningful IMO.

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
```diff
@@ -92,8 +97,8 @@ private static boolean isPublic(ExecutableElement method) {
         return method.getModifiers().contains( Modifier.PUBLIC );
     }
 
-    public static String getPropertyName(ExecutableElement getterOrSetterMethod) {
-        return ACCESSOR_NAMING_STRATEGY.getPropertyName( getterOrSetterMethod );
+    public static String getPropertyName(ExecutableElement getterOrHasserOrSetterMethod) {
```
gunnarmorling: "hasser"="presenceChecker"?

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
```diff
@@ -38,7 +36,7 @@
  */
 public class DefaultAccessorNamingStrategy implements AccessorNamingStrategy {
 
-    @Override
+     @Override
     public MethodType getMethodType(ExecutableElement method) {
```
gunnarmorling: That seems odd.

processor/src/main/java/org/mapstruct/ap/spi/PresenceCheckAccessorNamingStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.type.TypeKind;
+
+/**
+ * The default JavaBeans-compliant implementation of the {@link AccessorNamingStrategy} service provider interface.
+ *
+ * @author Sjaak Derksen
+ */
+public class PresenceCheckAccessorNamingStrategy
```
gunnarmorling: How comes that's a separate sub-class?

gunnarmorling:
@sjaakd Added some more minor comments but I think we are there? Anything remaining to discuss from your side? If not, awesome achievement :)

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
sjaakd: I reckon its good enough. Besides, as long as we don't make functional changes.. Remember that its also good to have examples. Although I think that most developers are smart enough to look one package further :smile:.

processor/src/main/java/org/mapstruct/ap/internal/services/Services.java
sjaakd: repaired this. Also moved `Services`to the `util` package.

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
sjaakd: done.

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
sjaakd: What do you mean? It implements the interface?

processor/src/main/java/org/mapstruct/ap/spi/PresenceCheckAccessorNamingStrategy.java
sjaakd: Reusing `DefaultAccessorNamingStrategy`.. Or do you prefer composition here?

core-common/src/main/java/org/mapstruct/SourceValuePresenceCheckStrategy.java
sjaakd: But it does apply to the source only. So what about `SourceNullCheckStrategy`?

core-common/src/main/java/org/mapstruct/MapperConfig.java
sjaakd: Ok. I'll extend the javadoc. Will also do doc updates in separate PR.

core-common/src/main/java/org/mapstruct/SourceValuePresenceCheckStrategy.java
sjaakd: Ok.. Perhaps it that's the only logical place were to apply it to. So `NullCheckStrategy` is ok to me

sjaakd: we do have a `NullValueMappingStrategy`.  So to stay in line.. what about `NullValueCheckStrategy`? (You see, I'm doubting myself a bit here)

sjaakd:
> Anything remaining to discuss from your side?

One thing.. I decide to name `SourcePresenceCheckStrategy` to `NullValueCheckStrategy` iso `NullCheckStrategy` as you proposed. Think this is more in line with the already existing `NullValueMappingStrategy`. WDYT?

sjaakd:
> If not, awesome achievement :)

Many thanks to elegant solution of @agudian and of course to the work carried out by @seanjob .. Speaking of which, I should add him to the contributors I reckon.

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
gunnarmorling: Yes, of course one always could copy it. But making it part of the public API/SPI imposes stricter compatability requirements onto us. That said, its behavior was guaranteed more or less also before, so probably no big difference.

core-common/src/main/java/org/mapstruct/SourceValuePresenceCheckStrategy.java
gunnarmorling: Works both for me. `NullValueCheckStrategy` seems nice considering the consistency argument.

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
gunnarmorling: Yepp, the alignment. Sorry, was one line off.

processor/src/main/java/org/mapstruct/ap/spi/PresenceCheckAccessorNamingStrategy.java
gunnarmorling: I am wondering why you did not simply add it to the existing class?

gunnarmorling:
> Speaking of which, I should add him to the contributors I reckon.

+1.

Ah, also a doc update is needed.

processor/src/main/java/org/mapstruct/ap/spi/PresenceCheckAccessorNamingStrategy.java
sjaakd: I think that that came from one of your previous comments (on @seanjob 's implementation).. But I'm not sure anymore. I think you did not want to make a definite choice for `has`.

So: in the current solution: even if you implement `has` methods, they will not be used for presence checking unless you make a deliberate choice. In that case: you have one predefined.

agudian: I'm all for moving this into the default implementation. Otherwise this would be an unused class unless someone puts it into a service-loader file, right?

sjaakd: > I'm all for moving this into the default implementation. Otherwise this would be an unused class unless someone puts it into a service-loader file, right?

That's right..  But is that a problem? Hibernate Validator has for instance a `ValueFormatterMessageInterpolator` which is not used by default (I believe). Not that this makes it right to do so. 

Anyway, if you look at it from the other side: what if you have accidentally a `has` property method.. Would you expect MapStruct to call it or do you want explicit control over that?

sjaakd:
@gunnarmorling , @agudian : there's one discussion topic open on whether to include the `has` check in the default behavior. See above. Can we come to an agreement so I continue the merge?

agudian:
What I understood from our discussion and agreement two weeks ago was that we discover and use the hasX-methods by default and therefore don't add an option in an annotation for it. Only if someone wants to use a different pattern (or disable it completely) they'd have to implement their own strategy and activate it via ServiceLocator. I still think that is a convenient and pretty safe approach. 
If we would not make this a default behavior, I would argue that an annotation option would be more consistent. Having to add the service-locator file to the processor classpath in order to activate a built-in feature feels a bit odd.

gunnarmorling:
Yes, we meant to enable support for hasXyz() by default. As the need to
disable it seems very rare, we agreed it'd be fine to require a custom SPI
annotation to do so.

sjaakd:
> Yes, we meant to enable support for hasXyz() by default. As the need to disable it seems very rare, we agreed it'd be fine to require a custom SPI annotation to do so.

Ok. I'll add an additional PR (together with the documentation) to fix this. But: remember: this might actually break some code generation when users already have  `has` defined on their source DTO's. **Should we issue a warning?**

agudian:
I would hope that a note in the migration notes would be enough...

gunnarmorling:
+1. And mentioning it in the announcement post. It's a Beta, so if we se
complaints, we still an envision a more conservative default approach.

sjaakd:
aye sir :smile:

PR #823, opened 2016-07-03
Add @WithServiceImplementation annotation in testutil to register SPI implementations in processor tests

agudian:
I've also added the test case for the accessor naming strategy from the integration-tests module to the processor test to demonstrate the new feature. We should keep that one integration test, though.

Speaking of integration tests: our `FullFeatureCompilationTest` needs to ignore any test mappers that rely on special SPI implementations. Thus, let's put those tests in a package `.spi.` to allow the `FullFeatureCompilationTest` to ignore them.

@sjaakd: As you can see, this now works without having to modify something in the processor classes. If you revert everything in `org.mapstruct.ap.testutil.*` and `Services` and the related change in `Executables` on your branch, it should be rebasable to this. Well, plus the changed name of the annotations... :wink:

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/testutil/WithServiceImplementation.java
+++ b/processor/src/test/java/org/mapstruct/ap/testutil/WithServiceImplementation.java
@@ -1,0 +1,44 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.testutil;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Specifies an implementation of an SPI to be used during the annotation processing.
+ *
+ * @author Andreas Gudian
+ */
+@Retention(RetentionPolicy.RUNTIME)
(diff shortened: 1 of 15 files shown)


sjaakd:
@agudian gr8 work. 👍

processor/src/test/java/org/mapstruct/ap/testutil/WithServiceImplementation.java
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * @return The service implementation class that is to be made available during the annotation processing.
+     */
+    Class<?> value();
+
+    /**
+     * @return The SPI that the service implementation provides. If omitted, then {@link #value()} is expected to
+     *         implement exactly one interface which will be considered the provided SPI implementation.
+     */
+    Class<?> provides() default Object.class;
```
sjaakd: now.. this is what I like.. :smile: . smartly designed to take scan for an interface on the value..

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
```diff
@@ -110,32 +118,48 @@ protected void setupDirectories() throws Exception {
...
                 "javax.inject",
                 "spring-beans",
                 "spring-context",
                 "joda-time" };
 
+        return filterBootClassPath( whitelist );
+    }
+
+    private static List<String> buildProcessorClasspath() {
+        String[] whitelist =
```
gunnarmorling: Just coming to my mind, seeing this again: We may take a look at the Shrinkwrap Maven resolver API in order to build this classpath instead of copying stuff via the dependency plug-in. Nothing important, just for a dark winter night ;)

gunnarmorling:
@agudian Can you open an issue and reference it in the commit message?

gunnarmorling:
Looks good, although I feel it'd be much simpler to just create the services file under src/test/resources (exposed by a specific classloader maybe)? Feel free to merge after referencing an issue number from the commit message. Thanks!

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
agudian: We don't do the dependency-plugin copy stuff anymore... for quite a while now... :)
But yeah, there might be better ways to create new and specific classloader hierarchies... That would be something for that plan of yours to move the testutils to a standalone project.

gunnarmorling: Ooops, ok. That's how far behind I am ;)

agudian:
> Looks good, although I feel it'd be much simpler to just create the services file under src/test/resources (exposed by a specific classloader maybe)?

Hmm, don't know. Creating that file is actually the easiest part here and it takes just a couple of lines of code. The nasty stuff was all that classloader-fiddling.. :/

Issue #801, opened 2016-04-14
Use classloader of given mapper type in Mappers#getMappers()


gunnarmorling:
We should use these classloaders in the given order:

* The CL of the given class object (in 99% of cases that's the right one)
* As a fallback, the TCCL
* As a fallback, MapStruct's own CL

jvz:
Using the CL of the provided class would certainly make more sense than the TCCL. I just came across this bug in an OSGi environment where my mapper is in one bundle (and ClassLoader) but used by another bundle (via Camel's type converters). This causes a sort of ClassLoader race condition based on which bundle causes the initialization of the mapper.

gunnarmorling:
Fixed in master. Thanks for your feedback, @jvz!

jvz:
FWIW, what I ended up doing as a workaround for 1.0.0 was this:

```java
public class MapperFactory {

	private static final String IMPLEMENTATION_SUFFIX = "Impl";

	/**
	 * Returns an instance of the given mapper type.
	 *
	 * @param clazz The type of the mapper to return.
	 * @param <T>   The type of the mapper to create.
	 * @return An instance of the given mapper type.
	 */
	public static <T> T getMapper(final Class<T> clazz) {
		try {
			return clazz.getClassLoader().loadClass(clazz.getName() + IMPLEMENTATION_SUFFIX).asSubclass(clazz).newInstance();
		} catch (final ClassNotFoundException | InstantiationException | IllegalAccessException e) {
			throw new RuntimeException(e);
		}
	}

	private MapperFactory() {
	}
}
```

PR #826, opened 2016-07-05
#801 Trying to load mapper implementation types via the mapper interf…

gunnarmorling:
…ace's loader first

### DIFF
--- a/core-common/src/main/java/org/mapstruct/factory/Mappers.java
+++ b/core-common/src/main/java/org/mapstruct/factory/Mappers.java
@@ -18,9 +18,11 @@
  */
 package org.mapstruct.factory;
 
-import org.mapstruct.Mapper;
-
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


core-common/src/main/java/org/mapstruct/factory/Mappers.java
```diff
@@ -69,32 +71,57 @@ private Mappers() {
             // - clazz is annotated with @Mapper
             //
             // Use privileged action
```
agudian: should we maybe remove these comments here?

gunnarmorling: +1. Opened a new issue for these.

agudian:
Aside from the checkstyle issue, it looks good, also the order of the classloaders makes sense to me :+1: 
Go ahead and push it to master when you're done...

gunnarmorling:
Rebased and applied. Thanks for reviewing, @agudian!

PR #830, opened 2016-07-06
669 docupdate, fixing name and default strategy with presence checker iso specialized strategy

sjaakd:
As agreed yesterday


core-common/src/main/java/org/mapstruct/NullValueCheckStrategy.java
```diff
@@ -50,12 +50,34 @@
...
+     * <p>
+     * <ol>
+     *   <li>a source value is directly assigned to a target</li>
+     *   <li>a source value assigned to a target by calling a type conversion on the target first</li>
+     * </ol>
+     * <p>
+     * <b>NOTE:</b> mapping methods (generated or hand written) are excluded from this null check. They are intended to
+     * handle a null source value as 'valid' input.
+     *
+     *//**
```
agudian: There's something wrong with these comments...

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -1512,6 +1512,33 @@ However, by specifying `nullValueMappingStrategy = NullValueMappingStrategy.RETU
 
 The strategy works in a hierarchical fashion. Setting `nullValueMappingStrategy` on mapping method level will override `@Mapper#nullValueMappingStrategy`, and `@Mapper#nullValueMappingStrategy` will override `@MappingConfig#nullValueMappingStrategy`.
 
+
+[[checking-source-property-for-null-arguments]]
+=== Controlling checking result for 'null' properties in bean mapping
+
+MapStruct offers control over when to generate a `null` check. By default (`nullValueCheckStrategy = NullValueMappingStrategy.ON_IMPLICIT_CONVERSION`) a `null` check will be generated for:
+
+* direct setting of source value to target value.
```
agudian: I thought "direct setting of source value to target value" does not produce a null-check unless the target is a primitive type and the source is not...?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -1512,6 +1512,33 @@ However, by specifying `nullValueMappingStrategy = NullValueMappingStrategy.RETU
...
+
+First calling a mapping method on the source property is not protected by a null check. Therefor generated mapping methods will do a null check prior to carrying out mapping on a source property. Handwritten mapping methods must take care of null value checking. They have the possibility to add 'meaning' to `null`. For instance: mapping `null` to a default value.
+
+The option `nullValueCheckStrategy = NullValueMappingStrategy.ALWAYS` will always include a null check, unless a source presence checker is defined on the source bean.
+
+The strategy works in a hierarchical fashion. `@Mapper#nullValueMappingStrategy` will override `@MappingConfig#nullValueMappingStrategy`.
+
+[[source-presence-check]]
+=== Source presence checking
+Some frameworks generate bean properties that have a source presence checker. Often this is in the form of a method `hasXYZ`, `XYZ` being a property on the source bean in a bean mapping method. MapStruct will call this `hasXYZ` instead of performing a `null` check when it finds such `hasXYZ` method.
```
agudian: Just wondering while I'm reading this: do we also call `hasXYZ` if the `XYZ` property is primitive? That would be pretty cool, as it would cover an important part for the support of Google Protobuf beans.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -1512,6 +1512,33 @@ However, by specifying `nullValueMappingStrategy = NullValueMappingStrategy.RETU
...
+
+[TIP]
+====
+The source presence checker name can be changed in the MapStruct service provider interface (SPI). It can also be deactivated in this way.
```
agudian: We might want to point to an example or add a short description here on how that would work.

agudian:
Looks good, just had some minor comments. The example on the SPI can surely wait for later... or we do it in a blog post or something. Feel free to merge once you looked over it again. :+1:

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: You mean like in this generated code :smile: :

```java
   @Override
    Target sourceToTarget(Source source) {
        if ( source == null ) {
            return null;
        }

        Target target = new Target();

        if ( source.hasSomePrimitiveDouble() ) {
            target.setSomePrimitiveDouble( source.getSomePrimitiveDouble() );
        }
        if ( source.hasSomeInteger() ) {
            target.setSomeInteger( source.getSomeInteger() );
        }
        if ( source.hasSomeList() ) {
            target.setSomeList( new ArrayList<String>( source.getSomeList() ) );
        }
        if ( source.hasSomeArray() ) {
            java.lang.String[] someArray = source.getSomeArray();
            target.setSomeArray( Arrays.copyOf( someArray, someArray.length ) );
        }

        return target;
    }
```

agudian: Great! :+1:

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: you're right.. I just ran a test to be sure..

```java
public class Source {

    private int s1;
    private Integer s2;
    private int s3;
    private Integer s4;
}
```

```java
public class Target {

    private Integer s1;
    private int s2;
    private int s3;
    private Integer s4;
}
```
generated
```java

    @Override
    public Target map(Source in) {
        if ( in == null ) {
            return null;
        }

        Target target = new Target();

        target.setS1( in.getS1() );
        if ( in.getS2() != null ) {
            target.setS2( in.getS2() );
        }
        target.setS3( in.getS3() );
        target.setS4( in.getS4() );

        return target;
    }
}
```

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: Perhaps better to put some doc on SPI in on general.. Shall we take that in a separate update?

core-common/src/main/java/org/mapstruct/NullValueCheckStrategy.java
sjaakd: There's something wrong with NetBeans refactoring.. Anyway.. fixed that.

Issue #828, opened 2016-07-06
Mappers cannot auto-generated within eclipse NEON


shane-chen:
Hi all,

we are using mapstruct 1.0.0.Final in our project. After upgrading the eclipse IDE to NEON and importing eclipse projects, some mappers cannot be auto-generated. Following is the error and a screenshot in the IDE:

```
Multiple markers at this line
	- Internal error in the mapping processor: java.lang.NoClassDefFoundError: org/eclipse/jdt/internal/compiler/apt/model/ElementImpl at 
	 org.mapstruct.ap.internal.util.workarounds.EclipseAsMemberOfWorkaround.asMemberOf(EclipseAsMemberOfWorkaround.java:60) at 
	 org.mapstruct.ap.internal.util.workarounds.SpecificCompilerWorkarounds.asMemberOf(SpecificCompilerWorkarounds.java:137) at 
	 org.mapstruct.ap.internal.util.workarounds.TypesDecorator.asMemberOf(TypesDecorator.java:148) at 
	 org.mapstruct.ap.internal.model.common.TypeFactory.getMethodType(TypeFactory.java:261) at 
	 org.mapstruct.ap.internal.model.common.TypeFactory.getParameters(TypeFactory.java:277) at 
	 org.mapstruct.ap.internal.model.common.TypeFactory.getSingleParameter(TypeFactory.java:273) at org.mapstruct.ap.internal.model.PropertyMapping
	 $PropertyMappingBuilder.getTargetType(PropertyMapping.java:502) at org.mapstruct.ap.internal.model.PropertyMapping
	 $PropertyMappingBuilder.build(PropertyMapping.java:166) at org.mapstruct.ap.internal.model.BeanMappingMethod
	 $Builder.applyPropertyNameBasedMapping(BeanMappingMethod.java:433) at org.mapstruct.ap.internal.model.BeanMappingMethod
	 $Builder.build(BeanMappingMethod.java:128) at 
	 org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMappingMethods(MapperCreationProcessor.java:361) at 
	 org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMapper(MapperCreationProcessor.java:143) at 
	 org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:114) at 
	 org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:73) at 
[...]

PR #831, opened 2016-07-06
#828 Fix NoClassDefFoundError when running in Eclipse Neon (4.6)

agudian:
I wasn't able to add a test for it, as there are no bundles of the new compiler version on Maven central, yet.
But I did test it locally with a fresh Eclipse Neon package and in Eclipse Mars (4.5.2).

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
@@ -36,7 +36,8 @@ import org.mapstruct.ap.internal.version.VersionInformation;
  * @author Andreas Gudian
  */
 public class SpecificCompilerWorkarounds {
-    private SpecificCompilerWorkarounds() { }
+    private SpecificCompilerWorkarounds() {
+    }
 
     /**
      * Tests whether one type is assignable to another, checking for VOID first.
@@ -120,7 +121,7 @@ public class SpecificCompilerWorkarounds {
     /**
      * Workaround for Bugs in the Eclipse implementation of {@link Types#asMemberOf(DeclaredType, Element)}.
      *
-     * @see <a href="https://bugs.eclipse.org/bugs/show_bug.cgi?id=382590">Eclipse Bug 382590</a>
+     * @see <a href="https://bugs.eclipse.org/bugs/show_bug.cgi?id=382590">Eclipse Bug 382590 (fixed in Eclipse 4.6)</a>
      * @see <a href="https://bugs.eclipse.org/bugs/show_bug.cgi?id=481555">Eclipse Bug 481555</a>
      */
     static TypeMirror asMemberOf(Types typeUtils, ProcessingEnvironment env, VersionInformation versionInformation,
@@ -134,7 +135,7 @@ public class SpecificCompilerWorkarounds {
             catch ( IllegalArgumentException e ) {
                 lastException = e;
                 if ( versionInformation.isEclipseJDTCompiler() ) {
-                    result = EclipseAsMemberOfWorkaround.asMemberOf( env, containing, element );
+                    result = EclipseClassLoaderBridge.invokeAsMemberOfWorkaround( env, containing, element );
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
```diff
@@ -149,4 +156,45 @@ static TypeMirror asMemberOf(Types typeUtils, ProcessingEnvironment env, Version
...
+                    ( (URLClassLoader) ClassLoaderHelper.class.getClassLoader() ).getURLs(), parent );
+            }
+
+            return classLoader;
+        }
+
+        private static Method getAsMemberOf(ClassLoader platformClassLoader) throws Exception {
+            if ( asMemberOf == null ) {
+                Class<?> workaroundClass = getOrCreateClassLoader( platformClassLoader )
+                    .loadClass( ECLIPSE_AS_MEMBER_OF_WORKAROUND );
```
sjaakd: I'm a bit lost here.. Why do we need reflection to get access to our own util class?

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
```diff
@@ -134,7 +140,8 @@ static TypeMirror asMemberOf(Types typeUtils, ProcessingEnvironment env, Version
             catch ( IllegalArgumentException e ) {
                 lastException = e;
                 if ( versionInformation.isEclipseJDTCompiler() ) {
-                    result = EclipseAsMemberOfWorkaround.asMemberOf( env, containing, element );
+                    result = (TypeMirror) ClassLoaderHelper.getAsMemberOf( element.getClass().getClassLoader() )
```
sjaakd: should we check for the Eclipse JDT compiler version and only then perform this trick?

gunnarmorling:
@shane-chen, any chance that you build this PR and confirm whether it addresses your issue? Just run `mvn clean install` and use the JAR from `processor/target` in Eclipse. Thanks!

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
agudian: Because our class itself and the classes it uses (ElementImpl and so on) need to be loaded from the same classloader hierarchy. As such a classloader doesn't exist, I need to create a new one - with the compiler's classloader as parent as I need to interact with those types directly.
With our workaround-class being loaded in that nwe classloader that has no relation to the CL of the rest of MapStruct, I can't cast that newly loaded class here to one of our types. Hence the need to do it via reflection.
The alternative would be to try and do everything that's being done in our workaround class on the eclipse specific types solely via reflection. But that's no fun either ;-).

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
agudian: I thought about it, but there are some differences in version counting when comparing the usage within the IDE and when using it via Maven or so. Plus I don't know what version exactly introduced that change. So I felt more safe to do it in a way that works for "all known" versions for now.

agudian:
@shane-chen, yeah, it would be great if you could give it a test drive. If you find anything that prevents you from updating from 1.0.0 to the upcoming 1.1.0-Beta2, let us know so we can either fix that or port this fix back and do a 1.0.2 release.

shane-chen:
@gunnarmorling @agudian thank you for your prompt reply. I've just tested the processor for our project in eclipse NEON. It works fine. When the 1.1.0-Beta2 is scheduled to be released?

gunnarmorling:
Awesome, thanks for the quick test and reporting back! The release should
be out in about two weeks from now.

processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/SpecificCompilerWorkarounds.java
sjaakd: Ok.. But.. you could also choose to wrap this part in a new `EclipseSpecificCompilerWorkaround` and delegate to `EclipseAsMemberOfWorkaround`. Advantage: the class loader stuff is Eclipse specific, so it would make sense to isolate that.

agudian:
@shane-chen Thanks for testing, Shengyao!

@sjaakd: I've moved the classloading and reflection to a new class as you suggested and it fells better that way. Wanna take another look?

Issue #836, opened 2016-07-19
Parameter mapping not reversible.


sjaakd:
Currently I get the error notice:
`Parameter nl.bro.entities.gm.gmw.GmwDeliveredLocationEntity entity cannot be reversed.`

when using: `@InheritInverseConfiguration`. 

This is akward. It means that all other mappings which can be automatically inherited must be defined by hand.

Better / more logical: ignore silently and exclude from inherit mapping.

sjaakd:
Actually the problem has been resolved by introducing `NestedTargetMapping`. However, by removing the checks for this issue, I stumbled on a new issue: #839. 

A reproducer (check generated code) is already present in PR #837. Checkout generated code for: 
`org.mapstruct.ap.test.nestedsource.parameter.LetterMapper`

PR #837, opened 2016-07-20
#836, #838 left overs reverse nested mapping

sjaakd:
See #836. Makes flatting (normalizing e.g. normalizing into an entity) and exploding (denormalizing into an dto) a bit more friendly by allowing usage of `@InheritInverseConfiguration`


sjaakd:
Hmm.. wrong test case. Need to provide a better example. Perhaps we do need another issue to tackle non existing empty constructors. Anyway..

sjaakd:
@agudian .. Do you see a chance to review this one still? I don't think its a dramatic one :smile:

agudian:
Looks good, go ahead with it! :+1:

sjaakd:
Thanks for reviewing 👍

PR #845, opened 2016-07-28
Enable support for Iterable classes which are not generic

osthus-sm:
(e.g. public class StringList extends List<String>)

In one scenario I have an auto generated class, which just extends List<String> or (List<Object> ...) and we want to map it. Currently I get the following error:
"error: Internal error in the mapping processor: java.util.NoSuchElementException          at java.util.ArrayList$Itr.next(ArrayList.java:854)     at org.mapstruct.ap.internal.util.Collections.first(Collections.java:77)      at org.mapstruct.ap.internal.model.IterableMappingMethod$Builder.build(IterableMappingMethod.java:102)"

Attached you find a possible solution with an additional testcase.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
@@ -91,10 +91,12 @@ public class IterableMappingMethod extends MappingMethod {
             Type sourceParameterType = first( method.getSourceParameters() ).getType();
             Type resultType = method.getResultType();
 
-            Type sourceElementType  = sourceParameterType.isArrayType() ? sourceParameterType.getComponentType()
-                : first( sourceParameterType.getTypeParameters() ).getTypeBound();
-            Type targetElementType = resultType.isArrayType() ? resultType.getComponentType()
-                : first( resultType.getTypeParameters() ).getTypeBound();
+            Type sourceElementType =
+                sourceParameterType.isArrayType() ? sourceParameterType.getComponentType() : first(
+                    sourceParameterType.determineTypeArguments( Iterable.class ) ).getTypeBound();
+            Type targetElementType =
+                resultType.isArrayType() ? resultType.getComponentType() : first(
+                    resultType.determineTypeArguments( Iterable.class ) ).getTypeBound();
 
             String loopVariableName =
                 Strings.getSaveVariableName( sourceElementType.getName(), method.getParameterNames() );
@@ -246,7 +248,7 @@ public class IterableMappingMethod extends MappingMethod {
             return sourceParameterType.getComponentType();
         }
         else {
-            return sourceParameterType.getTypeParameters().get( 0 ).getTypeBound();
+            return sourceParameterType.determineTypeArguments( Iterable.class ).get( 0 ).getTypeBound();
         }
     }
 
@@ -255,7 +257,7 @@ public class IterableMappingMethod extends MappingMethod {
             return getResultType().getComponentType();
         }
         else {
-            return getResultType().getTypeParameters().get( 0 );
+            return getResultType().determineTypeArguments( Iterable.class ).get( 0 );
         }
     }
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -793,7 +793,7 @@ public class Type extends ModelElement implements Comparable<Type> {
 
     /**
      * Searches for the given superclass and collects all type arguments for the given class
-     * 
+     *
      * @param superclass the superclass or interface the generic type arguments are searched for
      * @return a list of type arguments or null, if superclass was not found
      */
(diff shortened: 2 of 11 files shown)


agudian:
Just gave it a quick look, but it's very promising, thanks!

WDYT, could you also fix that for Maps? For cases like `StringToLongMap` or `StringKeyMap<V>` or `LongValueMap<K>` and that kind of things... That would probably call for a new test class then to keep it a bit separated.

agudian:
ok to test (to start the pull request builder)

osthus-sm:
Added support for the non generic maps, too and added another test class, is it okay this way?

agudian:
@osthus-sm: Thanks for your efforts! Sorry I didn't get to review your latest stuff, yet. I'm on vacation for the next week, the other devs here are traveling as well at the moment. I'll take a closer look the week after this.

processor/src/test/java/org/mapstruct/ap/test/collection/TestNonGenericList.java
```diff
@@ -0,0 +1,29 @@
...
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.collection;
+
+import java.util.ArrayList;
+
+/**
+ * @author Stefan May
+ */
+public class TestNonGenericList extends ArrayList<String> {
```
agudian: Could you rename this to `NonGenericList`? I think the prefix `Test` is not necessary here.

processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
```diff
@@ -246,7 +248,7 @@ public Type getSourceElementType() {
             return sourceParameterType.getComponentType();
         }
         else {
-            return sourceParameterType.getTypeParameters().get( 0 ).getTypeBound();
+            return sourceParameterType.determineTypeArguments( Iterable.class ).get( 0 ).getTypeBound();
```
agudian: Please use `first( x )` instead of `x.get( 0 )`, just as in the diff lines above. Same for line 260 below.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -790,4 +790,30 @@ public boolean hasEmptyAccessibleContructor() {
...
+    /**
+     * Searches for the given superclass and collects all type arguments for the given class
+     *
+     * @param superclass the superclass or interface the generic type arguments are searched for
+     * @return a list of type arguments or null, if superclass was not found
+     */
+    public List<Type> determineTypeArguments(Class<?> superclass) {
+        TypeMirror superclassMirror =
+            typeUtils.erasure( elementUtils.getTypeElement( superclass.getCanonicalName() ).asType() );
+        if ( typeUtils.isAssignable( superclassMirror, typeMirror )
```
agudian: I guess you use `isAssignable` in both directions because you are working with the erasure of `superclass`? I would guess that you could just String-compare the qualified type name instead to save us some CPU cycles (qualifiedName.equals( superclass.getName() )`. That would also eliminate the need to re-create the `TypeElement` / `TypeMirror` instances of `Map` / `Iterable` for each invocation.

processor/src/test/java/org/mapstruct/ap/test/collection/CollectionMappingTest.java
```diff
@@ -389,4 +391,56 @@ public void shouldEnumSetAsCopy() {
         assertThat( source.getEnumSet() ).containsOnly( Colour.BLUE, Colour.GREEN, Colour.RED );
         assertThat( target.getEnumSet() ).containsOnly( Colour.BLUE, Colour.GREEN );
     }
+
+    @Test
+    @IssueKey("TODO")
```
agudian: `@IssueKey("853")`, please... :smile:

processor/src/test/java/org/mapstruct/ap/test/collection/CollectionMappingTest.java
```diff
@@ -389,4 +391,56 @@ public void shouldEnumSetAsCopy() {
...
+        newTarget.setNonGenericStringList( nonGenericStringList );
+
+        Source mappedSource = SourceTargetMapper.INSTANCE.targetToSource( newTarget );
+
+        assertThat( mappedSource ).isNotNull();
+        assertThat( mappedSource.getStringList3() ).containsExactly( "Bill", "Bob" );
+    }
+
+    @Test
+    @IssueKey("TODO")
```
agudian: `@IssueKey("853")`

agudian:
Review done, and I only have a few remarks.

Could you please rebase your branch to the current master, squash your commits into one and prefix the commit message with "#853"?

gunnarmorling:
Hey @osthus-sm, any chance you could address the review remarks brought up by @agudian? Thanks a lot for your help, seems we are almost there :)

osthus-sm:
Hey, I'm currently busy, but will implement the changes in the week starting 4th of September. I should have more time than.

agudian:
> Hey, I'm currently busy, but will implement the changes in the week starting 4th of September. I should have more time than.

@osthus-sm, do you think you can find some time in the next days? We'd like to get out 1.1.0.Final next (at least that's our plan) and it would be great to include this. If you don't find the time on short notice, let me know and I'll just add a commit to adjust the small things myself.

agudian:
@osthus-sm I've rebased and squashed your commits and added the issue key as message prefix in the commit. I've added an additional commit to adapt some minor things that I pointed out in the review and added you to our list of contributors in the copyright.txt.

Again many thanks for your contribution, it's always welcome!

Issue #846, opened 2016-08-04
@InheritInverseConfiguration not working with @MappingTarget on interface


afdia:
This issue was introduced in 1.1.0.Beta2. Older versions (including 1.1.0.Beta1) work.

Example code:
```
public class Test {
	static interface BInterface {
		String getName();
		void setName(String s);
	}
	static class B implements BInterface {
		String name;
		public B() {}
		public String getName() {return name;}
		public void setName(String name) {this.name = name;}
	}
	static class A {
		String aName;
		public A() {}
		public String getaName() {return aName;}
		public void setaName(String name) {this.aName = name;}
	}
	@Mapper
	static interface MyMapper {
		@Mapping(source = "name", target = "aName")
		A convert(BInterface b);

		@InheritInverseConfiguration
		void convert(@MappingTarget BInterface target, A source);
	}
}
```

In Beta2 this shows the error `There is no suitable result type constructor for reversing this method.` at the `@InheritInverseConfiguration` annotation

Previous versions worked as expected and MapStruct doesn't need a constructor for this case, because the target object is created externally and passed to the method.

Changing the last method to
```
		@Mapping(source = "aName", target = "name")
		void convert(@MappingTarget BInterface target, A source);
```
works, therefore the issue seems to be linked to the `@InheritInverseConfiguration` annotation

sjaakd:
Oops. It's actually linked to one of the last two pr's. I included a check but obviously forgot the @mappintarget case. Put it on my name. It's not too difficult to fix.

Verstuurd vanaf mijn iPhone

agudian:
Just merged the fix by @sjaakd to master. @afdia, thanks for reporting!

PR #851, opened 2016-08-13
#846 Exclude updates methods from suitable constructor check

sjaakd:


sjaakd:
Apparently test classes should not start with TestABC ?

agudian:
:+1:

PR #854, opened 2016-08-17
Fixed ordering of beanProperties

tisoft:
The sorting of properties returned sometimes the wrong order, when there are several properties, that have no descendants and are considered equal in the comparator. I changed the GraphAnalyzer to create a node path, that contains a valid order and use that to sort the properties.

I wasn't able to extract a standalone test case that displays the old wrong behavior from our code base. But without this change we get wrongly sorted dependencies in our application.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -22,7 +22,6 @@ import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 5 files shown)


agudian:
Thanks for reaching out with a PR! Really appreciated!

But honestly I'd feel much more comfortable merging this if I understood what problem is that you're fixing.
Because with your changes the order is just based on the order in which the nodes are visited. And for nodes that are not descendents of each other (i.e. somehow in sibling subtrees), that order is semi-random, as it now depends on the order of the HashMap iteration instead of the compiler's order in the bean class (we WANT to use the compiler's order here for those cases).
My gut tells me that the reason that it's hard to create a reproducing test case is this swapping of a seemingly random order (compiler's order) with the semi-random order of the HashMap iteration.

So I hope you can understand us if we'd like to see a reproducing test case or a more elaborate bug report before merging a fix.

tisoft:
I have pushed a testcase that shows the wrong behaviour and also changed the implementatuion to keep the compile order as much as possible

agudian:
Great, thanks! I'll get to it tonight or so.

agudian:
@tisoft, I went with a different but hopefully easier solution and opened PR #857 for that. I also created a corresponding issue and another related one (#855 and #856). I kept your test-case, though! Just with a different class name and the test expectation is slightly different but within the bounds of our specification. Would you mind checking that out?

But huge thanks for finding and reporting this issue!

gunnarmorling:
Going to close this one as it's superseded by @agudian's work. Thanks for filing the report and adding the test case, @tisoft. It's part of Andreas' PR, too.

PR #857, opened 2016-08-21
#855 Fix property order when using dependsOn

### LINKED ISSUE #855: DependsOn creates the correct property order only in some cases
As demonstrated in #854 by @tisoft, `dependsOn` sometimes doesn't have any effect, even if no constant-mappings are involved (see #619 for that).

agudian:
This fixes #855 but uses a different approach than propsed in #854. This solution is hopefully a bit easier to understand and allows us to remove some now unecessary code from the implementation (more deletions than additions in the implementation classes... :smile:).

If this PR is merged, then #854 can be closed as well.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -18,6 +18,10 @@
  */
 package org.mapstruct.ap.internal.model;
 
+import static org.mapstruct.ap.internal.util.Collections.first;
+import static org.mapstruct.ap.internal.util.Collections.last;
+import static org.mapstruct.ap.internal.util.Strings.getSaveVariableName;
+
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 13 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -218,18 +219,10 @@ private void sortPropertyMappingsByDependencies() {
...
                         public int compare(PropertyMapping o1, PropertyMapping o2) {
-                            if ( graphAnalyzer.getAllDescendants( o1.getName() ).contains( o2.getName() ) ) {
-                                return 1;
-                            }
-                            else if ( graphAnalyzer.getAllDescendants( o2.getName() ).contains( o1.getName() ) ) {
-                                return -1;
-                            }
-                            else {
-                                return 0;
-                            }
```
agudian: This old implementation of the `Comparator` violates the API spec in this paragraph, which led to the bug:

> Finally, the implementor must ensure that `compare(x, y)==0` implies that `sgn(compare(x, z))==sgn(compare(y, z))` for all `z`.

tisoft:
@agudian This is looking good for me. I checked it out and used it to generate the mapper that had problems in our application and it's coming out correct.

gunnarmorling:
@agudian Could you split the commit so that @tisoft's test entities are committed under his name/email? Also add him to copyright.txt, please.

gunnarmorling:
@agudian The change itself looks good, thanks! The only thing would be to split up the commit as described above.

agudian:
Sure, done. Thanks again, @tisoft!

Issue #860, opened 2016-08-22
Use assertj instead of fest


filiphr:
While working on my last pull request I noticed that fest asserts are used for testing the framework. 

Would you be interested in replacing the fest asserts with [assertj](https://github.com/joel-costigliola/assertj)? Assertj is actually a fork of fest and is actively developed as opposed to fest. There  is also a [migration guide](http://joel-costigliola.github.io/assertj/assertj-core-migrating-from-fest.html) about migration from fest assert.

gunnarmorling:
So far I didn't perceive any shortcoming due to the lack of development in FEST Assert.

Agreed though that using an actively maintained lib would be nice. Would you like to give it a go? That'd be great, I cannot judge though how much work it'd be (we are still on FEST 1.4).

Thanks for the idea and any help!

filiphr:
I will try to do the migration assertj. I had a quick look and it seems that there aren't a lot of special usages of FEST Assert so it shouldn't be too complicated.

@gunnarmorling I have one question, which version should I use:
* AssertJ 1.x requires Java 6 or higher
* AssertJ 2.x requires Java 7 or higher (use the org.assertj.core.api.Java6Assertions entry point on Android)
* AssertJ 3.x requires Java 8 or higher (use the org.assertj.core.api.Java6Assertions entry point on Android)

2.x and 3.x have the same features (3.x having java 8 specifics).

I am asking because, I've seen java 8 in the tests, but in the pom I think the maven-compiler is configured for java 6.

filiphr:
@gunnarmorling It was much easier than I thought so I provided a pull request for the switch. If you are interested you can review it.

gunnarmorling:
Awesome, thanks a lot! Regarding the version, we use Java 8 for building but work with source level 1.6 for source and target (apart from the mapstruct-jdk8 module). Not sure where this leaves us exactly in terms of the AssertJ version, but it seems your PR works successfully using AssertJ 3.x?

filiphr:
Yes the PR is working with AssertJ 3.x. If you are OK I'll leave this version, as in any case there are java 8 classes in the tests

agudian:
Your PR is merged. Thanks a lot, @filiphr!

filiphr:
Great. We can close the issue now.

PR #863, opened 2016-08-25
#860 switch to using assertj instead of FEST assertions

filiphr:
This is a pull request for (#860) switching the assertions framework to the maintained assertj instead of FEST.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/testutil/assertions/JavaFileAssert.java
+++ b/processor/src/test/java/org/mapstruct/ap/testutil/assertions/JavaFileAssert.java
@@ -21,9 +21,9 @@ package org.mapstruct.ap.testutil.assertions;
 import java.io.File;
 import java.io.IOException;
 
-import org.fest.assertions.Assertions;
-import org.fest.assertions.FileAssert;
-import org.fest.assertions.StringAssert;
+import org.assertj.core.api.AbstractCharSequenceAssert;
+import org.assertj.core.api.Assertions;
+import org.assertj.core.api.FileAssert;
 
 import com.google.common.base.Charsets;
 import com.google.common.io.Files;
@@ -44,7 +44,7 @@ public class JavaFileAssert extends FileAssert {
     /**
      * @return assertion on the file content
      */
-    public StringAssert content() {
+    public AbstractCharSequenceAssert<?, String> content() {
         exists();
         isFile();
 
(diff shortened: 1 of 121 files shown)


processor/src/test/java/org/mapstruct/ap/testutil/assertions/JavaFileAssert.java
```diff
@@ -44,15 +44,16 @@ public JavaFileAssert(File actual) {
...
+    public AbstractCharSequenceAssert<?, String> content() {
         exists();
         isFile();
 
         try {
             return Assertions.assertThat( Files.toString( actual, Charsets.UTF_8 ) );
         }
         catch ( IOException e ) {
-            failIfCustomMessageIsSet( e );
+            //TODO shouldn't we just fail if something like this happens.
```
filiphr: @gunnarmorling Have a better look at this TODO. `failIfCustomMessageIsSet` does not exists in assertj anymore and I was not sure whether to directly fail with some message or do something else. Previously no custom message was set so even if an exception occurred the test did not fail.

agudian: Good catch! Let's just use `fail("Unable to read " + actual, e)`.

agudian:
Awesome, @filiphr, thanks for the PR! Would you replace the TODO with the `fail` call? You can directly amand your commit or squash the change to force-push it to your branch.

gunnarmorling:
Nice one, thanks a lot!

filiphr:
I am glad that you like the PR. I'll fix the TODO and do a force-push, tomorrow evening probably

filiphr:
@agudian I replaced the TODO with a fail message

gunnarmorling:
@filiphr I'm getting some failures in the integration tests, supposedly due to the requirement of JDK 8 by that new version (we run some integration tests on previous compiler versions). Didn't look into the details yet, but it seems we may actually have to use that version that run with JDK 7. Did you get the same when running `mvn clean install` on the entire project?

gunnarmorling:
Jenkins, test this please.

filiphr:
@gunnarmorling I did mvn clean install on the entire project. I did it again and realized that it didn't run with oracle_java_6 and oracle_java_7 that is why I didn't see the errors. A solution would be to use the 1.x branch which is requires java 6 or above. 2.x is for java 7 or higher.

Is it an option to use 1.x only in the integration tests?

filiphr:
My last commit should fix the failing tests

agudian:
ok to test

agudian:
I think it's a good idea to use the Java 6 compatible version only in the integration tests. Focus there isn't on fancy assertions anyway, but on correct compilation using the different compilers.

agudian:
Alright, the pull-request-builder looks healthy now as well. I'd say you squash the second commit again into the first one and I'll push it to master then... :+1:

agudian:
Never mind, forgot we have squash-and-merge for quite some time now here... :neckbeard:

Issue #843, opened 2016-07-27
Non effective code generated for nested property conversion


pavelbely:
Hi!

MapStruct generates code to call method twice instead of saving it result to a variable and then working with it.

Consider the interface
```java
@Mapper
public interface TagMapper {

	@Mapping(target = "date", source = "gitlabTag.commit.authoredDate")
	TagInfo gitlabTagToTagInfo(GitlabTag gitlabTag);
}
```

And the following is generated. Please mind double usage of gitlabTagCommitAuthoredDate() method.
```java
public class TagMapperImpl implements TagMapper {

    @Override
    public TagInfo gitlabTagToTagInfo(GitlabTag gitlabTag) {
        if ( gitlabTag == null ) {
            return null;
        }

        TagInfo tagInfo = new TagInfo();

        if ( gitlabTagCommitAuthoredDate( gitlabTag ) != null ) {
            tagInfo.setDate( new DateTime( gitlabTagCommitAuthoredDate( gitlabTag ) ) );
        }
        ...

        return tagInfo;
    }

    private Date gitlabTagCommitAuthoredDate(GitlabTag gitlabTag) {

        if ( gitlabTag == null ) {
            return null;
        }
        GitlabBranchCommit commit = gitlabTag.getCommit();
        if ( commit == null ) {
            return null;
        }
        Date authoredDate = commit.getAuthoredDate();
        if ( authoredDate == null ) {
            return null;
        }
        return authoredDate;
    }
}
```

I am using MapStruct 1.0.0.Final with Java 8.

agudian:
Still present in 1.1.0.Beta2. We should fix that.
@pavelbely, thanks for letting us know! Would you be interested in providing a fix?

pavelbely:
You are welcome, Andreas.
Sure, I will try to fix it.

sjaakd:
There should be such issue for the reverse nesting case. I noticed it just before merging but ran out of time. Perhaps we are talking about the same one.

Verstuurd vanaf mijn iPhone

sjaakd:
Could this be caused by an old version as well? I vaguely remember that such issue is fixed in the forward mapping (even before 1.0.0)...

Verstuurd vanaf mijn iPhone

sjaakd:
Hi, 

I tried to reproduce this on 1.1.0.Beta2.. This is the result:

```java
public class TagMapperImpl implements TagMapper {

    @Override
[...]

PR #869, opened 2016-08-27
#843 reproducer and solution

sjaakd:
I can't really make a reproducer for this one. You can't test for ineffective code. So I'm wandering whether the UT makes sense.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_843/Issue843Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_843/Issue843Test.java
@@ -1,0 +1,58 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_843/Issue843Test.java
```diff
@@ -0,0 +1,47 @@
...
+    GitlabTag.class,
+    TagMapper.class
+})
+@IssueKey("843")
+public class Issue843Test {
+
+    @Test
+    public void testMapperCreation() {
+        TagMapper.INSTANCE.gitlabTagToTagInfo( null );
+        // check the implemenation for ineffective code (duplication of forged method call).
```
agudian: You could use the `GeneratedSource` rule to obtain a `JavaFileAssert` with a `.content()` assert on the source code where you can use `containsOnlyOnce` to check that the invocation of the forged method is there only once. Or extend `JavaFileAssert` to do something else based on the generated file if you need it.

agudian: Another possibility is to add a counter in the getters and verify that they are not called more than expected.

agudian:
Always interesting to see how simple your fixes look like... :+1:.

I only have a suggestion on how to test it. And reword the commit message as well, please... :smile:

processor/src/test/java/org/mapstruct/ap/test/bugs/_843/Issue843Test.java
sjaakd: Of course.. I'll go for the latter one.

Issue #707, opened 2015-12-11
Private iterable methods are not generated if element mapping method(s) are qualified


gunnarmorling:
From [Stack Overflow](http://stackoverflow.com/questions/34200421/mapstruct-ambiguous-exception-when-using-helper-class):

When declaring two methods with same source/target types but with different qualifiers, a mapping method in another class requiring that mapping fails with an "ambigious mapping method" error. Apparently the qualifiers are not applied correctly, as it works when declaring all methods in the same mapper.

gunnarmorling:
Also we seem to fail to generate required private iterable mapping methods in case there several (qualified) element mapping methods involved. We should generate one iterable method per qualifier, e.g. with different names.

agudian:
The first thing reported smells like a wrong retention policy...?

gunnarmorling:
@agudian Ah, yes, indeed it works when changing it to `CLASS`, thanks! Let's add a note to the docs that qualifiers should always have `CLASS` at least.

gunnarmorling:
Reworded the issue title to just address the second issue.

afdia:
As I stumbled over the same issue with the retention policy, I would suggest to change the qualifier examples (TitleTranslator, EnglishToGerman, GermanToEnglish) in the documentation [1] from RetentionPolicy.SOURCE to RetentionPolicy.CLASS.

Perhaps you could also mention that if a qualifier annotation is located in a different file, it must have at least RetentionPolicy.CLASS to make it work. Perhaps you meant that with 
> Let's add a note to the docs that qualifiers

but I'm not sure if you were talking about JavaDoc or the documentation on the website.

[1] http://mapstruct.org/documentation/1.0/reference/html/index.html#selection-based-on-qualifiers

sjaakd:
@gunnarmorling :

So the documentation should be clarified. No problem. I think this unclarity (I made the same mistake) can cause a lot of misunderstanding.

> Also we seem to fail to generate required private iterable mapping methods in case there several (qualified) element mapping methods involved. We should generate one iterable method per qualifier, e.g. with different names.

Checkout the warning on P21 of our documentation: 'A class / method annotated with a qualifier will not qualify anymore for mappings that do not have the qualifiedBy element.'

So IIUC and in the light of the documentation, it would be very strange to generate a private method in case of iterable / map mappings if you cannot make a match based on the `qualifiedBy`. Right?

gunnarmorling:
[...]

PR #870, opened 2016-08-28
#707 Document to always use RetentionPolicy.CLASS in custom qualifier annotations

sjaakd:


documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -867,6 +867,11 @@ public class Titles {
 
 [WARNING]
 ====
+When the qualifier is placed in a different library, make sure it has the retention policy Class (`@Retention(RetentionPolicy.CLASS)`). Otherwise the qualifier will not end up in the jar and MapStruct cannot 'see' it during code generation.
```
agudian: I'd actually just say here and in the javadoc above that `@Retention(RetentionPolicy.CLASS)` should always be used on custom qualifier annotations, just to be on the safe side (they might suddenly be reused in a different module, or someone on the team uses Eclipse and due to the incremental compilation it reads the .class file instead of the .java file from the same project.

agudian:
And I have one further request... It's quite tedious for me to map all those PRs to their issues, as I don't grasp on the first look what's behind the issue numbers. Therefore commit messages and PR titles with issue-number plus some actualy describing message of the change would help me a lot (e.g. "#707 Document to always use RetentionPolicy.CLASS in custom qualifier annotations").
I know your perception of issue numbers is different here, but for me it usually is just a random number that I always have to look up... :grin:

sjaakd:
> I know your perception of issue numbers is different here, but for me it usually is just a random number that I always have to look up... 😁

Actually, my perception is not that different. Its just sloppiness from my side. I'll adapt the commit message.

sjaakd:
@agudian ..let me know if you're done with this one..? Then I'll force push the change..

sjaakd:
Oh.. BTW.. Its interesting to check the issue itself (#707), last comment. I'm not sure wether I agree with @gunnarmorling.. So I only did the documentation update.

BTW: thanks for reviewing.

agudian:
Yes, I'm done with the review here. I'm currently doing the reviews kinda on/off whenever I have a few minutes to spare, so I sometimes forget where I need to actually finish something... (I'm at home with the kids Monday till Wednesday this week as an easy ending to the summer hodidays).

Please go ahead with the squash and push... :+1:

sjaakd:
please test this

sjaakd:
test this please

Issue #787, opened 2016-03-16
Update readme


gunnarmorling:
- [x] Should show using compiler-plugin
- [x] Update general description
- [x] Point to docs
- [x] Show short example

gunnarmorling:
Done.

PR #874, opened 2016-08-29
#787 Enhance readme to use maven-compiler-plugin and add some badges.

agudian:
Adjusts the thing of #787 that I find most annoying. :grin:


sjaakd:
Looks ok to me.. Please merge

Issue #865, opened 2016-08-26
Collection source null check


wenerme:
```java

// mapper for member field
        if ( e.getMembers() != null ) {

            e.getMembers().clear();
            // may add null to target collection, check s.getMembers() != null is what I expected
            e.getMembers().addAll( projMemberDTOListToProjCoreUserEntityList( s.getMembers() ) );
        }

        else {

            e.setMembers( projMemberDTOListToProjCoreUserEntityList( s.getMembers() ) );
        }

// mapper source collection
    protected List<ProjCoreUserEntity> projMemberDTOListToProjCoreUserEntityList(List<ProjMemberDTO> list) {

        if ( list == null ) {

            return null;
        }

        List<ProjCoreUserEntity> list_ = new ArrayList<ProjCoreUserEntity>();

        for ( ProjMemberDTO projMemberDTO : list ) {

            list_.add( map( projMemberDTO ) );
        }

        return list_;
    }
```

wenerme:
For every collection field I have to assign an empty collection.

gunnarmorling:
Hi @wenerme, it's not quite clear to me what the issue is you are reporting. Could you describe more clearly what's the behavior you'd expect from MapStruct and what's the behavior you currently get instead? Thanks!

agudian:
And maybe also check this section of the docs: http://mapstruct.org/documentation/1.1/reference/html/index.html#mapping-result-for-null-arguments

agudian:
Now I understand:
* You have an update-method (mapping method with `@MappingTarget` parameter).
* A collection property needs to be mapped and for the element types there is a mapping method - the iterable mapping method is generated automatically (we call that a _forged method_ internally).
* The generated iterable mapping method returns `null` by default if the source is `null`. But as the result is directly passed to `addAll(..)`, you get a `NullPointerException`.

The right thing to do would be to perform a null-check on the result of the property-mapping method (user-specified or not) and then only call `addAll(..)` if it's not null.

wenerme:
Sorry for weekend delay, @agudian yes, I don't expected `addAll(null)`, this will cause NPE (JPA Persist Collection).

sjaakd:
Fixed with #878. @agudian : thanks for reviewing.

PR #876, opened 2016-08-31
#865 Collection NPE check for a calling update method.

sjaakd:


processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -447,14 +448,25 @@ private Assignment assignToCollection(Type targetType,
...
+                        // if the calling method is an update method and accesses a getter, make a local variable to
+                        // test NPE first.
+                        result = new LocalVarWrapper( result, method.getThrownTypes(), targetType );
+                    }
+                    else {
+                        // if not, asssign as new collecitin or direct
+                        result = new SetterWrapper( result, method.getThrownTypes() );
+                    }
+
+                    // TODO.. Reviewer, why is there a special collectionGetterName?
```
agudian: Hmm, it turns the setter method to the name of the getter. That might actually make sense, but I don't think we're checking if it actually exists. Hmm.

processor/src/test/java/org/mapstruct/ap/test/bugs/_865/ProjectEntity.java
```diff
@@ -0,0 +1,50 @@
...
+    public List<ProjCoreUserEntity> getCoreUsers() {
+        return coreUsers;
+    }
+
+    public void setCoreUsers(List<ProjCoreUserEntity> coreUsers) {
+        this.coreUsers = coreUsers;
+    }
+
+
+
```
agudian: Three newlines... :)

agudian:
Awesome, @sjaakd! Looks very good :+1:!

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: > Hmm, it turns the setter method to the name of the getter. That might actually make sense, but I don't think we're checking if it actually exists. Hmm.

But we should be able to derive this in an alternative way as well, right? We have the `Type#Map<String, ExecutableElement> getPropertyReadAccessors()`

processor/src/test/java/org/mapstruct/ap/test/bugs/_865/ProjectEntity.java
sjaakd: They're free :)..

sjaakd:
> Awesome, @sjaakd! Looks very good 👍!

Thanks.. I just saw there are remainders of other these (the no getter test) as well. I'll add a testcase for that or remove it.

Then I'll merge and push.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: I'll make an issue out of it to check if its really needed. See #878

sjaakd:
please test this

Issue #878, opened 2016-08-31
Check whether `collectionGetterName` is really needed


sjaakd:
We have the`Type#Map<String, ExecutableElement> getPropertyReadAccessors()` to do a similar job, right?

See #865 for more info

sjaakd:
I reckon we need to determine if this is really needed in an SPI before the CR1 is out..

sjaakd:
Fixed. @agudian thanks for reviewing.

PR #881, opened 2016-09-01
#878 remove AcessorNamingStrategy#getCollectionGetterName from SPI

sjaakd:
Seems not to be needed.. Why was it there in the first place?

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -462,7 +462,7 @@ public class PropertyMapping extends ModelElement {
                     // target accessor is setter, so wrap the setter in setter map/ collection handling
                     result = new SetterWrapperForCollectionsAndMaps(
                         result,
-                        Executables.getCollectionGetterName( targetWriteAccessor ),
+                        targetReadAccessor.getSimpleName().toString(),
                         newCollectionOrMap,
                         targetType,
                         existingVariableNames
--- a/integrationtest/src/test/resources/namingStrategyTest/strategy/src/main/java/org/mapstruct/itest/naming/CustomAccessorNamingStrategy.java
+++ b/integrationtest/src/test/resources/namingStrategyTest/strategy/src/main/java/org/mapstruct/itest/naming/CustomAccessorNamingStrategy.java
@@ -24,7 +24,7 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.type.TypeKind;
 
 import org.mapstruct.ap.spi.AccessorNamingStrategy;
-import org.mapstruct.ap.spi.MethodType;
+import org.mapstruct.ap.spi.DefaultAccessorNamingStrategy;
 
 /**
  * A custom {@link AccessorNamingStrategy} recognizing getters in the form of {@code property()} and setters in the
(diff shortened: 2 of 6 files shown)


sjaakd:
Oh.. As @agudian pointed out.. we might check the presence of a getter, to avoid NPE's and give a nice error message that a getMethod is really mandatory for update methods and collections and if a setter is also missing..

But first determining why this is there...

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -462,7 +462,7 @@ private Assignment assignToCollection(Type targetType,
                     // target accessor is setter, so wrap the setter in setter map/ collection handling
                     result = new SetterWrapperForCollectionsAndMaps(
                         result,
-                        Executables.getCollectionGetterName( targetWriteAccessor ),
+                        targetReadAccessor.getSimpleName().toString(),
```
agudian: Didn't we have a problem with the toString() method in different Compiler-Implementations? Well, all tests pass, so it seems to be okay. :grin:

agudian:
The method was added in 3ba4ded58c0db9cbb9f343f47680e35f4583a1dc by @gunnarmorling during a refactoring of that interface - I can't see the old version having a similar method.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: dunno.. Indeed apparently not 😉  But perhaps @gunnarmorling knows whether there was some kind of `deeper` reason why this ended up in the SPI. If that's not the case, we should not do it..

sjaakd:
> The method was added in 3ba4ded by @gunnarmorling during a refactoring of that interface - I can't see the old version having a similar method.

ah.. Perhaps he still knows.. I think the reason must then go in the javadoc. Lets see Monday what his opinion is.

sjaakd:
> The method was added in 3ba4ded by @gunnarmorling during a refactoring of that interface - I can't see the old version having a similar method.


There is a comment on Dec 4, 2014, on #365 related to 3ba4ded . In a bullitlist is were it appears first. And.. the code just prior to that looked like `this.targetGetterName = "get" + targetSetterName.substring( 3 );`. So I understand why the original commit replaced this by a call to the SPI via `Executables`. However, the setter is available via other means as shown in this PR. So IMHO it makes the SPI more complicated then needed.

agudian:
Could the `targetReadAccessor` there still have a different name now, e.g. as in your example customization (`items()` instead of `getItems()`)? If that's still possible (e.g. because that information is already considered when building up the source-model), we would clearly not need this method anymore.

sjaakd:
> Could the targetReadAccessor there still have a different name now, e.g. as in your example customization (items() instead of getItems())? If that's still possible (e.g. because that information is already considered when building up the source-model), we would clearly not need this method anymore.

Each potentential executable method is given to the strategy.. The strategy decides what it is (getter, hasser, setter, adder), See: `AccessorNamingStrategy `. ` MethodType getMethodType(ExecutableElement method)` and which property it belongs to `String getPropertyName(ExecutableElement getterOrSetterMethod);` Within a `Type` a map is build up that maps a property to a setter, getter and hasser. 

So. The answer is yes I guess. The strategy decides.. You could even decide to handle collections differently, if there would be a need for it..

gunnarmorling:
Hey, we cannot really remove that method in a minor (1.1) as it'd break existing implementations. Right now I don't quite remember why it's needed, but I'm sure there was a reason :)

sjaakd:
> Hey, we cannot really remove that method in a minor (1.1) as it'd break existing implementations. Right now I don't quite remember why it's needed, but I'm sure there was a reason :)

The same thing can be achieved via `#MethodType AccessorNamingStrategy #getMethodType(ExecutableElement method);`. Here you can check whether the argument is a collection getter and return different behavior if desired from normal getters.

My problem is that it confuses the SPI. We should at least mark it deprecated and add explanation to it. 

But, personally, since an SPI is not used that often, we would document it and just remove it. If the user used an `@Overide` annotation they would notice it compile time as well.

agudian:
> But, personally, since an SPI is not used that often, we would document it and just remove it. If the user used an `@Override` annotation they would notice it compile time as well.

At runtime, there shouldn't even be a problem, as the class would load anyway (at least I'm pretty sure it would :wink:), so users would have to adapt their custom implementation only if they update the mapstruct version there as well. So I would side with Sjaak here that the impact is low, and the number of affected users likewise and we would do us and other users more of a favor if we just remove it now.

And didn't we introduce it as experimental in 1.0.0.Final? Or was that before the final?

But I can understand if @gunnarmorling is against it for formal reasons - and in that case we should deprecate it and throw an exception in the default implementation ("method is not supposed to be called" or something).

gunnarmorling:
I feel very strong about backwards compatability in minor updates, I've to often been bitten by libs doing breaking changes, causing trouble on the user side. If the method is removed, someone implementing the interface will get a compile error if their implementation is marked with `@Override`. 

Thus we only can remove in 2.0, but we can mark it as deprecated, explaining that it will not be invoked by the engine any longer as of 1.1.

sjaakd:
No problem! I'll mark it Deprecated. With explanation.

sjaakd:
What shall I do with this PR..? Keep it open to 2.0? Push its branch to MapStruct?

agudian:
> What shall I do with this PR..? Keep it open to 2.0? Push its branch to MapStruct?

I think we should merge the parts of this PR that remove the usage, do the refactoring of the method in `Executables` and remove it from the example / test / documentation SPI implementations. So basically everything but actually removing it from the interface. Kind of #885 plus this one. :)

sjaakd:
> I think we should merge the parts of this PR that remove the usage, do the refactoring of the method in Executables and remove it from the example / test / documentation SPI implementations. So basically everything but actually removing it from the interface. Kind of #885 plus this one. :)

Ok..  To be honest.. I misread the comment of @gunnarmorling :smile:.  I'll update, fix and merge.

Issue #689, opened 2015-11-13
Auto generated Joda LocalDate to XMLGregorianCalendar mapper produces runtime errors


gbrehmer:
I don't know whether this is offically supported because this conversion is not mentioned in the documentation.
Nevertheless the generated code generates runtime errors, because the formated string can not be parsed by gregorian calendar.

Example generated method:
```
stringToXmlGregorianCalendar( DateTimeFormat.forPattern( DateTimeFormat.patternForStyle( "L-", Locale.getDefault() ) ).print( obj.getLocalDate() ).trim() , null ) 
```

String Output before conversion to XMLGregorianCalendar e.g.: "1. Januar 2015"
Exception during conversion: java.lang.IllegalArgumentException: For input string: "1. Ja"

Joda 2.9

gbrehmer:
The strange thing here is, that in another case (other mapper / output model) with same input & output type, it generates another (correct) mapping method:

```
dateToXmlGregorianCalendar( obj.getLocalDate().toDate() )
```

gunnarmorling:
I don't think it is supported. Would you be interested in helping out with that one?

agudian:
I guess what we need is are specific LocalDate<->XmlGregorianCalendar built-in mapping methods to avoid the funny choosing of two-step conversions here (which might be influenced by what built-ins were added for other properties to the same mapper before).

sjaakd:
I'm not sure whether the conversion makes sense from `LocalDate` to `XmlGregorianCalendar`. The latter one contains date and time. The first one only a date. So, in essence you have to choose what time part to generate in `XmlGregorianCalendar`. I'm not in favor of MapStruct making that choice. Especially considering time-zones.

I reckon the following conversions make sense:
 DateTime <> XmlGregorianCalendar

And doubtful about l:
XmlGregorianCalendar > `LocalDate` 
XmlGregorianCalendar > `LocalTime`
XmlGregorianCalendar <> `LocalDateTime`

The latter ones immediately raise the question: "Which `LocalDate`  and which `LocalTime`. According to the Joda Javadoc, these do not contain time zone information. And their name suggest that they should represent the date and time locally (at the locale). 

This means that MapStruct would:
 
1. have to calculate the time somehow (or use java classes to do so), given de timezone information present in the XmlGregorianCalendar, taking into account things like DST and the likes.. 

2. Or, alternatively leave time zone information out.

Anyway.. both choices are user choices, that we cannot make.

[...]

PR #884, opened 2016-09-05
#689, add built in methods to/from xmlGregorianCalendar and Joda Types.

sjaakd:


sjaakd:
I'm doubting whether to put commit: 	185b9c7 in. Determining conversion from `LocalDateTime`, `LocalDate` en `LocalTime` should perhaps be a user choice.

I'm also doubting whether to set the timezone field in xmlGregorianCalendar to UNDEFINED for the local Joda types or to the default. You can make argument for both cases.

sjaakd:
As an additional thought, we should perhaps disallow 2 step mappings:
* 2x builtin
* conversion / builtin
* 2x conversion

With 'simple' intermediate types, such as `String` and `Date`.. I

agudian:
Hmm, something in the tests is still dependent on the locale of the build system.

agudian:
> As an additional thought, we should perhaps disallow or deprioritize 2 step mappings ...

:+1: - this tends to create some funny conversions (and as some user reports suggest, that might not always be stable, but can depend the specific beans a single mapper deals with). Not sure though, if we should completely disable it, or just make it the "last resort" (but with deterministic behaviour for each specific pair of types).

sjaakd:
> or just make it the "last resort" (but with deterministic behaviour for each specific pair of types).

That was exactly my thinking. I think we now break our search on the first match in a 2 step mapping. What if we don't do that and add a `Selector` that sorts on the -in-between-type-? So last prio:
* String
then
*Date
etc..

agudian:
Sounds good.

Wdyt, should this PR be part of the CR1 release? And the change to the selector as well?

sjaakd:
> Wdyt, should this PR be part of the CR1 release? And the change to the selector as well?

I think this is a separate issue (I'll write one). Lets keep this one centered around Joda, WDYT?

agudian:
Anything left to be done here?

sjaakd:
> Anything left to be done here?

In principle not.. I'm just doubting about commit 185b9c7.. WDYT? Shall I leave it out or in?

agudian:
When in doubt, then leave it out. It's an old saying, right? :wink:

I guess we can still discuss that in our next hangout with Gunnar in time for the Final.

sjaakd:
Ok.. so we reschedule this one for the final, right? Tomorrow CR1?

agudian:
Yes, let's wait with this another week and do the release tomorrow.

gunnarmorling:
@sjaakd @agudian Where are we with this one? Is it good to go?

sjaakd:
same doubt as in #893

gunnarmorling:
Hi @sjaakd, regarding

> In principle not.. I'm just doubting about commit 185b9c7.. WDYT? Shall I leave it out or in?

If in doubt leave it out? I'm not too keen on adding more stuff for Joda anyways (as it's been replaced with Java 8's API), so if we aren't sure let's not add this one. Could you update the PR?

sjaakd:
@gunnarmorling .. I leave it out and merge then?

sjaakd:
Done!

sjaakd:
@gunnarmorling : do you still want to review the remainder?

gunnarmorling:
Go for it if you think it's good. Otherwise I'll take a look tomorrow.

gunnarmorling:
+1. Could you also add a mention of the new conversions in the docs? Thanks!

PR #885, opened 2016-09-05
#878 adding deprecated warning for getCollectionGetterName

sjaakd:


agudian:
Could you maybe make it more prominent that MapStruct won't call this method anymore and adapt the default implementation to throw an exception? E.g. `IllegalStateException( "This method is not intended to be called anymore and will be removed in future versions" )`. Then we should be on the safe side not to accidentally use it again in a year or so... :wink:

Issue #852, opened 2016-08-15
Add built-in conversion from java.util.LocalDate to java.util.Date


pjlarson:
There is no mapper for java.util.LocalDate to java.util.Date

Granted there are a lot of combinations of things you're attempting to cover.

Since I need this and stubbornly refuse to stop using mapstruct since I already have it in my POM, I will attempt to submit a pull request.

sjaakd:
Ok.. Are you aware its always possible to define a hand-written method to do this for you?

gunnarmorling:
> Since I need this and stubbornly refuse to stop using mapstruct since

That's some stubbornness I like :) A PR is surely welcome, the existing
built-in conversions around dates should provide some guidance. Thanks a
lot!

pjlarson:
@sjaakd I am aware and have done just that. But it feels a bit strange that all the combinations similar to mine are covered in the built-ins, but this particular one is not. Potentially I will have mapstruct in many many projects, and it would be great if I could have java.time.LocalDate<->java.util.Date built in rather than have to manage my own library and manual mappings.

sjaakd:
In that case, by all means: please provide a PR. Thanks!

gunnarmorling:
+1 Looking forward to this contribution :)

gunnarmorling:
@pjlarson Any update on this? Would be nice to have it for the 1.1 CR which we've planned soon-ish.

pjlarson:
Hi Gunnar. Unfortunately when I put in the fix it broke a test related to a
gregoriancalendarxml test. I left for vacation so not looking at it. Can
you mark it for the following release and I'll do a pull request when I get
back. I will need help understanding what broke.

sjaakd:
Hi Peter, do you want one of us to take a look and fix it? It's up for CR1 next week.

pjlarson:
It's on my home PC and I don't have access to it while on vacation. If only
I had pushed to my fork before I left...

I have someone at home who's not a software person. I'll see if I can talk
them through git over the phone. Stay tuned...

pjlarson:
Hi guys, so sorry for the delay. I created a pull request. There is a test failing for an unrelated conversion. I could not figure it out on my own.

gunnarmorling:
Going to look into the test failure right now.

gunnarmorling:
[...]

Issue #886, opened 2016-09-05
Have more control on 2 step builtin / conversions


sjaakd:
Idea. Consider the following 2 step matches:
- Conversion / Conversion
- Builtin / Conversion (and vice versa)
- Builtin / Builtin 
- The discovery mechanism of these 2 step matches should add all possibilities (and not bail out on the first match).
- A new selector should favor (prioritize) more complex types as in-between-type over simple ones such as: String, Date. Perhaps even issue a warning on these types.

sjaakd:
The question I'm having here is whether this one is not solved by removing the 2 conversions + adding more and more builtin methods... The latter avoids strange conversions from String to xmlDate to Date.. 

I'm closing this one and re-open it when something pops up..

PR #893, opened 2016-09-13
Expanding converters to cover java 8 LocalDate to java.util.Date

pjlarson:
Hi gents, sorry for the delay getting this to you. I have followed the pattern I could see in the code to add another converter for java.util.Date <-> java.time.LocalDate

All the tests pass except one is breaking related to an XML Gregorian Calendar converter. I don't know the project well enough to figure out what got broken. Seems like maybe it is using the new converter somewhere as an intermediate? Or perhaps I am not using the right methods for converting and it is leaving the extra time zone hanging on it.

For to issue #852

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaLocalDateToDateConversion.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaLocalDateToDateConversion.java
@@ -1,0 +1,39 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaLocalDateToDateConversion.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * SimpleConversion for mapping {@link java.time.LocalDate} to
+ * {@link java.util.Date} and vice versa.
+ */
+public class JavaLocalDateToDateConversion extends SimpleConversion {
+
+    @Override
+    protected String getToExpression(ConversionContext conversionContext) {
+        return "java.util.Date.from( <SOURCE>.atStartOfDay( java.time.ZoneOffset.UTC ).toInstant() )";
```
gunnarmorling: Could you avoid the fully-qualified names here and instead publish the required import types via `getFromConversionImportTypes()`/`getToConversionImportTypes()`?

gunnarmorling:
Could you also reword your commit message so it starts with the issue key "#852 Expanding...". Also an update to the reference documentation for this new conversion would be great.

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaLocalDateToDateConversion.java
gunnarmorling: I agree it's a corner case. Not so much concerned about `atStartOfDay` but rather about the TZ being used. UTC may be right in one case, but in others you may want to use the default TZ (or even inject one, which we don't support at all yet).

Still I'd say this one makes rather sense as a default, the user can always override it using their own method.

gunnarmorling:
Rebased and applied. Thanks @pjlarson!

I've added an explicit built-in conversion for `XMLGregorianCalendar` <> `LocalDate` (the test failure was caused by an (in this case unwanted) two-step conversion). Also updated the docs.

@sjaakd Could you create an issue for controlling when/how two-step conversion kicks in as we discussed? Thanks!

sjaakd:
> @sjaakd Could you create an issue for controlling when/how two-step conversion kicks in as we discussed? Thanks!

Hey @gunnarmorling .. checkout  #886

Issue #891, opened 2016-09-09
NullPointerException in implicit type conversions


stoerr:
When upgrading from MapStruct 1.1.0.Beta2 to 1.1.0.CR1 we suddenly got some NullPointerExceptions when MapStruct tries to convert an Integer to a Long. If you try to map the following, you get code that will throw an NPE in the .longValue() call when nested.getPropInt is null. This might be related to issue #843 . The correct way would probably be to save the result of srcNestedPropInt( src ) in a local variable and check it for being null, or to introduce a method that does such conversions in a nullsafe way.

    @Mapper
    public interface BuggyMapper {

        class Src {
            public SrcNested getNested() {
                return new SrcNested();
            }
        }

        class SrcNested {
            public Integer getPropInt() {
                return null;
            }
        }

        class Dest {
            public void setPropLong(Long l) {
            }
        }

        @Mapping(source = "nested.propInt", target = "propLong")
        Dest convert(Src src);
    }

generates the following buggy code:

    public class BuggyMapperImpl implements BuggyMapper {

        @Override
        public Dest convert(Src src) {
            if (src == null) {
                return null;
            }

            Dest dest = new Dest();

            dest.setPropLong(srcNestedPropInt(src).longValue());

            return dest;
        }

        private Integer srcNestedPropInt(Src src) {

            if (src == null) {
                return null;
            }
            SrcNested nested = src.getNested();
            if (nested == null) {
                return null;
            }
            Integer propInt = nested.getPropInt();
            if (propInt == null) {
                return null;
            }
            return propInt;
        }
    }

sjaakd:
Merged & solved.. @stoerr thanks for reporting.

PR #897, opened 2016-09-15
#891, Making local variables for nested forge methods calls when  null check is needed.

sjaakd:
Optimization and fixing an NPE in the generated code.


processor/src/main/java/org/mapstruct/ap/internal/conversion/SimpleConversion.java
```diff
@@ -39,7 +38,7 @@
     @Override
     public Assignment to(ConversionContext conversionContext) {
         String toExpression = getToExpression( conversionContext );
-        return AssignmentFactory.createTypeConversion(
+        return new TypeConversion(
```
sjaakd: new line

processor/src/main/java/org/mapstruct/ap/internal/conversion/SimpleConversion.java
```diff
@@ -49,7 +48,7 @@ public Assignment to(ConversionContext conversionContext) {
     @Override
     public Assignment from(ConversionContext conversionContext) {
         String fromExpression = getFromExpression( conversionContext );
-        return AssignmentFactory.createTypeConversion(
+        return new TypeConversion(
```
sjaakd: new line.

processor/src/main/java/org/mapstruct/ap/internal/model/MappingBuilderContext.java
```diff
@@ -81,21 +81,21 @@
...
-         * <li>Direct Assignment (empty TargetAssignment)</li>
+         * <li>SourceRHS Assignment (empty TargetAssignment)</li>
          * <li>null, no assignment found</li>
          * </ol>
          */
         @SuppressWarnings("checkstyle:parameternumber")
         Assignment getTargetAssignment(Method mappingMethod, String mappedElement, Type sourceType, Type targetType,
                                        String targetPropertyName, FormattingParameters formattingParameters,
-                                       SelectionParameters selectionParameters, String sourceReference,
+                                       SelectionParameters selectionParameters, SourceRHS sourceReference,
```
sjaakd: old parameter name 'sourceReference`

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -399,7 +402,7 @@ private Assignment assignToPlainViaAdder(Type sourceType, Assignment rightHandSi
                 result = new AdderWrapper(
                     result,
                     method.getThrownTypes(),
-                    getSourceRef(),
+                    getSourceRHS().getSourceReference(),
                     sourceType
```
sjaakd: can be optimized... mouth source type and source reference are now present in Freemarker templates.

agudian:
Looks alright to me (although I only understand half of it). Could you also rename the first commit? ("#891 Implementation and unit est")

sjaakd:
> Looks alright to me (although I only understand half of it). Could you also rename the first commit? ("#891 Implementation and unit est")

No problem.. I'm worried though. I tried to refactor the `AdderWrapper` as well to follow similar lines. Did not manage yet. I can see were it goes wrong. I'll do a follow up on that one some day. 

This whole stuff is getting too fuzzy . Perhaps for 2.0 we should refactor this completely. Get rid of the wrapper mechanism and split all cases out in individual templates for each case. Its perhaps a lot of work, but makes it more transparent. 

I'll merge

PR #898, opened 2016-09-15
#895 Fix generating forged iterable methods for mapping multi-dimensional Array types

### LINKED ISSUE #895: Mapstruct cannot map List<byte[]> to byte[][] and reverse
Given the following example class:
```java
public class Test {
	static class ByteObj {
		byte[][] name;
		public ByteObj() {}
		public byte[][] getName() {return name;}
		public void setName(byte[][] name) {this.name = name;}
	}
	static class ListObj {
		List<byte[]> name;
		public ListObj() {}
		public List<byte[]> getName() {return new ArrayList<>();}
		public void setName(List<byte[]> name) {this.name = name;}
	}
	@Mapper
	static abstract class MyMapper {
		abstract ListObj convert(ByteObj b);
		abstract ByteObj convert(ListObj b);
	}
}
```
Mapstruct will generate the following code:
```java

@Generated(
    date = "2016-09-15T18:12:41+0200",
    comments = "version: 1.1.0.CR1, compiler: Eclipse JDT (IDE) 1.2.100.v20160418-1457, environment: Java 1.8.0_101 (Oracle Corporation)"
)
class MyMapperImpl extends MyMapper {

    @Override
    ListObj convert(ByteObj b) {
        if ( b == null ) {
            return null;
        }

        ListObj listObj = new ListObj();

        listObj.setName( byte[]ArrayTobyteArrayList( b.getName() ) );

        return listObj;
    }

    @Override
    ByteObj convert(ListObj b) {
        if ( b == null ) {
            return null;
        }

        ByteObj byteObj = new ByteObj();

        byteObj.setName( byteArrayListTobyte[]Array( b.getName() ) );

        return byteObj;
    }

    protected List<byte[]> byte[]ArrayTobyteArrayList(byte[][] byte_) {
        if ( byte_ == null ) {
            return null;
        }

        List<byte[]> list = new ArrayList<byte[]>();
        for ( byte[] byte[] : byte_ ) {
            list.add( byte[] );
        }

        return list;
    }

    protected byte[][] byteArrayListTobyte[]Array(List<byte[]> list) {
        if ( list == null ) {
            return null;
        }

        byte[][] byte[]Tmp = new byte[][list.size()];
        int i = 0;
        for ( byte[] byte[] : list ) {
            byte[]Tmp[i] = byte[];
            i++;
        }

        return byte[]Tmp;
    }
}
```

If I change the class to
```java
public class Test {
	static class ByteObj {
		String[] name;
		public ByteObj() {}
		public String[] getName() {return name;}
		public void setName(String[] name) {this.name = name;}
	}
	static class ListObj {
		List<String> name;
		public ListObj() {}
		public List<String> getName() {return new ArrayList<>();}
		public void setName(List<String> name) {this.name = name;}
	}
	@Mapper
	static abstract class MyMapper {
		abstract ListObj convert(ByteObj b);
		abstract ByteObj convert(ListObj b);
	}
}
```
it generates the following correct mapper-class:
```java
@Generated(
    date = "2016-09-15T18:15:08+0200",
    comments = "version: 1.1.0.CR1, compiler: Eclipse JDT (IDE) 1.2.100.v20160418-1457, environment: Java 1.8.0_101 (Oracle Corporation)"
)
class MyMapperImpl extends MyMapper {

    @Override
    ListObj convert(ByteObj b) {
        if ( b == null ) {
            return null;
        }

        ListObj listObj = new ListObj();

        listObj.setName( stringArrayToStringList( b.getName() ) );

        return listObj;
    }

    @Override
    ByteObj convert(ListObj b) {
        if ( b == null ) {
            return null;
        }

        ByteObj byteObj = new ByteObj();

        byteObj.setName( stringListToStringArray( b.getName() ) );

        return byteObj;
    }

    protected List<String> stringArrayToStringList(java.lang.String[] string) {
        if ( string == null ) {
            return null;
        }

        for ( String string_ : string ) {
            list.add( string_ );
        }

        return list;
    }

    protected java.lang.String[] stringListToStringArray(List<String> list) {
        if ( list == null ) {
            return null;
        }

        String[] stringTmp = new String[list.size()];
        int i = 0;
        for ( String string : list ) {
            stringTmp[i] = string;
            i++;
        }

        return stringTmp;
    }

}
```

Therefore I guess the issue is that Mapstruct uses "byte[]" as part of method and variable names if byte[][] or List<byte[]> are used. This leads to broken mapper classes.

agudian:
Fixes #895.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
@@ -57,12 +57,12 @@ public class ForgedMethod implements Method {
      */
     public ForgedMethod(String name, Type sourceType, Type targetType, MapperConfiguration mapperConfiguration,
         ExecutableElement positionHintElement) {
-        String sourceParamName = Strings.decapitalize( sourceType.getName().replace( "[]", "" ) );
+        String sourceParamName = Strings.decapitalize( sourceType.getName() );
         String sourceParamSafeName = Strings.getSaveVariableName( sourceParamName );
         this.parameters = Arrays.asList( new Parameter( sourceParamSafeName, sourceType ) );
         this.returnType = targetType;
         this.thrownTypes = new ArrayList<Type>();
-        this.name = name;
+        this.name = Strings.sanitizeIdentifierName( name );
         this.mapperConfiguration = mapperConfiguration;
         this.positionHintElement = positionHintElement;
     }
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/IterableMappingMethod.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/IterableMappingMethod.ftl
@@ -54,7 +54,8 @@
 
     <#if resultType.arrayType>
         <#if !existingInstanceMapping>
-            <@includeModel object=resultElementType/>[] ${resultName} = new <@includeModel object=resultElementType/>[<@iterableSize/>];
+            <#assign elementTypeString><@includeModel object=resultElementType/></#assign>
+            ${elementTypeString}[] ${resultName} = new ${elementTypeString?keep_before('[]')}[<@iterableSize/>]${elementTypeString?replace('[^\\[\\]]+', '', 'r')};
         </#if>
     <#else>
         <#if existingInstanceMapping>
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.java
@@ -41,7 +41,6 @@ public class SetterWrapperForCollectionsAndMaps extends AssignmentWrapper {
 
     private final String targetGetterName;
     private final Assignment newCollectionOrMapAssignment;
-    private final Type targetType;
     private final String localVarName;
 
     public SetterWrapperForCollectionsAndMaps(Assignment decoratedAssignment,
@@ -53,7 +52,6 @@ public class SetterWrapperForCollectionsAndMaps extends AssignmentWrapper {
 
         this.targetGetterName = targetGetterName;
         this.newCollectionOrMapAssignment = newCollectionOrMapAssignment;
-        this.targetType = targetType;
         this.localVarName = Strings.getSaveVariableName( targetType.getName(), existingVariableNames );
         existingVariableNames.add( localVarName );
     }
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/Strings.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/Strings.java
@@ -150,7 +150,7 @@ public class Strings {
      * any Java keyword; starting with a lower-case letter
      */
     public static String getSaveVariableName(String name, Collection<String> existingVariableNames) {
-        name = decapitalize( name );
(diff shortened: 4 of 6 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
```diff
@@ -57,12 +57,12 @@
      */
     public ForgedMethod(String name, Type sourceType, Type targetType, MapperConfiguration mapperConfiguration,
         ExecutableElement positionHintElement) {
-        String sourceParamName = Strings.decapitalize( sourceType.getName().replace( "[]", "" ) );
```
sjaakd: hmm .. was this superflous?

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableMappingMethod.ftl
```diff
@@ -54,7 +54,8 @@
 
     <#if resultType.arrayType>
         <#if !existingInstanceMapping>
-            <@includeModel object=resultElementType/>[] ${resultName} = new <@includeModel object=resultElementType/>[<@iterableSize/>];
+            <#assign elementTypeString><@includeModel object=resultElementType/></#assign>
```
sjaakd: should we do this kind of manipulation in the FreeMarker template or in the java code?

sjaakd:
@agudian : ok.. 2 questions above. Furthermore: shouldn't we add a reproducer?

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
agudian: That moved into `Strings.sanatizeIdentifierName`, which is called by `Strings.getSaveVariableName` now - so this is obsolte here.

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableMappingMethod.ftl
agudian: Yeah, that was my first thought as well. But I would either have to pull in the `iterableSize` macro into the java code as well, or provide explicit properties for the prefix and suffix... both seemed to blur the separation of responsibilities quite a lot... Or do you have a nice idea on how to split this?

agudian:
> Furthermore: shouldn't we add a reproducer?

Sure, I just forgot to add it to the commit. My git-client has some other defaults here at home and I just didn't notice last night :grin:

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
sjaakd: ack.

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableMappingMethod.ftl
sjaakd: nope. think its good enough..

sjaakd:
looks good 👍 . Just add the TC and merge.. Thanks.

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
gunnarmorling: sanatize=sanitize

agudian: Right, it _did_ look funny. :+1:

Issue #765, opened 2016-02-25
Decorator class generator can not handle void/update methods


gbrehmer:
In the decorator not overridden void methods are producing compile errors in the generated decorator part. Default wrapper code generated return statement:

`public void updateMethod(Dto dto, Entity entity) {
  return delegate.updateMethod(dto, entity);
}`

Workaround is to override those methods in the decorator and to simply call the delegate.

sjaakd:
Ai.. This is an issue.

gunnarmorling:
Done.

PR #906, opened 2016-09-21
#765 Omitting "return" key word when returning from decorated void me…

gunnarmorling:
…thod

@agudian @sjaakd There you go. Feels like my first MapStruct PR for ages :)

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/decorator/PersonMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/decorator/PersonMapper.java
@@ -21,6 +21,7 @@ package org.mapstruct.ap.test.decorator;
 import org.mapstruct.DecoratedWith;
 import org.mapstruct.Mapper;
 import org.mapstruct.Mapping;
+import org.mapstruct.MappingTarget;
 import org.mapstruct.factory.Mappers;
 
 @Mapper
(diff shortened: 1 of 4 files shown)


processor/src/test/java/org/mapstruct/ap/test/decorator/PersonMapper.java
```diff
@@ -33,4 +34,6 @@
     PersonDto personToPersonDto(Person person);
 
     AddressDto addressToAddressDto(Address address);
+
+    void updateAddressFromDto(AddressDto dto, @MappingTarget Address address);
```
sjaakd: what about the combi: `Address updateAddressFromDto(AddressDto dto, @MappingTarget Address address);`.. Should we have a UT for that as well?

sjaakd:
@gunnarmorling .. welcome back :smile: .. Just one comment ..

processor/src/test/java/org/mapstruct/ap/test/decorator/PersonMapper.java
gunnarmorling: This should be covered by `shouldDelegateNonDecoratedMethodsToDefaultImplementation()` in the same test.

gunnarmorling:
Yeah :)

gunnarmorling:
Rebased and applied. Thanks for reviewing, @sjaakd!

Issue #858, opened 2016-08-22
Add built-in methods for converting java.sql.Date, java.sql.Time and java.sql.Timestamp


filiphr:
Are you willing to add built in method for the following conversions?

* Date <> java.sql.Date
* Date <> java.sql.Time
* Date <> java.sql.TimeStamp
* java.sql.Date <> Date
* java.sql.Time <> Date
* java.sql.TimeStamp <> Date

gunnarmorling:
Hey, yes, we'd be more than willing to accept a pull request providing
these built-in conversions!

gunnarmorling:
The PR is merged, leaving this issue open for tracking:

* Updating the reference docs
* Avoiding the usage of deprecated methods in generated code

@filiphr Do you think you could take a look into these issues above? If not let me know and I can take over. Thanks!

filiphr:
@gunnarmorling No problem. I just created a pull request for the reference documentation.

As for the deprecated methods, I am not sure what is deprecated. I am using `new Time( <SOURCE>.getTime() )`, which is not deprecated. Can you tell me where did you see that deprecated methods are used? How can I see that?

gunnarmorling:
> As for the deprecated methods, I am not sure what is deprecated.

@filiphr, ah, sorry. For some reason Eclipse marked the constructor usage as deprecated, I reckon it was because it didn't open it in the Java source editor properly. All good I'd say.

gunnarmorling:
Done. Many thanks for this contribution @filiphr! If there is anything else you'd like to hack on, just let me know and we can discuss some interesting issues :)

filiphr:
@gunnarmorling, no problem, it was really nice working on this. I am open for more hacking. Which issues  do you have in mind? We can also move the discussion to some chat/email as well.

gunnarmorling:
Hi @filiphr Awesome, great to hear that you are up for more hacking on MapStruct :)

One interesting thing would be to support direct access of (public) fields, see issue #557. This has been requested several times and I think it's a great starter to learn more about the inner workings of the tool. If you are interested, we can have a discussion on #557 itself or on the mapstruct-users mailing list.

This one will also be a good first step of eventually supporting usage of MapStruct alongside with Lombok (see #510) which currently is sort of PITA.

PR #908, opened 2016-09-22
#858 Update documentation with java.sql.* implicit type conversions

filiphr:


gunnarmorling:
@agudian How again can we add @filiphr to the whitelist for CI builds?

Issue #909, opened 2016-09-26
mapping does not work when parameter name is 'values' 


zivry:
A really non descriptive error is thrown, and I looked for the reason of problem for hours. 
apparently the name of the parameter cannot be 'values', I sill fail to understand why. 
The exception: 

[...]

PR #910, opened 2016-09-26
#909 Avoid using 'values' as parameter name to work around a supsected issue in Freemarker

agudian:
Not really a fix, but a workaround for #909. I guess it might hit us at some other places as well.

I'll create an issue in the Freemarker issue tracker tomorrow or so.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
@@ -30,12 +30,14 @@ import org.mapstruct.ap.internal.util.Collections;
 public class Parameter extends ModelElement {
 
     private final String name;
+    private final String originalName;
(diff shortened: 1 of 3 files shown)


agudian:
retest this, please.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
```diff
@@ -49,6 +51,10 @@ public String getName() {
         return name;
     }
 
+    public String getOriginalName() {
```
gunnarmorling: Is that used anywhere?

gunnarmorling:
@agudian Looks good. I'll just add a small comment and merge. If you've created the issue with FreeMarker, can you open one for us, too, so we track it and get rid of the workaround eventually. Thanks!

Issue #557, opened 2015-05-22
Support field access if visibility rules permit


gunnarmorling:
If e.g. the target object is a plain "struct" with public fields, it should supported to write to these using direct field access. This should only be supported if the visibility rules permit that access, i.e. no reflection-based work-around otherwise.

GilraenBurland:
+1 for this.

cardosso:
Is this being considered?

For me this is the current deal breaker with MapStruct and the only reason I'm still using Orika. I don't want to fill my Hibernate domain objects with unnecessary getters and setters just to be able to use a mapping framework. It's a pitty this feature is missing because otherwise this is a great library.

This should only be done on public accessible fields and could even be made configurable (ie disabled by default).

gunnarmorling:
Hey @cardosso, yes, it is considered for sure. We just did not yet find the time to implement it. If you are interested in helping out, let me know and I can help you to get started with hacking on MapStruct.

I will schedule this for the 1.1 release.

cardosso:
That's great news thanks.
I never looked into the MapStruct source code, I guess I could check it out and see if I could be of some help implementing it if you could give me some quick guidelines. Not sure if I will have much free time during the next days though.

gunnarmorling:
Hi, any help would be more than welcome. If you are interested, you may check out the resources given [here](http://mapstruct.org/contributing/#section-03) for getting started. The change would have to include adaptions the logic for identifying mappable properties as well as the templates for generating code working with properties.

xiemeilong:
+1

filiphr:
@gunnarmorling I had a look at this issue. And I think I can start working on it. I had a deeper look at the code and the inner workings. I think that in order to do this, we would need to add our own `Accessor` class which would wrap the `ExecutableElemement`(s) and the `VariableElements`. In our own `Accessor` we would add the methods which we need and get them from the underlying `Element`. 
The `Type` methods `getPropertyPresenceCheckers`, `getPropertyReadAccessors` and `getPropertyWriteAccessors` will need to be adjusted to return our wrapper. 

This is only the first step, I will have to investigate a bit more, about how to proceed once we have the fields for mapping. 

What do you think? Am I on the right path?

filiphr:
Another thing, we are using the `AccessorNamingStrategy` with the `ExecutableElement`(s). I guess this is public and should not be modified?

gunnarmorling:
> What do you think? Am I on the right path?

[...]

Issue #572, opened 2015-06-20
Generated mapper missing protected fields


scriptacus:
I have a class where a variable called "name" and its setter are both marked "protected".  The generated mapper Impl class does not use the "setName()" function even though it is accessible.  This seems like a bug.

gunnarmorling:
Atm. only public getters/setters are considered. Indeed we may relax this in some cases (mapper and beans in same package).

pkainulainen:
It would be useful if MapStruct would support protected and package private visibility constraints for getters and setters. I agree that it's not a good idea to support this if the mapper and the beans are not in the same package, but if they are, this would be a very useful feature.

AppSturz:
Hi, after one and a half year can you please give an update if this feature will be added in the (near) future?

Issue #37, opened 2013-06-09
Add "verbose" processor option to print out details if required


gunnarmorling:

sjaakd:
see also #177 for more ideas.

agudian:
At some point we talked about where to add that output, but I can't find it anymore.

I see two options:
* as comments in the generated java file, e.g. with a special handling of the templates (the template writer could check if such info is attached to the model and apply a special comment-template before)
* in a separate file to be placed somewhere (but where?)

sjaakd:
solved by PR..

filiphr:
Just for traceability. The PR that solved this is https://github.com/mapstruct/mapstruct/pull/1741

PR #928, opened 2016-10-10
Provide a way to use field mappng within mapstruct

filiphr:
This is a PR for #557.

The PR is not complete, there are some things which are missing:
- A lot of tests are missing as I was not sure how to do them properly. Should I only update the current tests by adding public fields, or create completely new tests?
- I am also not 100% satisfied with the `afterWriteVariable` and `finishWriteVariable` in the freemarker. Is it possible this to be done via some macro or something similar?
- Do we need to provide a strategy that turns on/off field mapping?

However, I did it so we can have a discussion and see if I am on the right path.


gunnarmorling:
HI @filiphr sorry for being slow, just seeing this PR right now. Going to look into it and provide answers to your questions asap. Thanks a lot for working on this!

filiphr:
@gunnarmorling no problem. Whenever you have time. Will wait for your input to continue forward

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -85,13 +87,13 @@
...
+    private Map<String, Accessor> readAccessors = null;
+    private Map<String, ExecutableElementAccessor> presenceCheckers = null;
 
-    private List<ExecutableElement> allExecutables = null;
-    private List<ExecutableElement> setters = null;
-    private List<ExecutableElement> adders = null;
-    private List<ExecutableElement> alternativeTargetAccessors = null;
+    private List<Accessor> allAccessors = null;
+    private List<Accessor> setters = null;
+    private List<Accessor> adders = null;
```
gunnarmorling: Would an adder not always be a method? I can't see how it could be a field.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -378,21 +380,33 @@ public boolean isAssignableTo(Type other) {
                     modifiableGetters.put( Executables.getPropertyName( getter ), getter );
                 }
             }
-            getters = Collections.unmodifiableMap( modifiableGetters );
+
+            //TODO decide if we want to have a mechanism for turning on/off direct field mapping
```
gunnarmorling: I think I'd give preference to calling getters/setters and fall back to field access if they aren't there and the field is accessible for the generated code (either public or protected and living in the same package).

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
```diff
@@ -190,19 +251,29 @@ private static void addEnclosedElementsInHierarchy(Elements elementUtils, List<E
...
             }
         }
 
         alreadyCollected.addAll( 0, safeToAdd );
     }
 
+    private static void addFields(List<Accessor> alreadyCollected, List<VariableElement> variablesToAdd) {
+        List<Accessor> safeToAdd = new ArrayList<Accessor>( variablesToAdd.size() );
+        for ( VariableElement toAdd : variablesToAdd ) {
+            //TODO should we do some check?
```
gunnarmorling: What kind of check do you have in mind?

processor/src/main/java/org/mapstruct/ap/internal/util/accessor/Accessor.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public interface Accessor {
+
+    TypeMirror getAccessedType();
+
+    Name getSimpleName();
+
+    Set<Modifier> getModifiers();
+
+    ExecutableElement getExecutable();
```
gunnarmorling: I reckon that'd return null for a field accessor?

processor/src/test/java/org/mapstruct/ap/test/abstractclass/AbstractDto.java
```diff
@@ -22,6 +22,10 @@
  * @author Andreas Gudian
  */
 public class AbstractDto implements Identifiable {
+
+    //CHECKSTYLE:OFF
```
gunnarmorling: Does CS complain about public fields? I'd say we should modify the rule set in this case.

gunnarmorling:
> A lot of tests are missing as I was not sure how to do them properly. Should I only update the current tests by adding public fields, or create completely new tests?

That's a good question. Maybe make some amendments to key tests from the existing ones? @agudian, @sjaakd any good ideas around it?

> I am also not 100% satisfied with the afterWriteVariable and finishWriteVariable in the freemarker. Is it possible this to be done via some macro or something similar?

I'll leave that to @sjaakd. Ideally the model element knows itself whether to add the braces or not.

> Do we need to provide a strategy that turns on/off field mapping?

Personally I don't think so. I'd use getter/setter access by default and use field access as fallback if possible. In a way people turn it on or off by the way how they model the source and target types.

> However, I did it so we can have a discussion and see if I am on the right path.

+1 That's much appreciated. I definitely think it's going into the right direction. Thanks a lot for tackling this one!

gunnarmorling:
Some more questions:

* How does this play together with decorators and `@BeforeMapping`/`@AfterMapping`? Some tests for that would be nice.
* Does it take visibility of fields into account? I.e. we always could access public fields, package private fields only if the generated class lives in the same class as the bean(s) and never private fields.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: Yes they will be, they won't be fields. They are filtered from the `allAccessors`. I was thinking that it is better if we have something more generic, and not have the specific implementations referenced in other places

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: It is already like that. For the property read accessors, first I go through the "getter" methods and then through the fields. A field is added only if a "getter" could not be found.

For the property write accessors it is similar. The fields are last checked, they are added at the end of the `getAlternativeTargetAccessors`. I think I am correctly handling the the strategy there. 

Maybe it would be good to add tests for the specific classes that we have

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
filiphr: I was thinking something like with the overriden methods. Whether we should check if a field is already added (shadowed). However, I think it is not needed, as later on the first found field will be used and the rest will be ignored

processor/src/main/java/org/mapstruct/ap/internal/util/accessor/Accessor.java
filiphr: Yes it returns `null` for field accessor. I will add documentation to everything once the details are resolved.

filiphr:
@gunnarmorling Thanks a lot for your feedback. This is an interesting issue to tackle.

> Personally I don't think so. I'd use getter/setter access by default and use field access as fallback if possible. In a way people turn it on or off by the way how they model the source and target types.

You have a good point here. Fields are used as fallback right now, if a method is there it should be used (should write tests though).

> How does this play together with decorators and @BeforeMapping/@AfterMapping? Some tests for that would be nice.

Honestly, I didn't try it. I will add some tests that use those 2 annotations.

> Does it take visibility of fields into account? I.e. we always could access public fields, package private fields only if the generated class lives in the same class as the bean(s) and never private fields.

Currently the fields are filtered with `Executables.isFieldAccessor`. And at this moment only not static public fields are considered, need to improve this to avoid final fields for writing. I think that taking protected and package protected fields into consideration should be tackled in #572

filiphr:
@gunnarmorling regarding the decorators. Do you have some test in mind? I am not seeing how the field accessor will change something here.

processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/ArrayCopyWrapper.ftl
```diff
@@ -20,11 +20,11 @@
 -->
 <#if (thrownTypes?size == 0) >
     <@includeModel object=ext.targetType/> ${localVarName} = <@_assignment/>;
-    ${ext.targetBeanName}.${ext.targetWriteAccessorName}( Arrays.copyOf( ${localVarName}, ${localVarName}.length ) );
+    ${ext.targetBeanName}.${ext.targetWriteAccessorName}${afterWriteVariable}Arrays.copyOf( ${localVarName}, ${localVarName}.length )${finishWriteStatement};
```
filiphr: @sjaakd These are the points which I don't really like in my PR. The getters are similar.

processor/src/main/resources/org/mapstruct/ap/internal/model/MethodReference.ftl
```diff
@@ -37,7 +37,7 @@
                 <#-- a class is passed on for casting, see @TargetType -->
                 <@includeModel object=ext.targetType raw=true/>.class<#t>
             <#elseif param.mappingTarget>
-                 ${ext.targetBeanName}.${ext.targetReadAccessorName}()
+                 ${ext.targetBeanName}.${ext.targetReadAccessorName}
```
filiphr: @sjaakd This is an example getter in the templates

processor/src/main/java/org/mapstruct/ap/internal/util/ValueProvider.java
```diff
@@ -0,0 +1,66 @@
...
+import org.mapstruct.ap.internal.util.accessor.Accessor;
+
+/**
+ * This a wrapper class which provides the value that needs to be used in the models.
+ *
+ * It is used to provide the read value for a difference kind of {@link Accessor}.
+ *
+ * @author Filip Hrisafov
+ */
+public class ValueProvider {
```
filiphr: @sjaakd This is something which I implemented for the getters, so either the getter with () or the field is returned. We need something similar for the assignment

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AssignmentWrapper.java
```diff
@@ -81,8 +82,26 @@ public AssignmentType getType() {
         return decoratedAssignment.getType();
     }
 
-     @Override
+    @Override
     public boolean isUpdateMethod() {
         return decoratedAssignment.isUpdateMethod();
     }
+
+    public String getAfterWriteVariable() {
```
filiphr: @sjaakd this and the `getFinishWriteStatement()` are used to return the way we need to start/finish an assignment. If you have some ideas I would be glad to hear them. Maybe you have something better in your refactorings.

sjaakd:
@filiphr .. I'll take a look tomorrow evening.. thanks...

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -443,28 +457,32 @@ public boolean isAssignableTo(Type other) {
                 else if ( Executables.isGetterMethod( candidate ) ) {
                         // the current accessor is a getter (no setter available). But still, an add method is according
                     // to the above strategy (SETTER_PREFERRED || ADDER_PREFERRED) preferred over the getter.
-                    Type targetType = typeFactory.getReturnType(
-                            typeFactory.getMethodType( (DeclaredType) typeMirror, candidate )
-                    );
+                    Type targetType = typeFactory.getReturnType( (DeclaredType) typeMirror, candidate );
```
sjaakd: Hmm.. is the `typeFactory.getMethodType` not required..? Or is this handled by type factory?

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -296,10 +320,28 @@ public Parameter getSingleParameter(DeclaredType includingType, ExecutableElemen
...
+        }
+        else {
+            type = getType( accessorType );
+        }
+
+        return type;
+    }
+
+    private boolean isExecutableType(TypeMirror accessorType) {
+        return accessorType.getKind() == TypeKind.EXECUTABLE && accessorType instanceof ExecutableType;
```
sjaakd: I reckon the && is not needed, right?

sjaakd:
@filiphr . Nice work!!. I've added a remark on your questions. Have a look if it makes sense. 

Some generals:
* I did not review the tests yet.. I'll do so soon (this weekend).
* As an add on (please different PR) we could (I think that @gunnarmorling mentioned that already) the long lasting discussion on Lombok. E.g. we could add a `@MapperConfiguration` `@Mapper` attribute  `supportLombok` or rather, something more Lombok agnostic as `useFieldsInAbsenceOfAccessors` (contest for better names) or something along those lines. That would inspect fields and make assumptions on getter/setter names which are not yet present when MapStruct is active in the compilation cycle.

filiphr:
@sjaakd Thanks for the remarks. I'll try out the CommonMacros.

As for Lombok, yes we could make another PR and find a nice way of reusing this new feature. Currently, only public not static and not final fields are considered as read/write accessors and this is done in the Util functions, maybe we need to move this filtering to somewhere were we have access to the entire Mapping configuration

filiphr:
@gunnarmorling, @sjaakd I rebased on our latest master and applied the changes from your comments. @sjaakd Thanks a lot for the `CommonMacros` it helped me a lot and I implemented the assignments much better 😄.

If you have some idea which of our current tests can be extended to work with `public` fields it would be great. There are already tests which I updated, but I am not sure if it is enough

processor/src/test/java/org/mapstruct/ap/test/nestedproperties/simple/SimpleMapper.java
```diff
@@ -31,6 +31,7 @@
     SimpleMapper MAPPER = Mappers.getMapper( SimpleMapper.class );
 
     @Mappings( { @Mapping( target = "longValue", source = "props.longValue" ),
+        @Mapping( target = "publicLongValue", source = "props.publicLongValue" ),
```
sjaakd: do you have a reverse case somewhere as well?

sjaakd:
Hey @filiphr .. Monumental task 👍 .. Only one comment on the reverse nested testing.. This is a tricky area. Perhaps one case to make sure nothing broke there. Otherwise, I'm ok. @gunnarmorling , @agudian WDYT?

processor/src/test/java/org/mapstruct/ap/test/nestedproperties/simple/SimpleMapper.java
filiphr: Is it enough if I add a reverse method with `@InheritInverseConfiguration` or should I make a completely new one?

sjaakd: keep it simple.. Just reverse (`@InheritInverseConfiguration`) and make sure there are no sneaky issues :).

filiphr:
@sjaakd The reverse mapping works now as well. The only thing that this is missing at the moment is to update the documentation. I'll do this over the weekend. 

Should we log somewhere if a user starts using field mappings?

sjaakd:
> Should we log somewhere if a user starts using field mappings?

We should.. But probably not as part of this PR. I think that would demand a new `ReportingPolicy.INFO`, since I think that a warning would be not really applicable. Perhaps open up a new issue?

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
sjaakd: I think you can safely fix this.. :smile:

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
```diff
@@ -145,14 +184,35 @@ private static TypeElement asTypeElement(TypeMirror mirror) {
      * @return the executable elements usable in the type
      */
     public static List<ExecutableElement> getAllEnclosedExecutableElements(Elements elementUtils, TypeElement element) {
-        List<ExecutableElement> enclosedElements = new ArrayList<ExecutableElement>();
+        List<ExecutableElement> executables = new ArrayList<ExecutableElement>();
+        for ( Accessor accessor : getAllEnclosedAccessors( elementUtils, element ) ) {
+            if ( accessor.getExecutable() != null ) {
+                executables.add( accessor.getExecutable() );
+            }
+
```
sjaakd: empty line

sjaakd:
@filiphr .. I've some very tiny comments left. Have a look.

The only additional thing that sprang to my mind was the `AccessorNamingStrategy`. I'm not sure how relevant that use case is, but it ties read-accessor/write-accessor/presence-checker to a property by naming convention. For the read-accessor/write-accessor there's obviously no issue when calling public fields (properties) directly. However.. are there scenarios conceivable were you would maintain a separate boolean next to the field were you would maintain whether the property is presence or not? And would you do that by some kind of naming convention?

sjaakd:
otherwise.. as far as I'm concerned you are ready to squash and go with this one..

agudian:
Why should we log something in case the field mapping kicks in?

I find the info-logging in annotation processors always a bit annoying, as they are barely visible anywhere in the IDEs and not at all visible in mvn/javac builds. They just pile up in the Eclipse workspace log files... 😕

sjaakd:
I think there is a more general need to log the decisions Mapstruct makes, for debug purposes. I , for instance, find it hard why MapStruct reaches a decision sometimes, and I have an advantage over a non author.

agudian:
@sjaakd: #37 should cover that, right?

sjaakd:
> @sjaakd: #37 should cover that, right?

Right! Thats what I meant.

filiphr:
@agudian Here I mentioned logging on warning level, because this is something new and we can write that in the log message. That way if something comes up it might be faster for the user or us to find what went wrong. Some frameworks have `@Beta` annotations to represent that the feature is in Beta, this of course cannot be used for use as they don't call our methods 😄 

@sjaakd  regarding the `AccessorNamingStrategy`. Using a field `Accessor` has lowest priority and if it is picked it doesn't go through the `AccessorNamingStrategy`. I also wrote a comment about the presence-checkers in #557 and reply from @gunnarmorling 

>Good question, I reckon that it doesn't need to be changed. A field named "foo" really can only represent a property named "foo" and it cannot be a "presence checker" or any of the other method types.

I also think that it doesn't really make sense to have another field which would represent a presence-checker for another field. Currently all the presence-checker are explicitly `ExecutableElementAccessor` -> only methods can be presence-checkers.

I'll fix the comments from @sjaakd and add the documentation

agudian:
For new features where we expect that they might not yet work in all of the possibile scenarios, we tried to release them in a beta. For new features where we were not sure if the API or the behavior really fits the bill, we marked those features as *experimental* both in the Javadoc of the annotations (if applicable) and in the reference doc (I just saw that the `@BeforeMapping`/`@AfterMapping` annotations still have that tag... we might want to change that 😉).
That, plus the wording in the release announcment and changelog should suffice, I'd say.

For this feature, are there any questionable decisions that could be changed in the future with impact on what's mapped and what not? If not, I would feel that just releasing it in a beta should be just fine.

filiphr:
>For this feature, are there any questionable decisions that could be changed in the future with impact on what's mapped and what not? If not, I would feel that just releasing it in a beta should be just fine.

I don't think so. The only thing that can change that is is the relaxing of the visibility type of the field. Currently only public, not static fields can be used as read accessors, and public, not static and not final as write accessors. However, in my opinion this can be tackled in the same time as #572

sjaakd:
So.. am I right to conclude we have consensus on this and we can go ahead and merge? If so: lets move on the the next :smile: ..  I want the large impact ones out of the way to continue refactoring

filiphr:
@sjaakd I just thought of one more test that I can add. A public field with getter and setter (I know it makes no sense), but in this case we should use the methods and not the field itself :).

To be faster, I can write the documentation in a separate PR, so we can merge this faster.

sjaakd:
> To be faster, I can write the documentation in a separate PR, so we can merge this faster.

That would be ok for me.. Perhaps as part of the documentation you can add an example to our examples? You can refer to that from the documentation.

filiphr:
>That would be ok for me.. Perhaps as part of the documentation you can add an example to our examples? You can refer to that from the documentation.

Sure I'll add an example to our examples. When I add the documentation.

My last commit fixes a small bug that I found. If there was a getter then the field was not considered as a write accessor.

In the following class:

```java
public class Source {

    public Integer fieldOnlyWithGetter;

    public Integer getFieldOnlyWithGetter() {
        return fieldOnlyWithGetter;
    }
}
```
The write accessor could not be found. I think that now it should be OK to merge

processor/src/test/java/org/mapstruct/ap/test/fields/FieldsMappingTest.java
```diff
@@ -18,28 +18,36 @@
...
  */
 @RunWith(AnnotationProcessorTestRunner.class)
 @IssueKey( "557" )
-@WithClasses( { Source.class, Target.class, SourceTargetMapper.class} )
+@WithClasses({ Source.class, Target.class, SourceTargetMapper.class, Invocation.class })
 public class FieldsMappingTest {
 
+    @Before
+    public void setUp() throws Exception {
+        Source.INVOCATIONS.clear();
```
agudian: Instead of the invocations class, you could just modify the passed values in the setter/getter methods, eg. add 42 or mulitply with `-1` or so... 😉. Then you can avoid the copy of the `Invocation` class.

filiphr: You are right. That is much better, I didn't think about that. I'll change it and squash everything afterwards

filiphr:
Thanks to everyone for the reviews. I removed the changes that @agudian wrote, and once the build is done. I will squash and merge everything into master

Issue #913, opened 2016-09-29
Generated code does not compile in 1.1.0.CR1 it compiled fine in 1.1.0.Beta2


aspan:
`List<String>` is incorrectly assigned to a `Set<String>` local variable in 1.1.0.CR1 in 1.1.0.Beta2 no such assignment is made.
## The objects to map:

``` java
public class Domain {
    private Set<String> strings = new HashSet<>();
    public Set<String> getStrings() {
        return strings;
    }
    public void setStrings(Set<String> strings) {
        this.strings = strings;
    }
}

public class Dto {
    private List<String> strings = new ArrayList<>();
    public List<String> getStrings() {
        return strings;
    }
    public void setStrings(List<String> strings) {
        this.strings = strings;
    }
}
```
## The mapper:

``` java
@Mapper
public interface DomainDtoMapper {
    Domain update(Dto source, @MappingTarget Domain target);
}
```
## 1.1.0.Beta2 generated code:

``` java
public class DomainDtoMapperImpl implements DomainDtoMapper {
    @Override
    public Domain update(Dto source, Domain target) {
        if ( source == null ) {
            return null;
        }
        if ( source.getStrings() != null ) {
            if ( target.getStrings() != null ) {
                target.getStrings().clear();
                target.getStrings().addAll( source.getStrings() );
            }
            else {
                target.setStrings( new HashSet<String>( source.getStrings() ) );
            }
        }
        return target;
    }
}
```
## 1.1.0.CR1 generated code:

``` java
public class DomainDtoMapperImpl implements DomainDtoMapper {
    @Override
    public Domain update(Dto source, Domain target) {
        if ( source == null ) {
            return null;
        }
        if ( source.getStrings() != null ) {
            Set<String> set = source.getStrings();
            if ( target.getStrings() != null ) {
                target.getStrings().clear();
                if ( set != null ) {
                    target.getStrings().addAll( set );
                }
            }
            else {
                target.setStrings( new HashSet<String>( source.getStrings() ) );
            }
        }
        return target;
    }
}
```
## The Diff left 1.1.0.Beta2, right 1.1.0.CR1:

``` diff
7a8
>             Set<String> set = source.getStrings();
10c11,13
<                 target.getStrings().addAll( source.getStrings() );
---
>                 if ( set != null ) {
>                     target.getStrings().addAll( set );
>                 }
```

`source.getStrings()` returns a `List<String>` not a `Set<String>`

sjaakd:
[...]

Issue #289, opened 2014-08-29
When mapping into a read-only collection, a NULL check of the source collection is missing


vgigerich:
I created a mapping into a collection which is only accessible by its getter.
The generated mapper does a NULL check on the target read-only collection, but I would have expected a NULL check on the source collection too.

Generated code looks like this:
```
if ( target.getTagetCollection() != null ) {
    target.getTargetCollection().addAll( toSourceType( source.getSourceCollection() ) );
}
```
Expected something like:
```
if ( target.getTagetCollection() != null  && source.getSourceCollection != null ) {
    target.getTargetCollection().addAll( toSourceType( source.getSourceCollection() ) );
}
```

sjaakd:
Or.. you add: `@Mapper(nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT)`

sjaakd:
merged

PR #932, opened 2016-10-19
#913 uncompilable code generated in 1.1.0.CR1 for collection properties

sjaakd:
The solution is twofold ( 1 & 2) but with some consequences ( 3 & 4):
1. It introduces the new ideas I have for templates. It uses the `<#nested>` in Freemarker to wrap statements in a null check or try-catch clause. The idea is to gradually phase out the other `xyzWrapper` and only keep an lefthandside (target) and righthandside (source), making stuff a lot simpeler.
2. Delegate the `NullValueMappingStrategy` to the forged Map/Iterable mapping methods and let them handle default situations. This requires that the user is warned for creating plausible null pointer situations. 
3. Choose (in case the user did not define a `NullValueMappingStrategy`  explicitly for forged methods the `NullValueMappingStrategy.RETURN_DEFAULT` iso  `NullValueMappingStrategy.RETURN_NULL`. This seems more in line with the idea that a used mapping method itself is responsible for handling null input.
4. There's a drawback, a testcase (#865) starts to fail that expects null to be returned when both source and target collection, which seems reasonable. Possibly bring back some null checks on source here.

Next step (after CR2): 

a. rebasing this on #903 it should be simple to get rid of the `NewCollectionOrMapWrapper` and `EnumSetCopyWrapper`. 

b. As further optimization: a local variable is not required when there's no NPE check.

c. Plain property mappings in a similar way?

There seems to be an issue with delegating the `NullValueMappingStrategy` from `@Mapper` to `@BeanMapping`

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -19,7 +19,6 @@
 package org.mapstruct.ap.internal.model;
 
 import static org.mapstruct.ap.internal.model.assignment.Assignment.AssignmentType.DIRECT;
-import static org.mapstruct.ap.internal.model.assignment.Assignment.AssignmentType.MAPPED;
 import static org.mapstruct.ap.internal.model.assignment.Assignment.AssignmentType.MAPPED_TYPE_CONVERTED;
 import static org.mapstruct.ap.internal.model.assignment.Assignment.AssignmentType.TYPE_CONVERTED;
 import static org.mapstruct.ap.internal.model.assignment.Assignment.AssignmentType.TYPE_CONVERTED_MAPPED;
@@ -38,7 +37,6 @@ import org.mapstruct.ap.internal.model.assignment.Assignment;
 import org.mapstruct.ap.internal.model.assignment.EnumConstantWrapper;
 import org.mapstruct.ap.internal.model.assignment.EnumSetCopyWrapper;
 import org.mapstruct.ap.internal.model.assignment.GetterWrapperForCollectionsAndMaps;
-import org.mapstruct.ap.internal.model.assignment.LocalVarWrapper;
 import org.mapstruct.ap.internal.model.assignment.NewCollectionOrMapWrapper;
 import org.mapstruct.ap.internal.model.assignment.NullCheckWrapper;
 import org.mapstruct.ap.internal.model.assignment.SetterWrapper;
@@ -416,8 +414,7 @@ public class PropertyMapping extends ModelElement {
             return result;
         }
 
-        private Assignment assignToCollection(Type targetType,
-                                            TargetWriteAccessorType targetAccessorType,
+        private Assignment assignToCollection(Type targetType, TargetWriteAccessorType targetAccessorType,
                                             Assignment rhs) {
 
             Assignment result = rhs;
@@ -457,46 +454,25 @@ public class PropertyMapping extends ModelElement {
                         targetType );
                 }
                 else {
-
-                    if ( method.isUpdateMethod() ) {
-                        // if the calling method is an update method and accesses a getter, make a local variable to
-                        // test NPE first.
-                        result = new LocalVarWrapper( result, method.getThrownTypes(), targetType );
-                    }
-                    else {
-                        // if not, asssign as new collecitin or direct
-                        result = new SetterWrapper( result, method.getThrownTypes() );
-                    }
-
                     // target accessor is setter, so wrap the setter in setter map/ collection handling
                     result = new SetterWrapperForCollectionsAndMaps(
                         result,
-                        targetReadAccessor.getSimpleName().toString(),
                         newCollectionOrMap,
-                        targetType,
-                        existingVariableNames
+                        method.getThrownTypes(),
+                        getSourcePresenceCheckerRef(),
+                        existingVariableNames,
+                        targetType
                     );
                 }
 
             }
             else {
                 // target accessor is getter, so wrap the setter in getter map/ collection handling
-                result = new GetterWrapperForCollectionsAndMaps(
-                    result,
-                    method.getThrownTypes(),
-                    ctx.getTypeFactory().asCollectionOrMap( targetType ),
-                    existingVariableNames
-                );
-            }
-
-            // For collections and maps include a null check, when the assignment type is DIRECT.
-            // for mapping methods (builtin / custom), the mapping method is responsible for the
-            // null check. Typeconversions do not apply to collections and maps.
-            if ( result.getType() == DIRECT ) {
-                result = new NullCheckWrapper( result,  getSourcePresenceCheckerRef() );
-            }
-            else if ( result.getType() == MAPPED && result.isUpdateMethod() ) {
-                result = new UpdateNullCheckWrapper( result, getSourcePresenceCheckerRef() );
+                result = new GetterWrapperForCollectionsAndMaps( result,
+                                                                 method.getThrownTypes(),
+                                                                 getSourcePresenceCheckerRef(),
+                                                                 existingVariableNames,
+                                                                 targetType);
             }
 
             return result;
@@ -623,7 +599,6 @@ public class PropertyMapping extends ModelElement {
             ExecutableElement element) {
 
             Assignment assignment = null;
-
             String name = getName( sourceType, targetType );
             name = Strings.getSaveVariableName( name, ctx.getNamesOfMappingsToGenerate() );
 
@@ -687,7 +662,7 @@ public class PropertyMapping extends ModelElement {
             return assignment;
         }
 
-        private String getName(Type sourceType, Type targetType) {
+        private String getName(Type sourceType, Type targetType ) {
             String fromName = getName( sourceType );
             String toName = getName( targetType );
             return Strings.decapitalize( fromName + "To" + toName );
@@ -701,6 +676,7 @@ public class PropertyMapping extends ModelElement {
             builder.append( type.getIdentification() );
             return builder.toString();
         }
+
     }
 
     public static class ConstantMappingBuilder extends MappingBuilderBase<ConstantMappingBuilder> {
@@ -770,12 +746,10 @@ public class PropertyMapping extends ModelElement {
                 else {
 
                     // target accessor is getter, so getter map/ collection handling
-                    assignment = new GetterWrapperForCollectionsAndMaps(
-                        assignment,
-                        method.getThrownTypes(),
-                        ctx.getTypeFactory().asCollectionOrMap( targetType ),
-                        existingVariableNames
-                    );
+                    assignment = new GetterWrapperForCollectionsAndMaps( assignment,
+                                                                         method.getThrownTypes(),
+                                                                         existingVariableNames,
+                                                                         targetType);
                 }
             }
             else {
@@ -840,12 +814,10 @@ public class PropertyMapping extends ModelElement {
             }
             else {
                 // target accessor is getter, so wrap the setter in getter map/ collection handling
-                assignment = new GetterWrapperForCollectionsAndMaps(
-                    assignment,
-                    method.getThrownTypes(),
-                    ctx.getTypeFactory().asCollectionOrMap( targetType ),
-                    existingVariableNames
-                );
+                assignment = new GetterWrapperForCollectionsAndMaps( assignment,
+                                                                     method.getThrownTypes(),
+                                                                     existingVariableNames,
+                                                                     targetType);
             }
 
             return new PropertyMapping(
@@ -859,6 +831,7 @@ public class PropertyMapping extends ModelElement {
                 null
             );
         }
+
     }
 
     // Constructor for creating mappings of constant expressions.
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/Assignment.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/Assignment.java
@@ -106,4 +106,5 @@ public interface Assignment {
 
 
     boolean isUpdateMethod();
+
 }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/GetterWrapperForCollectionsAndMaps.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/GetterWrapperForCollectionsAndMaps.java
@@ -18,14 +18,10 @@
  */
 package org.mapstruct.ap.internal.model.assignment;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
 import org.mapstruct.ap.internal.model.common.Type;
-import org.mapstruct.ap.internal.util.Strings;
 
 /**
  * This wrapper handles the situation were an assignment must be done via a target getter method because there
@@ -40,54 +36,39 @@ import org.mapstruct.ap.internal.util.Strings;
  *
  * @author Sjaak Derksen
  */
-public class GetterWrapperForCollectionsAndMaps extends AssignmentWrapper {
+public class GetterWrapperForCollectionsAndMaps extends WrapperForCollectionsAndMaps {
 
-    private final List<Type> thrownTypesToExclude;
-    private final Type localVarType;
-    private final String localVarName;
+    /**
+     * constructor for property mapping
+     *
+     * @param decoratedAssignment
+     * @param thrownTypesToExclude
+     * @param sourcePresenceChecker
+     * @param existingVariableNames
+     * @param targetType
+     */
+    public GetterWrapperForCollectionsAndMaps(Assignment decoratedAssignment,
+                                              List<Type> thrownTypesToExclude,
+                                              String sourcePresenceChecker,
+                                              Set<String> existingVariableNames,
+                                              Type targetType ) {
 
-
-    public GetterWrapperForCollectionsAndMaps(Assignment decoratedAssignment, List<Type> thrownTypesToExclude,
-                                              Type localVarType, Collection<String> existingVariableNames) {
-        super( decoratedAssignment );
-        this.thrownTypesToExclude = thrownTypesToExclude;
-        this.localVarType = localVarType;
-        this.localVarName = Strings.getSaveVariableName( "target" + localVarType.getName(), existingVariableNames );
-        existingVariableNames.add( localVarName );
-   }
-
-    @Override
-    public List<Type> getThrownTypes() {
-        List<Type> parentThrownTypes = super.getThrownTypes();
-        List<Type> result = new ArrayList<Type>( parentThrownTypes );
-        for ( Type thrownTypeToExclude : thrownTypesToExclude ) {
-            for ( Type parentThrownType : parentThrownTypes ) {
-                if ( parentThrownType.isAssignableTo( thrownTypeToExclude ) ) {
-                    result.remove( parentThrownType );
-                }
-            }
-        }
-        return result;
-    }
-
-    @Override
-    public Set<Type> getImportTypes() {
-        Set<Type> imported = new HashSet<Type>();
-        imported.addAll( super.getImportTypes() );
-        imported.add( localVarType ); /* is a local var */
-        imported.addAll( localVarType.getTypeParameters() );
-        return imported;
+        super( decoratedAssignment, thrownTypesToExclude, sourcePresenceChecker, existingVariableNames, targetType );
     }
 
     /**
-     * @return the targetType
+     * constructor for e.g. constants and expressions
+     *
+     * @param decoratedAssignment
+     * @param thrownTypesToExclude
+     * @param existingVariableNames
+     * @param targetType
      */
-    public Type getLocalVarType() {
-        return localVarType;
-    }
+    public GetterWrapperForCollectionsAndMaps(Assignment decoratedAssignment,
+                                              List<Type> thrownTypesToExclude,
+                                              Set<String> existingVariableNames,
+                                              Type targetType ) {
 
-    public String getLocalVarName() {
-        return localVarName;
+        super( decoratedAssignment, thrownTypesToExclude, null, existingVariableNames, targetType );
     }
-
 }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.java
@@ -18,12 +18,10 @@
  */
 package org.mapstruct.ap.internal.model.assignment;
 
-import java.util.Collection;
-import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import org.mapstruct.ap.internal.model.common.Type;
-import org.mapstruct.ap.internal.util.Strings;
 
 /**
  * This wrapper handles the situation were an assignment is done via the setter.
@@ -37,45 +35,33 @@ import org.mapstruct.ap.internal.util.Strings;
  *
  * @author Sjaak Derksen
  */
-public class SetterWrapperForCollectionsAndMaps extends AssignmentWrapper {
+public class SetterWrapperForCollectionsAndMaps extends WrapperForCollectionsAndMaps {
 
-    private final String targetGetterName;
     private final Assignment newCollectionOrMapAssignment;
-    private final String localVarName;
 
     public SetterWrapperForCollectionsAndMaps(Assignment decoratedAssignment,
-                                              String targetGetterName,
                                               Assignment newCollectionOrMapAssignment,
-                                              Type targetType,
-                                              Collection<String> existingVariableNames) {
-        super( decoratedAssignment );
+                                              List<Type> thrownTypesToExclude,
+                                              String sourcePresenceChecker,
+                                              Set<String> existingVariableNames,
+                                              Type targetType ) {
 
-        this.targetGetterName = targetGetterName;
+        super( decoratedAssignment, thrownTypesToExclude, sourcePresenceChecker, existingVariableNames, targetType );
         this.newCollectionOrMapAssignment = newCollectionOrMapAssignment;
-        this.localVarName = Strings.getSaveVariableName( targetType.getName(), existingVariableNames );
-        existingVariableNames.add( localVarName );
-    }
-
-    public String getTargetGetterName() {
-        return targetGetterName;
-    }
-
-    public Assignment getNewCollectionOrMapAssignment() {
-        return newCollectionOrMapAssignment;
     }
 
     @Override
     public Set<Type> getImportTypes() {
-        Set<Type> imported = new HashSet<Type>();
-        imported.addAll( getAssignment().getImportTypes() );
+        Set<Type> imported = super.getImportTypes();
         if ( newCollectionOrMapAssignment != null ) {
             imported.addAll( newCollectionOrMapAssignment.getImportTypes() );
         }
         return imported;
     }
 
-    public String getLocalVarName() {
-        return localVarName;
+    public Assignment getNewCollectionOrMapAssignment() {
+        return newCollectionOrMapAssignment;
     }
+
 
 }
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.ftl
@@ -18,36 +18,40 @@
      limitations under the License.
 
 -->
+<#import "../macro/CommonMacros.ftl" as lib>
+<@lib.handleExceptions>
 <#if ( ext.existingInstanceMapping ) >
-    <@_assignment targetWriteAccessorName = localVarName/>
-    if ( ${ext.targetBeanName}.${targetGetterName}() != null ) {
-        ${ext.targetBeanName}.${targetGetterName}().clear();
-        if ( ${localVarName} != null ) {
+    if ( ${ext.targetBeanName}.${ext.targetReadAccessorName}() != null ) {
+        <@lib.handleNullCheck>
+            ${ext.targetBeanName}.${ext.targetReadAccessorName}().clear();
             ${ext.targetBeanName}.${ext.targetReadAccessorName}().<#if ext.targetType.collectionType>addAll<#else>putAll</#if>( ${localVarName} );
+        </@lib.handleNullCheck>
+        <#if !ext.defaultValueAssignment??> <#-- the opposite (defaultValueAssignment) case is handeld inside lib.handleNullCheck -->
+        else {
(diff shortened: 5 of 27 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_289/Issue289Mapper.java
```diff
@@ -31,11 +33,9 @@
 
     void sourceToTargetWithoutSetter(Source source, @MappingTarget TargetWithoutSetter target);
 
+    @BeanMapping( nullValueMappingStrategy = NullValueMappingStrategy.RETURN_NULL )
```
sjaakd: should not be required anymore

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -681,7 +675,8 @@ private Assignment forgeMapMapping(Type sourceType, Type targetType, SourceRHS s
         private String getName(Type sourceType, Type targetType) {
             String fromName = getName( sourceType );
             String toName = getName( targetType );
-            return Strings.decapitalize( fromName + "To" + toName );
+            String with = nullValueMappingStrategy != null ? "With" + nullValueMappingStrategy.name() : "";
```
sjaakd: The generated name follows the enum constant. So capitals and underscore.. Should I decapitilize?

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -394,6 +399,19 @@ private void useLocalVarWhenNested(Assignment rightHandSide) {
             }
         }
 
+        private String getSafeName() {
+            String name;
+            if ( sourceReference.getPropertyEntries().isEmpty() ) {
+                name = sourceReference.getParameter().getName();
+            }
+            else {
+                name = sourceReference.getPropertyEntries().get( 0 ).getName();
```
sjaakd: use first()

sjaakd:
I'll process my own comments together with the others (when there).

gunnarmorling:
> rebasing this on #903 it should be simple to get rid of the NewCollectionOrMapWrapper and EnumSetCopyWrapper. 

Do you mean it the other way around I reckon? Rebase #903 on top of this?

gunnarmorling:
@sjaakd do you think you could share an example of the generated code (and how it differs from the previous one?

processor/src/test/java/org/mapstruct/ap/test/bugs/_289/Issue289Test.java
```diff
@@ -79,4 +82,46 @@ public void shouldLeaveNullTargetSetWhenSourceIsNullForCreateMethod() {
...
     }
+
+    @Test
+    @ExpectedCompilationOutcome(
+        value = CompilationResult.SUCCEEDED,
+        diagnostics = {
+            @Diagnostic(type = Issue289MapperTargetOnlyGetterWithWarning.class,
+                kind = javax.tools.Diagnostic.Kind.WARNING,
+                line = 33,
+                messageRegExp = "NullPointer risk when only a read accessor is available and applying "
```
gunnarmorling: This one makes me curious. Maybe we can talk about it in the call tonight?

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -681,7 +676,9 @@ private Assignment forgeMapMapping(Type sourceType, Type targetType, SourceRHS s
         private String getName(Type sourceType, Type targetType) {
             String fromName = getName( sourceType );
             String toName = getName( targetType );
-            return Strings.decapitalize( fromName + "To" + toName );
+            String with = nullValueMappingStrategy != null ?
+                "WithNvms" + camelizeEnumConstant( nullValueMappingStrategy ) : "";
```
agudian: "camelize"... that's flowerly... 🐫 👍

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -692,6 +689,43 @@ private String getName(Type type) {
...
+
+        private NullValueMappingStrategyPrism determineForgedNullValueMappingStrategy() {
+
+            NullValueMappingStrategyPrism nvms;
+
+            warnWhenTargetWriteAccessorIsGetterAndNullValueMappingIsExplicitReturnNull();
+            warnWhenUpdateMethodAndNullValueMappingIsExplicitReturnNull();
+
+            if ( nullValueMappingStrategy == null
+                && ( method.isUpdateMethod() || TargetWriteAccessorType.GETTER.equals( targetWriteAccessorType ) ) ) {
```
agudian: use `==` for enum equality check, please... 😃

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -692,6 +689,43 @@ private String getName(Type type) {
...
+            else {
+                nvms = nullValueMappingStrategy;
+            }
+            return nvms;
+        }
+
+        private void warnWhenTargetWriteAccessorIsGetterAndNullValueMappingIsExplicitReturnNull() {
+
+            if ( TargetWriteAccessorType.GETTER.equals( targetWriteAccessorType )
+                && NullValueMappingStrategyPrism.RETURN_NULL.equals( nullValueMappingStrategy ) ) {
```
agudian: use `==` for enum equality check, please... 😃

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -692,6 +689,43 @@ private String getName(Type type) {
...
+                ctx.getMessager().printMessage( method.getExecutable(),
+                    Message.PROPERTYMAPPING_POSSIBLE_NULL_VALUE_WHEN_NO_WRITE_ACCESSOR_TYPE,
+                    targetPropertyName );
+            }
+        }
+
+        private void warnWhenUpdateMethodAndNullValueMappingIsExplicitReturnNull() {
+
+            if ( method.isUpdateMethod()
+                && NullValueMappingStrategyPrism.RETURN_NULL.equals( nullValueMappingStrategy ) ) {
```
agudian: use `==` for enum equality check, please... 😃

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/GetterWrapperForCollectionsAndMaps.java
```diff
@@ -74,20 +65,18 @@ public GetterWrapperForCollectionsAndMaps(Assignment decoratedAssignment, List<T
...
-    }
-
-    public String getLocalVarName() {
-        return localVarName;
+    @Override
+    public boolean isIncludeNullCheck() {
+        // For collections and maps include a null check, when the assignment type is DIRECT.
+        // for mapping methods (builtin / custom), the mapping method is responsible for the
+        // null check. Typeconversions do not apply to collections and maps.
+        return AssignmentType.DIRECT.equals( getType() );
```
agudian: use `==` for enum equality check, please... 😃

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.java
```diff
@@ -65,17 +63,35 @@ public Assignment getNewCollectionOrMapAssignment() {
...
     }
 
-    public String getLocalVarName() {
-        return localVarName;
+    @Override
+    public boolean isIncludeNullCheck() {
+        // For collections and maps include a null check, when the assignment type is DIRECT.
+        // for mapping methods (builtin / custom), the mapping method is responsible for the
+        // null check. Typeconversions do not apply to collections and maps.
+        return AssignmentType.DIRECT.equals( getType() );
```
agudian: use `==` for enum equality check, please... 😃

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -53,6 +54,8 @@
     PROPERTYMAPPING_INVALID_PROPERTY_NAME( "No property named \"%s\" exists in source parameter(s)." ),
     PROPERTYMAPPING_NO_PRESENCE_CHECKER_FOR_SOURCE_TYPE( "Using custom source value presence checking strategy, but no presence checker found for %s in source type." ),
     PROPERTYMAPPING_NO_READ_ACCESSOR_FOR_TARGET_TYPE( "No read accessor found for property \"%s\" in target type." ),
+    PROPERTYMAPPING_POSSIBLE_NULL_VALUE_WHEN_NO_WRITE_ACCESSOR_TYPE( "NullPointer risk when only a read accessor is available and applying NullValueMappingStrategy.RETURN_NULL for property \"%s\" in target type.", Diagnostic.Kind.WARNING ),
+    PROPERTYMAPPING_POSSIBLE_NULL_VALUE_UPDATE_METHOD_ACCESSOR_TYPE( "NullPointer risk when using update method and applying NullValueMappingStrategy.RETURN_NULL for property \"%s\" in target type.", Diagnostic.Kind.WARNING ),
```
agudian: Could you give us an example of what exactly that means? It could perhaps be something for our to-be-created FAQ page...

processor/src/test/java/org/mapstruct/ap/test/bugs/_289/Issue289Mapper.java
```diff
@@ -35,7 +35,4 @@
 
     TargetElement sourceElementToTargetElement(SourceElement source);
 
-
-
-
```
agudian: 👍

processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.ftl
```diff
@@ -18,41 +18,41 @@
...
+<#import "../macro/SourceUtil.ftl" as source>
+<#import "../macro/TargetUtil.ftl" as target>
+<@source.handleExceptions>
 <#if ( ext.existingInstanceMapping ) >
-    <@_assignment targetWriteAccessorName = localVarName/>
     if ( ${ext.targetBeanName}.${targetGetterName}() != null ) {
         ${ext.targetBeanName}.${targetGetterName}().clear();
-        if ( ${localVarName} != null ) {
-            ${ext.targetBeanName}.${ext.targetReadAccessorName}().<#if ext.targetType.collectionType>addAll<#else>putAll</#if>( ${localVarName} );
+        <@source.handleNullCheck>
```
agudian: That's pretty cool! 👍

sjaakd:
> Do you mean it the other way around I reckon? Rebase #903 on top of this?

Yes.. First this one.. Later the `AdderWrapper` improvements. Then adapt this accordingly. Since this one solves an issue (reproducer included), we can include this in the CR. The remainder is for later...

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
sjaakd: There are two new test cases in the issue #289 reproducer. I will move them (in due time) to collection testing. Do they answer your question? I can of course externalize them..

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.java
sjaakd: Ok..

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: :smile: .. Sounds like caramelize..

sjaakd:
> I don't yet get how that'll help us with our customizability. But it sure makes the code a bit more readable.

I have some ideas on that area. But first make stuff understandable for ourselves again. If that works, I'm confident I can make the next move. Step by step..

sjaakd:
> @sjaakd do you think you could share an example of the generated code (and how it differs from the previous one?

The main problem was this one:

old:

```java
public class DomainDtoMapperImpl implements DomainDtoMapper {

    @Override
    public Domain update(Dto source, Domain target) {
        if ( source == null ) {
            return null;
        }

        if ( source.getStrings() != null ) { // strange double check
            Set<String> set = source.getStrings(); // strange assignment
            if ( target.getStrings() != null ) {
                target.getStrings().clear();
                if ( set != null ) {
                    target.getStrings().addAll( set );
                }
            }
            else {
                target.setStrings( new HashSet<String>( source.getStrings() ) );
            }
        }

        return target;
    }
}
```
 new:

```java
public class DomainDtoMapperImpl implements DomainDtoMapper {

    @Override
    public Domain update(Dto source, Domain target) {
        if ( source == null ) {
            return null;
        }

        if ( target.getStrings() != null ) {
            target.getStrings().clear();
            List<String> strings = source.getStrings();
            if ( strings != null ) {
                target.getStrings().addAll( strings );
            }
        }
        else {
            List<String> strings = source.getStrings();
            if ( strings != null ) {
                target.setStrings( new HashSet<String>( strings ) );
            }
        }

        return target;
    }
}
```

Furthermore, I decided that we could delegate the null value mapping strategy to the forged iterable / map mapping method. We do this for other annotation attributes as well (I remember a PR of some weeks ago).

By default (no annotation provided): we would select `RETURN_DEFAULT` for calls from update / getter methods and "RETURN_NULL" for calls from a setter method. 

On the call site, (in the generated code) this saves an NPE check and is in line with the idea that mapping methods should provide a decent null handling. Hence, if the warning if the user overrides this.

So take for instance  the reproducer of the old #289 issue.

old:

```java
public class Issue289MapperImpl implements Issue289Mapper {

    @Override
    public TargetWithoutSetter sourceToTargetWithoutSetter(Source source) {
        if ( source == null ) {
            return null;
        }

        TargetWithoutSetter targetWithoutSetter = new TargetWithoutSetter();

        if ( targetWithoutSetter.getCollection() != null ) {
            Collection<TargetElement> targetCollection = sourceElementCollectionToTargetElementCollection( source.getCollection() );
            if ( targetCollection != null ) {
                targetWithoutSetter.getCollection().addAll( targetCollection );
            }
        }

        return targetWithoutSetter;
    }

    @Override
    public void sourceToTargetWithoutSetter(Source source, TargetWithoutSetter target) {
        if ( source == null ) {
            return;
        }

            target.getCollection().clear();
            Collection<TargetElement> targetCollection = sourceElementCollectionToTargetElementCollection( source.getCollection() );
            if ( targetCollection != null ) { // check
                target.getCollection().addAll( targetCollection );
            }
        }
    }

    @Override
    public TargetWithSetter sourceToTargetWithSetter(Source source) {
        if ( source == null ) {
            return null;
        }

        TargetWithSetter targetWithSetter = new TargetWithSetter();

        targetWithSetter.setCollection( sourceElementCollectionToTargetElementCollection( source.getCollection() ) ); 

        return targetWithSetter;
    }

    @Override
    public TargetElement sourceElementToTargetElement(SourceElement source) {
        if ( source == null ) {
            return null;
        }

        TargetElement targetElement___ = new TargetElement();

        targetElement___.setValue( source.getValue() );

        return targetElement___;
    }

    protected Collection<TargetElement> sourceElementCollectionToTargetElementCollection(Collection<SourceElement> collection) {
        if ( collection == null ) {
            return null; // null is returned always
        }

        Collection<TargetElement> collection_ = new ArrayList<TargetElement>();
        for ( SourceElement sourceElement : collection ) {
            collection_.add( sourceElementToTargetElement( sourceElement ) );
        }

        return collection_;
    }
}
```

new:

``` java
public class Issue289MapperImpl implements Issue289Mapper {

    @Override
    public TargetWithoutSetter sourceToTargetWithoutSetter(Source source) {
        if ( source == null ) {
            return null;
        }

        TargetWithoutSetter targetWithoutSetter = new TargetWithoutSetter();

        if ( targetWithoutSetter.getCollection() != null ) {
            Collection<SourceElement> collection = source.getCollection();
            targetWithoutSetter.getCollection().addAll( sourceElementCollectionToTargetElementCollection( collection ) );
        }

        return targetWithoutSetter;
    }

    @Override
    public void sourceToTargetWithoutSetter(Source source, TargetWithoutSetter target) {
        if ( source == null ) {
            return;
        }

            Collection<SourceElement> collection = source.getCollection();
            target.getCollection().clear();
            target.getCollection().addAll( sourceElementCollectionToTargetElementCollection( collection ) ); // MapStruct will generate different forged methods depending on whether this is an update method / target dto without setter. This forged method will return an empty array on null input.
        }
    }

    @Override
    public TargetWithSetter sourceToTargetWithSetter(Source source) {
        if ( source == null ) {
            return null;
        }

        TargetWithSetter targetWithSetter = new TargetWithSetter();

        Collection<SourceElement> collection = source.getCollection();
        targetWithSetter.setCollection( sourceElementCollectionToTargetElementCollection_( collection ) ); // MapStruct will generate different forged methods depending on whether this is an update method / target dto without setter. This forged method will return null on null input.

        return targetWithSetter;
    }

    @Override
    public TargetElement sourceElementToTargetElement(SourceElement source) {
        if ( source == null ) {
            return null;
        }

        TargetElement targetElement___ = new TargetElement();

        targetElement___.setValue( source.getValue() );

        return targetElement___;
    }

    protected Collection<TargetElement> sourceElementCollectionToTargetElementCollection(Collection<SourceElement> collection) {
        if ( collection == null ) {
            return new ArrayList<TargetElement>();
        }

        Collection<TargetElement> collection_ = new ArrayList<TargetElement>();
        for ( SourceElement sourceElement : collection ) {
            collection_.add( sourceElementToTargetElement( sourceElement ) );
        }

        return collection_;
    }

    protected Collection<TargetElement> sourceElementCollectionToTargetElementCollection_(Collection<SourceElement> collection) {
        if ( collection == null ) {
            return null;
        }

        Collection<TargetElement> collection_ = new ArrayList<TargetElement>();
        for ( SourceElement sourceElement : collection ) {
            collection_.add( sourceElementToTargetElement( sourceElement ) );
        }

        return collection_;
    }
}
```

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.java
gunnarmorling: Seems Andreas got stuck in a time loop ;)

sjaakd: a twist.. in the fabric of space..

sjaakd:
@agudian , @gunnarmorling .. I remember why I delegated the `NullValueMappingStrategy` to the forged method. Because in a way, we already supported this. See commit 5825f33. So what do we do?

@agudian .. In retrospect its easier to always add the suffix. Do you mind?

agudian:
> In retrospect its easier to always add the suffix. Do you mind?

No, sounds good!

sjaakd:
> So what do we do?

I can still avoid the situations where we end up in a warning and leave the test case intact. The warnings only applied to the getter only / update methods (so where we use the .addAll at the call site). But what if we delegate the NVMS to the forged method only in the setter scenario? That would fix this. I'll add another commit.

sjaakd:
> So what do we do?

We could also leave: `org.mapstruct.ap.test.collection.forged.CollectionMappingTest.java` as is now. As @gunnarmorling pointed out, the NVMS is intended to work on the bean mapping method result, not on individual properties. So you could also claim that the failing test cases were made with the wrong assumption.

Anyway... WDYT?

gunnarmorling:
>  So you could also claim that the failing test cases were made with the wrong assumption.

What failing tests do you mean specifically?

sjaakd:
> What failing tests do you mean specifically?

checkout commit 5825f33, modication in `org.mapstruct.ap.test.collection.forged.CollectionMappingTest.java`. Expected null (because of NVMS) is now changed to empty.

sjaakd:
@gunnarmorling as discussed yesterday. 

1. Delegation of configuration parameters such as `NullValueMappingStrategy` runs from: config -> type (mapper) -> method (bean-mapping, iterable-mapping, map-mapping).

2. Hence a generated iterable-mapping, map-mapping should take its configuration parameters from either config or type (mapper).

This would lead to including possible NPE's on the call site ( generated property mapping ) before applying the addAll in case of getter only accessor and update methods. There are 2 ways to solve this:

1. Generate a different forged method for these cases (advantage, one check less, cleaner code, better understandable templates).

2. Include a null check at call site.

We decided in the hangout to go for option 1. I stuck to that solution.

---
If there are no further objections I'll continue and merge later today.

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -28,6 +28,7 @@
 public enum Message {
 
     // CHECKSTYLE:OFF
+// CHECKSTYLE:OFF
```
gunnarmorling: Seems not needed?

sjaakd: The `CHECKSTYLE:OFF` is needed for keeping the messages on one line (violates the line length). However, I did not touch this file functionally anymore..

gunnarmorling:
@sjaakd I've played a bit with the example provided by the issue reporter (`DomainDtoMapper`). If I change the type of `Domain#strings` into `Set<Long>`, a forged iterable mapping method is generated (as expected). This one returns an empty set though if the input is `null`. Shouldn't it return `null` though as the default NVMS is `RETURN_NULL`?

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
gunnarmorling: Yes, but why have it twice? It's already there a line before :)

sjaakd: Hmm. That's a good point :smile:

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -701,6 +692,24 @@ private String getName(Type type) {
...
+
+        private NullValueMappingStrategyPrism determineForgedNullValueMappingStrategy() {
+
+            NullValueMappingStrategyPrism nvms;
+
+            if ( method.isUpdateMethod() || TargetWriteAccessorType.GETTER == targetWriteAccessorType ) {
+                nvms = NullValueMappingStrategyPrism.RETURN_DEFAULT;
+            }
+            else {
+                if ( method.getMapperConfiguration().isMapToDefaultOnMapperAndMappingConfigLevel() ) {
```
gunnarmorling: Couldn't you simply do
    nvms = method.getMapperConfiguration().getNullValueCheckStrategy() == NullValueMappingStrategyPrism.RETURN_DEFAULT;

That method `isMapToDefaultOnMapperAndMappingConfigLevel()`, seems a bit dubious to me.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.java
```diff
@@ -65,17 +63,35 @@ public Assignment getNewCollectionOrMapAssignment() {
     }
 
     @Override
+    public List<Type> getThrownTypes() {
+        List<Type> parentThrownTypes = super.getThrownTypes();
+        List<Type> result = new ArrayList<Type>( parentThrownTypes );
+        for ( Type thrownTypeToExclude : thrownTypesToExclude ) {
+            for ( Type parentThrownType : parentThrownTypes ) {
+                if ( parentThrownType.isAssignableTo( thrownTypeToExclude ) ) {
```
gunnarmorling: Is this one related to the actual issue? And shouldn't this logic be applied in a more central location than only this particular assignment?

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: It was already there.. The `NullValueCheckStrategy` is a different thing than `NullValueMappingStrategy`..

sjaakd:
> @sjaakd I've played a bit with the example provided by the issue reporter (DomainDtoMapper). If I change the type of Domain#strings into Set<Long>, a forged iterable mapping method is generated (as expected). This one returns an empty set though if the input is null. Shouldn't it return null though as the default NVMS is RETURN_NULL?

I'll have a look later. Need to go now..

sjaakd:
> @sjaakd I've played a bit with the example provided by the issue reporter (DomainDtoMapper). If I change the type of Domain#strings into Set, a forged iterable mapping method is generated (as expected).

But it is a `Set` now.. So what do you mean by 'change the type of Domain#strings into Set'?

```java
public class Domain {

    private Set<String> strings = new HashSet<String>();

    public Set<String> getStrings() {
        return strings;
    }

    public void setStrings(Set<String> strings) {
        this.strings = strings;
    }
}

-----

> This one returns an empty set though if the input is null. Shouldn't it return null though as the default NVMS is RETURN_NULL?

So, IIUC you mean this:

```java
public class DomainDtoMapperImpl implements DomainDtoMapper {

    @Override
    public Domain update(Dto source, Domain target) {
        if ( source == null ) {
            return null;
        }

        if ( target.getStrings() != null ) {
            target.getStrings().clear(); // perhaps move this inside check
            List<String> strings = source.getStrings();
            if ( strings != null ) {
                target.getStrings().addAll( strings );
            }
           // start change
           else {
               target.setString( null );
           }
           // end change
        }
        else {
            List<String> strings = source.getStrings();
            if ( strings != null ) {
                target.setStrings( new HashSet<String>( strings ) );
            }
        }

        return target;
    }
}
```

gunnarmorling:
Ah it was escaped. I meant a Set of another type, requiring a Konversion.

sjaakd:
> Ah it was escaped. I meant a Set of another type, requiring a Konversion.

I see.. I'll check this out along with the above.

sjaakd:
Ok. I checkout out your comment. 

The generated code has the same problem as of the (non conversion) original case. Right? It both returns NVMS DEFAULT. 

Perhaps we should indeed always include a null check. But it should be on the result of the forged method. For the forged method I'm still in favor of generating the proper method based on the type NVMS.

But perhaps not based anymore on the fact its an update method / getter-only property / setter-and-getter-property. This requires some more work and adding some test cases. 

--------
Apart from this, I think that a method both being a create and update method at the same time (such as in this example) complicates stuff a lot without bringing user a lot of value. Perhaps we should remove this possibility. WDYT?

sjaakd:
@gunnarmorling do we want to set the target iterable property explicitly no null for update methods?

```java
        if ( target.getStrings() != null ) {
            List<String> strings = source.getStrings();
            if ( strings != null ) {
                target.getStrings().clear(); 
                target.getStrings().addAll( strings );
            }
           else {
               target.setString( null );
           }
        }
```

gunnarmorling:
Yes, I think for an update method that'd make sense. For a normal mapping
method I'd leave the target property alone if the source property is null.

gunnarmorling:
Hey @sjaakd, how is it going with this one? Do you still think we can aim for the CR 2 this week?

sjaakd:
> Yes, I think for an update method that'd make sense. For a normal mapping
> method I'd leave the target property alone if the source property is null.

I leave this one for now.. Need some more refactoring to enable this. Separate PR after release?


> Hey @sjaakd, how is it going with this one? Do you still think we can aim for the CR 2 this week?

Troubling.. Feels like fitting a too large carpet in a too small room. If you push down one corner, the other one pops up. Simply too many combinations.. The culprit is the test case assert commented out in the last commit above. It seems to conflict with #289 test case. The main question: when to clear a target in case of an update method and getter only?

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/Assignment.java
```diff
@@ -106,4 +107,10 @@
 
 
     boolean isUpdateMethod();
+
+    /**
+     *
+     * @return return whether to return a null check (true) or not (false).
+     */
+    boolean isIncludeNullCheck();
```
sjaakd: is not used, currently.. Can be removed again.

sjaakd:
@gunnarmorling ..see my comments above.. I guess its ready for review

sjaakd:
Wrt to my question about the test case adaptation in 562f572, I did some additional investigation.

Original generated code:

```java
        if ( source.getMapValues() != null ) {
            if ( preExistTarget.getMapValues() != null ) {
                preExistTarget.getMapValues().clear();
                Map<String, String> targetMap = source.getMapValues();
                if ( targetMap != null ) {
                    preExistTarget.getMapValues().putAll( targetMap );
                }
            }
        }
```

New code:
```java
        if ( preExistTarget.getMapValues() != null ) {
            preExistTarget.getMapValues().clear();
            Map<String, String> map = source.getMapValues();
            if ( map != null ) {
                preExistTarget.getMapValues().putAll( map );
            }
        }
```

The change is due to `PropertyMapping`, line 492 and following:
```java
            // For collections and maps include a null check, when the assignment type is DIRECT.
            // for mapping methods (builtin / custom), the mapping method is responsible for the
            // null check. Typeconversions do not apply to collections and maps.
            if ( result.getType() == DIRECT ) {
                result = new NullCheckWrapper( result,  getSourcePresenceCheckerRef() );  // << this statement
            }
            else if ( result.getType() == MAPPED && result.isUpdateMethod() ) {
                result = new UpdateNullCheckWrapper( result, getSourcePresenceCheckerRef() );
            }
```

I hava no idea why that code is there. Seems that we do not want to clear the target in case of a direct mapping being null. Do guys know the original reason? Or is this plain wrong? Shall I put this check back?

Note: in the future we might want to reconsider this (when we want to support merging of maps and iterables).

sjaakd:
test this please

agudian:
I vaguely remember that we had feature requests for such things: If you only have a partial structure coming in and want to update an existing object. But I can't remember if we really made its this way.

Because now it seems to be the better alternative would be to handle that on the user side and to either create more specific, slim, bean types for partial updates, or to define a second update method with those properties explicitly excluded.

So let's leave it out and add it to the migration notes?

sjaakd:
> So let's leave it out and add it to the migration notes?

Agreed. I think this stuff has become to tricky. 

@gunnarmorling : I removed the stuff (selecting a proper forged method) again.  So if you agree I want to go ahead and merge.. I'll do some final checking myself this evening.

sjaakd:
@gunnarmorling , @agudian I want to merge this one and finish this.. Any remaining remarks?

agudian:
Nope, go ahead! 👍

sjaakd:
Thanks!

gunnarmorling:
I can't remember the reasoning behind this :/ These things have become quite complex unfortunately. I hope we can find ways to simplify them a bit after 1.1 Final. Thanks for driving this home, @sjaakd!

Issue #921, opened 2016-10-06
Duplicate variable generated in Mapper. Compilation not possible. SNAPSHOT-Version 06.10.2016


svendig:
@sjaakd If there are two nested Mappings with the same resulttype, two variables with the same name are generated.
**The generated Mapper:**
```java
public class SourceTargetMapperImpl implements SourceTargetMapper {

    @Override
    public Target map(Source source) {
        if ( source == null ) {
            return null;
        }

        Target target = new Target();

        Integer integer = sourceSub2Prop( source );
        if ( integer != null ) {
            target.setProp2( BigDecimal.valueOf( integer ) );
        }

        //Same name for variable is used!!!!
        Integer integer = sourceSub1Prop( source );
        if ( integer != null ) {
            target.setProp1( BigDecimal.valueOf( integer ) );
        }

        return target;
    }

    private Integer sourceSub2Prop(Source source) {
        ...
    }

    private Integer sourceSub1Prop(Source source) {
        ...
    }
}
```

sjaakd:
Hmmm. that's not so nice.. I'll look into this.

gunnarmorling:
Fix is merged. @svendig could you build from master yourself and confirm that it works for you now?

svendig:
@gunnarmorling I can confirm, that the variables have different names now.

Issue #931, opened 2016-10-18
Wrong type for nested properties


oliversaggau:
When I try to map a nested property from source (long) to the target (Long) and the source or the nested object is null, the target gets the value 0L instead of null. I think the type of generated method in the impl is wrong. Instead of using the type of the target (Long) which actually is nullable, MapStruct uses the type of the source property (long) for the generated method.

I attached a simple model and a unit-test which covers this issue:
[src.zip](https://github.com/mapstruct/mapstruct/files/535922/src.zip)

sjaakd:
@oliversaggau thanks.. I'll have a look.

sjaakd:
Hmm. seems you are right.. 

1. The generated method (forged method) is correct. It's intention is just to get the content of the nested property.

2. Type conversion (in this case boxing) is used to map this to the `id` with type `Long`. Which is in essence also correct.

In short.. I need to think about this a bit more. For instance you might imagine a check in the calling generated method on the presence of `nested`. However, how would you handle a `nested.nestedX.id` with `nestedX` being `null`? Would you map `null` on any of the intermediate nesting levels also to null?

oliversaggau:
I think think the forged method is absolutly correct at the moment to return `0L` (so basically the default value) for any intermediate nesting level being `null`. I would definitly expect exactly that. But rather then mapping the default value of the source property I would expect the target having the default value of the target property.

So basically I would just change the forged method to have a result type that matches the target property instead of the source property. The type conversion (in this case boxing) would then just occur in the last return statement of the forged method (where the actual value is returned). Any intermediate nesting levels being `null` should still return the default value of the result type of the forged method. Just like it is at the moment, only with the result type being slightly different.

gunnarmorling:
PR merged to master.

sjaakd:
I was thinking of a generalized approach. If the target type extends Object and the source type is primitive, use the wrapped type as return type of the forged method instead of its corresponding primitive type. WDYT?

Verstuurd vanaf mijn iPhone

PR #934, opened 2016-10-20
#921 Avoid generation of duplicate local variable name

sjaakd:
Have a more meaningful name. And add it to the set of existing variables (forgotten initially).


gunnarmorling:
@sjaakd Looks good. I'm lacking one piece of info though: in which case would we assign the result of property mappings to local variables instead of passing them to the target getter right away?

gunnarmorling:
Merged; thanks, @sjaakd!

sjaakd:
> I'm lacking one piece of info though: in which case would we assign the result of property mappings to local variables instead of passing them to the target getter right away?

Hmm. I wish I could state exactly under which conditions. :cry:. It started out with:
1. converted
2. not direct 
3. not mapped.
4. not presence checked.
5. not source type primitive

but some more issues have been written.. So some more conditions are added. You added a condition in #931.. So.. It starts to be a forrest of if-then-else stuff. High in need of refactoring. On the positive side we added test cases for all of them.

PR #935, opened 2016-10-21
#931 nested forged methods return wrapped type when target is an object

sjaakd:
Currently a forged method created to handle source nesting will return a primitive even though the mapping property target is an object. Suppose `nested1.nested2.x` with `x` being primitive, the forged methods returns primitive. Any of `nested1` or `nested2` can be null. Most logically, the result should be null of the forged method if the ultimate mapping target is not primitive.

This PR checks the mapping target. If not primitive, it changes the return type of the forged method to the corresponding wrapped type.


gunnarmorling:
Hi @sjaakd, I think the following should be happening:

a) if any of the intermediate source nodes is null, leave the target property alone (i.e. it will be null)
b) if the leave source node is not set (i.e. is 0 for a primitive int), the target property should get its value (e.g. would be (Integer)0 or so).

This is to be inline with the case of non-nested mapping where we'd copy a primitive 0 from the source into a wrapper Integer on the target side. I think the current PR does a) but not b), right?

gunnarmorling:
Ah, no, sorry, it seems it's doing what I'm suggesting.

gunnarmorling:
But there still seems to be a problem. If "id" (from the test) is a String on the target side and "nested" is null, then the forged method will return null, but there is no null check in the main method, causing `String.valueOf( null )` to be propagated into the target, i.e. "null".

sjaakd:
> But there still seems to be a problem. If "id" (from the test) is a String on the target side and "nested" is null, then the forged method will return null, but there is no null check in the main method, causing String.valueOf( null ) to be propagated into the target, i.e. "null".

Hmm.. that's not ok... This is actually a conversion. A conversion should be preceded by a null check. The logic at the spot that generates the check should be adapted as well to cover this particular case... Well spotted. 👍

sjaakd:
I'll add another commit this evening with a test and solution for this particular case..

sjaakd:
@gunnarmorling .. Fixed. Testcase added.

svendig:
@sjaakd @gunnarmorling This change should be mentioned in the migration-notes because own mappings-methods, that use primitives as input are potentially no longer considered.

Here is an example, where "sub.prop" is a primitive boolean:
```java
@Mapper
public abstract class SourceTargetMapper {
	
	@Mapping(target ="prop1", source="sub1.prop", qualifiedByName={"b2YN"})
	public abstract Target map(Source source); 
	
	@Named("b2YN")
	String asYNString(boolean bool) {
		return bool?"Y":"N"; 
	}
}
```
In Beta2 the method `asYNString` is used:
```java
@Override
        if ( source == null ) {
            return null;
        }

        Target target = new Target();

        target.setProp1( asYNString( sourceSub1Prop( source ) ) );

        return target;
    }
```
In the current Snapshot it is not:
```java
 @Override
        if ( source == null ) {
            return null;
        }

        Target target = new Target();

        Boolean sub1 = sourceSub1Prop( source );
        if ( sub1 != null ) {
            //asYNString is no longer used, because the input type does not match Boolean!
            target.setProp1( String.valueOf( sub1 ) );
        }

        return target;
    }
```

For migration, the primitive input type of the method must be changed to the wrapper type.

sjaakd:
@svendig .. good point... although I'm a bit surprised in your particular scenario. MapStruct supports 2 step mapping / conversions. So, since there's only `String asYNString(boolean bool)` being called, I would have expected that MapStruct would have carried out a type conversion on for: `Boolean` to its wrapping type (`boolean`).

PR #937, opened 2016-10-25
#60 Automapping

navpil:
Mapping methods are auto-generated for some obvious mappings between beans that have the same properties.
New tests were added. Several older were updated, because with new functionality mapping was generated when tests were expecting it to fail.
There is still one failing test: ConversionTest.shouldFailOnNonMatchingWildCards() for eclipse only.
Also for some reason integration tests fail on my new test classes with compilation errors.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethodHistory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethodHistory.java
@@ -1,0 +1,60 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model.source;
+
+import org.mapstruct.ap.internal.model.common.Type;
+
+public class ForgedMethodHistory {
+
+    private final ForgedMethodHistory prevHistory;
+    private final String sourceElement;
+    private final String targetPropertyName;
+    private final Type targetType;
+    private final Type sourceType;
+
+    public ForgedMethodHistory(ForgedMethodHistory history, String sourceElement, String targetPropertyName,
+                               Type sourceType, Type targetType) {
+        prevHistory = history;
+        this.sourceElement = sourceElement;
+        this.targetPropertyName = targetPropertyName;
+        this.sourceType = sourceType;
+        this.targetType = targetType;
+    }
+
+    public String getSourceElement() {
+        return sourceElement;
+    }
+
+    public String getTargetPropertyName() {
+        return targetPropertyName;
+    }
+
+    public Type getTargetType() {
+        return targetType;
+    }
+
+    public Type getSourceType() {
+        return sourceType;
+    }
+
+    public ForgedMethodHistory getPrevHistory() {
+        return prevHistory;
+    }
+
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethodHistory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethodHistory.java
@@ -1,10 +1,10 @@
 /**
- *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  Copyright 2012-2016 Dmytro Polovinkin
  *  and/or other contributors as indicated by the @authors tag. See the
  *  copyright.txt file in the distribution for a full listing of all
  *  contributors.
  *
- *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  Licensed under the Apache License, Version q2.0 (the "License");
  *  you may not use this file except in compliance with the License.
  *  You may obtain a copy of the License at
  *
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethodHistory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethodHistory.java
@@ -1,5 +1,5 @@
 /**
- *  Copyright 2012-2016 Dmytro Polovinkin (http://github.com/navpil)
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
  *  and/or other contributors as indicated by the @authors tag. See the
  *  copyright.txt file in the distribution for a full listing of all
  *  contributors.
@@ -20,6 +20,11 @@ package org.mapstruct.ap.internal.model.source;
 
 import org.mapstruct.ap.internal.model.common.Type;
 
+/**
+ * Keeps the context where the ForgedMethod is generated, especially handy with nested forged methods
+ *
+ * @author Dmytro Polovinkin
+ */
 public class ForgedMethodHistory {
 
     private final ForgedMethodHistory prevHistory;
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethodHistory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethodHistory.java
@@ -1,10 +1,10 @@
 /**
- *  Copyright 2012-2016 Dmytro Polovinkin
+ *  Copyright 2012-2016 Dmytro Polovinkin (http://github.com/navpil)
  *  and/or other contributors as indicated by the @authors tag. See the
  *  copyright.txt file in the distribution for a full listing of all
  *  contributors.
  *
- *  Licensed under the Apache License, Version q2.0 (the "License");
+ *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
  *  You may obtain a copy of the License at
  *
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -46,7 +46,7 @@
         <org.apache.maven.plugins.surefire.version>2.19.1</org.apache.maven.plugins.surefire.version>
         <org.apache.maven.plugins.javadoc.version>2.10.3</org.apache.maven.plugins.javadoc.version>
         <org.springframework.version>4.0.3.RELEASE</org.springframework.version>
-        <org.eclipse.tycho.compiler-jdt.version>0.23.1</org.eclipse.tycho.compiler-jdt.version>
+        <org.eclipse.tycho.compiler-jdt.version>0.26.0</org.eclipse.tycho.compiler-jdt.version>
         <com.puppycrawl.tools.checkstyle.version>7.2</com.puppycrawl.tools.checkstyle.version>
         <add.release.arguments />
         <forkCount>1</forkCount>
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -47,7 +47,10 @@ import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.dependency.GraphAnalyzer;
 import org.mapstruct.ap.internal.model.dependency.GraphAnalyzer.GraphAnalyzerBuilder;
+import org.mapstruct.ap.internal.model.source.ForgedMethod;
+import org.mapstruct.ap.internal.model.source.ForgedMethodHistory;
 import org.mapstruct.ap.internal.model.source.Mapping;
+import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.model.source.PropertyEntry;
 import org.mapstruct.ap.internal.model.source.SelectionParameters;
 import org.mapstruct.ap.internal.model.source.SourceMethod;
@@ -78,11 +81,12 @@ public class BeanMappingMethod extends MappingMethod {
     private final boolean mapNullToDefault;
     private final Type resultType;
     private final NestedTargetObjects nestedTargetObjects;
+    private final boolean overridden;
 
     public static class Builder {
 
         private MappingBuilderContext ctx;
-        private SourceMethod method;
+        private Method method;
         private Map<String, Accessor> unprocessedTargetProperties;
         private Set<String> targetProperties;
         private final List<PropertyMapping> propertyMappings = new ArrayList<PropertyMapping>();
@@ -91,6 +95,8 @@ public class BeanMappingMethod extends MappingMethod {
         private SelectionParameters selectionParameters;
         private final Set<String> existingVariableNames = new HashSet<String>();
         private NestedTargetObjects nestedTargetObjects;
+        private Map<String, List<Mapping>> methodMappings;
+        private SingleMappingByTargetPropertyNameFunction singleMapping;
 
         public Builder mappingContext(MappingBuilderContext mappingContext) {
             this.ctx = mappingContext;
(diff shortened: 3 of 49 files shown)


agudian:
Really cool feature, @navpil! We'll need some time to digest it and get back to you in more detail. 

Regarding the failing integration tests: the classes that include the acutal test methods need to be named `*Test.java`, as those classes will be excluded when compiling the test mappers for the integration test.

navpil:
Thanks, I have renamed tests and integration tests passed.

agudian:
I had a first look in the meantime and played around a bit with it. It's really cool!

I do get the same test failure with the Eclipse compiler as you describe above, but still need to dive in deeper to see what's wrong there.

Still, this will probably be my favorite feature in release 1.2 👍

sjaakd:
So, guys how do we take this one step further. @navpil : is there a possibility that rebase your pull request to the latest MapStruct? We can of course help out as well.

agudian:
@sjaakd my first step would be to check the failure in `ConversionTest` and why the changes in the expected error messages where necessary. From far away it looks like something with the resolution of type variables got lost. I started looking at it two times already but had to interrupt after ten minutes each time 🙄.
Anyway, we can do that with or without the rebase - although that would be nice before going into the details of the PR.

filiphr:
Regarding the changed tests:
* *ErroneousCollectionMappingTest* - Change is OK. With the automapping kicking in we try to map the values from `AttributedString` to `String` and not the maps themselves. 
* *UpdateMethodsTest* - Change is also OK. With the automapping kicking in the `DepartmentDto` can be automatically mapped to the `DepartmentEntity` as they have the same field names. The new `UnmappableDepartmentDto` has `workers` instead of `employees` and therefore it cannot be automapped. Maybe we need to change something in the existing `DepartmentDto` so it does not get picked up by the automapping instead of adding a new Class.
* *ConversionTest* - Change is also OK. Due to the automapping kicking in we can map `WildCardSuperWrapper<Integer>` to `WildCardSuperWrapper<String>` via our conversions. However, something funny is happening with the eclipse compiler when we have unmappable types like `String` -> `TypeA`. This is what we have:

```java
public class WildCardSuperWrapper<T> {

    private T wrapped;

    public WildCardSuperWrapper(T wrapped) {
        this.wrapped = wrapped;
    }

    public T getWrapped() {
        return wrapped;
    }

    public void setWrapped(T wrapped) {
        this.wrapped = wrapped;
    }
}

public class ErroneousSource6 {

    private WildCardSuperWrapper<String> foo;

    public WildCardSuperWrapper<String> getFoo() {
        return foo;
    }

    public void setFoo(WildCardSuperWrapper<String> foo) {
        this.foo = foo;
    }
}

public class ErroneousTarget6 {

    private WildCardSuperWrapper<Integer> foo;

    public WildCardSuperWrapper<Integer> getFoo() {
        return foo;
    }

    public void setFoo(WildCardSuperWrapper<Integer> foo) {
        this.foo = foo;
    }
}
```

Via the automapping the following method is forged:
```java
protected WildCardSuperWrapper<Integer> stringWildCardSuperWrapperToIntegerWildCardSuperWrapper(WildCardSuperWrapper<String> wildCardSuperWrapper);
```
With the JDK compiler the `targetType` of the `wrapper` is correctly determined as `TypeA`. However, with the eclipse compiler the determined `targetType` is `T` and not `TypeA`. When the `targetType` is generic we do some other error:
* The `StringToXmlGregorianCalendar` builtIn method matches and we create the conversion

btw. If a target does not have a default public constructor, do we need to fail?

I think that the problem with the eclipse compiler is some bug, I tried with version `0.26.0` of the `tycho-compiler-jdt` and all the tests pass successfully. More specifically I think that the fix is in the new version of `org.eclipse.jdt.compiler.apt` that is in `tycho-compiler-jdt`.

Sorry for the lengthy comment, I didn't expect that it will be so long 😄.

filiphr:
Some other questions which I have before doing a deeper review:
* Do we want to make the automapping configurable (on or off)?
* How should we report the errors if something is not mappable? Do we need to display the path to the unmappable property or just display it like we do now?

navpil:
I tried to rebase/merge as good as I could and it seems that at least tests are not broken. I was not sure about the PropertyMapping.java where I (re)introduced the methods getSourceType() and getSourceElement(), which I use heavily when creating ForgeMethods. Is there a better alternative to those, or it's ok to have them there? (In particular PropertyMapping.java line 723 and 745).

Also it seems that we have too many constructors for MappingMethod.java

navpil:
@filiphr Currently the automapping is not confgurable, so it's always on. It's not that complicated to add This was discussed here: https://groups.google.com/forum/#!topic/mapstruct-users/8MiKbs-W5_c

When something is unmappable I tried to keep the error which user would get without automapping. Theoretically I could show the full path to the unmappable property in case there are several levels of automap, but it's not the case now.

sjaakd:
Hi,

These method should be available in SourceRHS (instance variable in property mapping.

Verstuurd vanaf mijn iPhone

filiphr:
@navpil thanks for the link. I didn't know that it was already discussed before, sorry for my misunderstanding.

@agudian, @sjaakd How should we proceed with the review? I did a review an the changed test and the failing eclipse test. Should we update the compiler in a separate PR or we do it as part of this one?

sjaakd:
> @agudian, @sjaakd How should we proceed with the review?

@filiphr I'm not in this weekend, so my first chance would be early next week.. I'm working on some tough refactoring of the `SetterWrapper` which I would like to continu on this evening.

navpil:
@sjaakd I've deleted the methods getSourceType() and getSourceElement() from the PropertyMapping.java.
I had to introduce the parameter for the getForgedMethodHistory(), which is not a problem by itself, but I've noticed that the sourceRHS/source which we pass into the forge*Mapping() methods is always the same as the "rightHandSide" field. If we used the field instead of passing the field as a parameter then the parameter list would be smaller and the getForgedMethodHistory() would also use the "rightHandSide" field instead of getting the "source" as a parameter.
I didn't feel brave enough to do this refactoring, but does it make sense to do it?

sjaakd:
It does! I actually want to move SourceRefetence out of property mapping and use SourceRHS in the property mapping builder. I'm first busy refactoring SetterWrapper and template. So hopefully later coming week!

Verstuurd vanaf mijn iPhone

agudian:
> I think that the problem with the eclipse compiler is some bug, I tried with version 0.26.0 of the tycho-compiler-jdt and all the tests pass successfully. More specifically I think that the fix is in the new version of org.eclipse.jdt.compiler.apt that is in tycho-compiler-jdt.

Ah, okay... 0.26.0 is based on the current Eclipse Neon release and I'd say it's okay to verify against the latest Eclipse version only. Especially if the older version has some problems here...

> When something is unmappable I tried to keep the error which user would get without automapping. Theoretically I could show the full path to the unmappable property in case there are several levels of automap, but it's not the case now.

I think that would be very cool to have, especially for people with deeper structures. Perhaps you can reuse `Message.PROPERTYMAPPING_MAPPING_NOT_FOUND` for this purpose and use dotted notations for the source and target property path?

sjaakd:
Suppose you want you want to use that error message to make a nested mapping... how would that work together with the auto mapping?

Verstuurd vanaf mijn iPhone

agudian:
If a mapping at some nested level doesn't work out of the box, you would have to define the mapping method for those types explicitly, just as it is the case now. @sjaakd, is that what you mean?

sjaakd:
Exactly! Thanks 

Verstuurd vanaf mijn iPhone

navpil:
Now I'm at the point where I can start trying to implement the dotted notation for error messages.

Let's say we have a Person who has a Car which has a Wheel. mapPerson() method is there, so mapCar() is forged, but the mapWheel() will fail because Wheels can't map.
Now we get the message like
Can't map property "org.mapstruct.test.Car car" to "org.mapstruct.test.CarDto car". Consider to declare/implement a mapping method: "org.mapstruct.test.CarDto map(org.mapstruct.test.Car value)".

How do we want the dotted notation to look like?
Can't map property "org.mapstruct.test.**Wheel car.wheel**" to "org.mapstruct.test.**WheelDto car.wheel**". Consider to declare/implement a mapping method: "org.mapstruct.test.CarDto map(org.mapstruct.test.Car value)".

agudian:
Exactly! Just the hint to the method to declare/implement:

Can't map property "org.mapstruct.test.Wheel car.wheel" to "org.mapstruct.test.WheelDto car.wheel". Consider to declare/implement a mapping method: "**org.mapstruct.test.WheelDto** map(**org.mapstruct.test.Wheel** value)".

But I guess that's what you meant, right?

navpil:
I had a problem with forged methods inside forged iterables. So if we have a list of wheels, then the forged iterable mapping will be generated, but it will create a forged method for actual mapping between wheels. And there it fell apart, because there was no targetPropertyName and no reasonable source name. 
I decided to simply strip away the package name and lowercase the classname to get those. Is this ok?

One more thing - I'm now parsing the sourceElement which I get from sourceRHS.getSourceErrorMessagePart() in order to create dotted notation for sources. It can come in several formats (one of them is "property "foo.Bar bar""). Is this ok to do, or I'd better do something else?

agudian:
> So if we have a list of wheels, then the forged iterable mapping will be generated, but it will create a forged method for actual mapping between wheels. And there it fell apart, because there was no targetPropertyName and no reasonable source name.

So if you had, for example a structure where you could call `person.getCar().getWheels().get(0).getPressure()` and a mapping method for `PersonDto` but you can't forge a method to map `Pressure` to `PressureDto`. You now implemented it as `car.wheels.wheelDto.pressure`? I think the path `car.wheels.pressure` should be just fine (so just leave out the notion of hopping from the list/iterable to the element, AFAIK there is no property path expression for that defined, right?). Or `car.wheels[].pressure`? Hmm, don't know if that would help...

> One more thing - I'm now parsing the sourceElement which I get from sourceRHS.getSourceErrorMessagePart() in order to create dotted notation for sources. It can come in several formats (one of them is "property "foo.Bar bar""). Is this ok to do, or I'd better do something else?

Hmm, can't you keep track of the path (on target and source side) with that History object?

navpil:
Let's have `Person `with `Car `with `List<Wheel>` where wheels don't map to each other. We have `PersonDto map(Person entity)` method, other are forged.
Should the error message look like this?
`Can't map List<org.mapstruct.test.Wheel> car.wheels" to List<org.mapstruct.test.WheelDto> car.wheels. Consider to declare/implement a mapping method: "org.mapstruct.test.WheelDto map(org.mapstruct.test.Wheel value) `
Note that the source and target types are different in the "Can't map" part and the "Consider to" part.

I guess we can reuse the same pattern if the keys/values of the map do not map to each other:
`Can't map Map<Position, org.mapstruct.test.Wheel> car.wheelsMap" to Map<Position, org.mapstruct.test.WheelDto> car.wheelsMap. Consider to declare/implement a mapping method: "org.mapstruct.test.WheelDto map(org.mapstruct.test.Wheel value) `

Actually I do keep track of the source and target property names in the History object, but before I didn't need the exact `sourceName`, because what the `Message.PROPERTYMAPPING_MAPPING_NOT_FOUND` requires for a source description is a single string, which conforms to what the `sourceRHS.getSourceErrorMessagePart()` returns. Now I'm tracking the `sourceName `separately. This means, however, that `Message.PROPERTYMAPPING_MAPPING_NOT_FOUND` will not receive what the `sourceRHS.getSourceErrorMessagePart()` would give it. That's logical - because that method didn't return the dotted notation anyway, but is it still ok?

sjaakd:
> Can't map List<org.mapstruct.test.Wheel> car.wheels" to List<org.mapstruct.test.WheelDto> car.wheels. Consider to declare/implement a mapping method: "org.mapstruct.test.WheelDto map(org.mapstruct.test.Wheel value)

Does it give the hint to implement a mapping for the element in the current MapStruct code base? (Sorry, I forgot to be honest).

> Now I'm tracking the sourceNameseparately. This means, however, that Message.PROPERTYMAPPING_MAPPING_NOT_FOUND will not receive what the sourceRHS.getSourceErrorMessagePart() would give it.

That's not (yet) in your PR yet, right? Or did I miss this somehow?

> That's logical - because that method didn't return the dotted notation anyway, but is it still ok?

I think what in the end is included `sourceReference.toString()`. So:
1. It will hint to the parameter, if its taken directly.
2. a property if its a non nested property
3. a nested property when nested.

Will that still work than? Or.. perhaps I don't understand your remark.

navpil:
@sjaakd Yes, mapstruct does give a hint for implementing the method. 
The dotted notation I'm talking about is still not in PR, so the `Message.PROPERTYMAPPING_MAPPING_NOT_FOUND` still gets the `sourceRHS.getSourceErrorMessagePart()`

I will look into the `sourceReference.toString()` and check whether it returns correct results for forged methods.

The trouble is that `Message.PROPERTYMAPPING_MAPPING_NOT_FOUND` requires a single string for a first parameter (the source part), which is currently the same as `sourceRHS.getSourceErrorMessagePart()`. However the latter may return `"collection element"` for example without any hint of a type or a name. Therefore I decided to use my own rules to determine what the string passed into the first parameter for `Message.PROPERTYMAPPING_MAPPING_NOT_FOUND` should be. What I currently pass there (not commited yet) is `history.getSourceType() + " " + history.getSourceDottedName()`.

sjaakd:
> The dotted notation I'm talking about is still not in PR, so the Message.PROPERTYMAPPING_MAPPING_NOT_FOUND still gets the sourceRHS.getSourceErrorMessagePart()

This PR is growing a bit too large :smile: 

@agudian, @filiphr, @gunnarmorling : shall we include the solution (above) first and take additional (like the commits) as additional Issue / PR? Then I can continue with my stuff without blocking each other in  merges.

agudian:
> Therefore I decided to use my own rules to determine what the string passed into the first parameter for Message.PROPERTYMAPPING_MAPPING_NOT_FOUND should be. What I currently pass there (not commited yet) is history.getSourceType() + " " + history.getSourceDottedName().

Sounds totally reasonable to me. Would you add that to this commit here?

I agree with @sjaakd that we should try to wrap this up quickly now and deal with any missing spots in seperate issues and PRs after that. It would of course be great if you could continue to help with that, @navpil 😄.
There are also some checkstyle issues (I think) that would now break the build... please check that before committing. Jenkins should also list that, if I write "ok to test" here.

agudian:
Oh and please set the version of the tycho compiler to `0.26.0` in `parent/pom.xml` to fix that remaining test?

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethodHistory.java
```diff
@@ -0,0 +1,60 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model.source;
+
+import org.mapstruct.ap.internal.model.common.Type;
+
+public class ForgedMethodHistory {
```
filiphr: You should add your name to the class and take credit for it 😉

sjaakd:
@navpil : are you able to find time to fix @agudian / @filiphr  last 2 comments and go ahead with your own proposed change for the `Message.PROPERTYMAPPING_MAPPING_NOT_FOUND`? If so, and when ready, let me know and I'll merge it 👍 .

navpil:
I'm looking at it now:
1. Set the tycho version
2. Change the author ofForgedMethodHistory

I want to change the formatting in some places, so codecheck passes.

After that I can continue with dotted notation. The problems is that about 20 tests fail, because of the [] and {} symbols I introduced for iterables and maps. 

I'm doing the dotted notation in a separate branch so we can included it here, or add it later.

agudian:
Okay, then let's get this one merged by tonight and handle the dotted error message in a separate PR afterwards.
@navpil, could you check the missing license headers reported by the build?

navpil:
I don't know why doesn't it like my license headers. I copy-pasted them from some other random file and changed the name. Nothing fails locally.

sjaakd:
Try: mvn licence format on the command line 😉

Verstuurd vanaf mijn iPhone

agudian:
The license header needs to remain unchanged, with Gunnars name in it. You place your name in the `@author` tag and we'll also add you to the copyright.txt.

Locally, mvn clean install should tellyou everything that the Jenkins finds as well.

sjaakd:
Hey @navpil . Thanks.. If it turns green I'm going to merge it.. :smile:

sjaakd:
Shall I have look? Its check style..

navpil:
@sjaakd I'm looking into it now. I didn't think I touched that class, but apparently I did.

sjaakd:
> @sjaakd I'm looking into it now. I didn't think I touched that class, but apparently I did.

Sure.. No problem.. I'll wait. I saw there's also some commented out code in `MapMappingMethod.java`.. Can you remove that as well?

sjaakd:
BTW: did you know you could run: `mvn checkstyle:checkstyle`.. Saves you a complete build :).

navpil:
@sjaakd I decided to run the complete build (except processor tests) so I would be sure that nothing else was broken. I also have a fresh system here, so it takes a bit longer than usual.

sjaakd:
> @sjaakd I decided to run the complete build (except processor tests) so I would be sure that nothing else was broken. I also have a fresh system here, so it takes a bit longer than usual.

No problem whatsoever. I'm looking into your stuff right now.  Thanks :).

sjaakd:
@navpil .. We want to add you to the copyright.txt. Can you disclose your real name ? :)

navpil:
@sjaakd My name is Dmytro Polovinkin

agudian:
Ah, just like in the `@author` tag 👍... we could say now that it was a trick question, uhm... 🙄 

😁

sjaakd:
wel Dmytro, welcome on board :). Nice work!!!

agudian:
🎆 🎉 

Thanks a lot, Dmytro!

agudian:
Oh, I'd say we leave the issue itself open until we figured out and merged the dotted notation for the error messages, right?

sjaakd:
Lets make a follow up one for that..

agudian:
We can add those nice checkmarks to #60 to track the missing pieces. Documentation update, for example. That all belongs to #60, so we don't really need extra issues for it, right?

sjaakd:
> We can add those nice checkmarks to #60 to track the missing pieces. Documentation update, for example. That all belongs to #60, so we don't really need extra issues for it, right?

That's actually a good idea.. Forgot about those entirely. Just opened #1001 ..  Sorry. Next time?

Issue #892, opened 2016-09-12
String.valueOf called for int to int mapping


piercena:
We have a source and target bean. Both has a parameter of type int on them. In the mapper impl that is generated using "mvn clean compile" the generated code is target.setType(String.valueOf(int)); even though both the source and target has an int type. In the target bean though there are two methods that set the same parameter on the target, one that takes in an int and one that takes in a string. The 1.0.0Final documentation says that String.valueOf(int) is only called implicitly when the source is of type int and the target is of type String. We believe this is a bug. Our current work around was to rename the method that took in the type String parameter and this fixed the issue.

sjaakd:
@piercena could you provide a reproducer?

piercena:
Since this was added for the 1.1.0.Final milestone does that mean this was found or do you still need the reproducer? I apologize for not providing that and completely forgetting to look back at this since I posted the issue.

gunnarmorling:
Yes, please do provide a reproducer. That will help very much. Thanks!

gunnarmorling:
Done.

PR #942, opened 2016-10-28
#892 If multiple setters exist, don't override the setter if it doesn't match the property getters type

agudian:


gunnarmorling:
@agudian So what is the strategy for handling this sort of conflict; if there is any target property with an assignment-compatible type, take that one? What happens in these cases:

* There are multiple assignment compatible candidates, e.g. `setFoo(int)` and `setFoo(long)`. Would it be ensured that the former would be used for mapping an `int` property? I reckon that yes because it would take the first assignment-compatible one and the call would be generated; the correct one will be linked by the compiler as per the Java method resolution rules.
* There are multiple non compatible candidates, `setFoo(String)` and `setFoo(BigInteger)`. I don't think we can choose one reasonably, rather an ambiguity error should be raised (though we don't provide a way to resolve the case currently). That would have to be fixed, too (could be done separately, e.g. by adding `@PropertyMapping#resultType()`. For now I'd be fine with an error clearly indicating the situation (it's really a special case).

agudian:
I think the "right" solution would be to promote all possible variants and let the method selection mechanism do its magic. That would cover the points you mentioned and allow for some error messages. But that would require some refactoring and it would make several methods more complicated.
As we're dealing with a strange edge case here, I went for a simple approach that only prevents returning a "worse" setter if there was one found previously already.

* If there is only one setter for the property name, it will be returned regardless of its argument type (as it has been before). 
* If there is more than one setter and there is a getter available to determine the type, the last setter with an assignment compatible argument type is returned. 
* If there is more than one setter and no getter is available, nothing changes compared to how it was before: the last setter is returned.

So it's not as smart as it could be, but it should suffice to prevent choosing the wrong setter... 😬

gunnarmorling:
I'm not sure I agree on automatically selecting a target setter if there are several ones which all need conversion. I don't think we can make any reasonable decision in this case. But we can leave that for another time, the change at hand already is an improvement.

gunnarmorling:
Rebased and applied. Thx, @agudian!

Issue #900, opened 2016-09-17
Fix integration-test execution with jigsawed JDK 9 in toolchain


agudian:
With the Jigsaw-enabled JDK 9 builds, we need to add `-addmods java.annotations.common` to the compiler args, as otherwise some of the annotations we add to the classes won't be available (e.g. `@Generated`).
We might also consider detecting if the annotation is available in the classpath and leave it out / comment it out in the generated classes to make it easier for our users.

sjaakd:
oops. wrong reference.. sorry

gunnarmorling:
Done.

PR #943, opened 2016-10-28
#900 Use "@Generated" only when it's available on the classpath.

agudian:
This allows the usage of MapStruct with JDK 9 Jigsaw without having to add `-addmods java.annotations.common` to javac if there is no other reason to do that.

Brushing up the integration tests to actually run with Jigsaw JDK 9 again is a bit icky and I'll do that later (requires `-addmods` for the JAXB tests, etc...).


gunnarmorling:
Nice one, thanks! I've added some words to the docs, too.

agudian:
> I've added some words to the docs, too.

Great, thanks 👍

PR #944, opened 2016-10-29
#787 Giving some love to the readme.md

gunnarmorling:


readme.md
```diff
@@ -3,37 +3,58 @@
...
+    CarMapper INSTANCE = Mappers.getMapper( CarMapper.class );
 
-Compared to writing mapping code from hand, MapStruct saves time by generating code which is tedious and error-prone to write. Following a convention over configuration approach, MapStruct uses sensible defaults but steps out of your way when it comes to configuring or implementing special behavior.
+    @Mapping(source = "numberOfSeats", target = "seatCount")
+    CarDto carToCarDto(Car car);
+}
+```
 
-Compared to dynamic mapping frameworks, MapStruct offers the following advantages:
+At compile time MapStruct will generate an implementation of this interface. The generated implementation uses plain Java method invocations for mapping between source and target objects, i.e. there is no reflection involved. By default, properties are mapped if they have the same name in source and target, but this and many other aspects can be controlled using `@Mapping` and a handful of other annotations.
```
sjaakd: Java 6 compatible method invocations

readme.md
```diff
@@ -3,37 +3,58 @@
...
+
+MapStruct saves you from writing mapping code by hand which is a tedious and error-prone task. The generator comes with sensible defaults and many built-in type conversions, but it steps out of your way when it comes to configuring or implementing special behavior.
+
+Compared to mapping frameworks working at runtime MapStruct offers the following advantages:
 
 * Fast execution by using plain method invocations instead of reflection
 * Compile-time type safety: Only objects and attributes mapping to each other can be mapped, no accidental mapping of an order entity into a customer DTO etc.
 * Self-contained code, no runtime dependencies
-* Clear error-reports at build time, if entities or attributes can't be mapped
+* Clear error-reports at build time if mappings are incorrect and incomplete
```
sjaakd: sub-bullits. 
1. incomplete (not all target properties are mapped).
2. incorrect (cannot find a proper mapping method / conversion).

I've come to really appreciate 1. as one of our strongest assets.

readme.md
```diff
@@ -3,37 +3,58 @@
...
-To learn more about MapStruct in two minutes, refer to the [project homepage](http://mapstruct.org). The [reference documentation](http://mapstruct.org/documentation) covers all provided functionality in detail. If you need help, come and join the [mapstruct-users](https://groups.google.com/forum/?hl=en#!forum/mapstruct-users) group.
-
 ## Requirements
 
 MapStruct requires Java 1.6 or later.
 
 ## Using MapStruct
 
-MapStruct is a Java annotation processor based on [JSR 269](https://jcp.org/en/jsr/detail?id=269) and as such can be used within command line builds (javac, Ant, Maven etc.) as well as from within your IDE.
+MapStruct works in command line builds (plain javac, via Maven, Gradle, Ant etc.) and IDEs.
```
sjaakd: IDE builds.

readme.md
agudian: Why should we differentiate between "Java 6 compatible method invocations" and "Java method invocations"? 🤔

gunnarmorling: It's also not quite clear to me. But also keep in mind that the readme is no reference documentation, it should be short and concise and provide the most important information, for all the details there is the reference guide.

gunnarmorling:
Thanks for reviewing. I've applied the remark with the sub items.

PR #946, opened 2016-11-01
Updates mapstruct-reference-guide: Revives caption

dgruntz:
Caption to example 17 got lost (due to a missing line). Adds the missing line in front of
    .Conversion from BigDecimal to String
so that this caption is printed as caption again.


gunnarmorling:
Merged; thanks a lot @dgruntz!

PR #952, opened 2016-11-04
Added BOM to make it easier for dependancy management

garethahealy:
Most frameworks now provide a BOM, which i think this is a good convention to follow.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+ Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ and/or other contributors as indicated by the @authors tag. See the
+ copyright.txt file in the distribution for a full listing of all
+ contributors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


agudian:
Hi Garet, thanks for the suggestion and the PR!

Do you think this would actually be used? BOMs are great, but in our case, with only one dependency (either `mapstruct` or `mapstruct-jdk8`) and no transitive deps, there isn't really anything to gain (except a couple of more lines in your pom). So I'd rather not add it to the README to keep it to the basics there.
Too bad we can't provide the compiler config with a BOM 😕 or at least manage the `mapstruct-processor` version for the usage in the compiler plugin config.

@gunnarmorling, @sjaakd, any thoughts?

gunnarmorling:
Hi, yes, I appreciate the contribution but I was about to answer along the same lines essentially.

I'm a strong proponent of BOMs, but I don't think it's worth it in this case. People only need two dependencies: _mapstruct_ (or _mapstruct-jdk8_) and _mapstruct-processor_, where the latter doesn't benefit from the BOM if it's used as a plug-in dependency AFAICS. So I'd rather not add a BOM and keep things simpler.

gunnarmorling:
I'm going to close this one as "Won't fix" for the reasons pointed out above. Nevertheless many thanks for your efforts to create this proposal!

Issue #955, opened 2016-11-12
Missing Type import for Map<k,v> when providing custom mapper


simontrudeau:
When declaring a mapper like so using MapStruct version 1.1.0.CR2:

Mapper:
```java
@Mapper(uses = {personMapper.class})
public interface CarMapper {
	SuperCar mapCar(Car source, @MappingTarget SuperCar destination);
}
```

Custom Mapper
```java
@Component
public class EslUserCustomFieldMapMergeMapper {

public Map<String,Person> merge(Map<String,Person> source, @MappingTarget Map<String,Person> destination){
		source.entrySet().forEach(entry -> destination.put(entry.getKey(), entry.getValue()));
		return destination;
	}
}
```

Car containing a Map of Person
```java
public class Car {
    private Map<String, Person> persons = new HashMap<String, Person>();
...
}
```

The generate code is missing import for type `Person`. The generate code (`CarMapperImpl.java`) contains the following and is missing *import* for the type `Person`, thus it doesn't compile:
```java
...
if ( destination.getPersons() == null ) {
            destination.setPersons( new HashMap<String, Person>() );
        }
carMapper.merge( source.getPersons(), destination.getPersons() );
...
```

Is it possible to add a hint or something so that the generator knows to add the import for the type `Person`?

filiphr:
If you add `@Mapper( imports = Person.class )`, mapstruct will add the import to mapper class and it will work.

simontrudeau:
It works, thank you.

sjaakd:
But.. is still an issue :smile:

sjaakd:
Fixed.

gunnarmorling:
@sjaakd, @agudian, shouldn't this be assigned to the 1.1.0.Final release?

gunnarmorling:
Assigned that milestone. When closing an issue please make always sure a milestone is assigned. Thanks! Same for #954.

sjaakd:
> Assigned that milestone. When closing an issue please make always sure a milestone is assigned. Thanks! Same for #954.

Ok.. Forgot.

gunnarmorling:
No problem. Just was wondering whether there was any intention behind it :)

PR #957, opened 2016-11-13
#955 Import type param type when call to upd. method for coll.impl type

sjaakd:


agudian:
Looks good! Would you please commit and push this on 1.1-master first and then use a merge-commit to integrate it back to master? That way we can always easily see what 1.1 fixes were integrated back already...

PR #958, opened 2016-11-13
#955 Import type param type when call to upd. method for coll.impl type

sjaakd:


sjaakd:
@agudian .. just to be sure.. you mean: just push the `merge-pull-request` button above when builded correctly.? that would cause a merge-commit right?

Issue #824, opened 2016-07-05
Target Bean Factory


JulianH:
It would be great if the target bean could be instantiated using a factory, which could use the source bean. The idea is, that the factory could instaniate the instance based on some existing information of the source. What i have in mind is, that the entity could be instantiated by an hibernate query and the mapping method will then update the mapped fields.

The factory which will instantiate the beans, for example the entity, could look like

``` java
public class EntityFactory {
   public <T> T createEntity(@TargetType Class<T> targetClass, @MappingSource Object source) {
      // return initialized instance of T
   }
}
```

The generated mapper could look like

``` java
public class EmployeeMapperImpl implements EmployeeMapper {


   public Employee convertFromEmployeeDto(EmployeeDto employeeDto) {
      if ( employeeDto== null ) {
         return null;
      }

      Employee employee = this.entityFactory.createEntity(Employee.class, employeeDto);
      ...
   }
}
```

@gunnarmorling if you give me some hints how to implement this feature, i will try to contribute this to mapstruct.

agudian:
Note: Currently we differentiate between factory methods and (custom) mapping methods simply by checking the parameters: if the method has no parameter or only one `@TargetType` parameter, it's a factory method - otherwise it's considred a mapping method that can be invoked for property mappings.
So we would have to come up with something else to distinguish these two cases. Perhaps by using an explicit `@FactoryMethod` annotation or something like that?

remmeier:
+1 such a feature would be extremely useful

sjaakd:
> Note: Currently we differentiate between factory methods and (custom) mapping methods simply by checking the parameters: if the method has no parameter or only one @TargetType parameter, it's a factory method - otherwise it's considred a mapping method that can be invoked for property mappings.

Hmm. wouldn't that be avoid by the suggestion above: so if the mapping method has only a parameter marked with `@MappingSource ` or icm with a parameter marked with `@TargetType ` its still a factory method?

agudian:
Every parameter that is not annotated otherwise is a mapping source, so creating a new annotation `@MappingSource` could be a bit confusing: should one always use that annotation for property mappings as well? Probably not... :wink: 

[...]

PR #961, opened 2016-11-18
#824 Target Bean Factory

remmeier:
Introduced @ObjectFactory annotation to distinguish between factory and mapping methods. So far this was decided based on the lack of parameters. @ObjectFactory lets a factory method carry the same
signature as a mapping method, letting it have access to the source objects.

### DIFF
--- a/core-common/src/main/java/org/mapstruct/ObjectFactory.java
+++ b/core-common/src/main/java/org/mapstruct/ObjectFactory.java
@@ -1,0 +1,73 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * By default beans are created during the mapping process with the default constructor.
+ * This method allows to mark a method as a factory method to create beans.
+ * The type of the factory method is determined by its return type. A factory is used
+ * to create a bean if their types match.
+ * <p>
+ * The factory method can retrieve the mapping sources by specifying
+ * them as parameters, exactly like mapping and lifecycle methods do. This allows the factory
+ * method to create target beans based on source beans.
+ * The use of source parameters opens up a variety of possibilities. For example,
+ * a factory method for entities can look up an
+ * EntityManager to check whether the entity already exists and then return its
+ * managed instance.
+ *
+ * <pre>
+ * {@literal @}ApplicationScoped // CDI component model
+ * public class ReferenceMapper {
+ *
+ *     {@literal @}PersistenceContext
+ *     private EntityManager em;
+ *
+ *     {@literal @}ObjectFactory
+ *     public SomeEntity resolve(SomeDto dto) {
+ *         SomeEntity entity = em.find(SomeEntity.class, dto.getId());
+ *         return entity != null ? entity : new SomeEntity();
+ *     }
+ * }
+ * </pre>
+ *
+ * If no such parameters
+ * are provided, the use of this annotation is optional.
+ *
+ * <p>
+ * If there are two factory methods, both serving the same type, one with no parameters
+ * and one taking sources as input, then the one with the source parameters is favored.
+ * If multiple factory method take sources as input, them the one is chosen which allows a
+ * valid assignment from mapping source parameters to those factory parameters. If
+ * there are multiple such factories, an ambiguity exception is thrown.
+ * </p>
+ *
+ * @author Remo Meier
+ * @since 1.2
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.METHOD)
+public @interface ObjectFactory {
+
+}
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -1410,7 +1410,8 @@ By default, the generated code for mapping one bean type into another will call 
 
 Alternatively you can plug in custom object factories which will be invoked to obtain instances of the target type. One use case for this is JAXB which creates `ObjectFactory` classes for obtaining new instances of schema types.
 
-To make use of custom factories register them via `@Mapper#uses()` as described in <<invoking-other-mappers>>. When creating the target object of a bean mapping, MapStruct will look for a parameterless method, or a method with only one `@TargetType` parameter that returns the required target type and invoke this method instead of calling the default constructor:
+To make use of custom factories register them via `@Mapper#uses()` as described in <<invoking-other-mappers>>. When creating the target object of a bean mapping, MapStruct will look for a parameterless method, 
+a method annotated with `@ObjectFactory`, or a method with only one `@TargetType` parameter that returns the required target type and invoke this method instead of calling the default constructor:
 
 .Custom object factories
 ====
@@ -1485,6 +1486,27 @@ public class CarMapperImpl implements CarMapper {
 }
 ----
 ====
+
+In addition, annotating a factory with `@ObjectFactory` lets you gain access to the mapping sources. 
+Source objects can be added as parameters in the same way as for mapping method. The `@ObjectFactory`
+annotation is necessary to let MapStruct know that the given method is only a factory method. 
+
+.Custom object factories with `@ObjectFactory`
+
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+public class DtoFactory {
+
+     @ObjectFactory
+     public CarDto createCarDto(Car car) {
+         return // ... custom factory logic 
+     }
+}
+----
+====
+
 
 == Advanced mapping options
 This chapter describes several advanced options which allow to fine-tune the behavior of the generated mapping code as needed.
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/HelperMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/HelperMethod.java
@@ -123,6 +123,16 @@ public abstract class HelperMethod implements Method {
     }
 
     /**
+     * object factory mechanism not supported for built-in methods
+     *
+     * @return false
+     */
+    @Override
+    public boolean isObjectFactory() {
+        return false;
+    }
+
+    /**
      * the conversion context is used to format an auxiliary parameter in the method call with context specific
      * information such as a date format.
      *
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
@@ -41,7 +41,7 @@ import org.mapstruct.ap.internal.model.source.Method;
 public abstract class MappingMethod extends ModelElement {
 
     private final String name;
-    private final List<Parameter> parameters;
+    private List<Parameter> parameters;
     private final Type returnType;
     private final Parameter targetParameter;
     private final Accessibility accessibility;
@@ -64,8 +64,14 @@ public abstract class MappingMethod extends ModelElement {
     protected MappingMethod(Method method, Collection<String> existingVariableNames,
                             List<LifecycleCallbackMethodReference> beforeMappingReferences,
                             List<LifecycleCallbackMethodReference> afterMappingReferences) {
+        this( method, method.getParameters(), existingVariableNames, beforeMappingReferences, afterMappingReferences );
+    }
+
+    protected MappingMethod(Method method, List<Parameter> parameters, Collection<String> existingVariableNames,
+                            List<LifecycleCallbackMethodReference> beforeMappingReferences,
+                            List<LifecycleCallbackMethodReference> afterMappingReferences) {
         this.name = method.getName();
-        this.parameters = method.getParameters();
+        this.parameters = parameters;
         this.returnType = method.getReturnType();
         this.targetParameter = method.getMappingTargetParameter();
         this.accessibility = method.getAccessibility();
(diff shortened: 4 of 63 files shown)


remmeier:
not yet ready to be pulled in. Have a look the changes whether the approach is ok.

I'm not quite sure about the best approach for MappingResolverImpl where I make use setParameters(...). Maybe there is a better way to do the generation of the factorymethod call with freemarker.

Sometimes I get java.lang.NoClassDefFoundError: org/mapstruct/ObjectFactory where I'm not quite sure where this is coming from.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
```diff
@@ -238,7 +241,11 @@ private SourceMethod(Type declaringMapper, ExecutableElement executable, List<Pa
         this.mapperToImplement = mapperToImplement;
     }
 
-    private Parameter determineMappingTargetParameter(Iterable<Parameter> parameters) {
+    private boolean determineIfIsObjectFactory(ExecutableElement executable) {
+		return executable.getAnnotation(ObjectFactory.class) != null || getTargetTypeParameter() != null || parameters.isEmpty();
```
agudian: Use the `ObjectFactoryPrism` here to determine if the executable is annotated with `ObjectFactory`.

Rule of thumb: we don't use any classes from the annotation modules (`mapstruct` and `mapstruct-jdk8`) in the processor (with the exception of the reference in the generator class for the prisms). That way the processor can run without having the annotations on its own classpath. And it usually doesn't have them, which would explain your `NoClassDefFound` errors.

processor/src/main/java/org/mapstruct/ap/internal/model/HelperMethod.java
```diff
@@ -121,6 +121,11 @@ public Parameter getMappingTargetParameter() {
     public Parameter getTargetTypeParameter() {
         return null;
     }
+    
+    @Override
+   	public boolean isObjectFactory() {
+   		return getTargetTypeParameter() != null || getParameters().isEmpty();
```
agudian: just return `false` here.

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
```diff
@@ -216,4 +216,9 @@ public MapperConfiguration getMapperConfiguration() {
     public boolean isUpdateMethod() {
         return getMappingTargetParameter() != null;
     }
+
+    @Override
+   	public boolean isObjectFactory() {
+   		return getTargetTypeParameter() != null || getParameters().isEmpty();
```
agudian: just return `false` here. We wouldn't create a forged factory method, at least not yet.

processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/BuiltInMethod.java
```diff
@@ -141,6 +141,11 @@ public Parameter getTargetTypeParameter() {
         return null;
     }
 
+    @Override
+	public boolean isObjectFactory() {
+		return getTargetTypeParameter() != null || getParameters().isEmpty();
```
agudian: just return `false` here. We have no built-in methods that act as factory methods.

remmeier:
thx for the input, I updated the PR accordingly. The no NoClassDefFoundError have gone away.

I added two further test cases with multiple input sources. I suspect maybe more can be done in that area.

processor/src/test/java/org/mapstruct/ap/test/factories/Bar6Factory.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.mapstruct.ObjectFactory;
+
+/**
+ * @author Remo Meier
+ */
+public class Bar6Factory {
+
+    @ObjectFactory
+    public Bar6 createBar6(Foo6A foo6A) {
```
agudian: This wouldn't work if you would require only `Foo6B` as argument type, right?

I think somewhere around the Lifecycle methods, we have some code that finds all possible (assignable) combinations for passing source/target arguments to all candidate methods. I think we need to do something very similar here (maybe even reuse the code?), but raise an error in case of multiple matches - at least I can't say from the top of my head how we could select a "best matching" one in case of ambiguities. Well, perhaps favor a factory method with arguments above one without any arguments. But other than that, how would we choose between one that takes Foo6A and one that takes Foo6B?

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
```diff
@@ -86,6 +86,10 @@ protected MappingMethod(Method method, List<LifecycleCallbackMethodReference> be
         this( method, method.getParameterNames(), beforeMappingReferences, afterMappingReferences );
     }
 
+    public void setParameters(List<Parameter> parameters) {
+        this.parameters = parameters;
```
agudian: Please leave the class to be immutable and either add a dedicated construtor or convert the class to use a builder instead.

core-common/src/main/java/org/mapstruct/ObjectFactory.java
```diff
@@ -0,0 +1,46 @@
...
+ * the closest one is chosen.
+ * <p>
+ * The factory method can optionally retrieve the mapping sources by specifying
+ * them as parameters, exactly like mapping methods do. This allows the factory
+ * method to create target beans based on source beans, which opens up a variety
+ * of possibilities. For example, a factory method for entities can look up an
+ * EntityManager to check whether the entity already exists and then return its
+ * managed instance.
+ *
+ * @author Remo Meier
```
agudian: Could you also add actual code examples here? Over the time we want to convert our javadoc to be more example driven, and I think we already have some in the existing annotations.

agudian: The doc should also mention that the annotation is optional for factory methods without parameters.

core-common/src/main/java/org/mapstruct/ObjectFactory.java
```diff
@@ -0,0 +1,46 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * By default beans are created during the mapping process with the default constructor.
+ * This method allows to mark a method as a factory method to create beans.
+ * The type of the factory method is determined by its return type. A factory is used
+ * to create a bean if their types match. If multiple factories match,
+ * the closest one is chosen.
```
agudian: Whatever we end up with, we should explain what "closest one" is referring to 🙂.

agudian:
Dear Jenkins, ok to test!

remmeier:
done, 

Parameter assigment is now shared with the lifecycle methods. If there are multiple factory methods, it chosen the one with a valid assigment. If multiple ones would be possible, an exception is thrown. The only exception is the factory method without parameters, that one is only chosen if no other factory is found.

otherwise:
- some smaller doc udpates.
- more test cases

agudian:
Nice, I've just checked it out and played around with the tests a bit. Feels really good.

Two more things we should support, I think: a `@TargetType` argument would be cool:
```
   @ObjectFactory
    public <T> T createGeneric(Foo7B foo7B, @TargetType Class<T> clazz) {
        return ...;
    }
```
That's a pretty easy part. In `MappingResolverImpl.getFactoryMethod`, you can add a `Parameter` for the target type and pass that together with the list of the source parameters into `getParameterAssignments`:
```
        List<Parameter> availableParams = new ArrayList<Parameter>( mappingMethod.getSourceParameters() );
        availableParams.add( new Parameter(
            null,
            typeFactory.classTypeOf( targetType ),
            false,
            true,
            false ) );
```

Then there's another thing that we should have to help users in case they face the *ambigious factory method* error message, and that's supporting a qualifier annotation (via `@BeanMapping(qualifiedBy = ..)` or `@BeanMapping(qualifiedByName = ..)`). Or does that work already? In any case, the qualifier would also have to overrule the factory-method-with-params over factory-method-without-params logic... And I didn't try that part out either, so it might work already 😀 .

Then we also need to add some tests below the package `org.mapstruct.ap.test.erroneous` that check the various error scenarios and the proper generation of the error messages (I've seen them while playing around, but we should have some tests there as well).

It would be great if you could follow up some more with this to bring it to the finish line, if you have the time. We'd really appreciate it.

gunnarmorling:
>  in case they face the ambigious factory method error message, and that's supporting a qualifier annotation (via @BeanMapping(qualifiedBy = ..) or @BeanMapping(qualifiedByName = ..))

+1 I'd prefer explicit resolution using a qualifier over the suggested logic (it seems not intuitive that one should be preferred over the other).

remmeier:
I committed something to clarify the qualifier (not usable currently).

There is a Bar10Factory and SourceTargetMapperAndBar2Factory has different methods mapping to Bar10. Is it intended like this? I did not yet really figure out how all those qualifiers are processed (and did also not use those in the past). MappingResolveRimpl.getFactoryMethod(...) does get called with SelectionParameters == null for foo10ToBar10Upper. I expected to see Bar10Qualifier. Looking at the code, it seems like I would have to use the BeanMapping annotation, which is then passed to MappingOptions and finally to the getFactoryMethod(...)  to get something different? Any hints how I have to proceed?

Once that is done, the the new qualifier attributes in ObjectFactory need to be considered for the selection process.

agudian:
Well, you've found us a documentation gap: `@BeanMapping#qualifiedBy` and `@BeanMapping#qualifiedByName` are not described with regards to factory methods in our reference docs.
But the Javadoc on those attributes gives a hint and we have an example in the tests (`org.mapstruct.ap.test.selection.qualifier.MovieFactoryMapper`). For your test, this would mean that you would the define the mapping methods the following way:

```java
    @BeanMapping(qualifiedBy = Bar10Qualifier.class)
    public abstract Bar10 foo10ToBar10Upper(Foo10 foo10);
    
    @BeanMapping(qualifiedByName = "Bar10NamedQualifier")
    public abstract Bar10 foo10ToBar10Camel(Foo10 foo10);
```
and the factory methods:
```java
    @Bar10Qualifier
    @ObjectFactory
    public Bar10 createBar10Upper(Foo10 foo10) {
        return new Bar10( foo10.getProp().toUpperCase() );
    }

    @Named("Bar10NamedQualifier")
    @ObjectFactory
    public Bar10 createBar10Camel(Foo10 foo10) {
        return new Bar10( foo10.getProp().charAt( 0 ) + foo10.getProp().toLowerCase().substring( 1 ) );
    }
```

So `@ObjectFactory` wouldn't have the attributes `qualifiedBy` and `qualifiedByName`. That's the mechanism used for mapping methods and the simple factory methods and it should apply here just the same.
As for the selection mechanism, the information about the settings of the qualifier attributes in `@BeanMapping` should be part of the `SelectionParameters` and evaluated in the `QualifierSelector`. So this might work already...?

As a side note for us, we should really refactor that `FactoryTest` after we're done with this PR... it's way too big for some time now and I bet the size of it already makes debugging your cases quite time consuming... 😉

remmeier:
I see, the other way around with the annotation. Yes, in this case everything works already. I added nevertheless the testcases, added support for TargetType and a ambiguous test.

agudian:
Okay, I have just two more tiny comments. Would be great if you could rebase to the head of the master branch and squash your commits into one.
Then this should be ready to merge from my side. 👍

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -1486,6 +1487,28 @@ public class CarMapperImpl implements CarMapper {
 ----
 ====
 
+Annotating a factory method further with `@ObjectFactory` lets you gain access to the mapping sources. 
+Source objects can be added as parameters in the same way as for mapping method. The `@ObjectFactory`
+annotation is necessary to let MapStruct know that the given method is only a factory method, not a 
+complete mapping method.
```
sjaakd: mapping method..  (leave complete out)

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -1486,6 +1487,28 @@ public class CarMapperImpl implements CarMapper {
 ----
 ====
 
+Annotating a factory method further with `@ObjectFactory` lets you gain access to the mapping sources.
```
sjaakd: further == additionally?

sjaakd: Or better: In addition, annotating a factory method with..

processor/src/test/java/org/mapstruct/ap/test/erroneous/ambiguousannotatedfactorymethod/AmbiguousAnnotatedFactoryTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.compilation.annotation.CompilationResult;
+import org.mapstruct.ap.testutil.compilation.annotation.Diagnostic;
+import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutcome;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+/**
+ * @author Sjaak Derksen
```
sjaakd: I kind of like the current one :smile:

sjaakd:
> could you guys please check out the API change and voice any concerns, please?

I'm ok 👍 ..

sjaakd:
> could you guys please check out the API change and voice any concerns, please?

Perhaps one general remark.. `ObjectFactory` sounds like an annotation that has more capabilities than creating target objects from source parameters. Should we be more precise in the name or do you see a broader use later on? To name one: I experimented a while ago with having a factory for mappers themselves.

gunnarmorling:
Are these factory methods supported for other component models (CDI, Spring), too? I.e. could a factory type providing a parameterized factory method be obtained via DI and `Mapper#uses()`?

processor/src/main/java/org/mapstruct/ap/internal/model/ParameterAssignmentUtil.java
```diff
@@ -0,0 +1,93 @@
...
+        for ( Parameter param : parameters ) {
+            if ( name.equals( param.getName() ) ) {
+                return param;
+            }
+        }
+
+        return null;
+    }
+
+    private static List<Parameter> findCandidateParameters(List<Parameter> candiateParameters, Parameter parameter) {
```
gunnarmorling: candiateParameters=candidateParameters

What is a candidate parameter? Some description would be nice.

remmeier:
done:
- split of test cases
- minor doc updates as suggested
- added @since 1.2
- simplified factory selection (no constructor factory handled like all the others)

squash to follow

remmeier:
I did not touch any of the CDI/Spring related things. I suppose if it work before, it works know.

agudian:
💯 🎉 
Merged. Thanks a lot, Remo! It's a great addition! Chime in again anytime if you find something else missing!

I've also pushed a commit to add your name to the list of contributors.

gunnarmorling:
Very cool, thanks @remmeier et al.!

sjaakd:
My congratulations as well. Nice piece of work!

remmeier:
thx as well. 

Is there any roadmap for a first 1.2 CR release? but there is no hurry, I can always release it locally.

gunnarmorling:
Hi Remo,

we haven't any fixed schedule for the 1.2 release yet (it will definitely
be much quicker than the 1.1 release). For the time being you can either
build yourself locally or download the snapshot from the Sonatype repo:


https://oss.sonatype.org/content/repositories/snapshots/org/mapstruct/mapstruct-processor/1.2.0-SNAPSHOT/

Issue #962, opened 2016-11-19
Add support to Java 8 Streams


svpace:
Add support for stream attributes. Allowing mappings like:
``
class From {
    public Stream<Foo> getFoos() {...}
}
``
to
``
class To {
    public void setBars(Stream<Bar> bars)  {...}
}
``
or even
``
class To {
    public void setBars(Set<Bar> bars)  {...}
}
``
by applying the proper Collector

agudian:
Hmm, interesting! It would probably make sense to generate the fluent kind of mapping with method references, such as `to.setBars( from.getFoos().map( this::toBar ) );`...
The Stream to Collection conversion could be done by adding appropriate built-in Conversions (again, for in-line usage).
Especially that last part shouldn't be too hard to get started with (in the code it's somehow similar to our numeric conversions). Would you care to try it and prepare a PR?

filiphr:
Support for this was added in #964

PR #956, opened 2016-11-13
#954 NVCS.ALWAYS and presence checker must not lead to inclusion of explicit null set

sjaakd:

Issue #877, opened 2016-08-31
Use a source-code formatter lib to format the generated code


agudian:
We're taking great pride in the readability of our generated code, and that includes a proper formatting. Currently, we're tweaking the ftl templates where we can until the result looks good and we have our `IndentationCorrectingWriter` that takes care of indentations and strips out double newlines.

But removing unecessary line breaks is always a pain, and we still don't have it right in some places. Plus, that tinkering makes the ftl files less readable in some cases and often we need to keep the missing line break at eof in those templates, which isn't pretty either.

So I just thought: we could also throw away that extra stuff, remove the workarounds from the ftl files and just pipe the generated java files through a code-formatter library.

For example, there's the fully configurable Eclipse code formatter that can be invoked programmatically. It's part of the JDT-core artifact that we currently only use in test-scope and as optional dependency during the compilation.
We could even allow customization of the code-style to be used at a later point (using a bloated config-annotation or by providing a config file via processor-option).

Thoughts?

natros:
There is also https://github.com/google/google-java-format the only problem is that formats the code to comply with the Google code style.

gunnarmorling:
I'm a bit torn. It's a bit sad that we should not be able to achieve a
proper formatting ourselves but as you say formatting errors keep to pop
up. I'm not thrilled by the thought of adding another dependency "just" for
formatting.

Still +1 for experimenting with it, I'd hope such lib would not add to much
weight to the processor JAR (when shaded in)?

gunnarmorling:
I recently learned about [Roaster](https://github.com/forge/roaster), which offers formatting functionality based on Eclipse. It may also be sufficient to use the plain Eclipse formatter directly, but then at least Roaster would show how to do it.

agudian:
The cool thing about Roaster is that it also has a nice API for creating an AST... which is something we might want to go for anyway... :)

anborg:
>I'm not thrilled by the thought of adding another dependency "just" for formatting. 

Adding another dependency for formatting looks excessive, perhaps need to look for simple custom code-formatter options.

gunnarmorling:
To be fair, that's past me speaking there :) Nowadays, I'd suggest to use the Maven formatter plug-in (or one of the other options, like Spotless), and be done with it.

filiphr:
[...]

PR #964, opened 2016-11-22
#962 Java 8 Stream conversions to Collections

filiphr:
This PR covers the easier part of #962 (converting a `Stream` to some collection type).  The implementation type are same as for the [collection mappings](http://mapstruct.org/documentation/1.1/reference/html/index.html#implementation-types-for-collection-mappings) (without the `Map` implementations of course 😄)

Documentation is not updated, once this is resolved I will update the documentation.

Also I am not sure why, but in the generated mappers there are no null checks for the Stream conversions. What am I doing wrong?

### DIFF
--- a/documentation/src/main/asciidoc/mapping-streams.asciidoc
+++ b/documentation/src/main/asciidoc/mapping-streams.asciidoc
@@ -1,0 +1,67 @@
+[[mapping-streams]]
+== Mapping Streams
+
+The mapping of `java.util.Stream` is done in a similar way as the mapping of collection types, i.e. by defining mapping
+methods with the required source and target types in a mapper interface.
+
+The generated code will contain the creation of a `Stream` from the provided `Iterable`/`Array` or will collect the
+provided `Stream` into an `Iterable`/`Array`. If a mapping method or an implicit conversion for the source and target
+element types exists, then this conversion will be done in `Stream.map`. The following shows an example:
+
+.Mapper with stream mapping methods
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+@Mapper
+public interface CarMapper {
+
+    Set<String> integerStreamToStringSet(Stream<Integer> integers);
+
+    List<CarDto> carsToCarDtos(Stream<Car> cars);
+
+    CarDto carToCarDto(Car car);
+}
+----
+====
+
+The generated implementation of the `integerStreamToStringSet` performs the conversion from `Integer` to `String` for
+each element, while the generated `carsToCarDtos()` method invokes the `carToCarDto()` method for each contained
+element as shown in the following:
+
+.Generated stream mapping methods
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+//GENERATED CODE
+@Override
+public Set<String> integerStreamToStringSet(Stream<Integer> integers) {
+    if ( integers == null ) {
+        return null;
+    }
+
+    return integers.stream().map( integer -> String.valueOf( integer ) )
+        .collect( Collectors.toCollection( HashSet<String>::new ) );
+}
+
+@Override
+public List<CarDto> carsToCarDtos(Stream<Car> cars) {
+    if ( cars == null ) {
+        return null;
+    }
+
+    return integers.stream().map( car -> carToCarDto( car ) )
+        .collect( Collectors.toCollection( ArrayList<CarDto>::new ) );
+}
+----
+====
+
+[WARNING]
+====
+If a mapping from a `Stream` to an `Iterable` or an `Array` is performed, then the passed `Stream` will be consumed
+and will no longer be possible to consume it.
+====
+
+The same implementation types as in <<implementation-types-for-collection-mappings>>, are used for the creation of the
+collection when doing `Stream` to `Iterable` mapping.
--- a/documentation/src/main/asciidoc/mapping-streams.asciidoc
+++ b/documentation/src/main/asciidoc/mapping-streams.asciidoc
@@ -4,9 +4,9 @@
 The mapping of `java.util.Stream` is done in a similar way as the mapping of collection types, i.e. by defining mapping
 methods with the required source and target types in a mapper interface.
 
-The generated code will contain the creation of a `Stream` from the provided `Iterable`/`Array` or will collect the
-provided `Stream` into an `Iterable`/`Array`. If a mapping method or an implicit conversion for the source and target
-element types exists, then this conversion will be done in `Stream.map`. The following shows an example:
+The generated code will contain the creation of a `Stream` from the provided `Iterable`/array or will collect the
+provided `Stream` into an `Iterable`/array. If a mapping method or an implicit conversion for the source and target
+element types exists, then this conversion will be done in `Stream#map()`. The following shows an example:
 
 .Mapper with stream mapping methods
 ====
@@ -25,7 +25,7 @@ public interface CarMapper {
 ----
 ====
 
-The generated implementation of the `integerStreamToStringSet` performs the conversion from `Integer` to `String` for
+The generated implementation of the `integerStreamToStringSet()` performs the conversion from `Integer` to `String` for
 each element, while the generated `carsToCarDtos()` method invokes the `carToCarDto()` method for each contained
 element as shown in the following:
 
@@ -59,9 +59,9 @@ public List<CarDto> carsToCarDtos(Stream<Car> cars) {
 
 [WARNING]
 ====
-If a mapping from a `Stream` to an `Iterable` or an `Array` is performed, then the passed `Stream` will be consumed
-and will no longer be possible to consume it.
+If a mapping from a `Stream` to an `Iterable` or an array is performed, then the passed `Stream` will be consumed
+and it will no longer be possible to consume it.
 ====
 
-The same implementation types as in <<implementation-types-for-collection-mappings>>, are used for the creation of the
+The same implementation types as in <<implementation-types-for-collection-mappings>> are used for the creation of the
 collection when doing `Stream` to `Iterable` mapping.
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/StreamMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/StreamMappingMethod.java
@@ -1,0 +1,349 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model;
+
+import static org.mapstruct.ap.internal.util.Collections.first;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import javax.lang.model.type.TypeKind;
+
+import org.mapstruct.ap.internal.model.assignment.Assignment;
+import org.mapstruct.ap.internal.model.assignment.Java8FunctionWrapper;
+import org.mapstruct.ap.internal.model.common.Parameter;
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.model.source.ForgedMethod;
+import org.mapstruct.ap.internal.model.source.FormattingParameters;
+import org.mapstruct.ap.internal.model.source.Method;
+import org.mapstruct.ap.internal.model.source.SelectionParameters;
+import org.mapstruct.ap.internal.prism.NullValueMappingStrategyPrism;
+import org.mapstruct.ap.internal.util.JavaStreamConstants;
+import org.mapstruct.ap.internal.util.Message;
+import org.mapstruct.ap.internal.util.Strings;
+
+/**
+ * A {@link MappingMethod} implemented by a {@link Mapper} class which maps one iterable or array type to Stream.
+ * The collection elements are mapped either by a {@link TypeConversion} or another mapping method.
+ *
+ * @author Filip Hrisafov
+ */
+public class StreamMappingMethod extends MappingMethod {
+
+    private final Assignment elementAssignment;
+    private final MethodReference factoryMethod;
+    private final boolean overridden;
+    private final boolean mapNullToDefault;
+    private final String loopVariableName;
+    private final SelectionParameters selectionParameters;
+    private final Set<Type> helperImports;
+
+    public static class Builder {
+
+        private Method method;
+        private MappingBuilderContext ctx;
+        private SelectionParameters selectionParameters;
+        private FormattingParameters formattingParameters;
+        private NullValueMappingStrategyPrism nullValueMappingStrategy;
+
+        public Builder mappingContext(MappingBuilderContext mappingContext) {
+            this.ctx = mappingContext;
+            return this;
+        }
+
+        public Builder method(Method sourceMethod) {
+            this.method = sourceMethod;
+            return this;
+        }
+
+        public Builder formattingParameters(FormattingParameters formattingParameters) {
+            this.formattingParameters = formattingParameters;
+            return this;
+        }
+
+        public Builder selectionParameters(SelectionParameters selectionParameters) {
+            this.selectionParameters = selectionParameters;
+            return this;
+        }
+
+        public Builder nullValueMappingStrategy(NullValueMappingStrategyPrism nullValueMappingStrategy) {
+            this.nullValueMappingStrategy = nullValueMappingStrategy;
+            return this;
+        }
+
+        public StreamMappingMethod build() {
+            //TODO the building of the methods is the same as the IterableMappingMethod, the only difference being
+            // the static extracted getElementType and the wrapper of the assignment using Java8FunctionWrapper
+
+            Type sourceParameterType = first( method.getSourceParameters() ).getType();
+            Type resultType = method.getResultType();
+
+            Type sourceElementType = getElementType( sourceParameterType );
+            Type targetElementType = getElementType( resultType );
+
+            String loopVariableName =
+                Strings.getSaveVariableName( sourceElementType.getName(), method.getParameterNames() );
+
+            Assignment assignment = ctx.getMappingResolver().getTargetAssignment(
+                method,
+                "stream element",
+                sourceElementType,
+                targetElementType,
+                null, // there is no targetPropertyName
+                formattingParameters,
+                selectionParameters,
+                new SourceRHS( loopVariableName, sourceElementType ),
+                false
+            );
+
+            if ( assignment == null ) {
+                if ( method instanceof ForgedMethod ) {
+                    // leave messaging to calling property mapping
+                    return null;
+                }
+                else {
+                    ctx.getMessager().printMessage( method.getExecutable(), Message.ITERABLEMAPPING_MAPPING_NOT_FOUND );
+                }
+            }
+            else {
+                if ( method instanceof ForgedMethod ) {
+                    ForgedMethod forgedMethod = (ForgedMethod) method;
+                    forgedMethod.addThrownTypes( assignment.getThrownTypes() );
+
+                }
+            }
+
+            assignment = new Java8FunctionWrapper(
+                assignment,
+                new ArrayList<Type>(),
+                ctx.getTypeFactory().getType( JavaStreamConstants.FUNCTION_FQN )
+            );
+
+            // mapNullToDefault
+            boolean mapNullToDefault = false;
+            if ( method.getMapperConfiguration() != null ) {
+                mapNullToDefault = method.getMapperConfiguration().isMapToDefault( nullValueMappingStrategy );
+            }
+
+            MethodReference factoryMethod = null;
+            if ( !method.isUpdateMethod() ) {
+                factoryMethod = ctx.getMappingResolver().getFactoryMethod( method, method.getResultType(), null );
+            }
+
+            List<LifecycleCallbackMethodReference> beforeMappingMethods =
+                LifecycleCallbackFactory.beforeMappingMethods( method, selectionParameters, ctx );
+            List<LifecycleCallbackMethodReference> afterMappingMethods =
+                LifecycleCallbackFactory.afterMappingMethods( method, selectionParameters, ctx );
+
+            Set<Type> helperImports = new HashSet<Type>();
+            if ( resultType.isIterableType() ) {
+                helperImports.add( ctx.getTypeFactory().getType( JavaStreamConstants.COLLECTORS_FQN ) );
+            }
+
+            if ( !sourceParameterType.isCollectionType() && !sourceParameterType.isArrayType() &&
+                sourceParameterType.isIterableType() ) {
+                helperImports.add( ctx.getTypeFactory().getType( JavaStreamConstants.STREAM_SUPPORT_FQN ) );
+            }
+
+            return new StreamMappingMethod(
+                method,
+                assignment,
+                factoryMethod,
+                mapNullToDefault,
+                loopVariableName,
+                beforeMappingMethods,
+                afterMappingMethods,
+                selectionParameters,
+                helperImports
+            );
+        }
+    }
+
+
+    private StreamMappingMethod(Method method, Assignment parameterAssignment, MethodReference factoryMethod,
+                                boolean mapNullToDefault, String loopVariableName,
+                                List<LifecycleCallbackMethodReference> beforeMappingReferences,
+                                List<LifecycleCallbackMethodReference> afterMappingReferences,
+                                SelectionParameters selectionParameters, Set<Type> helperImports) {
+        super( method, beforeMappingReferences, afterMappingReferences );
+        this.elementAssignment = parameterAssignment;
+        this.factoryMethod = factoryMethod;
+        this.overridden = method.overridesMethod();
+        this.mapNullToDefault = mapNullToDefault;
+        this.loopVariableName = loopVariableName;
+        this.selectionParameters = selectionParameters;
+        this.helperImports = helperImports;
+    }
+
+    public Parameter getSourceParameter() {
+        for ( Parameter parameter : getParameters() ) {
+            if ( !parameter.isMappingTarget() ) {
+                return parameter;
+            }
+        }
+
+        throw new IllegalStateException( "Method " + this + " has no source parameter." );
+    }
+
+    public Assignment getElementAssignment() {
+        return elementAssignment;
+    }
+
+    @Override
+    public Set<Type> getImportTypes() {
+        Set<Type> types = super.getImportTypes();
+        if ( elementAssignment != null ) {
+            types.addAll( elementAssignment.getImportTypes() );
+        }
+        if ( ( factoryMethod == null ) && ( !isExistingInstanceMapping() ) ) {
+            if ( getReturnType().getImplementationType() != null ) {
+                types.addAll( getReturnType().getImplementationType().getImportTypes() );
+            }
+        }
+
+        types.addAll( helperImports );
+
+        return types;
+    }
+
+    public boolean isMapNullToDefault() {
+        return mapNullToDefault;
+    }
+
+    public boolean isOverridden() {
+        return overridden;
+    }
+
+    public String getLoopVariableName() {
+        return loopVariableName;
+    }
+
+    public String getDefaultValue() {
+        TypeKind kind = getResultElementType().getTypeMirror().getKind();
+        switch ( kind ) {
+            case BOOLEAN:
+                return "false";
+            case BYTE:
+            case SHORT:
+            case INT:
+            case CHAR:  /*"'\u0000'" would have been better, but depends on platformencoding */
+                return "0";
+            case LONG:
+                return "0L";
+            case FLOAT:
+                return "0.0f";
+            case DOUBLE:
+                return "0.0d";
+            default:
+                return "null";
+        }
+    }
+
+    public MethodReference getFactoryMethod() {
+        return this.factoryMethod;
+    }
+
+    public Type getSourceElementType() {
+        return getElementType( getSourceParameter().getType() );
+    }
+
+    public Type getResultElementType() {
+        return getElementType( getResultType() );
+    }
+
+    public String getIndex1Name() {
+        return Strings.getSaveVariableName( "i", loopVariableName, getSourceParameter().getName(), getResultName() );
+    }
+
+    public String getIndex2Name() {
+        return Strings.getSaveVariableName( "j", loopVariableName, getSourceParameter().getName(), getResultName() );
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ( ( getResultType() == null ) ? 0 : getResultType().hashCode() );
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if ( this == obj ) {
+            return true;
+        }
+        if ( obj == null ) {
+            return false;
+        }
+        if ( getClass() != obj.getClass() ) {
+            return false;
+        }
+        StreamMappingMethod other = (StreamMappingMethod) obj;
+
+        if ( !getResultType().equals( other.getResultType() ) ) {
+            return false;
+        }
+
+        if ( getSourceParameters().size() != other.getSourceParameters().size() ) {
+            return false;
+        }
+
+        for ( int i = 0; i < getSourceParameters().size(); i++ ) {
+            if ( !getSourceParameters().get( i ).getType().equals( other.getSourceParameters().get( i ).getType() ) ) {
+                return false;
+            }
+
+            List<Type> thisTypeParameters = getSourceParameters().get( i ).getType().getTypeParameters();
+            List<Type> otherTypeParameters = other.getSourceParameters().get( i ).getType().getTypeParameters();
+
+            if ( !thisTypeParameters.equals( otherTypeParameters ) ) {
+                return false;
+            }
+        }
+
+        if ( this.selectionParameters != null ) {
+            if ( !this.selectionParameters.equals( other.selectionParameters ) ) {
+                return false;
+            }
+        }
+        else if ( other.selectionParameters != null ) {
+            return false;
+        }
+
+        return isMapNullToDefault() == other.isMapNullToDefault();
+    }
+
+    private static Type getElementType(Type parameterType) {
+        if ( parameterType.isArrayType() ) {
+            return parameterType.getComponentType();
+        }
+        else if ( parameterType.isIterableType() ) {
+            return first( parameterType.determineTypeArguments( Iterable.class ) ).getTypeBound();
+        }
+        else if ( parameterType.isStreamType() ) {
+            return first( parameterType.determineTypeArguments( JavaStreamConstants.STREAM_FQN ) ).getTypeBound();
+        }
+
+        throw new IllegalArgumentException( "Could not get the element type" );
+
+    }
+
+}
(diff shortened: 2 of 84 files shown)


sjaakd:
> Also I am not sure why, but in the generated mappers there are no null checks for the Stream conversions. What am I doing wrong?

Hmm. That should be included automatically. You might want to single-step through `assignToPlain`.  Although you might argue that `Optional.ofNillable` might be more in line with Java8. :smile:

filiphr:
@sjaakd Thanks for the pointer. However, the problem is not there it is in `assignToCollection`. As the `targetType` is `Collection` the entire conversion is `null` checked but not the `<SOURCE>` in the conversion.

Example:
```java
List<String> list = source.getStringStream().collect( Collectors.toList() );
if ( list != null ) {
    target.setStringStream( list );
}
```

The `<SOURCE>` in the previous snippet is `source.getStringStream()`.

I also tried with `NullValueCheckStrategy.ALWAYS` and same result. Is this expected or is it a bug?

> Although you might argue that Optional.ofNillable might be more in line with Java8. 😄

I agree, but I think that I shouldn't do a `null` check in the conversion itself?

sjaakd:
> As the targetType is Collection the entire conversion is null checked but not the <SOURCE> in the conversion.

ahh.. now I get it.. This (PR) is the first time we have a type conversion from collection to collection type. I believe it is (or at least was) even in the comments somewhere. I refactored these particular parts some commits ago.

For mapping methods, the responsibility for null checking is in the (forged) mapping method. 

So perhaps its not such a bad idea to do it in the conversion in Java8 style. Form experience: the placing of the null check is the result of a long discussion (see #956 for instance) and a bit of a can of worms. 

BTW: I'm still busy with refactoring this whole property mapping (step-by-step). I think it has become far to complex / nontransparent.

filiphr:
Done. Using Optional it looks much better now, thanks :smile:. The only thing is that, it is in one line and it is a bit long, I am not sure if it is possible to do it multi-line and to look nice.

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaStreamToCollectionConversion.java
```diff
@@ -41,7 +41,9 @@
 
     @Override
     protected String getToExpression(ConversionContext conversionContext) {
-        return "<SOURCE>.collect( Collectors." + collector( conversionContext ) + " )";
+        return "Optional.ofNullable( <SOURCE> )" +
```
sjaakd: could simply adding enters not be an option

sjaakd:
> Using Optional it looks much better now, thanks 😄

No problem.. I'll have a thorough look tomorrow. Thanks.

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaStreamToCollectionConversion.java
filiphr: It doesn't help. The `TypeConversion.ftl` has `compress single_line = true` which leads to making the `\n` into spaces (At least I think it is because of that)

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaStreamToCollectionConversion.java
```diff
@@ -0,0 +1,95 @@
...
+            collector = "toList()";
+        }
+        else if ( typeFactory.getType( Set.class ).isAssignableTo( targetType.erasure() ) ) {
+            collector = "toSet()";
+        }
+        else if ( acceptsTreeSet( conversionContext ) ) {
+            collector = "toCollection( TreeSet::new )";
+        }
+
+        //TODO what if we register wrong. In the end it won't compile, but should we throw an exception?
```
gunnarmorling: Yes, we should never generate code that doesn't compile. Instead we should raise a descriptive error about the issue before (missing conversion etc.).

processor/src/test/java/org/mapstruct/ap/test/conversion/java8stream/Source.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * @author Filip Hrisafov
+ */
+public class Source {
+
+
+    private List<Integer> ints;
+
+    private Stream<Integer> stream;
```
gunnarmorling: Do we have a test that invokes a method for converting the element type?

processor/src/test/java/org/mapstruct/ap/test/conversion/java8stream/StreamsTest.java
```diff
@@ -0,0 +1,100 @@
...
+        Target target = new Target();
+        target.setTargetStream( stream );
+        target.setStringStream( Arrays.asList( "4", "5", "6", "7" ) );
+        target.setInts( Arrays.asList( 1, 2, 3 ).stream() );
+        target.setIntegerSet( Collections.asSet( 1, 1, 2, 2, 4, 4 ) );
+        target.setStringCollection( Collections.asSet( "1", "1", "2", "3" ) );
+        target.setIntegerIterable( Arrays.asList( 10, 11, 12 ) );
+        target.setSortedSet( new TreeSet<Integer>( Arrays.asList( 12, 11, 10 ) ) );
+        target.setNavigableSet( new TreeSet<Integer>( Arrays.asList( 12, 11, 10 ) ) );
+
```
gunnarmorling: The double empty lines look odd (also below).

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaStreamToCollectionConversion.java
```diff
@@ -0,0 +1,95 @@
...
+        return "Optional.ofNullable( <SOURCE> )" +
+            ".map( s -> s.collect( Collectors." + collector( conversionContext ) + " ) )" +
+            ".orElse( null )";
+    }
+
+    private String collector(ConversionContext conversionContext) {
+        Type targetType = conversionContext.getTargetType();
+        TypeFactory typeFactory = conversionContext.getTypeFactory();
+        String collector = "";
+        // we must do erasure as JDK compiler does not work without it
```
sjaakd: Hmm. I'm not sure whether this is correct. `Iterable<String>` is not assignable to `Iterable<Object>`.. Is it the JDK compilation cycle that fails or the Eclipse? (Note both are executed)

processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
```diff
@@ -352,15 +366,15 @@ public boolean equals(Object obj) {
                     return false;
                 }
             }
-            else if ( !sourceType.equals( other.sourceType ) ) {
+            else if ( !sourceType.erasure().equals( other.sourceType.erasure() ) ) {
```
sjaakd: why is this needed?

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaStreamToCollectionConversion.java
```diff
@@ -0,0 +1,95 @@
...
+import org.mapstruct.ap.internal.util.Collections;
+import org.mapstruct.ap.internal.util.JavaStreamConstants;
+
+/**
+ * Conversion between {@link java.util.stream.Stream} and java utils: {@link Set}, {@link java.util.List},
+ * {@link java.util.Collection}.
+ *
+ * @author Filip Hrisafov
+ */
+public class JavaStreamToCollectionConversion extends SimpleConversion {
```
sjaakd: perhaps we should work with an abstract `JavaStreamToCollectionConversionBase` which has specific implementations for `Iterable, `Collection`, `Set`.. etc.  Then you could make the method `collector` abstract and removing the need for the current collector. This is done for other conversions as well (harmonize) and moves this if-elseif-elseif to the `Conversions`.

processor/src/test/java/org/mapstruct/ap/test/conversion/java8stream/Source.java
sjaakd: Indeed. I'm a bit worried on the erasure. Possible problems would show.

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaStreamToCollectionConversion.java
filiphr: We will already raise such error, as no conversion could be found before. The only reason for something like this to happen is if we register a new type and forget to update the `JavaStreamToCollectionConversion`. In order to avoid such things, we could maybe do something in the register methods and make sure that the providers supports the conversion that we are trying to register.

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaStreamToCollectionConversion.java
sjaakd: ah.. we should fix that someday. I'll try to take that on when I do the refactoring on the templates for the property mapping. For now.. leave it.

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaStreamToCollectionConversion.java
filiphr: This conversion is only for the same type currently. Without the `erasure()` the JDK compilation fails, and the Eclipse is OK. I think that the problem is that when we do `typeFactory.getType( Iterable.class )` we get a type like `Iterable<T>`. I am not sure what exactly is correct and how to check if the `targetType` is `Iterable`, or `Set`, or something else.

processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
filiphr: This is another point where I was not sure. If they are removed then the conversions are not found.
It seems that the type `Iterable<T>` is not the same to `Iterable<T>` (I don't know why)

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaStreamToCollectionConversion.java
sjaakd: @agudian .. could it  be that we need the `declared type` iso`type` comparison.?

processor/src/test/java/org/mapstruct/ap/test/conversion/java8stream/Source.java
filiphr: The erasure is a problem as it will allow a `List<String>` to be converted from/to `Stream<Integer>`. I think that we need to do something like in the `BuiltInMethod`(s) there the generics are handled nice.  Maybe even use them instead

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaStreamToCollectionConversion.java
agudian: See #877 🙄 😄

filiphr:
@gunnarmorling, @agudian, @sjaakd I rethought the approach about the implementation of this feature and I went in the same direction as the `IterableMappingMethod`. Should I just update this PR or you would prefer a different PR with the new approach?

The new approach works with mapping the element type considerts the different mappings:
* `Iterable` or `Array` type -> Stream type
* `Stream<TypeA>` to `Stream<TypeB>`
* `Stream` -> `Iterable` or `Array` type

filiphr:
I referenced a wrong issue with my last commit. 

I changed to approach for the stream mappings to be inline with the `IterableMappingMethod`. Mapping between different types work by using the stream `map` method.

I think that there are some formatting issues in the generated mappers that need to be resolved.

processor/src/test/java/org/mapstruct/ap/test/conversion/java8stream/Source.java
filiphr: With my latest commits the approach is changed and we have a test that maps `Stream<String>` to `List<Integer>`.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -836,4 +848,35 @@ public boolean hasEmptyAccessibleContructor() {
...
+     * with Java 6.
+     *
+     * @param superClassName the superclass or interface the generic type arguments are searched for
+     * @return a list of type arguments or null, if superclass was not found
+     */
+
+    public Collection<Type> determineTypeArguments(String superClassName) {
+        Class<?> superClass = null;
+        try {
+            superClass = Class.forName( superClassName );
```
gunnarmorling: What classes could we see here? If it's classes from the application under complication, `Class#forName()` should be avoided. `Elements#getTypeElement()` would be an alternative.

processor/src/test/java/org/mapstruct/ap/test/conversion/java8stream/Source.java
gunnarmorling: Does it also work by invoking another mapping method for the element type now? Or is that left for a follow-up PR?

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: I needed to get the type element of the `java.util.stream.Stream` being mapped. I can try the `Elements#getTypeElement()` to see if it works like that. I didn't know about that. I can't call our `determineTypeArguments(Class<?> superClass)` as the `Stream` does not compile with Java 6 and 7

processor/src/test/java/org/mapstruct/ap/test/conversion/java8stream/Source.java
filiphr: Yes, with the new approach it takes the same approach ad the `IterableMappingMethod` and checks for a mapping method for the element type and uses it. I will add a test for that in any case.

filiphr: @gunnarmorling I forgot that I already added that. Have a look at the `TargetElement` and `SourceElement`, they are mapped via another mapping method.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: @gunnarmorling I had a look at the method that we were using before, and the `Class` was only needed to get it's name. Therefore, I did some changes in my method and moved checks there.

gunnarmorling:
@filiphr Are you going to squash your commits into one (or a few ones)? E.g. the "removing imports" one seems not worth a separate entry in the commit history. Also can you prefix all commits that are going to remain with the issue number?

filiphr:
@gunnarmorling yes of course. I have to rebase everything as there is some conflict with the `TypeFactory`. We can also do squash and merge directly from the PR. I thought that it is easier for reviewing if I don't squash all the time.

filiphr:
@gunnarmorling I was thinking to have one commit. The commit with the issue 971 is with a wrong issue referenced.

sjaakd:
@filiphr What's the status on this one?

filiphr:
@sjaakd I was waiting for the new PRs to be merged. I will rebase this on top of the latest master. @gunnarmorling said that will have a look at it as well.

filiphr:
I squashed and rebased on the latest master branch. @gunnarmorling can you have a look at it?

gunnarmorling:
Thanks! Yes, I'll take another look tomorrow.

processor/src/test/java/org/mapstruct/ap/test/java8stream/StreamMappingTest.java
```diff
@@ -0,0 +1,244 @@
...
+        Target target = new Target();
+
+        Source source = SourceTargetMapper.INSTANCE.targetToSource( target );
+
+        assertThat( source ).isNotNull();
+        assertThat( source.getStringList() ).isNull();
+    }
+
+    @Test
+    public void shouldMapList() {
```
gunnarmorling: Hi @filiphr, I took a look at the generated code. That's what I see:

    protected List<String> stringStreamToStringList(Stream<String> stream) {
        if ( stream == null ) {
            return null;
        }

        list.addAll( stream.map( Function.identity() )
            .collect( Collectors.toCollection( ArrayList<String>::new ) )
        );

        return list;
    }

Why is it using the identity function here? And why is it assigning the result to a new list? I think the following would be better to generate:

    protected List<String> stringStreamToStringList(Stream<String> stream) {
        if ( stream == null ) {
            return null;
        }

        return stream.collect( Collectors.toCollection( ArrayList<String>::new ) );
    }

processor/src/test/java/org/mapstruct/ap/test/java8stream/StreamMappingTest.java
```diff
@@ -0,0 +1,244 @@
...
+        Source source = SourceTargetMapper.INSTANCE.targetToSource( target );
+
+        assertThat( source ).isNotNull();
+        assertThat( source.getStringSet() ).contains( "Bob", "Alice" );
+    }
+
+    @Test
+    public void shouldMapListToCollection() {
+        Source source = new Source();
+        source.setIntegerList( Arrays.asList( 1, 2 ).stream() );
```
gunnarmorling: setIntegerList=setIntegerStream?

processor/src/test/java/org/mapstruct/ap/test/java8stream/StreamMappingTest.java
```diff
@@ -0,0 +1,244 @@
...
+        Source source = SourceTargetMapper.INSTANCE.targetToSource( target );
+
+        assertThat( source ).isNotNull();
+        assertThat( source.getIntegerList() ).containsOnly( 1, 2 );
+    }
+
+    @Test
+    public void shouldMapIntegerSetToStringSet() {
+        Source source = new Source();
+        source.setAnotherIntegerSet( new HashSet<Integer>( Arrays.asList( 1, 2 ) ).stream() );
```
gunnarmorling: setAnotherIntegerSet=setAnotherIntegerStream (it seems to be a common issue in this test)

gunnarmorling: Instead of generating

    protected Set<String> integerStreamToStringSet(Stream<Integer> stream) {
        if ( stream == null ) {
            return null;
        }

        Set<String> set = new HashSet<String>();

        set.addAll( stream.map( integer -> String.valueOf( integer ) )
            .collect( Collectors.toCollection( HashSet<String>::new ) )
        );

        return set;
    }

How about this:

    protected Set<String> integerStreamToStringSet(Stream<Integer> stream) {
        if ( stream == null ) {
            return null;
        }

        return stream.map( integer -> String.valueOf( integer ) )
            .collect( Collectors.toCollection( HashSet<String>::new ) )
        );
    }

I.e. avoid the adding to another collection.

processor/src/test/java/org/mapstruct/ap/test/java8stream/StreamMappingTest.java
gunnarmorling: Another one I saw is this:

    protected Stream<String> stringArrayListToStringStream(ArrayList<String> arrayList) {
        if ( arrayList == null ) {
            return null;
        }

        Stream<String> stream = Stream.empty();

        stream = arrayList.stream()
        .map( Function.identity() );

        return stream;
    }

I think we could avoid the assignment of `Stream.empty()` which is overridden right in the next line? And the call to `map()` looks not properly indented (I know, it's nit-picking, sorry ;).

processor/src/test/java/org/mapstruct/ap/test/java8stream/forged/StreamMappingTest.java
```diff
@@ -0,0 +1,120 @@
...
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+/**
+ * Test for mappings between collection and stream types,
+ *
+ * @author Filip Hrisafov
+ */
+@IssueKey("962")
+@RunWith(AnnotationProcessorTestRunner.class)
+public class StreamMappingTest {
```
gunnarmorling: Could you use a distinct name for this test? There's another one of the same name.

agudian:
I usually use `Collectors.toList()` / `Collectors.toSet()` instead of `toCollection( .. )`.

I guess you're that specific here to reuse the same template for the general scenario, right? Including the `map( identity() )` part?

processor/src/test/java/org/mapstruct/ap/test/java8stream/wildcard/WildCardTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Test
+    @WithClasses({
+        ExtendsBoundSourceTargetMapper.class,
+        ExtendsBoundSource.class,
+        Target.class,
+        Plan.class,
+        Idea.class
+    })
+    public void shouldGenerateExtendsBoundSourceForgedStreamMethod() {
```
gunnarmorling: Minor thing, I'm seeing a FQN in the generated code here:

    protected List<Plan> ideaStreamToPlanList(Stream<? extends org.mapstruct.ap.test.java8stream.wildcard.Idea> {
        ...
    }

gunnarmorling: (Same for the super bounds test)

processor/src/main/java/org/mapstruct/ap/internal/model/StreamMappingMethod.java
```diff
@@ -0,0 +1,404 @@
...
+            return this;
+        }
+
+        public Builder nullValueMappingStrategy(NullValueMappingStrategyPrism nullValueMappingStrategy) {
+            this.nullValueMappingStrategy = nullValueMappingStrategy;
+            return this;
+        }
+
+        public StreamMappingMethod build() {
+            //TODO the building of the methods is the same as the IterableMappingMethod, the only difference being
```
gunnarmorling: Should we track it in a separate issue?

processor/src/main/java/org/mapstruct/ap/internal/model/StreamMappingMethod.java
```diff
@@ -0,0 +1,404 @@
...
+                beforeMappingMethods,
+                afterMappingMethods,
+                selectionParameters,
+                helperImports,
+                forgedMethod
+            );
+        }
+
+        private Assignment forgeMapping(SourceRHS sourceRHS, Type sourceType, Type targetType) {
+
```
gunnarmorling: Just a small thing: there are some superfluous empty lines like this one here and there.

processor/src/main/java/org/mapstruct/ap/internal/model/StreamMappingMethod.java
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    public Parameter getSourceParameter() {
+        for ( Parameter parameter : getParameters() ) {
+            if ( !parameter.isMappingTarget() ) {
+                return parameter;
+            }
+        }
+
+        throw new IllegalStateException( "Method " + this + " has no source parameter." );
```
gunnarmorling: Is this a "regular" error situation? If so, we should raise a proper compiler error instead.

processor/src/main/java/org/mapstruct/ap/internal/model/StreamMappingMethod.java
```diff
@@ -0,0 +1,404 @@
...
+
+    public boolean isOverridden() {
+        return overridden;
+    }
+
+    public String getLoopVariableName() {
+        return loopVariableName;
+    }
+
+    public String getDefaultValue() {
```
gunnarmorling: When is this one needed?

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -840,14 +862,27 @@ public boolean hasEmptyAccessibleContructor() {
      * @return a list of type arguments or null, if superclass was not found
      */
     public List<Type> determineTypeArguments(Class<?> superclass) {
-        if ( qualifiedName.equals( superclass.getName() ) ) {
+        return determineTypeArguments( superclass.getName() );
+    }
+
+    /**
+     * Searches for the given superclass and collects all type arguments for the given className.
+     * This is used for support with Java 8 streams, as must use the FQN in order to be able to compile
```
gunnarmorling: I'm not following why it's using the FQN now. When do we compile with Java 6?

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -447,6 +456,30 @@ else if ( method.isEnumMapping() ) {
         return mappingMethods;
     }
 
+    private StreamMappingMethod createStreamMappingMethod(SourceMethod method, MappingOptions mappingOptions) {
+        // TODO this is the same as the if with the isIterableMapping, I think that we need
```
gunnarmorling: Let's create an issue so we don't forget about it.

gunnarmorling:
Hey @filiphr, great work! Awesome to have the stream support :) I've added some comments inline, mostly details. Kudos for the super-thorough testing, too!

gunnarmorling:
> I usually use Collectors.toList() / Collectors.toSet() instead of toCollection( .. ).
>
> I guess you're that specific here to reuse the same template for the general scenario, right? Including the map( identity() ) part?

I also was about to suggest to use `toList()` etc., but their comments state that there are no guarantees regarding mutability etc. So I think we are better off using the explicit constructor references of specific types.

My guess on the identity function would be the same, would still be nice to avoid if reasonably doable.

processor/src/test/java/org/mapstruct/ap/test/java8stream/StreamMappingTest.java
filiphr: Will fix this issues in the entire test

processor/src/test/java/org/mapstruct/ap/test/java8stream/StreamMappingTest.java
filiphr: See my other comments about the same issue

processor/src/test/java/org/mapstruct/ap/test/java8stream/StreamMappingTest.java
filiphr: >Why is it using the identity function here?

Identity is used so I can have a general approach. Will fix this. 

>And why is it assigning the result to a new list?

The assigning to a new list is due to the before/after mapping possible calls. I can try and fix it by adding some checks if there are any before/after calls. Maybe add it as a TODO in #1012?

>I think we could avoid the assignment of Stream.empty() which is overridden right in the next line? 

Same for the `Stream.empty()`.  There could be calls to before/after mapping methods (Although it makes no sense to me to have a before mapping for `Stream`). I can do something similar as for the list.

>And the call to map() looks not properly indented (I know, it's nit-picking, sorry ;).

I am having problems to indent this correctly :(. I am not sure how I can do it correctly, @sjaakd maybe you can help out for this one. It is in line [114](https://github.com/filiphr/mapstruct/blob/java8-stream-conversions/processor/src/main/resources/org/mapstruct/ap/internal/model/StreamMappingMethod.ftl#L114), [117](https://github.com/filiphr/mapstruct/blob/java8-stream-conversions/processor/src/main/resources/org/mapstruct/ap/internal/model/StreamMappingMethod.ftl#L117) and [120](https://github.com/filiphr/mapstruct/blob/java8-stream-conversions/processor/src/main/resources/org/mapstruct/ap/internal/model/StreamMappingMethod.ftl#L120)

processor/src/test/java/org/mapstruct/ap/test/java8stream/forged/StreamMappingTest.java
filiphr: Of course. We have the same situation with the `CollectionsMappingTest`. I reused most of the tests in `org.mapstruct.ap.test.collection` 😄

processor/src/test/java/org/mapstruct/ap/test/java8stream/wildcard/WildCardTest.java
filiphr: Will look into it.

processor/src/main/java/org/mapstruct/ap/internal/model/StreamMappingMethod.java
filiphr: Already created one see #1012

processor/src/main/java/org/mapstruct/ap/internal/model/StreamMappingMethod.java
filiphr: I'll fix those thanks.

processor/src/main/java/org/mapstruct/ap/internal/model/StreamMappingMethod.java
filiphr: This should never happen. We do the exact same thing in the `IterableMappingMethod`, `MapMappingMethod` and `NestedPropertyMappingMethod`.

processor/src/main/java/org/mapstruct/ap/internal/model/StreamMappingMethod.java
filiphr: In the `StreamMappingMethod.ftl` when we need to reset an array.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: This actually works. I only need to use FQN in the `TypeFactory`. I thought that it will fail as we use the FQN in the `Conversions`. However, I ran the integration tests and it seems OK. I will revert the changes to this.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
filiphr: Done see #1012

filiphr:
Thanks for the through review guys.

>I usually use Collectors.toList() / Collectors.toSet() instead of toCollection( .. ).
>
>I guess you're that specific here to reuse the same template for the general scenario, right? Including the map( identity() ) part?

Yes to the both. I wanted to not have too many checks in the model. The `Function.identity()` part can be easily changed. I an do a check on the assignment and if it is a direct assignment do not add the `map`. Regarding the `Collectors.toList()` / `Collectors.toSet() `, I think it is much more complicated to make it like that. If you want we can create a new issue for that and tackle it separately. I think having `Collectors.toCollection( .. )` provides us more flexibility if we want to allow the users to provide custom default implementations.

filiphr:
The `map` method is now in the same line as `stream`, `Function.identity()` is not used anymore.

The only thing that is left is:
- [ ] Do we agree to keep using `Collectors.toCollection( .. )`?
- [ ] What about the before/after possible calls?
- [ ] The FQN `org.mapstruct.ap.test.java8stream.wildcard.Idea`

processor/src/test/java/org/mapstruct/ap/test/java8stream/wildcard/WildCardTest.java
filiphr: This is a different issue see #1014

gunnarmorling:
> Do we agree to keep using Collectors.toCollection( .. )

Yes, it makes sense to me.

> What about the before/after possible calls?

If possible I'd prefer to avoid the superfluous instantiations in the regular case (no before/after). But it coould be done as a follow-up task if you prefer.

filiphr:
>If possible I'd prefer to avoid the superfluous instantiations in the regular case (no before/after). But it coould be done as a follow-up task if you prefer.

I did it as part of this PR. Have a look at my last commit.

gunnarmorling:
@filiphr The generated source code looks functionally perfect now. Thanks a lot! The only remaining issue I spotted is a minor formatting glitch:

    return stream.map( string -> Long.parseLong( string ) )
    .collect( Collectors.toCollection( HashSet<Long>::new ) );

Could you indent the second line one level? Other than that it's great, it only needs a rebase (there is one conflict with current master). And do we have the doc update already?

filiphr:
Thanks for the review @gunnarmorling.

>Could you indent the second line one level?

For some reason I can't. I would appreciate some help here. You can see the template lines [here](https://github.com/filiphr/mapstruct/blob/java8-stream-conversions/processor/src/main/resources/org/mapstruct/ap/internal/model/StreamMappingMethod.ftl#L120-L121). There is a tab, in Line 121, but once it is included in the nested it does not add it to the template. I can write it in one line and it will look like this:

```java
return stream.map( sourceElement -> mapElement( sourceElement ) ).collect( Collectors.toCollection( ArrayList<TargetElement>::new ) );
```
WDYT?

>And do we have the doc update already?

No, but I will add a commit for it to this PR

gunnarmorling:
> For some reason I can't. I would appreciate some help here. 

Ah, I reckon it's due to [IndentationCorrectingWriter](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/writer/IndentationCorrectingWriter.java) which re-indents the generated source code at the end solely based on the nesting level as indicated by "{" and "}". It doesn't deal with multi-line chained invocations as we have them here.

If you feel adventureous, you may give it a try to fix it in the writer, but if not let's merge the PR as is and handle it separately. Just let me know what you prefer.

filiphr:
>If you feel adventureous, you may give it a try to fix it in the writer, but if not let's merge the PR as is and handle it separately. Just let me know what you prefer.

I had a look at it, I think it deserves as new PR 😄 and I created #1022 for it.

I rebased the PR and added 3 new commits, one for fixing a typo, another one for the documentation and the last one for support of the new `@Context`

agudian:
> If you feel adventureous, you may give it a try to fix it in the writer, but if not let's merge the PR as is and handle it separately. Just let me know what you prefer.

Or, we could revisit the idea to simply let a code formatter do the formatting for us, as proposed in #877. A first search didn't yield that many open source libs that seem usable for us there, with the exception of the Eclipse formatter... Not sure on the licensing though, if we start shading it into our processor.jar. But I could try integrating it to see how it behaves...

gunnarmorling:
@filiphr This one is good to go, right? Can you rebase, squash as you like and merge it?

filiphr:
@gunnarmorling If you are OK with the documentation then it is good to go. I will squash and merge it from github as there are no conflicts and I think that everything belongs into one commit

agudian:
better rebase locally, as you'll also have to update the legal headers for the new files... :)

documentation/src/main/asciidoc/mapping-streams.asciidoc
```diff
@@ -0,0 +1,67 @@
+[[mapping-streams]]
+== Mapping Streams
+
+The mapping of `java.util.Stream` is done in a similar way as the mapping of collection types, i.e. by defining mapping
+methods with the required source and target types in a mapper interface.
+
+The generated code will contain the creation of a `Stream` from the provided `Iterable`/`Array` or will collect the
```
gunnarmorling: `Array` should not be marked as literal. It's just an "array".

documentation/src/main/asciidoc/mapping-streams.asciidoc
```diff
@@ -0,0 +1,67 @@
...
+provided `Stream` into an `Iterable`/`Array`. If a mapping method or an implicit conversion for the source and target
+element types exists, then this conversion will be done in `Stream.map`. The following shows an example:
```
gunnarmorling: done via `Stream#map()` (let's use the hash pound when referencing to a method).

documentation/src/main/asciidoc/mapping-streams.asciidoc
```diff
@@ -0,0 +1,67 @@
...
+    Set<String> integerStreamToStringSet(Stream<Integer> integers);
+
+    List<CarDto> carsToCarDtos(Stream<Car> cars);
+
+    CarDto carToCarDto(Car car);
+}
+----
+====
+
+The generated implementation of the `integerStreamToStringSet` performs the conversion from `Integer` to `String` for
```
gunnarmorling: the `integerStreamToStringSet()` method.

documentation/src/main/asciidoc/mapping-streams.asciidoc
```diff
@@ -0,0 +1,67 @@
...
+
+    return integers.stream().map( car -> carToCarDto( car ) )
+        .collect( Collectors.toCollection( ArrayList<CarDto>::new ) );
+}
+----
+====
+
+[WARNING]
+====
+If a mapping from a `Stream` to an `Iterable` or an `Array` is performed, then the passed `Stream` will be consumed
```
gunnarmorling: `Array`=array

documentation/src/main/asciidoc/mapping-streams.asciidoc
```diff
@@ -0,0 +1,67 @@
...
+and will no longer be possible to consume it.
```
gunnarmorling: and it

documentation/src/main/asciidoc/mapping-streams.asciidoc
```diff
@@ -0,0 +1,67 @@
...
+====
+
+The same implementation types as in <<implementation-types-for-collection-mappings>>, are used for the creation of the
```
gunnarmorling: The "," is too much.

filiphr:
Thanks for the review guys. I have update the last documentation comments and squashed and merged

Issue #966, opened 2016-11-24
Avoid multiple newlines and missing newlines


agudian:
... both in our genereated code and in our hand written code.

PR following with some fixes for missing newlines in the generated code, the corresponding checkstyle settings and fixes for our code base following up.

PR #967, opened 2016-11-24
#966 Avoid missing newlines between methods and multiple newlines…

agudian:
…by adding a checkstyle rule, fixing current incarnations.

@sjaakd, I bet this will be your favourite feature of the build in the future... 😉


processor/src/main/java/org/mapstruct/ap/internal/conversion/BigDecimalToStringConversion.java
```diff
@@ -19,15 +19,16 @@
 package org.mapstruct.ap.internal.conversion;
```
gunnarmorling: Seems it's still missing some ;) Just saying, not that I'd care too much...

agudian: There isn't an empty line between the header and the package declaration anywhere, as far as I could see, so I thought we should leave it as it is.

This is more about the missing empty lines between method declarations (generated code and the processor code), and those white-outs in some files (there was one spot with five empty lines).

sjaakd:
> @sjaakd, I bet this will be your favourite feature of the build in the future... 😉

Hmm. I just moved to 1.1.0.Final.. And saw the collection mapping implementation. Somehow I missed it in the PR itself.. But you are more than right... 👍

sjaakd:
> Hmm. I just moved to 1.1.0.Final.. And saw the collection mapping implementation. Somehow I missed it in the PR itself.. But you are more than right... 👍

I meant the generated code.. But this one is alright..Please merge

processor/src/main/java/org/mapstruct/ap/internal/conversion/BigDecimalToStringConversion.java
filiphr: @gunnarmorling I think that is because there is no `PACKAGE_DEF` (no check after package declaration) in the tokens of the `EmptyLineSeparator` check

gunnarmorling: @agudian Let's add that config suggested by Filip then?

agudian: I think that enforced an empty line between the package decl and the legal header, but I'll check again.

filiphr: @agudian There is also `IMPORT` maybe this one is the one that makes sure there is only one line.

agudian: `IMPORT` does the trick for a couple of other spots 👍. Unfortunately, it doesn't seem to treat the `import static` properly and this part slips by... 😁

gunnarmorling:
Cool, you'll push another commit for it then and we are good to go?

agudian:
Thanks for the reviews! I've added the `IMPORT` check and will push it after the PRB is done. :)

agudian:
> Cool, you'll push another commit for it then and we are good to go?

I've already rebased and squashed it into the current commit here...

Issue #968, opened 2016-11-26
Formatting generated code property mapping for Collections


sjaakd:
Remove strange spacing and enters for code generated by `SetterWrapperForCollectionsAndMaps`.

sjaakd:
@filiphr thanks for reviewing.

PR #969, opened 2016-11-26
#968 removing strange enters and spaces in formatting.

sjaakd:
Should be an easy one to review :smile:

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.ftl
@@ -43,20 +43,16 @@
 -->
 <#macro callTargetWriteAccessor>
     <@lib.handleNullCheck>
-      <#if directAssignment>
-        ${ext.targetBeanName}.${ext.targetWriteAccessorName}( <@wrapLocalVarInCollectionInitializer/> );
-    <#else>
-        ${ext.targetBeanName}.${ext.targetWriteAccessorName}( ${localVarName} );
-    </#if>
+        ${ext.targetBeanName}.${ext.targetWriteAccessorName}( <#if directAssignment><@wrapLocalVarInCollectionInitializer/><#else>${localVarName}</#if> );
   </@lib.handleNullCheck>
 </#macro>
 <#--
   wraps the local variable in a collection initializer (new collection, or EnumSet.copyOf)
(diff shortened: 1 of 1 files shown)


processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMaps.ftl
```diff
@@ -43,20 +43,16 @@
 -->
 <#macro callTargetWriteAccessor>
     <@lib.handleNullCheck>
-      <#if directAssignment>
-        ${ext.targetBeanName}.${ext.targetWriteAccessorName}( <@wrapLocalVarInCollectionInitializer/> );
-    <#else>
-        ${ext.targetBeanName}.${ext.targetWriteAccessorName}( ${localVarName} );
-    </#if>
```
filiphr: The compress single_line does not work for this one?

sjaakd: probably it does.. only In this case the whole thing fits on one line.. After all its the same call, only the arg differs.

filiphr: I agree that it fits. What I meant is that it is easier to read it and understand when you look at the template only

sjaakd: Hey @filiphr ,

I tried your suggestion.. But there's a side effect. The compress also acts on the call site. So the result is (see Issue289 for instance):

```java
        Collection<TargetElement> collection = sourceElementCollectionToTargetElementCollection( source.getCollection() ); if ( collection != null ) { targetWithSetter.setCollection( collection ); }        
```
In one line. Perhaps an error in Freemarker..

sjaakd: @agudian : this is spotted by Checkstyle on the generated code. However, the current situation (leading up to this Issue / PR) isn't. Can we catch the current situation as well?

sjaakd: I added another issue #969 to detect generated code formatting problems.

PR #972, opened 2016-11-27
#971 Make sure that an Iterable or Map MapingMethod are equal if their source parameters are from the same type

### LINKED ISSUE #971: IterableMappingMethod uses List generated method for a Collection if the list is defined before the collection
Example:

```java
public class Target {
    List<Integer> integers;
    Collection<Integer> integersCollection;

   //The order of the getters is not really important here
}

public class Source {
    List<String> integers;
    List<String> integersCollection;

   //getters for the integers are before the getters of the integersCollection
}

@Mapper
public interface SourceTargetMapper {
    Source targetToSource(Target target);
}
```

The Implementation:
```java

public class CollectionMapperImpl implements CollectionMapper {

    @Override
    public Source targetToSource(Target target) {
        if ( target == null ) {
            return null;
        }

        Source source = new Source();

        List<String> list = integerListToStringList( target.getIntegers() );
        if ( list != null ) {
            source.setStringStream( list );
        }
        List<String> list_ = integerListToStringList( target.getIntegersCollection() );
        if ( list_ != null ) {
            source.setStringCollection( list_ );
        }

        return source;
    }

    protected List<String> integerListToStringList(List<Integer> list) {
        if ( list == null ) {
            return null;
        }

        List<String> list_ = new ArrayList<String>();
        for ( Integer integer : list ) {
            list_.add( String.valueOf( integer ) );
        }

        return list_;
    }
}
```

If the order of the getters is inversed in the `Source` then it works the first forged method is:
```java
protected List<String> integerCollectionToStringList(Collection<Integer> collection) {
        if ( collection == null ) {
            return null;
        }

        for ( Integer integer : collection ) {
            list.add( String.valueOf( integer ) );
        }

        return list;
}
```

And this one works both for the `List<Integer>` and the `Collection<Integer>`.

One solution of this would be to check if the `sourceParameters` in the `IterableMappingMethod` are equal, but this would always generate more specific methods (one for `List`, and another one for `Collection`), or do something more clever.

filiphr:
This one fixes #971. However, now we will generate a Mapping method for each of the types that we are converting for example there will be one method for each of: `Iterable`, `Collection`, `List` to a `List`. Where we could only use one method with a source parameter an `Iterable` that will do the mapping for the `Collection` and the `List` to a `List`.

Maybe we need to do some refactoring and for the `IterableMappingMethod` always use the most generic possible method.

Mapping any `Iterable`, `Collection` and/or `List` to a list in our case is exactly the same no matter what the parameter is.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
@@ -304,6 +304,9 @@ public class IterableMappingMethod extends MappingMethod {
         }
 
         for ( int i = 0; i < getSourceParameters().size(); i++ ) {
+            if ( !getSourceParameters().get( i ).getType().equals( other.getSourceParameters().get( i ).getType() ) ) {
+                return false;
+            }
             List<Type> thisTypeParameters = getSourceParameters().get( i ).getType().getTypeParameters();
             List<Type> otherTypeParameters = other.getSourceParameters().get( i ).getType().getTypeParameters();
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/MapMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/MapMappingMethod.java
@@ -319,8 +319,14 @@ public class MapMappingMethod extends MappingMethod {
         }
 
         for ( int i = 0; i < getSourceParameters().size(); i++ ) {
-            if ( !getSourceParameters().get( i ).getType().getTypeParameters().get( 0 )
-                .equals( other.getSourceParameters().get( i ).getType().getTypeParameters().get( 0 ) ) ) {
+            if ( !getSourceParameters().get( i ).getType().equals( other.getSourceParameters().get( i ).getType() ) ) {
+                return false;
+            }
+
+            List<Type> thisTypeParameters = getSourceParameters().get( i ).getType().getTypeParameters();
+            List<Type> otherTypeParameters = other.getSourceParameters().get( i ).getType().getTypeParameters();
+
+            if ( !thisTypeParameters.equals( otherTypeParameters ) ) {
                 return false;
             }
         }
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_971/CollectionSource.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_971/CollectionSource.java
@@ -1,0 +1,48 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.bugs._971;
+
+import java.util.List;
+
+/**
+ *
+ * @author Filip Hrisafov
+ */
+public class CollectionSource {
+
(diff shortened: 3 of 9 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
```diff
@@ -304,6 +304,9 @@ public boolean equals(Object obj) {
         }
 
         for ( int i = 0; i < getSourceParameters().size(); i++ ) {
+            if ( !getSourceParameters().get( i ).getType().equals( other.getSourceParameters().get( i ) ) ) {
```
sjaakd: I reckon its: `!getSourceParameters().get( i ).getType().equals( other.getSourceParameters().get( i ).getType()`, right?

processor/src/main/java/org/mapstruct/ap/internal/model/MapMappingMethod.java
```diff
@@ -319,6 +319,9 @@ public boolean equals(Object obj) {
         }
 
         for ( int i = 0; i < getSourceParameters().size(); i++ ) {
+            if ( !getSourceParameters().get( i ).getType().equals( other.getSourceParameters().get( i ).getType() ) ) {
+                return false;
+            }
             if ( !getSourceParameters().get( i ).getType().getTypeParameters().get( 0 )
```
sjaakd: You did not touch this.. but looking upon this code, line 325 cannot be correct. For `Iterable` types we decide to check all type parameters. Here only the first..? Feels not logical for maps. Can you copy the sam code here as for the `Iterable` method?

So:
```java
            List<Type> thisTypeParameters = getSourceParameters().get( i ).getType().getTypeParameters();
            List<Type> otherTypeParameters = other.getSourceParameters().get( i ).getType().getTypeParameters();

            if ( !thisTypeParameters.equals( otherTypeParameters ) ) {
                return false;
            }
```

sjaakd: And the sad thing is: I probably wrote this myself.. :sad: ...

sjaakd:
@filiphr .. With the above remarks fixed it should work.. When finished, please go ahead and merge.. 

But: make sure you merge to`master`. You are on `mapstruct:1.1-master`.. @agudian , correct me if I'm wrong.. But I take that after the final l `mapstruct:1.1-master` is dead.

We only want fast-forward merges. Do you know how to do this?

processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
filiphr: you are right. I'll fix it

processor/src/main/java/org/mapstruct/ap/internal/model/MapMappingMethod.java
filiphr: I'll update that part as well

filiphr:
@sjaakd I did it on purpose over the mapstruct:1.1-master, because this is a bug which will make the created code not compilable.

@agudian, @gunnarmorling  what should we do in this case? 

@sjaakd yes I'll do a fast-forward merge it is not a problem.

agudian:
Well we haven't decided, yet if we'll do a 1.1.1 bugfix release, but in case someone finds an issue worth fixing fast,  we should be in a position where we have all bugfixes available there already. So if it's not too much overhead, I'm all in for fixing this kind of bugs with a fast-forward on 1.1-master and then doing a merge-commit onto master for 1.2.
@gunnarmorling, do you have an opinion here?

sjaakd:
Hmm. in that case.. we should do perhaps this little one:https://github.com/mapstruct/mapstruct/commit/488eae2de27f22855ce71c78719cab40f4f161a4 as well.

agudian:
> Hmm. in that case.. we should do perhaps this little one:488eae2 as well.

I'd say only if it makes merging any follow-ups easier for us, as for the user it's just a cosmetic issue...?

sjaakd:
> I'd say only if it makes merging any follow-ups easier for us, as for the user it's just a cosmetic issue...?

ok.. You're right.. its cosmetic.. But anyway annoying that I missed it :smile:..
 
Than I would say.. merge the above via 1.1 to 1.0.. Does not matter whether we make a 1.1.1 or not.. We just open up the possibility and can decide later.. Right?

filiphr:
``Than I would say.. merge the above via 1.1 to 1.0.. Does not matter whether we make a 1.1.1 or not.. We just open up the possibility and can decide later.. Right?

You mean via 1.1 to 1.2, right? Yes it opens the possibility to make 1.1.1 without any problems with all the relevant bug fixes.

I'll merge it into 1.1

Issue #978, opened 2016-11-30
Problem mapping nested byte array to denormalized property


filiphr:
```java
@Mapper
public interface SimpleMapper {

    @Mapping( target = "props.byteArray", source = "byteArray" )
    SourceRoot toSourceRoot(TargetObject targetObject);
}
```

Will create a mapping for the byte array that looks like:

```java
public class SimpleMapperImpl implements SimpleMapper {
    @Override
    public SourceRoot toSourceRoot(TargetObject targetObject) {
        if ( targetObject == null ) {
            return null;
        }

        SourceRoot sourceRoot = new SourceRoot();

        SourceProps props = new SourceProps();
        sourceRoot.setProps( props );

        if ( targetObject.getByteArray() != null ) {
            byte[] props.byteArray = targetObject.getByteArray();
            props.setByteArray( Arrays.copyOf( props.byteArray, props.byteArray.length ) );
        }

        return sourceRoot;
    }
}
```

The `targetProperyName` is directly used from the mapping itself.

sjaakd:
ouch.. that should be simple to fix..  I prefer taking the property name from the target type nowadays..

filiphr:
I am working on it. Even if we use that target type we will have a problem when the mapping is reverse. I'll make a PR later tonight or tomorrow.

filiphr:
Closed via PR #979

Issue #941, opened 2016-10-27
Weird variable names generated for Issue289Test


gunnarmorling:
For some reason some private variables do have three underscores at the end of the name.

filiphr:
A suggestion from @sjaakd in PR #979:

>I was thinking, reading the code.. I see a lot of variable names like string, string_, string__, etc.. Perhaps we should start to use number suffixes.. So string_001, string_002 when we generate save variable names..

My suggestion and @agudian is to not use the underscore and the leading zeros, basically having: `string1, string2`.

Code snippet from @agudian for improving the naming (be more smart when naming the private variables:
```java
// should be named 'children'
        List<NodeDTO> list = nodeListToNodeDTOList( node.getChildren() ); 
        if ( list != null ) {
            nodeDTO.setChildren( list );
        }
// should be named 'attributes'
        List<AttributeDTO> list_ = attributeListToAttributeDTOList( node.getAttributes() );
        if ( list_ != null ) {
            nodeDTO.setAttributes( list_ );
        }
```

In `NodeMapperWithContextImpl`:

```java
protected AttributeDTO attributeToAttributeDTO(Attribute attribute) {
        AttributeDTO target = NodeMapperContext.getInstance( attribute, AttributeDTO.class );
        if ( target != null ) {
            return target;
        }

        if ( attribute == null ) {
            return null;
        }

        AttributeDTO attributeDTO_ = new AttributeDTO();

        NodeMapperContext.setInstance( attribute, attributeDTO_ );

        attributeDTO_.setNode( nodeToNodeDTO( attribute.getNode() ) );
        attributeDTO_.setName( attribute.getName() );
        attributeDTO_.setValue( attribute.getValue() );

        NodeMapperContext.cleanup();

        return attributeDTO_;
    }
```
And the `attributeDTO` is not conflicting.

sjaakd:
Not necessarily with the underscore before the number
Idea here is to make it a bit more easy parseable. You simply look for the last occurrence of an _.. If there is not, you add it followed by a sequence number 1. If there is, you try to parse the number behind it. If no success, you still add _, assuming there was an _ initially. If success, you take the number and increase it & replace the number.

agudian:
```java
    public static String getSaveVariableName(String name, Collection<String> existingVariableNames) {
        name = decapitalize( sanitizeIdentifierName( name ) );

        Set<String> conflictingNames = new HashSet<String>( KEYWORDS );
[...]

PR #979, opened 2016-12-01
#978 make sure that the safe variable name does not contain dots in it

filiphr:
If the name contains dots, then split it and join and camelize the parts

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
@@ -1,0 +1,99 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.util;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import org.junit.Test;
+
+/**
+ * @author Filip Hrisafov
+ */
+public class StringsTest {
+    @Test
+    public void testCapitalize() throws Exception {
+        assertThat( Strings.capitalize( null ) ).isNull();
+        assertThat( Strings.capitalize( "c" ) ).isEqualTo( "C" );
+        assertThat( Strings.capitalize( "capitalize" ) ).isEqualTo( "Capitalize" );
+        assertThat( Strings.capitalize( "AlreadyCapitalized" ) ).isEqualTo( "AlreadyCapitalized" );
+        assertThat( Strings.capitalize( "notCapitalized" ) ).isEqualTo( "NotCapitalized" );
+    }
+
+    @Test
+    public void testDecapitalize() throws Exception {
+        assertThat( Strings.decapitalize( null ) ).isNull();
+        assertThat( Strings.decapitalize( "c" ) ).isEqualTo( "c" );
+        assertThat( Strings.decapitalize( "capitalize" ) ).isEqualTo( "capitalize" );
+        assertThat( Strings.decapitalize( "AlreadyCapitalized" ) ).isEqualTo( "alreadyCapitalized" );
+        assertThat( Strings.decapitalize( "notCapitalized" ) ).isEqualTo( "notCapitalized" );
+    }
+
+    @Test
+    public void testJoin() throws Exception {
+        assertThat( Strings.join( new ArrayList<String>(), "-" ) ).isEqualTo( "" );
+        assertThat( Strings.join( Arrays.asList( "Hello", "World" ), "-" ) ).isEqualTo( "Hello-World" );
+        assertThat( Strings.join( Arrays.asList( "Hello" ), "-" ) ).isEqualTo( "Hello" );
+    }
+
+    @Test
+    public void testJoinAndCamelize() throws Exception {
+        assertThat( Strings.joinAndCamelize( new ArrayList<String>() ) ).isEqualTo( "" );
+        assertThat( Strings.joinAndCamelize( Arrays.asList( "Hello", "World" ) ) ).isEqualTo( "HelloWorld" );
+        assertThat( Strings.joinAndCamelize( Arrays.asList( "Hello", "world" ) ) ).isEqualTo( "HelloWorld" );
+        assertThat( Strings.joinAndCamelize( Arrays.asList( "hello", "world" ) ) ).isEqualTo( "helloWorld" );
+    }
+
+    @Test
+    public void testIsEmpty() throws Exception {
+        assertThat( Strings.isEmpty( null ) ).isTrue();
+        assertThat( Strings.isEmpty( "" ) ).isTrue();
+        assertThat( Strings.isEmpty( " " ) ).isFalse();
+        assertThat( Strings.isEmpty( "not empty" ) ).isFalse();
+    }
+
+    @Test
+    public void testGetSaveVariableNameWithArrayExistingVariables() throws Exception {
+        assertThat( Strings.getSaveVariableName( "int[]" ) ).isEqualTo( "intArray" );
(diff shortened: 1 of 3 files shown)


processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import org.junit.Test;
+
+/**
+ * @author Filip Hrisafov
+ */
+public class StringsTest {
```
agudian: 👍

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
```diff
@@ -0,0 +1,99 @@
...
+        assertThat( Strings.joinAndCamelize( Arrays.asList( "Hello", "world" ) ) ).isEqualTo( "HelloWorld" );
+        assertThat( Strings.joinAndCamelize( Arrays.asList( "hello", "world" ) ) ).isEqualTo( "helloWorld" );
+    }
+
+    @Test
+    public void testIsEmpty() throws Exception {
+        assertThat( Strings.isEmpty( null ) ).isTrue();
+        assertThat( Strings.isEmpty( "" ) ).isTrue();
+        assertThat( Strings.isEmpty( "not empty" ) ).isFalse();
+
```
agudian: `assertThat( Strings.isEmpty( " " ) ).isTrue();` ?

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Test
+    public void testGetSaveVariableNameWithCollection() throws Exception {
+        assertThat( Strings.getSaveVariableName( "int[]", new ArrayList<String>() ) ).isEqualTo( "intArray" );
+        assertThat( Strings.getSaveVariableName( "Extends", new ArrayList<String>() ) ).isEqualTo( "extends_" );
+        assertThat( Strings.getSaveVariableName( "prop", Arrays.asList( "prop", "prop_" ) ) ).isEqualTo( "prop__" );
+        assertThat( Strings.getSaveVariableName( "prop.font", Arrays.asList( "propFont", "propFont_" ) ) )
+            .isEqualTo( "propFont__" );
+
+
+    }
```
agudian: hmm, two empty lines... didn't they get reported during the build?

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
```diff
@@ -0,0 +1,99 @@
...
+    public void testGetSaveVariableNameWithArrayExistingVariables() throws Exception {
+        assertThat( Strings.getSaveVariableName( "int[]" ) ).isEqualTo( "intArray" );
+        assertThat( Strings.getSaveVariableName( "Extends" ) ).isEqualTo( "extends_" );
+        assertThat( Strings.getSaveVariableName( "prop", "prop", "prop_" ) ).isEqualTo( "prop__" );
+    }
+
+    @Test
+    public void testGetSaveVariableNameWithCollection() throws Exception {
+        assertThat( Strings.getSaveVariableName( "int[]", new ArrayList<String>() ) ).isEqualTo( "intArray" );
+        assertThat( Strings.getSaveVariableName( "Extends", new ArrayList<String>() ) ).isEqualTo( "extends_" );
```
agudian: Perhaps add another explicit example of passing a keyword, e.g. `"class"`?

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
agudian: Ah, branch 1.1... okay.

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
filiphr: Well this actually is not empty in our implementation. We do a `null` check and then `String#isEmpty()`

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
filiphr: Yes I went for this branch, as this can be a big problem for 1.1. There is no nice workaround for this

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
filiphr: Do we need to change our implementation for this?

agudian:
Great! 👍 Do you squash (at least the last two commits) and merge yourself?

filiphr:
I squashed everything into 2 commits, one to add the test and another one for the fix. Once the build is done, I will `rebase and merge` from Github

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
```diff
@@ -0,0 +1,99 @@
...
+        assertThat( Strings.getSaveVariableName( "Class" ) ).isEqualTo( "class_" );
+        assertThat( Strings.getSaveVariableName( "Case" ) ).isEqualTo( "case_" );
+        assertThat( Strings.getSaveVariableName( "Synchronized" ) ).isEqualTo( "synchronized_" );
+        assertThat( Strings.getSaveVariableName( "prop", "prop", "prop_" ) ).isEqualTo( "prop__" );
+    }
+
+    @Test
+    public void testGetSaveVariableNameWithCollection() throws Exception {
+        assertThat( Strings.getSaveVariableName( "int[]", new ArrayList<String>() ) ).isEqualTo( "intArray" );
+        assertThat( Strings.getSaveVariableName( "Extends", new ArrayList<String>() ) ).isEqualTo( "extends_" );
```
sjaakd: I was thinking, reading the code.. I see a lot of variable names like string, string_, string__, etc..  Perhaps we should start to use number suffixes.. So string_001, string_002 when we generate save variable names.. WDYT?

filiphr: I agree with you. I think number prefixes are better on the eye. However, I think that they should be `string1, string2, ..., string10` (camelCase), instead of `string_001` (snake_case).

agudian: That would be nice! Doesn't have to be now, but I'd like it. Not necessarily with the underscore before the number, and maybe leave out the leading zeros (implementation should be the same effort).
But there are also some spots where we could derive a more meaningful default name as well, e.g. for result beans in methods that return `List<NodeDTO>` we simply use `list` which might then be conflicting already. Same for some of the setter wrappers, e.g.:
```java
// should be named 'children'
        List<NodeDTO> list = nodeListToNodeDTOList( node.getChildren() ); 
        if ( list != null ) {
            nodeDTO.setChildren( list );
        }
// should be named 'attributes'
        List<AttributeDTO> list_ = attributeListToAttributeDTOList( node.getAttributes() );
        if ( list_ != null ) {
            nodeDTO.setAttributes( list_ );
        }
```

filiphr:
Jenkins it is ok to test

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
agudian: And in `NodeMapperWithContextImpl`, you might have noticed this variable as well:
```
   protected AttributeDTO attributeToAttributeDTO(Attribute attribute) {
        AttributeDTO target = NodeMapperContext.getInstance( attribute, AttributeDTO.class );
        if ( target != null ) {
            return target;
        }

        if ( attribute == null ) {
            return null;
        }

        AttributeDTO attributeDTO_ = new AttributeDTO();

        NodeMapperContext.setInstance( attribute, attributeDTO_ );

        attributeDTO_.setNode( nodeToNodeDTO( attribute.getNode() ) );
        attributeDTO_.setName( attribute.getName() );
        attributeDTO_.setValue( attribute.getValue() );

        NodeMapperContext.cleanup();

        return attributeDTO_;
    }
```
there is no conflict for `attribteDTO` anywhere...

sjaakd:
> here is no conflict for attribteDTO anywhere...

hmm.. I think Gunnar opened up an issue he spotted in the 289 bug mapper generated code.

agudian:
> hmm.. I think Gunnar opened up an issue he spotted in the 289 bug mapper generated code.

👍

agudian:
retest this, please

filiphr:
> hmm.. I think Gunnar opened up an issue he spotted in the 289 bug mapper generated code.

It is issue #941

sjaakd:
> Not necessarily with the underscore before the number

Idea here is to make it a bit more easy parseable. You simply look for the last occurrence of an `_`.. If there is not, you add it followed by a sequence number 1. If there is, you try to parse the number behind it. If no success, you still add `_`, assuming there was an `_` initially. If success, you take the number and increase it & replace the number.

filiphr:
@sjaakd I think that we can continue the discussion for the names in #941, if it is OK for you.

Can I merge this PR?

sjaakd:
> @sjaakd I think that we can continue the discussion for the names in #941, if it is OK for you.
> 
> Can I merge this PR?

eh.. yes.. good point. Of course. I'll move my comment to the #941.

PR #982, opened 2016-12-03
#557 Add documentation about using fields as read/write accessors

filiphr:
This is a separate PR for the documentation for the new field mappings ability introduced in #928

### DIFF
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -358,7 +358,6 @@ Collection-typed attributes with the same element type will be copied by creatin
 
 MapStruct takes all public properties of the source and target types into account. This includes properties declared on super-types.
 
-
 [[adding-custom-methods]]
 === Adding custom methods to mappers
 
@@ -515,7 +514,7 @@ considered as a write accessor.
 
 Small example:
 
-.Mappings with direct field access
+.Examples classes for mapping
 ====
 [source, java, linenums]
 [subs="verbatim,attributes"]
@@ -571,7 +570,7 @@ public interface CustomerMapper {
 
 For the configuration from above, the generated mapper looks like:
 
-.Generated mapper for mappings with direct field access
+.Generated mapper for example classes
 ====
 [source, java, linenums]
 [subs="verbatim,attributes"]
@@ -653,7 +652,6 @@ public class CustomerMapperImpl implements CustomerMapper {
 }
 ----
 ====
-
 
 You can find the complete example in the
 https://github.com/mapstruct/mapstruct-examples/tree/master/mapstruct-field-mapping[mapstruct-examples-field-mapping]
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -580,6 +580,12 @@ You can find the complete example in the
 https://github.com/mapstruct/mapstruct-examples/tree/master/mapstruct-field-mapping[mapstruct-examples-field-mapping]
 project on GitHub.
 
+[WARNING]
+====
+Using direct field mappings is considered experimental as of the 1.2.0.CR1 release. If you have any issues with the
+generated code please raise an issue in our https://github.com/mapstruct/mapstruct/issues[issue tracker].
+====
+
 [[retrieving-mapper]]
 == Retrieving a mapper
 
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -20,13 +20,13 @@ This work is licensed under the http://creativecommons.org/licenses/by-sa/4.0/[C
 
 MapStruct is a Java http://docs.oracle.com/javase/6/docs/technotes/guides/apt/index.html[annotation processor] for the generation of type-safe bean mapping classes.
 
-All you have to do is to define a mapper interface which declares any required mapping methods. During compilation, MapStruct will generate an implementation of this interface. This implementation uses plain Java method invocations for mapping between source and target objects, i.e. no reflection or similar.
+All you have to do is to define a mapper interface which declares any required mapping methods. During compilation, MapStruct will generate an implementation of this interface. This implementation uses plain Java method/field invocations for mapping between source and target objects, i.e. no reflection or similar.
 
 Compared to writing mapping code from hand, MapStruct saves time by generating code which is tedious and error-prone to write. Following a convention over configuration approach, MapStruct uses sensible defaults but steps out of your way when it comes to configuring or implementing special behavior.
 
 Compared to dynamic mapping frameworks, MapStruct offers the following advantages:
 
-* Fast execution by using plain method invocations instead of reflection
+* Fast execution by using plain method/field invocations instead of reflection
 * Compile-time type safety: Only objects and attributes mapping to each other can be mapped, no accidental mapping of an order entity into a customer DTO etc.
 * Clear error-reports at build time, if entities or attributes can't be mapped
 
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -350,14 +357,31 @@ public class CarMapperImpl implements CarMapper {
...
 
 As the example shows the generated code takes into account any name mappings specified via `@Mapping`. If the type of a mapped attribute is different in source and target entity, MapStruct will either apply an automatic conversion (as e.g. for the _price_ property, see also <<implicit-type-conversions>>) or optionally invoke another mapping method (as e.g. for the _driver_ property, see also <<mapping-object-references>>).
 
 Collection-typed attributes with the same element type will be copied by creating a new instance of the target collection type containing the elements from the source property. For collection-typed attributes with different element types each element will be mapped individually and added to the target collection (see <<mapping-collections>>).
 
 MapStruct takes all public properties of the source and target types into account. This includes properties declared on super-types.
 
+[[field-mappings]]
```
agudian: Let's put this section to the end of chapter 3 (below chapter `[[updating-bean-instances]]`. It's a feature several people have asked for, but I guess many users don't use public field access and therefor we should treat it more as an advanced "hey by the way, we can also do this" thing.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -350,14 +357,31 @@ public class CarMapperImpl implements CarMapper {
...
+=== Mappings via field
```
filiphr: Much better like that 😄, I'll change it

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -350,14 +357,31 @@ public class CarMapperImpl implements CarMapper {
...
+use the fields as read/write accessor if it cannot find suitable getter/setter methods for the property.
+
+A field is considered as a read accessor if it is `public` or `public final`. If a field is `static` it is not
+considered as a read accessors.
+
+A field is considered as a write accessor only if it is `public`. If a field is `final` and/or `static` it is not
+considered as a write accessor.
+
+You can find a complete example in the
+https://github.com/mapstruct/mapstruct-examples/tree/master/mapstruct-field-mapping[mapstruct-examples-field-mapping]
```
filiphr: The examples I wrote have exactly that. There is one bean with public fields, the target bean with getters and setters (This would be the use case that I see people using the most). Will add those to the documentation as well.

filiphr:
Thanks for the review @agudian. I removed all the parts that you mentioned, changed the section name and added code snippets. Do you think that their length is too much? Should I reduce them?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -500,6 +500,163 @@ The generated code of the `updateCarFromDto()` method will update the passed `Ca
...
+.Examples classes for mapping
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+public class Customer {
+
+    private Long id;
+    private String name;
+    private Collection<OrderItem> orderItems;
```
agudian: I guess the order item isn't really necessary to demonstrate the feature here.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -500,6 +500,163 @@ The generated code of the `updateCarFromDto()` method will update the passed `Ca
...
+    public String name;
+    public Long quantity;
+}
+
+@Mapper
+public interface CustomerMapper {
+
+    CustomerMapper MAPPER = Mappers.getMapper( CustomerMapper.class );
+
+    @Mapping(source = "orders", target = "orderItems")
```
agudian: Without the order, you could map `name` to `customerName` or something?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -500,6 +500,163 @@ The generated code of the `updateCarFromDto()` method will update the passed `Ca
...
+
+        Collection<OrderItem> collection = orderItemDtoListToOrderItemCollection( customerDto.orders );
+        if ( collection != null ) {
+            customer.setOrderItems( collection );
+        }
+        customer.setId( customerDto.id );
+        customer.setName( customerDto.name );
+
+        return customer;
+    }
```
agudian: You could abbreviate this to:
```java
// GENERATED CODE
public class CustomerMapperImpl implements CustomerMapper {

    @Override
    public Customer toCustomer(CustomerDto customerDto) {
        // ... 
        customer.setId( customerDto.id );
        customer.setName( customerDto.name );
        // ...
    }

    @Override
    public CustomerDto fromCustomer(Customer customer) {
        // ...
        customerDto.id = customer.getId();
        customerDto.name = customer.getName();
        // ...
    }
}
```

filiphr:
You are right. It is much better now. Thanks

agudian:
👍 go ahead and squash/merge!

filiphr:
@agudian Is the last addition with the "experimental" warning OK for you?

sjaakd:
@filiphr .. is there a reason to mark this as `experimental`? I think the beta upcoming beta release of 1.2 already expresses this.

Which triggers me.. perhaps we should have some basic guidelines what we regard as experimental. Up to now (IMHO) these are features for which we're not sure about their permanent character. Typically we discuss long on those. For myself, this is not a debatable one.

filiphr:
@sjaakd only because it is new. However, you are right the beta release expresses it and it will be in the release notes. I will remove that from there.

>Which triggers me.. perhaps we should have some basic guidelines what we regard as experimental. Up to now (IMHO) these are features for which we're not sure about their permanent character. Typically we discuss long on those. For myself, this is not a debatable one.

Setting up some basic guidelines would be great.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -500,6 +500,86 @@ The generated code of the `updateCarFromDto()` method will update the passed `Ca
...
 Collection- or map-typed properties of the target bean to be updated will be cleared and then populated with the values from the corresponding source collection or map.
 
+[[direct-field-mappings]]
+=== Mappings with direct field access
+
+MapStruct also supports mappings of `public` fields that have no getters/setters. MapStruct will
+use the fields as read/write accessor if it cannot find suitable getter/setter methods for the property.
+
+A field is considered as a read accessor if it is `public` or `public final`. If a field is `static` it is not
+considered as a read accessors.
```
gunnarmorling: accessors=accessor

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -500,6 +500,86 @@ The generated code of the `updateCarFromDto()` method will update the passed `Ca
 
 Collection- or map-typed properties of the target bean to be updated will be cleared and then populated with the values from the corresponding source collection or map.
 
+[[direct-field-mappings]]
+=== Mappings with direct field access
+
+MapStruct also supports mappings of `public` fields that have no getters/setters. MapStruct will
```
gunnarmorling: Does it support package visible accessors, too?

gunnarmorling:
Concering the "experimental" status, the idea is to use it in a final release (i.e. non-Beta) if the feature still may change as we need more feedback and real-world usage feedback.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
filiphr: No it doesn't it only accepts `public` visible accessors. I think that we agreed adding the ability to use `package`/`protected` accessors as part of #572?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
filiphr: We missed that one. Should I just push to master with this change, or create a new PR? This PR is already merged 😄

gunnarmorling:
Just push an update to master. Thanks! Sorry for commenting on the closed
PR, I'm a bit late to the party :)

gunnarmorling:
>  I think that we agreed adding the ability to use package/protected accessors
as part of #572 <https://github.com/mapstruct/mapstruct/issues/572>?

Ah, got it. That's cool.

filiphr:
>Just push an update to master. Thanks! Sorry for commenting on the closed
PR, I'm a bit late to the party :)

No problem. It is not nice to have typos and other problems. Fixed it in f949902

Issue #973, opened 2016-11-28
Bring `SetterWrapper` in line with new `SourceRHS`


sjaakd:

sjaakd:
fixed.

PR #984, opened 2016-12-04
#981 moving SourceReference out of PropertyMapping

### LINKED ISSUE #981: Bring SourceReference outside property mapping
And in line with TargetReference. As a consequence: when a mapping method is nested, its always assigned to a local variable, simplifying the handling quite a lot. This is needed as step up to #973.

sjaakd:
Following PR is general refactoring. I want to have this in place before I refactor the standard property mapping (see #973 ). 

Actually I wanted to move the `SourceReference` in outside of `PropertyMapping` to `BeanMappng` and use `SourceRHS` to construct `PropertyMapping`. Note `TargetReference` is also not used inside `PropertyMapping`. But I'm not there yet, so I can't close #981. I also don't want to make this PR too big. So lets take it step by step. This means that `getSourceRHS` and referred methods also moves to `BeanMapping`.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/MappingBuilderContext.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/MappingBuilderContext.java
@@ -75,7 +75,6 @@ public class MappingBuilderContext {
          * returns a parameter assignment
          *
          * @param mappingMethod target mapping method
-         * @param mappedElement used for error messages
          * @param targetType return type to match
          * @param targetPropertyName name of the target property
          * @param formattingParameters used for formatting dates and numbers
@@ -91,9 +90,8 @@ public class MappingBuilderContext {
          * <li>null, no assignment found</li>
          * </ol>
          */
-        @SuppressWarnings("checkstyle:parameternumber")
-        Assignment getTargetAssignment(Method mappingMethod, String mappedElement, Type targetType,
-                                       String targetPropertyName, FormattingParameters formattingParameters,
+        Assignment getTargetAssignment(Method mappingMethod, Type targetType, String targetPropertyName,
+                                       FormattingParameters formattingParameters,
                                        SelectionParameters selectionParameters, SourceRHS sourceRHS,
                                        boolean preferUpdateMethods);
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -203,6 +203,7 @@ public class PropertyMapping extends ModelElement {
         // initial properties
         private String defaultValue;
         private SourceReference sourceReference;
+        private SourceRHS rightHandSide;
         private FormattingParameters formattingParameters;
         private SelectionParameters selectionParameters;
 
@@ -228,14 +229,10 @@ public class PropertyMapping extends ModelElement {
 
         public PropertyMapping build() {
             // handle source
-            String sourceElement = getSourceElement();
-            Type sourceType = getSourceType();
-            SourceRHS sourceRHS = getSourceRHS();
-            if ( targetWriteAccessorType == TargetWriteAccessorType.ADDER && sourceType.isCollectionType() ) {
-                // handle adder, if source is collection then use iterator element type as source type.
-                sourceType = sourceType.getTypeParameters().get( 0 );
-                sourceRHS = new SourceRHS( sourceRHS.getSourceReference(), sourceType, existingVariableNames );
-            }
+            this.rightHandSide = getSourceRHS( sourceReference );
+            rightHandSide.setUseElementAsSourceTypeForMatching(
+                targetWriteAccessorType == TargetWriteAccessorType.ADDER );
+            Type sourceType = rightHandSide.getSourceType();
 
             // all the tricky cases will be excluded for the time being.
             boolean preferUpdateMethods;
@@ -248,22 +245,21 @@ public class PropertyMapping extends ModelElement {
 
             Assignment assignment = ctx.getMappingResolver().getTargetAssignment(
                 method,
-                sourceElement,
                 targetType,
                 targetPropertyName,
                 formattingParameters,
                 selectionParameters,
-                sourceRHS,
+                rightHandSide,
                 preferUpdateMethods
             );
 
             // No mapping found. Try to forge a mapping
             if ( assignment == null ) {
                 if ( (sourceType.isCollectionType() || sourceType.isArrayType()) && targetType.isIterableType() ) {
-                    assignment = forgeIterableMapping( sourceType, targetType, sourceRHS, method.getExecutable() );
+                    assignment = forgeIterableMapping( sourceType, targetType, rightHandSide, method.getExecutable() );
                 }
                 else if ( sourceType.isMapType() && targetType.isMapType() ) {
-                    assignment = forgeMapMapping( sourceType, targetType, sourceRHS, method.getExecutable() );
+                    assignment = forgeMapMapping( sourceType, targetType, rightHandSide, method.getExecutable() );
                 }
             }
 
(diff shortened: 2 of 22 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
```diff
@@ -122,14 +122,13 @@ public Assignment getTargetAssignment(Method mappingMethod, String mappedElement
...
             mappingMethod,
-            mappedElement,
             dateFormat,
             numberFormat,
             sourceRHS,
             criteria
         );
 
-        return attempt.getTargetAssignment( sourceRHS.getSourceType(), targetType );
+        return attempt.getTargetAssignment( sourceRHS.getSourceTypeForMatching(), targetType );
```
filiphr: I am a bit confused here. Is this the same thing we did before? Was it wrong before, or just the source type was the element when it was an `ADDER`?

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
```diff
@@ -61,7 +61,7 @@ public AdderWrapper( Assignment decoratedAssignment, List<Type> thrownTypesToExc
     public Set<Type> getImportTypes() {
         Set<Type> imported = new HashSet<Type>();
         imported.addAll( super.getImportTypes() );
-        imported.add( getSourceType() );
+        imported.add( getSourceType().getTypeParameters().get( 0 ) );
```
sjaakd: @filiphr an adder calls the add method on the target type. The argument is an element type from a collection. The collection is supposed to be set by a setter.

```java

      @Override
    public void toExistingTarget(Source source, Target target) {
        if ( source == null ) {
            return;
        }

        try {
            if ( source.getPets() != null ) {
                for ( String list : source.getPets() ) {
                    target.addPet( petMapper.toPet( list ) );
                }
            }
        }
        catch ( CatException e ) {
        }
        catch ( DogException e ) {
        }
    }
```

But.. the variable name should follow the element name as well.. I'll fix that as well.

sjaakd: So.. to answer the question.. the import is needed for the local var in the for loop.

processor/src/main/java/org/mapstruct/ap/internal/model/MappingBuilderContext.java
```diff
@@ -92,8 +91,8 @@
          * </ol>
          */
         @SuppressWarnings("checkstyle:parameternumber")
```
sjaakd: ack.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -553,36 +540,24 @@ else if ( propertyEntries.size() == 1 ) {
...
+
             }
         }
 
-        private String getSourcePresenceCheckerRef() {
+        private String getSourcePresenceCheckerRef( SourceReference sourceReference ) {
             String sourcePresenceChecker = null;
             if ( !sourceReference.getPropertyEntries().isEmpty() ) {
                 Parameter sourceParam = sourceReference.getParameter();
+                // TODO is first correct here?? shouldn't it be last since the remainer is checked
```
sjaakd: I have a feeling its correct. It was there, but a reminder for myself. I'll tackle  this in the follow up.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -806,7 +781,7 @@ public JavaExpressionMappingBuilder javaExpression(String javaExpression) {
         }
 
         public PropertyMapping build() {
-            Assignment assignment = new SourceRHS( javaExpression, null, existingVariableNames );
+            Assignment assignment = new SourceRHS( javaExpression, null, existingVariableNames, "" );
```
sjaakd: its what it says it is. Its used to identify the source part of an error message. I could use a better name (struggled with it). Ideas?

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
```diff
@@ -105,9 +105,9 @@ public MappingResolverImpl(FormattingMessager messager, Elements elementUtils, T
 
     @Override
     @SuppressWarnings("checkstyle:parameternumber")
```
sjaakd: Ok.. I'll remove

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
sjaakd: The problem is that the actual `SourceType`  is a collection. The type used for matching is the element. So I want `SourceRHS` to be consistent (for our own understanding, but also whenever we want to make this an templating SPI in the future). In the old solution, the `SourceType` in `SourceRHS` was simply replaced by the type of  the collection element when an ADDER was applicable.

But also here... if you have a better idea, its more than welcome.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
filiphr: Sorry I didn't express myself really well. In case  Why are we using an empty one here and not the `javaExpression`? In the `ConstantMappingBuilder` you have `String sourceErrorMessagePart = "constant '" + constantExpression + "'";` which is then used in the `SourceRHS`

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
filiphr: No the idea is great.

>In the old solution, the SourceType in SourceRHS was simply replaced by the type of the collection element when an ADDER was applicable.

This is what was missing in my understanding. I couldn't figure out how it was before.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
filiphr: Thanks for the explanation. As I wrote in the other comment, I had problem figuring out how it was created before. You explained it well with:

>In the old solution, the SourceType in SourceRHS was simply replaced by the type of the collection element when an ADDER was applicable.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: The only place were this message is used is in `MappingResolver`. This is not called in an expression since there is nothing to resolve. So there is never an error message to print. I guess this comes from generalizing stuff.

To avoid this, I could make an overloaded constructor to `sourceRHS`.. WDYT?

filiphr: You can leave the constructor. If it is there one needs to think about why it is there and it won't be forgotten. Empty is OK for me

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
sjaakd: No problem.. I want to make stuff more compressible. So if I need to add some javadoc or better variable names.. that's the least I can do. Any tip is welcome :)

filiphr: On the getter maybe you can add a Javadoc like: "The source type that needs to be used when looking for a target assignment"

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -203,6 +203,7 @@ protected boolean isFieldAssignment() {
         // initial properties
         private String defaultValue;
         private SourceReference sourceReference;
+        private SourceRHS rhs;
```
filiphr: Yes I saw that somewhere we use `rightHandSide` and now also `rhs`. I noticed this because some method parameters where you pass the `rhs` were renamed from `rightHandSide` to `rhs` 😄. Maybe use everywhere `rhs` or `rightHandSide` and then we are consistent

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
sjaakd: I'll add: `The source type that is to be used when resolving the mapping from source to target.` WDYT? We stay "in terminology" using terms resolving and mapping.

filiphr: I think it is OK

filiphr:
Once you update the JavaDoc of the getter it is OK for me.

Issue #975, opened 2016-11-28
Open a way to pass state information to subsequent mappings


sjaakd:
E.g. by a parameter and (new) annotation like: `@PassAsState` or something like that.

agudian:
How about `@MappingContext`? That's what we pondered with in the issue and PR regarding the handling of cycles in graph structures. Then that would require finishing up PR #911 (handling return values from before/after mapping methods) and allowing this state object to be passed in property mapping methods and those before/after methods.

sjaakd:
`@MappingContext` is perhaps a bit confusing. Its sole purpose is to let (generated)mapping coe ignore it and just happily pass it along. Perhaps.... `@ContextParameter`

Just a thought we should make it an annotation to use directly or which can act as a marker to stick to other annotations.

So:

`A map(B param1, @ContextParameter1 C param2, @ContextParameter2 D param3)` that could call  method which signature: `A1 map(B1 param1, @ContextParameter1 C param2, @ContextParameter2 D param3)` this way:

```java
A map(B param1, @ContextParameter1 C param2, @ContextParameter2 D param3) {

   A a = new A();
   a.set1( map( param1, param2, param3 )  );
   ..
   return a;

}

@Target(ElementType.METHOD)
@Retention(RetentionPolicy.CLASS)
@ContextParameter
public @interface ContextParameter1 {
}

@Target(ElementType.METHOD)
@Retention(RetentionPolicy.CLASS)
@ContextParameter
public @interface ContextParameter2 {
}

```

WDYT?

agudian:
A meta-annotation might just be a bit over-engineered here, IMHO. Using just one annotation should do the trick as well and keeps the mappers readable.
I guess we should take some time to come up with a good name, as obviously you found `@MappingContext` confusing and I think the *Parameter* in `@ContextParameter` is a bit stating the obvious (like annotating a constructor with `@Constructor` 😜).

sjaakd:
> A meta-annotation might just be a bit over-engineered here, IMHO

That might be true :). However, I propose to have an initial annotation which might double as meta annotation in the future. The only thing needed, for now, would be specifying the proper `@Target`.. 

As for the name.. We could decide it to keep it as simple as `@Context`..

agudian:
I'm gonna start working on this.

[...]

PR #986, opened 2016-12-05
Add documentation about cycle avoidance

filiphr:
This PR is to extend the documentation about the possibility to avoid cycles with MapStruct.

I think that we need to have something in our documentation that will show how one can make a mapper that can be used for avoiding cycles during mapping, especially because this feature is not out of the box and the user needs to add a Custom mapper.

Do you think that it is clear enough? Is the example too big?

### DIFF
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -1751,6 +1751,69 @@ public CarDto carToCarDto(Car car) throws GearException {
 
 Some **notes** on null checks. MapStruct does provide null checking only when required: when applying type-conversions or constructing a new type by invoking its constructor. This means that the user is responsible in hand-written code for returning valid non-null objects. Also null objects can be handed to hand-written code, since MapStruct does not want to make assumptions on the meaning assigned by the user to a null object. Hand-written code has to deal with this.
 
+[[avoiding-object-cycles]]
+=== Avoiding object cycles
+
+Using a combination of `@BeforeMapping` and `@AfterMapping` it is possible to write a generic Mapper which is able to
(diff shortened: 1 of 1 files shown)


sjaakd:
Hey Filip,

First of all: good initiative. Although, I should add, I'm not sure whether we should add this to the documentation. Personally I think we should keep the documentation sec of the description of the API.

I was discussing this kind of stuff with @gunnarmorling , I think somewhere 8 or 9 months go. We then decided we should have a blog. That blog should be about how to solve typical mapping problems with MapStruct. I'm not sure whether a blog is the good form, but I still think its a good idea to have a kind of collection of 
1. Problems
2. How to solve them with MapStruct.

This would be a very nice candidate. 

Often I'm surprised if I see people working with MapStruct how many times they favor for instance expressions over a hand-written mapping method. Also there a basic blog/ example would help.

WDYT? @gunnarmorling , @agudian .. Opinions?

filiphr:
@sjaakd I agree with you. This is not entirely part of our API. However, a lot of people are looking for things like this. We don't have to add it to the documentation, but it is one step towards adding it somewhere 😄 

When you mentioned blog I thought about something. We can have blog post where we go deeper into problems that MapStruct can solve (like the cycle mapping). However, I think that an F&Q would also be nice. In there we can have questions which are asked most frequently and how to solve typical problems with MapStruct. We can have a small concise answer with a link to a more detailed blog post.

agudian:
A blog post for mapstruct.org would sounds like a good idea. My secret hope is that we can implement #975 in time for the first or second beta, which would allow for a much nicer solution without thread-locals, depth counters and clean-ups... 🤓

gunnarmorling:
Closing this one as it's obsolete with the new context parameters.

PR #989, opened 2016-12-11
#973 refactoring setter wrapper and update wrapper

sjaakd:


sjaakd:
There are strange null checks in the calling the update methods when using expressions / constants. Need to check that.

processor/src/main/resources/org/mapstruct/ap/internal/model/macro/CommonMacros.ftl
```diff
@@ -93,15 +95,45 @@ Performs a default assignment with a default value.
...
 -->
-<#macro handleWriteAccesing>
-    <#t><#if fieldAssignment><#else>()</#if>
-</#macro>
\ No newline at end of file
+<#macro handleWriteAccesing><#if fieldAssignment><#else>()</#if></#macro>
+<#--
+  macro: initTargetObject
+
+  purpose: To factoryze or construct a new target object
```
gunnarmorling: factoryze=factorize

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapper.java
```diff
@@ -22,19 +22,41 @@
...
 public class SetterWrapper extends AssignmentWrapper {
 
     private final List<Type> thrownTypesToExclude;
+    private final boolean includeNullCheck;
+
+    public SetterWrapper(Assignment rhs, List<Type> thrownTypesToExclude, NullValueCheckStrategyPrism nvms,
+        boolean fieldAssignment, Type targetType) {
+        super( rhs, fieldAssignment );
+        this.thrownTypesToExclude = thrownTypesToExclude;
+        this.includeNullCheck = !rhs.getSourceReference().equals( rhs.getSourceParameterName() )
```
filiphr: I think that the checks should go in some method and maybe somehow add the list items that are there in the class Javadoc.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/GetterWrapperForCollectionsAndMaps.java
```diff
@@ -41,20 +40,17 @@
     /**
      * @param decoratedAssignment
      * @param thrownTypesToExclude
-     * @param existingVariableNames
      * @param targetType
      * @param fieldAssignment
      */
     public GetterWrapperForCollectionsAndMaps(Assignment decoratedAssignment,
```
filiphr: I noticed that in the `AdderWrapper` and `SetterWrapper` you renamed the `decoratedAssignment` to `rhs`. Should we do the renaming everywhere (the other `AssignmentWrapper`(s), and the `AssignmentWrapper` itself), or it does not make sense?

processor/src/main/resources/org/mapstruct/ap/internal/model/macro/CommonMacros.ftl
sjaakd: will fix.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapper.java
sjaakd: will fix.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/GetterWrapperForCollectionsAndMaps.java
filiphr: OK no problem. Completely agree with your reasoning.

sjaakd:
oops

Issue #621, opened 2015-08-13
@Mapping should support mutiple sources


gbrehmer:
Currently multiple sources are only supported by single map method. To use those mapper within other map methods the ```@Mapping``` annotation should support mutiple ```source``` parameters.
Current Workaround is to use java expression.

Example of mutiple sources:
```
@Mappings({@Mapping(source={"e.inner.prop.b", "e.inner.prop.c"}, target="bdto")}
ADto map(EntityA e);

BDto map(EntityB b,EntityC c);
```

mozinrat:
@gbrehmer please share you you achieve this by java expression.
I have a similar requirement 

@Mappings({ @Mapping(target = "guests", source = "hotelGuests")}) --> list to list mapping
    RoomInfoEntity mapToRoomInfo(RoomInfo roomInfo);

    @Mappings({@Mapping(target = "roomtypeName", source = "roomTypeName"),
            @Mapping(target = "rateplanName", source = "ratePlanName"),
            @Mapping(target = "currency", source = "price.currency"),
            @Mapping(target = "facilities", expression = "java( roomInfo.getFacilities().toString() )"),})
    RoomInfoEntity mapToRoomDetails(RoomDetails roomDetails);

    GuestEntity mapHotelGuest(HotelGuest guest);

gbrehmer:
@mozinrat 
my problem only occur, if you need multiple sources for a single mapping. Your example doesn't contain such a case, only one-to-one mappings. It seems that you use the wrong variable, because "roomInfo" doesn't match a method parameter name.

In my case I only have to use the generated mapper-method from java expression directly. E.g. expression = java(com.example.Mapper.INSTANCE.mapEntity(param.getSub().getVar1(),param.getSub().getVar2()))

Mapper.INSTANCE is a static method inside Mapper-Interface class to get generated Mapper instance (see official documentation)

sjaakd:
@gbrehmer :I've tried update methods, and stumbled on an interesting point..

```java
@Mapper
public interface MyMapper {

    @Mappings({
//        @Mapping(source = "e.inner.prop.b", target = "bdto"),
        @Mapping(source = "e.inner.prop.c", target = "bdto")
    })
    ADto map(EntityA e, @MappingTarget ADto a);

    void map(EntityB in, @MappingTarget BDto out);
    void map(EntityC in, @MappingTarget BDto out);
}
```
Check out the commented line.

[...]

PR #994, opened 2016-12-16
#975 Refactor method-matching…

agudian:
… to unify selection of mapping method, factories and lifecycle methods.

This is a step towards implementing #975, but in the light of keeping PRs as small as possible, this part can be reviewed and merged on its own.

The basic idea is to move what's in `ParameterAssignmentUtil` to the core method selection mechanism (`TypeSelector`, to be precise) and thus remove the need for the special selection handling for factory methods and lifecycle methods. So I deleted slightly more lines than I added... 😁 

Plus, the selection mechanism now also returns the parameter bindings that are used to determine how the parameters of the method to call are populated (for factory and lifecycle methods, that could be a source parameter of the mapping method).

I guess the `LifecycleCallbackMethodReference` and `MethodReference` can also be refactored to make `LifecycleCallbackMethodReference` extend `MethodReference` (or perhaps be replaced by). I'll look after that when I continue with #975.


sjaakd:
> but in the light of keeping PRs as small as possible,

Small is a relative concept :smile: .. I started a review, but I'll come back on it later this weekend..

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -34,17 +37,154 @@
...
+                                                                          List<Type> sourceTypes, Type targetType,
+                                                                          SelectionCriteria criteria) {
+
+        if ( methods.isEmpty() ) {
+            return methods;
+        }
+
+        List<MethodSelection<T>> result = new ArrayList<MethodSelection<T>>();
+
+        List<Parameter> availableParams;
```
sjaakd: I'm trying to follow what is happening here.. Could you add some comments..? What is relevant for factories? What is relevant for lifecycle methods? Or is this going to change when `Context` is added?

processor/src/main/java/org/mapstruct/ap/internal/model/LifecycleCallbackMethodReference.java
```diff
@@ -109,12 +91,8 @@ public String getTargetVariableName() {
         return declaringType != null ? Collections.asSet( declaringType ) : java.util.Collections.<Type> emptySet();
     }
 
-    public List<Parameter> getParameterAssignments() {
-        return parameterAssignments;
-    }
-
     public boolean hasMappingTargetParameter() {
-        for ( Parameter param : parameterAssignments ) {
+        for ( Parameter param : getParameterBindings() ) {
```
agudian: They represent what's bound to parameters of the method reference. Perhaps I should introduce a new class for that, e.g. `ParameterBinding` to differentiate it more clearly...

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
```diff
@@ -107,4 +101,27 @@ public boolean equals(Object obj) {
         return true;
     }
 
+    /**
+     * @param classTypeOf the type representing {@code Class<X>} for the target type {@code X}
+     * @return a {@link Parameter} usable for parameter bindings
+     */
+    public static Parameter forTargetTypeBinding(Type classTypeOf) {
+        return new Parameter( null, classTypeOf, false, true );
```
agudian: Yea.. one more reason to create a new class for that. What's better, `ParameterBinding` or `MethodReferenceArgument`?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/MethodSelection.java
```diff
@@ -16,28 +16,40 @@
...
+package org.mapstruct.ap.internal.model.source.selector;
 
 import java.util.List;
 
 import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.source.Method;
 
 /**
- * Represents a reference to a factory method.
+ * A selected method with additional metadata that might be required for further usage of the selected method.
```
agudian: Right

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/MethodSelection.java
```diff
@@ -16,28 +16,40 @@
...
-    private final List<Parameter> parameterAssignments;
+    public MethodSelection(T method) {
+        this.method = method;
+    }
+
+    public T getMethod() {
+        return method;
+    }
+
+    public List<Parameter> getParameterBindings() {
```
agudian: That holds the information about what parameters/arguments were successfully matched for the method call. When rendering the method call in the templates, that information is required to know what to put where (e.g. a source parameter of the mapping method, the target type class literal, the mapping target, or an assignment -- or later on any mapping context parameters).

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
agudian: Almost nothing changes here with the context parameters... 😁. But you're right, I'll add some more comments.

sjaakd: Thanks.

agudian:
I added another commit to add the `ParameterBinding` type and also work in the other suggestions you gave me.
Narrating all the changes in `TypeSelector` using comments is a bit tricky, but I tried to add some javadoc here and there and make the method names a bit more speaking (before they were basically copied from their old location). The method-matching of the lifecycle methods and the newly added factory methods with arguments used a similar logic for determining possible argument bindings. 
For the normal mapping method selection path, that was not really needed because `MethodMatcher` did some very specific checks directly on the source parameters and possible mapping target parameters. That's all gone now and handled generically for all parameters. To that end, the `TypeSelector` doesn't only pass through the `sourceTypes`, but the types for all method arguments that have been determined by `getCandidateParameterBindingPermutations`. Each of the possible permutations of the available parameter bindings is then checked with the `MethodMatcher` for type-assignability.

processor/src/main/java/org/mapstruct/ap/internal/model/common/ParameterBinding.java
```diff
@@ -0,0 +1,124 @@
...
+        this.type = parameterType;
+        this.variableName = variableName;
+        this.targetType = targetType;
+        this.mappingTarget = mappingTarget;
+    }
+
+    /**
+     * @return the name of the variable (or parameter) that is being used as argument for the parameter being bound.
+     */
+    public String getVariableName() {
```
sjaakd: This is never used, right?

sjaakd: Forget my last remark. It's used in the templates (of course).

sjaakd: But (and perhaps we should tackle this later).. If a localvar binds to a 'method to be called parameter', I would somehow expect that relation still to be there. Now the relation is only maintained because of the order in the list. Perhaps the `ParameterBinding` should wrap the original `Parameter`? Alternatively, we could maintain that relation where we use (e.g. method reference) in a map. Or, we should maintain the argument number somehow hard in the `Parameter` and the `ParameterBinding`.  I know there's no hard argument to do so, but I've a nagging feeling we should relate Parameter harder to ParameterBinding.

sjaakd:
@agudian .. I've added some food for thoughts in the comments. I'm not sure whether we should address that here and now. 

This (in general) is a huge step forward (in terms of separation of concerns). I think the selection process is scoped to a limited number of classes with a clear purpose again.
 
Towards the future: I was thinking that the `TypeSelector` and `matches` in `Method` share similar functionality. Responsibility of both classes feels a bit mixed. Perhaps that should be revisited some day. But all-in-all I think this is in good shape.

So: go ahead and merge.

-----------------------

Another thing that spang to my mind was: with this multiple sources, how far are we from creating one target object out of 2 or more sources in the property mapping? I saw some comments on an old issue coming by. I (myself) still have the following use case:

source
Date source.timeX
String source.timeXFormat
Date source.timeY
String source.timeYFormat


target
MyDate target.timeX
MyDate target.timeY


`(timeX, timeXFormat) -> timeX`  is the same mapping as `(timeY, timeYFormat) -> timeY`. It would be ideal if I could do this with multiple sources mapping into one target. Currently, to achieve this,  I pass the same source through three mapping methods: 
1. one for all other properties and maps source itself for timeX and timeY
2. one that maps specifically timeX and timeXFormat to timeX and ignores all other
3. one that maps specifically timeY and timeYFormat to timeYand ignores all other

(I'll add an example now I think of this to the examples repo, because it shows how verbose it becomes).

agudian:
> Towards the future: I was thinking that the `TypeSelector` and `matches` in `Method` share similar functionality. Responsibility of both classes feels a bit mixed. Perhaps that should be revisited some day. But all-in-all I think this is in good shape.

Right. Today, there's still the the impediment that `BuiltInMethod` has its own simplified implementation of that method. For that to be aligned, the BuiltInMethods would need to provide their correct `Parameter` list - now they only provide one source parameter there, leaving out the potential conversion-context parameter (e.g. for the date-format String literal). That's also part of the reason why I didn't yet make the `ParameterBinding` a `ModelElement` with its own template... Well that and the missing representation for the `assignment` to be added there.
So there's still room for improvement there. And that's also what we would need to revisit when adding support for multiple source parameters in the future. But at least the `MethodSelectors` stuff should now be ready for that as well.

I'll go ahead and squash these commits, but I wait with the final rebase / merge until the Automapping PR is merged. In the meantime I'll continue with the `@Context` stuff based on this state here... 😄 

Thanks for your thorough review, Sjaak!

sjaakd:
Nice work.!

filiphr:
I agree nice refactoring 😄.

By the way @sjaakd would you create an issue for your second part of [this comment](https://github.com/mapstruct/mapstruct/pull/994#issuecomment-267812320)? That will also improve the performance of the generated mappers as well, not just the verbosity 😄

agudian:
Does #621 cover it?

sjaakd:
> Does #621 cover it?

Yep.. And I'm not entirely sure whether we should pursue this 😄

filiphr:
I didn't completely understand the comment and I was thinking on something else.

e.g.

```java
   @Mapping( target = "fontType", source = "font.type"),
   @Mapping( target = "fontSize", source = "font.size"),
```

We create 2 methods for the mappings (one for `font.type`, and another one for `font.size`, whereas we could have done it in one by grouping the `font`.

PR #995, opened 2016-12-17
#557 add Javadoc for the Accessor

filiphr:
I forgot to add the Javadoc to the `Accessor` during the main PR. So this one is just for that one 😄

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
@@ -261,11 +261,13 @@ public class TypeFactory {
      * Get the ExecutableType for given method as part of usedMapper. Possibly parameterized types in method declaration
      * will be evaluated to concrete types then.
      *
+     * <b>IMPORTANT:</b> This should only be used from the Processors. as they are operating over executable elements.
+     * The internals should not be using this function and should not be using the {@link ExecutableElement} directly.
+     *
      * @param includingType the type on which's scope the method type shall be evaluated
      * @param method the method
      * @return the ExecutableType representing the method as part of usedMapper
      */
-    @Deprecated
     public ExecutableType getMethodType(DeclaredType includingType, ExecutableElement method) {
         TypeMirror asMemberOf = typeUtils.asMemberOf( includingType, method );
         return (ExecutableType) asMemberOf;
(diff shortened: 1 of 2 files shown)


agudian:
Oh I also have a tiny request regarding that older commit... you marked some methods as deprecated, but without any documentation on what else to use. Would be cool if you could switch the remaining call sites to the new method and remove the old ones, or document why they are still required but deprecated nonetheless.
I didn't see that spot in the original PR and just stubled across it last week - sorry for the late feedback there.

processor/src/main/java/org/mapstruct/ap/internal/util/accessor/Accessor.java
```diff
@@ -26,17 +26,39 @@
...
 public interface Accessor {
 
+    /**
+     * This returns the type that this accessor gives as a return.
+     *
+     * e.g. The {@link ExecutableElement#getReturnType()} if this is a method accessor,
+     * or {@link javax.lang.model.element.VariableElement#asType()} for field accessors.
+     *
+     * @return the type that the accessor gives as a return
+     */
```
agudian: If the accessor is for writing a property, this could also be the argument type of a setter method?

gunnarmorling:
> you marked some methods as deprecated, but without any documentation on what else to use

+1 Are these public (API/SPI) or internal methods? In case of the latter, we don't need deprecations anyways, just change it (apart from maybe a very short period of time where you may need to push some changes and merge someone else's PR or something like that).

sjaakd:
👍

processor/src/main/java/org/mapstruct/ap/internal/util/accessor/Accessor.java
filiphr: Not really. If it is a setter this will be `Void`. Maybe I need to rename it to `getReturnType`. Internally, there are different mechanics based on the parameters of the `ExecutableElement` that the `Accessor ` returns.

filiphr:
>Are these public (API/SPI) or internal methods?

They are internal ones. The `Accessor` is currently not exposed via the SPI. However, I think that if we want to be more flexible, me might need to use the `Accessor` instead of the `ExecutableElement` in the `AccessorNamingStrategy`. Therefore, I want to document the `Accessor` good.

I will fix the `@Deprecated` and add a commit to the PR. Sorry for not cleaning it up in the initial PR.

sjaakd:
Ok.. go ahead and merge

filiphr:
@sjaakd thanks for the review. @agudian is now better with the `@Deprecated`?

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -261,11 +261,13 @@ else if ( type.isPrimitive() ) {
      * Get the ExecutableType for given method as part of usedMapper. Possibly parameterized types in method declaration
      * will be evaluated to concrete types then.
      *
+     * <b>IMPORTANT:</b> This should only be used from the Processors. as they are operating over executable elements.
```
agudian: Please exchange the first period for a comma... then it's ready to be sqashed and merged 👍

filiphr:
Thanks for the review guys 😄

Issue #657, opened 2015-10-09
"No implementation type" Error Despite @BeanMapping(resultType=...) Being Set to Type with Default Constructor


mbach979:
MapStruct fails in the following case with a "No implementation type is registered for return type" error despite the @BeanMapping(resultType=...) being specified with a class that has an accessible default constructor.

```
interface i1
{
    public String getId();
    public void setId(String id);
}

class i1impl implements i1
{
    private String id;

    public i1impl()
    {
    }

    public String getId()
    {
        return id;
    }

    public void setId(String id)
    {
        this.id = id;
    }
}

class dto
{
    private String id;

    public String getId()
    {
        return id;
    }

    public void setId(String id)
    {
        this.id = id;
    }
}

@Mapper
public interface i1dtoMapper
{
    @BeanMapping(resultType=i1impl.class)
    public i1 dto2i1(dto src);
}
```

Looking at the code it seems that the BeanMappingMethod.build() has code to handle this by setting resultType on the resulting instance, but the validation of the method in MapperCreationProcessor.getMappingMethods() only checks for whether the BeanMappingMethod has a factoryMethod, and not whether the resultType is a concrete class.   

Hacking in an additional condition to look at the result type of the BeanMappingMethod and therefore not throwing the error seems to fix the problem.  The generated mapper correctly instantiates the @BeanMapping(resultType=..) specified class using the default constructor.

agudian:
Looks like you already have some insight here. Would you be willing to turn this into a pull-request including tests and possibly updates to the Javadoc of the annotation?

gunnarmorling:
So far, `@Mapping#resultType()` was only intended as means of narrowing down several matching factory methods. I don't see a problem with using it for the purpose you describe, so if you send in a PR I'd happily apply it.

Of course there is always the work-around of specifying the more specific (implementation) type in the mapping method definition, so it's not a blocker.

domdorn:
any updates here? 
I'm trying to do mapping to a interface and specify the implementation class somehow, but I'm unable 

code looks like this:

    @BeanMapping(resultType = ModifiableOpeningTime.class)
    OpeningTime mapOpeningTime(otherpackage.model.OpeningTime value);

ModifiableOpeningTime is a normal POJO, OpeningTime is a interface

thx

agudian:
@domdorn That shouldn't be too hard to do, but sending in a PR would speed things up... :)
[...]

PR #998, opened 2016-12-19
#657 Using bean mapping result types works when the map return type is not an interface

filiphr:
This PR basically makes sure that if the result type of the `BeanMappingMethod` is not an interface it can be considered that the methods has a "factory" method and therefore we do not need to report an error.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
@@ -365,7 +365,10 @@ public class MapperCreationProcessor implements ModelElementProcessor<List<Sourc
                     .build();
 
                 if ( beanMappingMethod != null ) {
-                    hasFactoryMethod = beanMappingMethod.getFactoryMethod() != null;
+                    // We can consider that the mapping method has a factory method if it has one,
+                    // or if the result type of the methods is not an interface
+                    hasFactoryMethod = beanMappingMethod.getFactoryMethod() != null ||
+                        !beanMappingMethod.getResultType().isInterface();
                     mappingMethods.add( beanMappingMethod );
                 }
             }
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -365,7 +365,10 @@ else if ( method.isEnumMapping() ) {
                     .build();
 
                 if ( beanMappingMethod != null ) {
-                    hasFactoryMethod = beanMappingMethod.getFactoryMethod() != null;
+                    // We can consider that the mapping method has a factory method if it has one,
+                    // or if the result type of the methods is not an interface
+                    hasFactoryMethod = beanMappingMethod.getFactoryMethod() != null ||
+                        !beanMappingMethod.getResultType().isInterface();
```
filiphr: Nothing. It works like it has been working for now. We will generate a Mapper and there will be a compilation problem. The issue with #657 is that in line [376](https://github.com/mapstruct/mapstruct/blob/af68fcf6169ffcbee24a45df1f2ef0796a930857/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java#L376) we explicitly call `reportErrorIfNoImplementationTypeIsRegisteredForInterfaceReturnType` and there we only check the `SourceMethod` and we fail if the return type is Interface and there is no implementation type. 

If we want to make sure that the result type is not an abstract class then we need to change something in the [BeanMappingMethod](https://github.com/mapstruct/mapstruct/blob/af68fcf6169ffcbee24a45df1f2ef0796a930857/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java#L165-L171) and add a check that the result type is not abstract. 

Should I add the check about an abstract methods to this PR as well?

sjaakd: As far as I'm concerned that would be a separate issue, right? Possibly there's a more general scenario in that case.

sjaakd:
I'm ok with this one.

agudian:
👍

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
filiphr: Create #998 for handling the abstract result type scenario

filiphr:
Thanks for the review

Issue #996, opened 2016-12-18
Bring `ArrayCopyWrapper` in line with new `SourceRHS`


sjaakd:
just like: #973

PR #1002, opened 2016-12-19
#996 refactor ArrayCopyWrapper

sjaakd:


agudian:
Looks totally like magic to me 😁 

Awesome how much code you are able to delete here, so 👍! Not that I can make much sense of what's going on at the call-site of the macros, but it's still easier to follow than our many Assignment variants.

All good to go from my side.

sjaakd:
> Not that I can make much sense of what's going on at the call-site of the macros

The common macro's should be understandable. So'll add some more comments/explanations. Anyway, any idea on how to make this more clear is welcome :smile:

There will be more pull requests on this topic :smile:

sjaakd:
I'll squash and merge. Thanks @agudian  for reviewing.

gunnarmorling:
Hi, one thought on protecting against progressions introduced by this kind of changes: How about adding some source files generated by 1.1 and compare the new ones against those?

We could have a set of files exposing most important key features and run this kind of test for them. That'd be a nice thing to have as safety net in the long term (I think Filip also asked about that).

One concern is fragility due to minor changes such as indentation, but then we are actually quite picky about these things anyways, so some tests failing due to somes for instance may actually be a good thing.

Any thoughts?

sjaakd:
> Hi, one thought on protecting against regressions introduced by this kind of changes: How about adding some source files generated by 1.1 and compare the new ones against those?

That's a good idea.

> One concern is fragility due to minor changes such as indentation, but then we are actually quite picky about these things anyways, so some tests failing due to somes for instance may actually be a good thing.

I share your concern. However, we should of leave room for improvements. This makes it at least a conscious improvement.

> Any thoughts? I'd bet @agudian would be up for the challenge for providing some infra for this kind of test :)

@agudian .. Ideas?


For now, I'd like to progress with these improvements on the general premise: less template is better :smile:. If we cannot do this as part of the build in 1.2, we could also do this manually.

sjaakd:
@gunnarmorling do you want to have this in place before I merge this one? Or can I continue?

gunnarmorling:
@sjaakd Given the number of changes I'd feel better if we had it. Let's see what @agudian has to say; I don't think it'd take too long to add some automated comparison-based tests for a few key files.

agudian:
As we already have the `GeneratedSource` test rule available, we'd only have to select those test mappers that we want to check and add the generated files from a 1.1 build to `src/test/resources/..` somewhere. Easy money. 😉 
Question is: what test mappers do we want to pick for this? That's the difficult part.

agudian:
Oh and yes we can do that retro-actively, at some point before the 1.2 release. I wouldn't block this PR for it.

sjaakd:
> Question is: what test mappers do we want to pick for this? That's the difficult part.

Perhaps not. I used the following representative mappers for the refactoring stuff:

* `org.mapstruct.ap.test.bugs._913.Issue913SetterMapperForCollectionsTest`
* `org.mapstruct.ap.test.array`
* `org.mapstruct.ap.test.collection.adder`
* `org.mapstruct.ap.test.updatemethods`
* `org.mapstruct.ap.test.conversion.string`

These were the ones which gave the most problems and I reckon pretty representative. Take the normal cases (not the erroneous ones, as far as they do actually generate code). 

Anyway: I agree with your remark: 

I wouldn't block this PR for it.

It doesn't make sense indeed, since:
1. the most troublesome stuff is already done (collections, null check and setters, ....).
2. The number of array tests (touched by this PR) are pretty limited and I checked them.

gunnarmorling:
Works for me; could you open an issue with the tests you mention above and
assign it to 1.2? Thanks!

sjaakd:
> Works for me; could you open an issue with the tests you mention above and
> assign it to 1.2? Thanks!

Sure, no problem. Checkout #1007 ..

sjaakd:
merged

PR #1006, opened 2016-12-20
#975 Introduce @Context annotation…

agudian:
… for passing context parameters through generated mapping methods to custom methods.

No doc update included, yet, so we can discuss the code changes first. What's also following up after that in a seperate PR is to add the ability to pick factory methods and `@BeforeMapping`/`@AfterMapping` methods directly from `@Context` parameter types (and to call them on the given context instances... (removing the need for that `CycleContextLifecycleMethods` bridge below in the test).

### DIFF
--- a/core-common/src/main/java/org/mapstruct/Context.java
+++ b/core-common/src/main/java/org/mapstruct/Context.java
@@ -1,0 +1,81 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks a parameter of a method to be treated as <em>mapping context</em>. Such parameters are passed to other mapping
+ * methods, {@link ObjectFactory} methods or {@link BeforeMapping}/{@link AfterMapping} methods when applicable and can
+ * thus be used in custom code. The {@link Context} parameters are otherwise ignored by MapStruct.
+ * <p>
+ * For generated code to call a method that is declared with {@link Context} parameters, the declaration of the mapping
+ * method being generated needs to contain at least those (or assignable) {@link Context} parameters as well. MapStruct
+ * will not create new instances of missing {@link Context} parameters nor will it pass {@code null} instead.
+ * <p>
+ * Example:
+ *
+ * <pre>
+ * <code>
+ * public abstract CarDto toCar(Car car, &#64;Context MyMappingContext context);
+ *
+ * &#64;BeforeMapping
+ * protected void registerVehicle(Vehicle mappedVehicle, &#64;Context MyMappingContext context) {
+ *     context.doSomethingWithTheVehicle( mappedVehicle );
+ * }
+ *
+ * &#64;BeforeMapping
+ * protected void logMappedVehicle(Vehicle mappedVehicle) {
+ *     // do something with the vehicle
+ * }
+ *
+ * &#64;BeforeMapping
+ * protected void notCalled(Vehicle mappedVehicle, &#64;Context DifferentMappingContextType context) {
+ *     // not called, because DifferentMappingContextType is not available within toCar(Car, MyMappingContext)
+ * }
+ *
+ * // generates:
+ *
+ * public CarDto toCar(Car car, MyMappingContext context) {
+ *     registerVehicle( car, context );
+ *     logMappedVehicle( car );
+ *
+ *     if ( car == null ) {
+ *         return null;
+ *     }
+ *
+ *     CarDto carDto = new CarDto();
+ *
+ *     // actual mapping code
+ *
+ *     return carDto;
+ * }
+ * </code>
+ * </pre>
+ *
+ * @author Andreas Gudian
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.CLASS)
+public @interface Context {
+
+}
--- a/core-common/src/main/java/org/mapstruct/Context.java
+++ b/core-common/src/main/java/org/mapstruct/Context.java
@@ -36,11 +36,16 @@ import java.lang.annotation.Target;
  *
  * <pre>
  * <code>
- * public abstract CarDto toCar(Car car, &#64;Context MyMappingContext context);
+ * // multiple &#64;Context parameters can be added
+ * public abstract CarDto toCar(Car car, &#64;Context VehicleRegistration context, &#64;Context Locale localeToUse);
+ *
+ * protected OwnerManualDto translateOwnerManual(OwnerManual ownerManual, &#64;Context Locale locale) {
+ *     // manually implemented logic to translate the OwnerManual with the given Locale
+ * }
  *
  * &#64;BeforeMapping
- * protected void registerVehicle(Vehicle mappedVehicle, &#64;Context MyMappingContext context) {
- *     context.doSomethingWithTheVehicle( mappedVehicle );
+ * protected void registerVehicle(Vehicle mappedVehicle, &#64;Context VehicleRegistration context) {
+ *     context.register( mappedVehicle );
  * }
  *
  * &#64;BeforeMapping
@@ -50,12 +55,13 @@ import java.lang.annotation.Target;
  *
  * &#64;BeforeMapping
  * protected void notCalled(Vehicle mappedVehicle, &#64;Context DifferentMappingContextType context) {
- *     // not called, because DifferentMappingContextType is not available within toCar(Car, MyMappingContext)
+ *     // not called, because DifferentMappingContextType is not available
+ *     // within toCar(Car, VehicleRegistration, Locale)
  * }
  *
  * // generates:
  *
- * public CarDto toCar(Car car, MyMappingContext context) {
+ * public CarDto toCar(Car car, VehicleRegistration context, Locale localeToUse) {
  *     registerVehicle( car, context );
  *     logMappedVehicle( car );
  *
@@ -65,7 +71,8 @@ import java.lang.annotation.Target;
  *
  *     CarDto carDto = new CarDto();
  *
- *     // actual mapping code
+ *     carDto.setOwnerManual( translateOwnerManual( car.getOwnerManual(), localeToUse );
+ *     // more generated mapping code
  *
  *     return carDto;
  * }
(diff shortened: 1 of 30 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
```diff
@@ -100,4 +110,36 @@ public boolean equals(Object obj) {
...
+    public static List<Parameter> getContextParameters(List<Parameter> parameters) {
+        List<Parameter> contextParameters = new ArrayList<Parameter>( parameters.size() );
+
+        for ( Parameter parameter : parameters ) {
+            if ( parameter.isMappingContext() ) {
+                contextParameters.add( parameter );
+            }
+        }
+
+        return contextParameters;
```
agudian: I wish we could use Java 8 for the processor, where stuff like this can be written as a one-liner... 😁

sjaakd:
First impression: 👍 . I'll have a more in depth look tomorrow (bit tired today).

agudian:
> Hey @agudian .. Very nice work. I did an initial review. One question: the "classic" factories also accept a `@TargetType`. Do they also accept the new `@Context`? Should they and should we have a unit test for it?

You're right, what else could those methods be there for... and we're ignoring the context params for most stuff anyway, so they should be ignored at that point as well. I've made sure of that and extended the test scenario.

sjaakd:
great. I'm done with reviewing. From my perspective: feel free to merge.

agudian:
Great, thanks for the review, @sjaakd!

@gunnarmorling, @filiphr, would you like to take a look as well, since this is a new feature?

core-common/src/main/java/org/mapstruct/Context.java
```diff
@@ -0,0 +1,81 @@
...
+ * <p>
+ * For generated code to call a method that is declared with {@link Context} parameters, the declaration of the mapping
+ * method being generated needs to contain at least those (or assignable) {@link Context} parameters as well. MapStruct
+ * will not create new instances of missing {@link Context} parameters nor will it pass {@code null} instead.
+ * <p>
+ * Example:
+ *
+ * <pre>
+ * <code>
+ * public abstract CarDto toCar(Car car, &#64;Context MyMappingContext context);
```
gunnarmorling: Maybe make it a bit more specifc, using something like a locale or time zone?

core-common/src/main/java/org/mapstruct/Context.java
```diff
@@ -0,0 +1,81 @@
...
+ *     return carDto;
+ * }
+ * </code>
+ * </pre>
+ *
+ * @author Andreas Gudian
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.CLASS)
+public @interface Context {
```
gunnarmorling: Can you add `@since 1.2`.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
```diff
@@ -34,18 +36,20 @@
     private final Type type;
     private final boolean mappingTarget;
     private final boolean targetType;
+    private final boolean mappingContext;
 
-    public Parameter(String name, Type type, boolean mappingTarget, boolean targetType) {
+    public Parameter(String name, Type type, boolean mappingTarget, boolean targetType, boolean mappingContext) {
```
gunnarmorling: I'm wondering whether we should create sub-types instead of using the booleans.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
gunnarmorling: Hum, maybe we raise the bar to 1.8 once Java 9 is out?

processor/src/test/java/org/mapstruct/ap/test/context/AutomappingNodeMapperWithContext.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+@Mapper(uses = CycleContextLifecycleMethods.class)
+public interface AutomappingNodeMapperWithContext {
+
+    AutomappingNodeMapperWithContext INSTANCE =
+        Mappers.getMapper( AutomappingNodeMapperWithContext.class );
+
+    NodeDTO nodeToNodeDTO(Node node, @Context CycleContext cycleContext, @Context FactoryContext factoryContext);
+
+    void nodeToNodeDTO(Node node, @MappingTarget NodeDTO nodeDto, @Context CycleContext cycleContext,
```
gunnarmorling: So far we preferred XyDto over XyDTO.

processor/src/test/java/org/mapstruct/ap/test/context/ContextParameterTest.java
```diff
@@ -0,0 +1,117 @@
...
+import org.mapstruct.ap.test.context.Node.Attribute;
+import org.mapstruct.ap.test.context.NodeDTO.AttributeDTO;
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+/**
+ * Tests the usage of the {@link Context} annotation in the following situations:
+ * <ul>
+ * <li>passing the parameter to property mapping methods
```
gunnarmorling: Does it work for update methods, too?

gunnarmorling:
Hey @agudian, nice stuff! Added some comments inline. I reckon you've planned to add another commit for the doc update?

Probably unrelated, I see a var name with two underscores in generated code: AttributeDTO attributeDTO__. I think there is an issue for that already.

processor/src/test/java/org/mapstruct/ap/test/context/AutomappingNodeMapperWithContext.java
agudian: Right. I started with a copy/paste from a differnt test... 😁

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
agudian: Hmm, sub types just for changing the result of three boolean accessors seems a bit over-dosed to me. How about making this constructor private and provide a more slim constructor or factory method? There's just one caller for this anyway...

agudian:
> I reckon you've planned to add another commit for the doc update?

I'm working on the second part of this feature (using factory/lifecycle methods defined on the context parameter's type) and will write the doc-update then, to have it consistent right away.


> Probably unrelated, I see a var name with two underscores in generated code: AttributeDTO attributeDTO__. I think there is an issue for that already.

Totally unrelated and we had an issue for it for a while. @sjaakd fixed it just yesterday. I've rebased and it looks much better now... 👍

agudian:
Let me know if this is okay to be squashed and merged... :)

agudian:
Anything else? I'd like to squash/merge this otherwise to make reviews for #1016 easier... 🙄

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
sjaakd: I'm a bit more conservative in this area. If we abandon 1.6 and 1.7 we should probably mark that with the transition to MapStruct 2.0. However, I do think that we could profit in the generated code already now. For instance, using `Optional` in the generated code for nested source properties when we detect java8.

processor/src/test/java/org/mapstruct/ap/test/context/ContextParameterTest.java
```diff
@@ -0,0 +1,117 @@
...
+    NodeDto.class,
+    NodeMapperWithContext.class,
+    AutomappingNodeMapperWithContext.class,
+    CycleContext.class,
+    FactoryContext.class,
+    CycleContextLifecycleMethods.class })
+@RunWith(AnnotationProcessorTestRunner.class)
+public class ContextParameterTest {
+
+    private static final int MATIC_NUMBER_OFFSET = 10;
```
filiphr: Should this be `MAGIC_NUMBER_OFFSET` instead of `MATIC_NUMBER_OFFSET`? There is also one variable name in the `FactoryContext` named `attributeMaticNumberOffset`. It is not a big deal, just something that confused me while having a look 😄

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
filiphr: As an addition to @sjaakd, we can also profit from the 1.7 diamond generics in generated code when we detect Java 7. Maybe we need to create an issue where we can track this

gunnarmorling:
Looks good to me. I'll unpack my presents and merge this PR later tonight :)

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
gunnarmorling: Yes, if we are going to require JDK 8 or later for the build, we'll need to raise the version to 2.0. Leveraging newer versions in the generated code is a good idea, I think we discussed it before but then abandoned it for some reason. I'd like it.

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -98,6 +98,7 @@
     RETRIEVAL_TYPE_VAR_RESULT( "Can't generate mapping method for a generic type variable target." ),
     RETRIEVAL_WILDCARD_SUPER_BOUND_SOURCE( "Can't generate mapping method for a wildcard super bound source." ),
     RETRIEVAL_WILDCARD_EXTENDS_BOUND_RESULT( "Can't generate mapping method for a wildcard extends bound result." ),
+    RETRIEVAL_CONTEXT_PARAMS_WITH_SAME_TYPE( "The types of @Context parameters must be unique." ),
```
gunnarmorling: What does this mean? I may not have two context parameters of the same type? What's the reason for this? I don't have a strong preference, just curious about the rationale behind it.

processor/src/test/java/org/mapstruct/ap/test/context/ContextParameterErroneousTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.mapstruct.ap.test.context.erroneous.ErroneousNodeMapperWithNonUniqueContextTypes;
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.compilation.annotation.CompilationResult;
+import org.mapstruct.ap.testutil.compilation.annotation.Diagnostic;
+import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutcome;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+/**
+ * Tests the errneous usage of the {@link Context} annotation in the following situations:
```
gunnarmorling: errneous=erroneous

I'll take care of it when merging.

processor/src/test/java/org/mapstruct/ap/test/context/ContextParameterTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Test
+    public void mappingWithContextCorrectlyResolvesCycles() {
+        Node root = buildNodes();
+        NodeDto rootDTO =
```
gunnarmorling: There's still some odd casing remaining. I'll handle it.

agudian:
Great, will you also take care of the typos Filip found and squash my
commits?

Otherwise I can handle it tomorrow night as well on my own... 🎄


Gunnar Morling <notifications@github.com> schrieb am Sa. 24. Dez. 2016 um
16:29:

gunnarmorling:
Yep, will do. Should I squash everything into one commit?

agudian:
Yes, please.. 😀

Gunnar Morling <notifications@github.com> schrieb am Sa. 24. Dez. 2016 um
16:47:

gunnarmorling:
Rebased, squashed and applied. Thanks, @agudian!

Issue #1001, opened 2016-12-19
Automapping follow-ups (error message and documentation)


sjaakd:
Follow up of #60:

- [x] create a dotted path notation for error messages where nested properties can't be mapped automatically
- [x] Adapt reference documentation
- [ ] Add entry to migration notes to note that there won't be an error message in case nested properties can be mapped automatically
- [x] ~~Make sure that configuration from prototype methods in `@MapperConfig` classes are applied to forged methods (I think that's currently not the case and never has been -- but we should tackle that soon as it becomes more important now).~~ This is now #1086
- [x] Align ForgedMethods with current ForgeMethod approach. ForgedMethods should not be properties on BeanMapping and PropertyMapping. The name should be generated based on the uniqueness of the method. See also discussions in #1011. A testcase should be added to verify this truly works (e.g. forcing the same name, with different property mappings). This could be done for instance by taking a DTO with the same name from different packages. 
- [x] `TestMultipleForgedMethodsTest` needs to do some assertions: create proper input data and assert the results. Rename the test class itself (`Test*Test`) and rename the classes `Bar`, `BarDto`, `Dto`... :)
- [x] Should work for update methods too

navpil:
How should we continue on this?
I already have some changes, which work, but they break some tests, which assume another format of the error message. But before I fix those tests I'd better know that the format I chose is the correct one. Should I do the PR with some tests failing so you can look at the current implementation (I have a dedicated test for these error messages)? Should we discuss it in the google group? Should we discuss it here?

agudian:
I'd say open a new PR right away and we can discuss it there directly with the modified code and the failing tests visible to all of us... :)

gunnarmorling:
>
> I'd say open a new PR right away and we can discuss it there directly with
> the modified code and the failing tests visible to all of us... :)
>
+1

agudian:
I've added another point about the test class `TestMultipleForgedMethodsTest` that I just stumbled upon... It looks a bit unfinished... 😉

sjaakd:
> Make sure that configuration from prototype methods in @MapperConfig classes are applied to forged methods (I think that's currently not the case and never has been -- but we should tackle that soon as it becomes more important now

[...]

PR #1010, opened 2016-12-22
#1001 Follow up on automapping: error messages with dotted properties

navpil:


agudian:
ok to test -- let's see what Jenkins says.

processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
```diff
@@ -202,6 +203,10 @@ private Assignment forgeMapping(SourceRHS sourceRHS, Type sourceType, Type targe
             return assignment;
         }
 
+        private String stubPropertyName(String fullyQualifiedName) {
+            return Strings.decapitalize( fullyQualifiedName.substring( fullyQualifiedName.lastIndexOf( '.' ) + 1 ) );
```
agudian: Doing this here for deriving a property name feels wrong... is it only used for adding a postfix after `[].` / `{}.` for iterable/maps?

processor/src/main/java/org/mapstruct/ap/internal/model/MapMappingMethod.java
```diff
@@ -214,6 +221,10 @@ private Assignment forgeMapping(SourceRHS sourceRHS, Type sourceType, Type targe
             return assignment;
         }
 
+        private String stubPropertyName(String fullyQualifiedName) {
+            return Strings.decapitalize( fullyQualifiedName.substring( fullyQualifiedName.lastIndexOf( '.' ) + 1 ) );
```
agudian: Hey, I know that one... If we really need it, the `Strings` class might actually a good place for it.

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/DottedErrorMessageTest.java
```diff
@@ -0,0 +1,117 @@
...
+        UnmappableDeepNestingMapper.class
+    })
+    @ExpectedCompilationOutcome(
+        value = CompilationResult.FAILED,
+        diagnostics = {
+            @Diagnostic(type = UnmappableDeepNestingMapper.class,
+                kind = javax.tools.Diagnostic.Kind.ERROR,
+                messageRegExp = "Can't map " + PROPERTY +
+                    " \".*Color house\\.roof\\.color\" to \".*house\\.roof\\.color\"\\. " +
+                    "Consider to declare/implement a mapping method: .*.")
```
agudian: Could you also include the line number attribute in the annotation and the suggested method in the reg-exp?

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/DottedErrorMessageTest.java
```diff
@@ -0,0 +1,117 @@
...
+    @WithClasses({
+        UnmappableDeepListMapper.class
+    })
+    @ExpectedCompilationOutcome(
+        value = CompilationResult.FAILED,
+        diagnostics = {
+            @Diagnostic(type = UnmappableDeepListMapper.class,
+                kind = javax.tools.Diagnostic.Kind.ERROR,
+                messageRegExp = "Can't map " + COLLECTION_ELEMENT +
+                    " \".*Wheel car\\.wheels\\[\\]\" to \".*car\\.wheels\\[\\]\"\\. " +
```
agudian: "Can't map collection element" and then having the path to the collection, `car.wheels` sounds pretty clear, no real need for `[]`, in my opinion.

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/DottedErrorMessageTest.java
```diff
@@ -0,0 +1,117 @@
...
+    @WithClasses({
+        UnmappableDeepMapKeyMapper.class
+    })
+    @ExpectedCompilationOutcome(
+        value = CompilationResult.FAILED,
+        diagnostics = {
+            @Diagnostic(type = UnmappableDeepMapKeyMapper.class,
+                kind = javax.tools.Diagnostic.Kind.ERROR,
+                messageRegExp = "Can't map " + MAP_KEY +
+                    " \".*Word dictionary\\.wordMap\\{\\}\\.word\" to \".*dictionary\\.wordMap\\{\\}\\.wordDto\"\\. " +
```
agudian: Same here... "Can't map map key" and the path to the map as `dictionary.wordMap` is sufficient, no real need for `{}` and a dummy property name for the key necessary, I'd say.

agudian:
Hmm, well the results look really good for the normal properties. 

I'm really not so sure if we should add the `[]` / `{}` to the path for iterables/maps and use a decapitalized element type name after that. I think I'd rather leave that out, as I don't think that it adds that much more information for the user. Or am I missing something?

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethodHistory.java
```diff
@@ -58,8 +54,51 @@ public Type getSourceType() {
...
+            else {
+                return prevHistory.getDottedTargetPropertyName();
+            }
+
+        }
     }
 
+    private String stripBrackets(String dottedName) {
+        return dottedName;
+//        if ( dottedName.endsWith( "[]" ) || dottedName.endsWith( "{}" )) {
```
filiphr: Do we really need this commented out code?

navpil:
That's why I was curious about how the error messages should look like. Imagine we can't map property of the wheel, let's say the TireType. Should we show it as user.car.wheels[].tireType or user.cars.wheels.tireType when wheels are actually the Iterable?
What if we can't map wheels? Should it show user.car.wheels[] or user.car.wheels? That's what the commented out code did -removed the [] and {} when "wheels" was the last property referenced, and not to remove them, when it's not the last.
Regarding maps: let's say we have unmappable key or value - do we need to differentiate between these two cases? Probably we want syntax like "user.dictionary{word:}.pronunciation can't map to user.dictionary{word:}.pronunciation" or just "user.dictionary.pronunciation can't map to user.dictionary.pronunciation"?
None of those are impossible to implement, we just need to agree on what is the best.

filiphr:
>Should we show it as user.car.wheels[].tireType or user.cars.wheels.tireType when wheels are actually the Iterable?

To me more clear would be using `user.car.wheels[].tireType`, as when I see `user.cars.wheels.tireType` I would first think that `tireType` is a property of some `wheels` object.

>What if we can't map wheels? Should it show user.car.wheels[] or user.car.wheels? 

Here I would expect to have `user.car.wheels`, as we want to show which property we couldn't map and it is not important to show whether it is an `Iterable` or not.

>Regarding maps: let's say we have unmappable key or value - do we need to differentiate between these two cases? 

I would say yes, then the user will know exactly what we can't map.

>Probably we want syntax like "user.dictionary{word:}.pronunciation can't map to user.dictionary{word:}.pronunciation" or just "user.dictionary.pronunciation can't map to user.dictionary.pronunciation"?

Here more clear to me would be `user.dictionary{word:}.pronunciation` than `user.dictionary.pronunciation`, for the same reason as the `Iterable`.

@navpil One question, when you generate the dotted path you use the names of the properties which we cannot map which are in the object, right? e.g. if `Car` had a `List<Wheel>` with the property name `myWheels` you would print `user.car.myWheels[].tireType`. I think that we only need to do decapitalization when we create a path for the `resultType` of an `Iterable` and/or the key/value of a `Map`.

navpil:
Yes, I generate the decapitalized names only when it's the collection element or map key/value.

navpil:
Error messages are currently implemented as per @filiphr comment. Some other tests were fixed.
There are at least three problems with the code:
1) Some integration tests fail (at least at my local machine) because of my new unmappable mappers.
2) I don't quite like the string-based programming in the ForgedMethodHistory in the getCorrectDottedPath() method
3) As I found out the Map<String, List<Room>> will not be mapped to Map<String, List<RoomDto>>. It will give errors that List<Room> can't be mapped to the List<RoomDto>. So there are some problems with automapping generics inside generics  - at least for maps. I have not investigated it yet.

navpil:
Don't mind the failed org.mapstruct.ap.test.collection.erroneous.ErroneousCollectionMappingTest.shouldFailOnNoElementMappingFound - that happened because of the last change to code, which I didn't expect to break the tests (bad idea - always, I know)

agudian:
> Some integration tests fail (at least at my local machine) because of my new unmappable mappers.

Rename the package `unmappables` to to `erroneous`, then the integration test will exclude those mappers from compilation.

agudian:
Hum, this is really a tough one. I get how we can help make the path expressions more readable and, most imporantly, unambigiuously, clear and helpful.

Not to die in back and forths here, but I'd like to sum up the cases that we currently have:

1. `Can't map property "org.mapstruct...Info computers[].info" to "org.mapstruct...InfoDto computers[].info". Consider to declare/implement a mapping method: "org.mapstruct...InfoDto map("org.mapstruct...Info source)".`, if the element type of `List<Computer> computers` has a property `Info info` that can't be mapped.
1. `Can't map collection element "org.mapstruct...Computer computers" to "org.mapstruct...ComputerDto computers". Consider to declare/implement a mapping method: "org.mapstruct...ComputerDto map("org.mapstruct...Computer source)".`, if the element type of `List<Computer> computers` can't be mapped.
1. `Can't map map key "org.mapstruct...Word dictionary.wordMap{word:}" to "org.mapstruct...WordDto dictionary.wordMap{wordDto:}". Consider to declare/implement a mapping method: "org.mapstruct...WordDto map(org.mapstruct...Word value)".`, if the key type of `Map<Word, WordDto> wordMap` can't be mapped.
1. `Can't map map value "org.mapstruct...ForeignWord dictionary.wordMap{:foreignWord}" to "org.mapstruct...ForeignWordDto dictionary.wordMap{:foreignWordDto}". Consider to declare/implement a mapping method: "org.mapstruct...ForeignWordDto map(org.mapstruct...ForeignWord value)".`, if the value type of `Map<Word, WordDto> wordMap` can't be mapped.
1. `Can't map property "org.mapstruct...Info userNameToUser{:user}.computers[].info" to "org.mapstruct...InfoDto userNameToUser{:userDto}.computers[].info". Consider to declare/implement a mapping method: "org.mapstruct...InfoDto map("org.mapstruct...Info source)".`.

I'm totally on board with 1. and 2.... But I'm still not sure about the representation of the key/value path in maps. Using the decapitalized type name is a nice way to show what's meant, but the distinction between key and value is only through the placement of the colon `:`. An alternative would be to use something like `{:key}` and `{:value}`, with the colon always prefixing, but the hint to the type name wouldn't be included anymore - which would be in line with the simple property paths and collection element paths. I'm not saying the alternative is better, but I'd like to mention it.

As a side note, I think "Can't map map value" reads a bit funny, so we should use "Can't map Map value" and "Can't map Collection element" instead...?

navpil:
Well, to me it's pretty much the same how we implement it. I tried to google some existing solution to the dot notation for maps to reuse it, but referring to the generic map's value as such is too specific and if some solution already exists, I could not find it. I also looked through Spring's SpEL and Dozer documentation to no avail.
It can be {:key}/{:value}, {myFoo:}/{:myBar} or even {:key(myFoo)}/{:value(myBar)}. All of them are a minute to implement and a bit more for fixing the test/run the check.
Which solution should we choose?

Regarding "map map key" I do like the phrase, but I will capitalize the first letter as you suggest to avoid confusion.

agudian:
Yes, I didn't find something authoritative on that either...

@gunnarmorling, @sjaakd, @filiphr, what do you think which variant we should use? (Check the previous two comments if it's TL;DR until now 😉)

> Regarding "map map key" I do like the phrase, but I will capitalize the first letter as you suggest to avoid confusion.
Great, thanks! 👍

agudian:
Cool, looks good!

filiphr:
I also think that it is OK. And I support @agudian summary and the last implemented idea about the map key and value representation

processor/src/main/java/org/mapstruct/ap/internal/util/Strings.java
```diff
@@ -172,6 +172,17 @@ public static String sanitizeIdentifierName(String identifier) {
     }
 
     /**
+     * Returns a stub property name from full class name by stripping away the package and decapitalizing the name
+     * For example will return <code>fooBar</code> for <code>com.foo.bar.baz.FooBar</code> class name
+     *
+     * @param fullyQualifiedName fully quialified class name, such as com.foo.bar.baz.FooBar
```
gunnarmorling: quialified=qualified

processor/src/main/java/org/mapstruct/ap/internal/util/Strings.java
```diff
@@ -172,6 +172,17 @@ public static String sanitizeIdentifierName(String identifier) {
     }
 
     /**
+     * Returns a stub property name from full class name by stripping away the package and decapitalizing the name
+     * For example will return <code>fooBar</code> for <code>com.foo.bar.baz.FooBar</code> class name
```
gunnarmorling: Instead of `<code>fooBar</code>`, could you use `{@code fooBar}`, please?

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/DottedErrorMessageTest.java
```diff
@@ -0,0 +1,186 @@
...
+        UnmappableDeepNestingMapper.class
+    })
+    @ExpectedCompilationOutcome(
+        value = CompilationResult.FAILED,
+        diagnostics = {
+            @Diagnostic(type = UnmappableDeepNestingMapper.class,
+                kind = javax.tools.Diagnostic.Kind.ERROR,
+                line = 26,
+                messageRegExp = "Can't map " + PROPERTY +
+                    " \".*Color house\\.roof\\.color\" to \".*house\\.roof\\.color\"\\. " +
```
gunnarmorling: Given the mapping is auto-generated, is there a need to state the property name twice? It cannot differ. How about this:

    Can't map property "house.roof.color" (source type "org.mapstruct.ap.test.nestedbeans.erroneous.Color", target type "org.mapstruct.ap.test.nestedbeans.erroneous.ColorDto"). Consider to declare/implement a mapping method "ColorDto map(Color value)".

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/DottedErrorMessageTest.java
```diff
@@ -0,0 +1,186 @@
...
+        UnmappableDeepMapKeyMapper.class
+    })
+    @ExpectedCompilationOutcome(
+        value = CompilationResult.FAILED,
+        diagnostics = {
+            @Diagnostic(type = UnmappableDeepMapKeyMapper.class,
+                kind = javax.tools.Diagnostic.Kind.ERROR,
+                line = 26,
+                messageRegExp = "Can't map " + MAP_KEY +
+                    " \".*Word dictionary\\.wordMap\\{:key\\}\" to \".*dictionary\\.wordMap\\{:key\\}\"\\. " +
```
gunnarmorling: How about

    Can't map key of map-typed property "dictionary.word map" (source type: ..., target type: ...). Consider to ...

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/DottedErrorMessageTest.java
agudian: Inherited Mapping customizations from a MapperConfig class could have been applied along the way (unsure if that works, yet, but that's on the list of open points).

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/DottedErrorMessageTest.java
agudian: Sounds good as well. But if an error is shown for a property below that key or value type, we still need that `{:key}` part in the property path. Not sure if it's worth the effort to differentiate all the variants in te code that generates the message (right now it uses the same pattern and inserts "property" / "collection element" / "Map key" / "Map value".

navpil: I agree with @agudian . Currently we use the same pattern for all messages and it might theoretically confuse the user why there are different error messages for similar cases.

gunnarmorling: Sorry for being slow with the response :/ I'm still not super-happy about the messages, but if that's the only thing missing let's merge it, we still can refine after that.

agudian:
@navpil I can take care of rebasing and merging it, if you like - the conflicts are in the same spots that I have to deal with in #1016 as well (had to already and will need to after this one again). I'd also squash your commits into one then. Is that okay for you? Authorship of the commit remains with you in any case.

navpil:
@agudian Sorry that I ignored this conversation for a while, but I had no free time. I will try to do fix this PR today. I'm still new to git, so the rebase process is not still crystal clear for me. But I'd like to do this myself to gain experience.

agudian:
Great, thanks!

navpil:
I hope I rebased/merged everything correctly. I did change the tests `ErroneousStreamMappingTest `and `ForgedStreamMappingTest`, please have a look whether the change is ok.

I did not change the error messages, so they will still duplicate the property name. However if there is no time pressure for this feature, I can look into the new way of showing the error messages. The change is not trivial because of need to fix all messages in tests and also something should be done with the way `ForgedMethodHistory` is created and used - we can't use the same pattern anymore. That's why I'd rather have us all agree on what the error messages should really look like for all situations (collections/stream elements, map key/values and nested properties for those) before actually doing the change. However in my opinion error messages are good now - they look the same for all occasions and I don't mind the duplication.

What should we do with nonmappable nested generics? In particular property of type `List<List<String>>` will fail to automap. Should I fix it here, or there should be another opened issue for that?

agudian:
> What should we do with nonmappable nested generics? In particular property of type `List<List<String>>` will fail to automap. Should I fix it here, or there should be another opened issue for that?

Let's tackle that in a seperate issue. Would you open up one?

I've pushed your commit to master now, the changes all look alright.

We can think about alternative error messages in a new issue as well - with that we could also do some refactoring to not rely on the string comparisons and the passing of the effectively unused source/target element names for collection elements / map components. I guess that would allow for nicer handling of different variants.

Thanks again for your contributions! We really appreciate your patience with us... 😉

Issue #1014, opened 2016-12-22
FQN is generated for super and extend bounded Types


filiphr:
Have a look at the `org.mapstruct.ap.test.collection.wildcard.ExtendsBoundSourceTargetMapperImpl` and `org.mapstruct.ap.test.collection.wildcard.SourceSuperBoundTargetMapperImpl`:

There is something like:
```java
protected List<Plan> ideaListToPlanList(List<? extends org.mapstruct.ap.test.collection.wildcard.Idea> list) {
```
and
```java
List<? super org.mapstruct.ap.test.collection.wildcard.Plan> list = ideaListToPlanList( source.getElements() );
```

gunnarmorling:
PR merged.

PR #1015, opened 2016-12-22
#1014 do not generate FQN for extends/super bound Types

filiphr:
The problem was that for extends/super bound `Type`(s) the name is `? extends/super java.lang.Integer`.

Do we need some more complex tests?


gunnarmorling:
Looks good. Merge it yourself? Thanks!

Issue #992, opened 2016-12-12
Remove @Experimental annotation from before/after mappings


gunnarmorling:

agudian:
Closing as duplicate, it's also mentioned as part of #983 and resolved in PR #1016.

Issue #1012, opened 2016-12-22
Cleanup the code duplication between the Stream and Iterable support


filiphr:
- [x] In `MapperCreationProcessor` creation of `IterableMappingMethod` and `StreamMappingMethod` are really similar. Extract a common Builder for the mapping methods and reuse as much as possible the code
- [x] The `StreamMappingMethod` and the `IterableMappingMethod` are almost identical, extract some common class between the two

gunnarmorling:
Done.

Issue #991, opened 2016-12-12
Split reference guide source into an adoc file per chapter


gunnarmorling:

gunnarmorling:
Also re-format it to use one sentence (or line of thought for longer sentences) per line.

Captain1653:
Maybe we steel need it. But I little didn't understand   about reformat. I have found two long lines: Chapter 10.6 and Chapter 12.2.  @gunnarmorling Is it what do you mean?

PR #1016, opened 2016-12-23
#975 Part 2: Context params used as source for factory and lifecycle methods

agudian:
Only the last commit is new here, the first commits are identical to what's open in PR #1006.

I've already worked in the changes in the Javadoc of `Context` and `ObjectFactory`. I'll do the adaptation of `BeforeMapping` / `AfterMapping` seperately in a PR for #992, if that's okay.

The introduction of `@Context` in the ref-docs will be included in this PR later on (probably after christmas), but I wanted to show you how it's coming along. Code-wise, I'm finished with it so you could play around if you like.

### DIFF
--- a/core-common/src/main/java/org/mapstruct/Context.java
+++ b/core-common/src/main/java/org/mapstruct/Context.java
@@ -25,14 +25,22 @@ import java.lang.annotation.Target;
 
 /**
  * Marks a parameter of a method to be treated as <em>mapping context</em>. Such parameters are passed to other mapping
- * methods, {@link ObjectFactory} methods or {@link BeforeMapping}/{@link AfterMapping} methods when applicable and can
- * thus be used in custom code. The {@link Context} parameters are otherwise ignored by MapStruct.
+ * methods, {@code @}{@link ObjectFactory} methods or {@code @}{@link BeforeMapping}/{@code @}{@link AfterMapping}
+ * methods when applicable and can thus be used in custom code.
  * <p>
- * For generated code to call a method that is declared with {@link Context} parameters, the declaration of the mapping
- * method being generated needs to contain at least those (or assignable) {@link Context} parameters as well. MapStruct
- * will not create new instances of missing {@link Context} parameters nor will it pass {@code null} instead.
+ * The type of an {@code @Context} parameter is also inspected for
+ * {@code @}{@link BeforeMapping}/{@code @}{@link AfterMapping} methods, which are called on the provided context
+ * parameter value if applicable.
  * <p>
- * Example:
+ * <strong>Note:</strong> no {@code null} checks are performed before calling before/after mapping methods on context
+ * parameters. The caller needs to make sure that no {@code null} are passed in that case.
+ * <p>
+ * For generated code to call a method that is declared with {@code @Context} parameters, the declaration of the mapping
+ * method being generated needs to contain at least those (or assignable) {@code @Context} parameters as well. MapStruct
+ * will not create new instances of missing {@code @Context} parameters nor will it pass {@code null} instead.
+ * <p>
+ * <strong>Example 1:</strong> Using {@code @Context} parameters for passing data down to hand-written property mapping
+ * methods and {@code @}{@link BeforeMapping} methods:
  *
  * <pre>
  * <code>
@@ -49,13 +57,8 @@ import java.lang.annotation.Target;
  * }
  *
  * &#64;BeforeMapping
- * protected void logMappedVehicle(Vehicle mappedVehicle) {
- *     // do something with the vehicle
- * }
- *
- * &#64;BeforeMapping
  * protected void notCalled(Vehicle mappedVehicle, &#64;Context DifferentMappingContextType context) {
- *     // not called, because DifferentMappingContextType is not available
+ *     // not called, because no context parameter of type DifferentMappingContextType is available
  *     // within toCar(Car, VehicleRegistration, Locale)
  * }
  *
@@ -63,7 +66,6 @@ import java.lang.annotation.Target;
  *
  * public CarDto toCar(Car car, VehicleRegistration context, Locale localeToUse) {
  *     registerVehicle( car, context );
- *     logMappedVehicle( car );
  *
  *     if ( car == null ) {
  *         return null;
@@ -75,6 +77,61 @@ import java.lang.annotation.Target;
  *     // more generated mapping code
  *
  *     return carDto;
+ * }
+ * </code>
+ * </pre>
+ * <p>
+ * <strong>Example 2:</strong> Using an {@code @Context} parameter with a type that provides its own {@code @}
+ * {@link BeforeMapping} methods to handle cycles in Graph structures:
+ *
+ * <pre>
+ * <code>
+ * // type of the context parameter
+ * public class CyclicGraphContext {
+ *     private Map&lt;Object, Object&gt; knownInstances = new IdentityHashMap&lt;&gt;();
+ *
+ *     &#64;BeforeMapping
+ *     public &lt;T extends NodeDto&gt; T getMappedInstance(Object source, &#64;TargetType Class&lt;T&gt; targetType) {
+ *         return (T) knownInstances.get( source );
+ *     }
+ *
+ *     &#64;BeforeMapping
+ *     public void storeMappedInstance(Object source, &#64;MappingTarget NodeDto target) {
+ *         knownInstances.put( source, target );
+ *     }
+ * }
+ *
+ * &#64;Mapper
+ * public interface GraphMapper {
+ *     NodeDto toNodeDto(Node node, &#64;Context CyclicGraphContext cycleContext);
+ * }
+ *
+ *
+ * // generates:
+ *
+ * public NodeDto toNodeDto(Node node, CyclicGraphContext cycleContext) {
+ *     NodeDto target = cycleContext.getMappedInstance( node, NodeDto.class );
+ *     if ( target != null ) {
+ *         return target;
+ *     }
+ *
+ *     if ( node == null ) {
+ *         return null;
+ *     }
+ *
+ *     NodeDto nodeDto = new NodeDto();
+ *
+ *     cycleContext.storeMappedInstance( node, nodeDto );
+ *
+ *     nodeDto.setParent( toNodeDto( node.getParent(), cycleContext ) );
+ *     List&lt;NodeDto&gt; list = nodeListToNodeDtoList( node.getChildren(), cycleContext );
+ *     if ( list != null ) {
+ *         nodeDto.setChildren( list );
+ *     }
+ *
+ *     // more mapping code
+ *
+ *     return nodeDto;
  * }
  * </code>
  * </pre>
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -924,6 +924,54 @@ public class CarMapperImpl implements CarMapper {
 ----
 ====
 
+[[passing-context]]
+=== Passing context or state objects to custom methods
+
+Additional _context_ or _state_ information can be passed through generated mapping methods to custom methods with `@Context` parameters. Such parameters are passed to other mapping methods, `@ObjectFactory` methods (see <<object-factories>>) or `@BeforeMapping` / `@AfterMapping` methods (see <<customizing-mappings-with-before-and-after>>) when applicable and can thus be used in custom code.
+
+`@Context` parameters are also searched for `@BeforeMapping` / `@AfterMapping` methods, which are called on the provided context parameter value if applicable.
+
+*Note:* no `null` checks are performed before calling before/after mapping methods on context parameters. The caller needs to make sure that `null` is not passed in that case.
+
+For generated code to call a method that is declared with `@Context` parameters, the declaration of the mapping method being generated needs to contain at least those (or assignable) `@Context` parameters as well. The generated code will not create new instances of missing `@Context` parameters nor will it pass a literal `null` instead.
+
+.Using `@Context` parameters for passing data down to hand-written property mapping methods
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+public abstract CarDto toCar(Car car, @Context Locale translationLocale);
+
+protected OwnerManualDto translateOwnerManual(OwnerManual ownerManual, @Context Locale locale) {
+    // manually implemented logic to translate the OwnerManual with the given Locale
+}
+----
+====
+
+MapStruct will then generate something like this:
+
+.Generated code
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+//GENERATED CODE
+public CarDto toCar(Car car, Locale translationLocale) {
+    if ( car == null ) {
+        return null;
+    }
+
+    CarDto carDto = new CarDto();
+
+    carDto.setOwnerManual( translateOwnerManual( car.getOwnerManual(), translationLocale );
+    // more generated mapping code
+
+    return carDto;
+}
+----
+====
+
+
 [[mapping-method-resolution]]
 === Mapping method resolution
 
@@ -1492,8 +1540,7 @@ By default, the generated code for mapping one bean type into another will call 
 
 Alternatively you can plug in custom object factories which will be invoked to obtain instances of the target type. One use case for this is JAXB which creates `ObjectFactory` classes for obtaining new instances of schema types.
 
-To make use of custom factories register them via `@Mapper#uses()` as described in <<invoking-other-mappers>>. When creating the target object of a bean mapping, MapStruct will look for a parameterless method, 
-a method annotated with `@ObjectFactory`, or a method with only one `@TargetType` parameter that returns the required target type and invoke this method instead of calling the default constructor:
+To make use of custom factories register them via `@Mapper#uses()` as described in <<invoking-other-mappers>>, or implement them directly in your mapper. When creating the target object of a bean mapping, MapStruct will look for a parameterless method, a method annotated with `@ObjectFactory`, or a method with only one `@TargetType` parameter that returns the required target type and invoke this method instead of calling the default constructor:
 
 .Custom object factories
 ====
@@ -1569,9 +1616,9 @@ public class CarMapperImpl implements CarMapper {
 ----
 ====
 
-In addition, annotating a factory with `@ObjectFactory` lets you gain access to the mapping sources. 
+In addition, annotating a factory method with `@ObjectFactory` lets you gain access to the mapping sources.
 Source objects can be added as parameters in the same way as for mapping method. The `@ObjectFactory`
-annotation is necessary to let MapStruct know that the given method is only a factory method. 
+annotation is necessary to let MapStruct know that the given method is only a factory method.
 
 .Custom object factories with `@ObjectFactory`
 
@@ -1583,7 +1630,7 @@ public class DtoFactory {
 
      @ObjectFactory
      public CarDto createCarDto(Car car) {
-         return // ... custom factory logic 
+         return // ... custom factory logic
      }
 }
 ----
@@ -1908,7 +1955,7 @@ Methods that are considered for inverse inheritance need to be defined in the cu
 
 If multiple methods qualify, the method from which to inherit the configuration needs to be specified using the `name` property like this: `@InheritInverseConfiguration(name = "carToDto")`.
 
-Expressions and constants are excluded (silently ignored). Reverse mapping of nested source properties is experimental as of the 1.1.0.Beta2. Reverse mapping will take place automatically when the source property name and target property name are identical. Otherwise, `@Mapping` should specify both the target name and source name. In all cases, a suitable mapping method needs to be in place for the reverse mapping.
+Expressions and constants are excluded (silently ignored). Reverse mapping of nested source properties is experimental as of the 1.1.0.Beta2 release. Reverse mapping will take place automatically when the source property name and target property name are identical. Otherwise, `@Mapping` should specify both the target name and source name. In all cases, a suitable mapping method needs to be in place for the reverse mapping.
 
 [NOTE]
 ====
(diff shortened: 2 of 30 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -111,7 +111,6 @@ else if ( Executables.isGetterMethod( accessor ) ) {
         protected TargetWriteAccessorType targetWriteAccessorType;
         protected Type targetType;
         protected Accessor targetReadAccessor;
-        protected TargetWriteAccessorType targetReadAccessorType;
```
agudian: Was marked as unused, so I removed it.

agudian:
Commit is rebased to the current master and ready for review.

One thing you might want to think about: as it is implemented now, any non-void methods without parameters on the context parameter type are considered as potential factory methods (as we do it on mappers and `uses` types as well). So it's consistent in that way, but perhaps users want to have context types with such methods (e.g. `getFirstCreatedInstance()`) or want to use context param types that are not under their control that by chance have such methods. That's why I had to change `FactoryContext#createNode` in the test to not collect it as potential factory method...

So should we make `@ObjectFactory` mandatory for factory methods on `@Context` parameter types?

agudian:
> Commit is rebased to the current master and ready for review.

uhm, what I mean with "ready for review" is that the code parts are done, but the ref-doc update is still missing. I'll start on that soonish and add it to this PR. 😁

agudian:
I've added the Javadoc changes to `@BeforeMapping` and `@AfterMapping`, and added the reference-documentation.

core-common/src/main/java/org/mapstruct/Context.java
```diff
@@ -25,14 +25,22 @@
...
  * Marks a parameter of a method to be treated as <em>mapping context</em>. Such parameters are passed to other mapping
- * methods, {@link ObjectFactory} methods or {@link BeforeMapping}/{@link AfterMapping} methods when applicable and can
- * thus be used in custom code. The {@link Context} parameters are otherwise ignored by MapStruct.
+ * methods, {@code @}{@link ObjectFactory} methods or {@code @}{@link BeforeMapping}/{@code @}{@link AfterMapping}
+ * methods when applicable and can thus be used in custom code.
  * <p>
- * For generated code to call a method that is declared with {@link Context} parameters, the declaration of the mapping
- * method being generated needs to contain at least those (or assignable) {@link Context} parameters as well. MapStruct
- * will not create new instances of missing {@link Context} parameters nor will it pass {@code null} instead.
+ * The type of an {@code @Context} parameter is also inspected for factory methods (with or without the {@code @}
```
filiphr: Nothing special, but why not use `{@code @}{@link Context}` like the other annotations?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -924,6 +924,54 @@ public class CarMapperImpl implements CarMapper {
...
+[[passing-context]]
+=== Passing context or state objects to custom methods
+
+Additional _context_ or _state_ information can be passed through generated mapping methods to custom methods with `@Context` parameters. Such parameters are passed to other mapping methods, `@ObjectFactory` methods (see <<object-factories>>) or `@BeforeMapping` / `@AfterMapping` methods (see <<customizing-mappings-with-before-and-after>>) when applicable and can thus be used in custom code.
+
+The type of an `@Context` parameter is also inspected for factory methods and `@BeforeMapping` / `@AfterMapping` methods, which are called on the provided context parameter value if applicable.
+
+*Note:* no `null` checks are performed before calling factory or before/after mapping methods on context parameters. The caller needs to make sure that no `null` are passed in that case.
+
+For generated code to call a method that is declared with `@Context` parameters, the declaration of the mapping method being generated needs to contain at least those (or assignable) `@Context` parameters as well. MapStruct will not create new instances of missing`@Context` parameters nor will it pass `null` instead.
```
filiphr: a space is missing in 
> missing`@Context`

filiphr:
Regarding your question @agudian, I think that maybe the others can help more then me.

And I forgot to add a note to me. The changes in the `IterableMappingMethod` need to be taken into consideration in #964.

core-common/src/main/java/org/mapstruct/Context.java
agudian: I didn't want to link if we are already within that file... 🤓

filiphr: oh yea you are right. I didn't see that I was reading in that file 😄

gunnarmorling:
WDYM by

> without parameters on the context parameter type

?

gunnarmorling:
> context param types that are not under their control that by chance have such methods

In which case are methods declared by a context parameter considered at all? So far I understood the context parameter(s) would just be passed through, but you seem to imply that their methods would potentially be invoked by generated code?

agudian:
> non-void methods without parameters on the context parameter type

That means a method without parameters that returns something other than `void` is declared on the type of the context parameter. We usually identify such methods as candidate factory methods (in mappers and types added with `@Mapper(uses = ...)`.

agudian:
> So far I understood the context parameter(s) would just be passed through, but you seem to imply that their methods would potentially be invoked by generated code?

That's the whole point of this PR 👍. You can have `@BeforeMapping` / `@AfterMapping` and `@ObjectFactory` methods declared on the type of the context parameter. Check the type `SelfContainingCycleContext` in the tests and it's usages for one example of what to do with that.

agudian:
Oh, I also added that as "Example 2" to the Javadoc of `@interface Context` :)

filiphr:
@agudian, when you will be doing the rebase over the master. Can you please make sure that the `StreamMappingMethod` has the new things applied in the `IterableMappingMethod` as well.

Once I do #1012 then it will be easier and changes that actually affect both methods will be in once place

core-common/src/main/java/org/mapstruct/Context.java
```diff
@@ -78,6 +80,61 @@
...
+ *         return target;
+ *     }
+ *
+ *     if ( node == null ) {
+ *         return null;
+ *     }
+ *
+ *     NodeDto nodeDto = new NodeDto();
+ *
+ *     cycleContext.storeMappedInstance( node, nodeDto );
```
gunnarmorling: How is it ensured that `getMappedInstance()` is called on line 113 and here then `storeMappedInstance()`?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -924,6 +924,54 @@ public class CarMapperImpl implements CarMapper {
 ----
 ====
 
+[[passing-context]]
+=== Passing context or state objects to custom methods
+
+Additional _context_ or _state_ information can be passed through generated mapping methods to custom methods with `@Context` parameters. Such parameters are passed to other mapping methods, `@ObjectFactory` methods (see <<object-factories>>) or `@BeforeMapping` / `@AfterMapping` methods (see <<customizing-mappings-with-before-and-after>>) when applicable and can thus be used in custom code.
+
+The type of an `@Context` parameter is also inspected for factory methods and `@BeforeMapping` / `@AfterMapping` methods, which are called on the provided context parameter value if applicable.
```
gunnarmorling: "`@Context` parameters are also searched for..."

But I'm not quite convinced of the overall idea.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -924,6 +924,54 @@ public class CarMapperImpl implements CarMapper {
...
+
+*Note:* no `null` checks are performed before calling factory or before/after mapping methods on context parameters. The caller needs to make sure that no `null` are passed in that case.
+
+For generated code to call a method that is declared with `@Context` parameters, the declaration of the mapping method being generated needs to contain at least those (or assignable) `@Context` parameters as well. MapStruct will not create new instances of missing `@Context` parameters nor will it pass `null` instead.
```
gunnarmorling: Can you put one sentence per line? That makes it easier to do diffs down the road.

"MapStruct will" sounds as if it s a runtime component. Better change it into something like "Code generated by MapStruct will not...".

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -1490,8 +1538,7 @@ By default, the generated code for mapping one bean type into another will call
 
 Alternatively you can plug in custom object factories which will be invoked to obtain instances of the target type. One use case for this is JAXB which creates `ObjectFactory` classes for obtaining new instances of schema types.
 
-To make use of custom factories register them via `@Mapper#uses()` as described in <<invoking-other-mappers>>. When creating the target object of a bean mapping, MapStruct will look for a parameterless method, 
-a method annotated with `@ObjectFactory`, or a method with only one `@TargetType` parameter that returns the required target type and invoke this method instead of calling the default constructor:
+To make use of custom factories register them via `@Mapper#uses()` as described in <<invoking-other-mappers>> or add a `@Context` parameter (see <<passing-context>>). When creating the target object of a bean mapping, MapStruct will look for a parameterless method, a method annotated with `@ObjectFactory`, or a method with only one `@TargetType` parameter that returns the required target type and invoke this method instead of calling the default constructor:
```
gunnarmorling: Better make it one sentence per line.

processor/src/test/java/org/mapstruct/ap/test/context/SelfContainingFactoryContext.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package org.mapstruct.ap.test.context;
+
+import org.mapstruct.Context;
+import org.mapstruct.ObjectFactory;
+import org.mapstruct.ap.test.context.Node.Attribute;
+import org.mapstruct.ap.test.context.NodeDto.AttributeDto;
+
+/**
+ * A type to be used as {@link Context} parameter to create NodeDTO and AttributeDTO instances
```
gunnarmorling: DTO=Dto.

processor/src/test/java/org/mapstruct/ap/test/context/ContextParameterTest.java
```diff
@@ -77,15 +81,35 @@ public void automappingWithContextCorrectlyResolvesCycles() {
...
         AutomappingNodeMapperWithContext.INSTANCE
-            .nodeToNodeDto( root, updated, new CycleContext(), new FactoryContext( 1, 10 ) );
+            .nodeToNodeDto( root, updated, new CycleContext(), new FactoryContext( 10, 10 ) );
+        assertResult( updated );
+    }
+
+    @Test
+    public void automappingWithSelfContainingContextCorrectlyResolvesCycles() {
+        Node root = buildNodes();
+        NodeDto rootDto = AutomappingNodeMapperWithSelfContainingContext.INSTANCE
```
gunnarmorling: What does "Automapping" mean?

processor/src/test/java/org/mapstruct/ap/test/context/AutomappingNodeMapperWithSelfContainingContext.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * @author Andreas Gudian
+ */
+@Mapper
+public interface AutomappingNodeMapperWithSelfContainingContext {
+
+    AutomappingNodeMapperWithSelfContainingContext INSTANCE =
+        Mappers.getMapper( AutomappingNodeMapperWithSelfContainingContext.class );
+
+    NodeDto nodeToNodeDto(Node node, @Context SelfContainingCycleContext cycleContext,
```
gunnarmorling: Should we really support multiple context parameters? I can see issues or at least very complex rules around method selection. A single context parameter should be enough, users can structure them as needed.

gunnarmorling:
> I admit the cycle detection use case is compelling, but I find the feature very tough to grok and use. Rules around it seem quite involved.

Specifically, could that use case not be solved by simply invoking the right methods on the context object from within `@BeforeMapping` methods?

agudian:
> Specifically, could that use case not be solved by simply invoking the right methods on the context object from within @BeforeMapping methods?

Yes it can be, and one of the existing test cases already demonstrates that. But I find this for some use cases to be a much better solution, as it is totally self-contained (data and logic in the same place) and I imagine that to be an easier solution to hand out for users wanting to handle cycles (and perhaps a starting point for automatically handling it in our code)

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
agudian: I'd gladly do so, I just followed the old habit from the awestruct days. Would it be okay if I did that for the whole file at once, in a seperate commit or PR? Then my text editor can do it for me and we have it fixed once and for all... 😁

gunnarmorling:
> Would it be okay if I did that for the whole file at once, in a seperate
commit or PR?

+1 for a separate PR. Maybe combining it with splitting the guide into one
file per chapter?

2017-01-10 22:20 GMT+01:00 Andreas Gudian <notifications@github.com>:

sjaakd:
I start to like the idea. A while ago I created a stageful mapper PR. The idea there was that we could implement handwritten methods (e.g. in an abstract-class-mapper) and use the context in these mappings. Use cases were:

1. I have 2 modes (quality regimes) in which a mapper should work.  The archive regime has a bigger set of possible values in general for a property than the new regime (more restricted). So take property X it maps String value A to K, B to L, C to M under the archive regime. But under the standard regime A maps to K, B to N, C to N. All other 'value' properties are mapped in a similar way. I would normally require a hand-written method to do each value mapping (with qualifiers pointing out which property it concerns (each one is string-to-string). In Andreas his solution I would do this with an `@AfterMapping` and hand the complete object. The context would have the complete object and do all stuff since it knows the quality regime. All value properties would be ignored.

2. ID generation. Some properties require a unique ID (counter) in a web service request. I now handle this with a ThreadLocal and an expression. I would use a IdGenerator (wrapping a counter) as context. Now I would again use a post processing method and call the getId (incrementing counter and return Id) on the IdGenerator. All concerned Id properties would be ignored.

These scenarios are mixed. So I would require 2 context objects, since their purpose is entirely different (nice separation of concerns). That covers (I believe one of your other remarks @gunnarmorling above).


So to make a long story short: I have already 2 use cases that I can cover quite easily with this solution. :smile:

sjaakd:
See my case 1 above: @agudian .. Is it possible to have an interface as `@Context` object and stick the annotation `@AfterMapping` to the interface?

agudian:
Absolutely! 🙂

sjaakd:
Nice 👍

agudian:
I've rebased and improved the documentation bits as Gunnar suggested, but I'll leave the line-breaks and split-ups for #991.
@gunnarmorling, I guess you'll need the weekend to think a bit more about it? Would a beer help? 🍻  😉

gunnarmorling:
@agudian thinking a bit more about this one, would it suffice to have support for lifecycle methods (before-mapping, after-mapping) on context objects? But factory methods and other mapping methods would be obtained from "used" mappers as before?

agudian:
Sure, it's a start. 😄 
Leaving out the factory methods seems a bit arbitrary to me, though. So I'd like to understand the rationale.

Is it because of the dilemma with the non-annotated factory methods I described above? I'd sleep better without using non-annotated factory methods as well.

But methods explicitly annotated with `@ObjectFactory`? Wouldn't it be strange to support the `@BeforeMapping`/`@AfterMapping` methods, but not `@ObjectFactory`? Or is it more strange if we would require `@ObjectFactory` on context parameters for factory methods without arguments, but leave them optional on _used_ mappers?

Or is it because we can't devise a proper life-like example? 😁

gunnarmorling:
Hey @agudian, sorry for the delay, finally coming back to this.

So my preference for only supporting lifecycle methods (before-mapping, after-mapping) on context objects is due to the fact that i'd expect those methods to be the ones actually interacting with the context. I.e. I don't see a strong reason for opening up more ways where mapping (and factory) methods should be defined; in a library, there should be one way to achieve a given goal. Your question around `@ObjectFactory` is a good example, why creating this issue if we could simply stick to the current ways of defining mapping and factory methods?

I guess one could say the same about the lifecycle methods, but here I find the use case more compelling indeed (as it allows for "plugging in" a context with its logic). So my personal preference is be to limit it to this - unless we can find a really good use case for the alternative.

We still can expand it in subsequent revisions if people ask for it.

agudian:
Okay, PR is updated to not search for factory methods on `@Context` params anymore. I did some rewriting of the `@ObjectFactory` javadoc that I left in... I didn't mention factory methods *on* context params there anyway.

core-common/src/main/java/org/mapstruct/Context.java
```diff
@@ -25,14 +25,22 @@
...
- * Example:
+ * <strong>Note:</strong> no {@code null} checks are performed before calling before/after mapping methods on context
+ * parameters. The caller needs to make sure that no {@code null} are passed in that case.
+ * <p>
+ * For generated code to call a method that is declared with {@code @Context} parameters, the declaration of the mapping
+ * method being generated needs to contain at least those (or assignable) {@code @Context} parameters as well. MapStruct
+ * will not create new instances of missing {@code @Context} parameters nor will it pass {@code null} instead.
+ * <p>
+ * <strong>Example 1:</strong> Using {@code @Context} parameters for passing data down to hand-written property mapping
+ * methods and {@code @} {@link BeforeMapping} methods:
```
gunnarmorling: Minor detail: the "@" seems to be handled a bit inconsistently. So far we always expressed it via "&#64;". Here I probably wouldn't bother adding it at all.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -924,6 +924,54 @@ public class CarMapperImpl implements CarMapper {
...
 ====
 
+[[passing-context]]
+=== Passing context or state objects to custom methods
+
+Additional _context_ or _state_ information can be passed through generated mapping methods to custom methods with `@Context` parameters. Such parameters are passed to other mapping methods, `@ObjectFactory` methods (see <<object-factories>>) or `@BeforeMapping` / `@AfterMapping` methods (see <<customizing-mappings-with-before-and-after>>) when applicable and can thus be used in custom code.
+
+`@Context` parameters are also searched for `@BeforeMapping` / `@AfterMapping` methods, which are called on the provided context parameter value if applicable.
+
+*Note:* no `null` checks are performed before calling before/after mapping methods on context parameters. The caller needs to make sure that no `null` are passed in that case.
```
gunnarmorling: ...to make sure that `null` is not passed...

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -924,6 +924,54 @@ public class CarMapperImpl implements CarMapper {
 ----
 ====
 
+[[passing-context]]
+=== Passing context or state objects to custom methods
+
+Additional _context_ or _state_ information can be passed through generated mapping methods to custom methods with `@Context` parameters. Such parameters are passed to other mapping methods, `@ObjectFactory` methods (see <<object-factories>>) or `@BeforeMapping` / `@AfterMapping` methods (see <<customizing-mappings-with-before-and-after>>) when applicable and can thus be used in custom code.
```
gunnarmorling: Can we stick to one sentence per line? Makes it easier with diffs down the road.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -924,6 +924,54 @@ public class CarMapperImpl implements CarMapper {
...
+*Note:* no `null` checks are performed before calling before/after mapping methods on context parameters. The caller needs to make sure that no `null` are passed in that case.
+
+For generated code to call a method that is declared with `@Context` parameters, the declaration of the mapping method being generated needs to contain at least those (or assignable) `@Context` parameters as well. The generated code will not create new instances of missing `@Context` parameters nor will it pass a literal `null` instead.
+
+.Using `@Context` parameters for passing data down to hand-written property mapping methods
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+public abstract CarDto toCar(Car car, @Context Locale translationLocale);
```
gunnarmorling: Nice example!

processor/src/test/java/org/mapstruct/ap/test/context/AutomappingNodeMapperWithSelfContainingContext.java
gunnarmorling: @agudian Seems you never got back to this one?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
agudian: I'd like to reformat that whole file with #991... It's even assigned to me already... 😁

agudian:
> Seems you never got back to this one?

What do you mean? (sorry, can't comment up there, I don't get a reply box for the last comment)

gunnarmorling:
My question was about whether multiple context parameters should be supported or just a single one. I can see more complexities coming in when doing the former. E.g. say your original method takes tour context parameters of the same type and calls another one which only takes one parameter of that same type. Which one is passed on?

agudian:
> My only remaining question is whether we should limit it to one context parameter. May make reasoning about ordering and so on a bit simpler.

I could imagine some use cases where you would want to handle different aspects in different branches of the data structure... I think Sjaak even had an example for this? As long as the types are different, there's no real problem here anyway. But I'd like to hear if users come back with use cases where they have more than one context parameter of the same type (e.g. different `Locale` parameters for different parts of their beans) and then see what we could do there - if it comes up.

agudian:
> E.g. say your original method takes tour context parameters of the same type and calls another one which only takes one parameter of that same type. Which one is passed on?

Ah, okay - currently you get an error if you declare a method with more than one context param of the same type.

agudian:
See `ContextParameterErroneousTest` (was part of the Part 1 PR)...

Issue #285, opened 2014-08-19
Selection of the proper factory method for generated (e.g. iterable to iterable) mappings.


sjaakd:
The solution for a mapping from List to List should take into consideration the selection of the proper factory JAXB factory methods to cover the case: 
`List<JAXBElement<String>> listOfStringToListOfJaxbElement(List<String> strings);`.  

The problem here is that calling context information that is contained in the `XmlElementDecl` (see also #135) needs to be passed to the `IterableMappingMethod` somehow. #119 describes a possible solution in the form of a new annotation: 

``` java
@IterableMapping(method=@MethodDescriptor(name="createOrderDetailsTypeDescription"))
List<JAXBElement<String>> listOfStringToListOfJaxbElement(List<String> strings);

@interface MethodDescriptor {
    Class<?> declaredOn() default void.class;
    String name() default "";
    Class<?>[] parameterTypes default {};
}
```

that could be used now #3 has been resolved.

sjaakd:
@filiphr thanks for reviewing

PR #1019, opened 2016-12-24
#285 JAXB selection mechanism for List

sjaakd:
XmlElementDeclSelector should receive its context from the calling property mapping. This is possible for forged methods. However for non-forged methods we should find another solution. Perhaps extend the annotation with jaxbCallingElemName or something along that lines.. Or leave it as is.

The equals methods needed to be implemented since which factory is called in a forged method is a discriminating element in its uniqueness.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/Field.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/Field.java
@@ -105,4 +105,27 @@ public class Field extends ModelElement {
         this.typeRequiresImport = typeRequiresImport;
     }
 
+    @Override
+    public int hashCode() {
+        int hash = 5;
+        hash = 43 * hash + (this.variableName != null ? this.variableName.hashCode() : 0);
+        return hash;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if ( this == obj ) {
+            return true;
+        }
+        if ( obj == null ) {
+            return false;
+        }
+        if ( getClass() != obj.getClass() ) {
+            return false;
+        }
+        final Field other = (Field) obj;
+        return !( (this.variableName == null) ?
+            (other.variableName != null) : !this.variableName.equals( other.variableName ) );
+    }
+
 }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
@@ -64,6 +64,7 @@ public class IterableMappingMethod extends MappingMethod {
         private FormattingParameters formattingParameters;
         private NullValueMappingStrategyPrism nullValueMappingStrategy;
         private ForgedMethod forgedMethod;
+        private String callingContextTargetPropertyName;
 
         public Builder mappingContext(MappingBuilderContext mappingContext) {
             this.ctx = mappingContext;
@@ -90,6 +91,11 @@ public class IterableMappingMethod extends MappingMethod {
             return this;
         }
 
+        public Builder callingContextTargetPropertyName(String callingContextTargetPropertyName) {
+            this.callingContextTargetPropertyName = callingContextTargetPropertyName;
+            return this;
+        }
+
         public IterableMappingMethod build() {
 
 
@@ -111,7 +117,7 @@ public class IterableMappingMethod extends MappingMethod {
             Assignment assignment = ctx.getMappingResolver().getTargetAssignment(
                 method,
                 targetElementType,
-                null, // there is no targetPropertyName
+                callingContextTargetPropertyName,
                 formattingParameters,
                 selectionParameters,
                 sourceRHS,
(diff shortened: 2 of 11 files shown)


sjaakd:
should probably also update the integration test.

processor/src/main/java/org/mapstruct/ap/internal/model/Field.java
```diff
@@ -105,4 +105,33 @@ public void setTypeRequiresImport(boolean typeRequiresImport) {
         this.typeRequiresImport = typeRequiresImport;
     }
 
+    @Override
+    public int hashCode() {
```
filiphr: Why only use the `variableName` in the `hashCode`? In `equals`, both the `variableName` and the `type` are used.

processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
```diff
@@ -379,6 +385,15 @@ else if ( other.selectionParameters != null ) {
             return false;
         }
 
+        if ( this.factoryMethod != null ) {
```
filiphr: Can you use `if (!Objects.equals(this.factoryMethod, other.factoryMethod))`? I know that we already have the same `if`s for the `selectionParameters`, but I find it much clearer when it is one liner with `Objects.equals`.

processor/src/main/java/org/mapstruct/ap/internal/model/MapMappingMethod.java
```diff
@@ -367,6 +367,15 @@ public boolean equals(Object obj) {
             }
         }
 
+        if ( this.factoryMethod != null ) {
```
filiphr: Same here.

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
```diff
@@ -248,4 +248,37 @@ public String toString() {
...
+            return true;
+        }
+        if ( obj == null ) {
+            return false;
+        }
+        if ( getClass() != obj.getClass() ) {
+            return false;
+        }
+        final MappingMethod other = (MappingMethod) obj;
+        if ( (this.name == null) ? (other.name != null) : !this.name.equals( other.name ) ) {
```
filiphr: Maybe for this `if`s we can use `Objects.equals`

processor/src/main/java/org/mapstruct/ap/internal/model/MethodReference.java
```diff
@@ -221,4 +221,30 @@ public boolean isCallingUpdateMethod() {
...
+            return true;
+        }
+        if ( obj == null ) {
+            return false;
+        }
+        if ( getClass() != obj.getClass() ) {
+            return false;
+        }
+        final MethodReference other = (MethodReference) obj;
+        if ( this.declaringMapper != other.declaringMapper && (this.declaringMapper == null
```
filiphr: Maybe here as well use `Objects.equals`

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
```diff
@@ -248,4 +248,37 @@ public String toString() {
     public List<ForgedMethod> getForgedMethods() {
         return forgedMethods;
     }
+
+    @Override
+    public int hashCode() {
+        int hash = 3;
+        return hash;
```
filiphr: Is it OK if we have a static hash?

processor/src/main/java/org/mapstruct/ap/internal/model/Field.java
sjaakd: I thought about that.. There's nothing more unique in the mapper class  on that level then the variables. The used mapper is a variable. Type would strictly be not needed.

processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
sjaakd: I thought about it when I wrote the code. Its perhaps more clear. At the other hand, I let the IDE do its work. I reckon its better at "not making errors" than I am :smile:. I can change it however. WDYT?

sjaakd:
@filiphr thanks for reviewing.

> Otherwise, I do not have a lot of experience with JAXB so I cannot say if it is ok or not. Code wise (apart of my comments) it is OK 😄.

Its just another selection parameter we hand down from the calling method to the mapping method. For forged methods we do have this context and this makes sense.

processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
filiphr: I just tried with my IDE, it didn't use `Objects.equals`. I also realized that `Objects` is since Java 7, so we can't use it 😄. Therefore it is OK as it is

processor/src/main/java/org/mapstruct/ap/internal/model/Field.java
filiphr: OK

filiphr:
>should probably also update the integration test.

Maybe also to this? 

What about the last comment (I deleted the ones with `Objects.equals`)?

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
sjaakd: Good question.  I will fix this.

sjaakd:
> What about the last comment (I deleted the ones with Objects.equals)?

No worries. I did see it. I'll add another commit this evening and merge later tonight.

Issue #1013, opened 2016-12-22
Errors for @Inherit(Reverse)Configuration based on non-abstract methods


sjaakd:
Several matching methods exist: merge(), mergeWithOutInsertedPart(). Specify a name explicitly.

`merge` is non abstract.

PR #1020, opened 2016-12-25
#1013 Inherit(Reverse)Configuration must only consider abstract methods

sjaakd:


filiphr:
@sjaakd this will also not consider `abstract` and/or other interface methods right? It will only consider methods that are annotated with `@Mapping`/`@Mappings`? 

Maybe people have 2 configurations, one default where everything is mapped and another one where some fields are ignored. Then the first one will no longer be selected, right?

sjaakd:
> @sjaakd this will also not consider abstract and/or other interface methods right? It will only consider methods that are annotated with @Mapping/@Mappings?

That's true. I act on the presence / absence of mappings (in general, so also iterable and map mappings). I reckon this is what we really want the check here. The idea also being that the fact that they are abstract should be checked at the moment that we fetch the `@Mapping` annotation.


> Maybe people have 2 configurations, one default where everything is mapped and another one where some fields are ignored. Then the first one will no longer be selected, right?

You have to indicate which one by pointing out a name. The `@InheritConfiguration` and `@InheritReverseConfiguration` have a parameter to resolve conflicts.

filiphr:
>You have to indicate which one by pointing out a name. The @InheritConfiguration and @InheritReverseConfiguration have a parameter to resolve conflicts.

Aren't we breaking the conflict with this one? The default one does not have `@Mapping` it just takes all, and with this PR it will never be considered as something that can be inherited.

sjaakd:
Perhaps I don't understand you. In general there is nothing to inherit if there are no mappings? Or do you mean that you would inherit the policy for unmapped target properties somehow for a single mapping method?

filiphr:
What I meant is the following we have this:

```java
public interface FullMappingMapper {
    CarEntity toCarEntity(CarDto carDto);
}

public interface PartialMappingMapper {
    @Mapping( target = "color", ignore = true)
    CarEntity toCarEntity(CarDto carDto);
}

@Mapper( uses = { FullMappingMapper.class, PartialMappingMapper.class } )
public interface MainMappingMapper {

    @InheritConfiguration(name = "toCarEntity")
    @Mapping(target = "auditTrail", constant = "fixed")
    CarEntity toCarEntityWithFixedAuditTrail(CarDto carDto);
}
````

What do we expect for this? What should happen?

Aside from my example I played around with the PR and my example will actually not work at all (no inheritance will be found), because in `MethodRetrievalProcessor.getReferencedMethod` during the build of the `SourceMethod` no `mappings` are passed. Is this expected or it is a side effect?

**Comment by sjaakd** [comment]
The `@InheritConfiguration` is only supposed to inherit the mappings. Both source an target should have an inheritance relation to the prototype methods.  If there are no mappings, it does not make sense to inherit.  Now, the idea is that you can specify those method as well on an interface marked as @MappingConfiguration.

**Comment by filiphr** [comment]
The `@InheritConfiguration` yes you are right, for my example it will only be picked up if the Mappers are registered via `Mapper.config`. That one only works if you register the configuration via `@Mapper(config=...)`. However, I think that the same example with `@InheritInverseConfiguration` instead should fail. Do you agree? Maybe I am understanding something wrong...

**Comment by sjaakd** [comment]
`@InheritConfiguration` and `@InheritInverseConfiguration` should work the same (only reverse target and source. But `@InheritInverseConfiguration` has more legacy. It could be that it picks up inverse mappers as well via the `uses`. We should align that behavior if that differs (and I'm in favor of the wider interpretation).  There's also some cycle risk. Pick up inherit-inverse-inherit etc. Lets discuss and solve that  in   separate issue. Could you write that? Perhaps there's already one.

As far as this issue is concerned. Has it introduced this behavior or was it already there? Otherwise I think we should merge / close this one.

**Comment by filiphr** [comment]
That's the thing this one no longer picks up inverse configuration from Mappers in `uses` (even if they are annotated), and I think that we should not break that.

File: processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java [hunk 1]
```diff
@@ -183,9 +183,15 @@ public Builder setDefininingType(Type definingType) {
...
 
-            MappingOptions mappingOptions =
-                new MappingOptions( mappings, iterableMapping, mapMapping, beanMapping, valueMappings );
-
+            MappingOptions mappingOptions = null;
+            if ( (mappings != null && !mappings.isEmpty())
+                || iterableMapping != null
+                || mapMapping != null
+                || beanMapping != null
+                || valueMappings != null ) {
```
filiphr: Here when the `valueMappings` are not null they can be empty. What exactly we need to check here? 

If we don't generate `mappings`. Then `@InheritInverseConfuguration` will _only_ pick up annotated methods from the same mapper as well.

sjaakd: I'll check it out.. And valueMappings empty should indeed be checked.

sjaakd: > That's the thing this one no longer picks up inverse configuration from Mappers in uses (even if they are annotated), and I think that we should not break that.

I tried the same.. By adding a new mapper: `UsedMapper` to `CarMapperWithAutoInheritance`: 

```java
@Mapper(
    uses = UsedMapper.class,
    config = AutoInheritedConfig.class
)
public interface CarMapperWithAutoInheritance {
    CarMapperWithAutoInheritance INSTANCE = Mappers.getMapper( CarMapperWithAutoInheritance.class );

    @Mapping(target = "color", source = "colour")
    CarEntity toCarEntity(CarDto carDto);

    @InheritInverseConfiguration(name = "toCarEntity2")
    CarDto toCarDto(CarEntity entity);

    @Mappings({
        @Mapping(target = "color", source = "colour"),
        @Mapping(target = "auditTrail", constant = "fixed")
    })
    CarEntity toCarEntityWithFixedAuditTrail(CarDto carDto);

    @Mapping(target = "color", source = "colour")
    void intoCarEntityOnItsOwn(CarDto carDto, @MappingTarget CarEntity entity);

    @InheritConfiguration(name = "toCarEntity")
    void intoCarEntity(CarDto carDto, @MappingTarget CarEntity entity);
}
```

```java
@Mapper
public interface UsedMapper {

    @Mapping(target = "color", source = "colour")
    CarEntity toCarEntity2(CarDto carDto);

}
```

And guess.. Null pointer.. :smile:  My best guess is that this method does not belong to this mapper. Since mappers are generated one (and this method does not belong to the mapper at hand) I'm asking myself the question whether we should allow this uses in relation to `@InheritInverseConfiguration`

sjaakd: see issue #1027

sjaakd:
@filiphr I played a bit around. Seemed that by only looking at mappings we miss the option that a method that has been selected has an `@InheritConfiguration` or `@InheritInverseConfiguration` (and no mapping on its own). So I went for my initial solution (only selecting non-abstract methods).

sjaakd:
Hmm. perhaps better to move the test methods from `NotUsedMapper` to the mapper itself. WDYT? Then I leave room to introduce a `NotUsedMapper` for #1027

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
```diff
@@ -590,6 +592,11 @@ public boolean isBeforeMappingMethod() {
         return Executables.isBeforeMappingMethod( getExecutable() );
     }
 
+    public boolean isAbstract() {
```
filiphr: This one needs to represent whether an implementation will be created for this method, right? 

Why don't we do `return overridesMethod() || !isDefault()`? Or maybe we need to do something else, as this will pick up Java 8 default methods. I think that we need to consider the Java 8 default methods the same way as we consider non abstract methods.

filiphr:
Sorry for my late response.

>Hmm. perhaps better to move the test methods from NotUsedMapper to the mapper itself. WDYT? Then I leave room to introduce a NotUsedMapper

If you do this then you can no longer use the same name right? I thought the problem was that methods with the same name were picked up.

sjaakd:
> If you do this then you can no longer use the same name right? I thought the problem was that methods with the same name were picked up.

You can always change the signature. E.g. add a (new) Context parameter or make it an update method..

filiphr:
>You can always change the signature. E.g. add a (new) Context parameter or make it an update method..

Well then it will work even without your change see `SourceMethod#reverse(SourceMethod)` and `SourceMethod#canInheritFrom(SourceMethod)`. Maybe I am misunderstanding something 😄. If you can do what you are saying do it.

sjaakd:
> Well then it will work even without your change see SourceMethod#reverse(SourceMethod) and SourceMethod#canInheritFrom(SourceMethod). Maybe I am misunderstanding something 😄. If you can do what you are saying do it

It is about getting 2 methods to qualify. And then discard the one which is not abstract. I added the reproducer. You can test it with - and without patch. Please do so. Just to confirm that we're on the same page :).

filiphr:
@sjaakd You were right. I made a mistake with the `getSourceParameters()`.

Did you also saw my comment for the new `isAbstract()` method?

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
sjaakd: `overridesMethod` would not work for methods defined in the configuration class.  `declaringMapper.getTypeElement().getKind().isInterface()` captures default methods as well, right?

sjaakd:
> Did you also saw my comment for the new isAbstract() method?

Missed it.. But answered it now :smile:

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
sjaakd: although `isAbstract` might then not be a good name.. I'll change that to `isAbstractOrDefault`.

sjaakd:
@filiphr are you ok with the answers?

filiphr:
I am on my mobile now. It made sense when I wrote my comment, I need to see the sources again. However, you can rename the method and go with it. I am OK with the answe

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
```diff
@@ -602,6 +604,14 @@ public boolean isBeforeMappingMethod() {
         return Executables.isBeforeMappingMethod( getExecutable() );
     }
 
+    /**
+     * @return returns true for interface methods (see jls 9.4) lacking a default or static modifier and for abstract
+     * methods
+     */
+    public boolean isAbstract() {
+        return executable.getModifiers().contains( Modifier.ABSTRACT );
```
filiphr: This is really nice, I didn't know this about interfaces 😄. We can maybe use the same construct on other places as well.

Issue #1007, opened 2016-12-20
Compare generated code to a fixed 1.1 baseline


sjaakd:
To avoid problems in generated code java files we should compare a selective set to a set of generated 1.1. generated source code java files. See #996 for more details.

filiphr:
Add tests for:

- [ ] org.mapstruct.ap.test.bugs._913.Issue913SetterMapperForCollectionsTest
- [x] org.mapstruct.ap.test.array
- [ ] org.mapstruct.ap.test.collection.adder
- [ ] org.mapstruct.ap.test.updatemethods
- [ ] org.mapstruct.ap.test.conversion.string

Are there any other tests that we need to add?

agudian:
I'd say create the merge commit locally, add the test fixes to  the test files and push to master 🙂.

agudian:
@filiphr I started merging those commits, as I would need to merge some other stuff soon as well... :)

agudian:
While merging, I found only improvements (yay!), with the exception to this change for an update-method (`ExternalSelectionTest`, in the `DepartmentMapperImpl`):

with 1.1, we generated the following:
```java
        if ( entity.getEmployees() == null ) {
            entity.setEmployees( new ArrayList<EmployeeEntity>() );
        }
        externalHandWrittenMapper.toEmployeeEntityList( dto.getEmployees(), entity.getEmployees() );
```

With the current master, we generate:
```java
        if ( dto.getEmployees() != null ) {
            if ( entity.getEmployees() == null ) {
                entity.setEmployees( new ArrayList<EmployeeEntity>() );
            }
            externalHandWrittenMapper.toEmployeeEntityList( dto.getEmployees(), entity.getEmployees() );
        }
        else {
            entity.setEmployees( null );
        }
```

That seems to be a good idea, but it definetly changes the outcome. There might be users who currently rely on the old behavior, so that should definitely be part of the migration notes. @sjaakd, did you make that change? Would you add something to the migration-notes wiki?

agudian:
Merged to master. Thanks, @filiphr!

sjaakd:
Sure ! I'll add that as soon as I've got the opportunity! Thanks 

Verstuurd vanaf mijn iPhone

PR #1023, opened 2016-12-27
#1007 add base for comparing generated code

filiphr:
This is just to start a discussion whether the approach for comparing the generated code against a fixed baseline.

Open questions:
- [x] Do we want to compare only eclipse to eclipse and javac to javac, or we expect to generate the same code for eclipse and javac? Yes, it is OK to expect the same file for javac and eclipse compilers.
- [x] Is it OK that we suppress the date and comment in the `@Generated`? No it is not. We are going to ignore those lines during the comparision

In #1002 these tests were written:
* org.mapstruct.ap.test.bugs._913.Issue913SetterMapperForCollectionsTest
* org.mapstruct.ap.test.array
* org.mapstruct.ap.test.collection.adder
* org.mapstruct.ap.test.updatemethods
* org.mapstruct.ap.test.conversion.string

Does something else make sense to be added as well?

I will add all the tests once we agree that the baseline for comparing is OK. We can also do that in a separate PR


sjaakd:
@filiphr looks good. Perhaps that @agudian should still take a look a `CompilingStatement`. I was thinking: should we have an explicit marker (such as you now have in the test case or should we simply check in the resources folder and  do the assertion when a file is present there?

agudian:
Great that you're picking this up, Filip!

I think the `@Generated` annotation should be generated as it is and the date/compiler dependent contents should be ignored in the comparison. It's totally okay to expect that javac and Eclipse generate the same file content.

Also, I would let the comparison ignore any formatting / whitespace changes. The formatting itself should be covered by our checkstyle check - the basic requirement for the checks here was to ensure that we don't break something functionally (e.g. missing null-checks or something). Plus, the comparison should ignore the legal headers that we could then add again to the committed files... :)

And as Sjaak suggests, the Test should fail if a comparison is expected to be done, but no reference file is there.

So I'd rather create a new `Rule` or extend the `GeneratedSource` rule to specifically include it in the test, e.g.:

```java
    @Rule
    public final GeneratedSource generatedSource = new GeneratedSource().withComparisonToFixtureFor( ScienceMapper.class );
```

WDYT?

filiphr:
Thanks for the comments guys.

@agudian I went with directly comparing the files as the error message from assertj was really nice (it was outputting the lines where something is not OK 😄). I will some regex for the date and comment in the `@Generated` and also about the legal headers.

Regarding the formatting/whitespace do you think that we should collapse everything into one line and compare that one, or per line check is enough?

>So I'd rather create a new Rule or extend the GeneratedSource rule to specifically include it in the test

This was my initial idea as well. However, I struggled with tests that have `@WithClasses` on methods. I guess that we can add a new Mapper to the `Rule` in a method in this case. 

I will update the PR with the new ideas

agudian:
> Regarding the formatting/whitespace do you think that we should collapse everything into one line and compare that one, or per line check is enough?

Yes, that should be enough. We can handle the rest of the differences by reformatting the fixture files manually in case something dramatic changes.

> This was my initial idea as well. However, I struggled with tests that have @WithClasses on methods. I guess that we can add a new Mapper to the Rule in a method in this case.

👍

filiphr:
I switched to using a `Rule` for the assertion instead of the `CompilingStatement`.

I was thinking about what kind of comparison to do, I think that it is OK to not ignore whitespaces/formatting, as when something changes in our formatting we will need to update the expected files and we can do this by replacing the expected with the new generated mapper.

WDYT?

processor/src/test/java/org/mapstruct/ap/testutil/runner/GeneratedSource.java
```diff
@@ -58,16 +86,35 @@ static void clearCompilingStatement() {
...
+
+    /**
+     * Adds more mappers that need to be compared.
+     *
+     * The comparison is done for mappers and the are compared against a Java file that matches the name of the
+     * Mapper that would have been created for the fixture.
+     *
+     * @param fixturesFor the classes that need to be compared with
+     */
+    public void addComparisonToFixtureFor(Class<?>... fixturesFor) {
```
agudian: I would guess we can make the rule usage more expressive with chaining calls to this method instead of passing the classes in the constructor.

processor/src/test/java/org/mapstruct/ap/testutil/runner/GeneratedSource.java
```diff
@@ -86,4 +133,38 @@ public JavaFileAssert forDecoratedMapper(Class<?> mapperClass) {
...
+        public void evaluate() throws Throwable {
+            next.evaluate();
+            handleFixtureComparison();
+        }
+    }
+
+    private void handleFixtureComparison() throws UnsupportedEncodingException {
+        for ( Class<?> fixture : fixturesFor ) {
+            String expectedFixture = getMapperName( fixture );
+            URL expectedFile = getClass().getClassLoader().getResource( expectedFixture );
```
agudian: How about we put the fixture files in `src/test/resources/fixtures` or so, to clarify the purpose of the .java files that we're putting there?

processor/src/test/java/org/mapstruct/ap/testutil/runner/GeneratedSource.java
filiphr: You are right. I will do that. Just one question should we use varargs just one argument?

processor/src/test/java/org/mapstruct/ap/testutil/runner/GeneratedSource.java
filiphr: That's a nice idea, I'll do that.

processor/src/test/java/org/mapstruct/ap/testutil/runner/GeneratedSource.java
agudian: I would have done it with one argument, but like your idea with the varargs!

filiphr:
Thanks for the review @agudian. I added chaining and `fixtures` root for the java files. 

If it is OK for you we can merge it and then add more mappers for testing in another PR.

processor/src/test/java/org/mapstruct/ap/testutil/runner/GeneratedSource.java
```diff
@@ -46,32 +46,15 @@
  */
 public class GeneratedSource implements TestRule {
 
+    private final static String FIXTURES_ROOT = "fixtures/";
```
agudian: A checkstyle error crept in here. Otherwise it's okay to go. 👍
Could you correct this while squashing / merging?

filiphr: I saw the build failure as well 😄. I fixed and I will squash and merge once the build is done

PR #1026, opened 2016-12-28
Attempt to use the Eclipse formatter on created Java files. 

agudian:
Breaks JDK 6 compatibility, though.

Hum, this took some fiddling, as we need to shade org.eclipse.jdt for the formatting, but also not for the Eclipse specific workarounds. Oh and the not-latest versions of `asm` created byte-code errors in some of the Eclipse AST classes... 🙄 

The downsides:
* boosts the processor fat-jar from 1,5 MB to 9 MB (already minimized, but there's still some slug ending up in the jar that we don't need...)
* in a fresh classloader, there's a considerable lag when doing the formatting of the first file. Doesn't add much to the total test execution time, though.
* I'm currently using the latest eclipse.jdt version for the formatter as well, which requires a JRE 7. We could use an older version just for the formatter... but it would create some trouble with executing tests in the IDE (where we only have the non-shaded dependencies around - we could release the formatter seperately, though, so that could be made work as well).

The upsides:
* we could remove quite a lot of fiddling and workarounds for style related stuff in the templates (e.g. missing NL at EOF, compress-statements, whitespace stripping, ...).
* we can configure the style of the generated code pretty accurately, including breaking long lines properly.
* we could even add easy support to pick up user-defined formatter configuration, if that is of any interest at all for anyone... 😉 

So what do you guys think, should I continue with this? Is it worth the effort?


sjaakd:
> So what do you guys think, should I continue with this? Is it worth the effort?

I honestly don't know. You spent considerable effort looking at the above.. But I think that we better invest in getting our code formatted nicely by default (without running a formatter). One of the risks I see is that we abandon the practice of having our templates neat and tidy (can we still see this?) and trust the eclipse formatter to do the work.

As an upside to this, I understand you package this. Can the user define its own templates for formatting the code?

agudian:
> One of the risks I see is that we abandon the practice of having our templates neat and tidy (can we still see this?) and trust the eclipse formatter to do the work.

That's actually one of the benefits that I see, as we're already spending a considerable time fiddling with the templates (and the indentation correcting writer - just look at the generated code without that writer in the pipeline 😉), and we will have to continuously do that. Adding support for multi-line statements might become very tricky as well. Having some proper formatter cleaning up would save us all that trouble and we could concentrate on the content and not the format in our templates.
I'd rather spend another one or two hours on this rather than have us fine-tune ftl files. But that's just my opinion on one of the points above 😁

> Can the user define its own templates for formatting the code?

It would be pretty simple to let the user override the formatter settings - it's just some properties that need to be added to a map... 😉

gunnarmorling:
I think the largely increased package size is not worth doing this. My hope is we can do it correcty ourselves with the formatting writter. So far it worked quite well, but admittedly it's not so easy for multi-line statements.

One idea: How about making it a functionality of the Eclipse plug-in to invoke the formatter after the MapStruct AP has generated some code?

filiphr:
>One idea: How about making it a functionality of the Eclipse plug-in to invoke the formatter after the MapStruct AP has generated some code?

I am not sure this makes sense. Yes the files will look good for the user in Eclipse. However, usually the sources are build in CI and there we won't have the Eclipse plug-in. The delivered files in the end, won't look the same way as in the IDE.

I can suggest the following. Why don't we make the formatting via an SPI, or something else that can be plugged in? This way the formatter will be a separate jar and if the user wants to use he/she can explicitly add it to the classpath. This way we don't increase the core mapstruct processor size and we still provide a way for the user to be able to nicely format the files, also with their own formatting guidelines.

gunnarmorling:
That's an interesting idea, I could get behind that.

2017-01-06 9:46 GMT+01:00 Filip Hrisafov <notifications@github.com>:

agudian:
I don't know, guys. At least for me the goal behind this was not to allow users to apply the formatting of their liking (it would have been a low hanging fruit to offer it as well, though).
My goal was to once and for all remove the need to manually tweak the ftl files for a decently formatted output by default. We just recently reverted a nicely readable Type.ftl back to an unreadable one-liner because the attempt to make it readable inadvertently added a whitespace between a type name and the `<` for the generics parameters. And stuff like that happens every once in a while. 😕
Perhaps we can find another formatter or even build our own parser / pretty printer?

agudian:
Uhm, just to be clear: I'm also okay with just leaving it as we have it today and keep an eye on the generated sources. Together with the checkstyle checks and the fixture files for several test mappers to compare against we aren't in a too bad shape for finding new issues.

But I guess for you guys the downsides from the list above outweigh the upsides, right?

gunnarmorling:
So my take on this is that adding that big dependency is a no-go.

It's not that we have a huge issue here, the indentation issue with multi-line invocations being the only real one coming to my mind right now (and I guess some templates may look nicer if we had smarter formatting after the fact).

My hope would be that we can find a way to do formatting well enough ourselves.

gunnarmorling:
So should we close this one?

agudian:
Yes, be gone with it 🙂

PR #1028, opened 2016-12-30
#1007 additional generated code tests

filiphr:
This are all the tests which are in #1007. Do we need to add more tests?

### DIFF
--- a/processor/src/test/resources/fixtures/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNcvsAlwaysMapperImpl.java
+++ b/processor/src/test/resources/fixtures/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNcvsAlwaysMapperImpl.java
@@ -1,0 +1,252 @@
+/**
+ *  Copyright 2012-2016 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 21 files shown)


processor/src/test/resources/fixtures/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNcvsAlwaysMapperImpl.java
```diff
@@ -0,0 +1,252 @@
...
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.annotation.Generated;
+
+@Generated(
+    value = "org.mapstruct.ap.MappingProcessor",
+    date = "2016-12-30T19:07:28+0100",
+    comments = "version: , compiler: javac, environment: Java 1.8.0_112 (Oracle Corporation)"
```
agudian: Which version did you use to generate these files? Was it 1.1?

agudian: 🙄 forget it, I just saw that this is based on the 1.1 already, so what else should it be... 😁

Will you merge everything back to master when the additional tests were added?

filiphr: >Which version did you use to generate these files? Was it 1.1?

Yes it was 1.1, and why the version isn't there is strange to me.

>Will you merge everything back to master when the additional tests were added?

I was planning to merge it locally add a commit for fixing the test failures (there will be as the generated code is changed) and add a PR. Or would you prefer we merge into master with test failures and then I provide a PR with the fixes only?

filiphr:
Sorry for my late response 😄. I will merge it into 1.1-master, and proceed as we agree for the merge into master

Issue #833, opened 2016-07-14
'Unknown property' error raised for ignored targets with no setter


bpiper:
If I have a Mapping that specifies a target property as ignored, and there is no setter for that property in the target class, I am getting an `Unknown property "myPropertyName" in return type` error raised.  This is not what I would expect if I have indicated that the property is to be ignored.

In `BeanMappingMethod.handleDefinedSourceMappings`, it is attempting to resolve the `targetWriteAccessor` for a mapping even though the mapping might be ignored.  The `targetWriteAccessor` is not actually used at all if the `mapping.isIgnored()` condition further down evaluates to true, hence there doesn't seem to be a legitimate reason for resolving the target accessor on an ignored mapping.

The workaround for this is of course to define a setter for the ignored property (but have it throw an exception, because I don't want anything actually calling it).  This isn't ideal though, as it could invite other things to use the setter instead of modifying the field via reflection.

agudian:
I'd say what you describe is not a problem but it is how it's supposed to work.
Read-only properties shouldn't need to be explicitly ignored anyway. If you get a warning for an unmapped property but that property is read-only, then my feeling is that there lies the bug instead. Do you get such a warning if you don't mention that property in the annotations?

bpiper:
Yes, you're quite right... I hadn't realised that target properties with no mutator would be implicitly ignored.  I'd lumped a couple of read-only properties in with a couple of writeable properties that I also wanted to ignore and assumed that Mapstruct would complain if I didn't... but turns out it doesn't.  

Arguably it would still be worth printing a more useful warning or error if someone adds an ignore mapping for a read-only property (e.g. "Target mapping for 'myProperty' with ignore = true is unnecessary as the property is read-only"), rather than implying that they need to add a mutator for that property... but I'm otherwise happy to close this.

sjaakd:
> Arguably it would still be worth printing a more useful warning or error if someone adds an ignore mapping for a read-only property

Good suggestion. Although, I reckon its a more broader case in which you specify a `@Mapping` that has no write accessor but does have a read accessor.

agudian:
Sjaak added a better message for this case, closing it. 
@bpiper thanks for the report!

PR #1031, opened 2017-01-05
#1029 Don't show an error when explicitly ignoring read-only properties

### LINKED ISSUE #1029: Mapper should not throw an error when explicitly ignoring read-only properties
screen shot 2017-01-03 at 2 35 43 pm

Ran the command `./gradlew assemble` and the annotation processor threw these warnings. Andreas Gudian suggested that they should not do so.

The unmapped properties are generated via the entities' constructors and are marked with `@Transient.` They do not contain accessor methods.

Thanks for reading.

------
Edit by @agudian to clarify what problem we're actually trying to fix here:

Read-Only properties are already ignored automatically and don't show up as unmapped target properties. But explicitly ignoring a read-only property using `@Mapping(target=.., ignore=true)` leads to the error "property X has no write accessor" and supresses any warnings about further unmapped target properties.

agudian:
Fixes #1029.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -255,17 +255,31 @@ public class BeanMappingMethod extends MappingMethod {
                         resultPropertyName = first( targetRef.getPropertyEntries() ).getName();
                     }
 
-                    if (!unprocessedTargetProperties.containsKey( resultPropertyName ) ) {
-                         boolean hasReadAccessor =
+                    if ( !unprocessedTargetProperties.containsKey( resultPropertyName ) ) {
+                        boolean hasReadAccessor =
                             method.getResultType().getPropertyReadAccessors().containsKey( mapping.getTargetName() );
-                        ctx.getMessager().printMessage( method.getExecutable(),
-                            mapping.getMirror(),
-                            mapping.getSourceAnnotationValue(),
-                            hasReadAccessor ? Message.BEANMAPPING_PROPERTY_HAS_NO_WRITE_ACCESSOR_IN_RESULTTYPE :
+
+                        if ( hasReadAccessor ) {
+                            if ( !mapping.isIgnored() ) {
+                                ctx.getMessager().printMessage(
+                                    method.getExecutable(),
+                                    mapping.getMirror(),
+                                    mapping.getSourceAnnotationValue(),
+                                    Message.BEANMAPPING_PROPERTY_HAS_NO_WRITE_ACCESSOR_IN_RESULTTYPE,
+                                    mapping.getTargetName() );
+                                errorOccurred = true;
+                            }
+                        }
+                        else {
+                            ctx.getMessager().printMessage(
+                                method.getExecutable(),
+                                mapping.getMirror(),
+                                mapping.getSourceAnnotationValue(),
                                 Message.BEANMAPPING_UNKNOWN_PROPERTY_IN_RESULTTYPE,
-                            mapping.getTargetName()
-                        );
-                        errorOccurred = true;
+                                mapping.getTargetName() );
+                            errorOccurred = true;
+                        }
+
                         continue;
                     }
 
--- a/processor/src/test/java/org/mapstruct/ap/test/ignore/ErroneousTargetHasNoWriteAccessorMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/ignore/ErroneousTargetHasNoWriteAccessorMapper.java
@@ -32,7 +32,7 @@ public interface ErroneousTargetHasNoWriteAccessorMapper {
     ErroneousTargetHasNoWriteAccessorMapper INSTANCE =
         Mappers.getMapper( ErroneousTargetHasNoWriteAccessorMapper.class );
 
-    @Mapping( target = "hasTallons", ignore = true )
+    @Mapping(target = "hasClaws", constant = "true")
     PreditorDto preditorToDto( Preditor preditor );
 
 }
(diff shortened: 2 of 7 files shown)


agudian:
PRB fails due to outdated license header template - needs #1033 to be merged first.

filiphr:
It looks OK to me

filiphr:
>PRB fails due to outdated license header template - needs #1033 to be merged first.

No you need to update `IgnorePropertyTest.shouldGiveWringingOnUnmapped` as now we do not fail the build if a property with a read accessor is ignored. To fail the build the property should not be ignored.

agudian:
> No you need to update IgnorePropertyTest.shouldGiveWringingOnUnmapped as now we do not fail the compilation if a property with a read accessor is ignored. To fail the build the property should not be ignored.

Yea, that as well. Hmm. It was introduced as part of #833 - but I still think that we should not error out there, but just happily ignore it if the user want's to ignore it... 😁 - so I'll remove that test.

@sjaakd, please shout if I'm missing something here (when you're back, that is).

filiphr:
>Yea, that as well. Hmm. It was introduced as part of #833 - but I still think that we should not error out there, but just happily ignore it if the user want's to ignore it... 😁 - so I'll remove that test.

I agree with you. I think that we should error out if the user does not ignore the property though. We don't have a test where a user targets some property (not ignoring it) without write accessor.

Something like `@Mapping(target = "outdated", constant="true")`

agudian:
> Something like `@Mapping(target = "outdated", constant="true")`

Good idea! I've changed the old test to do just that.

processor/src/test/java/org/mapstruct/ap/test/ignore/ErroneousTargetHasNoWriteAccessorMapper.java
```diff
@@ -32,7 +32,7 @@
     ErroneousTargetHasNoWriteAccessorMapper INSTANCE =
         Mappers.getMapper( ErroneousTargetHasNoWriteAccessorMapper.class );
 
-    @Mapping( target = "hasTallons", ignore = true )
+    @Mapping(target = "hasTallons", constant = "true")
```
gunnarmorling: [WTH](http://www.urbandictionary.com/define.php?term=tallon)?!

gunnarmorling:
Reading the title of #833, isn't this exactly the same as this one?

processor/src/test/java/org/mapstruct/ap/test/ignore/ErroneousTargetHasNoWriteAccessorMapper.java
agudian: Dutch humor? 😁

gunnarmorling: I'm shocked ;)

agudian:
> Reading the title of #833, isn't this exactly the same as this one?

Kinda. #833 introduced the error message that is shown here. The message is fine if a user tries to configure the mapping for a read-only target. But we missed that ignoring a read-only target should not lead to an error and suppressing the rest of the unmapped properties warnings.
So for this particular case here, we think it's better to leave the user alone and don't bother with a warning or error for a property that we're already ignoring automatically, even if the user explicitly adds the `ignore=true` for that read-only property.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -255,17 +255,19 @@ private boolean handleDefinedSourceMappings() {
...
                             method.getResultType().getPropertyReadAccessors().containsKey( mapping.getTargetName() );
-                        ctx.getMessager().printMessage( method.getExecutable(),
-                            mapping.getMirror(),
-                            mapping.getSourceAnnotationValue(),
-                            hasReadAccessor ? Message.BEANMAPPING_PROPERTY_HAS_NO_WRITE_ACCESSOR_IN_RESULTTYPE :
-                                Message.BEANMAPPING_UNKNOWN_PROPERTY_IN_RESULTTYPE,
-                            mapping.getTargetName()
-                        );
-                        errorOccurred = true;
+                        if ( !hasReadAccessor || !mapping.isIgnored() ) {
```
gunnarmorling: Can we de-compose this into two separate ifs? Together with the second one on `hasReadAccessor` inline I find i a bit tough to digest.

gunnarmorling:
Yes, that makes sense. If a target property is read-only, setting `ignore=true` shoudn't raise any warning or error. Any other explicit configuration for that property should result in an error.

agudian:
> Can we de-compose this into two separate ifs? Together with the second one on hasReadAccessor inline I find i a bit tough to digest.

I added a commit for that...

sjaakd:
Hmm,  and I thought I knew my synonyms! 😄. In my world it used to be the claw of a bird of prey. I'm getting old !! 😉

Verstuurd vanaf mijn iPhone

> Op 5 jan. 2017 om 22:35 heeft Gunnar Morling <notifications@github.com> het volgende geschreven:
> 
> I'm shocked ;)

filiphr:
The old test is OK. If you ask me you can merge it

processor/src/test/java/org/mapstruct/ap/test/ignore/ErroneousTargetHasNoWriteAccessorMapper.java
sjaakd: Me too :smile:. For the record, it should be [talon](https://en.wikipedia.org/wiki/Claw). With one 'l'. Although, that would not get me home-free in the urban dictionary either 😄

agudian: I'll change it to something less flowery... 🦄

agudian:
I've squashed / pushed a new commit with reworded properties in the test... 😁

Issue #1032, opened 2017-01-05
Update license headers to 2017


agudian:
... I'll take care of 1.1 and master...

sjaakd:
@agudian can this be closed?

agudian:
Oh, yes. Thanks for the reminder 👍

PR #1033, opened 2017-01-05
#1032 Update license headers to 2017

agudian:


filiphr:
Do we also need to update the [LICENSE.txt](https://github.com/mapstruct/mapstruct/blob/master/LICENSE.txt) as well?

agudian:
> Do we also need to update the LICENSE.txt as well?

Right. I'll update it and commit it directly, then merge it to master.

gunnarmorling:
Already did it for 1.1.

gunnarmorling:
Ok, so I'll leave the merge to master to you, @agudian. Thx!

agudian:
there's also one missing in the root pom.xml 🙄

PR #1034, opened 2017-01-06
Iterable and Stream Mapping methods duplication cleanup

filiphr:
This PR cleans up the duplication between the `StreamMappingMethod` and the `IterableMappingMethod` and their respective builders

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/WithElementMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/WithElementMappingMethod.java
@@ -39,7 +39,7 @@ import org.mapstruct.ap.internal.util.Strings;
  *
  * @author Filip Hrisafov
  */
-abstract class WithElementMappingMethod extends MappingMethod {
+public abstract class WithElementMappingMethod extends MappingMethod {
     private final Assignment elementAssignment;
     private final MethodReference factoryMethod;
     private final boolean overridden;
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/WithElementMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/WithElementMappingMethod.java
@@ -1,0 +1,195 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+import javax.lang.model.type.TypeKind;
+
+import org.mapstruct.ap.internal.model.assignment.Assignment;
+import org.mapstruct.ap.internal.model.common.Parameter;
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.model.source.ForgedMethod;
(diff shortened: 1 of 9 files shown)


agudian:
Looks good on a first glance. I don't see anything funny... 👍

processor/src/main/java/org/mapstruct/ap/internal/model/WithElementMappingMethod.java
```diff
@@ -0,0 +1,195 @@
...
+import org.mapstruct.ap.internal.model.assignment.Assignment;
+import org.mapstruct.ap.internal.model.common.Parameter;
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.model.source.ForgedMethod;
+import org.mapstruct.ap.internal.model.source.Method;
+import org.mapstruct.ap.internal.model.source.SelectionParameters;
+import org.mapstruct.ap.internal.util.Strings;
+
+/**
+ * A {@link MappingMethod} implemented by a {@link Mapper} class which does mapping of typed based methods.
```
gunnarmorling: "typed based methods", that sounds odd.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -74,6 +71,10 @@
 import org.mapstruct.ap.internal.util.Strings;
 import org.mapstruct.ap.internal.version.VersionInformation;
 
+import static org.mapstruct.ap.internal.prism.MappingInheritanceStrategyPrism.AUTO_INHERIT_FROM_CONFIG;
```
gunnarmorling: Don't we have them at the beginning?

processor/src/main/java/org/mapstruct/ap/internal/model/WithElementMappingMethod.java
```diff
@@ -0,0 +1,195 @@
...
+ * For example Iterable or Stream.
+ * The typed elements are mapped either by a {@link TypeConversion} or another mapping method.
+ *
+ * @author Filip Hrisafov
+ */
+public abstract class WithElementMappingMethod extends MappingMethod {
```
gunnarmorling: I stumbled a bit upon the name. Maybe something like `ContainerMappingMethod` or so?

gunnarmorling:
@filiphr looks good, only added some small remarks.

processor/src/main/java/org/mapstruct/ap/internal/model/WithElementMappingMethod.java
filiphr: I completely agree. What do you suggest for the Java doc. Maybe something with generic types?

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
filiphr: They are removed from there due to the IntelliJ code style settings moving them at the end. I always try to not commit changes like this, but some are creeping through my checks 😄. Will remove it in the next commit.

processor/src/main/java/org/mapstruct/ap/internal/model/WithElementMappingMethod.java
filiphr: I had problems coming up with the name as well 😄. `ContainerMappingMethod` is much better. Maybe `GenericContainerMappingMethod`. or is it too long and does not propagate the meaning about java generics?

gunnarmorling: Either works for me. Your call :)

gunnarmorling:
Rebased and applied. Thanks, @filiphr! I've reworked the last commit message to refer to the issue key.

filiphr:
@gunnarmorling thanks. Just so you know you could have squashed and merged everything directly from github. It even offers you to change the commit message in the GUI.

I usually do multiple commits in order to track the review process better 😄

gunnarmorling:
Yeah, I still prefer to do rebases locally most of the time. So I can take
a look at the generated code and get the full picture more easily.

Regarding squashing, I tend to not do it to not "steal" the number of
commits from you. I'd leave that to yourself (just do a force push once you
are done) so you can do it if you like.

filiphr:
OK good to know. I will do squashes within the PR then.

I don't consider "stealing" number of commits if you squash. I also don't want to be seen as increasing number of commits by making this smaller review commits.

>I'd leave that to yourself (just do a force push once you are done) so you can do it if you like.

This one is already merged. Should I do a squash on master?

gunnarmorling:
Nope, let's leave things as they are on master. We can do it differently
next time. Thanks!

PR #1037, opened 2017-01-07
#1005 Report an error if the resulting type in the BeanMappingMethod is abstract

### LINKED ISSUE #1005: Report an error if the resulting type in the BeanMappingMethod is abstract

filiphr:
Fixes #1005

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -181,13 +181,29 @@ public class BeanMappingMethod extends MappingMethod {
             if ( factoryMethod == null ) {
                 if ( selectionParameters != null && selectionParameters.getResultType() != null ) {
                     resultType = ctx.getTypeFactory().getType( selectionParameters.getResultType() );
-                    if ( !resultType.isAssignableTo( method.getResultType() ) ) {
+                    if ( resultType.isAbstract() ) {
+                        ctx.getMessager().printMessage(
+                            method.getExecutable(),
+                            beanMappingPrism.mirror,
+                            Message.BEANMAPPING_ABSTRACT,
+                            resultType,
+                            method.getResultType()
+                        );
+                    }
+                    else if ( !resultType.isAssignableTo( method.getResultType() ) ) {
                         ctx.getMessager().printMessage(
                             method.getExecutable(),
                             beanMappingPrism.mirror,
                             Message.BEANMAPPING_NOT_ASSIGNABLE, resultType, method.getResultType()
                         );
                     }
+                }
+                else if ( !method.isUpdateMethod() && method.getReturnType().isAbstract() ) {
+                    ctx.getMessager().printMessage(
+                        method.getExecutable(),
+                        Message.GENERAL_ABSTRACT_RETURN_TYPE,
+                        method.getReturnType()
+                    );
                 }
             }
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -189,6 +189,10 @@ public class Type extends ModelElement implements Comparable<Type> {
         return isVoid;
     }
 
+    public boolean isAbstract() {
+        return typeElement != null && typeElement.getModifiers().contains( Modifier.ABSTRACT );
+    }
+
     /**
      * @return this type's enum constants in case it is an enum, an empty list otherwise.
      */
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
@@ -30,6 +30,7 @@ public enum Message {
     // CHECKSTYLE:OFF
     BEANMAPPING_NO_ELEMENTS( "'nullValueMappingStrategy', 'resultType' and 'qualifiedBy' are undefined in @BeanMapping, define at least one of them." ),
     BEANMAPPING_NOT_ASSIGNABLE( "%s not assignable to: %s." ),
+    BEANMAPPING_ABSTRACT( "The result type %s may not be an abstract class nor interface." ),
     BEANMAPPING_UNKNOWN_PROPERTY_IN_RESULTTYPE( "Unknown property \"%s\" in return type." ),
     BEANMAPPING_PROPERTY_HAS_NO_WRITE_ACCESSOR_IN_RESULTTYPE( "Property \"%s\" has no write accessor." ),
     BEANMAPPING_SEVERAL_POSSIBLE_SOURCES( "Several possible source properties for target property \"%s\"." ),
@@ -77,6 +78,7 @@ public enum Message {
     DECORATOR_CONSTRUCTOR( "Specified decorator type has no default constructor nor a constructor with a single parameter accepting the decorated mapper type." ),
 
     GENERAL_NO_IMPLEMENTATION( "No implementation type is registered for return type %s." ),
+    GENERAL_ABSTRACT_RETURN_TYPE( "The return type %s is an abstract class or interface. Provide a non abstract / non interface result type or a factory method." ),
     GENERAL_AMBIGIOUS_MAPPING_METHOD( "Ambiguous mapping methods found for mapping %s to %s: %s." ),
     GENERAL_AMBIGIOUS_FACTORY_METHOD( "Ambiguous factory methods found for creating %s: %s." ),
     GENERAL_UNSUPPORTED_DATE_FORMAT_CHECK( "No dateFormat check is supported for types %s, %s" ),
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_1005/Issue1005Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_1005/Issue1005Test.java
@@ -1,0 +1,93 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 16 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -188,6 +188,22 @@ public BeanMappingMethod build() {
...
+                        ctx.getMessager().printMessage(
+                            method.getExecutable(),
+                            beanMappingPrism.mirror,
+                            Message.BEANMAPPING_ABSTRACT,
+                            resultType,
+                            method.getResultType()
+                        );
+                    }
+                }
+                else if ( !method.isUpdateMethod() && method.getReturnType().isAbstract() ) {
```
agudian: Are interfaces handled somewhere else already? I'd find it more clear if we'd also report an error (or the same error) if the given result type is an interface...

filiphr: There is one general place where interfaces are handled. That is in [here](https://github.com/mapstruct/mapstruct/blob/6b07dda829788387abef3ff8574066553ad7b646/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java#L538) in the `MapperCreationProcessor`. That one handles the other types as well.

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/QualifierTest.java
```diff
@@ -189,7 +189,12 @@ public void testFactorySelectionWithQualifier() {
                         kind = Kind.ERROR,
                         line = 37,
                         messageRegExp = "'nullValueMappingStrategy', 'resultType' and 'qualifiedBy' are undefined in "
-                            + "@BeanMapping, define at least one of them." )
+                            + "@BeanMapping, define at least one of them."),
+                @Diagnostic(type = ErroneousMovieFactoryMapper.class,
+                    kind = Kind.ERROR,
+                    line = 37,
+                    messageRegExp = "The result type .*\\.AbstractEntry is abstract. Provide non abstract result type" +
+                        " or a factory method.")
```
agudian: Could you reformat the complete `@ExpectedCompilationOutcome(..)`, please? The formatting is a bit strange above...

filiphr: Will do. I didn't want to make unnecessary formatting.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
agudian: That doesn't feel right. Here, we only check that the manually provided type is not an abstract class, but it could be an interface (which it mustn't be either).
In the linked code, we check that if the return type is an interface, the result type must be manually defined, although it would also have to be defined if the return type is an abstract class. But both would only be true if there was no factory method found, so I'm not so sure if that other place is really correct.

IMHO we should report errors for the following cases:
* The return type is an interface or abstract class AND 
* no factory method can be found AND
* no result type is specified OR the specified result type is an interface or an abstract class

The "interface or abstract class" could also be amended by " a class without a default constructor" (where we currently generate non-compilable code, btw).

filiphr: You are right. For #657, we said if the result type of the BeanMappingMethod is not an interface then we can consider that the method has a factory method. See this.

I think that with this PR all the errors that you mentioned are covered, just not in one place 😄. Should we structure that a bit better?

agudian: Yes, some clean-up would be great. 👍

And perhaps a test mapper with methods for all error variants?

filiphr:
@agudian I applied the reformatting and also changed the way we report errors. Some tests had to be extended (when the return type and the result type are interfaces), as we report errors twice in the execution.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -181,11 +181,45 @@ public BeanMappingMethod build() {
...
+                        );
+                    }
+                    else if ( !resultType.isAssignableTo( method.getResultType() ) ) {
+                        ctx.getMessager().printMessage(
+                        method.getExecutable(),
+                        beanMappingPrism.mirror,
+                        Message.BEANMAPPING_NOT_ASSIGNABLE, resultType, method.getResultType()
+                        );
+                    }
+                    else if ( resultType.isAbstract() ) {
```
gunnarmorling: I'd pull this before the second one, so that interface and abstract class check stay next to each other.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -189,6 +189,11 @@ public boolean isVoid() {
         return isVoid;
     }
 
+    public boolean isAbstract() {
```
gunnarmorling: Would that actually kick in for interfaces, too? I don't think we need keep the two cases apart, interfaces are just a form of abstract type. In the error message we could say "result type may not be an abstract class nor interface".

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -77,6 +79,8 @@
     DECORATOR_CONSTRUCTOR( "Specified decorator type has no default constructor nor a constructor with a single parameter accepting the decorated mapper type." ),
 
     GENERAL_NO_IMPLEMENTATION( "No implementation type is registered for return type %s." ),
+    GENERAL_ABSTRACT_RETURN_TYPE( "The return type %s is abstract. Provide non abstract result type or a factory method." ),
+    GENERAL_INTERFACE_RETURN_TYPE( "The return type %s is interface. Provide non abstract/interface result type or a factory method." ),
```
gunnarmorling: Seems "non abstract" is a left-over? I'd handle both cases together as lined out above.

gunnarmorling:
Looks great, @filiphr! Some small comments inline.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: Yes it can kick in for interfaces also. I will update the 2 places where we use this to have one message for interface and abstract class

processor/src/test/java/org/mapstruct/ap/test/bugs/_1005/Issue1005Test.java
```diff
@@ -0,0 +1,103 @@
...
+                kind = javax.tools.Diagnostic.Kind.ERROR,
+                line = 29,
+                messageRegExp = "The return type .*\\.HasKey is interface. Provide non abstract/interface result type" +
+                    " or a factory method."),
+            @Diagnostic(type = Issue1005ErroneousInterfaceReturnTypeMapper.class,
+                kind = javax.tools.Diagnostic.Kind.ERROR,
+                line = 29,
+                messageRegExp = "No implementation type is registered for return type .*\\.HasKey.")
+        })
+    public void shouldFailDueToInterfaceReturnType() throws Exception {
```
agudian: Should we really report two very similar error messages for the same code line? I think the second message isn't realy required here, as the one above has more context information and hints about what to do. Plus, some compilers (e.g. JDK) swallow all but the first error message for the same location. See `JdkCompilingStatement.filterExpectedDiagnostics`... :)

filiphr: Well we don't have to. However the second line comes from the [`MapperCreationProcessor#L538`](https://github.com/mapstruct/mapstruct/blob/6b07dda829788387abef3ff8574066553ad7b646/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java#L538). If I return `null` in the `BeanMappingMethod` when an error occurs then there will be only one error. Is that OK?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
filiphr: It will be done at the beginning and it abstract and interface will be considered the same

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
filiphr: Both cases will be handled together

processor/src/test/java/org/mapstruct/ap/test/bugs/_1005/Issue1005Test.java
filiphr: Sorry my bad returning `null` won't help. Do you have some idea how to avoid it? I can add a property to the `BeanMappingMethod` to store that an error occurred already. but I don't like that.

gunnarmorling: Can't you invoke the two checks at the same place, i.e. in `build()` of the bean mapping method builder? There you should be able to ensure only one error is raised using some local variable.

filiphr: The checks are already the same. I found a way. If there is a `BeanMappingMethod` constructed we can consider that it has a factory method, if there was some problem an error would have already been created

filiphr:
The build failure is strange...

filiphr:
Jenkins please retest this

filiphr:
retest this, please

filiphr:
@gunnarmorling, @agudian are you ok with the latest change? Now we assume that if a `BeanMappingMethod` is constructed it has a "factory method". If you are OK I will squash and merge. I will also fix one type in one of my comments 😄

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -77,6 +78,7 @@
     DECORATOR_CONSTRUCTOR( "Specified decorator type has no default constructor nor a constructor with a single parameter accepting the decorated mapper type." ),
 
     GENERAL_NO_IMPLEMENTATION( "No implementation type is registered for return type %s." ),
+    GENERAL_ABSTRACT_RETURN_TYPE( "The return type %s is abstract class or interface. Provide non abstract/non interface result type or a factory method." ),
```
agudian: `... is an abstarct class or interface. Provide a non abstract / non interface result type .. `?
(adding "an", "a", and spaces around "/")

filiphr: will do this in the merge

filiphr:
Thanks for the review @agudian, @gunnarmorling

Issue #1027, opened 2016-12-28
Mappers should not consider InheritInverseConfiguration in used mappers


sjaakd:
Leading to null pointers. I reckon because there's no source model present if you call a used mapper method in another mapper.

gunnarmorling:
Done.

PR #1038, opened 2017-01-08
#1027 do not consider Inherit(Inverse)Configuration in used mappers

sjaakd:
Currently MapStruct can refer to used mappers to inherit(reverse)configurations. However, it could be that these mappings are not available (for instance: the used mapper is part of another jar library or the concerning used mapper is not yet generated in the processor). I don't think we intended that so I explicitly disabled it.


gunnarmorling:
@sjaakd Did you run into an actual problem here? I agree it makes sense to limit it to the same mapper, but I'm curious whether it's a theoretical issue or a real one.

> or the concerning used mapper is not yet generated in the processor

But that shouldn't matter as we'd just compile against the interface and not the implementation?

gunnarmorling:
As discussed during today's Hangout, the issue stems from the fact that we don't build up any mapping meta-data for used mappers.

gunnarmorling:
Merged; thanks, @sjaakd!

sjaakd:
> Merged; thanks, @sjaakd!

No problem!

PR #1039, opened 2017-01-08
#1001 auto mapping support for update methods

sjaakd:
just a testcase to start with

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -635,14 +635,24 @@ public class PropertyMapping extends ModelElement {
             if ( sourceType.isPrimitive() || targetType.isPrimitive() ) {
                 return null;
             }
+
             String name = getName( sourceType, targetType );
+            List<Parameter> parameters = new ArrayList( method.getContextParameters() );
+            Type returnType;
+            if ( method.isUpdateMethod() ) {
+                parameters.add( Parameter.forForgedMappingTarget( targetType ) );
+                returnType = ctx.getTypeFactory().createVoidType();
+            }
+            else {
+                returnType = targetType;
+            }
             ForgedMethod forgedMethod = new ForgedMethod(
                 name,
                 sourceType,
-                targetType,
+                returnType,
                 method.getMapperConfiguration(),
                 method.getExecutable(),
-                method.getContextParameters(),
+                parameters,
                 getForgedMethodHistory( sourceRHS )
             );
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
@@ -125,6 +125,15 @@ public class Parameter extends ModelElement {
             ContextPrism.getInstanceOn( element ) != null );
     }
 
+    public static Parameter forForgedMappingTarget(Type parameterType) {
+        return new Parameter(
+            "mappingTarget",
+            parameterType,
+            true,
+            false,
+            false);
+    }
+
     /**
      * @param parameters the parameters to filter
      * @return the parameters from the given list that are considered 'source parameters'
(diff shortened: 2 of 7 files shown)


gunnarmorling:
ready for review.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -638,16 +639,33 @@ private Assignment forgeMapping(SourceRHS sourceRHS) {
...
-                targetType,
-                method.getMapperConfiguration(),
-                method.getExecutable(),
-                method.getContextParameters(),
-                getForgedMethodHistory( sourceRHS )
-            );
+            List<Parameter> parameters = new ArrayList( method.getContextParameters() );
+            ForgedMethod forgedMethod;
+            if ( method.isUpdateMethod() ) {
+                parameters.add( Parameter.forForgedMappingTarget( targetType ) );
```
filiphr: I would just add the new parameter if the `method` is update method. Everything else can stay as it was before. 

If the `mappingTargetParameter` is not `null` and it is when we do an update method, then we return it's type. The `TypeKind.VOID` is obsolete

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
```diff
@@ -303,4 +306,14 @@ public int hashCode() {
...
         return result;
     }
+
+    private Parameter determineMappingTargetParameter(Iterable<Parameter> parameters) {
+        for ( Parameter parameter : parameters ) {
+            if ( parameter.isMappingTarget() ) {
+                return parameter;
+            }
+        }
+
```
filiphr: 2 empty lines

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
```diff
@@ -303,4 +306,14 @@ public int hashCode() {
         result = 31 * result + ( name != null ? name.hashCode() : 0 );
         return result;
     }
+
+    private Parameter determineMappingTargetParameter(Iterable<Parameter> parameters) {
```
filiphr: Just a not for the future, we can extract this construct to some util. We use it in multiple places 😄

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: I think its more cleaner to return void. I struggled in the past with update methods that have both a return type and a mapping target. I still don't like that :).

filiphr: I agree with you. I forgot the `getReturnType`.

However, I will still reduce the amount of code. By either new variable which will be either the `targetType` or `TypeKind.VOID` or with a ternary on the same line 😄

sjaakd: Ok.. will update.

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
sjaakd: I could do it as static on Parameter. Just like @agudian did for context parameters as well.

filiphr: As you wish. We can also do it as part of another issue

sjaakd:
@filiphr : I covered your remarks and did some small additional refactoring. Have a look if you agree. After I'll continue to merge.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
```diff
@@ -125,6 +125,15 @@ public static Parameter forElementAndType(VariableElement element, Type paramete
             ContextPrism.getInstanceOn( element ) != null );
     }
 
+    public static Parameter forForgedMappingTarget( Type parameterType) {
```
agudian: 👍🙂

filiphr: Just the formatting looks a bit off. Is there a space after `(`?

filiphr:
@sjaakd everything looks fine, there is only one small formatting issue (maybe it is not and I am not seeing it correctly 😄)

sjaakd:
@agudian, @filiphr : thanks for reviewing

sjaakd:
@filiphr : fixed the spacing

Issue #510, opened 2015-03-20
fail if Project Lombok is used (projectlombok.org)


panov-andy:
Hi.
I'm using Project Lombok and unable generate mappers with maven.
I use @Data annotation (lombok) if it make sense.

gunnarmorling:
There has been a discussion about the combination of MapStruct + Lombok on the Google group recently: https://groups.google.com/forum/#!topic/mapstruct-users/91fBwD4hNZE. An example for using both together is here: https://github.com/mapstruct/mapstruct-examples

If this does not help, it'd be good to know what the exact issue is and how your set-up does look like.

bosyak:
Suddenly it's not an option for me to split project to modules.
Anyway I'll thing about it.
Thanks.

dave-lo:
I was able to get the mapstruct working with Lombok on versions < 1.16.0 because the annotationprocessor was still exposed. Starting 1.16.0+ the annotation processor is hidden so the maven compiler can't use it and thus this solution won't work

agudian:
That's too bad - it's likely that there is not much that we can do on our side. Perhaps the devs at lombok can take care of supporting annotation-processors again...

rzwitserloot:
We can't do much without more info on _how_ @dave-lo fixed this issue. I don't even know what the issue is :/

gunnarmorling:
Hey @rzwitserloot, thanks for dropping by. Yes, it would help to know what @dave-lo did do, e.g. it's not clear to me what "Starting 1.16.0+ the annotation processor is hidden" means. I suppose somehow he managed to establish an ordering of Lombok and then MapStruct.

Generally, the issue is that MapStruct generates code based on accessors of JavaBeans properties which don't exist yet in the source code when using Lombok's @Data annotation. I'm not sure whether an annotation processor such as MapStruct can reliably "see" bytecode generated by another participant of the same compilation (as that case generally is not foreseen by the compiler environment).

Another idea for making the integration better would be to let MapStruct take fields into account as well (we want that anyways) but let it generate accessor-based code instead of field-accessing code by means of some configuration (or e.g. when spotting the `@Data` annotation). That'd by my preferred approach as it should be more robust and e.g. also work in IDEs.

rzwitserloot:
[...]

PR #1042, opened 2017-01-14
#510 Adding experimental SPI for letting AST-modifying annotation pro…

gunnarmorling:
…cessors tell us about future modifications

An experiment of an SPI towards Lombok


processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -102,6 +110,20 @@ public TypeFactory(Elements elementUtils, Types typeUtils) {
         implementationTypes.put( ConcurrentNavigableMap.class.getName(), getType( ConcurrentSkipListMap.class ) );
     }
 
+    private static List<AstModifyingAnnotationProcessor> getAstModifyingAnnotationProcessors() {
+        List<AstModifyingAnnotationProcessor> processors = new ArrayList<AstModifyingAnnotationProcessor>();
+
+        ServiceLoader<AstModifyingAnnotationProcessor> loader = ServiceLoader.load(
+                AstModifyingAnnotationProcessor.class
```
rzwitserloot: Make this:

    ServiceLoader.load(AstModifyingAnnotationProcessor.class, AstModifyingAnnotationProcessor.class.getClassLoader());

Without the second param, you get the thread context loader and that's not the right one (all the APs share a special loader and we need that one).

rzwitserloot:
You need to fix one tiny thing in this change and it'll work together with the latest lombok edge release.

The thing to fix is commented in the change (https://github.com/mapstruct/mapstruct/pull/1042/commits/60823f0797efd3b985e1b3e781ad60943c5192fa) and the latest lombok edge release that works with this is at https://projectlombok.org/download-edge.html

agudian:
Nice, we'll try it out!

gunnarmorling:
Hey @rzwitserloot indeed it works if Lombok runs first and also - with the adjustment you suggested - when MapStruct comes first. Very nice!

I'm a tad concerned about your implementation of `AstModifyingAnnotationProcessor`, though. It ignores the type passed in and will return true as soon as `init()` of the processor has been called once.

It may pose a problem in cases where types which are enhanced by Lombok are only generated by another AP. Admittedly, that looks like a far-fetched case (why not generating that type completely then using that other AP?), but it could lead to the case where Lombok has been initialized (so you'll return `true` from `isTypeComplete()`) but not enhanced a given type.

gunnarmorling:
@agudian et al., I've pushed an update. It's good to go from side now.

gunnarmorling:
Pushed another commit to address #1045.

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -67,6 +70,7 @@
 
     private final Elements elementUtils;
     private final Types typeUtils;
+    private RoundContext roundContext;
```
agudian: Make it final as well?

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -565,4 +569,31 @@ static String trimSimpleClassName(String className) {
         return trimmedClassName;
     }
 
+    private boolean isCleared(TypeMirror type) {
```
agudian: Add a little javadoc to explain what specifically "cleared" means for the type?

processor/src/main/java/org/mapstruct/ap/spi/AstModifyingAnnotationProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+public interface AstModifyingAnnotationProcessor {
+
+    /**
+     * Whether the specified type has been fully processed by this processor or not (i.e. this processor will amend the
+     * given type's structure after this invocation).
+     *
+     * @param type The type of interest
+     * @return {@code true} if this processor has fully processed the given type, {@code false} otherwise.
+     */
+    boolean isTypeComplete(TypeMirror type);
```
agudian: Also needs to return `true` if the processor has no buisness with that type, right?

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
sjaakd: I think the name isCleared has a double connotation, right? IIUYC it means: "safe to use" or "ready to be used". There's also a connotation: erased.

processor/src/main/java/org/mapstruct/ap/spi/AstModifyingAnnotationProcessor.java
sjaakd: I perhaps miss something in the integration tests.. But I assumed that this needs to be somewhere implemented via SPI, right? I can't  find an implementation in the integration tests. And still it works?

gunnarmorling: @sjaakd there is currently no integration test. We could add one once this feature is implemented on the Lombok side. I guess I could add a custom processor somehow imitating what Lombok does, but it doesn't seem worth the trouble.

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
gunnarmorling: Ok.

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
gunnarmorling: @sjaakd, I see where you are coming from. Happy to change it, any suggestions?

gunnarmorling:
@agudian @sjaakd Rebased and added one more commit for addressing your review comments. Good to go fro my side. Thanks for reviewing!

sjaakd:
all ok from my side.

gunnarmorling:
Squashed and merged.

kevinmic:
Were these changes ever merged into the master branch?    There is the comment "Squashed and merged" but when I checkout the source code I can't find the changes.

PR #1052, opened 2017-01-24
#1001 Change TestMultipleForgedMethodsTest as suggested

navpil:
As a followup on #60 

In particular:
TestMultipleForgedMethodsTest needs to do some assertions: create proper input data and assert the results. Rename the test class itself (Test*Test) and rename the classes Bar, BarDto, Dto... :) (taken from #1001 )

Is it necessary to rename Bar/BarDto classes? I see lots of Bar classes in tests.


agudian:
Ok to test

agudian:
@navpil looks good!

We try to use some proper examples in the tests to hopefully make them more readable - especially for more complex tests it's easier to get confused if it's all just foo and bar and baz... 😉
If you can come up with something that maybe works with your current domain of houses and garages and vehicles, that would certainly be great 👍. If it's too hard, it probably won't be easier to grasp than what we have already, so I'd leave it to your discretion...

sjaakd:
Ok.. I'll continue and merge this one. The last commit covers your remarks I assume. 

@navpil Thanks!
@agudian Thanks for reviewing!

Issue #1050, opened 2017-01-23
MappingMethods refactoring


filiphr:
We need to extract as much as possible up to the `MappingMethod`. Candidates here are the equals and hashcode methods.

We also need to make sure that the `BeanMappingMethod` does not extend the `ContainerMappingMethod` (the idea of this method is to be used for `MappingMethods` similar to the `IterableMappingMethod` and `StreamMappingMethod`. Methods that are wrapping some underlying type in generics) introduced in #1046

sjaakd:
Perhaps change the name as well. `ContainerMappingMethod` . A bean mapping method can also take and return generic parameterized types.

sjaakd:
@filiphr .. I think this one can be closed, right?

filiphr:
Well we still need to make sure that the `BeanMappingMethod` does not extend the `ContainerMappingMethod`. I just saw that your last merge had it still in. I will make a PR for it

filiphr:
Done

PR #1053, opened 2017-01-24
#1001 name clashes demonstrator

sjaakd:
In #1001, one of the issue indicated alignment between forge method strategies. Here's the reason why. Perhaps this can be taken up in: #1050. See also #1011 for more information

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
@@ -1,0 +1,88 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model;
+
+import org.mapstruct.ap.internal.model.assignment.Assignment;
+import org.mapstruct.ap.internal.model.common.ParameterBinding;
+import org.mapstruct.ap.internal.model.source.ForgedMethod;
+import org.mapstruct.ap.internal.model.source.Method;
+
+/**
+ * @author Filip Hrisafov
+ */
+class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
+
+    protected B myself;
+    protected MappingBuilderContext ctx;
+    protected Method method;
+
+    AbstractBaseBuilder(Class<B> selfType) {
+        myself = selfType.cast( this );
+    }
+
+    public B mappingContext(MappingBuilderContext mappingContext) {
+        this.ctx = mappingContext;
+        return myself;
+    }
+
+    public B method(Method sourceMethod) {
+        this.method = sourceMethod;
+        return myself;
+    }
+
+    /**
+     * Creates a forged assignment from the provided {@code sourceRHS} and {@code forgedMethod}. If a mapping method
+     * for the {@code forgedMethod} already exists, then this method used for the assignment.
+     *
+     * @param sourceRHS that needs to be used for the assignment
+     * @param forgedMethod the forged method for which we want to create an {@link Assignment}
+     *
+     * @return See above
+     */
+    Assignment createForgedBeanAssignment(SourceRHS sourceRHS, ForgedMethod forgedMethod) {
+        BeanMappingMethod forgedMappingMethod = new BeanMappingMethod.Builder()
+            .forgedMethod( forgedMethod )
+            .mappingContext( ctx )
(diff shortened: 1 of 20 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/MapMappingMethod.java
```diff
@@ -210,15 +208,26 @@ private Assignment forgeMapping(SourceRHS sourceRHS, Type sourceType, Type targe
                 )
             );
 
+            BeanMappingMethod forgedMappingMethod = new BeanMappingMethod.Builder()
```
sjaakd: Don't we have a method for this.?

sjaakd: I see this coming back several times. Can't we put the method on the context and parameterize?

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -651,15 +648,26 @@ private Assignment forgeMapping(SourceRHS sourceRHS) {
                 getForgedMethodHistory( sourceRHS )
             );
 
+            BeanMappingMethod forgedMappingMethod = new BeanMappingMethod.Builder()
```
sjaakd: here again..

sjaakd:
@filiphr .. Looks nice 👍 . I've one comment, I think we can do some nice generics on the getMapping and move it to the mapping context. WDYT?

sjaakd:
Something like this:

```java
    public Assignment addForgedMethod( ForgedMethod forgedMethod, MappingMethod mappingMethod, Assignment sourceRHS ) {
  

            MappingMethod forgedMappingMethod = new BeanMappingMethod.Builder()
                .forgedMethod( forgedMethod )
                .mappingContext( this )
                .build();

            if ( !getMappingsToGenerate().contains( forgedMappingMethod ) ) {
                getMappingsToGenerate().add( forgedMappingMethod );
            }
            else {
                String existingName = getExistingMappingMethod( forgedMappingMethod ).getName();
                forgedMethod = new ForgedMethod( existingName, forgedMethod );
            }

            Assignment assignment = new MethodReference(
                forgedMethod,
                null,
                ParameterBinding.fromParameters( forgedMethod.getParameters() ) );

            assignment.setAssignment( sourceRHS );
            return assignment;
        }
    
}
```
iso exposign `getMappingsToGenerate`.. WDYT?

agudian:
Hmm, I'm not so sure that the mapping context is the right place for that. How about putting it into a new `*Factory` or another Builder class?

filiphr:
That looks really good. I would also extract the `private Assignment forgeMapping(SourceRHS sourceRHS, Type sourceType, Type targetType)` from the `MapMappingMethod.Builder` and the `ContainerMappingMethodBuilder` into the `AbstractMappingMethodBuilder`.

I am not sure that we can hide `getMappingsToGenerate` because we use them in the `MapperCreatioProcessor` and also when building `nestedPropertyMappings` in the `PropertyMapping` from line 438

sjaakd:
> I am not sure that we can hide getMappingsToGenerate because we use them in the MapperCreatioProcessor and also when building nestedPropertyMappings in the PropertyMapping from line 438

No problem.. I thought it was the only usage.

filiphr:
I extracted the creation to a common abstract builder, have a look

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Creates a forged assignment from the provided {@code sourceRHS} and {@code forgedMethod}. If a mapping method
+     * for the {@code forgedMethod} already exists, then this method used for the assignment.
+     *
+     * @param sourceRHS that needs to be used for the assignment
+     * @param forgedMethod the forged method for which we want to create an {@link Assignment}
+     *
+     * @return See above
+     */
+    Assignment createForgedBeanAssignment(SourceRHS sourceRHS, ForgedMethod forgedMethod) {
```
sjaakd: Is there a reason a bean assignment is special in relation to container method (still think this is a strange name) and mapmaping methods that it has a place here? Do you expect reuse on more places

sjaakd:
@filiphr look very nice. Just one question. Its good for me to merge.

sjaakd:
@agudian : do you have time to have a quick look as well?

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
filiphr: The `AbstractBaseBuilder` is not only for a container method. The `PropertyMappingBuilder` also extends it via the `MappingBuilderBase`. It is also used there. I was not sure where to put it. Maybe there is a better name for this. I named it `createForgedBeanAssignment`, because in the `PropertyMappingBuilder` we have `forgeMapMapping`, `forgeIterableMapping` and `forgeStreamMapping`, so it is a bit clearer when you see the usage there.

I also added to avoid the exact same creation of the `BeanMappingMethod`. However, I can remove this method and have the `BeanMappingMethod` created at the callers place (there are 2 such places). WDYT?

gunnarmorling:
Hey all, where are we with this one? Any further discussion needed?

filiphr:
I don't think that there are any further discussions, only if it is OK for you (@gunnarmorling) and @agudian for merging 😄

sjaakd:
I've got it prepared for merge I even rebase #1011 on top of that (standing by).. So if you guys are ok.. Just waiting for a verdict.

sjaakd:
@gunnarmorling You still want to review this one? Or can I merge?

sjaakd:
@gunnarmorling : I reckon its also ok for you. I'll continue to merge

gunnarmorling:
Go ahead if you think it's good.

sjaakd:
Done.

PR #1058, opened 2017-02-01
Cleanup duplicates in MappingMethods

filiphr:
I have created a new class for all the `MappingMethods` that have a factory method. There was a lot of duplication between all of the different methods.

Is it too much abstraction (with the new class)?

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethod.java
@@ -20,7 +20,6 @@ package org.mapstruct.ap.internal.model;
 
 import java.util.List;
 import java.util.Set;
-import javax.lang.model.type.TypeKind;
 
 import org.mapstruct.ap.internal.model.assignment.Assignment;
 import org.mapstruct.ap.internal.model.common.Parameter;
@@ -77,27 +76,6 @@ public abstract class ContainerMappingMethod extends NormalTypeMappingMethod {
 
     public String getLoopVariableName() {
         return loopVariableName;
-    }
-
-    public String getDefaultValue() {
-        TypeKind kind = getResultElementType().getTypeMirror().getKind();
-        switch ( kind ) {
-            case BOOLEAN:
-                return "false";
-            case BYTE:
-            case SHORT:
-            case INT:
-            case CHAR:  /*"'\u0000'" would have been better, but depends on platformencoding */
-                return "0";
-            case LONG:
-                return "0L";
-            case FLOAT:
-                return "0.0f";
-            case DOUBLE:
-                return "0.0d";
-            default:
-                return "null";
-        }
     }
 
     public abstract Type getResultElementType();
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethod.java
@@ -36,11 +36,8 @@ import org.mapstruct.ap.internal.util.Strings;
  *
  * @author Filip Hrisafov
  */
-public abstract class ContainerMappingMethod extends MappingMethod {
+public abstract class ContainerMappingMethod extends NormalTypeMappingMethod {
     private final Assignment elementAssignment;
-    private final MethodReference factoryMethod;
-    private final boolean overridden;
-    private final boolean mapNullToDefault;
     private final String loopVariableName;
     private final SelectionParameters selectionParameters;
 
@@ -49,11 +46,8 @@ public abstract class ContainerMappingMethod extends MappingMethod {
         List<LifecycleCallbackMethodReference> beforeMappingReferences,
         List<LifecycleCallbackMethodReference> afterMappingReferences,
         SelectionParameters selectionParameters) {
-        super( method, beforeMappingReferences, afterMappingReferences );
+        super( method, factoryMethod, mapNullToDefault, beforeMappingReferences, afterMappingReferences );
         this.elementAssignment = parameterAssignment;
-        this.factoryMethod = factoryMethod;
-        this.overridden = method.overridesMethod();
-        this.mapNullToDefault = mapNullToDefault;
         this.loopVariableName = loopVariableName;
         this.selectionParameters = selectionParameters;
     }
@@ -78,20 +72,7 @@ public abstract class ContainerMappingMethod extends MappingMethod {
         if ( elementAssignment != null ) {
             types.addAll( elementAssignment.getImportTypes() );
         }
-        if ( ( factoryMethod == null ) && ( !isExistingInstanceMapping() ) ) {
-            if ( getReturnType().getImplementationType() != null ) {
-                types.addAll( getReturnType().getImplementationType().getImportTypes() );
-            }
-        }
         return types;
-    }
-
-    public boolean isMapNullToDefault() {
-        return mapNullToDefault;
-    }
-
-    public boolean isOverridden() {
-        return overridden;
     }
 
     public String getLoopVariableName() {
@@ -119,10 +100,6 @@ public abstract class ContainerMappingMethod extends MappingMethod {
         }
     }
 
-    public MethodReference getFactoryMethod() {
-        return this.factoryMethod;
-    }
-
     public abstract Type getResultElementType();
 
     public String getIndex1Name() {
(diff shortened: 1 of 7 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethodWithFactory.java
```diff
@@ -0,0 +1,113 @@
...
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.model.source.Method;
+
+/**
+ * A {@link MappingMethod} that is used by the main mapping methods ({@link BeanMappingMethod},
+ * {@link MapMappingMethod}, {@link IterableMappingMethod} and {@link StreamMappingMethod}
+ *
+ * @author Filip Hrisafov
+ */
+public abstract class MappingMethodWithFactory extends MappingMethod {
```
sjaakd: Is it the fact that it has a factory that sets apart methods inheriting from this method? I reckon in essence these are generated non-enum / non-value mapping methods: MapMapping, BeanMapping, IterableMapping, StreamMapping right?

processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethod.java
```diff
@@ -36,11 +36,8 @@
  *
  * @author Filip Hrisafov
  */
-public abstract class ContainerMappingMethod extends MappingMethod {
```
filiphr: You mean `loopVariableName`, right? This one is actually needed:
* In `IterableMappingMethod.ftl` - here it is clear
* In `StreamMappingMethod.ftl` -for updating an array on line 103

Both of them have some form of element assignment and use the `SelectionParameters`. Theoretically other similar mappings (for example Guava collections) can benefit from the `ContainerMappingMethod`.

As for `getDefault()`, it was in the `IterableMappingMethod` and I pulled it up when I needed it for the `StreamMappingMethod`. I didn't know about `Type#getNull()`, had I known I would have used it. I can clean that up as well and use the one from the `Type`. We will just have to agree on the differences between the 2.

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethodWithFactory.java
filiphr: You are right, it is for those 4 Mappings. The factory was something they shared so I used that in the name. However, I am open for suggestions, I am sure it can be better named

processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethod.java
sjaakd: Ok.. but if the getDefault() moves what does remain here? And does it justify to have a common base class. I'm not in favor of long inheritance chains, and prefer composition of inheritance if possible. Anyway. Don't consider this as blocker.

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethodWithFactory.java
sjaakd: jls chapter 8.1 (https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html) makes a distinction between an 
```
ClassDeclaration:
    NormalClassDeclaration
    EnumDeclaration

```

so what about `NormalTypeMapper`.

sjaakd: excuse me : `NormalTypeMappingMethod`

processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethod.java
filiphr: This methods will remain:
* `getLoopVariableName`, `getIndex1Name` and `getIndex2Name` - for the arrays
* `getElementAssignment` - How the underlying element is assigned.
* `getResultElementType` - I think that this is the most important one. An abstract that needs to be implemented to get the `Type` of the result element. 
* The `selectionParameters`

I completely understand you not liking the long inheritance chain. Do you perhaps have some other suggestion? Maybe make the `MappingMethodWithFactory` (which will be renamed) generic and use that for the `MappingMethod` specifics?

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethodWithFactory.java
filiphr: That is interesting, I never knew this 😄. I'll rename it to `NormalTypeMappingMethod`.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -758,10 +758,11 @@ public String getNull() {
             return "0";
         }
         if ( "char".equals( getName() ) ) {
-            return "'\\u0000'";
+            //"'\u0000'" would have been better, but depends on platform encoding
+                return "0";
         }
         if ( "double".equals( getName() ) ) {
-            return "0.0";
+            return "0.0d";
```
filiphr: I alligned this one with the default value that was use before. I did the same for the `char`.

We also need to decide whether we want to do it like we did before with the `TypeMirror` `Kind` or like it is here.

filiphr:
@sjaakd I addressed your comments and renamed the `MappingMethodWithFactory`. I also am using the `Type#getNull()` methods instead of the `ContainerMappingMethod#getDefaultValue()`. I alligned the `getNull()` with how it was for the `getDefaultValue()`.

processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethod.java
sjaakd: > I completely understand you not liking the long inheritance chain. Do you perhaps have some other suggestion?

I think we should park this one. But on thing we might consider is to use composition instead of inheritance. Or a combination: we still have a base class `MappingMethod` and top level methods like `BeanMappingMethod`, `MapMappingMethod` but we compose other behavior out of the `ContainerMappingMethod` and ` NormalTypeMappingMethod`

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
sjaakd: Does it still work for the other scenarios (where `Type` is used directly)?

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableMappingMethod.ftl
```diff
@@ -35,7 +35,7 @@
                 <#if existingInstanceMapping>
                     <#-- we can't clear an existing array, so we've got to clear by setting values to default -->
                     for (int ${index2Name} = 0; ${index2Name} < ${resultName}.length; ${index2Name}++ ) {
-                        ${resultName}[${index2Name}] = ${defaultValue};
+                        ${resultName}[${index2Name}] = ${resultElementType.null};
```
sjaakd: 👍

sjaakd: resultElementType cannot be null?

sjaakd:
Just a few questions remaining. If you feel ok, please proceed.

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableMappingMethod.ftl
filiphr: I don't think so. Just have a look at the `getDefaultValue()` that was removed. We were doing `getResultElementType().getTypeMirror()...` so if it was `null` we would have had an NPE.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: Well nothing failed. The other scenario where this was used is the `NestedPropertyMappingMethod.ftl`.

processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethod.java
filiphr: Should I do this as part of this PR or create a new issue to do it for later?

sjaakd: Someday.  A new issue is fine, but not urgent. It works as is. That's what I meant with: 

> I think we should park this one.

filiphr: OK. I created #1066 for it

filiphr:
Thanks for the review @sjaakd

Issue #1061, opened 2017-02-03
RuntimeException when use no package (default)


diegobmd:
NullPointerException when use no package (default), on eclipseor build command line

gunnarmorling:
What is it you are using the default package for, the mapper interface or
any mapped type?

diegobmd:
For testar only

gunnarmorling:
Hum, what is "testar"?

diegobmd:
Kkkk, i m a brazilian is a corret by Phineas
I mean
Only for tests

diegobmd:
Kkkk, i m a brazilian is a corret by Phone
I mean
Only for tests

gunnarmorling:
Ok, maybe you can provide a small example exposing this issue? It's still not clear to me where you are using the default package. Are you saying that your JUnit test is located the default package? That should not make any difference to MapStruct really. Can you share the stack trace, too?

diegobmd:
I'm create on github
https://github.com/diegobmd/mapstructapt
RUN gradlew build

:compileJava
mapstructapt\src\main\java\SourceTargetMapper.java:26: error: Internal error in the mapping processor: java.lang.RuntimeException: javax.annotation.processing.FilerException: Illegal name .SourceTargetMapperImpl    at org.mapstruct.ap.internal.processor.MapperRenderingProcessor.createSourceFile(MapperRenderingProcessor.java:67)      at org.mapstruct.ap.internal.processor.MapperRenderingProcessor.writeToSourceFile(MapperRenderingProcessor.java:52)     at org.mapstruct.ap.internal.processor.MapperRenderingProcessor.process(MapperRenderingProcessor.java:42)       at org.mapstruct.ap.internal.processor.MapperRenderingProcessor.process(MapperRenderingProcessor.java:37)       at org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:223)         at org.mapstruct.ap.MappingProcessor.processMapperTypeElement(MappingProcessor.java:203)        at org.mapstruct.ap.MappingProcessor.proc
...

filiphr:
It actually it is not a `NullPointerException`. It is an exception because of the name. The Mapper does not have a package.

@gunnarmorling if someone wants to have a mapper in the default package we would have a problem. This are the reasons:
[...]

PR #1062, opened 2017-02-04
#1061 Support for generating Mappers in the default package

filiphr:
This PR makes sure that we generate a valid name for the Mappers that are created in the default package and that no `package` is added to the generated file.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperRenderingProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperRenderingProcessor.java
@@ -57,7 +57,11 @@ public class MapperRenderingProcessor implements ModelElementProcessor<Mapper, M
     }
 
     private void createSourceFile(GeneratedType model, ModelWriter modelWriter, Filer filer) {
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/MapperRenderingProcessor.java
```diff
@@ -57,7 +58,11 @@ private void writeToSourceFile(Filer filer, Mapper model) {
     }
 
     private void createSourceFile(GeneratedType model, ModelWriter modelWriter, Filer filer) {
-        String fileName = model.getPackageName() + "." + model.getName();
+        String fileName = "";
+        if ( !Strings.isEmpty( model.getPackageName() ) ) {
```
gunnarmorling: Seems your new method would fit nicely: `if( model.hasPackageName() ) { ... }`?

filiphr: Of course it would, I missed it 😄. I'll change that.

Issue #1056, opened 2017-01-30
java.lang.NegativeArraySizeException in indentation processing


stoerr:
When generating code I've got a NegativeArraySizeException that took me a while to find the cause for. The problem was that I had a closing parenthesis too many in an expression:

    @Mapping(target = "configurationPending", expression = "java(hasPendingSensorValue(src,SENSORS)))")

I'm not sure whether mapstruct should even try to recognize something like that during code generation, but I suggest to fix things that something as unimportant as the indentation doesn't break the build. The problem occurs both with 1.1.0.Final as well as 1.2.0-SNAPSHOT

The exception is:

    java.lang.NegativeArraySizeException
    at org.mapstruct.ap.internal.writer.IndentationCorrectingWriter.getIndentation(IndentationCorrectingWriter.java:101)
    at org.mapstruct.ap.internal.writer.IndentationCorrectingWriter.access$200(IndentationCorrectingWriter.java:47)
    at org.mapstruct.ap.internal.writer.IndentationCorrectingWriter$State$1.doOnEntry(IndentationCorrectingWriter.java:142)
    at org.mapstruct.ap.internal.writer.IndentationCorrectingWriter$State.onEntry(IndentationCorrectingWriter.java:354)
    at org.mapstruct.ap.internal.writer.IndentationCorrectingWriter.write(IndentationCorrectingWriter.java:75)
    ...

gunnarmorling:
Hey, agreed that we shouldn't run into `NegativeArraySizeException` in this case. The resulting code wouldn't compile anyways, though (as you have a closing brace too much), so it won't change much (apart from the less confusing exception).

stoerr:
Right, but the resulting compile error tells you where the problem is, while the NegativeArraySizeException just says "something is wrong somewhere in this file". 8-) So, I think you are doing a useful fix. Thank you!

agudian:
Fix landed on master. Thanks @stoerr for reporting and @gunnarmorling for the fix!

PR #1063, opened 2017-02-04
#1056 Making sure indentation level in formatting writer never goes b…

gunnarmorling:
…elow 0

Also had to provide a way to disable CheckStyle for this new test, as the additional brace would give a CS error.


gunnarmorling:
@sjaakd any thoughts from your side? Anyone up for merging this one :)

gunnarmorling:
Thanks!

PR #1069, opened 2017-02-06
#1001 Update documentation for the new automapping

filiphr:
Is this enough or we need to write more?

### DIFF
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -338,6 +338,7 @@ public class CarMapperImpl implements CarMapper {
         if ( car.getCategory() != null ) {
             carDto.setCategory( car.getCategory().toString() );
         }
+        carDto.setEngine( engineTtoEngineDto( car.getEngine() ) );
 
         return carDto;
     }
@@ -346,13 +347,31 @@ public class CarMapperImpl implements CarMapper {
     public PersonDto personToPersonDto(Person person) {
         //...
     }
+
+    private EngineDto engineToEngineDto(Engine engine) {
+        if ( engine == null ) {
+            return null;
+        }
+
+        EngineDto engineDto = new EngineDto();
+
+        engineDto.setHorsePower(engine.getHorsePower());
+        engineDto.setFuel(engine.getFuel());
+
+        return engineDto;
+    }
 }
 ----
 ====
 
 The general philosophy of MapStruct is to generate code which looks as much as possible as if you had written it yourself from hand. In particular this means that the values are copied from source to target by plain getter/setter invocations instead of reflection or similar.
 
-As the example shows the generated code takes into account any name mappings specified via `@Mapping`. If the type of a mapped attribute is different in source and target entity, MapStruct will either apply an automatic conversion (as e.g. for the _price_ property, see also <<implicit-type-conversions>>) or optionally invoke another mapping method (as e.g. for the _driver_ property, see also <<mapping-object-references>>).
+As the example shows the generated code takes into account any name mappings specified via `@Mapping`.
+If the type of a mapped attribute is different in source and target entity,
+MapStruct will either apply an automatic conversion (as e.g. for the _price_ property, see also <<implicit-type-conversions>>)
+or optionally invoke / create another mapping method (as e.g. for the _driver_ / _engine_ property, see also <<mapping-object-references>>).
+MapStruct will only create a new mapping method if and only if the source and target property are properties of a Bean and they themselves are Beans or simple properties.
+i.e. they are not `Collection` or `Map` type properties.
 
 Collection-typed attributes with the same element type will be copied by creating a new instance of the target collection type containing the elements from the source property. For collection-typed attributes with different element types each element will be mapped individually and added to the target collection (see <<mapping-collections>>).
 
@@ -805,7 +824,8 @@ When generating the implementation of a mapping method, MapStruct will apply the
 * If source and target attribute have the same type, the value will be simply copied from source to target. If the attribute is a collection (e.g. a `List`) a copy of the collection will be set into the target attribute.
 * If source and target attribute type differ, check whether there is a another mapping method which has the type of the source attribute as parameter type and the type of the target attribute as return type. If such a method exists it will be invoked in the generated mapping implementation.
 * If no such method exists MapStruct will look whether a built-in conversion for the source and target type of the attribute exists. If this is the case, the generated mapping code will apply this conversion.
-* Otherwise an error will be raised at build time, indicating the non-mappable attribute.
+* If no such method was found MapStruct will try to generate an internal method that will do the mapping between the source and target attributes
+* If MapStruct could not create a name based mapping method an error will be raised at build time, indicating the non-mappable attribute and its path.
 
 [[invoking-other-mappers]]
 === Invoking other mappers


documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -346,13 +347,29 @@ public class CarMapperImpl implements CarMapper {
...
 ----
 ====
 
 The general philosophy of MapStruct is to generate code which looks as much as possible as if you had written it yourself from hand. In particular this means that the values are copied from source to target by plain getter/setter invocations instead of reflection or similar.
 
-As the example shows the generated code takes into account any name mappings specified via `@Mapping`. If the type of a mapped attribute is different in source and target entity, MapStruct will either apply an automatic conversion (as e.g. for the _price_ property, see also <<implicit-type-conversions>>) or optionally invoke another mapping method (as e.g. for the _driver_ property, see also <<mapping-object-references>>).
+As the example shows the generated code takes into account any name mappings specified via `@Mapping`.
+If the type of a mapped attribute is different in source and target entity,
+MapStruct will either apply an automatic conversion (as e.g. for the _price_ property, see also <<implicit-type-conversions>>)
+or optionally invoke / create another mapping method (as e.g. for the _driver_ / _engine_ property, see also <<mapping-object-references>>).
```
filiphr: Only this lines differs from the previous version. The 3 lines above this one is just splitting the one-liner into multiple lines

filiphr:
@gunnarmorling, @agudian, @sjaakd Can you have a look at this PR for adding information about the forged named based mapping (automapping)?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -805,7 +822,8 @@ When generating the implementation of a mapping method, MapStruct will apply the
 * If source and target attribute have the same type, the value will be simply copied from source to target. If the attribute is a collection (e.g. a `List`) a copy of the collection will be set into the target attribute.
 * If source and target attribute type differ, check whether there is a another mapping method which has the type of the source attribute as parameter type and the type of the target attribute as return type. If such a method exists it will be invoked in the generated mapping implementation.
 * If no such method exists MapStruct will look whether a built-in conversion for the source and target type of the attribute exists. If this is the case, the generated mapping code will apply this conversion.
-* Otherwise an error will be raised at build time, indicating the non-mappable attribute.
+* If no such method was found MapStruct will generate an internal method that will do the mapping between the source and target attribute
```
sjaakd: What about: ```If no such method was found and **the source and target type adhere to the bean convention** MapStruct will generate an internal method that will do the mapping between the source and target attribute```

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: if and only if the source and target type adhere to the bean convention.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: and of course, other supported types (iterables, maps) also generate methods... But what I meant to say is we cannot generate arbitrary ones what this suggest..

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
gunnarmorling: +1 for using one line per sentence (or line of thought in a longer sentence). Diffs are much easier to follow that way.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
gunnarmorling: Can we add a sentence that explicitly creating the nested method is encouraged when it needs further configuration. I.e. we encourage not to configure it through the parent method (as enabled now by #1072), the reason being that this puts the configuration of the nested mapping into one place (method) where it can be be re-used from several methods at the upper level, instead of re-configuring the same things on all of those upper methods.

sjaakd: That's a good point.. feels like it could go into a 'Tip'.

filiphr: Isn't it better to reference what @sjaakd did in #1081 and add a 'Tip' there?

filiphr:
I updated the PR with the suggestions

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -346,13 +347,30 @@ public class CarMapperImpl implements CarMapper {
...
 
-As the example shows the generated code takes into account any name mappings specified via `@Mapping`. If the type of a mapped attribute is different in source and target entity, MapStruct will either apply an automatic conversion (as e.g. for the _price_ property, see also <<implicit-type-conversions>>) or optionally invoke another mapping method (as e.g. for the _driver_ property, see also <<mapping-object-references>>).
+As the example shows the generated code takes into account any name mappings specified via `@Mapping`.
+If the type of a mapped attribute is different in source and target entity,
+MapStruct will either apply an automatic conversion (as e.g. for the _price_ property, see also <<implicit-type-conversions>>)
+or optionally invoke / create another mapping method (as e.g. for the _driver_ / _engine_ property, see also <<mapping-object-references>>).
+MapStruct will only create a new mapping method if and only if the source and target property adhere to the bean convention.
```
gunnarmorling: What are we trying to say here? What about field support that we have now? And custom naming strategies?

filiphr: I am not sure what @sjaakd meant with that. Maybe we need to say: "MapStruct will only create a new mapping method if and only if the source and target properties match".  

I think you know what I mean by match (field matches a getter for it, custom naming strategy will make 2 properties match etc.). Do you have some other suggestion for this? How should we write it?

gunnarmorling: "match" suggests to their names are the same. @sjaakd, what is it you wanted to express here?

sjaakd: > what is it you wanted to express here?

my part only concerned the sentence about the bean convention (you know, property, getter, setter). 

The new `automapping` concerns beans. However, its incomplete. We have the forging mechanism also for iterables, maps and more recently streams. MapStruct cannot forge  a mapping for other 'classes of types' (although class is a bit of an overloaded name, but I reckon you know what I mean).

gunnarmorling: The thing is that we not require JavaBeans (in the sense of the spec) really. We can access fields directly (JavaBeans don't make any assumption on fields) and we can work with other naming patterns than mandated by JavaBeans, such as "withFoo()". Hence I found that sentence confusing.

sjaakd: I understand.. How should we rephrase it? We don't generate mappings always....  That's what I tried to point out.

filiphr: OK now I understand what @sjaakd meant. Maybe we can rephrase it into "MapStruct will only create a new mapping method if and only if the source and target are properties of a Bean and they themselves are Beans, i.e. it is not going to work if they are a `Collection` or a `Map`"

Issue #1060, opened 2017-02-01
Odd mapping generated for FishTankMapper


gunnarmorling:
Something looks weird there. Why does it generate a `fishTankToFishDto()` method? Two questions I see really:

* why does it try to map fish tank to fish DTO (the elements are on different levels in the tree)?
* why is this an update method instead of a regular mapping method with a return value?

sjaakd:
> why does it try to map fish tank to fish DTO (the elements are on different levels in the tree)?

Nested mappings were never limited to a symmetrical tree. It works exactly as before.

>why is this an update method instead of a regular mapping method with a return value?

Checkout: NestedTargetPropertiesTest#shouldMapNestedComposedTarget. I need update methods because the mapping for the first source parameter would influence the mapping for the next source parameter.

The generated code works exactly as before (even before the auto-mapping). I only added the FishTankMapper test case to show a problem in my initial solution, which was shown by tests of @snowe2010. See issue #1011. I only make use of the auto-mapping do simplify the code and to solve the problem in #1011.

For more info, see #1057 for more info.

gunnarmorling:
> Nested mappings were never limited to a symmetrical tree.

Yes, ok, but here the tree is symmetrical, right? So I'm wondering why we don't create a mapping method for each level in the tree. It works as I'd expect for `WaterPlant(Dto)`, it's not clear to me why it's different for `Fish`.

> I need update methods because the mapping for the first source parameter would influence the mapping for the next source parameter.

But we only have a single source parameter in the fish tank case, no?

> The generated code works exactly as before (even before the auto-mapping)

Hum, but the automatic mapping of fish and water plant without any user-provided method is only done since the introduction of auto-mapping, right?

sjaakd:
> Yes, ok, but here the tree is symmetrical, right? So I'm wondering why we don't create a mapping method for each level in the tree. It works as I'd expect for WaterPlant(Dto), it's not clear to me why it's different for Fish.

That's exactly why I wrote #1057 and send you guys mails. I think we never considered the combination nested source  / nested target properties icw auto-mapping. I just found out while solving an entirely different problem and was triggered by some remarks @agudian made about the eclipse plugin. 

You could indeed claim that the (defined by @Mapping) symmetrical mappings are a subset of the complete set of mappings. We have to recognize them in a bean mapping and they should be handled separately. I don't think this is easy.

> But we only have a single source parameter in the fish tank case, no?

[...]

PR #1071, opened 2017-02-07
#1060 Use update methods if there are multiple source parameters

filiphr:
This PR adds the ability to force `PropertyMappings` to create update methods, when the main method is not an update one.

If we do composite mapping (mapping with multiple source parameters), we need to make sure that they won't overwrite non-overlapping objects. Therefore we use update methods. However, by regular create methods with only one source parameter there is no need to create update methods.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
@@ -220,7 +220,6 @@ public abstract class MappingMethod extends ModelElement {
     @Override
     public int hashCode() {
         int hash = 7;
-        hash = 83 * hash + (this.name != null ? this.name.hashCode() : 0);
         hash = 83 * hash + (this.parameters != null ? this.parameters.hashCode() : 0);
         hash = 83 * hash + (this.returnType != null ? this.returnType.hashCode() : 0);
         return hash;
@@ -237,10 +236,10 @@ public abstract class MappingMethod extends ModelElement {
         if ( getClass() != obj.getClass() ) {
             return false;
         }
+        //Do not add name to the equals check.
+        //Reason: Whenever we forge methods we can reuse mappings if they are the same. However, if we take the name
+        // into consideration, they'll never be the same, because we create safe methods names.
         final MappingMethod other = (MappingMethod) obj;
-        if ( (this.name == null) ? (other.name != null) : !this.name.equals( other.name ) ) {
-            return false;
-        }
         if ( this.parameters != other.parameters &&
             (this.parameters == null || !this.parameters.equals( other.parameters )) ) {
             return false;
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/NormalTypeMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/NormalTypeMappingMethod.java
@@ -71,7 +71,7 @@ public abstract class NormalTypeMappingMethod extends MappingMethod {
     @Override
     public int hashCode() {
         final int prime = 31;
-        int result = 1;
+        int result = super.hashCode();
         result = prime * result + ( ( getResultType() == null ) ? 0 : getResultType().hashCode() );
         return result;
     }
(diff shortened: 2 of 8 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
```diff
@@ -238,9 +238,6 @@ public boolean equals(Object obj) {
             return false;
         }
         final MappingMethod other = (MappingMethod) obj;
-        if ( (this.name == null) ? (other.name != null) : !this.name.equals( other.name ) ) {
```
filiphr: I think that this was a bug that we had before. Whenever we forge methods we can reuse mappings if they are the same. However, if we take the name into consideration, they'll never be the same, because we create save method names.

processor/src/main/java/org/mapstruct/ap/internal/model/NormalTypeMappingMethod.java
```diff
@@ -89,6 +89,10 @@ public boolean equals(Object obj) {
         }
         NormalTypeMappingMethod other = (NormalTypeMappingMethod) obj;
 
+        if ( !super.equals( obj ) ) {
```
filiphr: This was a bug that was not calling the super equals methods and thus the `parameters` and the `returnType` was not taken into consideration for the equals methods.

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
sjaakd: Can you add some comment explaining this? Why you exclude name?

processor/src/main/java/org/mapstruct/ap/internal/model/NormalTypeMappingMethod.java
sjaakd: ok..

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -608,7 +619,7 @@ private Assignment forgeMapping(SourceRHS sourceRHS) {
             Type returnType;
             // there's only one case for forging a method with mapping options: nested target properties.
             // they should always forge an update method
-            if ( method.isUpdateMethod() || forgeMethodWithMappingOptions != null ) {
+            if ( method.isUpdateMethod() || forceUpdateMethod ) {
```
sjaakd: so if i understand you correctly you now have only update methods when you have nested target parameters and multiple source parameters, right? Is there a test case in which you check this?

processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
filiphr: Comment in the code?

The problem is that if take name into consideration and we always generate safe names then 2 forge methods will always be different. If I didn't remove it then the test for the `DomainDtoWithNvmsDefaultMapper` was failing because we were not reusing the forged (not via automapping, but for `Iterables`) `protected Set<Long> stringListToLongSet(List<String> list)`.

This is connected with adding the `super.equals( obj )` in the `NormalTypeMappingMethod`. What I can do is to add the `parameters` and `returnType` to the `equals` of the `NormalTypeMappingMethod`

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
filiphr: Yes, and of course if the method was an update method from before.

Well the `NestedTargetPropertiesTest` covers it. Now the `FishTankMapper` has a non update method and the `ChartEntryToArtist` an update methods. Maybe I need to add a fixture test for the `NestedTargetPropertiesTest`? That way we can actually test what is created is OK.

filiphr: I just realized that I need to change the comment above the if as well. It is no longer correct.

sjaakd:
> Comment in the code?

Yep. I understand the problem. But its easy forget the reasoning. So a comment would trigger somebody (like myself) to implement and add this to the equals method.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: Ok

sjaakd:
👍 go ahead and merge.

filiphr:
Thanks for the review @sjaakd

Issue #1057, opened 2017-02-01
Control automapping from root level by allowing @Mapping


sjaakd:
Besides the prototype methods in a shared config, it should be possible to control the auto-mapping from the root object graph method. With control I mean:
1. ignore a target
2. direct MapStruct to select the right target if names between source and target do not match.

Currently we have nested source / nested target mappings `@Mapping` to allow arbitrary selection if properties in a source object graph and map those to arbitrary properties in the target object graph. This is done by `@Mapping( target = "tbean.propX.propX1", source = "sbean.propY.propY1.propY11" )`. For all clarity: it does not seem to make sense add an ignore option like for instance: `@Mapping( target = "tbean.propX.propX2", ignore = true )`, since this property would not have been mapped anyway.

Auto-mapping changes things. MapStruct starts to generate mappings automatically based on names. It is essentially different from the nested source / nested target concept however.  It requires the source object graph to mirror its target object graph (they are more or less supposed to be symmetrical).  MapStruct generates a mapping between each of the properties.

So suppose source:
```
sbean.propX.propX1 -> tbean.propX.propX1
sbean.propX.propX2 -> tbean.propX.propX2
sbean.propY.propY1 -> tbean.propB.probB1
```

Now suddenly it would make sense to define mappings.
`@Mapping( target = "tbean.propX.propX2", ignore = true)`
`@Mapping( target = "tbean.propB.probB1", source="sbean.propY.propY1")`

1. Do we want to implement this? In principle we have the prototype methods to do this, but that mechanism might be cumbersome to use
2.  Do we want to use the same notation? Also for ignores?. Any problems foreseen here?
3. Is there a mixed scenario conceivable in which the nested source / nested target concept is mixed with auto-mappings? If so, if we reuse the current mechanism we possibly need to use update mappings.

agudian:
I commented that at another point already, but for the sake of the discussion here:

I would expect that the following two declarations would create the same results (just with a different visibility of `fishToFishDto`:

```java
    @Mapping(target = "fish.kind", source = "fish.type")
    FishTankDto toFishTankDto(FishTank source);
```
is basically just a short-cut to this:
```java
    FishTankDto toFishTankDto(FishTank source);

    @Mapping(target = "kind", source = "type")
    FishDto fishToFishDto(Fish fish);
```

And to get the blood pressure up, I'd even say that supporting this behaviour is more important than supporting the selected nested-source to nested-target mapping 😱.

[...]

PR #1072, opened 2017-02-10
#1057 Controlling forged name based bean mapping from root level with @Mapping

sjaakd:
very tiny first start... @filiphr .. if you want have a look.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -46,6 +46,7 @@ import org.mapstruct.ap.internal.model.dependency.GraphAnalyzer.GraphAnalyzerBui
 import org.mapstruct.ap.internal.model.source.ForgedMethod;
 import org.mapstruct.ap.internal.model.source.ForgedMethodHistory;
 import org.mapstruct.ap.internal.model.source.Mapping;
+import org.mapstruct.ap.internal.model.source.MappingOptions;
 import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.model.source.PropertyEntry;
 import org.mapstruct.ap.internal.model.source.SelectionParameters;
@@ -574,6 +575,12 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         }
 
         private ReportingPolicyPrism getUnmappedTargetPolicy() {
+            MappingOptions mappingOptions = method.getMappingOptions();
+            if ( mappingOptions.getBeanMapping() != null &&
+                mappingOptions.getBeanMapping().getReportingPolicy() != null ) {
+                return mappingOptions.getBeanMapping().getReportingPolicy();
+            }
+
             MapperConfiguration mapperSettings = MapperConfiguration.getInstanceOn( ctx.getMapperTypeElement() );
 
             return mapperSettings.unmappedTargetPolicy( ctx.getOptions() );
@@ -584,12 +591,11 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             // fetch settings from element to implement
             ReportingPolicyPrism unmappedTargetPolicy = getUnmappedTargetPolicy();
 
-            //we handle forged methods differently than the usual source ones. in
-            if ( method instanceof ForgedMethod ) {
+            //we handle automapping forged methods differently than the usual source ones. in
+            if ( method instanceof ForgedMethod && ( (ForgedMethod) method ).isAutoMapping() ) {
 
                 ForgedMethod forgedMethod = (ForgedMethod) this.method;
-                if ( forgedMethod.isAutoMapping()
-                    && ( targetProperties.isEmpty() || !unprocessedTargetProperties.isEmpty() ) ) {
+                if ( targetProperties.isEmpty() || !unprocessedTargetProperties.isEmpty() ) {
 
                     if ( forgedMethod.getHistory() == null ) {
                         Type sourceType = this.method.getParameters().get( 0 ).getType();
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -664,7 +664,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             ReportingPolicyPrism unmappedTargetPolicy = getUnmappedTargetPolicy();
 
             //we handle automapping forged methods differently than the usual source ones. in
-            if ( method instanceof ForgedMethod && ( (ForgedMethod) method ).isAutoMapping() ) {
+            if ( method instanceof ForgedMethod && ( (ForgedMethod) method ).isForgedNamedBased() ) {
 
                 ForgedMethod forgedMethod = (ForgedMethod) this.method;
                 if ( targetProperties.isEmpty() || !unprocessedTargetProperties.isEmpty() ) {
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -147,6 +147,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 applyParameterNameBasedMapping();
             }
 
+            // Process the unprocessed defined targets
             handleUnprocessedDefinedTargets();
 
             // report errors on unmapped properties
@@ -220,11 +221,13 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         }
 
         /**
-         * If there were nested defined targets that have not been handled. The we need to process them at he end.
+         * If there were nested defined targets that have not been handled. Then we need to process them at the end.
          */
         private void handleUnprocessedDefinedTargets() {
             Iterator<Entry<String, List<Mapping>>> iterator = unprocessedDefinedTargets.entrySet().iterator();
 
+            // For each of the unprocessed defined targets forge a mapping for each of the
+            // method source parameters. The generated mappings are not going to use forged name based mappings.
             while ( iterator.hasNext() ) {
                 Entry<String, List<Mapping>> entry = iterator.next();
                 String propertyName = entry.getKey();
@@ -315,6 +318,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             boolean errorOccurred = false;
             Set<String> handledTargets = new HashSet<String>();
 
+            // first we have to handle nested target mappings
             if ( method.getMappingOptions().hasNestedTargetReferences() ) {
                 handleDefinedNestedTargetMapping( handledTargets );
             }
@@ -355,6 +359,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             unprocessedSourceParameters.removeAll( holder.getProcessedSourceParameters() );
             propertyMappings.addAll( holder.getPropertyMappings() );
             handledTargets.addAll( holder.getHandledTargets() );
+            // Store all the unprocessed defined targets.
             for ( Entry<PropertyEntry, List<Mapping>> entry : holder.getUnprocessedDefinedTarget().entrySet() ) {
                 if ( entry.getValue().isEmpty() ) {
                     continue;
@@ -424,6 +429,8 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             }
 
             // its a constant
+            // if we have an unprocessed target that means that it most probably is nested and we should
+            // not generated any mapping for it now. Eventually it will be done though
             else if ( mapping.getConstant() != null && !unprocessedDefinedTargets.containsKey( propertyName ) ) {
 
                 propertyMapping = new ConstantMappingBuilder()
@@ -441,6 +448,8 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             }
 
             // its an expression
+            // if we have an unprocessed target that means that it most probably is nested and we should
+            // not generated any mapping for it now. Eventually it will be done though
             else if ( mapping.getJavaExpression() != null && !unprocessedDefinedTargets.containsKey( propertyName ) ) {
 
                 propertyMapping = new JavaExpressionMappingBuilder()
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -89,6 +89,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         private final Set<String> existingVariableNames = new HashSet<String>();
         private Map<String, List<Mapping>> methodMappings;
         private SingleMappingByTargetPropertyNameFunction singleMapping;
+        private final Map<String, List<Mapping>> unprocessedDefinedTargets = new HashMap<String, List<Mapping>>();
 
         public Builder mappingContext(MappingBuilderContext mappingContext) {
             this.ctx = mappingContext;
@@ -145,6 +146,8 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 // map parameters without a mapping
                 applyParameterNameBasedMapping();
             }
+
+            handleUnprocessedDefinedTargets();
 
             // report errors on unmapped properties
             reportErrorForUnmappedTargetPropertiesIfRequired();
@@ -214,6 +217,50 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 beforeMappingMethods,
                 afterMappingMethods
             );
+        }
+
+        /**
+         * If there were nested defined targets that have not been handled. The we need to process them at he end.
+         */
+        private void handleUnprocessedDefinedTargets() {
+            Iterator<Entry<String, List<Mapping>>> iterator = unprocessedDefinedTargets.entrySet().iterator();
+
+            while ( iterator.hasNext() ) {
+                Entry<String, List<Mapping>> entry = iterator.next();
+                String propertyName = entry.getKey();
+                if ( !unprocessedTargetProperties.containsKey( propertyName ) ) {
+                    continue;
+                }
+                List<Parameter> sourceParameters = method.getSourceParameters();
+                boolean forceUpdateMethod = sourceParameters.size() > 1;
+                for ( Parameter sourceParameter : sourceParameters ) {
+                    SourceReference reference = new SourceReference.BuilderFromProperty()
+                        .sourceParameter( sourceParameter )
+                        .name( propertyName )
+                        .build();
+
+                    MappingOptions mappingOptions = extractAdditionalOptions( propertyName, true );
+                    PropertyMapping propertyMapping = new PropertyMappingBuilder()
+                        .mappingContext( ctx )
+                        .sourceMethod( method )
+                        .targetWriteAccessor( unprocessedTargetProperties.get( propertyName ) )
+                        .targetReadAccessor( getTargetPropertyReadAccessor( propertyName ) )
+                        .targetPropertyName( propertyName )
+                        .sourceReference( reference )
+                        .existingVariableNames( existingVariableNames )
+                        .dependsOn( mappingOptions.collectNestedDependsOn() )
+                        .forgeMethodWithMappingOptions( mappingOptions )
+                        .forceUpdateMethod( forceUpdateMethod )
+                        .forgedNamedBased( false )
+                        .build();
+
+                    if ( propertyMapping != null ) {
+                        unprocessedTargetProperties.remove( propertyName );
+                        iterator.remove();
+                        propertyMappings.add( propertyMapping );
+                    }
+                }
+            }
         }
 
         /**
@@ -291,6 +338,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 // In order to avoid: "Unknown property foo in return type" in case of duplicate
                 // target mappings
                 unprocessedTargetProperties.remove( handledTarget );
+                unprocessedDefinedTargets.remove( handledTarget );
             }
 
             return errorOccurred;
@@ -307,7 +355,12 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             unprocessedSourceParameters.removeAll( holder.getProcessedSourceParameters() );
             propertyMappings.addAll( holder.getPropertyMappings() );
             handledTargets.addAll( holder.getHandledTargets() );
-
+            for ( Entry<PropertyEntry, List<Mapping>> entry : holder.getUnprocessedDefinedTarget().entrySet() ) {
+                if ( entry.getValue().isEmpty() ) {
+                    continue;
+                }
+                unprocessedDefinedTargets.put( entry.getKey().getName(), entry.getValue() );
+            }
         }
 
         private boolean handleDefinedMapping(Mapping mapping, Set<String> handledTargets) {
@@ -318,6 +371,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
             TargetReference targetRef = mapping.getTargetReference();
             PropertyEntry targetProperty = first( targetRef.getPropertyEntries() );
+            String propertyName = targetProperty.getName();
 
             // unknown properties given via dependsOn()?
             for ( String dependency : mapping.getDependsOn() ) {
@@ -361,7 +415,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                         .dependsOn( mapping.getDependsOn() )
                         .defaultValue( mapping.getDefaultValue() )
                         .build();
-                    handledTargets.add( targetProperty.getName() );
+                    handledTargets.add( propertyName );
                     unprocessedSourceParameters.remove( sourceRef.getParameter() );
                 }
                 else {
@@ -370,7 +424,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             }
 
             // its a constant
-            else if ( mapping.getConstant() != null ) {
+            else if ( mapping.getConstant() != null && !unprocessedDefinedTargets.containsKey( propertyName ) ) {
 
                 propertyMapping = new ConstantMappingBuilder()
                     .mappingContext( ctx )
@@ -387,7 +441,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             }
 
             // its an expression
-            else if ( mapping.getJavaExpression() != null ) {
+            else if ( mapping.getJavaExpression() != null && !unprocessedDefinedTargets.containsKey( propertyName ) ) {
 
                 propertyMapping = new JavaExpressionMappingBuilder()
                     .mappingContext( ctx )
@@ -500,6 +554,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                                 .defaultValue( mapping != null ? mapping.getDefaultValue() : null )
                                 .existingVariableNames( existingVariableNames )
                                 .dependsOn( mapping != null ? mapping.getDependsOn() : Collections.<String>emptyList() )
+                                .forgeMethodWithMappingOptions( extractAdditionalOptions( targetPropertyName, false ) )
                                 .build();
 
                             unprocessedSourceParameters.remove( sourceParameter );
@@ -523,6 +578,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 if ( propertyMapping != null ) {
                     propertyMappings.add( propertyMapping );
                     targetPropertyEntriesIterator.remove();
+                    unprocessedDefinedTargets.remove( targetPropertyName );
                 }
             }
         }
@@ -560,14 +616,30 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                             .selectionParameters( mapping != null ? mapping.getSelectionParameters() : null )
                             .existingVariableNames( existingVariableNames )
                             .dependsOn( mapping != null ? mapping.getDependsOn() : Collections.<String>emptyList() )
+                            .forgeMethodWithMappingOptions( extractAdditionalOptions( targetProperty.getKey(), false ) )
                             .build();
 
                         propertyMappings.add( propertyMapping );
                         targetPropertyEntriesIterator.remove();
                         sourceParameters.remove();
+                        unprocessedDefinedTargets.remove( targetProperty.getKey() );
                     }
                 }
             }
+        }
+
+        private MappingOptions extractAdditionalOptions(String targetProperty, boolean restrictToDefinedMappings) {
+            MappingOptions additionalOptions = null;
+            if ( unprocessedDefinedTargets.containsKey( targetProperty ) ) {
+
+                Map<String, List<Mapping>> mappings = new HashMap<String, List<Mapping>>();
+                mappings.put(
+                    targetProperty,
+                    unprocessedDefinedTargets.get( targetProperty )
+                );
+                additionalOptions = MappingOptions.forMappingsOnly( mappings, restrictToDefinedMappings );
+            }
+            return additionalOptions;
         }
 
         private Accessor getTargetPropertyReadAccessor(String propertyName) {
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
@@ -128,28 +128,39 @@ public class NestedTargetPropertyMappingHolder {
             Set<String> handledTargets = new HashSet<String>();
             List<PropertyMapping> propertyMappings = new ArrayList<PropertyMapping>();
 
+            // first we group by the first property in the target properties and for each of those
+            // properties we get the new mappings as if the first property did not exist.
             GroupedTargetReferences groupedByTP = groupByTargetReferences( method.getMappingOptions() );
             Map<PropertyEntry, List<Mapping>> unprocessedDefinedTarget = new HashMap<PropertyEntry, List<Mapping>>();
 
             for ( Map.Entry<PropertyEntry, List<Mapping>> entryByTP : groupedByTP.poppedTargetReferences.entrySet() ) {
                 PropertyEntry targetProperty = entryByTP.getKey();
+                //Now we are grouping the already popped mappings by the source parameter(s) of the method
                 GroupedBySourceParameters groupedBySourceParam = groupBySourceParameter(
                     entryByTP.getValue(),
                     groupedByTP.singleTargetReferences.get( targetProperty )
                 );
                 boolean forceUpdateMethod = groupedBySourceParam.groupedBySourceParameter.keySet().size() > 1;
 
+                // All not processed mappings that should have been applied to all are part of the unprocessed
+                // defined targets
                 unprocessedDefinedTarget.put( targetProperty, groupedBySourceParam.notProcessedAppliesToAll );
                 for ( Map.Entry<Parameter, List<Mapping>> entryByParam : groupedBySourceParam
                     .groupedBySourceParameter.entrySet() ) {
 
                     Parameter sourceParameter = entryByParam.getKey();
 
+                    // Lastly we need to group by the source references. This will allow us to actually create
+                    // the next mappings by popping source elements
                     GroupedSourceReferences groupedSourceReferences = groupByPoppedSourceReferences(
                         entryByParam.getValue(),
                         groupedByTP.singleTargetReferences.get( targetProperty )
                     );
 
+                    // For all the groupedBySourceReferences we need to create property mappings
+                    // from the Mappings and not restrict on the defined mappings (allow to forge name based mapping)
+                    // if we have composite methods i.e. more then 2 parameters then we have to force a creation
+                    // of an update method in our generation
                     for ( Map.Entry<PropertyEntry, List<Mapping>> entryBySP : groupedSourceReferences
                         .groupedBySourceReferences
                         .entrySet() ) {
@@ -180,6 +191,9 @@ public class NestedTargetPropertyMappingHolder {
                         handledTargets.add( entryByTP.getKey().getName() );
                     }
 
+                    // For the nonNested mappings (assymetric) Mappings we also forge mappings
+                    // However, here we do not forge name based mappings and we only
+                    // do update on the defined Mappings.
                     if ( !groupedSourceReferences.nonNested.isEmpty() ) {
                         MappingOptions nonNestedOptions = MappingOptions.forMappingsOnly(
                             groupByTargetName( groupedSourceReferences.nonNested ),
@@ -220,24 +234,43 @@ public class NestedTargetPropertyMappingHolder {
          * The target references are popped. The {@code List<}{@link Mapping}{@code >} are keyed on the unique first
          * entries of the target references.
          *
-         * So, take
+         * <p>
+         * <p>
+         * Group all target references by their first property and for each such mapping use a new one where the
+         * first property will be removed from it. If a {@link org.mapstruct.Mapping} cannot be popped, i.e. it
+         * contains a non nested target property just keep it as is (this is usually needed to control how an
+         * intermediary level can be mapped).
          *
-         * targetReference 1: propertyEntryX.propertyEntryX1.propertyEntryX1a
-         * targetReference 2: propertyEntryX.propertyEntryX2
-         * targetReference 3: propertyEntryY.propertyY1
-         * targetReference 4: propertyEntryZ
+         * <p>
+         * <p>
+         * We start with the following mappings:
          *
-         * will be popped and grouped into entries:
+         * <pre>
+         * {@literal @}Mapping(target = "fish.kind", source = "fish.type"),
+         * {@literal @}Mapping(target = "fish.name", ignore = true),
+         * {@literal @}Mapping(target = "ornament", ignore = true ),
+         * {@literal @}Mapping(target = "material.materialType", source = "material"),
+         * {@literal @}Mapping(target = "document", source = "report"),
+         * {@literal @}Mapping(target = "document.organisation.name", source = "report.organisationName")
+         * </pre>
          *
-         * propertyEntryX - List ( targetReference1: propertyEntryX1.propertyEntryX1a,
-         * targetReference2: propertyEntryX2 )
-         * propertyEntryY - List ( targetReference1: propertyEntryY1 )
+         * We will get this:
          *
-         * The key will be the former top level property, the MappingOptions will contain the remainders.
+         * <pre>
+         * // All target references are popped and grouped by their first property
+         * GroupedTargetReferences.poppedTargetReferences {
+         *     fish:     {@literal @}Mapping(target = "kind", source = "fish.type"),
+         *               {@literal @}Mapping(target = "name", ignore = true);
+         *     material: {@literal @}Mapping(target = "materialType", source = "material");
+         *     document: {@literal @}Mapping(target = "organization.name", source = "report.organizationName");
+         * }
          *
-         * If the target reference cannot be popped it is stored in a different map. That looks like:
-         *
-         * propertyEntryZ - List ( targetReference4: propertyEntryZ )
+         * //This references are not nested and we they stay as they were
+         * GroupedTargetReferences.singleTargetReferences {
+         *     document: {@literal @}Mapping(target = "document", source = "report");
+         *     ornament: {@literal @}Mapping(target = "ornament", ignore = true );
+         * }
+         * </pre>
          *
          * @param mappingOptions that need to be used to create the {@link GroupedTargetReferences}
          *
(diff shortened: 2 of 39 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -137,14 +137,12 @@ public BeanMappingMethod build() {
                 return null;
             }
 
-            if ( !method.getMappingOptions().isRestrictToDefinedMappings() ) {
```
filiphr: I think that we should always apply `property` and `parameter` name based mappings. The problem with getting the `fish.name` from the `FishTank` should be solved by doing the popping of source reference when we create the forged methods

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -608,12 +612,11 @@ private void reportErrorForUnmappedTargetPropertiesIfRequired() {
             // fetch settings from element to implement
             ReportingPolicyPrism unmappedTargetPolicy = getUnmappedTargetPolicy();
 
-            //we handle forged methods differently than the usual source ones. in
-            if ( method instanceof ForgedMethod ) {
+            //we handle automapping forged methods differently than the usual source ones. in
+            if ( method instanceof ForgedMethod && ( (ForgedMethod) method ).isAutoMapping() ) {
```
filiphr: I think we should only do this for forged automappings and do the normal reporting for the other forged methods. Especially with the addition to ignore unmapped target properties for our own ForgedMethods (which needs to be revised)

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
```diff
@@ -74,7 +80,14 @@ public static MappingOptions empty() {
      * @return MappingOptions with only regular mappings
      */
     public static MappingOptions forMappingsOnly( Map<String, List<Mapping>> mappings ) {
```
filiphr: Maybe we need to pass the `BeanMapping`. I think that we only need to explicitly ignore when we have multiple source parameters.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
sjaakd: I think you're right for this case 👍 

But now consider:

`@Mapping(target = "song.artist.label.studio.city", source = "chartEntry1.city" ),`

Suppose there would be a property: `number` in `studio` and a `number` in `chartEntry1` I would not expect them to be mapped automatically, right? And which one would be valid? the one from `chartEntry1` or `chartEntry2`?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
sjaakd: With my latest change it becomes more apparent: 
* the case here is forged bean mapping method
* more-over, this executed only when its a 'name-based' mapping. 

So what if we implement a method on method called: `isForgedNameBased`?

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
```diff
@@ -225,7 +225,7 @@ public ForgedMethodHistory getHistory() {
     }
 
     public boolean isAutoMapping() {
-        return mappingOptions.getValueMappings().isEmpty();
+        return MappingOptions.empty().equals( mappingOptions );
```
sjaakd: hey. interesting. what's the diff?

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMapping.java
```diff
@@ -61,12 +63,25 @@ public static BeanMapping fromPrism(BeanMappingPrism beanMapping, ExecutableElem
...
             resultTypeIsDefined ? beanMapping.resultType() : null );
 
-        return new BeanMapping(cmp, nullValueMappingStrategy );
+        //TODO Do we want to add the reporting policy to the BeanMapping as well? To give more granular support?
+        return new BeanMapping( cmp, nullValueMappingStrategy, null );
     }
 
-    private BeanMapping( SelectionParameters selectionParameters, NullValueMappingStrategyPrism nvms ) {
+    /**
+     * This method should be used to generate BeanMappings for our internal generated Mappings. Like forged update
```
sjaakd: forged update? Or all forgedBeanMappings

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMapping.java
```diff
@@ -61,12 +63,25 @@ public static BeanMapping fromPrism(BeanMappingPrism beanMapping, ExecutableElem
...
+     * methods.
+     *
+     * @return bean mapping that needs to be used for Mappings
+     */
+    public static BeanMapping forMappingsOnly() {
```
sjaakd: forForgedMethod ?

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
sjaakd: I don't get your comment here.. What do you mean?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
filiphr: You are also right 😄. I think that we need to distinguish between composite (multiple) source parameters and single source parameters. For single source parameters, we need to ignore it, otherwise apply it. 

I think we can only change how we define the `MappingOptions`, because at the time we created them we know whether there are multiple source parameters or not. Something like `forceUpdateMethod`

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
filiphr: New method on `ForgedMethod` or on `Method`? 

`isForgeNameBased` would be similar to the `isAutoMapping` right?

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
```diff
@@ -182,6 +195,7 @@ public boolean hasNestedTargetReferences() {
     public Map<Parameter, MappingOptions> groupBySourceParameter() {
 
         Map<Parameter, List<Mapping>> mappingsKeyedByParameterType = new HashMap<Parameter, List<Mapping>>();
```
sjaakd: `mappingsKeyedByParameterType` = `mappingsKeyedByParameter`. This is deliberate. Look at the equals in `Parameter` and what makes Parameter unique..

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
filiphr: Well `mappingOptions.getValueMappings()` is a list of `@ValueMapping` which is used enums. However, even though the modified check is more "correct" it is still not correct, because we create non-automapping `ForgedMethod` with `null` `MappingOptions`. I think that we need some explicit parameter or a new `AutoMappingForgedMethod` that will explicitly show that it was created due to automapping.

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMapping.java
filiphr: Basically all our forged mappings. However, after thinking a bit more. We might _only_ need to have `ReportingPolicy.IGNORE` for composite methods.

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMapping.java
filiphr: 😄, that was my first name and I renamed it. After the comments, it suits it more.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
sjaakd: > You are also right 😄. I think that we need to distinguish between composite (multiple) source parameters and single source parameters. For single source parameters, we need to ignore it, otherwise apply it.

I think the distinction is between symmetric and asymmetric (keep on using that term, sorry.)  In would not expect `number` to be mapped in any occasion. Since `number` in `chartEntry` would perhaps reflect the number in the charts, while `number` in `studio` would perhaps be a `street number` or something along that lines.

The risk of such confusion is not so high if you are at the same level in an object graph (source level matches target level) because you are most likely talking about the same things. In an asymmetric situation you are cherry-picking mappings. It would be very risky to assume something about the other mappings at that level..

Hope I make sense somehow.. :smile:

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
filiphr: We can rename that as well. It was there from before...

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
sjaakd: As an addition to my own comment.. `city` above is now a simple type (String) but what if it would be a bean itself (on both sides). Then it would again be logical to resume  name based mappings inside `city`.

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
filiphr: This is connected with the `isRestrictToDefinedMappings()` discussion. In a symmetric case we want to not explicitly ignore unmapped target properties. However, in assymetric we want to have it explicitly.

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
sjaakd: ahh.. of course.. I meant getMappings.. but this is far more safe..

sjaakd: 👍

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -304,31 +306,79 @@ private void handleDefinedNestedTargetMapping(Set<String> handledTargets) {
 
                 Map<Parameter, MappingOptions> optionsBySourceParam = entryByTP.getValue().groupBySourceParameter();
                 boolean forceUpdateMethod = optionsBySourceParam.keySet().size() > 1;
+                PropertyEntry targetProperty = entryByTP.getKey();
                 for ( Entry<Parameter, MappingOptions> entryByParam : optionsBySourceParam.entrySet() ) {
```
filiphr: I think that we need to extract all this loops and creation of mapping options into one method / class instead of doing it one by one. And most importantly I think that we have to add some unit tests only for that. I think that the entire logic here is quite complex.

sjaakd: I  think you did a great job already 👍 .. There is room for improvement. Need to take some time to check the generated code. In the mean time I pushed a small nitty-witty check style / javadoc commit..

sjaakd:
Hey @filiphr ... I checked the generated code.. Something strange is indeed happening..

```java
     fishTankDto.setFish( fishTankToFishDto( source ) );
        fishTankDto.setQuality( waterQualityToWaterQualityDto( source.getQuality() ) );
        fishTankDto.setQuality( fishTankToWaterQualityDto( source ) );
```

I would expect that `fish` would be mapped to `fish` and `quality` to `quality` (in one method)..  They are symmetric at  this level.. I Need to debug later on..

sjaakd:
Could it be: you now have GroupedSourceReferences. I would expect however that mappings are in the same MappingOptions. Shouldn't this functionality 'GroupedSourceReference' not simply be in `MappingOptions`? Just a quick thought... (without further checking)..

filiphr:
>I would expect that fish would be mapped to fish and quality to quality (in one method).. They are symmetric at this level.. I Need to debug later on..

I found the problem I was always creating a mapping for non-nested. My last commit solves this.

filiphr:
>Could it be: you now have GroupedSourceReferences. I would expect however that mappings are in the same MappingOptions. Shouldn't this functionality 'GroupedSourceReference' not simply be in MappingOptions? Just a quick thought... (without further checking)...

As I commented in the `MappingOptions`. I think that we need to extract all that logic out of the `MappingOptions`. I think it is too complex to be there.

I added `GroupedSourceReferences` because when we are popping the source references we need to keep the non-nested mappings somehow and not pop them. That is why that one has 2 main properties:
* `Map<PropertyEntry, MappingOptions> groupedByProperty` -> Which stores the mapping options for each level of the source property entries and for which we need to create a new `SourceReference` with the read accessor for the properties
* `MappingOptions nonNestedOptions` -> Are all the `Mapping`(s) for non-nested source properties. For this we need to create a new method with the source of the mapping method.

filiphr:
@sjaakd I know why the `expression` is failing. It is because in `MappingOptions.groupBySourceParameter` we have no source parameter. And `appliesToAll` is never applied, the result is an empty map, because of this we never generate the mapping there. The same problem will occur for the other mappings if they are the only one which are defined. 

Maybe in a case like this we need to generate at lease a mapping for the current source parameter.

filiphr:
With my last commit theoretically all the tests should pass. Tomorrow I will cleanup the commits so this can be reviewed.

sjaakd:
> MappingOptions.groupBySourceParameter we have no source parameter

So, it was a similar case as for the 'ignore'?

In principle MapStruct should apply autmapping in this symmetrical case.  So quality should be mapped to quality, report to report. They carry the same name. On top of that the expression should be applied.

Makes me wonder what happens if report on the target side would have had a different name. E.g. document. Guess it would have to look like this:

```java
    @Mappings({
        @Mapping(target = "fish.kind", source = "fish.type"),
        @Mapping(target = "fish.name", expression = "java(\"Jaws\")"),
        @Mapping(target = "plant", ignore = true ),
        @Mapping(target = "ornament", ignore = true ),
        @Mapping(target = "material", ignore = true),
        @Mapping(target = "quality.documen", source="quality.report" ),
        @Mapping(target = "quality.document.organisation.name", expression = "java(\"Dunno\")" )
    })
    FishTankDto map( FishTank source  );

```

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/mixed/AutomappingAndNestedTest.java
```diff
@@ -200,7 +204,7 @@ public void shouldAutomapAndHandleSourceAndTargetPropertyNestingAndExpresion() {
 
         assertThat( target.getQuality() ).isNotNull();
         assertThat( target.getQuality().getReport() ).isNotNull();
-        assertThat( target.getQuality().getReport().getVerdict() ).isEqualTo( "PASSED" );
```
sjaakd: Hey @filiphr .. this is not right.. should be automapped.. 
quality --> quality automapped
report --> report automapped
verdict is a property both in report and reportdto. So Should be automated.

sjaakd:
I've written the test case above, but in doing so, spotted a 'wrong fix' (I think)  in my original test case. I saw you corrected verdict.. But that should actually be mapped.

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/mixed/AutomappingAndNestedTest.java
filiphr: You are right. Maybe if we don't have anything we just need to leave it and let MapStruct do name based mapping here

sjaakd: That's the idea.. just pass the popped MappingOptions to the property mapping. Mapstruct will forge a new BeanMapping with these popped MappingOptions. 

Considered from a `BeanMapping` method perspective :

If the top of the stack is reached (popping yields null), its either a 
1. plain `a` to `b` mapping or nested source mapping `a.a1` to `b` (handled by property mapping itself).
2. name based / parameter based mapping 
3. or a nested target `a` to `b.b1` mapping (for that one the name based / parameter based mapping needs to stop at 'b' level', but resumes when you map `a` to `b1` which is, at that recurrence level exactly case 1. again.

Hope.. I make some sense..

 

If there's something defined on a certain level, the BeanMapping will consider it. For the remainder: name based property mapping + forging auto mappings when its a bean.

filiphr:
@sjaakd the test case is great. And we hit another problem with it. For your test case we are not passing the normal mappings to our popped generated mappings. We eliminate those, for this to work we will have to pass those for generation down the lines as well. For your example we are ignoring the `a` to `b` mapping because we first do the defined nested target mappings for `a.a1` to something.

filiphr:
For the problem with the expressions I think I know why they are not being generated. The thing is that we are passing the source to those generated methods, because we don't know how to map them. I think that to solve this problem we need to consider passing mapping options to the forged named based mappings instead of doing it in the handle.

sjaakd:
Hey @filiphr Nice work... 

javadoc (warnings) and errors: checksyle.

```
[WARNING] Javadoc Warnings
[WARNING] /scratch/jenkins/workspace/mapstruct-pullrequest-builder/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java:230: warning: no @param for mappingOptions
[WARNING] public GroupedTargetReferences groupByTargetReferences(MappingOptions mappingOptions) {
[WARNING] ^
[WARNING] /scratch/jenkins/workspace/mapstruct-pullrequest-builder/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java:264: warning: no @param for mappings
[WARNING] public GroupedBySourceParameters groupBySourceParameter(List<Mapping> mappings,
[WARNING] ^
[WARNING] /scratch/jenkins/workspace/mapstruct-pullrequest-builder/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java:264: warning: no @param for singleTargetReferences
[WARNING] public GroupedBySourceParameters groupBySourceParameter(List<Mapping> mappings,
[WARNING] ^
[WARNING] /scratch/jenkins/workspace/mapstruct-pullrequest-builder/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java:264: warning: no @param for sourceParameters
[WARNING] public GroupedBySourceParameters groupBySourceParameter(List<Mapping> mappings,
[WARNING] ^
[INFO] 
[INFO] --- maven-checkstyle-plugin:2.17:checkstyle (check-style) @ mapstruct-processor ---
[INFO] Starting audit...
[ERROR] /scratch/jenkins/workspace/mapstruct-pullrequest-builder/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java:355: Line has trailing spaces [RegexpSingleline]
```

I would advise to squash it all in one commit.. Splitting in UT and solution does not bring any clarity. You could consider to put the javadoc failures (unrelated to this PR / Issue to a separate commit. A separate PR would be overkill.

I don't think its to complex for one commit. For me: initially it looks ok. I'll will have a more detailed look tomorrow. But perhaps its even better that somebody else will take a look as well.d

filiphr:
Thanks for the last review @sjaakd. I have fixed the Javadoc warnings and the checkstyle error.

I have cleaned up the commits a bit.  Now we have 6 commits:
* All the tests into one
* What you commited towards the fix
* What I commited towards the fix
* The `ReportingPolicy` in the internal `BeanMapping`
* The change of `automapping` into `forgedNameBased` and how we control it
* The last source fixtures

I have removed 2 commits the `RoundContext` Javadoc cleanup, and the `MappingOptions` empty instance (I'll create a separate PR for those 2).

Additionally I moved all the logic for handling into the `NestedTargetPropertyMappingHolder` and I create `MappingOptions` only in the final loop.

Let me know what you think.

@gunnarmorling, @agudian can someone of you do a review of this? I think that it is important to have a new pair of eyes looking at it :)

sjaakd:
@filiphr .. I was just going through your code. 

First of all: I've think you've done a gr8 job. Excellent!

Then, although we discussed it a lot, I'm having a hard time to understand what case is exactly were. 

You did structure it nicely. However, I feel that we could be more  clear in adding javadoc to methods (and if required comments in the code). 

This javadoc / these comments should explain what we are doing from a defined mapping pov. So at one spot you are handling: `@Mapping( target = "a.a1", source "a.b1" )` as comment, or `@Mapping(target="a", source="b")` or `@Mapping(target="a.a1", expression/constant="blah")` or asymmetric target (aka nested target): `@Mapping( target="a.a1.a11", source = "b" )` (or any nested combination on deeper level.  We should also explain that the asymmetric source  is handled by the property mapping. 

We should, at the proper place, explain the popping mechanism, and that we pop just as long that we have the symmetric case `@Mapping(target="a", source="b")` or an asymmetric case (nested target) or nested source left (see above)..

It would be especially nice if we could read this as a kind of logical order of steps somehow from the `BeanMappingBuilder#build` method.

For all clarity: I'm not proposing to refactor the code itself. Just to make clear which steps we take in the `BeanMappingBuilder#build`.

WDYT?

filiphr:
@sjaakd That is a good idea. I will go through the `BeanMappingMethod.Builder` and the `NestedTargetPropertyMappingHolder` and write examples like you explained (I will use the ones that we have in the `FishTankMapper*`).

btw I also added `org.mapstruct.ap.internal.util.Extractor` which basically is `java.util.function.Function` so we can at least use some generics and avoid code duplication :).

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -218,5 +375,123 @@ public NestedTargetPropertyMappingHolder build() {
...
+                        if ( !map.containsKey( key ) ) {
+                            map.put( key, new ArrayList<Mapping>() );
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    private static class GroupedBySourceParameters {
```
sjaakd: This class should not be private. You expose it from this class in the api.

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -218,5 +375,123 @@ public NestedTargetPropertyMappingHolder build() {
...
+     * <pre>
+     * sourceReference 4: propertyEntryZ
+     * sourceReference 5: propertyEntryZ2
+     * </pre>
+     *
+     * <br />
+     * If Mappings that should apply to all were found, but no grouping was found, they will be located in a
+     * different list:
+     */
+    private static class GroupedSourceReferences {
```
sjaakd: Same here. class should be public.

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMapping.java
```diff
@@ -61,12 +63,25 @@ public static BeanMapping fromPrism(BeanMappingPrism beanMapping, ExecutableElem
             beanMapping.qualifiedByName(),
             resultTypeIsDefined ? beanMapping.resultType() : null );
 
-        return new BeanMapping(cmp, nullValueMappingStrategy );
+        //TODO Do we want to add the reporting policy to the BeanMapping as well? To give more granular support?
```
sjaakd: If I understand you correctly, you now just delegate it down to the @MappingConfig, to the @Mapper to the @BeanMapping? 

I (vaguely) remember that I discussed this earlier somehow with @gunnarmorling ... He was not in favor at that point in time. However, it could be another strategy...

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
```diff
@@ -320,6 +320,17 @@ public SourceReference copyForInheritanceTo(SourceMethod method) {
         return new SourceReference( replacement, propertyEntries, isValid );
     }
 
+    public SourceReference pop() {
+        if ( propertyEntries.size() > 1 ) {
```
sjaakd: hey.. is this also done now for the TargetProperties popping? I wanted to do that, but did not get to it.

sjaakd:
> btw I also added org.mapstruct.ap.internal.util.Extractor which basically is java.util.function.Function so we can at least use some generics and avoid code duplication :).

👍 saw that.. !

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: Where do I expose it? The class is only used internally between the for loops. It is not exposed outside of the `Builder`. In the end of the `build` new object (`NestedTargetPropetyMappingHolder`) with the needed properties is created.

sorry if I misunderstood something

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMapping.java
filiphr: Exactly. However, this is not yet exposed to `org.mapstruct.BeanMapping`. It is only used for our internally generated `ForgedMethods` where we don't apply name based mapping and therefore we need to explicitly ignore the unmapped target properties.

Exposing it in the public API via `org.mapstruct.BeanMapping` can be an option, but it is a different issue. I wrote the in order to see if you guys want to have something like this and whether I need to create an issue

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
filiphr: Yes it is. I think you got to it 😄

filiphr:
I added some more Javadoc / comments. Let me know if they are good enough

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: Thanks for the email. Actually the problem was that the methods returning the class should not have been public. They are only used within the `NestedTargetPropertyMappingHolder` so I made them `private`

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
```diff
@@ -0,0 +1,358 @@
...
+
+    @Override
+    public FishTankDto map(FishTank source) {
+        if ( source == null ) {
+            return null;
+        }
+
+        FishTankDto fishTankDto = new FishTankDto();
+
+        fishTankDto.setMaterial( fishTankToMaterialDto( source ) );
```
gunnarmorling: Thanks for that great test!

One thing I'm wondering is why `fishTankToMaterialDto()` takes the entire source object (`FishTank`), wheres `fishToFishDto()` only receives the object it's interested in. Not a blocker at all, just curious about it :)

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapper.java
```diff
@@ -0,0 +1,59 @@
...
+@Mapper
+public interface FishTankMapper {
+
+    FishTankMapper INSTANCE = Mappers.getMapper( FishTankMapper.class );
+
+    @Mappings({
+        @Mapping(target = "fish.kind", source = "fish.type"),
+        @Mapping(target = "fish.name", ignore = true),
+        @Mapping(target = "ornament", source = "interior.ornament"),
+        @Mapping(target = "material.materialType", source = "material"),
```
gunnarmorling: I added this:

    @Mapping(target = "material.manufacturer",  constant = "MMM" ),

Then the test still passes, although it expects that the manufacturer is null. It seems that constant isn't propagated.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
```diff
@@ -0,0 +1,358 @@
...
+            return null;
+        }
+        Ornament ornament = interior.getOrnament();
+        if ( ornament == null ) {
+            return null;
+        }
+        return ornament;
+    }
+
+    protected Fish fishDtoToFish(FishDto fishDto) {
```
gunnarmorling: Why is it that some of the methods are protected and others are private?

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
```diff
@@ -0,0 +1,358 @@
...
+
+        Fish fish = new Fish();
+
+        fish.setType( fishDto.getKind() );
+
+        return fish;
+    }
+
+    private String waterQualityReportDtoOrganisationName(WaterQualityReportDto waterQualityReportDto) {
+
```
gunnarmorling: Would be nice to avoid the empty line.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        Fish fish = new Fish();
+
+        fish.setType( fishDto.getKind() );
+
+        return fish;
+    }
+
+    private String waterQualityReportDtoOrganisationName(WaterQualityReportDto waterQualityReportDto) {
```
gunnarmorling: Shouldn't the name be "waterQualityReportDto*To*OrganisationName"?

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/mixed/_target/WaterQualityOrganisationDto.java
```diff
@@ -16,33 +16,16 @@
...
-    }
-
-    public WaterPlant getPlant() {
-        return plant;
-    }
-
-    public void setPlant(WaterPlant plant) {
-        this.plant = plant;
-    }
+    private String approval;
```
gunnarmorling: Is that property used anywhere actually?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -216,6 +221,52 @@ else if ( !method.isUpdateMethod() && method.getReturnType().isAbstract() ) {
         }
 
         /**
+         * If there were nested defined targets that have not been handled. The we need to process them at he end.
```
gunnarmorling: "The*n* we need to process them at *t*he end".

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMapping.java
gunnarmorling: Can you give an example where it would be useful? Probably best to discuss separately, though. If you like create an issue for the discussion's sake, we can close it at the end if we agree it's not needed.

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/mixed/_target/WaterQualityOrganisationDto.java
filiphr: Indirectly by not generating an error for it because it is not mapped. However, I think that you are right, we need to make sure that it is actually `null` in the tests as well.

gunnarmorling:
I've removed the milestone from this PR. Let's only assign milestones to issues, otherwise it's getting too confusing.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
filiphr: That is a bit strange, it seems to affect all `NestedPropertyMappingMethod`. I created #1082 for it.

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapper.java
sjaakd: Hmm that's not right.. It should indeed set the constant. However, if nothing is set, it should silently ignore in this case.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
filiphr: This method is generated by the `NestedPropertyMappingMethod` which are used for nested source properties. The names of this methods are created by joining all the element names from the `SourceReference`. Have a look at the [PropertyMapping#L487](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java#L487). The forged method has a source `waterQualityReportDto` and the mapping:
```java
@Mapping(target = "organisation.name", source = "organisationName")
```

Therefore the name is like it is. If you think that it needs to be different, I say we update #1082

sjaakd:
> I've removed the milestone from this PR. Let's only assign milestones to issues, otherwise it's getting too confusing.

And still it feels strange to assign issues to a milestone instead of solutions :smile:..

gunnarmorling:
> And still it feels strange to assign issues to a milestone instead of solutions 😄..

But that's what's usually done, no? E.g. in JIRA you also assign issues to a "fix version". People would like to know whether the issues they reported are part of a release or not. Also there can be several PRs for one issue, you don't want to have all of these in the milestone history. That all said, I feel issues and pull requests are not separated well enough in GitHub (shared id sequence etc.). I'd prefer them to be better distinguishable, but well :)

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
filiphr: The `private` methods are the methods generated by the `NestedPropertyMappingMethod`, in [`NestedPropertyMappingMethod.ftl#L21`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl#L21) we explicitly use `private`. The other methods are actually forged `BeanMappingMethod`(s) which use the accessibility from a `ForgedMethod` which actually is `protected`. If you want we can put this in #1082 as well.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
gunnarmorling: I think having the "To" in there would make the name better to grasp. But no blocker for sure.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
gunnarmorling: Do you know what we've started to use `protected` for forged bean mapping methods? I'd prefer them to be private as this gives us more flexibilty in changing things later (e.g. changing the name etc.). If they are protected, people could rely on them in sub-classes. The less we expose the better.

sjaakd:
> But that's what's usually done, no?

It was just a small joke from my side. No worries, I agree with your approach. But you are right: its nice that you can refer to an issue from a PR but it should have been separate flows that can be easily cross referenced.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
filiphr: The reason for this are their mappings:

```java
@Mapping(target = "fish.kind", source = "fish.type"),
@Mapping(target = "material.materialType", source = "material"),
```

The first mapping is symmetric, which means that we allow forged name based mapping between `fishTankDto.getFish()` and `fishTank.getFish()`.

The second mapping is assymetric, and therefore we send the entire source object in the mapping. However this mapping does not forge name based mappings, it is only restricted to the defined mappings.  The extracted mapping for this will look like:

```java
@Mapping(target = "materialType", source = "material")
```

gunnarmorling:
Ok, you got me there :)

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
filiphr: Well the accessibility comes from the `ForgedMethod` which uses `protected`. They have been introduced for forging collection methods (back in 2014). Theoretically, we could make sure that the forged methods that we generate here have `private` accessibility by adding that to the constructor, or exposing a setter for that. 

WDYT?

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
filiphr: Sorry I wrote the wrong mapping it should be:

```java
@Mapping(target = "organisationName", source = "organisation.name")
```

This really is not a `waterQualityDto` to `organisationName`. This is `organisationName = waterQualityDto.getOrganisation().getName()` with `null` checks.

gunnarmorling: Ah, understood. `getWaterQualityReportDtoOrganisationName()` would convey this a bit better. @sjaakd, WDYT?

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
gunnarmorling: Yes, the more private the better. @sjaakd do you remember why the original forged methods are protected?

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapper.java
filiphr: I updated the PR and added a fix for this. This example is now added to the `FishTankMapperConstant` and the fix is in the `NestedTargetPropertyMappingHolder#L483-491`

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
filiphr: Fixed

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -218,5 +508,123 @@ public NestedTargetPropertyMappingHolder build() {
...
+         */
+        private <K> void populateWithSingleTargetReferences(Map<K, List<Mapping>> map,
+            List<Mapping> singleTargetReferences, Extractor<SourceReference, K> keyExtractor) {
+            if ( singleTargetReferences != null ) {
+                //This are non nested target references only their property needs to be added as they most probably
+                // define it
+                for ( Mapping mapping : singleTargetReferences ) {
+                    if ( mapping.getSourceReference() != null && mapping.getSourceReference().isValid()
+                        && !mapping.getSourceReference().getPropertyEntries().isEmpty() ) {
+                        //TODO is this OK? Why there are no propertyEntries? For the Inverse LetterMapper for example
```
filiphr: @gunnarmorling, @sjaakd Did you had a look at this one as well? Is this an issue or not? Does it make sense that the source references property entries are empty?

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMapping.java
filiphr: @gunnarmorling you are right seprate issue is better I created #1083 for that

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
filiphr: I created #1084 to discuss this

filiphr:
@gunnarmorling thanks for the review. I have addressed your comments and update the PR (fixes are already squashed in the commits that they belong to 😄). Let me know if you think that there are too many commits. I made the commits in such way so we have some flow in what was done to solve this.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
sjaakd: totally agree. Dunno how difficult it is to handle. I think the same logic is re-used. I think that now this is handled in `AbastractMehthodBuilderBase#String getName(Type sourceType, Type targetType)`.. Perhaps we should add an enum and pass that along the call stack: e.g. MethodStyle (GETTER, MAPPER)..

filiphr:
I forgot to add. I rebased it on top of the current master and updated the fixtures accordingly

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
sjaakd: me neither.. just adapt and see if test cases start to fail?

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
sjaakd: ignore? constant? expression?

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
filiphr: @sjaakd it is pretty easy. I think that it is handled in [`PropertyMapping#L487`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java#L487). The `AbastractMehthodBuilderBase#String getName(Type sourceType, Type targetType)` is for name forged mappings. However this is nested source property and uses the `NestedPropertyMappingMethod`

sjaakd: ah.. then perhaps we should still address it here.. Anyway.. leave it up to you. I can imagine you want to close this one as well.

filiphr: `NestedPropertyMappingMethod` is something that we had from before. That is why I want to address it on a different place and wrap this one up 😄

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: When we have ignore, constant or expression the source reference is `null`.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/mixed/FishTankMapperImpl.java
sjaakd: ack.. I even added some extension to it :smile: .. Now.. time to do something healthy .. Going for a small run.

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
sjaakd: and when we only map a parameter (see my doc PR on #1057)?

filiphr: OK I created #1082 for that one.

filiphr:
It has been rebased with the latest master

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
```diff
@@ -151,11 +93,12 @@ public ForgedMethod(String name, Type sourceType, Type returnType, MapperConfigu
      *            parameters
      * @param history a parent forged method if this is a forged method within a forged method
      * @param mappingOptions the mapping options for this method
+     * @param forgedNameBased forges a name based (matched) mapping method
      */
-    private ForgedMethod(String name, Type sourceType, Type returnType, MapperConfiguration mapperConfiguration,
+    public ForgedMethod(String name, Type sourceType, Type returnType, MapperConfiguration mapperConfiguration,
```
agudian: Next time someone visits this class, it should be refactored to either use a builder or dedicated static creator methods with proper names... :)

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/mixed/AutomappingAndNestedTest.java
```diff
@@ -0,0 +1,281 @@
...
+    Interior.class,
+    Ornament.class,
+    WaterQuality.class,
+    WaterQualityReport.class,
+    FishTankWithNestedDocumentDto.class,
+    WaterQualityWithDocumentDto.class,
+    FishTankMapper.class,
+    FishTankMapperConstant.class,
+    FishTankMapperExpression.class,
+    FishTankMapperWithDocument.class
```
agudian: I guess at some point adding all classes of a package (ignoring *Test classes) would make things more readable, right? We should add that to our test framework...

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
filiphr: Good idea. I created #1088 for this

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/mixed/AutomappingAndNestedTest.java
filiphr: Good idea, I created #1089 for it

PR #1074, opened 2017-02-12
Wrap mapping options groupings

filiphr:
In order to clean up the #1072 PR we need to extract the generation of the groupings into a separate class. After this one is resolved, then we can more easily do the clean up

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -46,7 +46,6 @@ import org.mapstruct.ap.internal.model.dependency.GraphAnalyzer.GraphAnalyzerBui
 import org.mapstruct.ap.internal.model.source.ForgedMethod;
 import org.mapstruct.ap.internal.model.source.ForgedMethodHistory;
 import org.mapstruct.ap.internal.model.source.Mapping;
-import org.mapstruct.ap.internal.model.source.MappingOptions;
 import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.model.source.PropertyEntry;
 import org.mapstruct.ap.internal.model.source.SelectionParameters;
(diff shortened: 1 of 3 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -0,0 +1,212 @@
...
+
+import static org.mapstruct.ap.internal.util.Collections.first;
+
+/**
+ * This is a helper class that holds the generated {@link PropertyMapping}(s) and all the information associated with
+ * it for nested target properties.
+ *
+ * @author Filip Hrisafov
+ */
+public class NestedTargetPropertyMappingHolder {
```
sjaakd: Is it only nested target properties..? Or is it handling defined mappings in general?

filiphr: I made this explicitly for `handleDefinedNestedTargetMapping` so we can have this complex logic in one place

sjaakd:
Ok! Then go ahead and merge this one !

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -0,0 +1,212 @@
...
+
+        /**
+         * Splits the MappingOptions into possibly more MappingOptions based on each source method parameter type.
+         *
+         * Note: this method is used for forging nested update methods. For that purpose, the same method with all
+         * joined mappings should be generated. See also: NestedTargetPropertiesTest#shouldMapNestedComposedTarget
+         *
+         * @return the split mapping options.
+         */
+        public Map<Parameter, MappingOptions> groupBySourceParameter(List<Mapping> mappings) {
```
filiphr: You are right, I think that it would be better. Just to return `Map<Parameter, List<Mapping>>` and then we just create the `MappingOptions` before we create the `PropertyMapping`

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -298,38 +297,15 @@ else if ( reportErrorOnTargetObject( mapping ) ) {
 
         private void handleDefinedNestedTargetMapping(Set<String> handledTargets) {
 
-            Map<PropertyEntry, MappingOptions> optionsByNestedTarget =
-                method.getMappingOptions().groupByPoppedTargetReferences();
-            for ( Entry<PropertyEntry, MappingOptions> entryByTP : optionsByNestedTarget.entrySet() ) {
+            NestedTargetPropertyMappingHolder holder = new NestedTargetPropertyMappingHolder.Builder()
```
filiphr: I am not sure. I think that we need the "old" routine. The only thing is that we need to consider non-nested for creating the `PropertyMapping` in case we have just renaming like:

```java
@Mapping(target = "quality.document", source = "quality.report"),
@Mapping(target = "quality.document.organisation", source = "quality.report.org")
```

sjaakd: That's not what I meant.. We do need the "old" routine.. But what is less clear is that there are 2 paths to get to the old routine.. One direct (target is not nested) and one after popping via nested.  (`method.getMappingOptions().hasNestedTargetReferences()` considers all mappings)..

sjaakd: So.. sorting them into nested targets and non nested targets first thing would clarify that. Especially if you combine that with working on `Map<String,List<Mapping>`> iso `MappingOptions`.

sjaakd: and you can do something later on as well to check if sources are also nested (when a target is nested), to check for the symmetrical case.

PR #1077, opened 2017-02-16
General cleanups

filiphr:
These are some general cleanups that we found during #1072. 

@gunnarmorling do I need to make an issue and reference it?


gunnarmorling:
Looks good, thanks :)

Issue #1073, opened 2017-02-12
Duplicate NestedPropertyMappingMethod are created


filiphr:
Let's say that we have this mappings (They don't really make sense, but they convey my point):

```java
public interface FishTankMapper {

   @Mapping(target = "ornament", source = "interior.ornament")
   FishTankDto map(FishTank source);

   @Mapping(target = "ornament", source = "source.interior.ornament")
   FishTankDto map2(FishTank source);
}
```

The generated code looks like:

```java
public class FishTankMapperImpl implements FishTankMapper {

    @Override
    public FishTankDto map(FishTank source) {
       //null checks and other mappings
       FishTankDto fishTankDto = new FishTankDto();
       Ornament ornament = sourceInteriorOrnament( source );
       if ( ornament != null ) {
           fishTankDto.setOrnament( ornament );
       }

        return fishTankDto;
    }

    @Override
    public FishTankDto map2(FishTank source) {
        //null checks and other mappings
        FishTankDto fishTankDto = new FishTankDto();
        Ornament ornament = sourceInteriorOrnament1( source );
        if ( ornament != null ) {
            fishTankDto.setOrnament( ornament );
        }

        return fishTankDto;
    }
```
In the `map2` method we should reuse the `sourceInteriorOrnament` method instead of creating a new one.

gunnarmorling:
Done.

PR #1079, opened 2017-02-17
#1073 Change NestedPropertyMappingMethod to not use name in its equality

filiphr:
`NestedPropertyMappingMethod` was using the name in its `equals` method. This is not correct, because we always generate safe names, which leads to never reusing a `NestedPropertyMappingMethod`. 

With this PR a `NestedPropertyMappingMethod` is considered equals if it additionally has equal `SafePropertyEntry` entries.


sjaakd:
@filiphr : just one question.. For the remainder looks ok!

filiphr:
@sjaakd I can't reply on the comment. It looks like you have commented into my branch on the commit directly 😄. 

In any case we cannot use the same equals for the `sourceParameters`, because for those we only compare their `Type` and we don't use their `equals` methods

gunnarmorling:
Rebased and applied. Nice improvement, thanks, @filiphr!

PR #1081, opened 2017-02-19
#1057 documentation for Controlling forged name based bean mapping from root level with @Mapping

sjaakd:
Missing: I need to look what kind of formatting I should use if I use references to method signatures in-line.

Missing: example. Need to bring fishtank example to example repository as soon as we agree on #1057.


sjaakd:
And.. It needs to be rebased on the #1001 (#1069).

sjaakd:
Not my favorite job, so reviewers: feel free to add commits to make it better 👍

sjaakd:
> you can use intermediary methods.

and throw some intermediate iterables as well. :smile:

PR #1094, opened 2017-02-20
#1091 Map ANY_REMAINING Enum source to null

### LINKED ISSUE #1091: MappingConstants ANY_REMAINING and NULL not playing well together
I have an enum `FirstEnum` which looks like :

- START
- STOP
- REPORT

and an enum `SecondEnum` which looks like : 
- START
- STOP

I want to map the values of the `FirstEnum` to the `SecondEnum` and output `null` when the value of `FirstEnum` doesn't exist in `SecondEnum`. It should map the values as follow: 
- null -> null
- START -> START
- STOP -> STOP
- REPORT -> null

I tried : 

```
    @ValueMappings({
        @ValueMapping(source = MappingConstants.ANY_REMAINING, target = MappingConstants.NULL),
    })
    SecondEnum toSecondEnum(FirstEnum firstEnum);
```

It generates the code: 

```

    @Override
    public SecondEnum toSecondEnum(FirstEnum firstEnum) {
        if ( firstEnum == null ) {
            return null;
        }

        SecondEnum secondEnum;

        switch ( firstEnum ) {
            case START: secondEnum = SecondEnum.START;
            break;
            case STOP: secondEnum = SecondEnum.STOP;
            break;
            default: secondEnum = SecondEnum.<NULL>;
        }

        return secondEnum;
    }

```

As you can see, the generated code contains `default: secondEnum = SecondEnum.<NULL>` which doesn't compile.
I would have expect to output ` default: secondEnum = null;` instead

filiphr:
Fixes #1091. 

`ValueMapping` defined like:
```java
@ValueMapping(source = MappingConstants.ANY_REMAINING, target = MappingConstants.NULL)
```

Should create compilable code and should map all remaining values to `null`.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
@@ -105,7 +105,9 @@ public class ValueMappingMethod extends MappingMethod {
                     nullTarget = nullTargetValue.getTarget();
                 }
                 if ( defaultTargetValue != null ) {
-                    defaultTarget = defaultTargetValue.getTarget();
+                    // If the default target value is NULL then we should map it to null
+                    defaultTarget = MappingConstantsPrism.NULL.equals( defaultTargetValue.getTarget() ) ? null :
+                        defaultTargetValue.getTarget();
                 }
                 else {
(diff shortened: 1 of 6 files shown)


filiphr:
Thanks for the review @gunnarmorling

PR #1095, opened 2017-02-21
#1059 Add ftlvariable to the ftl files so we can have autocomplete with IntelliJ

### LINKED ISSUE #1059: Add ftlvariable to the ftl files so we can have autocomplete with IntelliJ
In order for IntelliJ to link the Java classes and the ftl files one needs to add an `ftlvariable` comment to the ftl file. 

It looks like this
```
<#-- @ftlvariable name="" type="org.mapstruct.ap.internal.model.common.Type" -->
```

Another option is to use an external file called `freemarker_implicit.ftl` and have something like this
```
[#-- @ftlvariable name="" type="org.mapstruct.ap.internal.model.common.Type" file="org/mapstruct/ap/internal/model/common/Type.ftl" --]
```

I am more for the first option as you immediately see to what it is linked.

If you don't like it then I won't do it 😄.

filiphr:
Fixes #1059.

Adds better support for autocomplete in ftl files in IntelliJ


gunnarmorling:
Merged, thanks, @filiphr!

PR #1096, opened 2017-02-21
Fix checkstyle error

filiphr:
It is a bit strange how there was no error in #1094. This fixes the build


gunnarmorling:
Odd indeed.

filiphr:
It was still failing :). I'll make another one

Issue #1105, opened 2017-02-26
Fixing some template issues


sjaakd:
I was continuing my refactor activities around #904 (adapted text). Before I'm going to dig in further into this I spotted a problem between two templates.

~~There was an extra enter in `SourceRHS` which was somehow fixed in `MethodReference` by declaring the whole `MethodReference` as `<@compress single_line=true>` (including surrounding its macro's). This needs to be fixed.~~

Just refactoring  `MethodReference`, remove the strange `<@compress single_line=true>` including surrounding its macro's.

sjaakd:
done and closed

PR #1106, opened 2017-02-26
#1105 cleaning up MethodReference freemarker template

sjaakd:


sjaakd:
> the formatting was not OK

@filiphr .. the formatting was not the problem (in the end).  Its just a first small step. I want to have the current free marker templates ok before I (for instance) move them.

filiphr:
@sjaakd, sorry I understood it wrong. If you are planning to change something in them I would say add a fixture. However, if it is only moving then maybe we don't need a fixture.

sjaakd:
No problem.. I'll go ahead and merge this one..

Issue #1103, opened 2017-02-26
1.2.0-Beta1 Automapping Feature: cycle tree let the compiler crash with StackOverflowError.


brabenetz:
Recursive Mappings can produce a compiler crash "Internal error in the mapping processor: java.lang.StackOverflowError":
[MapStructRecursionMapper.java.txt](https://github.com/mapstruct/mapstruct/files/801850/MapStructRecursionMapper.java.txt)

navpil:
@brabenetz Thanks for finding this bug.

I have added the PR which fixes this particular issue. Probably some other recursive cases should be handled.

agudian:
Merged the fix by @navpil. Thanks @brabenetz for reporting the issue!

PR #1107, opened 2017-02-27
#1103 

navpil:
Try to fix the issue #1103 "1.2.0-Beta1 Automapping Feature: cycle tree let the compiler crash with StackOverflowError."

Initially when there was a non-recursive implementation there was a HashSet which excluded already created Forged Mappings. With the recursive implementation I see no way except using some global state or passing the set of created forged methods as a parameter.
I decided to use the first solution. This fix is added only for the usual property mappings, I did not check how would the recursive iterable mappings behave.

Is this a fine approach?

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
@@ -57,6 +57,14 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
      * @return See above
      */
     Assignment createForgedAssignment(SourceRHS sourceRHS, ForgedMethod forgedMethod) {
+
+        if ( ctx.getForgedMethodsUnderCreation().containsKey( forgedMethod ) ) {
+            return createAssignment( sourceRHS, ctx.getForgedMethodsUnderCreation().get( forgedMethod ) );
+        }
+        else {
+            ctx.getForgedMethodsUnderCreation().put( forgedMethod, forgedMethod );
+        }
+
         MappingMethod forgedMappingMethod;
         if ( MappingMethodUtils.isEnumMapping( forgedMethod ) ) {
             forgedMappingMethod = new ValueMappingMethod.Builder()
@@ -72,7 +80,9 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
                 .build();
         }
 
-        return createForgedAssignment( sourceRHS, forgedMethod, forgedMappingMethod );
+        Assignment forgedAssignment = createForgedAssignment( sourceRHS, forgedMethod, forgedMappingMethod );
+        ctx.getForgedMethodsUnderCreation().remove( forgedMethod );
+        return forgedAssignment;
     }
 
     Assignment createForgedAssignment(SourceRHS source, ForgedMethod methodRef, MappingMethod mappingMethod) {
@@ -87,9 +97,14 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
             methodRef = new ForgedMethod( existingName, methodRef );
         }
 
+        return createAssignment( source, methodRef );
+    }
+
+    private Assignment createAssignment(SourceRHS source, ForgedMethod methodRef) {
         Assignment assignment = MethodReference.forForgedMethod(
             methodRef,
-            ParameterBinding.fromParameters( methodRef.getParameters() ) );
+            ParameterBinding.fromParameters( methodRef.getParameters() )
+        );
         assignment.setAssignment( source );
 
         return assignment;
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
@@ -96,24 +96,27 @@ public class Parameter extends ModelElement {
 
     @Override
     public int hashCode() {
-        int hash = 5;
-        hash = 23 * hash + (this.name != null ? this.name.hashCode() : 0);
-        return hash;
+        int result = name != null ? name.hashCode() : 0;
+        result = 31 * result + ( type != null ? type.hashCode() : 0 );
+        return result;
     }
 
     @Override
-    public boolean equals(Object obj) {
-        if ( obj == null ) {
+    public boolean equals(Object o) {
+        if ( this == o ) {
+            return true;
+        }
+        if ( o == null || getClass() != o.getClass() ) {
             return false;
         }
-        if ( getClass() != obj.getClass() ) {
+
+        Parameter parameter = (Parameter) o;
+
+        if ( name != null ? !name.equals( parameter.name ) : parameter.name != null ) {
             return false;
         }
-        final Parameter other = (Parameter) obj;
-        if ( (this.name == null) ? (other.name != null) : !this.name.equals( other.name ) ) {
-            return false;
-        }
-        return true;
+        return type != null ? type.equals( parameter.type ) : parameter.type == null;
+
     }
 
     public static Parameter forElementAndType(VariableElement element, Type parameterType) {
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
@@ -87,6 +87,7 @@ public enum Message {
     GENERAL_UNSUPPORTED_DATE_FORMAT_CHECK( "No dateFormat check is supported for types %s, %s" ),
     GENERAL_VALID_DATE( "Given date format \"%s\" is valid.", Diagnostic.Kind.NOTE ),
     GENERAL_INVALID_DATE( "Given date format \"%s\" is invalid. Message: \"%s\"." ),
+    GENERAL_NOT_ALL_FORGED_CREATED( "Internal Error in creation of Forged Methods, it was expected all Forged Methods to finished with creation, but %s did not" ),
 
     RETRIEVAL_NO_INPUT_ARGS( "Can't generate mapping method with no input arguments." ),
     RETRIEVAL_DUPLICATE_MAPPING_TARGETS( "Can't generate mapping method with more than one @MappingTarget parameter." ),
--- a/processor/src/test/java/org/mapstruct/ap/test/nestedbeans/RecursionTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/nestedbeans/RecursionTest.java
@@ -1,0 +1,131 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.nestedbeans;
+
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mapstruct.ap.test.nestedbeans.recursive.RecursionMapper;
+import org.mapstruct.ap.test.nestedbeans.recursive.TreeRecursionMapper;
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+@RunWith(AnnotationProcessorTestRunner.class)
+public class RecursionTest {
+
+    @WithClasses({
+        RecursionMapper.class
+    })
+    @Test
+    @IssueKey("1103")
+    public void testRecursiveAutoMap() {
+        RecursionMapper.RootDto rootDto = new RecursionMapper.RootDto(
+            new RecursionMapper.ChildDto( "Sub Root", new RecursionMapper.ChildDto(
+                "Sub child", null
+            ) )
+        );
+        RecursionMapper.Root root = RecursionMapper.INSTANCE.mapRoot( rootDto );
+
+        assertRootEquals( rootDto, root );
+    }
+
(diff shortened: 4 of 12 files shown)


gunnarmorling:
ok to test

agudian:
I vaguely remember that we check the available methods if they already match the soure/target type before creating a forged method, but in the near past added the mapping options to the hashCode/equals for forged methods. Is that maybe a spot where we can fix this as well? @filiphr?

agudian:
Btw, @navpil, huge thanks for keeping in touch with your feature 👍

filiphr:
@navpil You are right we changed a bit how the `ForgedMethods` are created. I also started working on something yesterday evening. It is a similar approach to yours. Have a look at my https://github.com/filiphr/mapstruct/commit/ebc1920eb87dc4b6c3e0bbd2d0ded9a2cede11dc. I am using a map, because the name of a `ForgedMethod` is not used in the `equals`/`hashCode` methods and we always generate a safe name for the methods. I think that with a set we could theoretically create assignments that don't match the created methods.

My fix spot is also at another place, let me know what you think.

navpil:
@filiphr Actually the name of the ForgedMethod is used in the hashCode and equals. Anyway - map keys and sets work the same (HashSet is actually a HashMap with no values) so our approaches are quite the same. The spot is almost the same too - only a method call apart. 
The only real difference I see is that you remove the ForgedMethod from the UnderCreation collection and I don't. I don't know which one is better. What will happen if in some other place, not in the recursion path, we decide to generate another forged method with the same parameters? Should we know that such method was already created, can it cause problems if we don't create another forged method, or is it irrelevant?

filiphr:
@navpil You are right the `ForgedMethod` does use the name in the `equals`/`hashCode` methods. However, I think that it shouldn't. We removed the name from those methods for some other `Mapping` methods. I used a `Map` only because I thought that the `name` was not used in the `equals`/`hashCode` methods.

I did it in that spot because there are other places apart from the `PropertyMapping` that create forged mappings.


>What will happen if in some other place, not in the recursion path, we decide to generate another forged method with the same parameters? 

If we generate a `ForgedMethod` with the same parameters then we check the created `MappingMethod` in `createForgedAssignment()` and if one already exists (`name` is not taken into consideration) then we just change the `ForgedMethod` that we are going to use in the `assignment`.

>Should we know that such method was already created, can it cause problems if we don't create another forged method, or is it irrelevant?

I don't think that it matters whether we create a different `ForgedMethod` or not. 

Btw. I decided to remove it so I can check in the `MapperCreationProcessor` if all `ForgedMethod`(s) finished with creation successfully (I am not sure if this is needed, I just thought as a safeguard that will notify us if something is not OK).

sjaakd:
> @navpil You are right the ForgedMethod does use the name in the equals/hashCode methods. However, I think that it shouldn't. We removed the name from those methods for some other Mapping methods. I used a Map only because I thought that the name was not used in the equals/hashCode methods.

The reason here: we want MapStruct to be able to reuse prior generated methods. That is, their signature is the same and their content (property mappings, incl. their order) is the same. The name not matter. Note that generating the same method with different names can lead to problems (conflicting methods in the selection process).

navpil:
@filiphr So it looks like your solution is closer to what we need than mine. Putting the check into the AbstractBaseBuilder and clearing the UnderCreation collection do make sense. 
Next the 'name' should be removed from the ForgedMethod equals/hashCode and I'd really suggest to use the Set instead of Map (Map is kind of confusing, since we don't use values). 
Should I adjust my PR to reflect your changes, or we can just replace it with yours?
What about other cases? I will look into the case where user has a list of users as a property. Other suggestions?

navpil:
I checked that with with @filiphr solution there will be no problem with mapping tree like objects (when user has a list of users himself).

filiphr:
@navpil feel free to adjust the PR with the changes. We don't have to replace it with mine, use what you need from there.

As for the `Set` vs `Map`. If we remove the `name` from the `equals` and `hashCode` checks, then we must use `Map`. In my "solution" the value of the `underCreation` map is used to create an assignment. If we only use `Set` then you can theoretically create an `assignment` that won't compile.

For example:
* You have `ForgedMethod` with name `carToCarDto`, you put this method into the `Map` with itself as key and value. 
* In the next step you have a `ForgedMethod` with name `carToCarDto1`. 

Both methods are equal and have the same hashCode. If you use a `Set` then you will use the name of the second method `carToCarDto1`, but you are going to create a mapping method for `carToCarDto` (the first method is used to create the mapping). If on the other side we use a `Map`, then when you use the second method to get a value you will get back the first `ForgedMethod` and you are going to create an assignment with the name `carToCarDto`.

WDYT?

navpil:
@filiphr Finally I got it. Sorry, I did not notice you actually use the map values, so that's why I was confused about map vs set difference. As for PR - I will update it to reflect your changes.

navpil:
I had to change how the `equals/hashCode` is being generated in the `Parameter` class. It took only name into account which caused `ForgedMethod` equality be wrong. So now it takes name and type into account. I was not sure whether name should have been removed or not, so I left it there.
This change broke one test, in particular `NestedTargetPropertiesTest`. But the reason was not that the generated mapper was wrong, but that the method names were not what as expected. I guess the reason was that the `hashCode` change changed the way objects are stored in the set buckets and were iterated through in the other order than before. So created mapper was different from the reference mapper `fixtures\org\mapstruct\ap\test\nestedtargetproperties\ChartEntryToArtistImpl.java` in particular - methods `chartEntryTopSong1` and `chartEntryToSong2` were swapped.
You can see I commented out the Rule in the test and the test itself passed.
What should be done with that?

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -84,6 +84,7 @@
     GENERAL_UNSUPPORTED_DATE_FORMAT_CHECK( "No dateFormat check is supported for types %s, %s" ),
     GENERAL_VALID_DATE( "Given date format \"%s\" is valid.", Diagnostic.Kind.NOTE ),
     GENERAL_INVALID_DATE( "Given date format \"%s\" is invalid. Message: \"%s\"." ),
+    GENERAL_NOT_ALL_FORGED_CREATED( "Internal Error in e creation of Forged Methods, it was expected all Forged Methods to finished with creation, but %s did not" ),
```
agudian: `Error in e creation` --> I guess the " e " can be removed...

processor/src/test/java/org/mapstruct/ap/test/nestedtargetproperties/NestedTargetPropertiesTest.java
```diff
@@ -51,11 +50,11 @@
...
-    @Rule
-    public GeneratedSource generatedSource = new GeneratedSource().addComparisonToFixtureFor(
-        ChartEntryToArtist.class,
-        ChartEntryToArtistUpdate.class
-    );
+//    @Rule
+//    public GeneratedSource generatedSource = new GeneratedSource().addComparisonToFixtureFor(
+//        ChartEntryToArtist.class,
+//        ChartEntryToArtistUpdate.class
+//    );
```
agudian: Hmm, that's not good. Can you find out where we're iterating over a map or set that doesn't contain any ordering? Typically, we use `LinkedHashMap` or somehting similar that preserves the `put` order when iterating over it.

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
```diff
@@ -56,12 +56,23 @@ public B method(Method sourceMethod) {
...
     Assignment createForgedBeanAssignment(SourceRHS sourceRHS, ForgedMethod forgedMethod) {
+
+        if ( ctx.getForgedMethodsUnderCreation().containsKey( forgedMethod ) ) {
+            return createAssignment( sourceRHS, ctx.getForgedMethodsUnderCreation().get( forgedMethod ) );
+        }
+        else {
+            ctx.getForgedMethodsUnderCreation().put( forgedMethod, forgedMethod );
+        }
+
+
```
agudian: One empyt line is sufficient... 😁

processor/src/test/java/org/mapstruct/ap/test/nestedtargetproperties/NestedTargetPropertiesTest.java
agudian: @navpil now that I've read your comment above a bit more careful: if you can make sure that the order remains "stable" in a sense that it's not depending on any HashMap/HashSet iteration, then it's totally okay to adapt the reference impl file.
But if this is a result of HashMap/HashSet iteration (which can vary based on the JDK version, among other things), then we should somehow fix it.

filiphr: @navpil the problem are the `HashMap`(s) used in the `NestedTargetPropertyMappingHolder`. If you change them to `LinkedHashMap`(s) then all is good.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -18,17 +18,11 @@
  */
 package org.mapstruct.ap.internal.model;
 
-import static org.mapstruct.ap.internal.model.assignment.Assignment.AssignmentType.DIRECT;
```
filiphr: Apart from the imports and one empty line there are no changes to this file. Can you please revert all changes to this one?

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
```diff
@@ -95,25 +95,28 @@ public boolean isMappingContext() {
...
+        if ( name != null ? !name.equals( parameter.name ) : parameter.name != null ) {
             return false;
         }
-        return true;
+        return type != null ? type.equals( parameter.type ) : parameter.type == null;
+
+    }
+
+    @Override
+    public int hashCode() {
```
filiphr: Why change the order of the `equals` / `hashCode`? Why not just add the needed lines to the methods?

processor/src/main/java/org/mapstruct/ap/internal/model/source/FormattingParameters.java
```diff
@@ -40,4 +40,28 @@ public String getNumber() {
         return number;
     }
 
+    @Override
+    public boolean equals(Object o) {
```
filiphr: Why are these needed for this class? Where are they used?

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -100,8 +101,8 @@
     RETRIEVAL_TYPE_VAR_RESULT( "Can't generate mapping method for a generic type variable target." ),
     RETRIEVAL_WILDCARD_SUPER_BOUND_SOURCE( "Can't generate mapping method for a wildcard super bound source." ),
     RETRIEVAL_WILDCARD_EXTENDS_BOUND_RESULT( "Can't generate mapping method for a wildcard extends bound result." ),
-    RETRIEVAL_CONTEXT_PARAMS_WITH_SAME_TYPE( "The types of @Context parameters must be unique." ),
```
filiphr: Why are there line changes for this one and `INHERITCONFIGURATION_CYCLE`?

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/RecursionTest.java
```diff
@@ -0,0 +1,88 @@
...
+    @IssueKey("1103")
+    public void testRecursiveAutoMap() {
+        RecursionMapper.Root root = RecursionMapper.INSTANCE.mapRoot(
+            new RecursionMapper.RootDto(
+                new RecursionMapper.ChildDto( "Sub Root", new RecursionMapper.ChildDto(
+                    "Sub child", null
+                ) )
+            ) );
+
+        Assert.assertEquals(
```
filiphr: Can you use object accessors instead of the equals method for the checks? You can have a look at the latest `NestedSimpleBeansMappingTest`.

Then you won't need the `equals` and `hashCode` methods in the objects and the class will be smaller :)

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/recursive/RecursionMapper.java
```diff
@@ -0,0 +1,174 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.nestedbeans.recursive;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.ReportingPolicy;
+import org.mapstruct.factory.Mappers;
+
+@Mapper(unmappedTargetPolicy = ReportingPolicy.IGNORE)
```
filiphr: I think that the `unmappedTargetPolicy` is not needed.

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/recursive/TreeRecursionMapper.java
```diff
@@ -0,0 +1,191 @@
...
+ */
+package org.mapstruct.ap.test.nestedbeans.recursive;
+
+import java.util.List;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.ReportingPolicy;
+import org.mapstruct.factory.Mappers;
+
+@Mapper(unmappedTargetPolicy = ReportingPolicy.IGNORE)
```
filiphr: Same here. I am don't think the target policy is needed.

processor/src/main/java/org/mapstruct/ap/internal/model/source/FormattingParameters.java
navpil: True to say I blindly copypasted this from the changeset you linked earlier. I will remove it and check that all tests still pass.

filiphr: No problem. I didn't realize I have committed that as well, I was in a hurry to commit and push then and might have pushed something not needed :)

filiphr:
@navpil will you have time to finish this PR? We want to release Beta2 and this is the last issue that is planned for it. If you don't have the time, let us know, so we can finish it.

agudian:
@navpil do you have some time rebase and work in the remaining comments? We're planning to do the Beta2 release tonight, but I'd really like to see this fix included.
If you already know that you can't do it in time, one of us can take over, but it's usually nicer for everyone if we don't steal your credit 😉. Just let us know... 👍

navpil:
I will do this today

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
navpil: I usually delete equals/hashCode and generate them from the IntelliJ template instead of adding needed lines, I did not notice the methods were swapped.

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/RecursionTest.java
navpil: I will do how you ask, though as I think of it - the best solution here would be map dto to entity and back (we assume they have the same number of properties) and then compare them by the equals method. Something like
`RootDto rootDto = createTestRootDto();
assertEquals(rootDto, mapper.mapEntityToDto( mapper.mapDtoToEntity( rootDto ) );`
This way there is no need to write own equality methods (possibly error prone) and no need to create an additional object of different type, but same parameters.

agudian:
Awesome, @navpil, thanks a lot! Release is coming up now... 👍

Issue #1111, opened 2017-03-01
List<List<Bean>> generate method error


johnwatsondev:
#### ENV
MacOS 10.12.3

Android Studio 2.2.3
Build #AI-145.3537739, built on December 2, 2016
JRE: 1.8.0_112-release-b05 x86_64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o

lombokVersion = '1.16.14'
mapstructVersion = '1.2.0.Beta1'

#### Mapper class
```
@Mapper public interface MaintainSetAndDetailEntity2MaintainSetAndDetailBeanMapper
    extends MaintainItemInfoEntity2MaintainItemInfoBeanMapper {
  MaintainSetAndDetailEntity2MaintainSetAndDetailBeanMapper MAPPER =
      Mappers.getMapper(MaintainSetAndDetailEntity2MaintainSetAndDetailBeanMapper.class);

  List<MaintainItemInfoBean> maintainItemInfoEntityListCast(
      List<MaintainItemInfoEntity> maintainItemInfoEntityList);

  List<List<MaintainItemInfoBean>> listList(
      List<List<MaintainItemInfoEntity>> maintainItemInfoEntityListList);
}
```

#### Generated class
```
  @Override

  public List<List<MaintainItemInfoBean>> listList(
      List<List<MaintainItemInfoEntity>> maintainItemInfoEntityListList) {

    if (maintainItemInfoEntityListList == null) {

      return null;
    }

    List<List<MaintainItemInfoBean>> list = new ArrayList<List<MaintainItemInfoBean>>();

    for (List<MaintainItemInfoEntity> list : maintainItemInfoEntityListList) {

      list.add(maintainItemInfoEntityListCast(list));
    }

    return list;
  }
```

#### Solution
Please change the local variable name `list` to `some_prefix_list` in for-loop criteria.

Thanks for hard work.

gunnarmorling:
@sjaakd We could even consider to forge the middle (List -> List) method. Not sure though whether it's worth the efforts. Probably better to avoid the nested collections and model it using proper classes.

sjaakd:
> We could even consider to forge the middle (List -> List) method.

That doesn't work.. I think forging only works when a bean mapping method refers to an iterable. Here an iterable requires another iterable mapping. That's not implemented. Perhaps an issue?

gunnarmorling:
Yes, that's what I meant, should we support it? We can open an issue, but
[...]

PR #1112, opened 2017-03-01
#1111 local variabele should not clash with loop variable

sjaakd:


agudian:
Looks good, Sjaak!

You could also move `Source` and `Target` into the test-mapper as inner classes, making the simple test a bit more concise (and you won't need to list them in `@WithClasses` in that case). And maybe add a short check that the list of lists is mapped correctly, as we didn't have that anywhere else, yet?

Anyway, I leave that at your own discretion - feel free to merge 👍

sjaakd:
@agudian thanks for reviewing. I adapted the test case... I'll merge tomorrow when the CI build succeeds...

Issue #1113, opened 2017-03-01
Explore AST-based approach for code generation


gunnarmorling:
Hi all, creating this issue to share some experiment I've done with you and get a discussion on this started.

So my impression is that the template based approach for code generation served us well in the beginning but when looking at the templates of the "assignment" model, we clearly hit a limit where it's getting very hard to grasp what the behavior is and how the generated code will look like.

Hence I've spent some time looking into an AST-based alternative. The idea is to assemble an AST (abstract syntax tree) representing the code to be generated. This tree can then easily be written out. A nice side effect is that we get more control over the formatting (pretty printing) as that happens using a (customizable) visitor of the AST. On each level, the AST nodes are examinable in the debugger, showing that sub-part of the tree.

I'm going to send a PR which shows how this could look like (just adapted a very small sub-set of the model elements).

PR #1114, opened 2017-03-01
DON'T MERGE: #1113 Small example for AST-based code generation

gunnarmorling:
This uses the [JavaParser](https://github.com/javaparser/javaparser/) project for assembling an AST of the code to be generated. The build fails and only a few tests pass. Take a look at `OnewayTest` for one that works.

Essentially, each `ModelElement` has a `getAst()` method which returns the AST of this sub-tree (calling `getAst()` of any sub-nodes). I think it works well for the inner, more logic-intensive model elements (`SetterWrapper` and friends). For the higher-level nodes (e.g. `BeanMappingMethod`) it's not so clear, the templates may be more easily comprehensible here. But one could think of actually _parsing_ templates here (not FreeMarker templates necessarily) instead of assembling these ASTs from scratch and then only replace those nodes which need to be customized.

Hope it makes sense. Looking forward to your thoughts. I think it'd be a great win for the entire `Assignment` hierarchy, I'm feeling we're hitting a dead end with templates here.

### DIFF
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -44,6 +44,12 @@
             <artifactId>freemarker</artifactId>
         </dependency>
 
+<dependency>
+    <groupId>com.github.javaparser</groupId>
+    <artifactId>javaparser-core</artifactId>
+    <version>3.1.0</version>
+</dependency>
+
         <!-- Compile-only; Using "provided" scope as there is no such scope in Maven;
         these dependencies are not required at runtime, only for prism generation
         and tests -->
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/Annotation.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/Annotation.java
@@ -25,6 +25,10 @@ import java.util.Set;
 import org.mapstruct.ap.internal.model.common.ModelElement;
 import org.mapstruct.ap.internal.model.common.Type;
 
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.expr.Name;
+import com.github.javaparser.ast.expr.NormalAnnotationExpr;
+
 /**
  * Represents a Java 5 annotation.
  *
@@ -59,5 +63,12 @@ public class Annotation extends ModelElement {
 
     public List<String> getProperties() {
         return properties;
+    }
+
+    @Override
+    public Node getAst(Context context) {
+        NormalAnnotationExpr node = new NormalAnnotationExpr();
+        node.setName( new Name( type.isImported() ? type.getName() : type.getFullyQualifiedName() ) );
+        return node;
     }
 }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -24,6 +24,7 @@ import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -62,6 +63,24 @@ import org.mapstruct.ap.internal.util.Message;
 import org.mapstruct.ap.internal.util.Strings;
 import org.mapstruct.ap.internal.util.accessor.Accessor;
 import org.mapstruct.ap.internal.util.accessor.ExecutableElementAccessor;
+
+import com.github.javaparser.ast.Modifier;
+import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.ast.body.MethodDeclaration;
+import com.github.javaparser.ast.body.VariableDeclarator;
+import com.github.javaparser.ast.expr.BinaryExpr;
+import com.github.javaparser.ast.expr.BinaryExpr.Operator;
+import com.github.javaparser.ast.expr.Expression;
+import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.expr.NullLiteralExpr;
+import com.github.javaparser.ast.expr.ObjectCreationExpr;
+import com.github.javaparser.ast.expr.VariableDeclarationExpr;
+import com.github.javaparser.ast.stmt.BlockStmt;
+import com.github.javaparser.ast.stmt.ExpressionStmt;
+import com.github.javaparser.ast.stmt.IfStmt;
+import com.github.javaparser.ast.stmt.ReturnStmt;
+import com.github.javaparser.ast.stmt.Statement;
+import com.github.javaparser.ast.type.ClassOrInterfaceType;
 
 /**
  * A {@link MappingMethod} implemented by a {@link Mapper} class which maps one bean type to another, optionally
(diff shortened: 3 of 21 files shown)


sjaakd:
Hey @gunnarmorling .. To be honest.. I'm not sure. I'm still refactoring the wrappers. I don't think we should have wrappers (trying to get lost of them soon, but was occupied with this auto mapping). There should only be a RHS / LHS combination. That should make the whole property assignment a large deal simpler even. Note, the wrapping is in fact already gone on the targetLHS (wrapper) side. Combine that with a clear selection mechanism (instead of the if-then-else constructions now... Also passing an assignment obfuscates stuff. If each TargetLHS has just one backing bean... 

The way I see it is that the current problem is in the rather large templates of bean mapping, iterable mapping and map mapping. They should be high on the refactoring list to make them readable again. 

The problem trading templates for java code is that you can not see directly what's generated from the java code. Templates are better in that respect. For myself, I can relate the generated code easier with the template (since it still looks more or less the same).

gunnarmorling:
>  That should make the whole property assignment a large deal simpler even.

That would be great. Currently I'm having a hard time to understand what's going on there (admittedly I didn't work on this much recently, so that may explain a bit ;).

 >The problem trading templates for java code is that you can not see directly what's generated from the java code. Templates are better in that respect. For myself, I can relate the generated code easier with the template

It's interesting that you say that. That's why I started with templates originally, but I don't find it to be true any longer for the assignments. Due to their highly normalized nature (extracted re-used parts, deep nesting of template hierarchy) I often have no idea how the generated code will look like. I find debugging got hard due to that. With the AST approach you can see its current state at any time in the debugger (`Node#toString()` emits the represented Java code).

But I can relate to your concern about the outcome not being foreseeable when just looking at the Java code of complex AST operations. That's what I meant to mitigate when talking about parsing "templates" as the starting ground for AST manipulations. E.g. in `BeanMappingMethod` we could have this (strongly simplified):

    class BeanMappingMethod {

        private final static String TEMPLATE =
            "public Object name(Object source) {" +
            "    if ( source == null ) {" +
            "         return null;" +
            "    }" +
            "    Object result = new Object();" +
            "    Object mappingsPlaceHolder = null;" +
            "    return result;" +
            "}";

        MethodDeclaration getAst() {
            MethodDeclarationAst ast = parser.parseMethod( TEMPLATE );
            // now alter that template AST by setting name, replace parameter type(s), insert mappings etc.
            return ast;
        }
    }

I'd envision this gives you the best of both: understanding of the structure by looking at `TEMPLATE` in the source code as well as understanding during debugging.

Of course one wouldn't always start with such template, for the assignments/property mappings themselves it may make more sense to go programmatically all the way. But for the higher level nodes (mapping methods, mappers) one would do it by amending such AST loaded from a template. That's what JavaParser folks recommend, too.

sjaakd:
> deep nesting of template hierarchy

That is what I want to get rid of as well. Like said, I want to have 2 components for a property. A SourceRHS (containing only the the source side of things (2 step, inner/outer method / type conversion / direct assignment) and the stuff what we now call assignment wrapper. I don't want to hand objects as an assignment between those.

> I'd envision this gives you the best of both: understanding of the structure by looking at TEMPLATE in the source code as well as understanding during debugging.

Great 👍 Like that.

--------
Having said that.. What if we tackle it as follows:
1. I'd like to finish the improvements that I'm talking about (coming iterations, preferable before the 1.2.final. But I expect it to be several weeks work still.

2. Having this simplified (better understandable) structure, use that as a base for exploring replacement your style. In that way we have a clear starting point. Perhaps we can even do it one template after another. I even think we don't have to go to 2.0 to do that.

WDYT?

agudian:
One thing we also need to consider when waging the different options is the extensibility... in case we can't or don't want to make the model at various levels mutable or decoratable, we could instead or additionally allow modifying the constructed AST...

filiphr:
@gunnarmorling this looks interesting. I think that the idea of using templates is interesting. Having the option to do `Node#toString()` in the debugger certainly helps :).

Did you perhaps had a look at [JavaPoet](https://github.com/square/javapoet)? It looks like that it is not as powerfull as JavaParser and does not provide an AST. However, by looking at the manual it looks interesting and the programmatic creation looks readable.

filiphr:
btw I forgot to write. Did you notice that the latest version of JavaParser requires Java 8?

gunnarmorling:
Thanks for your feedback, everyone!

@sjaakd:

> I'd like to finish the improvements that I'm talking about (coming iterations, preferable before the 1.2.final. But I expect it to be several weeks work still.
> Having this simplified (better understandable) structure, use that as a base for exploring replacement your style.

+1, that was my thinking as well. Your current work seems to be a good basis for the AST change once you're done.

> In that way we have a clear starting point.

Right. That'd help a lot because currently it's tough for me to grasp the semantics of the individual assignment nodes. So how will it look like after your work? Will there be just two elements left, one for the source and one for the target of a property propagation?

> Perhaps we can even do it one template after another.

Yes, that works if one starts from the bottom up. I.e. lower nodes (e.g. `PropertyMapping`) can be migrated to the AST approach, whereas higher nodes (e.g. `Mapper`) can still be template based. That's what I did for this small PoC (only that I did it all the way up, supporting only a subset of the functionality for each node).

> I even think we don't have to go to 2.0 to do that.

+1

@agudian:

> we could instead or additionally allow modifying the constructed AST...

Right, I actually wanted to mention that but forgot :) Indeed we could expose the AST as the basis for custom amendments by users. It'd allow them to do whatever they want. I'm only feeling a bit uncomfortable about exposing types of a dependency (JavaParser) to our users. But then it doesn't really make sense to provide our own abstraction which would be the same very much. Something to think about :)

@filiphr:

> Did you perhaps had a look at JavaPoet?

Yes, I did. But I think it falls short where we'd benefit most from a proper programmatic approach: actual method implementations. That's not modeled by JavaPoet, so we'd be back to String manipulation AFAIU. So it didn't seem useful for our case, where the most complex part is in the generation of method implementations.

>  Did you notice that the latest version of JavaParser requires Java 8?

That's true. We would either have to mandate Java 8 (I wouldn't mind, but I think Sjaak is still using MapStruct with earlier versions?) or we'd have to use a previous version of JavaParser.

sjaakd:
> Will there be just two elements left, one for the source and one for the target of a property propagation?

Yes.. That is the idea. :smile:

* One `SourceRHS`, reusable in all spots were we have assignments (including in the `Iterable`, and `MapMapping` methods)

* Several `TargetLHS`'es. Specialized along the same lines now. However, they would all implement the same basic `TargetLHS` interface.

* It is the pair (`XyzTargetLHS` - `SourceRHS`) that makes up a property mapping. So a TargetLHS does no longer wrap the `SourceRHS`.

* All the checks, currently in the property mapping leading to an `XyzTargetLHS` would be part of the `XyzTargetLHS`. So each `XyzTargetLHS` becomes responsible for its own selection. This could for instance be done by giving each `XyzTargetLHS` its own standard factory (as part of the `TargetLHS` interface. So something like `TargetLHS.factory` and add those `factories` to a chain of responsibility. Once a factory is selected, it returns a new  `XyzTargetLHS` and its added together with `SourceRHS` in the `PropertyMapping`. Further selection stops.

* This chain of responsibility could eventually  become part of the SPI. A user could place its own `TargetLHS` (factory) in the chain and reorder the priority.

* Don't propagate the Assignment via a template model

sjaakd:
> but I think Sjaak is still using MapStruct with earlier versions?

No.. I'm not.. We moved to java 8 last summer. But if we need to go to Java 8 than we probably should introduce this in 2.0. But remember, you guys all seem to be on the cutting  (sometimes bleeding) edge of Java. A lot of users aren't. In marketing terms: we could use quite some market segment going to java 8 only.

gunnarmorling:
We could have an option to emit Java 6 compatible code, also when running on 8. Not sure whether it's worth it, though. Java 8 is widely accepted at this point, we've moved to 8 with all the projects I'm working on during the day (Hibernate, Bean Validation, but also WildFly and many others). Haven't heard of complaints. People stuck on older versions can always use an older MapStruct release.

Anyways, I'm closing this one for now. Thanks for the feedback everyone, we seem to have a good path forward. Nice!

Issue #1086, opened 2017-02-19
Use configuration from prototype methods for forged name based methods


filiphr:
This issue is extracted from one task at #1001.

Make sure that configuration from prototype methods in @MapperConfig classes are applied to forged methods (I think that's currently not the case and never has been -- but we should tackle that soon as it becomes more important now)

filiphr:
This was closed by a mistake

sjaakd:
Yes

sjaakd:
sorry.. closed the wrong one..

jean-tissot:
Hi, I'd like to contribute to this issue.
What should I now before doing this (besides what I read in the contribution guidelines)?
I was able to add a [test case](https://github.com/jean-tissot/mapstruct/blob/8a820d0c8b1d4ae78f4e92689fbc3383b12b6c8e/processor/src/test/java/org/mapstruct/ap/test/inheritfromconfig/InheritFromConfigTest.java#L203) of what I would like to implement, but I'm having a hard time finding where to start the implementation (I suppose it's somewhere near `MapperCreationProcessor#getMapper`?)

PR #1117, opened 2017-03-02
#1102 Add support for Auto forging of Enum properties

### LINKED ISSUE #1102: 1.2.0-Beta1 Automapping Feature: doesn't work with enums
enums are not Mapped (Error "Consider to declare/implement a mapping method: ..."):
[MapStructEnumMapper.java.txt](https://github.com/mapstruct/mapstruct/files/801846/MapStructEnumMapper.java.txt)

The Docu "mapstruct-reference-guide.asciidoc" says:

	MapStruct will only create a new mapping method if and only if the source and target property
	are  properties of a Bean and they themselves are Beans or simple properties.
	i.e. they are not Collection or Map type properties.

Maybe enums doesn't fall under "simple properties"?

filiphr:
Fixes #1102 

Maybe we need to do the same that it is done in `MapperCreationProccessor.getMappingMethods()`

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
@@ -56,7 +56,7 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
      *
      * @return See above
      */
-    Assignment createForgedBeanAssignment(SourceRHS sourceRHS, ForgedMethod forgedMethod) {
+    Assignment createForgedAssignment(SourceRHS sourceRHS, ForgedMethod forgedMethod) {
         MappingMethod forgedMappingMethod;
         if ( MappingMethodUtils.isEnumMapping( forgedMethod ) ) {
             forgedMappingMethod = new ValueMappingMethod.Builder()
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
@@ -21,6 +21,7 @@ package org.mapstruct.ap.internal.model;
 import org.mapstruct.ap.internal.model.assignment.Assignment;
 import org.mapstruct.ap.internal.model.common.ParameterBinding;
 import org.mapstruct.ap.internal.model.source.ForgedMethod;
+import org.mapstruct.ap.internal.model.source.MappingMethodUtils;
 import org.mapstruct.ap.internal.model.source.Method;
 
 /**
@@ -56,10 +57,20 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
      * @return See above
      */
     Assignment createForgedBeanAssignment(SourceRHS sourceRHS, ForgedMethod forgedMethod) {
-        BeanMappingMethod forgedMappingMethod = new BeanMappingMethod.Builder()
-            .forgedMethod( forgedMethod )
-            .mappingContext( ctx )
-            .build();
+        MappingMethod forgedMappingMethod;
+        if ( MappingMethodUtils.isEnumMapping( forgedMethod ) ) {
+            forgedMappingMethod = new ValueMappingMethod.Builder()
+                .method( forgedMethod )
+                .valueMappings( forgedMethod.getMappingOptions().getValueMappings() )
+                .mappingContext( ctx )
+                .build();
+        }
+        else {
+            forgedMappingMethod = new BeanMappingMethod.Builder()
+                .forgedMethod( forgedMethod )
+                .mappingContext( ctx )
+                .build();
+        }
 
         return createForgedAssignment( sourceRHS, forgedMethod, forgedMappingMethod );
     }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractMappingMethodBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractMappingMethodBuilder.java
@@ -73,7 +73,7 @@ public abstract class AbstractMappingMethodBuilder<B extends AbstractMappingMeth
             true
         );
 
-        return createForgedBeanAssignment( sourceRHS, forgedMethod );
+        return createForgedAssignment( sourceRHS, forgedMethod );
     }
 
     private String getName(Type sourceType, Type targetType) {
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
@@ -381,9 +381,7 @@ public class SourceMethod implements Method {
 
     public boolean isEnumMapping() {
         if ( isEnumMapping == null ) {
-            isEnumMapping = getSourceParameters().size() == 1
-                && first( getSourceParameters() ).getType().isEnumType()
-                && getResultType().isEnumType();
+            isEnumMapping = MappingMethodUtils.isEnumMapping( this );
         }
         return isEnumMapping;
     }
--- a/processor/src/test/java/org/mapstruct/ap/test/nestedbeans/NestedSimpleBeansMappingTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/nestedbeans/NestedSimpleBeansMappingTest.java
@@ -18,10 +18,12 @@
  */
 package org.mapstruct.ap.test.nestedbeans;
 
+import org.junit.Rule;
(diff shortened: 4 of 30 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
```diff
@@ -381,9 +381,7 @@ public boolean isMapMapping() {
 
     public boolean isEnumMapping() {
         if ( isEnumMapping == null ) {
-            isEnumMapping = getSourceParameters().size() == 1
-                && first( getSourceParameters() ).getType().isEnumType()
-                && getResultType().isEnumType();
+            isEnumMapping = MappingMethodUtils.isEnumMapping( this );
```
gunnarmorling: What's the motivation for extracting the static helper method?

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/UserDtoMapperSmartImpl.java
```diff
@@ -0,0 +1,221 @@
...
+
+        List<org.mapstruct.ap.test.nestedbeans.other.WheelDto> list1 = new ArrayList<org.mapstruct.ap.test.nestedbeans.other.WheelDto>();
+        for ( Wheel wheel : list ) {
+            list1.add( wheelToWheelDto1( wheel ) );
+        }
+
+        return list1;
+    }
+
+    protected org.mapstruct.ap.test.nestedbeans.other.CarDto carToCarDto1(Car car) {
```
gunnarmorling: Any idea why we don't import `CarDto` here? Probably unrelated to the issue at bay, so could you open an issue for keeping track of it?

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/NestedSimpleBeansMappingTest.java
```diff
@@ -41,6 +44,13 @@
 @RunWith(AnnotationProcessorTestRunner.class)
 public class NestedSimpleBeansMappingTest {
 
+    @Rule
+    public final GeneratedSource generatedSource = new GeneratedSource().addComparisonToFixtureFor(
```
gunnarmorling: There doesn't seem to be an actual test for the generated functionality? Having the assertion on the generated source is great, but there should be a test the actual functionality, too.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
filiphr: Because I also need it in the `AbstractBaseBuilder` line 61. In the method `createForgedBeanAssignment`, which I should probable rename to `createForgedAssignment` now. 

I would have extracted the method to a `default` on the `Method`, but no Java 8 for us yet :).

What is your suggestion for reusing the logic?

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/NestedSimpleBeansMappingTest.java
filiphr: What do you mean? That I don't call the `getter` and check if they are actually what we expect? This tests for this actually work with the overridden `equals` on the Dtos. I can change the assertions to actually call the `getters` instead

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/UserDtoMapperSmartImpl.java
filiphr: There are `CarDto` from 2 different packages in here that's why. Those were added to make sure that we are going to generate the method names accordingly.

There is no import for the second CarDto because it is in the same package as the mapper.

filiphr:
@gunnarmorling thanks for the comments. I replied to them inline.

Regarding your question for `@ValueMapping`. With the current implementation it will only work if the source type can be mapped automatically to the target, i.e. the source enum is a subset of the target. 

I see 2 possibilities for dealing with different source and target types:
* If we implement #1086 then the users will have to define it in those prototypes
* Extend the `@ValueMapping` with a `sourceClass` and `targetClass`. Then those should then be filtered and passed down to the forged methods accordingly.

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
```diff
@@ -55,11 +56,21 @@ public B method(Method sourceMethod) {
      *
      * @return See above
      */
-    Assignment createForgedBeanAssignment(SourceRHS sourceRHS, ForgedMethod forgedMethod) {
-        BeanMappingMethod forgedMappingMethod = new BeanMappingMethod.Builder()
-            .forgedMethod( forgedMethod )
-            .mappingContext( ctx )
-            .build();
+    Assignment createForgedAssignment(SourceRHS sourceRHS, ForgedMethod forgedMethod) {
```
filiphr: I renamed the method to match it's new functionality. Is it misleading that we have 2 method with the same name, but different parameters?

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractMappingMethodBuilder.java
```diff
@@ -73,7 +73,7 @@ Assignment forgeMapping(SourceRHS sourceRHS, Type sourceType, Type targetType) {
             true
         );
 
-        return createForgedBeanAssignment( sourceRHS, forgedMethod );
+        return createForgedAssignment( sourceRHS, forgedMethod );
```
sjaakd: Hey guys.. I was thinking. Should forging not be part of the selector or `MappingResolverImpl`? So, if you cannot select something else.. Just forge a method. Then it would solve it not only for `BeanMapping`, but also `IterableMapping` en `MapMapping`..

filiphr: I was thinking the same @sjaakd. That is what I meant with:
>Maybe we need to do the same that it is done in MapperCreationProccessor.getMappingMethods()

If you all agree, I can theoretically do it as part of this PR, or you prefer a new issue?

sjaakd:
I would do it in a separate pr! I issued a new issue earlier today handling list-of-lists. Perhaps it lends itself better for tackling this one..

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/UserDtoMapperSmartImpl.java
gunnarmorling: Ah yes, makes sense. I was surprised by such apparent issue, so it's good to know it actually all is good :)

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/NestedSimpleBeansMappingTest.java
gunnarmorling: Yes, I would have expected a test which accesses the state of the target object.

gunnarmorling:
> With the current implementation it will only work if the source type can be mapped automatically to the target, i.e. the source enum is a subset of the target.

Ok, that makes sense. If that's not doable, an error should be raised, suggesting to create a mapping method for the enum types.

> Extend the @ValueMapping with a sourceClass and targetClass.

I'd rather have people implement a proper enum mapping method, hinted by an error message as described above.

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
gunnarmorling: Seems alright to me :)

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/NestedSimpleBeansMappingTest.java
filiphr: @gunnarmorling with my latest commit the state of the target object is accessed and asserted

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
filiphr: @gunnarmorling you didn't get back to me about this one.

filiphr:
@gunnarmorling, @sjaakd I updated the PR to access the state of the target object when asserting. I also adapted the `Message.VALUE_MAPPING_UNMAPPED_SOURCES` to display the path of the property when there is a history.

Let me know what you think about the latest changes.

If someone else is merging this, I am completely OK (and I would prefer) to squash everything into one commit

PR #1125, opened 2017-03-06
#1124 Don't pass context parameters to forged methods for nested property mapping methods

### LINKED ISSUE #1124: Mapping nested property to simple property does not work when used with context
In 1.2.0-Beta1 when I am passing a context to a mapper using @Context annotation and I am applying a custom mapping from nested property like "entity.id" to "entityId" property (@Mapping(source = "entity.id", target = "id")) generated code does not compile as the invocation of method which extracts "entity.id" value expects context to be passed but it does not get any.

[TestMapper.java.txt](https://github.com/mapstruct/mapstruct/files/821703/TestMapper.java.txt)
[TestMapperImpl.java.txt](https://github.com/mapstruct/mapstruct/files/821708/TestMapperImpl.java.txt)

agudian:
`NestedPropertyMappingMethod` currently can't handle lifecycle methods anyway, so we can skip adding the context parameters.

Fixes #1124.

@sjaakd, I guess once the `SourceRHS` is refactored, it might be easier to add handling of the actual source call (e.g. the property accessor invocation, or as in this case, a call to the nested property mapping method via `MethodReference`) - is that on your map as well?
Initially I just assumed that lifecycle methods are considered in nested property mapping methods, but adding that right now would require some icky code... I guess we should revisit that later on... :)

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -48,6 +48,7 @@ import org.mapstruct.ap.internal.model.source.ForgedMethodHistory;
 import org.mapstruct.ap.internal.model.source.FormattingParameters;
 import org.mapstruct.ap.internal.model.source.MappingOptions;
 import org.mapstruct.ap.internal.model.source.Method;
+import org.mapstruct.ap.internal.model.source.ParameterProvidedMethods;
 import org.mapstruct.ap.internal.model.source.PropertyEntry;
 import org.mapstruct.ap.internal.model.source.SelectionParameters;
 import org.mapstruct.ap.internal.model.source.SourceReference;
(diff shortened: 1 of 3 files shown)


sjaakd:
> I guess once the SourceRHS is refactored, it might be easier to add handling of the actual source call (e.g. the property accessor invocation, or as in this case, a call to the nested property mapping method via MethodReference)

The idea is to give `SourceRHS` the following properties:
* `MappingMeans inner`
* `MappingMeans  outer`
* `String sourceReference`

MappingMeans would then have properties (mutual exclusive):
* `TypeConversion conversion`
* `MethodReference  methodCall`

Not sure on the name `MappingMeans` and perhaps we should rename `MethodReference` to `MethodCall` (seems to cover the content better). 

WDYT? I guess `sourceReference` should probably be a list somehow (to reflect for instance a context parameter), if I understand you correctly.

sjaakd:
BTW: merge when you feel its ok. I'm ok.

agudian:
Thanks for reviewing, guys! I didn't have time to press the merge button the last couple of days... :)

To answer Sjaaks question:
> I guess sourceReference should probably be a list somehow (to reflect for instance a context parameter), if I understand you correctly.

Being able to pass a list here won't be necessary, I guess. The context-parameters are handled seperately with their own list, see `Method#getContextParameters()` and its usages (e.g. how it is usually passed to the constructor of `ForgedMethod`).

Issue #1121, opened 2017-03-03
TypeConversion seems to fail when going to 1.2.0.Beta1


iersel:
When going to MapStruct 1.2.0.Beta1 we get the following errors:

`[ERROR] Can't map property "java.math.BigInteger numberOfLayerComponents" to "java.lang.Integer numberOfLayerComponents". Consider to declare/implement a mapping method: "java.lang.Integer map(java.math.BigInteger value)".`

Seems that typeconversion no longer works between `Integer` and `BigDecimal`

While it compiled on 1.1.0.Final without a problem.

iersel:
Note: MapStruct used to generate:

```java
        if ( sources.getNumberOfLayerComponents() != null ) {
            targets.setNumberConstituents( BigDecimal.valueOf( sources.getNumberOfLayerComponents() ) );
        }
```

iersel:
We are using an update method (with `@MappingTarget`)

agudian:
Great, thanks for reporting! We'll need to look into it for the next release.

sjaakd:
@iersel I can't reproduce it. Lets have a better look on tuesday.

sjaakd:
I've managed to reproduce it. It seems to happen when source and target are in a different jar than the mapper. Need to check how to set this up in unit test.Perhaps using bean-style objects in the java lib (and hope it works)? Candidates?

sjaakd:
problem is most likely in `Type#hashCode` which uses `TypeMirror#hashCode`. JavaDoc `TypeMirror`:  `Types should be compared using the utility methods in Types. There is no guarantee that any particular type will always be represented by the same object. `

Ergo: if they are in different jar's these are most likely not the same objects.. Verifying now..

sjaakd:
Should be solved..

PR #1127, opened 2017-03-07
#1121 TypeMirror should not be used as hashCode base for Type

sjaakd:
Creating a reproducer for this issue is a bit tricky. It requires a new jar containing the source and target for the mapping. Perhaps this is a good idea to have.

For the time being we'll have to do with the explanation in the `Type` class below.

### DIFF
--- a/integrationtest/src/test/java/org/mapstruct/itest/tests/ExternalBeanJarTest.java
+++ b/integrationtest/src/test/java/org/mapstruct/itest/tests/ExternalBeanJarTest.java
@@ -1,0 +1,35 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.itest.tests;
+
+import org.junit.runner.RunWith;
+import org.mapstruct.itest.testutil.runner.ProcessorSuite;
+import org.mapstruct.itest.testutil.runner.ProcessorSuiteRunner;
+
+/**
+ *
+ * See: https://github.com/mapstruct/mapstruct/issues/1121
+ *
+ * @author Sjaak Derksen
+ */
+@RunWith( ProcessorSuiteRunner.class )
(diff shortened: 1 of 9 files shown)


filiphr:
@sjaakd do we need to have the source in one jat and the target in another one? 

An idea for a reproducer would be to have 2 new modules, one for source and another one for target, then we use those modules as test dependencies. 

Or maybe we have some validation checks (Checkstyle or something else) that would prevent us from using hashCode on the Mirrors.

sjaakd:
> do we need to have the source in one jat and the target in another one?

@filiphr : no .. just a jar containing both `Source` and `Target` . The `Mapper` should be in another jar.

filiphr:
@sjaakd then we can maybe have one module with all sources and targets and the processor will contain only tests and not dtos and normal classes

iersel:
We've build the pull requests and the solution works.

agudian:
I guess the easiest way to make sure we don't break it in the future would be a multi-module project in the integration-tests, executed with only one compiler variant (to keep it brief).

sjaakd:
@agudian I've added the requested tests.

agudian:
Hey cool! Could you add a comment to the test pointing to the fixed issue? And does it indeed fail without the code change?

But it looks good, feel free to merge!

sjaakd:
> And does it indeed fail without the code change?

I'm afraid so 😄 .. At least. works on my machine.. 


> And does it indeed fail without the code change?

Done.

gunnarmorling:
@sjaakd Could describe a bit what's the issue specifically? I.e. which type mirrors are different in the case of the different JARs you describe? Your fix seems good, I'd just like to better understand what's the issue here.

sjaakd:
> I'd just like to better understand what's the issue here.

I'm not completely sure either. But I did some troubleshooting. We have all this conversions in the `ConversionProvider` in a map. The map is keyed by a composite key consisting out of source / target type. A match was not made when the `Source` and `Target` requiring a type conversion were in a different jar than the mapper itself.

Suspected the `hashCode` because a class can be multiple times on the class path when packaged in a different jar. Ordering in class loader is important but suspected that `TypeMirror` would make a distinction regardless of ordering somehow. Found evidence in the javedoc of `TypeMirror`. So I looked in the blame of `Type` and found out that the `Type#hashCode` was changed 3 months ago. That's about it..

gunnarmorling:
Ok, I see. What I find odd though is that this about the built-in
conversion of JDK types, hence I'm wondering where multiple type mirrors
for those come from. But either way it seems better to not rely on
Type#hashCode(), that's for sure.

sjaakd:
ok.. I got it to work locally as well by a JDK 7 update.. I'll proceed to merge..

sjaakd:
> Ok, I see. What I find odd though is that this about the built-in
> conversion of JDK types, hence I'm wondering where multiple type mirrors
> for those come from. But either way it seems better to not rely on
> Type#hashCode(), that's for sure.

@gunnarmorling you are right.. It is odd. Almost feels like a bug in the compiler somehow. The fact that `Source` and `Target` are in a different jar should not matter for their attributes which are java-types. Seems that the fact they are member of `Source` and `Target` makes them somehow different.

Having said that: although the underlying cause is unclear: the solution works. Makes me  a bit uncomfortable as well. Something to monitor.

filiphr:
@sjaakd [this](https://bugs.openjdk.java.net/browse/JDK-7159498) JDK bug report might actually be connected to what we are experiencing.

I also checked and we are actually using `TypeMirror#equals()` on 2 places and `TypeMirror#hashCode()` on one more place. I'll open an issue for that.

Issue #1126, opened 2017-03-07
MappingTarget and immutable collections


zelark:
I am trying to apply MapStruct to our project, but I run into a problem. We use our own ORM. If a property is a list, It returns either an emptyList() or an immutable list of values (unfortunately it works this way). The problem is the following generated code for update methods won't work in such case:
```java
        if ( entity.getFilePaths() != null ) {
            List<String> list = dto.getStorageFilePaths();
            if ( list != null ) {
                entity.getFilePaths().clear(); // this throws an exception
                entity.getFilePaths().addAll( list );
            }
            else {
                entity.setFilePaths( null );
            }
        }
        else {
            List<String> list = dto.getStorageFilePaths();
            if ( list != null ) {
                entity.setFilePaths(       new ArrayList<String>( list )
                );
            }
        }
```

The question is how can we tell MapStruct just to generate code like in the else block for update methods?

Also I found #153 where clear and addAll came from. And it seems in #728 describing the same feature which I'd like to get.

zelark:
Hi guys, can you help, please?

sjaakd:
@zelark .. I'll try to have a look later this weekend.

sjaakd:
@zelark .. for the current setup I don't have a definite solution.. However, you might want to try to define a hand written method for all `List<X>` to `List<Y>` you need along this line:

```java
@Mapper
public abstract class ExampleMapper {


    @Mapping(target = "integer", source = "bigDecimal")
    abstract void map(Source source, @MappingTarget Target target);


    List<String> map(List<String> source ) {
        if ( source != null && !source.isEmpty() ) {
            return new ArrayList();
        }
        return Collections.emptyList();
    }
}
```

That will generate the following code

```java
public class ExampleMapperImpl extends ExampleMapper {

    @Override
    void map(Source source, Target target) {
        if ( source == null ) {
            return;
        }

        if ( source.getBigDecimal() != null ) {
            target.setInteger( source.getBigDecimal().intValue() );
        }
        if ( target.getList() != null ) {
            List<String> list = map( source.getList() );
            if ( list != null ) {
                target.getList().clear();
                target.getList().addAll( list );
            }
            else {
                target.setList( null );
            }
        }
        else {
            List<String> list = map( source.getList() );
            if ( list != null ) {
                target.setList( list );
            }
        }
    }
}
```

Alternatively you might want to have a look at `@BeforeMapping` , initializing all collection types that have a non-empty, non-null source.

[...]

Issue #487, opened 2015-03-03
Follow up of #160, mapping method selection should select IterableMapping and MapMapping as well.


sjaakd:
Allow update method selection mechanism for properties to work on collections and maps.

sjaakd:
There are several cases to consider here:
1. Lists
2. Sets
3. Arrays
4. Maps.

Lets start with 1. 

Assume we have a
- source `List` with elements `[A, C, D]` 
- target `List` with elements `[A', B' D']`
- there is a mapping method (both create and update)  available for the elements `X -> X`]

I would expect the result to look like:
`[A'', C' D'']`
Where: 
- `A'` and `D'` are updated to `A''` respectively `D''`
- `B'` is deleted, 
- `C'` is created

All in all the order is maintained.

For 2. Its likewise, only the order is irrelevant.

For 3. The array size needs to be the same. Any deviation of order cannot be mapped and should throw an exception (which?)

For 4. The should be the determining factor. In absence of a key in the target, a new entry will be added. Superflous keys will be removed.

This will all depends on meaningful `hashcode` and `equals` methods in the target objects. 

-------------------
Solution wise:

We also need some intermediate results. The source collection / map first needs to be mapped completely (we could call a forged method for that if a mapping is absent). And then we execute the merging logic along the lines that is described above.

nocquidant:
FYI GeDa uses matcher objects.
http://www.inspire-software.com/confluence/display/GeDA/Annotations+API#AnnotationsAPI-@DtoCollection.dtoToEntityMatcher

But the O(n^2) could be improved I guess.

sjaakd:
> FYI GeDa uses matcher objects

Lets first try to tackle the simple case. Would you be interested to help out on this one?

nocquidant:
I'm afraid I do not find the time to do it. Hats off ;)

sjaakd:
As an in-between solution we could consider selection of handwritten methods with target type collection / map as a first step.

agudian:
@sjaakd, we should close this, right?

sjaakd:
Nope.. This one concerns also generation of these mappings. So this is only the first (small) step.

agudian:
Ok, then I'd suggest setting the milestone to 1.1... :)

sjaakd:
> Ok, then I'd suggest setting the milestone to 1.1... :)

ack.

marcoschulte:
We might find the time to implement that feature.

[...]

PR #1133, opened 2017-03-12
#1126 adding the option to handle immutable collections in update methods

sjaakd:
A new `CollectionMappingStrategy.TARGET_IMMUTABLE` has been introduced. The solution does not call the clear/addAll on the target collection. Instead it uses the setter on the target.

Pending: while handling the nice table in the documentation: what do we do when we encounter conflicting situations: e.g. there's no setter? Probably we should generate an error. 

I'll add that scenario, when we agree in general on this approach.

### DIFF
--- a/core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
+++ b/core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
@@ -30,7 +30,8 @@ public enum CollectionMappingStrategy {
      * {@code orderDto.setOrderLines(order.getOrderLines)}.
      * <p>
      * If no setter is available but a getter method, this will be used, under the assumption it has been initialized:
-     * {@code orderDto.getOrderLines().addAll(order.getOrderLines)}.
+     * {@code orderDto.getOrderLines().addAll(order.getOrderLines)}. This will also be the case when using
+     * {@link MappingTarget} (updating existing instances).
      */
     ACCESSOR_ONLY,
 
@@ -51,5 +52,13 @@ public enum CollectionMappingStrategy {
      * Identical to {@link #SETTER_PREFERRED}, only that adder methods will be preferred over setter methods, if both
      * are present for a given collection-typed property.
      */
-    ADDER_PREFERRED;
+    ADDER_PREFERRED,
+
+    /**
+     * Identical to {@link #SETTER_PREFERRED}, however the target collection will not be cleared and accessed via
+     * addAll in case of updating existing bean instances, see: {@link MappingTarget}.
+     *
+     * Instead the target accessor (e.g. set) will be used on the target bean to set the collection.
+     */
+    TARGET_IMMUTABLE;
 }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -469,7 +469,8 @@ public class Type extends ModelElement implements Comparable<Type> {
             // the current target accessor can also be a getter method.
             // The following if block, checks if the target accessor should be overruled by an add method.
             if ( cmStrategy == CollectionMappingStrategyPrism.SETTER_PREFERRED
-                || cmStrategy == CollectionMappingStrategyPrism.ADDER_PREFERRED ) {
+                || cmStrategy == CollectionMappingStrategyPrism.ADDER_PREFERRED
+                || cmStrategy == CollectionMappingStrategyPrism.TARGET_IMMUTABLE ) {
 
                 // first check if there's a setter method.
                 Accessor adderMethod = null;
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -466,7 +466,7 @@ Specifying the parameter in which the property resides is mandatory when using t
 Mapping methods with several source parameters will return `null` in case all the source parameters are `null`. Otherwise the target object will be instantiated and all properties from the provided parameters will be propagated.
 ====
 
-MapStruct also offers the possibility to directly refer to a source parameter. 
+MapStruct also offers the possibility to directly refer to a source parameter.
 
 .Mapping method directly referring to a source parameter
 ====
@@ -1338,7 +1338,7 @@ public Map<Long, Date> stringStringMapToLongDateMap(Map<String, String> source) 
 [[collection-mapping-strategies]]
 === Collection mapping strategies
 
-MapStruct has a `CollectionMappingStrategy`, with the possible values: `ACCESSOR_ONLY`, `SETTER_PREFERRED` and `ADDER_PREFERRED`.
+MapStruct has a `CollectionMappingStrategy`, with the possible values: `ACCESSOR_ONLY`, `SETTER_PREFERRED`, `ADDER_PREFERRED` and `TARGET_IMMUTABLE`.
 
 In the table below, the dash `-` indicates a property name. Next, the trailing `s` indicates the plural form. The table explains the options and how they are apply to the presence/absense of a `set-s`, `add-` and / or `get-s` method on the target object:
 
@@ -1366,6 +1366,13 @@ In the table below, the dash `-` indicates a property name. Next, the trailing `
 |add-
 |get-s
 |get-s
+
+|`TARGET_IMMUTABLE`
+|set-s
+|exception
+|set-s
+|exception
+|set-s
 |===
 
 Some background: An `adder` method is typically used in case of http://www.eclipse.org/webtools/dali/[generated (JPA) entities], to add a single element (entity) to an underlying collection. Invoking the adder establishes a parent-child relation between parent - the bean (entity) on which the adder is invoked - and its child(ren), the elements (entities) in the collection. To find the appropriate `adder`, MapStruct will try to make a match between the generic parameter type of the underlying collection and the single argument of a candidate `adder`. When there are more candidates, the plural `setter` / `getter` name is converted to singular and will be used in addition to make a match.
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -52,6 +52,7 @@ import org.mapstruct.ap.internal.model.source.ParameterProvidedMethods;
 import org.mapstruct.ap.internal.model.source.PropertyEntry;
 import org.mapstruct.ap.internal.model.source.SelectionParameters;
 import org.mapstruct.ap.internal.model.source.SourceReference;
+import org.mapstruct.ap.internal.prism.CollectionMappingStrategyPrism;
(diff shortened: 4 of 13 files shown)


sjaakd:
@gunnarmorling : can you check whether you agree with extending the `CollectionMappingStragegy`.

core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
```diff
@@ -30,7 +30,8 @@
      * {@code orderDto.setOrderLines(order.getOrderLines)}.
      * <p>
      * If no setter is available but a getter method, this will be used, under the assumption it has been initialized:
-     * {@code orderDto.getOrderLines().addAll(order.getOrderLines)}.
+     * {@code orderDto.getOrderLines().addAll(order.getOrderLines)}. This will also be the case in case when using
```
filiphr: Wording: "This will also be the case when using..." there is extra "in case"

core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
```diff
@@ -51,5 +52,12 @@
      * Identical to {@link #SETTER_PREFERRED}, only that adder methods will be preferred over setter methods, if both
      * are present for a given collection-typed property.
      */
-    ADDER_PREFERRED;
+    ADDER_PREFERRED,
+
+    /**
+     * Identical to {@link #SETTER_PREFERRED}, however a target collection will not cleared be accessed via addAll.
```
filiphr: Wording: "...will not be cleared and accessed via addAll"

core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
```diff
@@ -51,5 +52,12 @@
...
+     * {@code orderDto.getOrderLines().clear(); orderDto.getOrderLines().addAll(order.getOrderLines)}. Instead the
```
filiphr: Maybe we should write the code with:

```
<pre>
orderDto.getOrderLines().clear();
orderDto.getOrderLines().addAll(order.getOrderLines);
</pre>
```

This way it will be on 2 lines when viewing the Javadoc

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -469,7 +469,8 @@ public boolean isAssignableTo(Type other) {
             // the current target accessor can also be a getter method.
             // The following if block, checks if the target accessor should be overruled by an add method.
             if ( cmStrategy == CollectionMappingStrategyPrism.SETTER_PREFERRED
-                || cmStrategy == CollectionMappingStrategyPrism.ADDER_PREFERRED ) {
+                || cmStrategy == CollectionMappingStrategyPrism.ADDER_PREFERRED
+                || cmStrategy == CollectionMappingStrategyPrism.TARGET_IMMUTABLE ) {
```
filiphr: Shouldn't we do a check, that the strategy is not `TARGET_IMMUTABLE` on line 482? Maybe this is connected with your comment about the exceptions when no setter exists

agudian:
> Pending: while handling the nice table in the documentation: what do we do when we encounter conflicting situations: e.g. there's no setter? Probably we should generate an error.

I agree with you and Filip, we should then generate an error... what whould be the point of declaring a target collection as immutable and then don't have a way to set it (setter / public field, or soonish a constructor)... :)

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
sjaakd: The check is added, but inside property mapping.. 

Looking at this code, I'm thinking.. Is the table correct in the documentation? The documentation has a column "Only target add- Available".. That is not possible. The adder is derived from the presence of either a setter and/or a getter. Note: candidates only contains getters/setters... WDYT?

If this is wrong.. we should handle it in a separate PR..

sjaakd:
Ok.. added the error scenario + test case. I think the fundamental question remains.. Do we want this extra enum value?

zelark:
> what whould be the point of declaring a target collection as immutable and then don't have a way to set it (setter / public field, or soonish a constructor)... :)

it may be worth to take into account, but, again, In our project we have add methods instead of set ones for the parent-child relation. Today, I checked out what would happened if I set ```CollectionMappingStrategy``` in ```ADDER_PREFERRED```. And I got the following code for an update method:
```java
    ...
        if ( dto.getValues() != null ) {
            for ( ValueDTO value : dto.getValues() ) {
                entity.addValue( dtoToValue( value ) );
            }
        }
    ...
```

where the ```addValue``` method looks as following:
```
void addValue(ValueEntity value) {
    value.setParentKey(this.PrimaryKey());
}
```
Some weird thing here is that we don't delete the old children but simply add new ones. Does it mean ```ADDER_PREFERRED``` don't suppose to be used for updating, or we should get rid of the old children manually? If the latter is true, maybe it makes sense to use ```add-``` when only target ```add-``` available and ```TARGET_IMMUTABLE``` specified.

core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
```diff
@@ -51,5 +52,16 @@
...
      * are present for a given collection-typed property.
      */
-    ADDER_PREFERRED;
+    ADDER_PREFERRED,
+
+    /**
+     * Identical to {@link #SETTER_PREFERRED}, however the target collection will not be cleared and accessed via
+     * addAll.
+     * <pre>
+     * orderDto.getOrderLines().clear();
```
gunnarmorling: I'd rather show the code that _gets_ generated instead of the one which isn't.

core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
```diff
@@ -51,5 +52,16 @@
...
+     * orderDto.getOrderLines().addAll(order.getOrderLines);
+     * </pre>.
+     * Instead the target accessor (e.g. set) will be used on the target bean to set the collection.
+     */
+    TARGET_IMMUTABLE;
```
gunnarmorling: Should it be named `SETTER_ONLY`?

core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
```diff
@@ -51,5 +52,16 @@
      * Identical to {@link #SETTER_PREFERRED}, only that adder methods will be preferred over setter methods, if both
      * are present for a given collection-typed property.
      */
-    ADDER_PREFERRED;
+    ADDER_PREFERRED,
+
+    /**
+     * Identical to {@link #SETTER_PREFERRED}, however the target collection will not be cleared and accessed via
```
gunnarmorling: > however the target collection will not be cleared

That does only apply to update methods, right?

core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
sjaakd: True.. I removed the code snippet all together.. Does not help.

core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
sjaakd: Good point. Thought about that one. I'd like to express that the use case related to immutable collections.  I'd like to stick to TARGET_IMMUTABLE (if you don't have any objections).

core-common/src/main/java/org/mapstruct/CollectionMappingStrategy.java
sjaakd: True.. added that.. We should however align our terminology.. I now used the frasing: `updating existing bean instances` which is in line with the documentation.

However, I would like to make some kind of glossary for the terms we introduced.  Update method (or perhaps merge method) would sound better in the long run, don't you think?

sjaakd:
If you guys agree, I'll continue to merge later..

sjaakd:
> Some weird thing here is that we don't delete the old children but simply add new ones. 

This is true.. update method were developed later.

> Does it mean ADDER_PREFERRED don't suppose to be used for updating, or we should get rid of the old children manually?

How would MapStruct be able to decide which entity to remove?

>  If the latter is true, maybe it makes sense to use add- when only target add- available and TARGET_IMMUTABLE specified.

TARGET_IMMUTABLE will apply to all collections in the Mapper. If you have a need to distinguish, you might need to separate the desired behaviors in different mappers.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: @sjaakd thinking a bit about this. `ADDER_PREFERRED` does not mean `list.add(something)`, it means `entity.add(something)`, right? Theoretically, an adder should also work (if the user implements it correctly). 

@sjaakd I would suggest the following

| Option | Only target set-s Available | Only target add- Available | Both set-s / add- Available | No set-s / add- Available | Existing Target(@TargetType) |
| ------------- | ------------- | ------------- | ------------- |  ------------- | ------------- | 
| `TARGET_IMMUTABLE` | set-s | add | set-s | exception | set-s |

We will only throw exception if there is no adder or setter. WDYT?

sjaakd: > Thinking a bit about this. ADDER_PREFERRED does not mean list.add(something), it means entity.add(something), right?

That's true.. is that not clear from the doc?


`Only target add- Available` .. This cannot be. An adder can only be there when there is also an getter and/or setter..  So I think the table is wrong to start with.. So I think the column should disappear al together.  I'll do that in a separate PR.. Want to close this one.

filiphr: No it is clear. And you are right `Only target add- Available` is wrong, that cannnot happen (at least with the current implementation).

zelark:
> This is true.. update method were developed later.
Hm, if it's not a bug, I think, it should be documented at least. Actually, I didn't expect such behavior.

> How would MapStruct be able to decide which entity to remove?
A good question to think. That's related only to update methods and when ```ADDER_PREFERRED``` is used.

filiphr:
@zelark it is written in the documentation in [Updating existing bean instances](http://mapstruct.org/documentation/stable/reference/html/#updating-bean-instances).

It says:

>Collection- or map-typed properties of the target bean to be updated will be cleared and then populated with the values from the corresponding source collection or map.

Is that what you are looking for, or it is something else?

zelark:
I mean if you specify ```ADDER_PREFERRED``` you'll get the following in your update method:
```java
    @Override
    public Car updateCar(CarDto dto, Car car) {
        ...
        if ( dto.getPersons() != null ) {
            for ( PersonDto person : dto.getPersons() ) {
                car.addPerson( dtoToPerson( person ) );
            }
        }
       ...
```
It don't clear the collection.

sjaakd:
> It doesn't clear the collection.

That's probably not consistent with other places where collections are used. I'll open up a new issue for that.. #1142.

Note: update method and collections / maps are not trivial. We've got a long standing issue on that #487. I'll just have to consider adder behavior as well.

Issue #1135, opened 2017-03-12
Support overloading for target


quantumlexa:
Mapstruct  does support overloading of target setters(see: Issue892Test test) but choosing preferred type logic is based on getter type of target - not source, which looks a bit strange.
org.mapstruct.ap.internal.model.common.Type#getPropertyWriteAccessors: line 465

Please see unit test:
Source:
```
public class Source {

    private Date updatedOn;

    public Source(Date updatedOn) {
        this.updatedOn = updatedOn;
    }

    public Date getUpdatedOn() {
        return updatedOn;
    }

    public void setUpdatedOn(Date updatedOn) {
        this.updatedOn = updatedOn;
    }
}
```

Target:
```
public class Target {

    private long updatedOn;

    public long getUpdatedOn() {
        return updatedOn;
    }

    public void setUpdatedOn( long updatedOn ) {
        this.updatedOn = updatedOn;
    }

    public void setUpdatedOn( Date updatedOn ) {
        if (updatedOn == null) {
            return;
        }
        this.updatedOn = updatedOn.getTime();
    }

}
```
Mapper:
```
@Mapper
public interface SourceTargetMapper {

    SourceTargetMapper INSTANCE = Mappers.getMapper( SourceTargetMapper.class );

    @Mapping(target = "updatedOn", source = "updatedOn")
    Target sourceToTarget( Source source );

}
```
UnitTest:
```
@WithClasses({
        SourceTargetMapper.class,
        Source.class,
        Target.class
})
@RunWith(AnnotationProcessorTestRunner.class)
public class OverloadingTest {

    @Test
    public void testShouldGenerateCorrectMapperImplementation() {
        Source source = new Source( new Date() );
        Target target = SourceTargetMapper.INSTANCE.sourceToTarget( source );
        Assert.assertTrue( target.getUpdatedOn() > 0 );
    }

}
```

therottman:
+1

rivierej:
+1

aioobe:
This was never merged? Is there a particular reason?

filiphr:
The PR for this hasn't been merged since the MapStruct team hasn't had the time to look into it in detail. The PR is not a small change (functionality wise). Therefore, we need to be careful when merging it.

onurcelik98:
Is there an estimation about when this may be merged?

neboskreb:
November 2025. Bump up.

PR #1136, opened 2017-03-12
#1135 added test and implementation for overloading

quantumlexa:
Mapstruct does support overloading of target setters(see: Issue892Test test) but choosing preferred type logic is based on getter type of target - not source, which looks a bit strange.
Here is straightforward implementation of this

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -437,6 +437,11 @@ public class Type extends ModelElement implements Comparable<Type> {
         return presenceCheckers;
     }
 
+    public Map<String, Accessor> getPropertyWriteAccessors( CollectionMappingStrategyPrism cmStrategy ) {
+        return getPropertyWriteAccessors( cmStrategy, Collections.<String, Accessor>emptyMap() );
+    }
+
+
     /**
      * getPropertyWriteAccessors returns a map of the write accessors according to the CollectionMappingStrategy. These
      * accessors include:
@@ -450,7 +455,8 @@ public class Type extends ModelElement implements Comparable<Type> {
      * @param cmStrategy collection mapping strategy
      * @return an unmodifiable map of all write accessors indexed by property name
      */
-    public Map<String, Accessor> getPropertyWriteAccessors( CollectionMappingStrategyPrism cmStrategy ) {
+    public Map<String, Accessor> getPropertyWriteAccessors( CollectionMappingStrategyPrism cmStrategy,
+                                                            Map<String, Accessor> sourceAccessor ) {
         // collect all candidate target accessors
         List<Accessor> candidates = new ArrayList<Accessor>( getSetters() );
         candidates.addAll( getAlternativeTargetAccessors() );
@@ -495,13 +501,47 @@ public class Type extends ModelElement implements Comparable<Type> {
             }
 
             Accessor previousCandidate = result.get( targetPropertyName );
-            if ( previousCandidate == null || preferredType == null || ( targetType != null
-                && typeUtils.isAssignable( preferredType.getTypeMirror(), targetType.getTypeMirror() ) ) ) {
+
+
+            if (previousCandidate != null
+                    && sourceAccessor != null
+                    && sourceAccessor.get( targetPropertyName ) != null
+                    && isBetter( previousCandidate, candidate,
+                    sourceAccessor.get( targetPropertyName ).getAccessedType() ) ) {
+                result.put( targetPropertyName, candidate );
+                continue;
+            }
+
+
+            if (previousCandidate == null || preferredType == null || (targetType != null
+                    && typeUtils.isAssignable( preferredType.getTypeMirror(), targetType.getTypeMirror() ) ) ) {
                 result.put( targetPropertyName, candidate );
             }
         }
 
         return result;
+    }
+
+    private TypeMirror extractTypeMirror( Accessor accessor ) {
+        return accessor.getExecutable().getParameters().get( 0 ).asType();
+    }
+
+    private boolean isBetter( Accessor previousCandidate, Accessor nextCandidate, TypeMirror preferredType ) {
+        TypeMirror previousType = extractTypeMirror( previousCandidate );
+        TypeMirror nextType = extractTypeMirror( nextCandidate );
+
+        if (typeUtils.isSameType( previousType, preferredType )) {
+            return false;
+        }
+
+        if (typeUtils.isSameType( nextType, preferredType )) {
+            return true;
+        }
+        if (typeUtils.isAssignable( nextType, preferredType )
+                && !typeUtils.isAssignable( previousType, preferredType )) {
+            return true;
+        }
+        return false;
     }
 
     private Type determinePreferredType(Accessor readAccessor) {
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
@@ -26,6 +26,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.LinkedHashMap;
 
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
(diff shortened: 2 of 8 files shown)


gunnarmorling:
ok to test

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -495,18 +501,54 @@ else if ( Executables.isFieldAccessor( candidate ) && ( Executables.isFinal( can
...
                 result.put( targetPropertyName, candidate );
             }
         }
 
         return result;
     }
 
-    private Type determinePreferredType(Accessor readAccessor) {
+    private TypeMirror extractTypeMirror( Accessor accessor ) {
+        return accessor.getExecutable().getParameters().get( 0 ).asType();
```
filiphr: This is wrong it can lead to NPE. The `accessor` does not always have an executable. MapStruct can also map public field and that is why we have introduced the `Accessor`.  I think that this can be done by only using the MapStruct `Type` and not the `TypeMirror`. Maybe you can also use `Type` in the `isBetter` check as well.

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
```diff
@@ -203,7 +203,10 @@ private static TypeElement asTypeElement(TypeMirror mirror) {
      *
      * @return the executable elements usable in the type
      */
-    public static List<Accessor> getAllEnclosedAccessors(Elements elementUtils, TypeElement element) {
+    public static List<Accessor> getAllEnclosedAccessors( Elements elementUtils, TypeElement element ) {
+        if (element == null) {
```
filiphr: Did you have problems with an NPE here?

processor/src/test/java/org/mapstruct/ap/test/overloading/OverloadingTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+import java.util.Date;
+
+@WithClasses({
+        SourceTargetMapper.class,
+        SourceTargetMapperPropertyBean.class,
```
filiphr: What is the difference between the `SourceTargetMapper` and the `SourceTargetMapperPropertyBean` is it only the `@Mapping` annotation? Btw. when the source and target have the same name you don't need it. If you want to test that it works for different name you can just add new fields :).

processor/src/test/java/org/mapstruct/ap/test/overloading/OverloadingTest.java
```diff
@@ -0,0 +1,54 @@
...
+        Assert.assertTrue( target.getUpdatedOn() > 0 );
+    }
+
+    @Test
+    @WithClasses({
+            SourceTargetMapperPropertyBean.class,
+            Source.class,
+            Target.class
+    })
+    public void testShouldGenerateCorrectMapperImplementationJsr() {
```
filiphr: Why do you need this test method? Also the `@WithClasses` on the method is not needed, because those classes are already added on the class level. We usually add `@WithClasses` to methods when we want to test errors within the mappers.

processor/src/test/java/org/mapstruct/ap/test/overloading/OverloadingTest.java
```diff
@@ -0,0 +1,54 @@
...
+        Target.class
+})
+@RunWith(AnnotationProcessorTestRunner.class)
+public class OverloadingTest {
+
+    @Test
+    public void testShouldGenerateCorrectMapperImplementation() {
+        Source source = new Source( new Date() );
+        Target target = SourceTargetMapper.INSTANCE.sourceToTarget( source );
+        Assert.assertTrue( target.getUpdatedOn() > 0 );
```
filiphr: Can you use assertj `Assertions` here? I would also add the implementation class as a comparison fixture to the `GeneratedSource` rule.  You can have a look at the `ArrayMappingTest` and the `ScienceMapperImpl.java` fixture for how it works.

agudian: As this PR adresses only a behavioral change and doesn't structurally change the generated code, I don't think we need that fixture here - we should try keep the usage of those to a minimum, as we'll always need to keep maintaining them in the future... 🙂

filiphr: @agudian is right for the fixture. Just to make sure that we don't miss this in the future, maybe the `setUpdateOn(long updateOn)` in the `Target` should throw an exception. That method should not be called.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
```diff
@@ -612,4 +615,19 @@ public boolean isAbstract() {
     public boolean isUpdateMethod() {
         return getMappingTargetParameter() != null;
     }
+
+    public Map<String, Accessor> getGetSourceReadAccessors() {
```
filiphr: It looks like your last version was reverted to before you moved the `getSourceReadAccessors` to the `Method` interface.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -450,7 +455,8 @@ public boolean isAssignableTo(Type other) {
      * @param cmStrategy collection mapping strategy
      * @return an unmodifiable map of all write accessors indexed by property name
      */
-    public Map<String, Accessor> getPropertyWriteAccessors( CollectionMappingStrategyPrism cmStrategy ) {
+    public Map<String, Accessor> getPropertyWriteAccessors( CollectionMappingStrategyPrism cmStrategy,
```
filiphr: This will only work when the `source` and `target` properties are the same.

If you modify the name in either `Source` or `Target` and accordingly use `@Mapping` then it'll fail.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -495,15 +501,49 @@ else if ( Executables.isFieldAccessor( candidate ) && ( Executables.isFinal( can
...
+                    && typeUtils.isAssignable( preferredType.getTypeMirror(), targetType.getTypeMirror() ) ) ) {
                 result.put( targetPropertyName, candidate );
             }
         }
 
         return result;
     }
 
+    private TypeMirror extractTypeMirror( Accessor accessor ) {
+        return accessor.getExecutable().getParameters().get( 0 ).asType();
```
filiphr: In theory you should not be able to get up to here if your `Accessor` is field. However, I think that a `null` check needs to be done on the `accessor.getExecutable()`.

filiphr:
@agudian, I had another look at the PR. There are some issues which are addressed in my second review. However, I don't think that there is an issue with the selectors:

* If none of the `source` accessors can be matched to the target then the type of the `getter` is used
* If a `source` accessor is used, then a direct mapping will always be done, because the type of the source is either the same as the target, or it is assignable to the target.

@quantumlexa Can you perhaps add more properties and make sure that all the branches in `Type#isBetter` are tested?

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
quantumlexa: yes

processor/src/test/java/org/mapstruct/ap/test/overloading/OverloadingTest.java
quantumlexa: removed body of test at all. It should compile ant that's enough.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -437,6 +437,11 @@ public boolean isAssignableTo(Type other) {
         return presenceCheckers;
     }
 
+    public Map<String, Accessor> getPropertyWriteAccessors( CollectionMappingStrategyPrism cmStrategy ) {
```
quantumlexa: done

quantumlexa:
due to some errors with git branch - recreated it in: https://github.com/mapstruct/mapstruct/pull/1136

filiphr:
PR moved to #1167

Issue #1118, opened 2017-03-02
Forge iterable / map mappings from inside an iterable / map mappings


sjaakd:
To handle lists of lists, like discussed in #1111

filiphr:
As suggested by @sjaakd  way to do this would be to:

> move the forging into a selector or MappingResolverImpl. So, if you cannot select something else.. Just forge a method.

PR #1137, opened 2017-03-12
#1104 use unmapped target policy for forged name based mappings

### LINKED ISSUE #1104: 1.2.0-Beta1 Automapping Feature: Doesn't work with "Unmapped target property" on the child Bean.
An "Unmapped target property" on the child Bean shows the error `Consider to declare/implement a mapping method: ...`:
[MapStructWithHirarchyMapper.java.txt](https://github.com/mapstruct/mapstruct/files/801853/MapStructWithHirarchyMapper.java.txt)

The `annotation @Mapping(ignore = true, target = "child.id")` solves that issue.  
But, ...

1. There should not be an Error. A warning like `Unmapped target property: "child.id"` would be more appropriate.
	
2. If unmappedTargetPolicy = ReportingPolicy.IGNORE then there should not even be a warning.

Background:
This was the root error I searched for: My Domain Objects all extends a MappedSuperClass which contains the Database Id.
This MappedSuperClass should not be relevant for the mapping, because the DB-Id is not exposed to the DTOs.

filiphr:
Fixes #1104.

This PR is a starter for supporting unmapped target policy for the forged named based mappings.

There are some tests that are failing and I wanted to discuss with you about the approach to "fix" them or remove them. 

In my opinion most of the tests (the ones that are failing because the compilation succeeds) need to be updated and have the unmapped target property message with warning.

One main change is that an error is no longer reported if the target properties for a forged method are empty. I think that we should behave the same way as the user has written the method.

For example for the `Issue590Test`. One can define a method like: 

```java
void map(String source, @MappingTarget xmlFormatter);
```

and we are going to generate an implementation for it. Yes it'll be empty and it'll look like:

```java
@Override
public void map(String source, XMLFormatter xmlFormatter) {
    if ( source == null ) {
        return;
    }
}
```

Maybe I should add the generated Sources for those tests to the source code?

WDYT?


gunnarmorling:
@filiphr What is your suggestion to do for the `String` - `XMLFormatter` case?

We could raise a warning when generating a (forged) method without any property mappings at all, as it indicates that something is fishy. But otherwise I'm not sure how we could detect that something is wrong.

It's this kind of situation which is why I'm still skeptical about the auto-mapping feature (and why we should have an option to turn it off).

filiphr:
@gunnarmorling In my opinion I don't see a difference between a mapping like `String` - `XMLFormatter` and `Source` - `Target` where the target is empty (makes no sense to have empty target, but still :)).

I was thinking to just mark the test as it is a success. I agree with you having a target that is empty is fishy. Therefore, I think that we should always emit a warning when there are no target properties, not only for (forged) name based mappings. We can even suggest the method that needs to be created.

A user can still write a method like

```java
XMLFormatter map(String source);
```
and MapStruct will generate it empty. 

Regarding the turning off. I also think that it'll be nice to have the option. Maybe it'll be best to do #1118 (or at least move the forged mapping in the MappingResolver). If we do that then we can more easily apply the turning of. Moreover, this PR can benefit from turned of auto mappings (all those tests will not fail).

agudian:
> We could raise a warning when generating a (forged) method without any property mappings at all, as it indicates that something is fishy. But otherwise I'm not sure how we could detect that something is wrong.

I don't know... then the only way to remove the warning would be to manually add a method declaration for that source/target type pair... But that would be funny as well. How about having another option in an annotation to actively acknowledge and suppress the warning? It could be applied to manually declared methods for that type pair as well. And one should be able to put that annotation to a prototype method as well...

agudian:
Btw, my suggestion would be to let this PR only handle the unmapped target policy thing, and let's discuss the warning and a proper suppression mechanism in a seperate issue.

filiphr:
>Btw, my suggestion would be to let this PR only handle the unmapped target policy thing, and let's discuss the warning and a proper suppression mechanism in a seperate issue.

If you agree then I will add back the `targetProperties.isEmpty()` check and for forged methods we will output an error like we were doing before. Only then we can move the discussion for that in a new issue, it doesn't make sense to fix the tests when we know that it is not a good approach to leave it like that :)

gunnarmorling:
> I don't know... then the only way to remove the warning would be to manually add a method declaration for that source/target type pair...

Right, but that's the most reasonable thing to do, no? In which case would you have a mapping without any properties?

But +1 for discussing this separately.

> I don't know... then the only way to remove the warning would be to manually add a method declaration for that source/target type pair...

+1

filiphr:
I have created a new issue for when the target is empty.

Just to add to the discussion, the user has to implement a method, if he/she just declares one then it would be the same.

I updated the PR and left the empty target properties. Please have a look at the modified tests and let me know if you agree with them

agudian:
Just checking on my phone here on the subway, but does unmappedTargetProperty=IGNORE work as well?
And could you rename the Erroneous* mappers that now only report warnings?

filiphr:
@agudian Yes `unmappedTargetProperty=IGNORE` works. Have a look at the `UnmappableIgnoreXXXMapper`(s).

For the `Erroneous*` mappers. Should I rename them so they can run in the integration tests or just the naming? I am asking, because they are still in a `erroneous` package.

agudian:
> Yes unmappedTargetProperty=IGNORE works. Have a look at the UnmappableIgnoreXXXMapper(s).

Cool 👍 

> For the Erroneous* mappers. Should I rename them so they can run in the integration tests or just the naming? I am asking, because they are still in a erroneous package.

Yes, please rename and move out of that package... :)

agudian:
Looks good otherwise, feel free to merge when you're done 🎉

filiphr:
Merged. Thanks a lot for the review @gunnarmorling, @sjaakd and @agudian :)

PR #1146, opened 2017-03-20
#993 Add support for disabling the generation of forged mapping methods

### LINKED ISSUE #993: Add a switch to turn off auto-mappings
Follows up on #60.

filiphr:
Fixes #993.

Please have a look at the field name in the API and the Javadoc. The field is a bit verbose, but I couldn't think of something else. I am open to ideas :)

### DIFF
--- a/core-common/src/main/java/org/mapstruct/Mapper.java
+++ b/core-common/src/main/java/org/mapstruct/Mapper.java
@@ -159,4 +159,23 @@ public @interface Mapper {
      * @return strategy how to do null checking
      */
     NullValueCheckStrategy nullValueCheckStrategy() default ON_IMPLICIT_CONVERSION;
+
+    /**
+     * If MapStruct could not find another mapping method or apply an automatic conversion it will try to generate a
+     * sub-mapping method between the two beans. If this property is set to {@code true} MapStruct will not try to
+     * automatically generate sub-mapping methods.
+     * <p>
+     * Can be configured by the {@link MapperConfig#disableSubMappingMethodsGeneration()} as well.
+     * <p>
+     * Note: If you need to use {@code disableSubMappingMethodsGeneration} please contact the MapStruct team at
+     * <a href="http://mapstruct.org">mapstruct.org</a> or
+     * <a href="https://github.com/mapstruct/mapstruct">github.com/mapstruct/mapstruct</a> to share what problem you
+     * are facing with the automatic sub-mapping generation.
+     *
+     * @return whether the automatic generation of sub-mapping methods is disabled
+     *
+     * @since 1.2
+     */
+    boolean disableSubMappingMethodsGeneration() default false;
+
 }
--- a/core-common/src/main/java/org/mapstruct/MapperConfig.java
+++ b/core-common/src/main/java/org/mapstruct/MapperConfig.java
@@ -146,4 +146,22 @@ public @interface MapperConfig {
      * @return strategy how to do null checking
      */
     NullValueCheckStrategy nullValueCheckStrategy() default ON_IMPLICIT_CONVERSION;
+
+    /**
+     * If MapStruct could not find another mapping method or apply an automatic conversion it will try to generate a
+     * sub-mapping method between the two beans. If this property is set to {@code true} MapStruct will not try to
+     * automatically generate sub-mapping methods.
+     * <p>
+     * Can be overridden by {@link Mapper#disableSubMappingMethodsGeneration()}
+     * <p>
+     * Note: If you need to use {@code disableSubMappingMethodsGeneration} please contact the MapStruct team at
+     * <a href="http://mapstruct.org">mapstruct.org</a> or
+     * <a href="https://github.com/mapstruct/mapstruct">github.com/mapstruct/mapstruct</a> to share what problem you
+     * are facing with the automatic sub-mapping generation.
+     *
+     * @return whether the automatic generation of sub-mapping methods is disabled
+     *
+     * @since 1.2
+     */
+    boolean disableSubMappingMethodsGeneration() default false;
 }
--- a/processor/src/test/java/org/mapstruct/ap/test/nestedbeans/ErroneousDisabledViaConfigHouseMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/nestedbeans/ErroneousDisabledViaConfigHouseMapper.java
@@ -1,0 +1,30 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.nestedbeans;
+
+import org.mapstruct.Mapper;
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper(config = DisableConfig.class)
+public interface ErroneousDisabledViaConfigHouseMapper {
+
+    HouseDto houseToHouseDto(House house);
+}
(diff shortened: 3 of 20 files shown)


core-common/src/main/java/org/mapstruct/MapperConfig.java
```diff
@@ -146,4 +146,17 @@ MappingInheritanceStrategy mappingInheritanceStrategy()
...
+     * forged mapping method for the properties. If this property is set to {@code true} MapStruct will not try to
+     * automatically generate forged mapping methods.
+     *
+     * Can be overridden by {@link Mapper#disableForgedMappingMethodsGeneration()}
+     *
+     * @return whether the automatic generation of forged mapping methods is disabled
+     *
+     * @since 1.2.0
+     */
+    boolean disableForgedMappingMethodsGeneration() default false;
```
sjaakd: has the term "forged" leaked into the documentation? I thought it was for internal use only...

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/ErroneousDisabledViaConfigHouseMapper.java
```diff
@@ -0,0 +1,27 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.nestedbeans;
+
+import org.mapstruct.Mapper;
+
+@Mapper(config = ErroneousDisableConfig.class)
```
sjaakd: is the configuration erroneous too?

core-common/src/main/java/org/mapstruct/MapperConfig.java
```diff
@@ -146,4 +146,17 @@ MappingInheritanceStrategy mappingInheritanceStrategy()
      * @return strategy how to do null checking
      */
     NullValueCheckStrategy nullValueCheckStrategy() default ON_IMPLICIT_CONVERSION;
+
+    /**
+     * If MapStruct could not find another mapping method or apply an automatic conversion it will try to generate a
+     * forged mapping method for the properties. If this property is set to {@code true} MapStruct will not try to
```
sjaakd: those properties

sjaakd: perhaps we should make clear somehow that some times of mapping methods are still generated (iterable, map, stream, enum?) This only applies to bean properties, right? @gunnarmorling isn't that kind of confusing?

sjaakd:
@filiphr .. I've reviewed. There are some questions left.. Thanks for the solution.

core-common/src/main/java/org/mapstruct/MapperConfig.java
filiphr: It hasn't. Maybe something else would be better (it is too long as well :)). maybe `disableSubMappingMethodsGeneration`. I am open to suggestions :)

processor/src/test/java/org/mapstruct/ap/test/nestedbeans/ErroneousDisabledViaConfigHouseMapper.java
filiphr: No it isn't. I'll rename it

core-common/src/main/java/org/mapstruct/MapperConfig.java
filiphr: What do you mean those properties? I mean about the `disableForgedMappingMethodsGeneration` property of the config.

And yes we are still generating (iterable, map, stream, enum). The only thing that we are not generating are forged mappings.

filiphr:
Thanks for the review @sjaakd. I think that it would be good if @agudian and @gunnarmorling have a look as well, as this adds to the public API. 

Also I am open for suggestions for the name of the property :)

agudian:
Here I go again 🙂... I still think of this as a temporary thing to unblock users who ran into trouble with the new feature, e.g. because the generated mapping is wrong or actual problems with the mappings would be hidden - but all those problems should be reported to us and we should seek better solutions for those specific cases... Yes, I know you guys probably think differently about it...

But that's why I think we should make it clear that in hopefully one or two Final releases, we're going to remove the switch again. And I would add the disclaimer to the Javadoc (this is temporary, and please report back why you think you need to use it so it can be "fixed")...

WDYT? Perhaps we should have a Hangout this?

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -252,6 +252,13 @@ Supported values are:
 
 If a policy is given for a specific mapper via `@Mapper#unmappedTargetPolicy()`, the value from the annotation takes precedence.
 |`WARN`
+
+|`mapstruct.
+disableForgedMappingMethodsGeneration`
```
agudian: I think having it in the annotations is enough... Gunnar has quite some distast for those processor-options anyway 🙂

core-common/src/main/java/org/mapstruct/MapperConfig.java
sjaakd: iterable, map, stream, enum can also be forged.. hence my confusion...

filiphr:
>But that's why I think we should make it clear that in hopefully one or two Final releases, we're going to remove the switch again. And I would add the disclaimer to the Javadoc (this is temporary, and please report back why you think you need to use it so it can be "fixed")...

You are right, the only use case I see in using is to make sure that when new properties are added to the classes they have to manually map them and also bugs that might be there.

>WDYT? Perhaps we should have a Hangout this?

I agree maybe we should have a Hangout, we haven't had one in a longer time. Who wants to organise it :)?

core-common/src/main/java/org/mapstruct/MapperConfig.java
gunnarmorling: This should be addressed by not using the term "forged" here.

filiphr:
I renamed the property to `disableSubMappingMethodsGeneration` and removed the `forged` from it's Javadoc. Have a look at it again. The processor Option is also out. 

@agudian I just forgot to add the text for contacting us.

How does this sound:

>If you decide to disableSubMappingMethodsGeneration please contact us at [mapstruct.org](http://mapstruct.org) to let us know if there are any problems and if we can improve the functionality of the feature

agudian:
Yeah, coming up with a good notice there isn't easy... 

How about this variant of your proposal?
> Note: If you need to use {@code disableSubMappingMethodsGeneration} please contact the MapStruct team at <a href="http://mapstruct.org/community/getting-help/">mapstruct.org</a> or -link-to-github- to share what problem you are facing with the automatic sub-mapping generation.

filiphr:
Much better :). I'll change it and push the change in few minutes.

sjaakd:
agreed!

filiphr:
Guys even if you all agree on this. Do not merge it yet. I have to add tests for the `Iterable` and `Map` mapping methods. There it does not work.

sjaakd:
> I have to add tests for the Iterable and Map mapping methods. There it does not work.

I'm confused. You mean: if you set `disableSubMappingMethodsGeneratio=true` it will:
1. still generate Iterable, Map and Enum (recently) mapping methods
2. not generate  Iterable, Map and Enum (recently) mapping methods
?

filiphr:
No we will have an internal error. We are going to return a `null` assignment and we are going to wrap it into some setter. Then during code generation there will be an NPE

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -159,4 +159,22 @@
      * @return strategy how to do null checking
      */
     NullValueCheckStrategy nullValueCheckStrategy() default ON_IMPLICIT_CONVERSION;
+
+    /**
+     * If MapStruct could not find another mapping method or apply an automatic conversion it will try to generate a
+     * sub-mapping method between the two beans. If this property is set to {@code true} MapStruct will not try to
+     * automatically generate sub-mapping methods.
+     * <p>
+     * If you decide to {@code disableSubMappingMethodsGeneration} please contact us at
```
gunnarmorling: I would prefer to have this note only in the reference documentation, not the JavaDocs.

filiphr: Currently we have no mention of the automatic sub-mapping in the reference documentation. Do you think that we should add one? Maybe the one from the release post

filiphr:
I've fixed the problem with the `Collection` and `Map` mapping methods not working when automapping is disabled. Now I think the PR is complete, of course details from the Javadoc need to be cleared up

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -159,4 +159,23 @@
...
+     * Can be configured by the {@link MapperConfig#disableSubMappingMethodsGeneration()} as well.
+     * <p>
+     * Note: If you need to use {@code disableSubMappingMethodsGeneration} please contact the MapStruct team at
+     * <a href="http://mapstruct.org">mapstruct.org</a> or
+     * <a href="https://github.com/mapstruct/mapstruct">github.com/mapstruct/mapstruct</a> to share what problem you
+     * are facing with the automatic sub-mapping generation.
+     *
+     * @return whether the automatic generation of sub-mapping methods is disabled
+     *
+     * @since 1.2.0
```
gunnarmorling: Can we just say since "1.2"?

filiphr:
Thanks a lot for the review :)

PR #1156, opened 2017-03-23
#1155 Field accessors should be considered when resolving nested target properties

### LINKED ISSUE #1155: Field mapping does not work for nested target properties
For the example in #1148 if you use only public fields then we are not finding the entire path for the properties

filiphr:
Fixes #1155


filiphr:
Thanks for the review @sjaakd

PR #1158, opened 2017-03-26
#725 Flag the Annotation element with the wrong dateFormat

### LINKED ISSUE #725: Validate value given for date format is valid

filiphr:
Fixes #725 

I wanted to add the column number to our diagnostics. However, that does not work correctly :). 

* The eclipse compiler finds out the column. However, it is never used :). There might be a bug report, but I can't find it. I just looked at their sources
* As for javac it will work in JDK 9, see [JDK-6388543](https://bugs.openjdk.java.net/browse/JDK-6388543)

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethodBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethodBuilder.java
@@ -28,7 +28,7 @@ import java.util.Set;
 import org.mapstruct.ap.internal.model.assignment.Assignment;
 import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.source.ForgedMethod;
-import org.mapstruct.ap.internal.model.source.FormattingParameters;
+import org.mapstruct.ap.internal.model.common.FormattingParameters;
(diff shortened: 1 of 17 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/common/HasFormattingParameters.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.AnnotationValue;
+import javax.lang.model.element.Element;
+
+/**
+ * Formatting parameters for the common subpackage.
+ *
+ * @author Filip Hrisafov
+ */
+public interface HasFormattingParameters {
```
agudian: I don't quite get why we need this interface... Do we need one in the common package? Could we also just move `FormattingParameters` to the common package? My main concern is that `HasFormattingParameters` reads a bit funny and I guess it's hard to come up with a better name... 🙄

filiphr: We actually don't, moving the `FormattingParameters` to the `common` package works. I'll do that. I did it due to some package cycle problems.

filiphr:

Issue #122, opened 2014-02-03
Suggest property name in error message when referring to a non-existent property in @Mapping


gunnarmorling:
The AP raises an error in case a non-existent property is referenced via `@Mapping#source()` or `target()`. It would be awesome to display the name of the most similar existing property in the error message: 

```
"No property 'foubar' exists. Did you mean 'foobar'?"
```

filiphr:
Fixed by #1168

Issue #1153, opened 2017-03-23
Exception when mapping non existing target..


sjaakd:
Internal error in the mapping processor: java.util.NoSuchElementException at
     java.util.ArrayList$Itr.next(ArrayList.java:854)  	at
     org.mapstruct.ap.internal.util.Collections.first(Collections.java:77)  	at
     org.mapstruct.ap.internal.model.NestedTargetPropertyMappingHolder$Builder.groupByTargetReferences(NestedTargetPropertyMappingHolder.java:288)  	at
     org.mapstruct.ap.internal.model.NestedTargetPropertyMappingHolder$Builder.build(NestedTargetPropertyMappingHolder.java:133)  	at
     org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.handleDefinedNestedTargetMapping(BeanMappingMethod.java:359)  	at
     org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.handleDefinedMappings(BeanMappingMethod.java:325)  	at
     org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.build(BeanMappingMethod.java:137)  	at
     org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMappingMethods(MapperCreationProcessor.java:373)  	at
     org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMapper(MapperCreationProcessor.java:152)  	at
     org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:123)  	at
     org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:80)  	at
     org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:280)  	at
     org.mapstruct.ap.MappingProcessor.processMapperTypeElement(MappingProcessor.java:260)  	at
     org.mapstruct.ap.MappingProcessor.processMapperElements(MappingProcessor.java:226)  	at
     org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:162)  	at
     com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:794)  	at
     com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:705)  	at
[...]

PR #1168, opened 2017-04-04
#122 Suggest property name in error message

navpil:
Suggest property name in error message when referring to a non-existent property in @Mapping (issue #122 )

I did some basic functionality. Please check whether this is something we want and suggest other tests if I missed some cases.


filiphr:
@navpil can you please rebase the PR on top of the latest master? Additionally, do you think that you can split it into 2 commits, one that does the actual enhancement and the other one with all the style changes?

Thanks :)

navpil:
I can't do that right now, but I will be available in about a week. I hope we are not in a hurry.

filiphr:
@navpil no worries, we are not in a hurry. In any case there are some other pending PRs that we want to merge before going to Beta3, if we need to release Beta3 earlier we can do it without this and include this in the next version :)

navpil:
@filiphr I rebased on latest master and split into two commits. I had to change the TargetReference because message printing was moved there. Code I added is a little bit dirty, but it works.

processor/src/main/java/org/mapstruct/ap/internal/model/source/TargetReference.java
```diff
@@ -246,9 +247,22 @@ else if ( targetWriteAccessor == null ) {
             }
         }
 
-        private void reportMappingError(Message msg, Object... objects) {
-            messager.printMessage( method.getExecutable(), mapping.getMirror(), mapping.getSourceAnnotationValue(),
-                msg, objects );
+        private void reportMappingError() {
+            if ( errorMessage == Message.BEANMAPPING_UNKNOWN_PROPERTY_IN_RESULTTYPE ) {
```
filiphr: Are you saying that this is dirty? You can theoretically do something like:

```java
Object[] args;

if (errorMessage == Message.BEANMAPPING_UNKNOWN_PROPERTY_IN_RESULT_TYPE) {
    args = new Object[]{ mapping.getTargetName(), mostSimilarProperty() };
}
else {
    args = new Object[]{ mapping.getTargetName() };
}

messager.printMessage( method.getExecutable(), mapping.getMirror(), mapping.getSourceAnnotationValue(), errorMessage, args);
```

filiphr:
@navpil awesome. Yes we changed the location where we report the error. Thanks for rebasing it. I wrote one comment on the location I think you meant that it is dirty.

By the way the way you can also store the error arguments like we do with the `errorMessage` and use the result type of the current `writeAccessor`. The reason for this is that we have nested target properties, and I think that we need to find the most similar word for the nested property and not the result type of the method.

processor/src/main/java/org/mapstruct/ap/internal/model/source/TargetReference.java
navpil: Actually I did not quite like the if-statement itself. It's like we set the error message depending on some conditions in one place (`setErrorMessage()`), but then we again check on error condition message to set the correct arguments in another place (`reportMappingError()`). In theory we could create some ErrorMessage object with the error message with argument list in `setErrorMessage()` or even create a strategy of reporting the error, but I think that in this case it would be an overcomplication. If we can live with this `if` statement we may leave it as is.

navpil:
> By the way the way you can also store the error arguments like we do with the errorMessage and use the result type of the current writeAccessor. The reason for this is that we have nested target properties, and I think that we need to find the most similar word for the nested property and not the result type of the method.

@filiphr Sorry, but I did not quite understand this paragraph. Do you mean that we should use Accessor's `TypeMirror getAccessedType()` instead of `mapping.getTargetName()` in `reportMappingError()` for finding the most similar property? Or how?

filiphr:
@navpil sorry for my bad explanation :). I'll try to explain better.

We have the following example:

```java
public interface ErroneousIssue1153Mapper {

    @Mapping( target = "nestedTarget2.writable2", source = "nestedSource.writable"),
    Target map(Source source);

    class Source {

        public static class NestedSource {
            public String writable;
        }

        public NestedSource nestedSource;
    }

    class Target {

        public static class NestedTarget {
            private String writable;

            public void setWritable(String writable) {
                this.writable = writable;
            }
        }

        private NestedTarget nestedTarget2;

        public NestedTarget getNestedTarget2() {
            return nestedTarget2;
        }

        public void setNestedTarget2(NestedTarget nestedTarget2) {
            this.nestedTarget2 = nestedTarget2;
        }
    }
}
```

The error is that the `writable2` cannot be found in the `NestedTarget`. Currently we report the entire `mapping.getTargetName()` and we don't specify which type it is (there is #1185 for this). Now you check the most similar word between `nestedTarget2.writable2` and the properties from the `Target`. However, correct in this case would be to find the most similar word between `writable2` and the properties of the `NestedTarget`. 

Have a look at the `Issue1153Test` if you want more information, it tests #1153. 

I think the best would be to also adapt the following tests which are testing the messages that you changed as well:
* `ErroneousMappingsTest#shouldFailToGenerateMappings()` - for target and source
* `Issue1153Test#shouldReportErrorsCorrectly()` - for target only
* `Issue1029Test#reportsProperWarningsAndError()` - for target only

I think if you adapt those tests you are going to see the problem.

Maybe you will need to do something like you proposed in your https://github.com/mapstruct/mapstruct/pull/1168#discussion_r115938107, and also pass the `entryNames[i]` and the `nextType` to `setErrorMessage(...)` so you can find the most similar word and use the `propertyReadAccessors` from the `nextType`.

processor/src/main/java/org/mapstruct/ap/internal/model/source/TargetReference.java
filiphr: This is a bit connected to my answer in https://github.com/mapstruct/mapstruct/pull/1168#issuecomment-300857751

navpil:
@filiphr I did not adapt all tests and decided to write a few more of my own tests. Please have a look at tests and the implementation. I am a bit uncertain about correctness of similar name calculation in SourceReference (lines 168-186). It kind of works, but looks a bit error prone to me, especially I don't clearly understand all the logic before it (in method build()).

filiphr:
@navpil no problem for the tests, we can adapt them later on (I think having it on all of the tests will make it work correctly). As for the correctness it looks ok, in the steps before the build the `entryNames` will contain the last valid path.

If you want you can simplify the `for` loops and using `Strings#join` instead, you can also use `Collections#last`, the `Collections` is a MapStruct util as well. However, we can do it later on as well. Just let me know if you want to make other changes to the PR, otherwise, I am going to merge it tomorrow.

navpil:
@filiphr I did not adapt those test, because I like having tests testing one thing, so if functionality changes (for example message format is changed) we need to fix only those tests which tested that particular functionality. I don't rigorously follow the rule, but I usually try.
As for simplifying the for loops and using Collections.last() - that will not work in my case since I actually need those indexes and I don't want to join all strings so I'd better leave it as is. Imagine the following mapper:

    @Mapping(source = "garage.colour.rgb", target = "garage.color.rgb")

Note the British _colour_ spelling, which is wrong in this case. Prefix will be `garage.` and most similar word will be `color`, and the error message will be 
`No property named "garage.colour.rgb" exists in source parameter(s). Did you mean "garage.color"?`

There are two questions here:
1. Do we need tests for this particular case? I tried the case locally and it works.
2. Is the error message actually OK? Note that we don't check for all other properties in the chain (such as `rgb`) and suggest only partial property name.

If we don't need this test and if the message is OK, you may merge the PR.

filiphr:
@navpil I completely agree with:

>@filiphr I did not adapt those test, because I like having tests testing one thing, so if functionality changes (for example message format is changed) we need to fix only those tests which tested that particular functionality. I don't rigorously follow the rule, but I usually try.

However, I would say that tests that use `messageRegExp` in the `@ExpectedCompilationOutcome` are testing the message format. It can happen that those tests are testing some corner cases that are maybe not covered with more simpler tests. Additionally, when we test for a compilation error, we should strive in making sure that all the information that that compilation error emitted should be tested.

As for your questions:

>1. Do we need tests for this particular case? I tried the case locally and it works.

I think that it would be good to have such a test case. We might already have something that tests that, but those tests are not failing as we wrap the `messageRegExp` with `.*`

>2. Is the error message actually OK? Note that we don't check for all other properties in the chain (such as rgb) and suggest only partial property name.

I think that the error is OK. We can't go up the chain as we have nothing to follow. I am not sure if it will be clear to the user on which level the property is wrong, but I think it'll be OK.

As for the simplification, we can do it later on, we can use `subList` or something like that.

navpil:
@filiphr Tests adjusted, new case added

filiphr:
Thanks a lot for the PR @navpil it has been integrated into master with https://github.com/mapstruct/mapstruct/commit/75fa2fb0f70578b7e6ec0fe17b987eff2352891f and https://github.com/mapstruct/mapstruct/commit/5dd0097cc9bf2da28d8dec14c9a6a3321f3a46a9

Issue #1164, opened 2017-03-30
Missing import when using generic on generic element


influence160:
My class **SearchResultDto** contains this attribute
`  private List<GenericRessource<CriteriaDto>> criterias;`

i have a mapper like this 

```
@Mapper(componentModel = "spring", uses = {CriteriaWsMapper.class, .....(other classes)})
public abstract class SearchResultWsMapper extends GenericWsMapper<SearchResult, SearchResultDto> {
....
}
```
this  mapper works fine with mapstructure 1.0.0.Final and generates this file 

```
package mydomain.ws.mappers;

import mydomain.domain.entities.SearchResult;
import mydomain.ws.dto.SearchResultDto;
import javax.annotation.Generated;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.stereotype.Component;

@Generated(
    value = "org.mapstruct.ap.MappingProcessor",
    date = "2017-03-30T16:27:59+0200",
    comments = "version: 1.0.0.Final, compiler: javac, environment: Java 1.8.0_121 (Oracle Corporation)"
)
@Component
public class SearchResultWsMapperImpl extends SearchResultWsMapper {

    @Autowired
    private PieceWsMapper pieceWsMapper;
    @Autowired
    private CapaciteWsMapper capaciteWsMapper;
    @Autowired
    private ArgumentaireHuileWsMapper argumentaireHuileWsMapper;

    @Override
    public SearchResultDto toDto(SearchResult entity) {
        if ( entity == null ) {
            return null;
        }

        SearchResultDto searchResultDto = new SearchResultDto();

        searchResultDto.setCriterias( mapCriterias( entity.getCriterias() ) );
        searchResultDto.setGroupePieces( mapGroup( entity.getGroupePieces() ) );
        searchResultDto.setAfficherAvertissement( entity.isAfficherAvertissement() );
        searchResultDto.setPiecesAssociees( pieceWsMapper.toDtos( entity.getPiecesAssociees() ) );
        searchResultDto.setCapaciteVariantes( capaciteWsMapper.toDtos( entity.getCapaciteVariantes() ) );
        searchResultDto.setArgumentairesHuiles( argumentaireHuileWsMapper.toDtos( entity.getArgumentairesHuiles() ) );

        return searchResultDto;
    }
}

```

when i try to migrate to 1.1.0.Final the generated file is :

```
package mydomain.ws.mappers;

import mydomain.domain.entities.SearchResult;
[...]

PR #1175, opened 2017-04-10
#1164 Make sure we don't do a null check all the time for SetterWrapperForCollections

filiphr:
When using set in the SetterWrapperForCollections do null check only direct assignment or when the NullValueCheckStrategy is ALWAYS.

This one fixed the second part of #1164

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
@@ -1,0 +1,185 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model;
+
+import org.mapstruct.ap.internal.model.assignment.Assignment;
+import org.mapstruct.ap.internal.model.assignment.ExistingInstanceSetterWrapperForCollectionsAndMaps;
+import org.mapstruct.ap.internal.model.assignment.GetterWrapperForCollectionsAndMaps;
+import org.mapstruct.ap.internal.model.assignment.SetterWrapperForCollectionsAndMaps;
+import org.mapstruct.ap.internal.model.assignment.SetterWrapperForCollectionsAndMapsWithNullCheck;
+import org.mapstruct.ap.internal.model.assignment.UpdateWrapper;
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.model.source.Method;
+import org.mapstruct.ap.internal.prism.CollectionMappingStrategyPrism;
+import org.mapstruct.ap.internal.prism.NullValueCheckStrategyPrism;
+import org.mapstruct.ap.internal.util.Message;
+import org.mapstruct.ap.internal.util.accessor.Accessor;
+
+/**
+ * A builder that is used for creating an assignment to a collection.
+ *
+ * The created assignments to the following null checks:
+ * <ul>
+ *     <li>source-null-check - For this the {@link SetterWrapperForCollectionsAndMapsWithNullCheck} is used when a
+ *     direct assignment is done or the {@link org.mapstruct.NullValueCheckStrategy} is
+ *     {@link org.mapstruct.NullValueCheckStrategy#ALWAYS}. It is also done in
+ *     {@link ExistingInstanceSetterWrapperForCollectionsAndMaps} which extends
+ *     {@link SetterWrapperForCollectionsAndMapsWithNullCheck}</li>
+ *     <li>target-null-check - Done in the {@link ExistingInstanceSetterWrapperForCollectionsAndMaps}</li>
+ *     <li>local-var-null-check - Done in {@link ExistingInstanceSetterWrapperForCollectionsAndMaps}, and
+ *     {@link SetterWrapperForCollectionsAndMapsWithNullCheck}</li>
+ * </ul>
+ *
+ * A local-var-null-check is needed in the following cases:
+ *
+ * <ul>
+ *     <li>Presence check with direct assignment - We need a null check before setting, because we use the copy
+ *     constructor</li>
+ *     <li>Presence check for existing instance mapping - We need the null check because we call addAll / putAll.</li>
+ *     <li>No Presence check and direct assignment - We use the copy constructor</li>
+ *     <li>No Presence check and {@link org.mapstruct.NullValueCheckStrategy#ALWAYS} - the user requested one</li>
+ * </ul>
+ *
+ * @author Filip Hrisafov
+ */
+public class CollectionAssignmentBuilder {
+    private MappingBuilderContext ctx;
+    private Method method;
+    private Accessor targetReadAccessor;
+    private Type targetType;
+    private String targetPropertyName;
+    private PropertyMapping.TargetWriteAccessorType targetAccessorType;
+    private Assignment rhs;
+
+    public CollectionAssignmentBuilder mappingBuilderContext(MappingBuilderContext ctx) {
+        this.ctx = ctx;
+        return this;
+    }
+
+    public CollectionAssignmentBuilder method(Method method) {
+        this.method = method;
+        return this;
+    }
+
+    public CollectionAssignmentBuilder targetReadAccessor(Accessor targetReadAccessor) {
+        this.targetReadAccessor = targetReadAccessor;
+        return this;
+    }
+
+    public CollectionAssignmentBuilder targetType(Type targetType) {
+        this.targetType = targetType;
+        return this;
+    }
+
+    public CollectionAssignmentBuilder targetPropertyName(String targetPropertyName) {
+        this.targetPropertyName = targetPropertyName;
+        return this;
+    }
+
+    public CollectionAssignmentBuilder targetAccessorType(PropertyMapping.TargetWriteAccessorType targetAccessorType) {
+        this.targetAccessorType = targetAccessorType;
+        return this;
+    }
+
+    public CollectionAssignmentBuilder rightHandSide(Assignment rhs) {
+        this.rhs = rhs;
+        return this;
+    }
+
+    public Assignment build() {
+        Assignment result = rhs;
+
+        CollectionMappingStrategyPrism cms = method.getMapperConfiguration().getCollectionMappingStrategy();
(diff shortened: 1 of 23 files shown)


sjaakd:
@filiphr .. I did a first small review.. I'll have a better look later this weekend.. It looks good, although I'm not sure whether we should add the explicit setting to null. I remember vaguely that there's some problem there. But I might as well be mistaken.

filiphr:
@sjaakd you commented in my branch and not the PR so I am inline the answers in the comment here.

> is it wise to change this behavior at this point in time? Its not part of this problem right?

Not exactly the same, but in our tests we are expecting that. I also updated the `Domain` there to have empty default `List` and `Set` for the `strings` and `longs`. If we leave the `Domain` like I updated and remove the explicit `null` set then the tests will fail. This is a normal mapping method, not an update method. I am not sure what one should expect here. I think that this is inline with the existing instance mapping where we do the explicit check to `null`.

> funny thing is here: the ExistingInstanceSetterWrapperForCollections.ftl actually uses the getter. So what I originally meant was that this behavior would migrate from the SetterWrapperForCollections to the GetterWrapperForCollections. Not sure whether this is possible. WDYT?

The getter is used to do the update, but we also have an `else` that uses the setter when the target collection that we are updating is `null` and another `else` to set to `null` when it is not a `defaultValueAssignment`, no `sourcePresenceCheckerReference` and no `includeSourceNullCheck`. The `GetterWrapperForCollectionsAndMaps.ftl` does not have those possibilities.


BTW I noticed that there are still some obsolete assignments, like in the case when we have a `presenceCheck`, there we don't need to do a `null` check on the value. I will push changes for that soon.

processor/src/test/java/org/mapstruct/ap/test/bugs/_913/Issue913SetterMapperForCollectionsTest.java
```diff
@@ -63,6 +63,8 @@
      * The null value mapping strategy on type level (Mapper) should generate forged methods for the
      * conversion from string to long that return null in the entire mapper, so also for the forged
      * mapper. Note the default NVMS is RETURN_NULL.
+     *
+     * The result of the forged methods should be applied directly to the target object.
```
sjaakd: Don't know were to place this comment. Wanted to put it on the test case names :smile: .  Initially I organized the testcases around their mappers. That's probably not valid anymore with the advent of the new mapper.

processor/src/test/java/org/mapstruct/ap/test/bugs/_913/Issue913SetterMapperForCollectionsTest.java
```diff
@@ -173,6 +175,7 @@ public void shouldReturnDefaultForNvmsReturnDefaultForUpdate() {
         Dto dto = new Dto();
         Domain domain = new Domain();
         Set<Long> longIn = new HashSet<Long>();
+        longIn.add( 10L );
```
sjaakd: is this a testcase improvement or did I miss something? (Adding values to the set).

processor/src/test/resources/fixtures/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNvmsDefaultMapperImpl.java
```diff
@@ -39,18 +39,15 @@ public Domain create(Dto source) {
         Domain domain = new Domain();
 
         if ( source != null ) {
-            Set<Long> set = stringListToLongSet( source.getStringsInitialized() );
```
sjaakd: Nice thing about fixtures is that you can give comments on the generated code :smile: ..Somewhere I made a remark that we should distinguish between: 
* source-null-check, 
* target-null-check and 
* local-var-null-check (or something along that line). 

The latter being the result of a mapping method call or type conversion call. Do we still have that scenario somewhere (e.g. is that `handleLocalVarNullCheck`) ? I guess, if we have a type conversion (after mapping method call)  we still should do null checks and we should also do a source-null-check before we hand the source variable to a type-conversion.

 Ergo should my original  comment be addressed somewhere in the templates or code?

processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
```diff
@@ -0,0 +1,163 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
```
sjaakd: Think the builder is a good idea, But should we express somehow in the name of the class that this belongs to the property mapping ?

processor/src/test/java/org/mapstruct/ap/test/bugs/_913/Issue913SetterMapperForCollectionsTest.java
filiphr: I was also not sure about the location for this. Maybe I need to go through all of the test methods and validate that the Javadoc makes sense to the latest implementation.

processor/src/test/java/org/mapstruct/ap/test/bugs/_913/Issue913SetterMapperForCollectionsTest.java
filiphr: It is just an improvement. You didn't miss anything, the test passed without changing anything. I just added this, to be sure that clear is called. The reason is that we check if the `longs` are empty and if they are empty from the start, you don't know if it is because empty was added or because of before :)

processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
filiphr: Well we could theoretically move this to `org.mapstruct.ap.internal.model.property.mapping` package or maybe name it `PropertyMappingCollectionAssignmentBuilder`. What do you prefer?

Theoretically, I would refactor the entire assignment into it's own Builder that will invoke the needed builders for: `assignToCollection`, `assignToArray` and `assignToPlain`. `PropertyMapping` is getting big slowly :)

sjaakd: I was thinking of using a chain of responsibility. Each chain element is a current wrapper and triggering on its own criteria (the current if-then-elses).

processor/src/test/resources/fixtures/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNvmsDefaultMapperImpl.java
filiphr: I only removed the null check from the default NCVS after an invocation of a mapping method. I didn't change the logic on other places for when a null check is done or not. That is still being handled in the `handleLocalVarNullCheck`. What I did is just omit the explicit set to a variable when we were not doing a null check. The decision logic is now in Java instead of the tempaltes.

>Somewhere I made a remark that we should distinguish between:

Distinguish where?
Well the names of the wrappers are somehow depicting that:

* source-null-check - For this the `SetterWrapperForCollectionsAndMapsWithNullCheck` is used when a direct assignment is done or the NVCS is `ALWAYS`. It is also done in `ExistingInstanceSetterWrapperForCollectionsAndMaps` which extends `SetterWrapperForCollectionsAndMapsWithNullCheck`. Maybe we need to call it `SetterWrapperForCollectionsAndMapsWithSourceNullCheck`
* target-null-check - Done in the `ExistingInstanceSetterWrapperForCollectionsAndMaps`
* local-var-null-check - Done in `ExistingInstanceSetterWrapperForCollectionsAndMaps` and `SetterWrapperForCollectionsAndMapsWithSourceNullCheck`

From my point of view a local var null check is needed when:
* Presence check with direct assignment - We need a null check before setting, because we use the copy constructor
* Presence check for existing instance mapping - We need the null check because we call `addAll` / `putAll`.
* No Presence check and direct assignment - We use the copy constructor
* No Presence check and NCVS `ALWAYS` - the user requested one

All other cases we don't need one.

processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
filiphr: That is too a possible way to go. I am fine with any approach, if it reduces the complexity of the class :)

filiphr:
>@filiphr .. Nice / huge job.

Thanks :).

I replied back to the comments. 

> I was thinking though, whether we should have local-var-null checks on type conversions and if we have testcases for that

What exactly do you mean with this? We have type conversions in the tests. We map `List<String>` from `Dto` to a `Set<Long>` from Domain.

> The fixtures seem to concern calling mapping methods only, before calling the target accessor. So perhaps that's an omission. WDYT?

I don't think that I understood what you mean with this, sorry. What do you think that is missing from the fixtures?

sjaakd:
> I don't think that I understood what you mean with this, sorry. What do you think that is missing from the fixtures?

 I've checked the fixtures more thoroughly.. Seems everything is there. To clarify: I was looking for instances were we do a local-var-null-check. Like here (`DomainDtoWithNvmsNullMapperImpl`, line 109) for  instance:

```java
        if ( target.getLongsInitialized() != null ) {
            Set<Long> set1 = stringListToLongSet( source.getStringsInitialized() );
            if ( set1 != null ) {
                target.getLongsInitialized().clear();
                target.getLongsInitialized().addAll( set1 );
            }
            else {
                target.setLongsInitialized( null );
            }
        }
        else {
            Set<Long> set1 = stringListToLongSet( source.getStringsInitialized() );
            if ( set1 != null ) {
                target.setLongsInitialized( set1 );
            }
        }

```

There are too many scenarios 😢 ..

processor/src/test/resources/fixtures/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNvmsDefaultMapperImpl.java
sjaakd: Ok.. shall we add everything beneath `Distinguish were?`  to the java doc of the new builder :smile:

processor/src/test/java/org/mapstruct/ap/test/bugs/_913/Issue913SetterMapperForCollectionsTest.java
sjaakd: Or take that in an improvement PR. Perhaps we missed some combinations as well.. When I wrote the testcases I tried to follow a structured approach.. But there are most likely some caveats.

processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
sjaakd: > That is too a possible way to go. I am fine with any approach, if it reduces the complexity of the class :)

I've scheduled that for some refactoring (rainy day) :smile:

processor/src/test/resources/fixtures/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNvmsDefaultMapperImpl.java
filiphr: That is a good idea. I will put the text in the builder :)

processor/src/test/java/org/mapstruct/ap/test/bugs/_913/Issue913SetterMapperForCollectionsTest.java
filiphr: That is a good idea. I can make a separate PR only with the Javadoc changes of the test methods

filiphr:
@sjaakd thanks for the review. I am going to move the list in the javadoc and will merge afterwards.

PR #1183, opened 2017-04-21
#1154 Add SPI for excluding types/elements from automatic sub-mapping generation

### LINKED ISSUE #1154: Do not try to create mappings for objects in the java package
When we generate missing mappings we shouldn't try to go into objects from the java package.

For example if we have:

```java
public class MyDate {

    private String someProperty;

    //getters and setters omitted for brevity
}
```
```java
public class Entity {
    
    private MyDate date;

    //getters and setters omitted for brevity
}
```

```java
public class EntityDto {
    
    private Date date;

    //getters and setters omitted for brevity
}
```

```java
@Mapper
public interface MyMapper {

    EntityDto map(Entity entity);
}
```

We are going to generate:

```java
protected Date myDateToDate(MyDate myDate) {
        if ( myDate == null ) {
            return null;
        }

        Date date = new Date();

        return date;
}
````

In my opinion we shouldn't try to instantiate and create objects from the java package. We also then continue and try to see if there are properties that can be mapped between `Date` and `MyDate`.

### PR DESCRIPTION
**Pull request by filiphr**
Fixes #1154.

With this PR we are adding an SPI that can be used to exclude `TypeElement`(s) from the automapping, i.e. MapStruct will not try to generate an automatic sub-mapping method if a type is excluded.

The default implementation excludes packages from the `java` and `javax` sub-packages

### DIFF
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -3,6 +3,8 @@
 :toc: right
 :sectanchors:
 :Author: Gunnar Morling, Andreas Gudian, Sjaak Derksen, Filip Hrisafov and the MapStruct community
+:processor-dir: ../../../../processor
+:processor-ap-test: {processor-dir}/src/test/java/org/mapstruct/ap/test
 
 [[Preface]]
 == Preface
@@ -2463,4 +2465,59 @@ To use a custom SPI implementation, it must be located in a seperate .jar file t
 [TIP]
 Fore more details: There's the above example is present in our our examples repository (https://github.com/mapstruct/mapstruct-examples).
 
+[mapping-exclusion-provider]
+=== Mapping Exclusion Provider
 
+MapStruct offers the possibility to override the `MappingExclusionProvider` via the Service Provider Interface (SPI).
+A nice example is to not allow MapStruct to create an automatic sub-mapping for a certain type,
+i.e. MapStruct will not try to generate an automatic sub-mapping method for an excluded type.
+
+[NOTE]
+====
+The `DefaultMappingExclusionProvider` will exclude all types under the `java` or `javax` packages.
+This means that MapStruct will not try to generate an automatic sub-mapping method between some custom type and some java type.
+====
+
+.Source object
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+include::{processor-ap-test}/nestedbeans/exclusions/custom/Source.java[tag=documentation]
+----
+====
+
+.Target object
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+include::{processor-ap-test}/nestedbeans/exclusions/custom/Target.java[tag=documentation]
+----
+====
+
+.Mapper definition
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+include::{processor-ap-test}/nestedbeans/exclusions/custom/ErroneousCustomExclusionMapper.java[tag=documentation]
+----
+====
+
+We want to exclude the `NestedTarget` from the automatic sub-mapping method generation from MapStruct.
+
+.CustomMappingExclusionProvider
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+include::{processor-ap-test}/nestedbeans/exclusions/custom/CustomMappingExclusionProvider.java[tag=documentation]
+----
+====
+
+To use a custom SPI implementation, it must be located in a seperate .jar file
+together with the file `META-INF/services/org.mapstruct.ap.spi.MappingExclusionProvider` with the fully qualified name of your custom implementation as content
+(e.g. `org.mapstruct.example.CustomMappingExclusionProvider`).
+This .jar file needs to be added to the annotation processor classpath
+(i.e. add it next to the place where you added the mapstruct-processor jar).
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -817,7 +817,7 @@ When generating the implementation of a mapping method, MapStruct will apply the
 * If source and target attribute have the same type, the value will be simply copied from source to target. If the attribute is a collection (e.g. a `List`) a copy of the collection will be set into the target attribute.
 * If source and target attribute type differ, check whether there is a another mapping method which has the type of the source attribute as parameter type and the type of the target attribute as return type. If such a method exists it will be invoked in the generated mapping implementation.
 * If no such method exists MapStruct will look whether a built-in conversion for the source and target type of the attribute exists. If this is the case, the generated mapping code will apply this conversion.
-* If no such method was found MapStruct will try to generate an internal method that will do the mapping between the source and target attributes
+* If no such method was found MapStruct will try to generate an automatic sub-mapping method that will do the mapping between the source and target attributes
 * If MapStruct could not create a name based mapping method an error will be raised at build time, indicating the non-mappable attribute and its path.
 
 include::controlling-nested-bean-mappings.asciidoc[]
@@ -2460,7 +2460,7 @@ public class CustomAccessorNamingStrategy extends DefaultAccessorNamingStrategy 
 ====
 The `CustomAccessorNamingStrategy` makes use of the `DefaultAccessorNamingStrategy` (also available in mapstruct-processor) and relies on that class to leave most of the default behaviour unchanged.
 
-To use a custom SPI implementation, it must be located in a seperate .jar file together with the file `META-INF/services/org.mapstruct.ap.spi.AccessorNamingStrategy` with the fully qualified name of your custom implementation as content (e.g. `org.mapstruct.example.CustomAccessorNamingStrategy`). This .jar file needs to be added to the annotation processor classpath (i.e. add it next to the place where you added the mapstruct-processor jar).
+To use a custom SPI implementation, it must be located in a separate JAR file together with the file `META-INF/services/org.mapstruct.ap.spi.AccessorNamingStrategy` with the fully qualified name of your custom implementation as content (e.g. `org.mapstruct.example.CustomAccessorNamingStrategy`). This JAR file needs to be added to the annotation processor classpath (i.e. add it next to the place where you added the mapstruct-processor jar).
 
 [TIP]
 Fore more details: There's the above example is present in our our examples repository (https://github.com/mapstruct/mapstruct-examples).
@@ -2475,7 +2475,7 @@ i.e. MapStruct will not try to generate an automatic sub-mapping method for an e
 [NOTE]
 ====
 The `DefaultMappingExclusionProvider` will exclude all types under the `java` or `javax` packages.
-This means that MapStruct will not try to generate an automatic sub-mapping method between some custom type and some java type.
+This means that MapStruct will not try to generate an automatic sub-mapping method between some custom type and some type declared in the Java class library.
 ====
 
 .Source object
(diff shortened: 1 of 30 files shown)

### CODE REVIEW

File: processor/src/main/java/org/mapstruct/ap/spi/DefaultMappingExclusionProvider.java [hunk 1]
```diff
@@ -0,0 +1,47 @@
...
+package org.mapstruct.ap.spi;
+
+import java.util.regex.Pattern;
+import javax.lang.model.element.Name;
+import javax.lang.model.element.TypeElement;
+
+/**
+ * The default implementation of the {@link MappingExclusionProvider} service provider interface.
+ *
+ * With the default implementation, MapStruct will exclude all classes under the under the {@code java} and {@code
```
sjaakd: Mapstruct will not consider classes in the Java / Javax package as source / target for an automatic sub mapping. (Or something along that line).

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
```diff
@@ -50,7 +50,27 @@ public B method(Method sourceMethod) {
...
+     * Checks if MapStruct is allowed to generate an automatic sub-mapping between {@code sourceType} and @{code
+     * targetType}.
+     * This will evaluate to {@code true}, when:
+     * <li>
+     * <ul>Automatic sub-mapping methods generation is not disabled</ul>
+     * <ul>MapStruct is allowed to generate an automatic sub-mapping between the {@code sourceType} and {@code
+     * targetType}</ul>
+     * </li>
+     *
+     * @param sourceType the source type that we want to generate a sub-mapping from
```
sjaakd: I'm not in favor of using 'we'.. Perhaps: `candidate source type to generate a sub-mapping from`?

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
```diff
@@ -50,7 +50,27 @@ public B method(Method sourceMethod) {
         return myself;
     }
 
-    boolean isDisableSubMappingMethodsGeneration() {
+    /**
+     * Checks if MapStruct is allowed to generate an automatic sub-mapping between {@code sourceType} and @{code
```
sjaakd: have we explained the term `sub-mapping` anywhere already (documentation).. Can we refer to that.?

filiphr: Well we added `disableSubMappingMethodsGeneration` to the `@Mapper` and `@MapperConfig` with #1146. We don't have it in the documentation

filiphr: Do you think that is enough?

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
```diff
@@ -60,8 +60,8 @@ public B method(Method sourceMethod) {
      * targetType}</ul>
      * </li>
      *
-     * @param sourceType the source type that we want to generate a sub-mapping from
-     * @param targetType the target type that we want to generate a sub-mapping for
+     * @param sourceType candidate source type to generate a sub-mapping from
+     * @param targetType candidate target type to to generate a sub-mapping for
```
sjaakd: to to?

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
sjaakd: yep.

agudian:
Looks good, Filip! 
The exclusion of java/javax types should be added to the docs as well, I guess. Together with a section on automatic sub-type mapping in general, if we don't have that yet 😉

filiphr:
We don't have a nice section where we explain it well. I can add one and put the exclusion of `java` / `javax` types in it, and then update this PR.

gunnarmorling:
@filiphr anything missing for this one? Or can we merge it?

processor/src/main/java/org/mapstruct/ap/spi/DefaultMappingExclusionProvider.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.regex.Pattern;
+import javax.lang.model.element.Name;
+import javax.lang.model.element.TypeElement;
+
+/**
+ * The default implementation of the {@link MappingExclusionProvider} service provider interface.
+ *
+ * With the default implementation, MapStruct will not consider classes in the {@code java} and {@code javax} package
+ * as source / target for an automatic sub-mapping. The only exception is the {@link java.util.Collection},
```
gunnarmorling: Where does this exception reflect in the code? I cannot seem to find it in this implemention. What about `Iterable` btw.? Shouldn't this entire concept only apply to bean mapping methods (as opposed to collection mapping methods) anyways?

processor/src/main/java/org/mapstruct/ap/spi/MappingExclusionProvider.java
```diff
@@ -0,0 +1,42 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.spi;
+
+import javax.lang.model.element.TypeElement;
+
+/**
+ * A service provider interface that is used to control if MapStruct is allowed to generate automatic sub-mapping for
```
gunnarmorling: I think "sub-mapping" should shortly be explained here, so people looking at this interface don't need out-of-band knowledge to grasp this contract.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -2463,4 +2465,59 @@ To use a custom SPI implementation, it must be located in a seperate .jar file t
 [TIP]
 Fore more details: There's the above example is present in our our examples repository (https://github.com/mapstruct/mapstruct-examples).
 
+[mapping-exclusion-provider]
+=== Mapping Exclusion Provider
 
+MapStruct offers the possibility to override the `MappingExclusionProvider` via the Service Provider Interface (SPI).
+A nice example is to not allow MapStruct to create an automatic sub-mapping for a certain type,
```
gunnarmorling: Same remark as below, I think we should explain the term "sub-mapping" on first usage. It's not self-explanatory really.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -2463,4 +2465,59 @@ To use a custom SPI implementation, it must be located in a seperate .jar file t
...
+i.e. MapStruct will not try to generate an automatic sub-mapping method for an excluded type.
+
+[NOTE]
+====
+The `DefaultMappingExclusionProvider` will exclude all types under the `java` or `javax` packages.
+This means that MapStruct will not try to generate an automatic sub-mapping method between some custom type and some java type.
```
gunnarmorling: s/java type/type declared in the Java class library/

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -2463,4 +2465,59 @@ To use a custom SPI implementation, it must be located in a seperate .jar file t
...
+.Mapper definition
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+include::{processor-ap-test}/nestedbeans/exclusions/custom/ErroneousCustomExclusionMapper.java[tag=documentation]
+----
+====
+
+We want to exclude the `NestedTarget` from the automatic sub-mapping method generation from MapStruct.
```
gunnarmorling: s/from MapStruct//

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -2463,4 +2465,59 @@ To use a custom SPI implementation, it must be located in a seperate .jar file t
...
+.CustomMappingExclusionProvider
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+include::{processor-ap-test}/nestedbeans/exclusions/custom/CustomMappingExclusionProvider.java[tag=documentation]
+----
+====
+
+To use a custom SPI implementation, it must be located in a seperate .jar file
```
gunnarmorling: s/seperate .jar/separate JAR/

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -2463,4 +2465,59 @@ To use a custom SPI implementation, it must be located in a seperate .jar file t
...
+together with the file `META-INF/services/org.mapstruct.ap.spi.MappingExclusionProvider` with the fully qualified name of your custom implementation as content
+(e.g. `org.mapstruct.example.CustomMappingExclusionProvider`).
+This .jar file needs to be added to the annotation processor classpath
```
gunnarmorling: s/.jar/JAR/

processor/src/main/java/org/mapstruct/ap/spi/DefaultMappingExclusionProvider.java
gunnarmorling: Should we really expose our default implementation in the SPI btw.? Do you think it helps people much as opposed to implementing the interface directly themselves? I think I'd prefer to keep it internal unless there's a good case for it.

processor/src/main/java/org/mapstruct/ap/spi/MappingExclusionProvider.java
filiphr: I'll write more about sub-mapping here.

processor/src/main/java/org/mapstruct/ap/spi/DefaultMappingExclusionProvider.java
filiphr: You are right, we are in any case not descending into `Collection` and the likes. I'll remove that.

As for the default implementation, I followed the same approach as the `AccessorNamingStrategy`. However, you are right, maybe we should move it internally, if people want to use this SPI they can implement the interface and add the Java type it is not that complex :).

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
filiphr: I used the same text that we have for the [`CustomAccessorNamingStrategy`](http://mapstruct.org/documentation/stable/reference/html/#_custom_accessor_naming_strategy) in our reference documentation.  Should I change those places as well?

btw. I just saw that we have `sepErate` in the reference documentation as well :)

filiphr:
@gunnarmorling  regarding:

>@filiphr anything missing for this one? Or can we merge it?

Only your comments are left :). I'll update the PR and if you are OK with the additions to the Javadoc we can merge it :)

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
filiphr: In the [`Mapping object references`](http://mapstruct.org/documentation/dev/reference/html/#mapping-object-references) section we have the routine that MapStruct uses to create a mapping. The last step before raising an error is:

> * If no such method was found MapStruct will try to generate an internal method that will do the mapping between the source and target attributes

Will it suffice if we have:

* If no such method was found MapStruct will try to generate an automatic sub-mapping method that will do the mapping between the source and target attributes

instead? Or you think we need more context.  I can also write the routine steps in the `MappingExclusionProvider` as well, so when someone sees the interface they can have more content.

processor/src/main/java/org/mapstruct/ap/spi/MappingExclusionProvider.java
```diff
@@ -0,0 +1,42 @@
...
+ * a given {@link TypeElement}.
+ *
+ * @author Filip Hrisafov
+ * @since 1.2
+ */
+public interface MappingExclusionProvider {
```
gunnarmorling: Let's mark it as "Experimental" or "Incubating", allowing us to change it in future versions.

filiphr:
Thanks everyone for the review :)

PR #1186, opened 2017-04-23
Use anchors for the sections

filiphr:
Uses `:sectanchors:` to add anchors to the sections. See [this](http://asciidoctor.org/docs/user-manual/#anchors) for more information.

And ads a small change to the authors :)


filiphr:
By the way if there is no anchor defined for the section it creates one, see the ones for section 2.1 for example.

Do we want to add defined anchors to all the sections?

Issue #777, opened 2016-03-10
Set initial capacity for new collection element in collection mapping


ugrave:
In collection mappings the new created target collection shout be initialize with a capacity of the source collection. For collections with many elements a lot of re-sizing is done.

Example
Mapper:
`List<DataDto> map(List<Data> source);`
Generated Mapper:

```
List<DataDto> list = new ArrayList<DataDto>( source.size() /* + growCapacity */ );
 for ( Data data : source ) {
   list.add( map( data ) );
}
```

Maybe its also a good point to have an optionn to make the result list larger then the source list(see growCapacity attribute)
This could be made configurable (for example with the IterableMapping annotation).

agudian:
I'm not sure if that growCapacity thing is something we should add as an option, but properly initializing the with the expected size for certain implementations (`ArrayList`, `HashSet`, `HashMap`) would surely make sense.
For the growCapacity, I could imagine that we allow factory methods to have the source size being passed as argument, so users could implement what they need there.

You already know what's coming next: if you find the time, we'd gladly apply a patch of yours... :+1:

gunnarmorling:
+1 for specifying the right size (though it's not super-easy for sets/maps). Regarding a parameter for grow capacity, I side with @agudian: It seems a bit too specific.

Any help for using the correct size is very welcome!

filiphr:
The initial capacity constructor is used where needed. I will close this issue. I think that the part about the `growCapacity` can go in a new issue. 

Just to clarify for `Map`(s) and `Set`(s) we use: `Math.max((int) (c.size()/.75f) + 1, 16)`, something similar to what Guava does and what the Java Collections use when creating a new `HashMap` / `HashSet` from a collection

PR #1192, opened 2017-05-04
#777 Use initial capacity when creation new Collection / Map

filiphr:
This one solves the first part of #777 (using the constructors with initial capacity) when instantiating the new Collections / Maps. For the second part we need to discuss.

This also improves the performance of the generated code. The performance is improved because there the underlying `Arrays` in the `Collection`s / `Map`s will not need to be increased during the mapping.

### DIFF
--- a/processor/src/test/resources/fixtures/org/mapstruct/ap/test/collection/defaultimplementation/SourceTargetMapperImpl.java
+++ b/processor/src/test/resources/fixtures/org/mapstruct/ap/test/collection/defaultimplementation/SourceTargetMapperImpl.java
@@ -1,0 +1,272 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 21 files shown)


processor/src/test/resources/fixtures/org/mapstruct/ap/test/collection/defaultimplementation/SourceTargetMapperImpl.java
```diff
@@ -0,0 +1,272 @@
...
+import java.util.concurrent.ConcurrentNavigableMap;
+import java.util.concurrent.ConcurrentSkipListMap;
+import javax.annotation.Generated;
+
+@Generated(
+    value = "org.mapstruct.ap.MappingProcessor",
+    date = "2017-05-06T00:20:29+0200",
+    comments = "version: , compiler: javac, environment: Java 1.8.0_131 (Oracle Corporation)"
+)
+public class SourceTargetMapperImpl implements SourceTargetMapper {
```
filiphr: I added this fixture because it contains the defaults for all collections/maps that we have. And I think that we should have one place where we test the way we instantiate them. If you think that it is not needed, I can remove it

filiphr:

agudian:
I wonder why travis didn't build the PR... 🤔

filiphr:
oh sorry I didn't notice that. It did the first one. I think that it missed the second one, because the same commit is part of my example PR for coveralls on my branch 😄

PR #1195, opened 2017-05-07
Coverage analysis setup

### LINKED ISSUE #1194: Setup coverage analysis

filiphr:
Fixes #1194 

I decided to integrate with [codecov.io](https://codecov.io/). I think that they have better integration with github and much better GUI. Also with them you can easily configure the bot to use our own bot. If we want to do that it can be configured like [this](https://docs.codecov.io/docs/team-bot)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        # basic
+        threshold: 0.05
+comment: off
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ jdk:
   - oraclejdk8
 install: true
 script: mvn clean install -DprocessorIntegrationTest.toolchainsFile=etc/toolchains-travis-jenkins.xml -B -V
+after_success:
+  - mvn jacoco:report && bash <(curl -s https://codecov.io/bash) || echo "Codecov did not collect coverage reports"
 
 sudo: false
 cache:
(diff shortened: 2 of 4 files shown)


gunnarmorling:
@filiphr seems this PR failed its own check as done by codecov ;) ?

filiphr:
@gunnarmorling the `codecov/project` check has failed because there is no previous analysis. I can try an push one analysis from my machine to codecov and then see how it will be reported. Is that OK?

gunnarmorling:
@filiphr We can also integrate it and see how it works for future PRs. What's the difference between "project" and "patch"? And will there always that huge comment with the colored boxes be added? I think my overarching question is this: is this a really useful feature or just noise that gets added to each PR and which needs good knowledge to interpret it? Personally I probably wouldn't need it (looking into coverage now and then is fine for me), but if you think it's a valuable feature we'll benefit from, let's do it.

filiphr:
@gunnarmorling `codecov/project` is the coverage on the entire code base when the PR would be merged, and `codecov/patch` is the coverage only on the new diff, i.e. the changes in the PR. 

As for the comment, we can customize what comes into it, or even completely remove it. It doesn't need to be there :). 

> Personally I probably wouldn't need it (looking into coverage now and then is fine for me), but if you think it's a valuable feature we'll benefit from, let's do it.

What wouldn't you need? The comment or the entire coverage check for the PR? Sorry I didn't understand you well

gunnarmorling:
As discussed, let's remove the commenting and then go for it. Thanks, @filiphr!

filiphr:
Thanks for the review

Issue #1198, opened 2017-05-10
Support @Named variables to address mapper inheritance when super is not located in same jar


jmuis:
Hi, 

We have a case where we need to have a mapper interface with 2 different implementations, which implementation will be chosen runtime is basically dependent upon the way we need to retrieve additional data for the mapper (target property uses a java expression to fetch mapping data).

Because these 2 implementations resist in 2 different jars and the mapper interface in a 3rd jar reflection the actual parameter names get lots through the java compiler and end up as arg0, arg1, arg2, etc.  Example mapper method:   void mapMyData(Type1 arg0, Type2 arg2, ....) where the original interface specifies mapMyData(Type1 type1, Type2 type2, ... )

You will now have to use arg0, arg1, etc. in your java expression or you will get a compile error.
You can avoid this when using java 8 and adding parameter names compiler options, but we would like to avoid this very much. 

Project structure: 
- api
  + interface my-mapper-interface(...) {}
- implementation-variant-1
  +@Mapping(target="...", expression="java(...)") 
     interface mapper(...) extends my-mapper-interface {}
- implementation-variant-2
     @Mapping(target="...", expression="java(...)") 
  + interface mapper(...) extends my-mapper-interface {}

Possible solution (will add a merge request later) is to use @Named annotations on the parameter, which would then be used by the annotation processor to not use the java generated parameter name (arg0) but the value of the @Named annotation.

filiphr:
As I wrote in the PR #1199 I find this feature really interesting and useful. The implementation looks pretty simple as well.

I would only prefer using some other annotation like `@Param` or something similar instead of the `@Named` annotation. @gunnarmorling, @agudian, @sjaakd what do you think? 

@jmuis is using a different annotation acceptable for you?

agudian:
Hmm, interesting. We talked about a similar problem in https://github.com/mapstruct/mapstruct/pull/847

filiphr:
@agudian where is this talked about in the linked PR? Sorry, I couldn't find it :(.

Are you talking about this:

>For instance my particular use case would most likely be with jackson annotated constructor fields. > In this case I would like to retrieve the jackson annotations and somehow create the constructor
mapping from them

If that is the case, it is similar, but not that much. I also didn't understand what was the consensus there. Did you agree on something or not?

gunnarmorling:
[...]

PR #1199, opened 2017-05-10
#1198: Support @Named variables to address mapper inheritance when su…

jmuis:
Just noticed imports ordering has changed, but please have a look at the general idea, more than happy to clean it up, add documentation, etc. if suggestion / improvement will be accepted at all.

### DIFF
--- a/core-common/src/main/java/org/mapstruct/Named.java
+++ b/core-common/src/main/java/org/mapstruct/Named.java
@@ -89,7 +89,7 @@ import java.lang.annotation.Target;
  * @see MapMapping#keyQualifiedByName()
  * @see MapMapping#valueQualifiedByName()
  */
-@Target( { ElementType.TYPE, ElementType.METHOD } )
+@Target( { ElementType.TYPE, ElementType.METHOD, ElementType.PARAMETER } )
 @Retention( RetentionPolicy.CLASS )
 @Qualifier
 public @interface Named {
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
@@ -18,16 +18,16 @@
  */
 package org.mapstruct.ap.internal.model.common;
 
+import org.mapstruct.ap.internal.prism.ContextPrism;
+import org.mapstruct.ap.internal.prism.MappingTargetPrism;
+import org.mapstruct.ap.internal.prism.NamedPrism;
+import org.mapstruct.ap.internal.prism.TargetTypePrism;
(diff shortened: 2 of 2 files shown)


filiphr:
@jmuis thanks for the PR. I think that the idea is really good. Especially to support cross jar compilations.

The implementation looks really simple :). Apart from adding documentation and tests. I think we'll need an integration test with multiple modules for this to be tested as well.

I also think that we need to use a different Annotation. I don't think that the `@Named` is correct here. Maybe something like `@Param`.  @gunnarmorling, @agudian, @sjaakd what do you think?

jmuis:
Hi, 

Thanks for your quick & positive feedback. 
We had some discussion around @Named as well, but actually it does make sense IMHO, as you are naming a parameter. Obviously a new annotation is possible as well, technically I have nothing against it. It might just be a little confusing if you already have something around which is called @Named.
 
How would you do unit testing? Create a 'dummy' project just containing base mappers?

filiphr:
@jmuis let's see what the others say about the annotation. I think that it will confuse user, especially if you look at the Javadoc of `@Named`, it is meant to be used as a `@Qualifier`, it is even annotated with that. It is a way to give a name to a method that can be then used in `@Mapping#qualifiedByName()` and the other variants. That is the only reason why I think that it might confuse users to use that.

As for the unit testing, you don't have to create a "dummy" project. We have our own runner `AnnotationProcessorTestRunner` that invokes the javac compiler and the eclipse compiler during tests. just have a look at any test to see how it works. Apart from the test, you will need to create classes and mapper that will map them. 

Apart from the "unit" test. I will suggest to create an integration test as well. For that you can have a look at the tests in the `integrationtest` module. Those are a bit more complex 😄. You will need to create a test and use the `ProcessorSuiteRunner` to say where the folder is. For this you will need to create a "dummy" maven project. Have a look at the `NamingStrategyTest`. I would suggest running that test from a command line via maven 😄.

If something is not clear, just let me know and I can help you out.

agudian:
Hmm, doesn't this work with JDK 8 out of the box? Or do you need to pass an additional compiler arg when compiling? Ah, found it: http://docs.oracle.com/javase/8/docs/technotes/tools/windows/javac.html#options (it speaks about making the parameter name available for reflection, but I would guess that it could do the trick for us as well)... Did you try that?

agudian:
> Hmm, doesn't this work with JDK 8 out of the box?...... Did you try that?

Sorry, you wrote that in the issue already.

gunnarmorling:
Not going to merge this as discussed on the issue itself. I think Java 8 parameter names should be the way to go and we shouldn't provide an alternative to that standard mechanism.

Issue #1197, opened 2017-05-09
Optimize understandability of our source code


sjaakd:
It seems we are dragging around the following `Map<String, List<Mapping>>`, were the key identifies a target name and the list of mappings is associated with that target.

However: the only place were we seem to use that is in the 'old-style' enum mapping (before `@ValueMapping`). Goal of this issue is to get rid of this relict. It will make the code a whole lot easier to understand.

sjaakd:
As a suggestion, perhaps we can rewrite the @Mapping to `@ValueMapping` in an earlier phase (e.g. when creating the `SourceMethod`)

sjaakd:
Related to #1301, #1307

PR #1202, opened 2017-05-17
#1129 Do not use equals and hashCode from TypeMirror

### LINKED ISSUE #1129: Don't use equals and hashCode from TypeMirror
This is connected to the fix to #1127. There is [this](https://bugs.openjdk.java.net/browse/JDK-7159498) JDK bug report that might be causing java lang `ClassType` that are actually same to be different. 

According to the javadoc of [`TypeMirror#equals()`](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/type/TypeMirror.html#equals-java.lang.Object-) one should use [`Types#isSameType()`](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/util/Types.html#isSameType-javax.lang.model.type.TypeMirror-javax.lang.model.type.TypeMirror-) for semantic comparison.

There is one more place where we use `TypeMirror#hashCode()`:
 * [SelectionParameters#L72](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/source/SelectionParameters.java#L72)

There are other places where we use `TypeMirror#equals`:

* [SelectionParameters#L97](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/source/SelectionParameters.java#L97)
* [Type#L570](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java#L570)
* [MethodMatcher#L398](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/source/MethodMatcher.java#L398)

filiphr:
Fixes #1129

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/EnumMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/EnumMappingMethod.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Set;
 
 import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Types;
 
 import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.source.EnumMapping;
@@ -99,7 +100,7 @@ public class EnumMappingMethod extends MappingMethod {
                 }
             }
 
-            SelectionParameters selectionParameters = getSelecionParameters( method );
+            SelectionParameters selectionParameters = getSelecionParameters( method, ctx.getTypeUtils() );
 
             Set<String> existingVariables = new HashSet<String>( method.getParameterNames() );
             List<LifecycleCallbackMethodReference> beforeMappingMethods =
@@ -110,13 +111,13 @@ public class EnumMappingMethod extends MappingMethod {
             return new EnumMappingMethod( method, enumMappings, beforeMappingMethods, afterMappingMethods );
         }
 
-        private static SelectionParameters getSelecionParameters(SourceMethod method) {
+        private static SelectionParameters getSelecionParameters(SourceMethod method, Types typeUtils) {
             BeanMappingPrism beanMappingPrism = BeanMappingPrism.getInstanceOn( method.getExecutable() );
             if ( beanMappingPrism != null ) {
                 List<TypeMirror> qualifiers = beanMappingPrism.qualifiedBy();
                 List<String> qualifyingNames = beanMappingPrism.qualifiedByName();
                 TypeMirror resultType = beanMappingPrism.resultType();
-                return new SelectionParameters( qualifiers, qualifyingNames, resultType );
+                return new SelectionParameters( qualifiers, qualifyingNames, resultType, typeUtils );
             }
             return null;
         }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -572,7 +572,7 @@ public class Type extends ModelElement implements Comparable<Type> {
                         continue;
                     }
                     VariableElement arg = executable.getParameters().get( 0 );
-                    if ( arg.asType().equals( typeArg ) ) {
+                    if ( typeUtils.isSameType( arg.asType(), typeArg ) ) {
                         candidates.add( adder );
                     }
                 }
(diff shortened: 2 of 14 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/EnumMappingMethod.java
```diff
@@ -99,7 +100,7 @@ else if ( mappedConstants.size() == 1 ) {
                 }
             }
 
-            SelectionParameters selectionParameters = getSelecionParameters( method );
+            SelectionParameters selectionParameters = getSelecionParameters( method, ctx.getTypeUtils() );
```
sjaakd: selecionParameters = selectionParameters

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -572,7 +572,7 @@ private Accessor getAdderForType(Type collectionProperty, String pluralPropertyN
                         continue;
                     }
                     VariableElement arg = executable.getParameters().get( 0 );
-                    if ( arg.asType().equals( typeArg ) ) {
+                    if ( typeUtils.isSameType( arg.asType(), typeArg ) ) {
```
sjaakd: the `equals` method already uses `isSameType`..

filiphr: `arg.asType()` is a `TypeMirror` and not our `Type`. `arg` is also not our object but  `javax.lang.model.element.VariableElement`

processor/src/main/java/org/mapstruct/ap/internal/model/EnumMappingMethod.java
filiphr: I can change that as well, but with #1197 this class will go away. Right?

sjaakd: ok

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
sjaakd: OK, but isn't it strange that we use `TypeMirror` from other places iso `Type`

filiphr: This is only here locally, we need to get the type of the `ExecutableElement`

processor/src/main/java/org/mapstruct/ap/internal/model/source/SelectionParameters.java
```diff
@@ -67,9 +71,8 @@ public TypeMirror getResultType() {
     @Override
     public int hashCode() {
         int hash = 3;
-        hash = 97 * hash + (this.qualifiers != null ? this.qualifiers.hashCode() : 0);
         hash = 97 * hash + (this.qualifyingNames != null ? this.qualifyingNames.hashCode() : 0);
-        hash = 97 * hash + (this.resultType != null ? this.resultType.hashCode() : 0);
+        hash = 97 * hash + (this.resultType != null ? this.resultType.toString().hashCode() : 0);
```
gunnarmorling: Why is it using the name here?

gunnarmorling:
Looks good. Should we add a rule in the build that forbids any calls of `TypeMirror#equals()`?  We could use the Forbidden APIs plug-in (not sure whether we already use that one or not).

processor/src/main/java/org/mapstruct/ap/internal/model/source/SelectionParameters.java
filiphr: Which name do you mean? The `hashCode` is also not deterministic for `TypeMirror` and the `resultType` is a `TypeMirror`

gunnarmorling:
By "name" I meant the toString() invocation. But it seems alright.

filiphr:
>Looks good. Should we add a rule in the build that forbids any calls of TypeMirror#equals()? We could use the Forbidden APIs plug-in (not sure whether we already use that one or not).

Thanks. I think a rule would be good to have. Are you talking about [this](https://github.com/policeman-tools/forbidden-apis) plugin? I haven't used a plugin like this. I can try to set it up.

I mentioned in the issue about using sonarqube. With sonarqube we can hook it in with the PR. However, in build it might be better.

gunnarmorling:
Yes, that's the one.

> However, in build it might be better.

Right, that way we'll notice about any such issues right when building
locally.

PR #1204, opened 2017-05-19
Update assertj and used dedicated method in tests

filiphr:
Just a small bump of AssertJ and switch to using a new dedicated method in AssertJ instead of our own


agudian:
Dammit, now I accidentally merged it with a message not starting with the issue number. I shouldn't do that on the phone, even less so while walking... 😁

filiphr:
@agudian well it's my fault, I didn't create an issue for this and thus no issue number 😟

PR #1205, opened 2017-05-20
General code cleanups:

filiphr:
* #122 Use util methods when possible
* Fix some warnings in Javadoc generation
* Don't use raw classes when not needed
* Add .yml, binding.xjb and .asciidoc files to license check exclusion

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
@@ -20,6 +20,7 @@ package org.mapstruct.ap.internal.model.source;
 
 import static org.mapstruct.ap.internal.model.source.PropertyEntry.forSourceReference;
 import static org.mapstruct.ap.internal.util.Collections.first;
+import static org.mapstruct.ap.internal.util.Collections.last;
 
 import java.util.ArrayList;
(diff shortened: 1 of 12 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
```diff
@@ -165,28 +166,23 @@ public SourceReference build() {
...
-                        prefix += sourcePropertyNames[i] + ".";
-                    }
+                    List<String> elements = new ArrayList<String>(
+                        Arrays.asList( sourcePropertyNames ).subList( 0, notFoundPropertyIndex )
+                    );
+                    elements.add( mostSimilarWord );
                     reportMappingError(
                         Message.PROPERTYMAPPING_INVALID_PROPERTY_NAME, mapping.getSourceName(),
-                        prefix + mostSimilarWord
+                        Strings.join( elements, "." )
```
agudian: I'm on my phone now, but is Strings.join available in JDK 6 already?

agudian: Now I got it, it's our own class 👍

filiphr:
Thanks for the Review @agudian. And yes `Strings` is our own class 😄.

Issue #1170, opened 2017-04-04
Bounded wildcard like Collection<? extends OrderItem> is not generated properly


cornzy:
Easy to reproduce in example project "mapstruct-examples-field-mappings".
Change org.mapstruct.example.dto.Customer.getOrderItems() return type from `Collection<OrderItem>` to `Collection<? extends OrderItem>`.

With the default CollectionMappingStrategy (ACCESSOR_ONLY) this may be hard to generate because you could not add anything to such a list. We are using ADDER_PREFERRED, in this case the generated code would be simple to fix:

  * The adder Method should be found properly
  * The new collection to add the mapped elements needs to be of type `Collection<OrderItem>` (do not copy `? extends`)
  * In some cases the type of the for-loop variable is also `? extends OrderItem` (cannot reproduce with the simple example yet)

cornzy:
If you have any hint how to fix it I may have a look in the code for my own :)

filiphr:
Can you also make an example that does not work. I think that I know how to reproduce it, but just to be on the same page.

For the adder the generated code looks like:
```java
    for ( ? extends OrderItem order : customer.getOrderItems() ) {
        customerDto.addOrder( orderItemMapper.fromOrder( order ) );
    }
```

The new collection that elements should be added to is of type `Collection<OrderItem>`. Or you mean when the setter is also `? extends OrderItem` then the new collection is with element `? extends OrderItem`?

As for the code, to fix the problem with the adder, you would need to look at the [`AdderWrapper`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java) and it's FreeMarker model [`AdderWrapper.ftl`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/AdderWrapper.ftl). Most probably the fix is in the ftl and you need to take the `typeBound` of the type parameters there.

[...]

Issue #147, opened 2014-02-25
AssertionError: isSubtype 15 when executing under JDK 6


agudian:
I tried a locally built version of MapStruct (more or less my latest pull-request) with my project and got the following exception:

I compiled mapstruct with JDK 6 and used the same JDK to perform the source-generation.

```
An annotation processor threw an uncaught exception.
Consult the following stack trace for details.
java.lang.AssertionError: isSubtype 15
        at com.sun.tools.javac.code.Types$5.visitType(Types.java:351)
        at com.sun.tools.javac.code.Types$5.visitType(Types.java:332)
        at com.sun.tools.javac.code.Types$DefaultTypeVisitor.visitWildcardType(Types.java:3201)
        at com.sun.tools.javac.code.Type$WildcardType.accept(Type.java:416)
        at com.sun.tools.javac.code.Types$DefaultTypeVisitor.visit(Types.java:3199)
        at com.sun.tools.javac.code.Types.isSubtype(Types.java:328)
        at com.sun.tools.javac.code.Types.isSubtype(Types.java:312)
        at com.sun.tools.javac.model.JavacTypes.isSubtype(JavacTypes.java:80)
        at org.mapstruct.ap.model.common.TypeFactory.getType(TypeFactory.java:125)
        at org.mapstruct.ap.model.common.TypeFactory.getTypeParameters(TypeFactory.java:250)
        at org.mapstruct.ap.model.common.TypeFactory.getType(TypeFactory.java:172)
        at org.mapstruct.ap.model.common.TypeFactory.getReturnType(TypeFactory.java:238)
        at org.mapstruct.ap.processor.MapperCreationProcessor.alternativeTargetAccessorMethodsIn(MapperCreationProcessor.java:976)

        at org.mapstruct.ap.processor.MapperCreationProcessor.reportErrorIfMappedPropertiesDontExist(MapperCreationProcessor.java:
464)
        at org.mapstruct.ap.processor.MapperCreationProcessor.getBeanMappingMethod(MapperCreationProcessor.java:339)
        at org.mapstruct.ap.processor.MapperCreationProcessor.getMappingMethods(MapperCreationProcessor.java:199)
        at org.mapstruct.ap.processor.MapperCreationProcessor.getMapper(MapperCreationProcessor.java:103)
        at org.mapstruct.ap.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:92)
[...]

PR #1209, opened 2017-05-22
#1170 Fix wildcards in collection adder mappings

cornzy:
Fix collection mapping using ADDER with collections using wildcard bounds like List<? extends Item>.


processor/src/test/resources/fixtures/org/mapstruct/ap/test/bugs/_1170/AdderSourceTargetMapperImpl.java
```diff
@@ -0,0 +1,150 @@
...
+import org.mapstruct.ap.test.bugs._1170._target.Target;
+import org.mapstruct.ap.test.bugs._1170.source.Source;
+import org.mapstruct.factory.Mappers;
+
+@Generated(
+    value = "org.mapstruct.ap.MappingProcessor",
+    date = "2017-05-21T21:54:47+0200",
+    comments = "version: , compiler: javac, environment: Java 1.8.0_131 (Oracle Corporation)"
+)
+public class AdderSourceTargetMapperImpl implements AdderSourceTargetMapper {
```
filiphr: The test fails because this file is wrong. I think you used an older master to test it. We no longer do a `null` per default. Also I am not sure if we really need a fixture here. We usually use fixtures when we want to test that the generated code looks like we are expecting. @gunnarmorling, @agudian, @sjaakd do you think that we need one here?

processor/src/test/resources/fixtures/org/mapstruct/ap/test/bugs/_1170/PetMapperImpl.java
```diff
@@ -0,0 +1,29 @@
...
+package org.mapstruct.ap.test.bugs._1170;
+
+import javax.annotation.Generated;
+
+@Generated(
+    value = "org.mapstruct.ap.MappingProcessor",
+    date = "2017-05-19T23:07:12+0200",
+    comments = "version: 1.2.0-SNAPSHOT, compiler: javac, environment: Java 1.8.0_131 (Oracle Corporation)"
+)
+public class PetMapperImpl extends PetMapper {
```
filiphr: This we definitely don't need, you are also not using it in the fixture check.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1170/PetMapper.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Map;
+
+import org.mapstruct.Mapper;
+
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * @author Cornelius Dirmeier
+ */
+@Mapper
```
filiphr: This doesn't need to be a `@Mapper`. Just as info one can use any class in `@Mapper#uses`.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -375,14 +375,16 @@ public Type erasure() {
...
-    // TODO This doesn't yet take wild card types into account; e.g. ? extends Integer wouldn't be assignable to Number
-    // atm.
     public boolean isAssignableTo(Type other) {
         if ( equals( other ) ) {
             return true;
         }
 
-        return typeUtils.isAssignable( typeMirror, other.typeMirror );
+        TypeMirror typeMirrotToMatch = isWildCardExtendsBound() ? getTypeBound().typeMirror : typeMirror;
+        TypeMirror otherTypeMirrotToMatch = other.isWildCardSuperBound() ?
```
filiphr: I think that there is no test that tests when `other` is `wildCardSuperBound`.  Can you add a case like that as well?

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -375,14 +375,16 @@ public Type erasure() {
...
      */
-    // TODO This doesn't yet take wild card types into account; e.g. ? extends Integer wouldn't be assignable to Number
-    // atm.
     public boolean isAssignableTo(Type other) {
         if ( equals( other ) ) {
             return true;
         }
 
-        return typeUtils.isAssignable( typeMirror, other.typeMirror );
+        TypeMirror typeMirrotToMatch = isWildCardExtendsBound() ? getTypeBound().typeMirror : typeMirror;
```
filiphr: There is a type in the variable `typeMirrotToMatch`, it should be `typeMirrorToMatch` the same for the `otherTypeMirrotToMatch`, it should be `otherTypeMirrorToMatch`.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
cornzy: I do not have any practicle use case for this scenario. I think it is right but anyhow if the second type is `? super AnyType` it resolves to `Object` and hence any type would be assignable to this? Very theoretically. Maybe it would be better to leave the super-wildcard-case and only implement the extends-wildcard-case.
Further more if there would be a `List<? super AnyType>` (for which I do not have any practical use case) any iteration over that list must use Object and hence our Type#getTypeBound would lead to the wrong type.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
cornzy: there is a typo in the word typo of your comment ;) (done)

processor/src/test/java/org/mapstruct/ap/test/bugs/_1170/PetMapper.java
cornzy: but it did not work without @Mapper annotation. Do I need to configure it otherwise?

cornzy:
why are travis checks failing? When I call local build works without any problems. Or do I have to build additional modules?

processor/src/test/java/org/mapstruct/ap/test/bugs/_1170/PetMapper.java
filiphr: For now it has to be like this, because we use the default component model. I created #1210 for this.

filiphr:
>why are travis checks failing? When I call local build works without any problems. Or do I have to build additional modules?

The integration tests are using maven toolchains for running, on Travis it is setup to run with JDK 6, 7 and 8 plus the eclipse compiler 6, 7 and 8. You most probably don't have toolchains setup and on top of that, I doubt you have JDK6 (I don't have it as well :)). 

The failure is the same as in #147 there is some bug in the JDK6 that has some issues with generics. @agudian, @sjaakd, @gunnarmorling Do you have an idea what we need to do in our `TypesDecorator` to get around this problem?

cornzy:
@filiphr I already have a toolchains.xml in my .m2 folder and there is also an oracle jdk 6 configured - also need it for some other legazy projects ;). But it seems like the maven build does not use my toolchain. Any hint how to configure that correctly?

filiphr:
@cornzy didn't expect that 😄. Have a look at the [toolchains-jenkins](https://github.com/mapstruct/mapstruct/blob/master/etc/toolchains-travis.xml) that we have. Make sure that the info in the `<provides>` tag is done similarly. We actually parse this before and use it in the runner.

gunnarmorling:
If it's a bug that only is exposed in JDK 6 but not in later versions I personally would be fine with ignoring this. I.e. we'd have to exclude this test somehow from running on JDK 6.

filiphr:
@gunnarmorling yes it only fails on the Open JDK 6 (Travis does not support the Oracle JDK 6 as it is EOL :)). 

@cornzy If you want exclude it have a look at `FullFeatureCompilationTest` line 71, you'll need to add another line for the package of the new test.

filiphr:
@cornzy Do you need some help with this one? Do you think you'll have time to wrap it up?

gunnarmorling:
Hey @cornzy, @filiphr, where are we standing with this one?

filiphr:
Thanks a lot for the PR @cornzy. It has been merged into master. It will be part of the 1.2.0.CR1 which should come out soon 😄

PR #1217, opened 2017-05-30
#1215 Add import correctly for array types

### LINKED ISSUE #1215: Regression: Missing import when upgrading to 1.2.0.Beta2
Hi, we were using Mapstruct version 1.1 but we are upgrading to 1.2 because we need Lombok support. After the upgrade, old mappers stopped working.

The problem is demonstrated in https://github.com/juriad/mapstruct-missing-import-1.2

Basically, this (in 1.1):
```
if ( dto.getTags() != null ) {
  some.entity.Tag[] tags = dto.getTags();
  entity.setTags( Arrays.copyOf( tags, tags.length ) );
}
```
was replaced with this (1.2):
```
Tag[] tags = dto.getTags();
if ( tags != null ) {
  entity.setTags( Arrays.copyOf( tags, tags.length ) );
}
```

Notice that in version 1.2, the type Tag[] is not fully qualified, which causes a compilation problem.

filiphr:
Fixes #1215.

This makes sure that a type is considered as an import event if if is of an array type. The array suffix `[]` is always stripped in the import name of the `Type`

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
@@ -199,6 +199,14 @@ public abstract class GeneratedType extends ModelElement {
         return importedTypes;
     }
 
+    public SortedSet<String> getImportTypeNames() {
+        SortedSet<String> importTypeNames = new TreeSet<String>();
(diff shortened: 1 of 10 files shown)


filiphr:

PR #1225, opened 2017-06-15
#611 Allow nested declaration of Mappers

### LINKED ISSUE #611: Nested Mapper Interface not supported
Hi, 

If the mapper interface (or abstract class) is nested into another class/interface, mapstruct won't work. 

For example, consider the following simple class:

``` java
import org.mapstruct.Mapper;

public class ClassWithNestedMapper {
    @Mapper
    public interface MyMapper {
        MyMapper INSTANCE = Mappers.getMapper(MyMapper.class);
    }

        System.out.println(MyMapper.INSTANCE);
    }
}
```

A call to the main method will fail :

``` java
java.lang.ClassNotFoundException: ClassWithNestedMapper$MyMapperImpl
    at org.mapstruct.factory.Mappers.getMapper(Mappers.java:82)
    at mapper.SecondClassWithNestedMapper$MyMapper.<clinit>(SecondClassWithNestedMapper.java:9)
    ... 1 more
Caused by: java.lang.ClassNotFoundException: ClassWithNestedMapper$MyMapperImpl
    at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
    at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
    at java.security.AccessController.doPrivileged(Native Method)
    at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
    at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
    at org.mapstruct.factory.Mappers.getMapper(Mappers.java:77)
    ... 2 more
```

because the `getMapper(...)` method tries to load class

``` java
clazz.getName() + IMPLEMENTATION_SUFFIX 
```

which resolves to `ClassWithNestedMapper$MyMapperImpl`

Though the fact is that the implementation class is generated as a standalone file - not nested into ClassWithNestedMapper obviously.

IMHO, the bug is on the generation side. Because two different classes in the same package may have nested interfaces with the same name. And I don't talk about interfaces nested into interfaces...
So I guess the solution will have to change the naming convention for the generated implementation and/or dispatch implementations into sub-packages according to the mapper nest hierarchie.

Example 2 : `ClassWithNestedMapper$MyMapper` and `AnotherClassWithNestedMapper$MyMapper` will collide in generating `MyMapperImpl.java` in the same directory.

``` java
import org.mapstruct.Mapper;

public class AnotherClassWithNestedMapper {
    @Mapper
    interface MyMapper {
        MyMapper INSTANCE = Mappers.getMapper(MyMapper.class);
    }

        System.out.println(MyMapper.INSTANCE);
    }
}

```

**Side note**
By the way, the generated mapper implementation does not compile anyway : it declares that it implements `MyMapper` instead of `ClassWithNestedMapper.MyMapper` ...!

I thank you in advance for your time and kind feedback. I think I can give a hand and draft a solution if that's needed.

Cheers!

Tillerino:
Fixes #611.

The method name "getFlatName" is supposed to imply that it squashes/flattens several class names into one.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_611/Issue611Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_611/Issue611Test.java
@@ -1,0 +1,71 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.bugs._611;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+/**
+ * @author Tillmann Gaida
+ */
+@IssueKey("611")
+@RunWith(AnnotationProcessorTestRunner.class)
+public class Issue611Test {
+    /**
+     * Checks if an implementation of a nested mapper can be loaded at all.
+     */
+    @Test
+    @WithClasses(SomeClass.class)
+    public void mapperIsFound() throws Exception {
+        assertNotNull( SomeClass.InnerMapper.INSTANCE );
+    }
+
+    /**
+     * Checks if an implementation of a nested mapper can be loaded which is nested into an already
+     * nested class.
+     */
+    @Test
+    @WithClasses(SomeClass.class)
+    public void mapperNestedInsideNestedClassIsFound() throws Exception {
+        assertNotNull( SomeClass.SomeInnerClass.InnerMapper.INSTANCE );
+    }
+
+    /**
+     * Checks if it is possible to load two mapper implementations which have equal simple names
+     * in the same package.
+     */
+    @Test
+    @WithClasses({ SomeClass.class, SomeOtherClass.class })
+    public void rightMapperIsFound() throws Exception {
+        SomeClass.InnerMapper.Source source1 = new SomeClass.InnerMapper.Source();
+        SomeOtherClass.InnerMapper.Source source2 = new SomeOtherClass.InnerMapper.Source();
+
+        SomeClass.InnerMapper.Target target1 = SomeClass.InnerMapper.INSTANCE.toTarget( source1 );
+        SomeOtherClass.InnerMapper.Target target2 = SomeOtherClass.InnerMapper.INSTANCE.toTarget( source2 );
+
+        assertTrue( target1 instanceof SomeClass.InnerMapper.Target );
+        assertTrue( target2 instanceof SomeOtherClass.InnerMapper.Target );
+    }
+}
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_611/Issue611Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_611/Issue611Test.java
@@ -18,28 +18,30 @@
  */
 package org.mapstruct.ap.test.bugs._611;
 
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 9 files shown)


filiphr:
@Tillerino Thanks for the PR. I changed the base branch to `master`. Can you please do a rebase on top of master and push the changes, so a build via Travis is triggered?

We are currently nearing the release of 1.2.0.RC1, so we won't be doing anymore updates for 1.1.0.Final.

I'll do a review of this during the weekend.

Tillerino:
Oh I was wondering why Travis wasn't doing anything.

Rebased, tests are running. :+1:

Btw, is there a roadmap for 1.2.0?

processor/src/test/java/org/mapstruct/ap/testutil/runner/GeneratedSource.java
```diff
@@ -31,6 +31,7 @@
 import org.mapstruct.ap.testutil.assertions.JavaFileAssert;
 
 import static org.assertj.core.api.Assertions.fail;
+import static org.mapstruct.factory.Mappers.getFlatName;
```
filiphr: As I wrote in the `Mapper`, I would write an implementation here and hide the one in `Mappers`

filiphr:
>Btw, is there a roadmap for 1.2.0?

Right now the `1.2.0.Beta3` is out. Next would be the RC1 hopefully end of this month or until mid of July. 

We are looking for feedback for the new changes in 1.2.0 so if you can try the beta and report problems it would be great.

Tillerino:
While this is still open: why not use the same `$` to separate outer from inner class(es) in the implementation's name? It is a legal identifier and the [Java specification 3.8](https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.8) merely states

> The $ character should be used only in mechanically generated source code or, rarely, to access pre-existing names on legacy systems.

Since this is clearly _mechanically generated source code_ there is no reason to shy away from this special character and I feel like the purpose of a `$` is very clear as opposed to a `_`.

This would also require _no change_ to `Mappers` or `GeneratedSource` since these were compatible with that kind of natural naming from the start.

To be honest, I went with Gunnar's suggestion because I wanted the least resistance for this PR but the `$` separator makes much more sense to me.

processor/src/test/java/org/mapstruct/ap/test/bugs/_611/Issue611Test.java
```diff
@@ -0,0 +1,42 @@
...
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+/**
+ * @author Tillmann Gaida
+ */
+@IssueKey("611")
+@RunWith(AnnotationProcessorTestRunner.class)
+@WithClasses(SomeClass.class)
```
Tillerino: Yes, absolutely.

filiphr:
To be honest I don't know what to say about using `$` maybe it is OK to be used in this case. Let's see what the other think about this.

core-common/src/main/java/org/mapstruct/factory/Mappers.java
```diff
@@ -119,4 +124,22 @@ private Mappers() {
...
+    /**
+     * <p>
+     * Returns the simple name of a class with all enclosing classes' simple names prepended,
+     * separated by an underscore.
+     * </p>
+     * <p>
+     * Example: returns <code>Map_Entry</code> for {@link java.util.Map.Entry}.
+     * </p>
+     */
+    public static String getFlatName(Class<?> klass) {
```
Tillerino: Added unit test with two levels.

:100: on making `getFlatName` private (pollutes API), but maybe we don't need it at all (see PR comment)

Tillerino:
Any additional opinions? I guess we're mainly looking for a word from @gunnarmorling

filiphr:
Apart from the comments I wrote nothing else form my side. 

We are going to have a meeting where we go over the PRs next week. I'll keep you updated. I don't see any major issues for this to be rejected. Thanks a lot for the patience.

filiphr:
@Tillerino we discussed with @gunnarmorling  and the team and we think that using the `$` makes sense. Can you do that and push all the other changes with the tests. Once that is  done I can merge this.

gunnarmorling:
+1 on using "$"; when writing my original comment, I wasn't aware of the fact that it is a legal identifier character in source files, but given the JLS reference you linked to, it makes sense very much. Thanks!

core-common/src/main/java/org/mapstruct/factory/Mappers.java
```diff
@@ -119,4 +124,22 @@ private Mappers() {
...
+        String flatName = klass.getSimpleName();
+        while (klass.isMemberClass()) {
+            klass = klass.getEnclosingClass();
+            flatName = klass.getSimpleName() + "_" + flatName;
```
gunnarmorling: +1 for using a string builder in case the class is an inner one (i.e. let's avoid the string builder for the default case of a top-level class).

Tillerino:
Will do :+1:

Tillerino:
Changed to `$` separator, used `StringBuilder` for concatenation. Using `$` also made a couple of changes redundant. Is this good to go?

filiphr:
@Tillerino thanks a lot. It is great. The only thing is that we use AssertJ assertions and not JUnit. They provide much better error message and much more possibilities. I can do that during the merge if it is OK for you as well.

Tillerino:
Sure, go ahead. 👍

filiphr:
Thank you for the contribution @Tillerino. Pull request has been integrated into `master`.

PR #1229, opened 2017-06-23
Immutable builder concept

smartytime:
(Edited)
I altered my original PR somewhat significantly, so I'll just restate the newish approach:

First, my goals, which I probably failed somewhat miserably:
* Avoid making large fundamental changes to mapstruct.  I wanted builders to be something that almost "sat on top" of the core mapstruct code.
* At a minimum, support the most simple and common patterns for immutables/builders... including lombok (the use case driving my work).  Maintainers may disagree with me, but I'd be okay having mapstruct be more opinionated if it kept mapstruct simpler and more stable.

## High-Level Thought Process

A typical generated BeanMappingMethod structure follows this pattern:
```
[resultType] instanceVarName = [resultTypeCtrInvocation or factoryMethod];
instanceVarName.[ writeAccessorInvocation ];
...
return instanceVarName;
```

In the case of a builder, the "write" operations happen on a different Class altogether, so it seemed like the mappingTarget Type would need to know when it was immutable and needed a builder, and also have some basic information about how the build should happen.  Generally, though, the template would need to change to look like this:

```
[builderType] instanceVarName = [builderCtr or builderFactoryMethod];
instanceVarName.[ builderWriteAccessorInvocation ];
...
return instanceVarName.[buildMethodInvocation];
```

So, at a minimum, if a Type could recognize that it needed a builder, you could get away with:

* A factory method that produced an instance of the Builder
* A way to recognize writeAccessors for builders (the template technically wouldn't even need to change)
* Tracking the "build" method name, so you could invoke it before returning/assigning the result in the template.

The first and third items are relatively trivial, although I recognize they break one of the core tenants of mapstruct: _don't require annotations on domain classes_.  I started out with annotating mapping methods, but didn't like it and ended up moving the annotation to either the Builder class, or the Target class.

It will work either way - and I can see the value in agreeing to _not_ annotated domain classes.  

## Implementation

Anyhow, back to the second item... supporting "builder" Accessors as valid for mappings.  Here's the approach I took:

1. Added a "builderType" property on Type.  When the Type is immutable and needs a builder, the builderType property contains a reference to the Type for the builder class.  This property is picked up from the annotation.
2. When a Type has a builder, it will include any accessors from the builder class when it builds mappings.  Here, the framework needed to be able to differentiate between builder setter methods and normal pojo setter methods.  I'm not sure the best way to accomplish, but what made sense to me was hard-coding the following logic:  a) the accessor is public, b) the accessor is in a known builder class, c) the accessor has a single parameter, d) the accessor returns an instance of the builder class itself.  d is a little tricky; if we take this approach, I'll need some direction.
3. Added a "builderOptions" property on Type.  If a Type has a builder, this property stores metadata about the builder, currently: how to instantiate, and how to invoke build() to get the actual target instance.
4. Modified the template accordingly.

I've added a bunch of tests based on comments from @filiphr , which have helped me understand mapstruct better and unwind some of the mistakes I made in my first approach.  Currently, I've verified that builders are working for these use cases:

* Basic property mapping
* Nested target property mapping
* Abstract/generic target mappings
* Abstract/generic builders

### AfterMapping
I could see two approaches:
* Passing the builder into the method. 
* Passing the immutable object, then assume the ```@AfterMapping``` method would "thaw" it back into a builder, and return another Immutable instance.  Obviously you couldn't have a void return type in this case.

### ObjectFactory for Builders
Self-explanatory.  Currently, I just added a method name on the target class that is expected to return a builder instance.  

### ObjectFactory for abstract targets (where the immutable is the implementation)
I got the framework to pick up an ObjectFactory that returned the Builder type as a valid ObjectFactory for an abstract target, but the code generation doesn't work.  When the mapper looks like this:
```
AbstractTarget toAbstract(ConcreteSource concrete);

default ConcreteTarget.Builder concreteTargetBuilder() {
    return ConcreteTarget.builder();
}

```

it returns this:
```
AbstractType abstractType = factoryThatReturnsBuilder();
abstractType.builderSetter( someProp );
return abstractType;
```

In this case, the framework would have to know explicitly that the concrete immutable/builder should be used, rather than using the abstract classes accessors.  Maybe this should be working?  It does work if you actually map the abstract target using a mapping method instead of an ObjectFactory, eg:

```
ConcreteTargetThatContainsAbstractChildProperty toTarget(ConcreteSource concrete);

ConcreteChildProperty toChildProperty(ConcreteChildSource);
```


core-common/src/main/java/org/mapstruct/MappedByBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
+@Experimental
+public @interface MappedByBuilder {
+  Class builderClass();
+
+  String buildMethod() default "build";
+
+  String staticBuildMethod() default "builder";
```
smartytime: oops - this should be named "staticBuilderMethod" or "staticFactoryMethod" might be better.

filiphr:
Thanks a lot for the PR @smartytime. I also saw the issue and your detailed writeup. 

We'll have a deeper look into the PR and will get back to you in the issue comments and on the PR.

processor/src/main/java/org/mapstruct/ap/internal/util/accessor/AbstractAccessor.java
```diff
@@ -32,9 +35,20 @@
 abstract class AbstractAccessor<T extends Element> implements Accessor {
 
     protected final T element;
+    private final TypeMirror builderType;
 
     AbstractAccessor(T element) {
         this.element = element;
+        if (element.getEnclosingElement().getKind() == ElementKind.CLASS) {
```
smartytime: Yeah.  I'm going to try to get this working without adding builderType to Accessor.  Should I just update this PR if I make changes?  Or wait for other's feedback?

filiphr: If you are doing any changes feel free to update the PR. We can sync up along the way 😀

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
```diff
@@ -111,9 +110,21 @@ public static boolean isPresenceCheckMethod(Accessor method) {
...
+                && executable.getParameters().size() == 1
+                && (ACCESSOR_NAMING_STRATEGY.getMethodType( executable ) == MethodType.SETTER ||
+                isBuilderSetterMethod( method ) );
+    }
+
+    private static boolean isBuilderSetterMethod(Accessor method) {
+        if ( method.isBuilder() ) {
+            String methodReturnType = method.getExecutable().getReturnType().toString();
+            String builderType = method.getBuilderType().toString();
+            return methodReturnType.equals( builderType );
```
smartytime: I'd like to add a bunch of test cases for this feature... I suppose there might be limitations to what features in mapstruct work when using immutable builders?  What are some other things you think I should test that might expose where this approach is going to break?

processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
```diff
@@ -76,9 +76,12 @@
     	</#if>
     	<@includeModel object=callback targetBeanName=resultName targetType=resultType/>
     </#list>
-    <#if returnType.name != "void">
```
smartytime: So, currently it's this:
```
if( returnType.isMappedByBuilder() ) {
...
}
else if( returnType.name != "void" ) {
...
}
```
and are you saying it should be this?

```
if( returnType.name != "void" ) {
  if ( returnType.isMappedByBuilder() ) {

  }
  else {
    
  }
}
```

I think that makes sense - you can read my updated initial comment - at the bottom I talked about ```@AfterMapping``` methods, which might use a builder and return void.

filiphr: This part here is not connected with `@AfterMapping`, `@AfterMapping` occurs before this. This part here generates the return of the method. 

Yes I am saying that it should be like the second example. The reason for this is that if the return type of the method for example:

```
void update(Source source, @MappingTarget Target target);
```

If for this method the `returnType` is `void`, but the `resultType` is `Target`. If we do like it is now and there is a builder for the `resultType` then there will be a compilation error.

filiphr:
My comments are without looking at the code. I'll write comments for the code directly.

> Avoid making large fundamental changes to mapstruct. I wanted builders to be something that almost "sat on top" of the core mapstruct code.

I'll go through the implementation and get back to you, maybe we can make it like that 😄 

> At a minimum, support the most simple and common patterns for immutables/builders... including lombok (the use case driving my work). Maintainers may disagree with me, but I'd be okay having mapstruct be more opinionated if it kept mapstruct simpler and more stable.

For us it is not a problem to be more opinionated. It is completely OK to find all the needed methods during compilation time. There is a lot of information in the code and it is not that complex. If something makes more problems we could extend once we get more feedback from the users.

## High-Level Thought Process

I would say that the template would stay the same, with the only exception is that now the `returnType` is `mappingToType`

```
[mappingToType] instanceVarName = [mappingToTypeCtrInvocation or factoryMethodForMappingToType];
instanceVarName.[ writeAccessorInvocation ];
...
return instanceVarName.[buildMethodInvocation]
// or return instanceVarName;
```

If we need to use builder then `mappingToType` would be the Builder Type, otherwise it is the `resultType`. Currently. we have [`BeanMapping#resultType`](http://mapstruct.org/documentation/stable/api/org/mapstruct/BeanMapping.html#resultType--), so the type that we use is not always the type that the method returns. We can use the same concept.

For your three points. I think that MapStruct can easily do all off those without the help of the user. Of course we'll have to define some rules and create the `BuilderOptions` from those rules.

## Implementation

1. I agree here with you. The only thing is whether we stay with the annotation on the class or we go with a different approach (maybe the SPI?)
1. I think that we should solve this via the `AccessorNamingStrategy` somehow. The reason is that not all builders use just the simple name for the method. Some use `with` some might use something else. If a user uses Builder with Lombok then in the Strategy he / she can have this:

```
@Override
    Element parent = getterOrSetterMethod.getEnclosingElement();
    if ( parent.getKind() != ElementKind.PACKAGE && parent.getKind().isClass() &&
        parent.getEnclosingElement().getAnnotation( Builder.class ) != null ) {

        return getterOrSetterMethod.getSimpleName().toString();
    }
    return Introspector.decapitalize( methodName.substring( methodName.startsWith( "is" ) ? 2 : 3 ) );
}
```

Or for AutoValue

```
@Override
    Element parent = getterOrSetterMethod.getEnclosingElement();
    if ( parent.getAnnotation( AutoValue.Builder.class ) != null ) {
        return getterOrSetterMethod.getSimpleName().toString();
    }
    return Introspector.decapitalize( methodName.substring( methodName.startsWith( "is" ) ? 2 : 3 ) );
}
```
1. This is OK
1.  Also OK

## Known Issues and other things to think through

### AfterMapping

Starting from 1.2.0 `@AfterMapping` can return a value and if this happens we return this value instead. Theoretically we can always pass the builder and if something is returned return that. 

### ObjectFactory for abstract targets (where the immutable is the implementation)

Currently in order to map abstract classes / interfaces one needs to use `@BeanMapping(resultType)`. 

Do you have a test case for this? Is this something for AutoValue support?

smartytime:
Closing.  I'm going to start a new PR due to significant change in approach.

PR #1230, opened 2017-06-24
#1227 Make sure that only types belonging to java.lang are not imported

### LINKED ISSUE #1227: import is not generated for sub-packages of java.lang
I tried to map java.lang.management.ThreadInfo:

`void map(final ThreadInfo threadInfoSrc, @MappingTarget final ThreadDto threadDto);`

even `@Mapper(imports = ThreadInfo.class` didn't work.

The generated class didn't include the import statement.
I use MapStruct 1.1.0.Final with java8

filiphr:
Fixes #1227.

I am not sure if we need to add the types from the other `java.lang` sub-packages. Like `java.lang.ref`, `java.lang.reflect` etc.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
@@ -235,7 +235,9 @@ public abstract class GeneratedType extends ModelElement {
         }
 
         if ( typeToAdd.getPackageName() != null ) {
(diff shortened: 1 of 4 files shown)


gunnarmorling:
Looks good to me. I think `java.lang` is the only package implicitly imported. Thanks, @filiphr!

filiphr:
I meant whether we need to add tests for those packages, like the one in the `java.lang.management`

gunnarmorling:
Ah, got you. You've added a test for one, that's alright :)

PR #1234, opened 2017-06-28
#1185 display the specific type for a property for which no write accessor exists or it is unknown

### LINKED ISSUE #1185: Report the specific type for which a nested target property has no write accessor or not exists
With #1184 the reporting of the error is moved into the `TargetReference`. The error reporting stays the same and the entire target name is reported.

For a Mapping like:
```java
@Mappings( {
    @Mapping( target = "readOnly", source = "sourceRead"),
    @Mapping( target = "writable2", source = "writable"),
    @Mapping( target = "nestedTarget.readOnly", source = "nestedSource.nested"),
    @Mapping( target = "nestedTarget.writable2", source = "nestedSource.writable"),
} )
Target map(Source source);
```

The error messages look like:
* `Property "readOnly" has no write accessor.` 
* `Unknown property "writable2" in return type.`.
* `Property "nestedTarget.readOnly" has no write accessor.` 
* `Unknown property "nestedTarget.writable2" in return type.`.

It would be much better if MapStruct reports that like:
* `Property "readOnly" has no write accessor in "com.example.Target".`
* `Unknown property "writable2" in "com.example.Target".`
* `Property "readOnly" has no write accessor in "com.example.NestedTarget".`
* `Unknown property "writable2" in "com.example.NestedTarget".`

For a discussion: Do we want to keep the same error messages if there are no nested target mappings, or it is still better to display the return type in the error message?

filiphr:
Fixes #1185.

Just for info we are displaying the current type only and not nested. The error message contains the entire path of the problematic property.


processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -31,8 +31,8 @@
     BEANMAPPING_NO_ELEMENTS( "'nullValueMappingStrategy', 'resultType' and 'qualifiedBy' are undefined in @BeanMapping, define at least one of them." ),
     BEANMAPPING_NOT_ASSIGNABLE( "%s not assignable to: %s." ),
     BEANMAPPING_ABSTRACT( "The result type %s may not be an abstract class nor interface." ),
-    BEANMAPPING_UNKNOWN_PROPERTY_IN_RESULTTYPE( "Unknown property \"%s\" in return type. Did you mean \"%s\"?" ),
-    BEANMAPPING_PROPERTY_HAS_NO_WRITE_ACCESSOR_IN_RESULTTYPE( "Property \"%s\" has no write accessor." ),
+    BEANMAPPING_UNKNOWN_PROPERTY_IN_RESULTTYPE( "Unknown property \"%s\" in \"%s\". Did you mean \"%s\"?" ),
```
agudian: How about `"Unknown property \"%s\" in result type %s. Did you mean \"%s\"?"` (adding “result type” again and removing the quotes around the type, as we usually only quote the property names... or do we?)

filiphr: You are right. It is better like this.

PR #1235, opened 2017-06-28
#1086 Mention the fact that prototype methods are not considered for automatic sub mapping methods

### LINKED ISSUE #1086: Use configuration from prototype methods for forged name based methods

filiphr:
Fixes #1086.

Something similar would be included in the Release notes as well.

### DIFF
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -825,6 +825,12 @@ When generating the implementation of a mapping method, MapStruct will apply the
 In order to stop MapStruct from generating automatic sub-mapping methods, one can use `@Mapper( disableSubMappingMethodsGeneration = true )`.
 ====
 
+[NOTE]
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -825,6 +825,12 @@ When generating the implementation of a mapping method, MapStruct will apply the
 In order to stop MapStruct from generating automatic sub-mapping methods, one can use `@Mapper( disableSubMappingMethodsGeneration = true )`.
 ====
 
+[NOTE]
+====
+During the generation of automatic sub-mapping methods <<shared-configurations>> will not be taken into consideration.
```
agudian: “...into consideration, yet.”?

Issue #883, opened 2016-09-04
generated variable names with ı instead i


Runomu:
I'm using mapstruct in spring-boot with component model 'spring'.  My problem is generated variable names have ı instead i. i'm using IDEA and encoding for editor is already set to UTF-8. i have also set encoding to UTF-8 in maven properties. But i can't still generate concrete classes with 'i' instead of 'ı'. After Build success it becomes 'ı'. And if i try to deploy, make module in IDEA fails and all 'ı' in code become question marks. Code looks like following: 

`
  @Generated(
       value = "org.mapstruct.ap.MappingProcessor",
      date = "2016-09-05T00:38:27+0300",
      comments = "version: 1.0.0.Final, compiler: javac, environment: Java 1.8.0_74 (Oracle Corporation)"
    )
   @Component
    public class VeininvoiceMapperImpl extends VeininvoiceMapper {

```
@Autowired
private LocalDateMapper localDateMapper;
@Autowired
private BoxTypeMapper boxTypeMapper;
@Autowired
private InvoiceStatusCodeMapper ?nvoiceStatusCodeMapper;
@Autowired
private SenderLabelMapper senderLabelMapper;
@Autowired
private PostboxLabelMapper postboxLabelMapper;
@Autowired
private VeinpartnerMapper veinpartnerMapper;

@Override
public Invoice veininvoiceToDto(Veininvoice veininvoice) {
    if ( veininvoice == null ) {
        return null;
    }

    Invoice ?nvoice = new Invoice();

    ?nvoice.setInvoiceStatusCode( ?nvoiceStatusCodeMapper.statusToDto( veininvoice.getInvoicestatuscode() ) );
    ?nvoice.setBoxType( boxTypeMapper.boxtypeToDto( veininvoice.getBoxtype() ) );
    ?nvoice.setPartner( veinpartnerMapper.veinpartnerToDto( veininvoice.getVeinpartner() ) );
    ?nvoice.setInvoiceNr( veininvoice.getInvoiceNr() );
    ?nvoice.setIsarchive( veininvoice.getIsarchive() );
    ?nvoice.setUblversionid( veininvoice.getUblversionid() );
    ?nvoice.setCustomizationid( veininvoice.getCustomizationid() );
    ?nvoice.setCopyindicator( veininvoice.getCopyindicator() );
    ?nvoice.setIssuedate( localDateMapper.issueDateToCalendar( veininvoice.getIssuedate() ) );
[...]

PR #1236, opened 2017-06-29
#883 generated variable names with ı instead i

shalugin:


processor/src/main/java/org/mapstruct/ap/internal/util/Strings.java
```diff
@@ -20,10 +20,7 @@
 
 import static org.mapstruct.ap.internal.util.Collections.asSet;
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.*;
```
filiphr: The build fails because of wild card imports, they are not allowed in the build

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
```diff
@@ -19,16 +19,21 @@
...
+import java.util.Locale;
 
 import org.junit.Test;
 
 /**
  * @author Filip Hrisafov
  */
 public class StringsTest {
+
+    private static Locale TURKEY_LOCALE = getTurkeyLocale();
```
filiphr: This should be `private static final` that is one of the reasons why the build fails.

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
```diff
@@ -110,5 +115,42 @@ public void findMostSimilarWord() throws Exception {
         );
         assertThat( mostSimilarWord ).isEqualTo( "fullName" );
     }
+    @Test
+    public void capitalizeEnglish() {
+        Locale.setDefault(Locale.ENGLISH);
+        String international = Strings.capitalize("international");
+        assertEquals("International", international);
```
filiphr: We use AssertJ for the tests

processor/src/test/java/org/mapstruct/ap/internal/util/StringsTest.java
```diff
@@ -110,5 +115,42 @@ public void findMostSimilarWord() throws Exception {
...
+    @Test
+    public void decapitalizeEnglish() {
+        Locale.setDefault(Locale.ENGLISH);
+        String international = Strings.decapitalize("International");
+        assertEquals("international", international);
+    }
+
+    @Test
+    public void capitalizeTurkish() {
+        Locale.setDefault(TURKEY_LOCALE);
```
filiphr: I think that we need to store the current default `Locale` in a `Before` method and then restore it after an `After` method. This will make sure that we don't "leak" this set.

agudian: Yes, definitely rescue and restore the default setting in `@Before` and `@After` methods 👍

processor/src/main/java/org/mapstruct/ap/internal/util/Strings.java
```diff
@@ -89,11 +86,11 @@ private Strings() {
     }
 
     public static String capitalize(String string) {
-        return string == null ? null : string.substring( 0, 1 ).toUpperCase() + string.substring( 1 );
+        return string == null ? null : string.substring( 0, 1 ).toUpperCase( Locale.ENGLISH ) + string.substring( 1 );
```
agudian: The current Javadoc of `toUpperCase` suggests using `Locale.ROOT` instead of `Locale.ENGLISH`...

shalugin:
@filiphr @agudian Big thanks for review! And thank you for this awesome project!

filiphr:
Thanks for the changes @shalugin and thanks for the kind words, it really means a lot when we get positive feedback 😄.

@agudian, can you accept this?

gunnarmorling:
Looks good to me - rebased and applied. Thanks a lot for your contribution and the nice words, @shalugin! I've squashed your two commits into one and added another one for simplifying the look-up of the locale in the test.

PR #1237, opened 2017-06-30
#1131 Use assignment source type for update methods factories

### LINKED ISSUE #1131: Source parameter for @ObjectFactory method is not well defined
It seems that the source parameter to object factory method has different meaning depending if it is used from method with @MappingTarget annotation or not. For methods with @MappingTarget parent of object which is going to be mapped into newly created object is passed as source and in case of methods without @MappingTarget annotation the object to be mapped is passed itself.
This can be seen in the example below where copyB() method implementation passes BDTO instance as source and in mergeA() method implementation passes ADTO instance.

```java
@Mapper
public interface VerifyMappings {
	class AEntity {
		private BEntity b;
		public BEntity getB() {
			return b;
		}
		public void setB(BEntity b) {
			this.b = b;
		}
	}
	class BEntity {}
	class ADTO {
		private BDTO b;
		public BDTO getB() {
			return b;
		}
		public void setB(BDTO b) {
			this.b = b;
		}
	}
	class BDTO {}

	ADTO copyA(AEntity source);
	BDTO copyB(BEntity source);

	void mergeA(AEntity source, @MappingTarget ADTO target);
	void mergeB(BEntity source, @MappingTarget BDTO target);

	@ObjectFactory
	default BDTO createB(Object source) {
		return new BDTO();
	}
}
```

filiphr:
Fixes #1131. 

With this PR the assignment type is also considered for the update methods.

Some open questions:

* See the `TODO` in the `Issue1131Mapper`. Should we consider the method parameters as well or not?
* The move of the `Assignment` to the common package. I am not that satisfied with this, maybe we need to pass the `SourceRHS` instead of the generic `Assignment`?


processor/src/test/java/org/mapstruct/ap/test/bugs/_1131/Issue1131Mapper.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public abstract void merge(Source source, @MappingTarget Target target);
+
+    @ObjectFactory
+    protected Target.Nested create(Source.Nested source) {
+        return new Target.Nested( "from object factory" );
+    }
+//TODO Should we report ambiguous methods if we use the ones below/
+//    @ObjectFactory
+//    protected Target.Nested createWithObject(Object source) {
```
filiphr: I just realized `Object` should be `Source`? Do we need to consider a method like this or not?

agudian: A method with `Source` as parameter type and Target.Nested as result should not be considered, I’d say... Imagine there was a non-update method for the Nested types defined, then such a factory method wouldn't be considered either.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1131/Issue1131Mapper.java
```diff
@@ -0,0 +1,49 @@
...
+//        throw new IllegalArgumentException( "Should not use create with object" );
+//    }
+//
+//    @ObjectFactory
+//    protected Target.Nested createDefault() {
```
agudian: This one could be defined, but would not be chosen as there is an overloaded variant with the source parameter, which then takes precedence (I think that’s how we’re describing it in the javadoc now)... 🙂

agudian:
Just added small remarks on your TODO in the test. I’ll play around with it a bit another night and also take a closer look then. But the changes make sense for me (scrolling through the diff). 

Thanks a lot for taking over! 🙌

processor/src/test/java/org/mapstruct/ap/test/bugs/_1131/Issue1131Mapper.java
filiphr: Maybe I am doing something wrong. If I uncomment this then it would be considered

processor/src/test/java/org/mapstruct/ap/test/bugs/_1131/Issue1131Mapper.java
filiphr: Same here, if the parameter is `Source`, then there is ambiguous  methods compile error

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -81,11 +86,15 @@ public TypeSelector(TypeFactory typeFactory) {
...
-    private List<ParameterBinding> getAvailableParameterBindingsFromMethod(Method method, Type targetType) {
-        List<ParameterBinding> availableParams = new ArrayList<ParameterBinding>( method.getParameters().size() + 2 );
+    private List<ParameterBinding> getAvailableParameterBindingsFromMethod(Method method, Type targetType,
+        Assignment assignment) {
+        List<ParameterBinding> availableParams = new ArrayList<ParameterBinding>( method.getParameters().size() + 3 );
 
         availableParams.addAll( ParameterBinding.fromParameters( method.getParameters() ) );
         addMappingTargetAndTargetTypeBindings( availableParams, targetType );
+        if ( assignment != null ) {
+            availableParams.add( ParameterBinding.fromAssignment( assignment ) );
```
agudian: I think if athe assignment is considered, then we need to ignore the source parameters of the method (we should still include ˋ@Contextˋ params, though)...

processor/src/test/java/org/mapstruct/ap/test/bugs/_1131/Issue1131Mapper.java
agudian: I need to look into the code, can’t remember what the trick was... maybe tonight but more likely tomorrow night...

agudian: Hehe, well... I double checked and in the Javadoc of the new `@ObjectFactory` annotation, we're writing:
```
If there are two factory methods, both serving the same type, one with no parameters and one taking sources as input, then the one with the source parameters is favored. If there are multiple such factories, an ambiguity error is shown.
```

I searched for the piece of code implementing that, and I searched for a test checking this. But I didn't find anything. I tried it out (with and without your changes) and it's just not working. 😁

Seems like we'll need to fix that as well...

Or did I miss anything?

filiphr: You are not missing anything. It seems that we don't have a test for that. It seems that it won't work in other context as well 😄. Should we fix it as part of this issue or create a new one?

processor/src/main/java/org/mapstruct/ap/internal/model/common/Assignment.java
```diff
@@ -16,13 +16,11 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License.
  */
-package org.mapstruct.ap.internal.model.assignment;
+package org.mapstruct.ap.internal.model.common;
```
filiphr: Maybe instead of using `Assignment` in the `common` package we could use the `SourceRHS`. I think that in that case no change in a package will be needed. What do you think?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/InheritanceSelector.java
```diff
@@ -41,11 +42,11 @@
                                                                           Type targetType,
                                                                           SelectionCriteria criteria) {
 
-        if ( sourceTypes.size() != 1 ) {
+        if ( sourceTypes.size() != 1 && criteria.getAssignment() == null) {
```
filiphr: I am not entirely sure if we need this here, maybe it is wrong. I think it works without that as well. 

On top of that. This together with #1242 will lead to a runtime error on line 56.

filiphr:
I reversed the change on the `InheritanceSelector`. However, I am in doubt what is correct there. From what I see that is used only for mapping methods and not for ObjectFactories and Lifecycle callbacks. However, in this case we can actually say what is the source type (based on the assignment), maybe we could do it on other locations as well, but I am not sure.

filiphr:
I just remembered one more thing. When there are ambiguous factory methods how can a user select one? Or we don't support this?

processor/src/test/java/org/mapstruct/ap/test/bugs/_1131/Issue1131MapperWithContext.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper
+public abstract class Issue1131MapperWithContext {
+    public static final Issue1131MapperWithContext INSTANCE = Mappers.getMapper( Issue1131MapperWithContext.class );
+
+    public static class MappingContext {
+        @ObjectFactory
```
agudian: This annotation can be removed, as we don't look for `@ObjectFactory` methods on `@Context` types (only for before/after-mapping methods).

agudian:
> I reversed the change on the `InheritanceSelector`. However, I am in doubt what is correct there.

I think that is the right thing. The InheritanceSelector is used for only those methods with one single source type (can only be a mapping method) and chooses the best fit. For lifecycle methods, all applicable methods are used, and for factory methods we simply didn't implement it that way and would throw the ambiguity error right now. Not sure if that's a wise choice, but it's nothing we can't fix later on.

agudian:
> I just remembered one more thing. When there are ambiguous factory methods how can a user select one? Or we don't support this?

The `@Qualifier` annotations or `@Named` annotation can be used for factory methods as well... (for that we do have tests, stumbled upon them yesterday...)

agudian:
Okay, the code looks good to me (you decide on the Assignment / SourceRHS thing, I'm not that deep into it that I could form an opinion... 😉 ).

WDYT, should we also add a sentence or something to the Javadoc or the reference docs?

filiphr:
>The `@Qualifier` annotations or `@Named` annotation can be used for factory methods as well... (for that we do have tests, stumbled upon them yesterday...).

You are right. The only thing now is that this won't work for update methods... Should we take this somehow into consideration? We would need to get the `MappingOptions` from the method in the assignment if possible...

> (you decide on the Assignment / SourceRHS thing, I'm not that deep into it that I could form an opinion... 😉 ).

I'll see how it looks with the `SourceRHS` 😄. I am not sure what I prefer more, I don't like the change of the package that much.


>WDYT, should we also add a sentence or something to the Javadoc or the reference docs?

In the Javadoc of what? Maybe we can add it in the Object Factories section of the reference docs, and I would definitely write something in the release notes on GitHub. In behavior changes.

agudian:
> You are right. The only thing now is that this won't work for update methods...

Hmm, right. How about using the `@Mapping#qualifiedBy` here as well? That shouldn't interfere with selecting mapping methods, 

> In the Javadoc of what? Maybe we can add it in the Object Factories section of the reference docs, and I would definitely write something in the release notes on GitHub. In behavior changes.

In the Javadoc of `@ObjectFactory`, as that specifically says "a mapping methods source parameter" - we could instead write "the mapping source parameter" (removing the "method", which makes the meaning more local to the thing that is currently being mapped)... Would that be enough? I'd like to avoid too complicated stuff in Javadocs, as people have a tendency to not read it if it's too long... 😁  The reference docs probably contain the same wording.

filiphr:
>Hmm, right. How about using the `@Mapping#qualifiedBy` here as well? That shouldn't interfere with selecting mapping methods,

I don't think that it would be good to do that. In the Javadoc of `@Mapping#qualifiedBy` it says that it is used for selecting mapping methods. Maybe leave it for now as I think that there are some other places that are affected by exactly the same thing.

Looks good for the Javadoc. I'll do some updates.

One more thing which I remembered is that we are creating the `UpdateWrapper` on 2 other places:
* `CollectionAssignmentBuilder` for creating `Iterable`(s), `Map`(s)
* `ConstantMappingBuilder`

I think that we need to do it for those as well. WDYT? This would lead make it obligatory to use the `Assignment`

filiphr:
For the `ConstantMappingBuilder` it is OK as we don't have a type that we could use. However, for the `CollectionAssignmentBuilder` I think that we need to consider adding something.

filiphr:
Some final touches 😄.

* The `CollectionAssignmentBuilder` now uses the same type of factory methods for updated
* We must use the `SourceRHS` for the factory method because we need the getter on the source. If we use the assignment than we can get some of the wrapper which is not what we need.
* I had to move both the `Assignment` and `SourceRHS` to the `org.mapstruct.ap.internal.model.common` package as there were cycles if I left `SourceRHS` where it was, or if I moved only that one.
* I also update the reference documentation and the Javadoc on the `@ObjectFactory`

@agudian Can you please have one more look at this one 😄. Thanks a lot.

filiphr:
Squashed and merged. Thanks a lot for the review @agudian

PR #1239, opened 2017-07-02
#543 Make sure that the bound type is imported if the type is wildcard

### LINKED ISSUE #543: wildcard generics generated code contains fully qualified names.
Checkout generated code from `org.mapstruct.ap.test.collection.wildcard`.

filiphr:
Fixes #543


gunnarmorling:
Applied - thanks, @filiphr! I've added one more commit to add an assertion that the import created as expected.

PR #1241, opened 2017-07-09
#1196 Fix build problems on Netbeans

### LINKED ISSUE #1196: new surefire argline breaks debugging in netbeans
The following change (but then commented in) makes in-IDE-debugging impossible in netbeans..

```java
            <plugin>
                <artifactId>maven-surefire-plugin</artifactId>
<!--                <configuration>
                     to get jacoco report we need to set argLine in surefire,
                        without this snippet the jacoco argLine is lost
                    <argLine>@{argLine} -Xms1024m -Xmx3072m</argLine>
                </configuration>-->
            </plugin>
```

I know this was added for jacocco., but I think we should tweak the configuration a  little bit (lifecycle?). Don't know how Eclipse behaves.

filiphr:
Fixes #1196.

This setup works on Netbeans (installed and tested by me). I think that it'll work on Travis.

Another possible solution would be the user to set an empty `argLine`,  or the `propertyName` that we would define for jacoco in their own `settings.xml`. We can write this on our website or even in our `README`

### DIFF
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -35,6 +35,9 @@
 
     <properties>
         <prism.directory>${project.build.directory}/generated-sources/prims</prism.directory>
+        <!-- Netbeans has a problem when we use late binding with @ in the surefire arg line.
+            Therefore we set this empty property here-->
+        <jacocoArgLine />
     </properties>
 
     <dependencies>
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Thanks, @filiphr!

sjaakd:
👍

Issue #1244, opened 2017-07-10
mapping does not work when parameter name is 'size'


grumpy-programmer:
Related to #909
mapstruct 1.2.0.Beta3
lombok 1.16.14

Example code:
```
import lombok.Data;
import lombok.Generated;

@Mapper(componentModel = "spring")
public interface SizeMapper {

    @Mapping(source = "width", target = "width")
    @Mapping(source = "height", target = "height")
    SizeDto toSizeDto(Size size);
}

@Data
@Generated
public class SizeDto {

    private Integer width;

    private Integer height;
}

@Data
@Generated
public class Size {

    private Integer width;

    private Integer height;
}
```
Compilation exception:

[...]

PR #1245, opened 2017-07-10
#1244 Fix problems with special word for FreeMarker in some cases

filiphr:
FreeMarker has problems getting values from a map when the key
is "values" or "size" (there might be other cases, this are the ones we know).
Therefore, instead of using a Map, MapStruct will pass the Parameter to the
function and return List<PropertyMapping>

Fixes: #1244

I decided for a different approach in solving this, as there might be other cases where FreeMarker fails. It should be much safer like this. We just pass the `Parameter` to a function and we get back a `List<PropertyMapping>`.


filiphr:
>👍 Way better approach then the way I did with the `values` problem.

Thanks, I wanted to find the root cause and it seemed not that complex 😄. I think that for now we can be sure that there are not other FreeMarker specialties lurking around 😁

PR #1246, opened 2017-07-10
#1242 Fix favoring of a single factory method with source params before others without source params.

### LINKED ISSUE #1242: Ambiguous factory method for @ObjectFactory with and without parameters
In the javadoc of the new `@ObjectFactory` we state:

> If there are two factory methods, both serving the same type, one with no parameters and one taking sources as input, then the one with the source parameters is favored. If there are multiple such factories, an ambiguity error is shown.

However, when that situation actually occurs. Currently we have compile error ambiguous factory methods. Should we update the implementation or the Javadoc?

agudian:
From the Javadoc of `@ObjectFactory`: If there are two factory methods, both serving the same type, one with no parameters and one taking sources as input, then the one with the source parameters is favored. If there are multiple such factories, an ambiguity error is shown.

Fixes #1242.

The test relies on #1237 being merged, so the PR will fail initially. I'll rebase after the review and after the other PR is merged.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/FactoryParameterSelector.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/FactoryParameterSelector.java
@@ -1,0 +1,62 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model.source.selector;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.model.source.Method;
+
+/**
+ * For factory methods, the candidate list is checked if it contains a method with a source parameter which is to be
+ * favored compared to factory methods without a source parameter. It returns the original list of candidates in case of
+ * ambiguities.
+ *
+ * @author Andreas Gudian
+ */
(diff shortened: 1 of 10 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/FactoryParameterSelector.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public class FactoryParameterSelector implements MethodSelector {
+
+    @Override
+    public <T extends Method> List<SelectedMethod<T>> getMatchingMethods(Method mappingMethod,
+            List<SelectedMethod<T>> methods,
+            List<Type>sourceTypes,
+            Type targetType,
+            SelectionCriteria criteria) {
+        if ( !criteria.isObjectFactoryRequired() || methods.size() <= 1 ) {
```
filiphr: For the future I think that we need to extract a `supports` method on the `MethodSelector` almost all of them have some condition that just returns. However, it might be an overkill as well.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1242/Issue1242Test.java
```diff
@@ -0,0 +1,83 @@
...
+    TargetFactories.class
+})
+public class Issue1242Test {
+    @Test
+    public void factoryMethodWithSourceParamIsChosen() {
+        SourceA sourceA = new SourceA();
+        sourceA.setB( new SourceB() );
+
+        TargetA targetA = new TargetA();
+        Mappers.getMapper( Issue1242Mapper.class ).mergeA( sourceA, targetA );
```
filiphr: Does this slow down the tests a bit?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/FactoryParameterSelector.java
agudian: Yeah, maybe. If it makes stuff more readable for us 🙂

processor/src/test/java/org/mapstruct/ap/test/bugs/_1242/Issue1242Test.java
agudian: What exactly do you mean?

filiphr:
@agudian can you please rebase this on master. There is some issue with codecov. I don't know why but the commit it is picking up is not correct.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1242/Issue1242Test.java
filiphr: Well we call twice `Mapper.getMapper( Issue1242Mapper.class )` . Don't we recommend users having a `public static final` field to store the reference. In any case, I didn't mean it for you to change it, just wanted to mention it.

agudian:
Rebased :)

filiphr:
Merged, thanks for the PR @agudian

Issue #1085, opened 2017-02-19
Investigate why there are no property entries in the source reference


filiphr:
In #1072 there is a TODO in the `NestedTargetPropertyMappingHolder`. It seems that it is possible that the property entries of a non-`null` valid `SourceReference` can be empty. Investigate why is that the case

filiphr:
If the mapping is ignore, constant and/or expression then the `SourceReference` is `null`

sjaakd:
and add meaningful comment why :smile:

filiphr:
I know why they are empty. When we map the source parameter to some property then we have a valid `SourceReference` but no property entries 😄. See also PR #1248.


I'll close this one.

PR #1248, opened 2017-07-12
#1247 Make sure that nested target mappings works correctly for multiple source parameters

### LINKED ISSUE #1247: Several mappings ignored when defined another mapping with embedded paths
when try to map `DtoIn` to `DtoOut`
```
@Data
public class DtoIn {
    private String data;
    private String data2;
}
```

```
@Data
public class DtoOut {
    private String data;
    private InternalDto internal;
}
```

```
@Data
public class InternalDto {
    private String data2;
    private InternalData internalData;
}
```

```
@Data
public class InternalData {
    private List<String> list;
}
```

with mapper
```
@Mapper
public interface DtoMapper {

    @Mapping(target = "internal", source = "in")
    @Mapping(target = "internal.internalData.list", source = "list")
    DtoOut map(DtoIn in, List<String> list);
}
```

generates:
```
public class DtoMapperImpl implements DtoMapper {

    @Override
    public DtoOut map(DtoIn in, List<String> list) {
        if ( in == null && list == null ) {
            return null;
        }

        DtoOut dtoOut = new DtoOut();

        if ( in != null ) {
            dtoOut.setData( in.getData() );
        }
        if ( list != null ) {
            dtoOut.setInternal( stringListToInternalDto( list ) );
        }

        return dtoOut;
    }

    protected InternalData stringListToInternalData(List<String> list) {
        if ( list == null ) {
            return null;
        }

        InternalData internalData = new InternalData();

        List<String> list1 = list;
        if ( list1 != null ) {
            internalData.setList( new ArrayList<String>( list1 ) );
        }
        else {
            internalData.setList( null );
        }

        return internalData;
    }

    protected InternalDto stringListToInternalDto(List<String> list) {
        if ( list == null ) {
            return null;
        }

        InternalDto internalDto = new InternalDto();

        internalDto.setInternalData( stringListToInternalData( list ) );

        return internalDto;
    }
}
```

mapping `@Mapping(target = "internal", source = "in")` was ignored

filiphr:
Fixes #1247

This one also clarifies #1085.  A valid `SourceReference` can have no property entries when the entire source parameter is mapped to some property.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
@@ -19,8 +19,9 @@
 package org.mapstruct.ap.internal.model;
 
 import java.util.ArrayList;
-import java.util.LinkedHashMap;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -56,7 +57,8 @@ public class NestedTargetPropertyMappingHolder {
         Extractor<SourceReference, PropertyEntry>() {
             @Override
             public PropertyEntry apply(SourceReference sourceReference) {
-                return first( sourceReference.getPropertyEntries() );
+                return sourceReference.getPropertyEntries().isEmpty() ? null :
+                    first( sourceReference.getPropertyEntries() );
             }
         };
 
@@ -117,6 +119,8 @@ public class NestedTargetPropertyMappingHolder {
         private Method method;
         private MappingBuilderContext mappingContext;
         private Set<String> existingVariableNames;
+        private List<PropertyMapping> propertyMappings;
+        private Set<String> handledTargets;
 
         public Builder method(Method method) {
             this.method = method;
@@ -135,8 +139,8 @@ public class NestedTargetPropertyMappingHolder {
 
         public NestedTargetPropertyMappingHolder build() {
             List<Parameter> processedSourceParameters = new ArrayList<Parameter>();
-            Set<String> handledTargets = new HashSet<String>();
-            List<PropertyMapping> propertyMappings = new ArrayList<PropertyMapping>();
+            handledTargets = new HashSet<String>();
+            propertyMappings = new ArrayList<PropertyMapping>();
 
             // first we group by the first property in the target properties and for each of those
             // properties we get the new mappings as if the first property did not exist.
@@ -164,7 +168,7 @@ public class NestedTargetPropertyMappingHolder {
                     // Lastly we need to group by the source references. This will allow us to actually create
                     // the next mappings by popping source elements
                     GroupedSourceReferences groupedSourceReferences = groupByPoppedSourceReferences(
-                        entryByParam.getValue(),
+                        entryByParam,
                         groupedByTP.singleTargetReferences.get( targetProperty )
                     );
 
@@ -229,6 +233,13 @@ public class NestedTargetPropertyMappingHolder {
                         handledTargets.add( entryByTP.getKey().getName() );
                     }
 
+                    handleSourceParameterMappings(
+                        groupedSourceReferences.sourceParameterMappings,
+                        targetProperty,
+                        sourceParameter,
+                        forceUpdateMethod
+                    );
+
                     unprocessedDefinedTarget.put( targetProperty, groupedSourceReferences.notProcessedAppliesToAll );
                 }
             }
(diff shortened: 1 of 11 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -516,10 +582,55 @@ else if ( mappingsKeyedByProperty.isEmpty() && nonNested.isEmpty() ) {
...
                 nonNested,
-                notProcessedAppliesToAll
+                notProcessedAppliesToAll,
+                sourceParameterMappings
             );
         }
 
+        /**
+         * Extracts all relevant single target references and populates the {@code sourceParameterMappings} if needed.
+         * A relevant single target reference mapping is a mapping that has a valid source reference and the is for
```
agudian: “... and the is for...” - something is missing there 🙂

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -496,7 +547,22 @@ else if ( mapping.getSourceReference() == null ) {
...
+            Parameter sourceParameter = entryByParam.getKey();
+            List<Mapping> singleTargetReferencesToUse =
+                extractSingleTargetReferencesToUseAndPopulateSourceParameterMappings(
+                    singleTargetReferences,
+                    sourceParameterMappings,
+                    hasNoMappings,
+                    sourceParameter
+                );
+
+
```
agudian: Two empty lines 😉

processor/src/test/java/org/mapstruct/ap/test/bugs/_1247/Issue1247Mapper.java
```diff
@@ -0,0 +1,42 @@
...
+@Mapper
+public interface Issue1247Mapper {
+
+    Issue1247Mapper INSTANCE = Mappers.getMapper( Issue1247Mapper.class );
+
+    @Mappings( {
+        @Mapping(target = "internal", source = "in"),
+        @Mapping(target = "internal.internalData.list", source = "list")
+    } )
+    DtoOut map(DtoIn in, List<String> list);
```
agudian: It’s pretty amazing that this actually does what it’s supposed to do 😳

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: Checkstyle not working again 😄.  I'll fix it

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: It should be "... and is for ...". Is that clear enough, or should I reword it?

processor/src/test/java/org/mapstruct/ap/test/bugs/_1247/Issue1247Mapper.java
filiphr: I was also surprised, this things are really pushing the limits of MapStruct, I am glad that we are able to do it 😄

filiphr:
Thanks @agudian. I've spent so much time on this, it is a bit easier now 😄. I am really glad that MapStruct can do these really complex mappings.

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
agudian: That’s okay 👌

agudian:
Alright, cool! Go ahead and merge if you’re done with the cosmetics 👍

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -229,6 +233,13 @@ public NestedTargetPropertyMappingHolder build() {
                         handledTargets.add( entryByTP.getKey().getName() );
                     }
 
+                    handleSourceParameterMappings(
```
sjaakd: so, IIUC, if no nesting level remains on source side, you assume the source parameter to be mapped directly?

processor/src/test/java/org/mapstruct/ap/test/bugs/_1247/Issue1247Mapper.java
sjaakd: should we throw in some constants, expressions and the likes as well?

BTW: @agudian : I agree..  However, it has grown a bit too complex. I think we need to consider some serious refactoring after 1.3 (getting rid of this dreaded enum solution, which is now making it unnecessary complex first).

sjaakd: Perhaps even break with the current enum mappings. And yes.. We should go for java 8 and introduce lamda's in our own code as well.

sjaakd: I'm so happy with the test cases. They allow for refactoring.

sjaakd:
@filiphr .. I was just thinking about the relation to #1148. I remember that we discussed whether we should generate source parameter to nested targets whenever `popping the nested source level ends`. 

In my solution I was struggling to fix this. Does this by coincidence fix this as well?

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: Yes, with this one for some reason a source parameter is into the groupings, but it was not included in the mappings before I assume that the source parameter is mapped to the target property. For this reason the mappings are empty (if they are not it would be treated as properties of the parameter)

processor/src/test/java/org/mapstruct/ap/test/bugs/_1247/Issue1247Mapper.java
filiphr: I can add some methods with `contstants`, `expressions`, etc. Maybe use some other target property so we keep this consistent.

filiphr:
>@filiphr .. I was just thinking about the relation to #1148. I remember that we discussed whether we should generate source parameter to nested targets whenever popping the nested source level ends.
>
>In my solution I was struggling to fix this. Does this by coincidence fix this as well?

I think that this covers that case as well. As now if this was in the nested mappings and there are no other alternative we would do a mapping between the source parameter and the target property

sjaakd:
> I think that this covers that case as well. As now if this was in the nested mappings and there are no other alternative we would do a mapping between the source parameter and the target property

Ok.. I remember we had a discussion on `AbstractBaseBuilder.java`  on this `methodRef.getMappingOptions().isRestrictToDefinedMappings()`.. Is that check stil needed?

filiphr:
Yes that is still needed.

sjaakd:
Then I've no further comments. Nice piece of work.! 👍 Thanks @filiphr ..

filiphr:
Thanks @sjaakd. I've updated the test with some more examples. I'll leave the build to finish and I can do the merge tomorrow evening.

filiphr:
PR integrated into `master` thanks a lot for the review.

PR #1252, opened 2017-07-16
#1251 Deploy SNAPSHOTS from Travis CI

### LINKED ISSUE #1251: Configure Travis to deploy SNAPSHOT versions
We need to configure Travis CI to do deployment when something is commited on master

filiphr:
Fixes #1251.

Merge only after the `SONATYPE_USERNAME` and `SONATYPE_PASSWORD` have been added on Travis. [Here](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings) can be seen how to do them. I can do it as well, we just need to agree how we are going to do it (add a new mapstruct-ci user to sonatype, or somehow different).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,13 @@ script: mvn clean install -DprocessorIntegrationTest.toolchainsFile=etc/toolchai
 after_success:
   - mvn jacoco:report && bash <(curl -s https://codecov.io/bash) || echo "Codecov did not collect coverage reports"
 
+deploy:
+  provider: script
+  script: "test ${TRAVIS_TEST_RESULT} -eq 0 && mvn -s etc/travis-settings.xml -DskipTests=true deploy"
+  skip_cleanup: true
+  on:
+    branch: master
+
 sudo: false
 cache:
   directories:
--- a/etc/travis-settings.xml
+++ b/etc/travis-settings.xml
@@ -1,0 +1,11 @@
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
+    <servers>
+        <server>
+            <id>sonatype-nexus-snapshots</id>
+            <username>${env.SONATYPE_USERNAME}</username>
+            <password>${env.SONATYPE_PASSWORD}</password>
+        </server>
+    </servers>
+</settings>
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -534,6 +534,7 @@
                         <exclude>**/LICENSE.txt</exclude>
                         <exclude>**/mapstruct.xml</exclude>
                         <exclude>**/toolchains-*.xml</exclude>
+                        <exclude>**/travis-settings.xml</exclude>
                         <exclude>**/eclipse-formatter-config.xml</exclude>
                         <exclude>**/forbidden-apis.txt</exclude>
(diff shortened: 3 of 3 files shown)


.travis.yml
```diff
@@ -6,6 +6,13 @@ script: mvn clean install -DprocessorIntegrationTest.toolchainsFile=etc/toolchai
 after_success:
   - mvn jacoco:report && bash <(curl -s https://codecov.io/bash) || echo "Codecov did not collect coverage reports"
 
+deploy:
+  provider: script
+  script: "test ${TRAVIS_TEST_RESULT} -eq 0 && mvn -s etc/travis-settings.xml -DskipTests=true deploy"
```
filiphr: Maybe we need to add some "quick" profile that will skip all the checks so this line goes faster.

agudian: Can’t we just use mvn clean deploy when building from master to avoid the second mvn call? Or can’t we distinguish that in the travis.yml?

filiphr: Theoretically we can, I just have to check how we can define different build script conditionally, we don't want to deploy from PRs

gunnarmorling:
My preference is to have a dedicated "mapstruct-ci" user whose credentials are used. Has anyone requested such account already, or should I do it?

agudian:
Would be nice if you created one, Gunnar, so that you can keep control of the account through your email address...

gunnarmorling:
Ok, I've created an account and asked to assign it to our group id. I'll ping you once it's set up.

gunnarmorling:
Btw. this should also allow us a few more things:

* Set up a job on Travis for doing releases (instead of doing them locally)
* Set up a job for re-publishing the website after a commit was pushed to master

agudian:
I'd be curious about your ideas of how to kick off a release build using Travis. I'm no big fan of pushing a release on each commit to master (for various reasons, including that Maven Central shouldn't be polluted that way). So we'd need a mechanism to start the release goals...
I drafted a rough idea on how such a thing could look like here: adobe/S3Mock#1 - any thoughts on that?

gunnarmorling:
I don't have a good answer in terms of what's doable on Travis, but I can
describe what we do on Jenkins for Hibernate: there we have a parameterized
build job which is started manually when wishing to do a release. It takes
parameters for the version to be released, the new (snapshot) version and
things like that. Then everything happens automatically: doing the release,
creating a tag, uploading the distribution, deploying to nexus etc. The
only thing to be done by hand is writing the announcement. It would be
great if the same could be done on Travis.

agudian:
That's the thing, on Travis you can't define an arbitrary parameterized job to be triggered manually. At least not that I know of... :-/

gunnarmorling:
Agh, I see :/

filiphr:
Did you link the `mapstruct-ci` user with our @mapstruct-robot? 

Regarding the publishing, an interesting auto publishing tool is [shipkit](https://github.com/mockito/shipkit). Maintained by the Mockito team and used by Mockito for publishing to Maven Central. It automates quite a lot.

Regarding Travis CI, one can trigger a build via an API, see [here](https://docs.travis-ci.com/user/triggering-builds/). Theoretically, we could use that to trigger a release build. I think that a parameterized build could be performed by using the API.

gunnarmorling:
Interesting note on the API usage. Would be fun to explore, though I'm not sure whether it's worth the effort. It's not that much to automate here, I think. WDYT?

filiphr:
It would be interesting if someone does an UI for the Travis API 😄  (maybe it already exists). Doing the release locally it's not complex, just run a command and it is done 😄 , I think for now we are OK not having it automated 100%

gunnarmorling:
+1

filiphr:
All done, thanks all for the review 😄 .

Nest steps use the mapstruct-ci user to publish the website immediately on pushes to master 😄

PR #1263, opened 2017-07-30
Constructor Injection for Annotation Based Component Model

kevcodez:
This PR replaces field injection with constructor injection for annotation based component model processors (Spring, CDI, JSR330).

Example of the generated source:

```java
public class Jsr330SourceTargetMapperImpl implements Jsr330SourceTargetMapper {

    private CustomMapper customMapper;
    private org.mapstruct.ap.test.references.samename.b.CustomMapper customMapper1;
 
    @Inject
    public Jsr330SourceTargetMapperImpl(CustomMapper customMapper, org.mapstruct.ap.test.references.samename.b.CustomMapper customMapper1) {
        this.customMapper = customMapper;
        this.customMapper1 = customMapper1;
    }
...
}
```

```java
public class PersonMapperImpl extends PersonMapperDecorator implements PersonMapper {

    @Inject
    public PersonMapperImpl(@Named("org.mapstruct.ap.test.decorator.jsr330.PersonMapperImpl_") PersonMapper delegate) {
        this.delegate = delegate;
    }
```

@gunnarmorling @filiphr

Please note, the processing tests are working just fine, however, since this is the first time I am making a change in MapStruct, I may have forgot something.  Please review to make sure I did not miss anything and the direction I am heading here is fine.

Furthermore, I do not know if Guice or Spring need a public constructor, just as CDI does. I added the `publicEmptyConstructor` property to do just this. If Spring does NOT need one, contrary to CDI, the SpringComponentProcessor can just overwrite the getter.

I will be on vacation for the next week, I will finish this task and implement review feedback once I return ;)


filiphr:
I wrote in a review comment, but also here. Spring and Guice don't need an empty constructor. Therefore, I would make the `publicEmptyConstructor` default to `false` and only set it to `true` where it is needed (CDI).

I am not sure if we want to add fixture tests for this or some other tests....

kevcodez:
@filiphr really appreciate the feedback. I agree to all points and will change in a week. 

I would suggest to introduce an option to use either field or constructor injection. All tests, processing and integration work fine with constructor injection without a single change in tests, IMO it should be the default one in a newer release, maybe start with field injection and change the default upon release. 

When supporting both, setter and constructor injection, do you see any further issue here?

processor/src/main/resources/org/mapstruct/ap/internal/model/AnnotationMapperReference.ftl
```diff
@@ -19,7 +19,4 @@
      limitations under the License.
 
 -->
-<#list annotations as annotation>
-<#nt><@includeModel object=annotation/>
-</#list>
 private <@includeModel object=type/> ${variableName};
```
kevcodez: This can only be final, if there is no default public constructor, as in CDI. I will change, so that the fields are final, if no public constructor is needed

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -99,23 +121,100 @@ protected void adjustDecorator(Mapper mapper) {
...
+                annotationMapperReference.getVariableName(),
+                qualifiers,
+                annotationMapperReference.isUsed() );
+
+            mapperReferenceIterator.add( annotationMapperReference1 );
+        }
+    }
+
+    protected boolean publicEmptyConstructorWhenConstructorInjectionIsUsed() {
+        return true;
```
kevcodez: Sure, will make the default false

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -99,23 +121,100 @@ protected void adjustDecorator(Mapper mapper) {
...
+            mapperReferenceIterator.remove();
+
+            List<Annotation> qualifiers = new ArrayList<Annotation>();
+            for ( Annotation annotation : annotationMapperReference.getAnnotations() ) {
+                if ( !mapperReferenceAnnotationsTypes.contains( annotation.getType() ) ) {
+                    qualifiers.add( annotation );
+                }
+            }
+
+            AnnotationMapperReference annotationMapperReference1 = new AnnotationMapperReference(
```
kevcodez: Indeed, will fix

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -99,23 +121,100 @@ protected void adjustDecorator(Mapper mapper) {
...
+        ListIterator<AnnotationMapperReference> mapperReferenceIterator = annotationMapperReferences.listIterator();
+
+        Set<Type> mapperReferenceAnnotationsTypes = new HashSet<Type>();
+        for ( Annotation annotation : mapperReferenceAnnotations ) {
+            mapperReferenceAnnotationsTypes.add( annotation.getType() );
+        }
+
+        while ( mapperReferenceIterator.hasNext() ) {
+            MapperReference mapperReference = mapperReferenceIterator.next();
+            AnnotationMapperReference annotationMapperReference = (AnnotationMapperReference) mapperReference;
```
kevcodez: Must have overseen that when refactoring, will fix

processor/src/main/resources/org/mapstruct/ap/internal/model/AnnotationMapperReference.ftl
filiphr: Yes you are right. I forgot about CDI.

filiphr:
@kevcodez the only issue is not to have too many options, as it would make things more complicated. I know that with setter injection we would solve cyclic dependency problem and make the generated classes more testable. However, I am a big opponent of cyclic dependency and if one wants good testable code, then one should have a clean solution (of course this is my personal opinion).

For the option, we could add a new `enum` in the API and have a field in the `@Mapper` and `@MapperConfig` annotations that can be used to configure. Maybe something like `injectionType` and the enum can have values: `DEFAULT`, `FIELD`, `CONSTRUCTOR`, or without the `DEFAULT`.  In the first beta phase we would use `FIELD` as default, but then we would switch to `CONSTRUCTOR` and in the end the release should use `CONSTRUCTOR`. I think we need something more granular, because of the cycle dependencies, the users should be able to change the injection type on a case, by case level.

afdia:
Just a quick note about Dagger (because the previous comments only mentioned Spring, Guice, CDI):

Dagger2 requires you to write a constructor which is annotated with `@Inject` otherwise it will not work (see https://google.github.io/dagger//users-guide.html )

> Classes that lack `@Inject` annotations cannot be constructed by Dagger.

This was also true for Dagger1 (see https://square.github.io/dagger/)

> Classes that lack `@Inject` annotations cannot be constructed by Dagger, even if they have a no-argument constructor.

Therefore I would suggest to set publicEmptyConstructor to true and make sure the empty constructor is also always annotated with `@Inject` or at least the option should not be limited to CDI.

kevcodez:
@afdia Thanks for the hint! The doc states that you need to add the `@Inject` parameter to the constructor. It does not seem to state that a second, empty constructor is needed. Can you elaborate?

@filiphr Regarding the configuration, how can I access the configuration in the processor? I found the Options class, that can be accessed in the processor, but I am unsure if this is the correct and only spot to get that information.

afdia:
Maybe I misunderstood what the publicEmptyConstructor option does.
You're right that Dagger doesn't need a second empty constructor, but the difference to other DI frameworks such as Guice is the following:

Guice will inject any class which has no constructor or just a single no-arg constructor without any annotation (therefore the `@Inject` annotation is just used to specify which constructor to use if none of these 2 conditions are met). Dagger requires you to always write one constructor which is annotated with `@Inject`.
So classes without constructor or with a single no-arg constructor would work with Guice but not with Dagger.

I thought the option is used to create a `@Inject` annotated constructor for classes which would otherwise have no constructor at all.

kevcodez:
@afdia The public empty constructor is needed in CDI **in addition** to the constructor annotated with `@Inject` and the fields to be injected as parameters. I should probably rename that.

kevcodez:
@filiphr Made the injection strategy configurable. The default injection strategy is field injection for now.

Two things we should probably discuss:

1. Testing

I could write a few additional tests for constructor and field injection. Otherwise, we could maybe run every annotation based component model test with both, field and constructor injection. What do you think?

IMO, I should write some additional test cases (e.g. for the case below). Since constructor injection is going to be the default in a future release, the tests are sooner or later going to run on constructor injection only, anyway.

2. Injecting additional fields when using Constructor Injection.

The constructor injection works fine for all current tests, including decorators (example in first comment). However, imagine following mapper

```java
@Mapper(componentModel = "spring", injectionStrategy = CONSTRUCTOR, uses = { BarMapper1.class, BarMapper2.class }
abstract class FooMapper {
   
    private final FooService fooService;

    @Inject
    public FooMapper(@MyQualifier FooService fooService) {
        this.fooService = fooService;
    }

    // use fooService
}
```

How do we want to handle this? Similar to a decorator, generate multiple classes?

When everything is fine, I will also change the documentation and add some JavaDoc.

processor/src/main/java/org/mapstruct/ap/internal/model/AnnotatedConstructor.java
```diff
@@ -0,0 +1,79 @@
...
+        this.annotations = annotations;
+        this.publicEmptyConstructor = publicEmptyConstructor;
+    }
+
+    @Override
+    public Set<Type> getImportTypes() {
+        Set<Type> types = new HashSet<Type>();
+
+        for ( MapperReference mapperReference : mapperReferences ) {
+            types.add( mapperReference.getType() );
```
filiphr: Use `mapperReference#getImportTypes()` instead

processor/src/main/java/org/mapstruct/ap/internal/model/AnnotatedConstructor.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        for ( Annotation annotation : annotations ) {
+            types.add( annotation.getType() );
```
filiphr: Use `mapperReference#getImportTypes()` instead

processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
```diff
@@ -169,6 +169,10 @@ public Accessibility getAccessibility() {
         return accessibility;
     }
 
+    public void setConstructor(Constructor constructor) {
+        this.constructor = constructor;
```
filiphr: We need to call `getImportTypes` on the `Constructor`. For that we also need to add the signature to the `Constructor` interface as well.

processor/src/main/java/org/mapstruct/ap/internal/prism/InjectionStrategyPrism.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.prism;
+
+public enum InjectionStrategyPrism {
```
filiphr: You can add the `@author` tag here 😄

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -50,8 +55,11 @@
     public Mapper process(ProcessorContext context, TypeElement mapperTypeElement, Mapper mapper) {
         this.typeFactory = context.getTypeFactory();
 
-        String componentModel = MapperConfiguration.getInstanceOn( mapperTypeElement )
-                                                   .componentModel( context.getOptions() );
+        String componentModel =
+            MapperConfiguration.getInstanceOn( mapperTypeElement ).componentModel( context.getOptions() );
```
filiphr: You can extract `MapperConfiguration.getInstanceOn( mapperTypeElement )` into a variable and use it instead of creating it twice

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -65,21 +73,42 @@ public Mapper process(ProcessorContext context, TypeElement mapperTypeElement, M
...
+        ListIterator<MapperReference> iterator = referencedMappers.listIterator();
         while ( iterator.hasNext() ) {
             MapperReference reference = iterator.next();
             iterator.remove();
-            iterator.add( replacementMapperReference( reference, annotations ) );
+            iterator.add( replacementMapperReference( reference, annotations, injectionStrategy ) );
+        }
+
+        if ( injectionStrategy == InjectionStrategyPrism.CONSTRUCTOR ) {
+            if ( !referencedMappers.isEmpty() ) {
```
filiphr: I would extract everything between lines 90 and 105 into a new method

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -88,34 +117,119 @@ protected void adjustDecorator(Mapper mapper) {
...
             for ( Field field : decorator.getFields() ) {
-                replacement.add( replacementMapperReference( field, annotations ) );
+                replacement.add( replacementMapperReference( field, annotations, injectionStrategy ) );
             }
         }
         decorator.setFields( replacement );
     }
 
+    private AnnotatedConstructor buildAnnotatedConstructorForMapper(Mapper mapper) {
+        List<AnnotationMapperReference> mapperReferencesForConstructor = new ArrayList<AnnotationMapperReference>();
```
filiphr: You can use the size of the referenced mappers as initial size of the list

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -88,34 +117,119 @@ protected void adjustDecorator(Mapper mapper) {
...
+
+        return new AnnotatedConstructor(
+            mapper.getName(),
+            mapperReferencesForConstructor,
+            mapperReferenceAnnotations,
+            additionalPublicEmptyConstructor() );
+    }
+
+    private AnnotatedConstructor buildAnnotatedConstructorForDecorator(Decorator decorator) {
+        List<AnnotationMapperReference> mapperReferencesForConstructor = new ArrayList<AnnotationMapperReference>();
```
filiphr: You can use the size of the fields as an initial size for the list

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -88,34 +117,119 @@ protected void adjustDecorator(Mapper mapper) {
...
+            mapperReferenceAnnotations,
+            additionalPublicEmptyConstructor() );
+    }
+
+    /**
+     * Removes constructor parameter annotations that are already on the constructor. For example, for CDI, the
+     * javax.inject.Inject annotation is on the constructor and must not be on the constructor parameters.
+     *
+     * @param annotationMapperReferences list of annotations that are be on the constructor parameter
+     * @param mapperReferenceAnnotations annotations that are be on the constructor
```
filiphr: I would write here `annotations to annotate the constructor with`

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -88,34 +117,119 @@ protected void adjustDecorator(Mapper mapper) {
...
+            mapperReferencesForConstructor,
+            mapperReferenceAnnotations,
+            additionalPublicEmptyConstructor() );
+    }
+
+    /**
+     * Removes constructor parameter annotations that are already on the constructor. For example, for CDI, the
+     * javax.inject.Inject annotation is on the constructor and must not be on the constructor parameters.
+     *
+     * @param annotationMapperReferences list of annotations that are be on the constructor parameter
```
filiphr: The description here is a bit confusing to me. At the moment, I don't have an idea for something better 😟

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -88,34 +117,119 @@ protected void adjustDecorator(Mapper mapper) {
...
+
+            List<Annotation> qualifiers = new ArrayList<Annotation>();
+            for ( Annotation annotation : annotationMapperReference.getAnnotations() ) {
+                if ( !mapperReferenceAnnotationsTypes.contains( annotation.getType() ) ) {
+                    qualifiers.add( annotation );
+                }
+            }
+
+            mapperReferenceIterator.add(
+                new AnnotationMapperReference(
```
filiphr: I would extract a new method on the `AnnotationMapperReference` where you can pass the new annotations and you get a new instance (in order to reduce this class a bit).

It can be something like:

```java
public AnnotationMapperReference withNewAnnotations(List<Annotation> annotations) {
        return new AnnotationMapperReference(
            getType(),
            getVariableName(),
            annotations,
            isUsed(),
            fieldFinal,
            includeAnnotationsOnField
        );
}
```

processor/src/main/java/org/mapstruct/ap/internal/util/MapperConfiguration.java
```diff
@@ -155,6 +156,15 @@ public NullValueCheckStrategyPrism getNullValueCheckStrategy() {
         }
     }
 
+    public InjectionStrategyPrism getInjectionStrategy() {
```
filiphr: I think that we need to have a test where we would test this as well. Not a unit test, but via the `@MapperConfig`

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
kevcodez: Will try to come up with something better

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
kevcodez: Good idea, will do

processor/src/main/java/org/mapstruct/ap/internal/prism/InjectionStrategyPrism.java
kevcodez: Didn't put any effort into JavaDoc yet, as the implementation might change drastically (prior to your review ;)). Since the basic implementation seems to be fine, I will add some JavaDoc to the new classes.

processor/src/main/java/org/mapstruct/ap/internal/util/MapperConfiguration.java
kevcodez: @filiphr No problem, I appreciate any feedback :) I will implement the review feedback soon

Regarding the tests: I agree, the easiest way is to copy the existing mappers and once the field injection is removed in a later release, we can just delete those mappers.

Regarding the injection of additional fields: The easiest way would be to not handle this case at all and possibly mix constructor with setter injection. However, this would also impact testability. A user would have to call a constructor and call the additional setters. I can't come up with an easy way of handling the injection of additional fields, though.

Regarding the decorators: If I understand you correctly, this is also implemented, see the example implementations in my first post. Otherwise, could you please give a sample mapper and the supposed implementation that MapStruct will generate?

Maybe another opinion regarding tests and the injection of additional fields from @gunnarmorling  @agudian @sjaakd would be nice :)

kevcodez:
@filiphr No problem, I appreciate any feedback :) I will implement the review feedback soon

Regarding the tests: I agree, the easiest way is to copy the existing mappers and once the field injection is removed in a later release, we can just delete those mappers.

Regarding the injection of additional fields: The easiest way would be to not handle this case at all and possibly mix constructor with setter injection. However, this would also impact testability. A user would have to call a constructor and call the additional setters. I can't come up with an easy way of handling the injection of additional fields, though.

Regarding the decorators: If I understand you correctly, this is also implemented, see the example implementations in my first post. Otherwise, could you please give a sample mapper and the supposed implementation that MapStruct will generate?

Maybe another opinion regarding tests and the injection of additional fields from @gunnarmorling @agudian @sjaakd would be nice :)

filiphr:
@kevcodez I am not sure if the field injection will ever be removed, we will definitely switch to constructor injection eventually, but removing the field injection can theoretically lead to unsolvable problems with cyclic dependencies for someone (and we don't want to do that 😄 ).

For the tests you saw that there are tests in the main processor module and the `mapstruct-integrationtest` module, right? 😄 

Regarding the other field injection, we can create an issue as an enhancement for that. In order not to make this review bigger. And if you want to work on that then feel free to do it additionally 😄 . Regarding testability, you are right, one will need to invoke the correct constructor and use the setters, but in my opinion that is acceptable.

Regarding the decorators, I didn't mean about what you are referring to. Maybe we can have a new issue for this and work on that separately. Have a look at `PersonMapperDecorator` and the test that is generated for the test methods that uses it. In the decorator we have:

```java 
    public PersonMapperImpl() {
        this( new PersonMapperImpl_() );
    }

    private PersonMapperImpl(PersonMapperImpl_ delegate) {
        super( delegate );
        this.delegate = delegate;
    }
```

Is it worth doing something similar for the others as well. For example for Spring you would have:

```java
    @Autowired
    public PersonMapperImpl(@Qualifier("delegate") PersonMapper delegate) {
        super( delegate ); //in the current solution it is without this.
        this.delegate = delegate;
    }
```

That way there won't be a big difference between `default` and the others.

processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
kevcodez: Done.

kevcodez:
@filiphr I implemented review feedback and did an integration test aswell for field and constructor injection. Should I write another Test using MapperConfig and InjectionStrategy? I will try to add another test in the processor module aswell.

Where do you want me to documentate the possibility of defining an injectionStrategy within the mapstruct-reference-guide? Under **5.2. Mapping object references**?

integrationtest/src/test/java/org/mapstruct/itest/tests/InjectionStrategyTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.mapstruct.itest.testutil.runner.ProcessorSuite;
+import org.mapstruct.itest.testutil.runner.ProcessorSuite.ProcessorType;
+import org.mapstruct.itest.testutil.runner.ProcessorSuiteRunner;
+
+/**
+ * @author Kevin Grüneberg
+ *
+ */
+@RunWith( ProcessorSuiteRunner.class )
+@ProcessorSuite( baseDir = "injectionStrategyTest", processorTypes = ProcessorType.ORACLE_JAVA_8 )
```
kevcodez: No, this was just for testing, I changed it back.

integrationtest/src/test/resources/injectionStrategyTest/src/main/java/org/mapstruct/itest/injectionstrategy/constructor/CustomerConstructorMapper.java
```diff
@@ -0,0 +1,37 @@
...
+import org.mapstruct.itest.injectionstrategy.CustomerEntity;
+
+/**
+ *
+ * @author Kevin Grüneberg
+ */
+@Mapper(componentModel = "spring", uses = GenderConstructorMapper.class, injectionStrategy = InjectionStrategy.CONSTRUCTOR)
+public interface CustomerConstructorMapper {
+
+    @Mapping(source = "gender", target = "gender")
```
kevcodez: Oh, this was also just temporary :D Removed it!

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -160,6 +160,8 @@
      */
     NullValueCheckStrategy nullValueCheckStrategy() default ON_IMPLICIT_CONVERSION;
 
+    InjectionStrategy injectionStrategy() default InjectionStrategy.FIELD;
```
kevcodez: Sure, adding it

kevcodez:
@filiphr Can you take another look? (Waiting for Travis aswell)

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -160,6 +160,16 @@
      */
     NullValueCheckStrategy nullValueCheckStrategy() default ON_IMPLICIT_CONVERSION;
 
+    /**
+     * Determines whether to use field or constructor injection. This is only used on annotated based component models
+     * such as CDI, Spring and JSR330.
+     *
+     * Can be overridden by the one on {@link MapperConfig}.
```
filiphr: It is the other way around. This overrides the value in the `@MapperConfig`

kevcodez: Fxd.

Would be great if you can help a bit with the test.

Sure, so this PR will be on hold for a while, right?

filiphr: I'll push a test case to this no worries.

filiphr:
>Sure, so this PR will be on hold for a while, right?
  
I'll see what I can do. Maybe I'll make a branch for the `1.2.0` things and push this to master. Then we can do a release from the `1.2.0` branch and have this in master.

kevcodez:
@filiphr Do as you please. I am not in a hurry :)

filiphr:
@kevcodez I've pushed the changes for the tests. I moved the integration tests you had into the processor tests. I hope it is fine for you. We already have basic integration tests with spring / cdi, so I think that this is sufficient.

Don't worry when I do the merge, I will squash the commits and you will still be the author 😄 .

Btw. if the build fails, would you be able to rebase everything on top of master? There were some changes in Travis and I think that we have some problems with the earlier setup (master is updated and OK). Thanks

kevcodez:
@filiphr Glad to hear that. Sure, if there is anything I can do to help you there, just mention me

kevcodez:
I messed up my branches, got the patches though. Fixing it...

filiphr:
@kevcodez no worries. I have the commit on my computer. I'll commit them again tomorrow. I'll also merge this pretty soon, we are almost there with the 1.2.0.Final if we don't release it this week I'll do something else so we don't have open PRs 😄

kevcodez:
@filiphr Thanks a lot!

core-common/src/main/java/org/mapstruct/InjectionStrategy.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Strategy for handling injection. This is only used on annotated based component models such as CDI, Spring and
+ * JSR330.
+ *
+ * @author Kevin Grüneberg
+ */
+public enum InjectionStrategy {
+
+    FIELD,
```
gunnarmorling: Some comment would be nice.

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -161,6 +161,16 @@
     NullValueCheckStrategy nullValueCheckStrategy() default ON_IMPLICIT_CONVERSION;
 
     /**
+     * Determines whether to use field or constructor injection. This is only used on annotated based component models
+     * such as CDI, Spring and JSR330.
```
gunnarmorling: s/JSR330/JSR 330/

core-common/src/main/java/org/mapstruct/MapperConfig.java
```diff
@@ -148,6 +148,18 @@ MappingInheritanceStrategy mappingInheritanceStrategy()
     NullValueCheckStrategy nullValueCheckStrategy() default ON_IMPLICIT_CONVERSION;
 
     /**
+     * Determines whether to use field or constructor injection. This is only used on annotated based component models
+     * such as CDI, Spring and JSR330.
```
gunnarmorling: Same here.

filiphr:
@kevcodez we have finally released `1.2.0.Final`. It certainly took longer that I expected. Thanks for your patience 😄 

There is some conflict in this PR. Can you please rebase this on top of the current master and then I would be able to merge it? If you can apply the comments from @gunnarmorling it would be great, otherwise, I'll do them.

kevcodez:
@filiphr @gunnarmorling No problem ;)

I merged the current master branch. I also added JavaDoc on the fields and fixed the typo regarding "JSR 330". Do you want me to squash everything together aswell?

filiphr:
Thanks a lot @kevcodez. I'll perform the squashing no worries :). (will merge it manually and let you know)

kevcodez:
@filiphr Can youtell me the exact actions you take to squash the commits in this case (commits inbetween)? So you do not have to do it next time ;)

filiphr:
Of course. You can try it out if you want to as well. What I basically do is [rebasing](https://git-scm.com/book/en/v2/Git-Branching-Rebasing).

So in this case I would rebase [interactive](https://git-scm.com/docs/git-rebase#git-rebase--i).
* `git rebase -i upstream/master`
* This gives you a list of commits that you can modify and rearrange:
```
pick 771237e4 Constructor Injection for Annotation Based Component Model
pick 590a6801 Allow configuring injection strategy (Constructor/Field) for annotation based model processors. The default strategy is field injection.
pick 6f5d5e9b Unit-Test / JavaDoc
pick 30e763d8 Fix JavaDoc
pick 82eddf28 Move from integration tests into processor tests
pick faf2bd79 #571 Add test for default injection strategy
pick 47cbeb8b JavaDoc
```

I would then pick what I want to do:

```
r 771237e4 Constructor Injection for Annotation Based Component Model
f 590a6801 Allow configuring injection strategy (Constructor/Field) for annotation based model processors. The default strategy is field injection.
f 6f5d5e9b Unit-Test / JavaDoc
f 30e763d8 Fix JavaDoc
f 82eddf28 Move from integration tests into processor tests
f 47cbeb8b JavaDoc
pick faf2bd79 #571 Add test for default injection strategy
```

The `r`(reword) would give me the option to edit the commit message (Add the issue ID and amend the message). Keep in mind that `#` is a comment character, so you would either need to change it, or do something else. The `f`(fixup) means that the commit would be merged into the one above it. And `pick` would leave the commit as is.

Once you do that, you should end up with 2 commits, you then need to force push them `git force push`.

Let me know if you want to ry it out

kevcodez:
@filiphr Will try this evening

kevcodez:
@filiphr Unfortunately, I do not have the time to merge it for the next couple of days, could you please do it? 🙏

filiphr:
I am doing it now, just finishing with compilation and will merge it. If you do squash it will ask you for the message that you want to use for the squash and you would be able to combine it.

https://github.com/filiphr/mapstruct/commits/master How does it look?

By the way I see that you are using your master branch for doing changes, which then forces you to do merge commits from the upstream (mapstruct) branch. I would suggest you to delete the branch you have and checkout the upstream from scratch. Then when doing changes I would create branches. And I would use the master just to have it synced with upstream. That way it would be easier for you to manage your fork. This is just my advice for you and how I am working with those stuff :)

filiphr:
Merged manually via 0192bdaf836212108edbb97a5e55ce0ca1b61dd1 and 8143aa81e21ef3385449858959ae73182c29f66f. Thanks a lot @kevcodez and thank you for your patience 😉

PR #1272, opened 2017-08-16
#1269 use update methods for different sources for nested targets

### LINKED ISSUE #1269: [1.2.0.CR1] Error on mapping generation with multiple sources in target & sub target fields
Hello,

I'm facing an issue with the mapping generation of multiple sources mapped in a target and sub target.
You can find in attachment my case my case with which I have the problem (inspired by the _Example 25. Mapper controlling nested beans mappings (FishTankMapperWithDocument)_ )

[mapstruct-nested-bean-mappings-vehicle.zip](https://github.com/mapstruct/mapstruct/files/1220684/mapstruct-nested-bean-mappings-vehicle.zip)

`public interface VehicleMapper {
  @Mappings({
      @Mapping(source = "vehicleTypeInfo" , target = "vehicleInfo"),
      @Mapping(source = "images" , target = "vehicleInfo.images"),
  })
  VehicleDto map(Vehicle in);
}`

The generated mapping:

`    
    @Override
    public VehicleDto map(Vehicle in) {
        if ( in == null ) {
            return null;
        }

        VehicleDto vehicleDto = new VehicleDto();

        vehicleDto.setVehicleInfo( vehicleTypeInfoToVehicleInfoDto( in.getVehicleTypeInfo() ) );
        vehicleDto.setVehicleInfo( vehicleToVehicleInfoDto( in ) );

        return vehicleDto;
    }

     protected VehicleInfoDto vehicleToVehicleInfoDto(Vehicle vehicle) {
        if ( vehicle == null ) {
            return null;
        }

        VehicleInfoDto vehicleInfoDto = new VehicleInfoDto();

        vehicleInfoDto.setImages( vehicleImageListToVehicleImageDtoList( vehicle.getImages() ) );

        return vehicleInfoDto;
    }
`

As you can see, the second call to setVehicleInfo create a new VehicleInfoDto by setting only images and of course the set erase the previous call to setVehicleInfo .

## Solutions:

What I propose but probably tricky is a mean to call a custom mapper:

`
  //  The mapping I would like to call from the method above
  VehicleInfoDto mapVehicleInfo(VehicleTypeInfo vhInfo, List<VehicleImage> images);
`
The generation of the mapping above is exactly what I expect but this method is currently not call in the mapping of VehicleToVehicleDTO ( VehicleDto map(Vehicle in) )

`
public VehicleInfoDto mapVehicleInfo(VehicleTypeInfo vhInfo, List<VehicleImage> images) {
  if(vhInfo == null && images == null) {
    return null;
  } else {
    VehicleInfoDto vehicleInfoDto = new VehicleInfoDto();
    if(vhInfo != null) {
      vehicleInfoDto.setName(vhInfo.getName());
      vehicleInfoDto.setDoors(vhInfo.getDoors());
    }

    if(images != null) {
      vehicleInfoDto.setImages(this.vehicleImageListToVehicleImageDtoList1(images));
    }

    return vehicleInfoDto;
  }
}

protected List<VehicleImageDto> vehicleImageListToVehicleImageDtoList1(List<VehicleImage> list) {
    if ( list == null ) {
        return null;
    }

    List<VehicleImageDto> list1 = new ArrayList<VehicleImageDto>( list.size() );
    for ( VehicleImage vehicleImage : list ) {
        list1.add( vehicleImageToVehicleImageDto( vehicleImage ) );
    }

    return list1;
}
`

Otherwise, as a basic solution, I expect something like below:

`
@Override
public VehicleDto map(Vehicle in) {
    if ( in == null ) {
       return null;
    }

    VehicleDto vehicleDto = new VehicleDto();
    out.setVehicleInfo(vehicleTypeInfoToVehicleInfoDto( in.getVehicleTypeInfo() ));
    if (in.getSizedPictures() != null) {
      if (out.getVehicleInfo() == null) {
        out.setVehicleInfo(new VehicleInfoDto());
      }
      out.getVehicleInfo().setImages(vehicleImageListToVehicleImageDtoList1( in.getImages() ) );
    }
  }
  return out;
}

protected VehicleInfoDto vehicleTypeInfoToVehicleInfoDto(VehicleTypeInfo vehicleTypeInfo) {
    if ( vehicleTypeInfo == null ) {
        return null;
    }

    VehicleInfoDto vehicleInfoDto = new VehicleInfoDto();

    vehicleInfoDto.setName( vehicleTypeInfo.getName() );
    vehicleInfoDto.setDoors( vehicleTypeInfo.getDoors() );

    return vehicleInfoDto;
}

protected List<VehicleImageDto> vehicleImageListToVehicleImageDtoList1(List<VehicleImage> list) {
    if ( list == null ) {
        return null;
    }

    List<VehicleImageDto> list1 = new ArrayList<VehicleImageDto>( list.size() );
    for ( VehicleImage vehicleImage : list ) {
        list1.add( vehicleImageToVehicleImageDto( vehicleImage ) );
    }

    return list1;
}
`

Thank to tell me if both cases are possible or at least the second.

Best regards, 
Xavier

filiphr:
Fixes #1269

We are basically resorting to creating update methods in case a case like in the issue occurs. The generated mapper will look like the one in [this](https://gist.github.com/filiphr/f795d28730aaa7d1b83b3b17db042546) gist

I have added a note in the documentation that it can happen that an ignoring unmapped target policy is used.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
@@ -180,9 +180,15 @@ public class NestedTargetPropertyMappingHolder {
                         .groupedBySourceReferences
                         .entrySet() ) {
                         PropertyEntry sourceEntry = entryBySP.getKey();
+                        boolean forceUpdateMethodOrNonNestedReferencesPresent =
+                            forceUpdateMethod || !groupedSourceReferences.nonNested.isEmpty();
+                        // if an update method is forced or there are non nested source references then
+                        // we need to create mapping options for forged methods (i.e. Ignore unmapped target properties
+                        // because it the nonNested properties might contain the target properties)
                         MappingOptions sourceMappingOptions = MappingOptions.forMappingsOnly(
                             groupByTargetName( entryBySP.getValue() ),
-                            forceUpdateMethod
+                            forceUpdateMethod,
+                            forceUpdateMethodOrNonNestedReferencesPresent
                         );
                         SourceReference sourceRef = new SourceReference.BuilderFromProperty()
                             .sourceParameter( sourceParameter )
@@ -192,11 +198,13 @@ public class NestedTargetPropertyMappingHolder {
                             .name( targetProperty.getName() )
                             .build();
 
+                        // if an update method is forced or there are non nested source references then we must create
+                        // an update mapping method because the nonNested references can reference the same object
                         PropertyMapping propertyMapping = createPropertyMappingForNestedTarget(
                             sourceMappingOptions,
                             targetProperty,
                             sourceRef,
-                            forceUpdateMethod
+                            forceUpdateMethodOrNonNestedReferencesPresent
                         );
 
                         if ( propertyMapping != null ) {
@@ -219,11 +227,16 @@ public class NestedTargetPropertyMappingHolder {
                             .name( targetProperty.getName() )
                             .build();
 
+                        boolean forceUpdateMethodForNonNested =
+                            forceUpdateMethod || !groupedSourceReferences.groupedBySourceReferences.isEmpty();
+                        // If an update method is forced or there are groupedBySourceReferences then we must create
+                        // an update method. The reason is that they might be for the same reference and we should
+                        // use update for it
                         PropertyMapping propertyMapping = createPropertyMappingForNestedTarget(
                             nonNestedOptions,
                             targetProperty,
                             reference,
-                            forceUpdateMethod
+                            forceUpdateMethodForNonNested
                         );
 
                         if ( propertyMapping != null ) {
--- a/documentation/src/main/asciidoc/controlling-nested-bean-mappings.asciidoc
+++ b/documentation/src/main/asciidoc/controlling-nested-bean-mappings.asciidoc
@@ -85,3 +85,9 @@ Instead of configuring everything via the parent method we encourage users to ex
 This puts the configuration of the nested mapping into one place (method) where it can be reused from several methods in the upper level,
 instead of re-configuring the same things on all of those upper methods.
 ====
+
+[NOTE]
+====
+In some cases the `ReportingPolicy` that is going to be used for the generated nested method would be `IGNORE`.
+This means that it is possible for MapStruct not to report unmapped target properties in nested mappings.
+====
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_1269/Issue1269Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_1269/Issue1269Test.java
@@ -1,0 +1,78 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.bugs._1269;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mapstruct.ap.test.bugs._1269.dto.VehicleDto;
+import org.mapstruct.ap.test.bugs._1269.dto.VehicleImageDto;
+import org.mapstruct.ap.test.bugs._1269.dto.VehicleInfoDto;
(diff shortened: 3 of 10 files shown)


filiphr:
@sjaakd I've added you to this review as we have been working together on this part 😄

sjaakd:
> @sjaakd I've added you to this review as we have been working together on this part 😄

I'll do a review asap. 😄

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -180,9 +180,15 @@ public NestedTargetPropertyMappingHolder build() {
                         .groupedBySourceReferences
                         .entrySet() ) {
                         PropertyEntry sourceEntry = entryBySP.getKey();
+                        boolean forceUpdateMethodOrNonNestedReferencesPresent =
+                            forceUpdateMethod || !groupedSourceReferences.nonNested.isEmpty();
+                        // if an update method is forced or there are non nested source references then
```
sjaakd: Q1. The only places were we currently fall back on update method is when we have 2 source arguments in the mapping method right?

Q2. This is a new situation. I think we need to explain what actually triggers this update methods. There have been many discussions in the past why we use them in a chain that only uses create methods.

I'm not sure.. In general. were would this lead is in the future? Should we actually allow this situation?

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -180,9 +180,15 @@ public NestedTargetPropertyMappingHolder build() {
...
+                        // we need to create mapping options for forged methods (i.e. Ignore unmapped target properties
+                        // because it the nonNested properties might contain the target properties)
```
sjaakd: I think we need to put examples in the comments. I'm kind of lost what happens here.

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -180,9 +180,15 @@ public NestedTargetPropertyMappingHolder build() {
...
                         MappingOptions sourceMappingOptions = MappingOptions.forMappingsOnly(
                             groupByTargetName( entryBySP.getValue() ),
-                            forceUpdateMethod
+                            forceUpdateMethod,
```
sjaakd: the parameter name is `restrictToDefinedMappings` however we put in`forceUpdateMethod` how do they relate?.  Same goes for the next arg.

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: Q1: Yes until now the only place where we fall back to update method was having 2 source methods.

Q2: Exactly this is another corner case. All these cases are quite complex and are forcing us to do some complex logic in there.

I agree with you, this can be more problematic in the future. 

What we could do is not to allow mappings that are not consistent mappings to different sources should not be allowed maybe. This means that we would have to do checks at the beginning and if we see something like this we error out. So if we have not symmetric mappings to sources (mappings to different sources within the levels)

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: Again, I completely agree with you. It is complex (maybe even too much). I can put some examples here, or even mention the tests that are testing this

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: It was like that before. Maybe we need to rename it to `restrictToDefinedMappings` instead.

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
sjaakd: > What we could do is not to allow mappings that are not consistent mappings to different sources should not be allowed maybe.

I'm just thinking: what are frameworks like dozer doing in these cases? Are we missing something?

Anyway.. If we can support a case in a simple manner (like above) we probably should.

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
sjaakd: ok..

sjaakd:
@filiphr : apart from the remarks I head about naming and putting in some references to the test cases, I don't have anything to add. Please feel free to go ahead.

I do think we need to refactor this stuff one day. There's already a PR for removing the mappings originating from the old enum mapping solution. Whence that is done, we could see how understandable the code is and add references to test cases where applicable.

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: I am not entirely sure what the other frameworks do in such a case. Maybe we need to test it 😄

filiphr:
I'll apply those remarks and push the changes.

I agree we need to refactor this and simplify it if possible. We have enough test cases I think to be able to do it correctly 😄

filiphr:
@sjaakd I update the comments can you have a look again if it is better?

sjaakd:
@filiphr go ahead..

Issue #1255, opened 2017-07-20
Wrong prototype types identification for mapper configuration


leomillon:
Hello,

I have a problem with the prototype definition in MapperConfig since the 1.2.0.Beta2 version.

When I try to reproduce the exemple 65 of the documentation ( [Example 65. Mapper configuration class with prototype methods](http://mapstruct.org/documentation/dev/reference/html/#shared-configurations)), it works well in 1.2.0.Beta1 but not in the next version (1.2.0.CR1 included).

Here is a sample code to show the problem (Java8, JUnit/AssertJ).

Model :
```java
public abstract class AbstractA {
  private String field1;

  public String getField1() {
    return field1;
  }

  public void setField1(String field1) {
    this.field1 = field1;
  }
}

public class SomeA extends AbstractA {
  private String field2;

  public String getField2() {
    return field2;
  }

  public void setField2(String field2) {
    this.field2 = field2;
  }
}

public class SomeB {
  private String field1;
  private String field2;

  public String getField1() {
    return field1;
  }

  public void setField1(String field1) {
    this.field1 = field1;
  }

  public String getField2() {
    return field2;
  }

  public void setField2(String field2) {
    this.field2 = field2;
  }
}
```

Mappers and Config :
```java
@MapperConfig(
    mappingInheritanceStrategy = MappingInheritanceStrategy.AUTO_INHERIT_FROM_CONFIG
)
public interface SomeMapperConfig {

  @Mappings({
      @Mapping(target = "field1", ignore = true)
  })
  AbstractA toAbstractA(SomeB source);
}

@Mapper(config = SomeMapperConfig.class)
public interface SomeMapper {

  static SomeMapper instance() {
    return Mappers.getMapper(SomeMapper.class);
  }

  SomeA toSomeA(SomeB source);

  SomeB toSomeB(SomeA source);
}
```

Tests :
```java
public class SomeMapperTest {
  @Test
  public void should_map_someB_to_someA_without_field1() throws Exception {
    SomeB someB = new SomeB();
    someB.setField1("value1");
    someB.setField2("value2");

    SomeA someA = SomeMapper.instance().toSomeA(someB);

    assertThat(someA.getField1())
        .isNotEqualTo(someB.getField1())
        .isNull();
    assertThat(someA.getField2()).isEqualTo(someB.getField2());
  }

  @Test
  public void should_map_someA_to_someB() throws Exception {
    SomeA someA = new SomeA();
    someA.setField1("value1");
[...]

Issue #72, opened 2013-11-24
Allow to exclude a property from mapping


gunnarmorling:
Something like the following would be useful in conjunction with customized mappings via decorators in order to avoid warnings due to unmapped properties:

```
@Mapping(target="derivedProperty", ignore=true)
OrderDto orderToOrderDto(Order order);
```

PR #1276, opened 2017-08-17
#1255 ReverseMapping should not apply automatically.

sjaakd:
The problem reproduces by means of the provided test case. However, it seems to be deliberate functionality, at one time introduced after issue #72.

It is possible to include a `source` name besides `target` in the `@Mapping`. Also, ignore should be applied when there is a target property with the same name in the source. The testcases in `org.mapstruct.ap.test.ignore.IgnorePropertyTest` start to fail once the fix in the second [commit](https://github.com/sjaakd/mapstruct/commit/9db22395ae3b354efa27ff5ac8bdd4065c0d50f1) is in place. 

@gunnarmorling , @agudian , @filiphr : do we still want this functionality? Seems to me it is more straightforward and better explainable to 'ignore' the `ignore=true` in the reverse mapping just as we do for expressions and constants. But that would be breaking existing functionality.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
@@ -398,15 +398,9 @@ public class Mapping {
     public Mapping reverse(SourceMethod method, FormattingMessager messager, TypeFactory typeFactory) {
 
         // mapping can only be reversed if the source was not a constant nor an expression nor a nested property
-        if ( constant != null || javaExpression != null ) {
+        // and the mapping is not a 'target-source-ignore' mapping
+        if ( constant != null || javaExpression != null || ( isIgnored && sourceName == null ) ) {
             return null;
-        }
-
-        // should only ignore a property when 1) there is a sourceName defined or 2) there's a name match
-        if ( isIgnored ) {
-            if ( sourceName == null && !hasPropertyInReverseMethod( targetName, method ) ) {
-                return null;
-            }
         }
 
         Mapping reverse = new Mapping(
--- a/processor/src/main/java/org/mapstruct/ap/internal/prism/MappingInheritanceStrategyPrism.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/prism/MappingInheritanceStrategyPrism.java
@@ -25,6 +25,32 @@ package org.mapstruct.ap.internal.prism;
  * @author Andreas Gudian
  */
 public enum MappingInheritanceStrategyPrism {
-    EXPLICIT,
-    AUTO_INHERIT_FROM_CONFIG;
+
+    EXPLICIT( false, false, false ),
+    AUTO_INHERIT_FROM_CONFIG( true, true, false ),
+    AUTO_INHERIT_REVERSE_FROM_CONFIG( true, false, true ),
+    AUTO_INHERIT_ALL_FROM_CONFIG( true, true, true );
+
+    private final boolean autoInherit;
+    private final boolean applyForward;
+    private final boolean applyReverse;
+
+    MappingInheritanceStrategyPrism(boolean isAutoInherit, boolean applyForward, boolean applyReverse) {
+        this.autoInherit = isAutoInherit;
+        this.applyForward = applyForward;
+        this.applyReverse = applyReverse;
+    }
+
+    public boolean isAutoInherit() {
+        return autoInherit;
+    }
+
+    public boolean isApplyForward() {
+        return applyForward;
+    }
+
+    public boolean isApplyReverse() {
+        return applyReverse;
+    }
+
 }
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
@@ -59,6 +59,7 @@ import org.mapstruct.ap.internal.prism.DecoratedWithPrism;
 import org.mapstruct.ap.internal.prism.InheritConfigurationPrism;
 import org.mapstruct.ap.internal.prism.InheritInverseConfigurationPrism;
 import org.mapstruct.ap.internal.prism.MapperPrism;
+import org.mapstruct.ap.internal.prism.MappingInheritanceStrategyPrism;
 import org.mapstruct.ap.internal.prism.NullValueMappingStrategyPrism;
 import org.mapstruct.ap.internal.processor.creation.MappingResolverImpl;
 import org.mapstruct.ap.internal.util.FormattingMessager;
@@ -67,7 +68,6 @@ import org.mapstruct.ap.internal.util.Message;
 import org.mapstruct.ap.internal.util.Strings;
 import org.mapstruct.ap.internal.version.VersionInformation;
 
-import static org.mapstruct.ap.internal.prism.MappingInheritanceStrategyPrism.AUTO_INHERIT_FROM_CONFIG;
 import static org.mapstruct.ap.internal.util.Collections.first;
 import static org.mapstruct.ap.internal.util.Collections.join;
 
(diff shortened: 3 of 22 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
```diff
@@ -398,16 +398,16 @@ private boolean hasPropertyInReverseMethod(String name, SourceMethod method) {
...
         }
 
         // should only ignore a property when 1) there is a sourceName defined or 2) there's a name match
-        if ( isIgnored ) {
-            if ( sourceName == null && !hasPropertyInReverseMethod( targetName, method ) ) {
-                return null;
-            }
-        }
+//        if ( isIgnored ) {
+//            if ( sourceName == null && !hasPropertyInReverseMethod( targetName, method ) ) {
```
sjaakd: so this code seems to be specifically there to test the conditions I mentioned.

agudian:
I’m for doing it as we discussed, not reverse the ignore... and document that change in the release notes 🙂

sjaakd:
> I’m for doing it as we discussed, not reverse the ignore... and document that change in the release notes 🙂

I'm myself also in favor. But perhaps we should explicitly exclude the combination of `source-target-ignore` (just as I think we do for `source-target-constant` and `source-target-expression`). And we should adapt the `org.mapstruct.ap.test.ignore.IgnorePropertyTest`

filiphr:
I have another idea. Why not reverse the ignore if and only if there is a source parameter.

So when we have:

```java
@Mapping(target = "target", source = "source", ignore = true)
```

Then we are going to produce a reverse mapping. However, if we have:

```java
@Mapping(target = "target", ignore = true)
```

Basically, the second point on line 405 will not longer hold. 

Then we won't reverse it. I think that that way one can configure what they want to reverse, i.e. if the reverse should not be inversely inherited then they should not write a source parameter, but if they want to inherit the ignore to the reverse method, than they also need to provide a source parameter.

Theoretically, your reproducer should be applied, because you have both `target` and `source`. However, in the issue there is no source, so it won't be used. and it should work as it should (I think 😄 )

sjaakd:
Its both: name based mapping and/or presence source parameter in mapping

filiphr:
>Its both: name based mapping and/or presence source parameter in mapping

Sorry I didn't understood you well? What is both?

Btw. I also forgot, we would need to update the Javadoc on the `MappingInheritanceStrategy#AUTO_INHERIT_FROM_CONFIG` as well. See the original message in the issue for that. The text there is not completely correct.

sjaakd:
I meant, if @Mapping only has a target and ignore =true but source and target property name are the same, the reverse mapping is ignored as well. This is what's actually happening in the reported problem.

Verstuurd vanaf mijn iPhone

filiphr:
Now I understand you. Yes that is what I meant as well. When the `source` and `target` name are same and the user wants to ignore in both sides then they have to write `@Mapping(source = "field1", target = "field1", ignore = true)`.

sjaakd:
So: to If I understand you correctly: currently  @Mapping(source = "field1", ignore = true) 
Is sufficient when name are the same and you want to make this explicit like:  @Mapping(source = "field1", target = "field1", ignore = true). 
As well. @agudian and myself want to remove the option completely.

> Op 18 aug. 2017 om 23:56 heeft Filip Hrisafov <notifications@github.com> het volgende geschreven:
> 
> @Mapping(source = "field1", target = "field1", ignore = true)

filiphr:
Yes I think that making it explicit like `@Mapping(source = "field1", target = "field1", ignore = true)` might be more clearer for users, they might expect something like this to work. Whereas with `@Mapping(target = "field1", ignore = true)`, one would expect that it is ignored only in one direction.

Basically what I am saying is to remove the `!hasPropertyInReverseMethod( targetName, method )` check.

However, if you and @agudian think that it is better to completely remove that go for it.

sjaakd:
> Yes I think that making it explicit like @Mapping(source = "field1", target = "field1", ignore = true) might be more clearer for users, they might expect something like this to work. Whereas with @Mapping(target = "field1", ignore = true), one would expect that it is ignored only in one direction.

If you extrapolate on that thought: why not support `@Mapping(source = "field1"", ignore = true)` as well, so that a field is mapped in the forward mapping and suppressed in the reverse mapping. That would also cover fields which are not present in the source mappings like database keys. 

The only problem for this: we attached everything on the target mapping. Therefor defining 2 mappings (forward and reverse) in the shared mapping configuration (in which optionally the reverse mapping inherits everything from the forward mapping and particular fields are overriden) might in the end be more clear. Don't know if we have limitations on that though.

filiphr:
You are right, defining a specific mapping (that would inherit all except `ignore` from the forward mapping) might be clearer.

So if we handle the reversing like you and @agudian are proposing and mention this in the release notes it should be fine.

I think that we would need some tests as well to make sure that this constellation works.

sjaakd:
still to do: finish discussion in #1255 
Opinion @gunnarmorling ?

sjaakd:
Need to check the (failing) integration. Otherwise its ready for review I reckon.

processor/src/test/java/org/mapstruct/ap/test/inheritfromconfig/InheritFromConfigTest.java
```diff
@@ -141,6 +141,35 @@ public void explicitInheritedMappingIsAppliedInReverseDirectlyFromConfig() {
     }
 
     @Test
+    @IssueKey( "1255" )
+    @WithClasses({ CarMapperReverseWithAutoInheritance.class, AutoInheritedReverseConfig.class } )
+    public void autoInheritedMappingIsAppliedInReverseDirectlyFromConfig() {
```
filiphr: I think that we need to add the reverse method to the `CarMapperReverseWithAutoInheritance` and validate that the inheritance is not inherited

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -446,41 +446,48 @@ private void mergeInheritedOptions(SourceMethod method, MapperConfiguration mapp
...
+                        true,
+                        method,
+                        messager,
+                        typeFactory );
+                }
+                else if ( applicableReversePrototypeMethods.size() > 1 ) {
+                    messager.printMessage(
+                        method.getExecutable(),
+                        Message.INHERITINVERSECONFIGURATION_MULTIPLE_PROTOTYPE_METHODS_MATCH,
+                        Strings.join( applicablePrototypeMethods, ", " ) );
```
filiphr: This should be `applicableReversePrototypeMethods` instead of `applicablePrototypeMethods`. In any case, I think it would be better if we can extract it into a method. Something like:

```java
private void applyInheritedOptions(SourceMethod method, MappingOptions mappingOptions,
        List<SourceMethod> applicableMethods, boolean isInverse) {
        if ( applicableMethods.size() == 1 ) {
            mappingOptions.applyInheritedOptions(
                first( applicableMethods ).getMappingOptions(),
                isInverse,
                method,
                messager,
                typeFactory
            );
        }
        else if ( applicableMethods.size() > 1 ) {
            messager.printMessage(
                method.getExecutable(),
                isInverse ? Message.INHERITINVERSECONFIGURATION_MULTIPLE_PROTOTYPE_METHODS_MATCH :
                    Message.INHERITCONFIGURATION_MULTIPLE_PROTOTYPE_METHODS_MATCH,
                Strings.join( applicableMethods, ", " )
            );
        }
    }
```

And we just call it with the appropriate parameters in the ifs. The logic is the same with the only difference in the methods and the `isInverse`

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -47,6 +47,7 @@
     PROPERTYMAPPING_DUPLICATE_TARGETS( "Target property \"%s\" must not be mapped more than once." ),
     PROPERTYMAPPING_EMPTY_TARGET( "Target must not be empty in @Mapping." ),
     PROPERTYMAPPING_SOURCE_AND_CONSTANT_BOTH_DEFINED( "Source and constant are both defined in @Mapping, either define a source or a constant." ),
+    PROPERTYMAPPING_SOURCE_AND_IGNORE_BOTH_DEFINED( "Source and ignore are both defined in @Mapping, make explicit in reverse mapping when the intent is to ignore the reverse mapping." ),
```
filiphr: This is not used. Did you forget to push some things maybe 😄 ?

processor/src/test/java/org/mapstruct/ap/test/ignore/ErroneousCombinationOfSourceAndIgnoreMapper.java
```diff
@@ -0,0 +1,44 @@
...
+import org.mapstruct.Mapping;
+import org.mapstruct.Mappings;
+import org.mapstruct.factory.Mappers;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+@Mapper
+public interface ErroneousCombinationOfSourceAndIgnoreMapper {
```
filiphr: This is also not used. Linked with the message I think

sjaakd:
> Maybe you added them to the copied sources by mistake

uhh. Yep..

sjaakd:
I'll also add some extra test cases.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
filiphr: I don't understand what you mean. The calling side would look like:

```java
else if ( inheritanceStrategy.isAutoInherit() ) {

            if ( inheritanceStrategy.isApplyForward() ) {
                applyInheritedOptions( method, mappingOptions, applicablePrototypeMethods, false );
            }

            if ( inheritanceStrategy.isApplyReverse() ) {
                applyInheritedOptions( method, mappingOptions, applicableReversePrototypeMethods, true );
            }
        }
```
Which is exactly the same. With the only difference those code that is the same is in a method now 😄

processor/src/test/java/org/mapstruct/ap/test/ignore/ErroneousCombinationOfSourceAndIgnoreMapper.java
sjaakd: I'll remove that.. missed that initially..

sjaakd:
test this please

filiphr:
Are the integration tests passing on your side?

Btw, with Travis you don't need to tell it to run the tests. They are always automatically run 😄.

processor/src/test/java/org/mapstruct/ap/test/inheritfromconfig/InheritFromConfigTest.java
sjaakd: Can you check if its complete now.. ?

processor/src/main/java/org/mapstruct/ap/internal/prism/MappingInheritanceStrategyPrism.java
```diff
@@ -25,6 +25,32 @@
  * @author Andreas Gudian
  */
 public enum MappingInheritanceStrategyPrism {
```
filiphr: Do we need a unit test to make sure that the parameters here are correct? I know they would break the other tests. I can add it later on as well

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -446,41 +446,48 @@ private void mergeInheritedOptions(SourceMethod method, MapperConfiguration mapp
...
-                    true,
-                    method,
-                    messager,
-                    typeFactory );
-            }
-            else if ( applicableReversePrototypeMethods.size() > 1 ) {
-                messager.printMessage(
-                    method.getExecutable(),
-                    Message.INHERITINVERSECONFIGURATION_MULTIPLE_PROTOTYPE_METHODS_MATCH,
-                    Strings.join( applicablePrototypeMethods, ", " ) );
```
filiphr: I just realized that this was wrong all along 😄. It seems we don't have a test for this error case

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
filiphr: This one is still here. We don't need it right?

processor/src/test/java/org/mapstruct/ap/test/bugs/_1255/Issue1255Test.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+@IssueKey("1255")
+@RunWith(AnnotationProcessorTestRunner.class)
+@WithClasses({
+    AbstractA.class,
+    SomeA.class,
+    SomeB.class,
+    SomeMapper.class,
+    SomeMapperConfig.class})
+@WithSingleCompiler( org.mapstruct.ap.testutil.runner.Compiler.JDK )
```
filiphr: This should not be used, I supposed you were only using it locally?

sjaakd: ach.. forgot to remove it..

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
sjaakd: true... I'll have a look.. Dunno if I've time left.. Bit tired..

filiphr:
@sjaakd I think that it is fine. Just remove the `@WithSingleCompiler` and the obsolete message and it is good for merging. I'll make an issue for the missing test case, there was none from the get go as well

sjaakd:
> @sjaakd I think that it is fine. Just remove the @WithSingleCompiler and the obsolete message and it is good for merging. I'll make an issue for the missing test case, there was none from the get go as well

I've just added the test case. So I'll squash all the stuff in one commit if CI succeeds.

Issue #610, opened 2015-08-01
Add `Mapping#unmappedSourcePolicy` to control failures to map newly added fields.


johnament:
Adding this so that unmapped fields can optionally fail a build if no mapping present for the attribute.

gunnarmorling:
Yes, this has been asked for before. In case you are interested in helping out with that one, let us know :)

velo:
Anyone doing this? I think I will give it a shot!

gunnarmorling:
@velo I am not aware of anyone working on this, so if like to give it a try, that's awesome! Let me know if you hit any roadblocks and I'll be glad to help.

spoerri:
@velo did you make any progress? We'd like to use this too...

velo:
I end up not needing it... only unmappedTargetPolicy was enough for me

gunnarmorling:
Hey @spoerri maybe you'd like to give this a shot?

spoerri:
No time for near future... and i'd probably go for https://github.com/mapstruct/mapstruct/issues/621 first.

filiphr:
This has been integrated with #1282. Thanks @spoerri for the PR. I have opened #1317 to follow the ignoring.

PR #1282, opened 2017-08-27
#610 unmappedSourcePolicy

spoerri:
Just like the unmappedTargetPolicy feature, defaulting the new configuration to IGNORE for backwards compatibility.

### DIFF
--- a/core-common/src/main/java/org/mapstruct/Mapper.java
+++ b/core-common/src/main/java/org/mapstruct/Mapper.java
@@ -55,6 +55,15 @@ public @interface Mapper {
     Class<?>[] imports() default { };
 
     /**
+     * How unmapped properties of the source type of a mapping should be
+     * reported. The method overrides an unmappedSourcePolicy set in a central
+     * configuration set by {@link #config() }
+     *
+     * @return The reporting policy for unmapped source properties.
+     */
+    ReportingPolicy unmappedSourcePolicy() default ReportingPolicy.IGNORE;
+
+    /**
      * How unmapped properties of the target type of a mapping should be
      * reported. The method overrides an unmappedTargetPolicy set in a central
      * configuration set by {@link #config() }
--- a/core-common/src/main/java/org/mapstruct/MapperConfig.java
+++ b/core-common/src/main/java/org/mapstruct/MapperConfig.java
@@ -58,6 +58,14 @@ public @interface MapperConfig {
     Class<?>[] uses() default { };
 
     /**
+     * How unmapped properties of the source type of a mapping should be
+     * reported.
+     *
+     * @return The reporting policy for unmapped source properties.
+     */
+    ReportingPolicy unmappedSourcePolicy() default ReportingPolicy.IGNORE;
+
+    /**
      * How unmapped properties of the target type of a mapping should be
      * reported.
      *
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -257,6 +257,18 @@ Supported values are:
 
 If a policy is given for a specific mapper via `@Mapper#unmappedTargetPolicy()`, the value from the annotation takes precedence.
 |`WARN`
+
+|`mapstruct.unmappedSourcePolicy`
+|The default reporting policy to be applied in case an attribute of the source object of a mapping method is not populating an attribute of the target object.
+
+Supported values are:
+
+* `ERROR`: any unmapped source property will cause the mapping code generation to fail
+* `WARN`: any unmapped source property will cause a warning at build time
+* `IGNORE`: unmapped source properties are ignored
+
+If a policy is given for a specific mapper via `@Mapper#unmappedSourcePolicy()`, the value from the annotation takes precedence.
+|`IGNORE`
 |===
 
 === Using MapStruct on Java 9
--- a/processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
@@ -91,6 +91,7 @@ import org.mapstruct.ap.internal.util.TypeHierarchyErroneousException;
     MappingProcessor.SUPPRESS_GENERATOR_TIMESTAMP,
     MappingProcessor.SUPPRESS_GENERATOR_VERSION_INFO_COMMENT,
     MappingProcessor.UNMAPPED_TARGET_POLICY,
+    MappingProcessor.UNMAPPED_SOURCE_POLICY,
     MappingProcessor.DEFAULT_COMPONENT_MODEL
 })
 public class MappingProcessor extends AbstractProcessor {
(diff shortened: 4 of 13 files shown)


processor/pom.xml
```diff
@@ -356,6 +356,14 @@
                     </parameter>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>1.8</source>
```
spoerri: master doesn't compile on my machine without this

filiphr:
Thanks for this @spoerri, I'll have a deeper look tonight or during the week. In any case your commit with removing the source / target version is wrong. I don't know why the build fails on your machine, but on mine and on Travis it is OK.

I'll have a look why there is an NPE. As for the `UnmappedSourceTest` it fails with eclipse because there are actually 3 errors / warnings. It doesn't fail with JDK because it works a bit differently there 😄 .

spoerri:
Thanks @filiphr! That commit with removing the version in pom.xml was just to revert my own adding of the version in the first place. I thought it may have been creating other problems, and the presence of core-jdk8 vs core-common made me hesitate to just use java8 across the board.

filiphr:
@spoerri I just saw that commit, I didn't check the changes well sorry.

Yes MapStruct needs Java 6 currently, we are planning to switch to Java 8, but for the next major version probably.

As for `core-jdk8` and `core-common`. They are there because we release 2 jar for the annotations. In one of the jars the `@Mappings` and `@ValueMappings` annotations, have the Java 8 `@Repeatable` annotation 😄 . The `mapstruct-processor` is always Java 6 compatible.

filiphr:
@spoerri I found the issue with the `UnmappedSourceTest`. I should have seen it even sooner. The problem is there the reporting line. This:

```java
@ExpectedCompilationOutcome(
    value = CompilationResult.SUCCEEDED,
    diagnostics = {
        @Diagnostic(type = SourceTargetMapper.class,
            kind = Kind.WARNING,
            line = 33,
            messageRegExp = "Unmapped source property: \"qux\""),
        @Diagnostic(type = SourceTargetMapper.class,
            kind = Kind.WARNING,
            line = 35,
            messageRegExp = "Unmapped source property: \"bar\"")
    }
)
```
is correct.

The order also matters. So first comes the `sourceToTarget` method and it's warnings, then the `targetToSource` and it's warnings.

filiphr:
That other issue is that you are trying to get accessors from a primitive type. You'll need to skip getting the accessors for primitive types. Like we do [here](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java#L503).

Something like this is not needed for the target because the target cannot be a primitive type. There might be some other corner cases as well, like with an array.

spoerri:
Very cool, thank you

filiphr:
@spoerri thanks for doing it this. I think that it is OK. I have only one remark, a test with `@MappingConfig` is missing. Can you add one like that?

Thanks for checking the arrays. I know that we had one issue with double arrays, but I just remembered the we fixed it a bit up the chain so code should not get affected 😄

spoerri:
Np...

Is my third test with `@ProcessorOption(name = "mapstruct.unmappedSourcePolicy", value = "ERROR")` not sufficient for testing @MappingConfig? Or do you mean something like duplicating processor/src/test/java/org/mapstruct/ap/test/mapperconfig and tweaking for unmappedSourcePolicy?

filiphr:
Your 3rd test test the option that is passed via parameters during compilation, and it is great.

We need something like: `Issue880Test` where we can be sure that the configuration from the `@MapperConfig` is used and not the one from the `ProcessorOption`

spoerri:
Argh, now it fails in the integration tests. Any idea?

spoerri:
I worked around it. Please merge!

filiphr:
Thanks a lot for this @spoerri. I'll have a better look during the weekend, but overall I think it is good :). 

The only thing is that we might need to extend the message to make it similar as the target properties where we write the method that was being mapped.

spoerri:
With this feature, it'd also be nice to be able to ignore a source property which doesn't map to any target property.

Which approach would we prefer?

1. make the target() parameter of Mapping optional, so that we can use @Mapping(source="foo", ignore=true)
2. ignore the value of the target parameter when ignore=true and unmappedSourcePolicy is not IGNORE
3. support a sentinel value for the target parameter for this use case

filiphr:
>With this feature, it'd also be nice to be able to ignore a source property which doesn't map to any target property.

You are right. We recently had this discussion for another bug 😄 .

I think mix of 1 and 2 would be the best. `target` can be optional only when `ignore=true` and `unmappedSourcePolicy` is not `IGNORE`.

What do you mean support sentinel value for target?

gunnarmorling:
Where are we with that one; is it something for 1.2 still or for the time after that?

filiphr:
@gunnarmorling I think that this one is OK, I need to have a better look at it, but it is definitely for after 1.2. There are some questions which are open, but nothing major

spoerri:
I tried to add the support for ignoring unmapped source fields, but I got a little stuck :(

Is it wise to continue fixing issues related to the missing the target field, or should I try a completely different approach?

filiphr:
@spoerri thanks a lot for your patience. We have finally released `1.2.0.Final` and we can start merging :).

Let's move handling of ignoring of a source property into a separate PR / issue. Can you create one with your https://github.com/mapstruct/mapstruct/pull/1282#issuecomment-328345251 and reverse the last commit? Can you also maybe rebase this on top of master?

@gunnarmorling is the `unmappedTargetPolicy` compiler options the one you want to remove from 2.0? If yes, I would say that we shouldn't expose `unmappedSourcePolicy` in this PR

spoerri:
> reverse the last commit? Can you also maybe rebase

You got it. Created separate PR for ignore handling: #1316

filiphr:
Thanks a lot for the PR @spoerri. Thank you for the patience 😄

gunnarmorling:
Hi @filiphr,

> @gunnarmorling is the unmappedTargetPolicy compiler options the one you want to remove from 2.0? If yes, I would say that we shouldn't expose unmappedSourcePolicy in this PR

Yes, indeed. I don't think these options should be processor options but only be configurable via `@Mapper` and `@MapperConfig`. The only processor options remaining should be `suppressGeneratorTimestamp` and `suppressGeneratorVersionInfoComment` and potentially other things that deal with technicalities of the code generation process. But all actual mapping options should be configurable only via the annotations.

filiphr:
@gunnarmorling will remove them now directly on master. I have also added the 2 options to the #1307 issue as well. Maybe we need to document them and output warnings in the 1.3 release

gunnarmorling:
Yes, that's a very good idea.

filiphr:
@gunnarmorling functionality removed in https://github.com/mapstruct/mapstruct/commit/a3e7b0a20a9b37a2120088287cca76866d12e070

PR #1287, opened 2017-09-03
Handle no suitable empty constructor during BeanMappingMethod creation

### LINKED ISSUE #1283: Mapstruct: @InheritInverseConfiguration --> There is no suitable result type constructor for reversing this method
I try to use mapstruct feature @InheritInverseConfiguration. But it gives me the following error: There is no suitable result type constructor for reversing this method. But the generated Impl look pretty good.
```
@Mapper(uses = {
  BOIdMapper.class,
  DateMapper.class,
  CarVOFactory.class,
})
public interface CarMapper {
   CarMapper INSTANCE = Mappers.getMapper(CarMapper .class);

   @Mappings({
     @Mapping(source = "VOKey", target = "id"),
     @Mapping(source = "gueltAb", target = "gueltigAb")
   })
   Car map(CarVO carVO);

   @InheritInverseConfiguration
   CarVO map(Car car);

   public class CarVOFactory {
      public CarVO createCarVO() {
         return VOFactory.createVO(CarVO.class);
      }
   }
}
```

This is the POJO Car
```
public class Car {

    private String id;
    private LocalDate gueltigAb;
    ...
}
```
This is the other Car-Class:
```
public interface CarVO extends ... {
    void setVOKey (VOKey<CAR> voKey);
}
```
And this is the generated Impl:
```
@Generated(
    date = "2017-08-29T08:55:10+0200",
    comments = "version: 1.1.0.Final, compiler: Eclipse JDT (IDE) 1.2.100.v20160418-1457, environment: Java 1.8.0_102 (Oracle Corporation)"
)

    private final BOIdMapper bOIdMapper = new BOIdMapper();
    private final DateMapper dateMapper = new DateMapper();
    private final CarVOFactory CarVOFactory = new CarVOFactory();

    @Override
    public Car map(CarVO CarVO) {
        if ( CarVO == null ) {
            return null;
        }
        Car Car = new Car();
        Car.setGueltigAb( dateMapper.asLocalDate( CarVO.getGueltAb() ) );
        Car.setId( bOIdMapper.asId( CarVO.getVOKey() ) );
        return Car;
   }
    @Override
    public CarVO map(Car Car) {
        if ( Car == null ) {
            return null;
        }
        CarVO CarVO = CarVOFactory.createCarVO();
        return CarVO;
   }

```

[Linked](https://stackoverflow.com/questions/45912643/mapstruct-inheritinverseconfiguration-there-is-no-suitable-result-type-con)

filiphr:
Fixes #1283.

This PR fixes #1283 by moving the no suitable empty constructor check during the `BeanMappingMethod` creation (only if there is no factory method). With this we additionally get the same check for non-reverse methods and `BeanMapping#resultType`

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -863,8 +863,7 @@ public class Type extends ModelElement implements Comparable<Type> {
             hasEmptyAccessibleContructor = false;
             List<ExecutableElement> constructors = ElementFilter.constructorsIn( typeElement.getEnclosedElements() );
             for ( ExecutableElement constructor : constructors ) {
-                if ( (constructor.getModifiers().contains( Modifier.PUBLIC )
-                    || constructor.getModifiers().contains( Modifier.PROTECTED ) )
+                if ( !constructor.getModifiers().contains( Modifier.PRIVATE )
                     && constructor.getParameters().isEmpty() ) {
                     hasEmptyAccessibleContructor = true;
                     break;
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
@@ -89,7 +89,7 @@ public enum Message {
     GENERAL_VALID_DATE( "Given date format \"%s\" is valid.", Diagnostic.Kind.NOTE ),
     GENERAL_INVALID_DATE( "Given date format \"%s\" is invalid. Message: \"%s\"." ),
     GENERAL_NOT_ALL_FORGED_CREATED( "Internal Error in creation of Forged Methods, it was expected all Forged Methods to finished with creation, but %s did not" ),
-    GENERAL_NO_SUITABLE_CONSTRUCTOR( "%s does not have a suitable empty constructor." ),
+    GENERAL_NO_SUITABLE_CONSTRUCTOR( "%s does not have an accessible empty constructor." ),
 
     RETRIEVAL_NO_INPUT_ARGS( "Can't generate mapping method with no input arguments." ),
     RETRIEVAL_DUPLICATE_MAPPING_TARGETS( "Can't generate mapping method with more than one @MappingTarget parameter." ),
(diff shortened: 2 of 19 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -863,8 +863,7 @@ public boolean hasEmptyAccessibleContructor() {
             hasEmptyAccessibleContructor = false;
             List<ExecutableElement> constructors = ElementFilter.constructorsIn( typeElement.getEnclosedElements() );
             for ( ExecutableElement constructor : constructors ) {
-                if ( (constructor.getModifiers().contains( Modifier.PUBLIC )
-                    || constructor.getModifiers().contains( Modifier.PROTECTED ) )
+                if ( !constructor.getModifiers().contains( Modifier.PRIVATE )
```
agudian: How do we know here that the constructor is actually accessible from the package of the mapper?

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -89,6 +89,7 @@
     GENERAL_VALID_DATE( "Given date format \"%s\" is valid.", Diagnostic.Kind.NOTE ),
     GENERAL_INVALID_DATE( "Given date format \"%s\" is invalid. Message: \"%s\"." ),
     GENERAL_NOT_ALL_FORGED_CREATED( "Internal Error in creation of Forged Methods, it was expected all Forged Methods to finished with creation, but %s did not" ),
+    GENERAL_NO_SUITABLE_CONSTRUCTOR( "%s does not have a suitable empty constructor." ),
```
agudian: I think we can omit the “suitable”, as you already mention “empty”. Maybe add “visible” or “accessible”?

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: That is the case, we don't really know. And I think it is difficult to check this because the test is later on. Actually one of our test have package protected classes and package protected static classes. The modifiers are then empty (and that is why it started failing).

I am not sure if it is worth it to add more complex checks? It will create a compile problem in the generated file in any case.

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
filiphr: I agree "accessible" sounds better than "suitable". I'll change it. Do you think that we need to give a suggestion what they should do? To me it is pretty clear, but I don't know 😄

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
agudian: Sure, that’s okay as well.

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
agudian: Yes, I think it’s pretty obvious... 🙂

agudian:
Looks good! Feel free to squash and merge!

filiphr:
Thanks a lot for the review @agudian

PR #1288, opened 2017-09-03
#1281 Do not depend on deprecated sonatype parent

### LINKED ISSUE #1281: Don't depend on Sonatype OSS Parent POM any longer
As per [this guide](http://central.sonatype.org/pages/apache-maven.html#deprecated-oss-parent) this parent POM is deprecated and not maintained anymore. Instead, we should pull all the required configuration into our own parent POM. The required changes can be taken from ModiTect, where I just [applied them](https://github.com/moditect/moditect/compare/69e20d236231a3156233418069aa823183e4f003...a02ec309f7d9b9192bc3fa2bd4307c1fe55e453a).

filiphr:
Fixes #1281

### DIFF
--- a/build-config/pom.xml
+++ b/build-config/pom.xml
@@ -33,15 +33,7 @@
     <packaging>jar</packaging>
     <name>MapStruct Build Configuration</name>
 
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-deploy-plugin</artifactId>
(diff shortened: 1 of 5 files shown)


filiphr:
Thanks @agudian, I'll try to run a deploy for the snapshots and stage a release just to make sure that it works correctly

filiphr:
Thanks a lot for the review @agudian. I have also tested this with the snapshot version an deployed to nexus staging. It worked great 😄

Issue #1273, opened 2017-08-16
NullValueMappingStrategy with Lists


kevcodez:
Given the following example

```
public class Dto {

    List<Long> longs = new ArrayList<>();

    public List<Long> getLongs() {
        return longs;
    }

    public void setLongs(List<Long> longs) {
        this.longs = longs;
    }
}
```

```
public class Entity {

    List<Long> longs = new ArrayList<>();

    public List<Long> getLongs() {
        return longs;
    }

    public void setLongs(List<Long> longs) {
        this.longs = longs;
    }
}
```

```
@Mapper(nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT)
public interface EntityMapper {

    Dto asTarget(Entity entity);

}
```

The generated mapper looks like this:

```public class EntityMapperImpl implements EntityMapper {
    public EntityMapperImpl() {
    }

    public Dto asTarget(Entity entity) {
        Dto dto = new Dto();
        if (entity != null) {
            List<Long> list = entity.getLongs();
            if (list != null) {
                dto.setLongs(new ArrayList(list));
            } else {
                dto.setLongs((List)null);
            }
        }

        return dto;
    }
}
```

Shouldn't the list set to an empty list rather than null? Or do I have a misunderstanding of the null value mapping strategy?

filiphr:
I think that you might be right...

Although in the JavaDoc it is written:

>If `null` is passed to a mapping method, a default value will be returned. The value depends on the kind of the annotated method:

Where **passed** is the key word. Maybe it should be considered for the other locations as well.

kevcodez:
> However, by specifying nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT on @BeanMapping, @IterableMapping, @MapMapping, or globally on @Mapper or @MappingConfig, the mapping result can be altered to return empty default values. This means for:
> Iterables / Arrays: an empty iterable will be returned. 
> The strategy works in a hierarchical fashion. Setting nullValueMappingStrategy on mapping method level will override @Mapper#nullValueMappingStrategy, and @Mapper#nullValueMappingStrategy will override @MappingConfig#nullValueMappingStrategy.

Since the strategy can also be used on an IterableMapping or set globally on a Mapper, I expected the list to be empty rather than null. I am not sure if this is working as intended for now, or if this is a bug.

filiphr:
[...]

PR #1293, opened 2017-09-10
Fixes issue #1273 using NVMS#RETURN_DEFAULT when using collections

kevcodez:
Fix for NullValueMappingStrategy#RETURN_DEFAULT when using collections.

This is a proposed fix for https://github.com/mapstruct/mapstruct/issues/1273.

When mapping a collection using NullValueMappingStrategy#RETURN_DEFAULT and the source is null, the target will be an empty collection.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
@@ -30,6 +30,7 @@ import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.model.source.SelectionParameters;
 import org.mapstruct.ap.internal.prism.CollectionMappingStrategyPrism;
 import org.mapstruct.ap.internal.prism.NullValueCheckStrategyPrism;
+import org.mapstruct.ap.internal.prism.NullValueMappingStrategyPrism;
 import org.mapstruct.ap.internal.util.Message;
 import org.mapstruct.ap.internal.util.accessor.Accessor;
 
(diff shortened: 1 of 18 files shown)


kevcodez:
@filiphr Could you please review? This might be relevant for 1.2.0.Final I suppose

filiphr:
@kevcodez I didn't see your message until now :). In any case, I have already reviewed it, have a look at my review.

kevcodez:
@filiphr Would you add the IterableCreation to the appropriate Wrapperclasses (Java)? I started doing that, but this causes package cycles, thus I placed it in the PropertyMapping.

filiphr:
@kevcodez you don't need the `IterableCreation`. That one is mostly used for the size and we don't need that here. I mentioned it only because in there one can see how to do the implementation type.

I would just pass a boolean that will be `mapNullToDefault` or something like that (we have that in the `MappingMethod`s).

If that boolean is `true` I would call the macro `initTargetObject` from the `CommonMacros.ftl`, otherwise just return `null`.

filiphr:
@kevcodez we are going to include this PR in the 1.2.0.Final release 😄. When will you have time to go over my comments?

kevcodez:
@filiphr hopefully tomorrow. If you get no reply til monday, I wont be able to make it

kevcodez:
@filiphr I implemented your feedback, can you have another look?

filiphr:
@kevcodez thanks, I noticed the typo later. I have pushed an update. Also don't worry I am going to squash everything in once commit and you are going to be the author of everything 😄

kevcodez:
@filiphr Sure, it is fine. I used the eclipse formatter XML from MapStruct in IntelliJ. Good to know :)

filiphr:
We actually have an IntelliJ formatter in git [here](https://github.com/mapstruct/mapstruct/blob/master/etc/mapstruct.xml), which is actually up to date. We need to make sure that we document this better 😄

filiphr:
@kevcodez I have merged this into master. Thanks a lot for the PR :). I've also added you to our copyright https://github.com/mapstruct/mapstruct/commit/98bdc3612f5beb6d89596a280e5974cd2dce282b

PR #1296, opened 2017-09-16
Builders Round 2

ericmartineau:
@filiphr 

This represents an altered approach from the first time.  Like you suggested, I used a BuilderProvider SPI to determine whether something is a builder, and also used a BuilderAccessorNamingStrategy to allow for builder setter conventions that are different from regular POJOs.

Currently, this all lives within the Type instance - I removed any trace from the mapping annotations.  Basically, it's transparent to the user and should just work.

Builder factory methods are picked up as ObjectFactory methods automatically, so it ended up requiring pretty minimal changes to the template.

Would love feedback on this newer approach.  I'm still working through some build issues.


filiphr:
@smartytime Thanks a lot for the new PR. You'll have to bear with me, I'll need some time to have a look at it. I hope I'll be able to get to it by the end of next week.

Something from my initial view:
* Why are the `Filters` and `Executables` not static anymore?
* I think that all implementations of the public SPI interfaces should be in the public package something like the `DefaultAccessorNamingStrategy`, otherwise users will need to use internal packages or implement everything by themselves (which is not what we want). I'll have a better look at this as well.

filiphr:
Btw. what kind of build problems are you having? Maybe I can help you out there?

smartytime:
I figured out my build problem.  A dumb issue with maven `${user.home}` directory.

I made a couple decisions that I knew might not go over well.  I'm still trying to understand the scope of things, and I'm sure I'm doing bad things out of ignorance.  They're easy enough to change - so please don't feel bashful in your critique.  

Your comment about Executables sent me back looking through things, and I think I can pretty easily change it back.  The short version is that builder logic makes Executable logic a little more complicated, and Executables started to feel more like `ExecutableFactory` (like `TypeFactory`)

The one thing I liked about making Executables an instance factory is that it allows you to construct it up-front with the appropriate `Types`, `Elements`, and then callers don't have to pass those references around as much.  But, I also like the simplicity.  So, for now I'll change it back.

smartytime:
Also, moving the default SPI implementation had more to do with circular dependencies - the change above should solve that.

smartytime:
I backed out the static Executables change and put a flag on `Accessor`. 

I ran into a bigger problem when writing an integration test for Lombok.

The default BuilderProvider works with lombok by convention, but my integrationtest failed because Lombok hasn't fully processed the  Builder annotation yet.  The `Type` method that checks other annotation processors returns `true` - so I'm guessing that Lombok claims to be finished with an Element once the getters and setters are generated.

I'm working on a Lombok-specific provider now, and had a few questions:

1. Would it be appropriate for BuilderProvider to throw TypeHierarchyErroneousException?  If so, I'll need to move that exception into a shared package.  If not, we'll need another unchecked exception that can push a TypeElement into a later round when it hasn't been fully processed.
2. Should I add modules for the different builder providers into the mapstruct project?  They are all going to be essentially a single class, so an entire project feels a little overkill.  They can be built easily without even adding the dependencies to Mapstruct
3. I don't think we should force the user to pick a single builder provider.  They don't always have control over their entity classes, and might inherit some combination of entities using different builders.  In my use case, we have custom builders and Lombok builders.  What do you think about adding an array of BuilderProvider implementation classes to `@Mapping` (and also available as a global setting) and ditching spi?  E.g.,
```
@Mapping(builderProviders = { LombokBuilders.class, MyCustomBuilders.class })
```
The default value could still be `DefaultBuilderProvider`, and we could also detect builder libraries in the classpath.
4. I'm wrestling with package cycles, particularly the until and spi packages.  I can't reuse anything from util inside my spi implementation.  I could:
- move everything inside util that has dependencies to other packages into a separate package
- split the spi implementation from the interfaces and use reflection to instantiate default providers
- Create a new shared util package that contains classes used by both spi and internal.
- Duplicate some logic in both places

Thanks again for your patience.  This has been a learning experience for me.

filiphr:
@smartytime my main concern with the `Types` and `Executables` was that it brings a lot of changes (and the PR is already big enough 😄 ). It would just make it easier for review if we don't change things which are not needed. And you are right it might be better for those to be singletons, but that is for another issue I would say.

The Lombok stuff are also tricky, I won't lose too much time on those. The thing with Lombok is that they did something else (to the AST of the types) apart from implementing the `AstModifyingAnnotationProcessor, maybe they didn't do it for the builders. I would work with already generated builder or `AutoValue` or some other Builders generators that work like a "normal" APT.

For your questions:

1. Maybe we can use a flag in the `BuilderMapping` that will tell the invoker that a type is incomplete. How would you know in the `BuilderProvider` that you need to throw the exception? Maybe we can move it.
1. Adding the builder providers to the MapStruct codebase would mean that we would need to support them and would need to pay attention to version changes and stuff like that. I am not entirely sure that we need to add something specific to a specific library. We could always add different builder provider implementations to the examples repository and users can just copy them. We could also create a different repository and different module for them.
1. We don't have to have only one `BuilderProvider` we can also have a list. This is what we are doing for the `AstModifyingAnnotationProcessor`, the only thing is how would we know which one to use for which type? We could theoretically add a `supports` and `order` methods them we can order them and use the first provider that has supports true. We could add them as an array, but that would mean that one would alway need to add the providers that you need in all your mappers.
1. What do you need from the `util`? The `spi` package is the public package that we support, so it should not use anything that is internal. I would try to leave the `spi` package as light as possible. Maybe we can move some complex logic to the internal (glancing at the `DefaultBuilderProvider` it looks complex a bit).

>Thanks again for your patience. This has been a learning experience for me.

No problem thank you for tackling this. And also for being patient with my review. 

I still haven't had the chance to look at it better (I would do it this week). In the meantime, feel free to push changes that you have (even if it breaks the build :)).

smartytime:
>  it brings a lot of changes (and the PR is already big enough 😄 ).

Very true.  I was really excited with this SPI approach because it seemed like it would make the change _smaller_... I'm going to work to remove everything but the bare essentials from the PR, and it's obvious that there are certain classes like `Type` where changes should be minimized.

> How would you know in the BuilderProvider that you need to throw the exception? 

In the case of lombok, it's easy to see that the class was annotated with `@lombok.Builder`, but the elements that lombok creates are missing (the inner class, the static `builder()` method, or the `build()` method.  My opinion is that this situation warrants a bubbling exception rather than dirtying the BuilderProvider API with the concern of 'readiness'.  But it would be easy enough to add another method to `BuilderProvider` like `isReady`, and add that check inside of `TypeFactory.canBeProcessed`.  I'll let you make that decision.

> I won't lose too much time on those (lombok)

Unfortunately, I'm stuck with lombok for the time being.  But, I was able to get it work simply by pushing processing of the `Type` to a later round.  The logic seems deterministic (at least for lombok), so I'm feeling okay about it.

> We could also create a different repository and different module for them.

This works for me.  Do you want me to create a repo?  Or do you want to create one under the `mapstruct` umbrella and I can set the project up?  I'll look at `AstModifyingAnnotationProcessor` as a guide.

> how would we know which one to use for which type? 

At least for the three major providers, each has an easily identifiable annotation: `@lombok.Builder`, `@com.google.auto.value.AutoValue`, or `@org.immutables.value.Value.Immutable`.  And from the annotation, we know exactly how the final code will be structured, so it's easy to tell if the annotation processor has completed or not.  

In my first PR, I used `String` to represent the builder factory `builder()` and `build()` methods.  In this PR, I changed those `String` values into `ExecutableElement` and then produced a `MethodReference` to try to align better with `ObjectFactory` for the builder initialization.  My point is that the annotations tell us everything we need to know about the builder (at least for code generation) - it would definitely be simpler if we used `String` -- for example, we wouldn't care if lombok hasn't processed yet, but might lose the ability to register the `builder()` method as an `ObjectFactory` and we'd have to put that in the template, like it was in the last PR.  

> could theoretically add a supports and order methods them we can order them and use the first provider that has supports true. 

Especially for keeping the footprint smaller, why don't we start here?  If there's a `BuilderProvider` on the classpath, it gets loaded into a list, and then we run through each provider in order.  Even the current default `BuilderProvider` implementation could be one of the 3rd-party implementations (so you don't have to maintain it as part of mapstruct core) 

[A slight tangent ]  

At one point, I was trying to break the concept of 'builder' down into something more fundamental... In a sense, a builder is just a submapping.  You're going from mutable > builder > target.  But there's one key difference.  Instead of mapping properties from the builder to the target, it acts more like an `ObjectFactory`, where an instance of a builder produces a new instance of the target.  I see this somewhat like a scoped provider in a DI framework:  given a `Provider` instance, I invoke a method to produce the fully-formed output.  This is how it would be hard-coded:
```
Target.Builder fromSource(Source source);

default Target toTarget(Target.Builder builder) {
  return builder.build();
}
```
For the second method, you could create a way to describe this relationship, maybe as an MappedInstanceFactory?

```
@MappedInstanceFactory(factoryMethod = "build")
Target toTarget(Target.Builder builder);
```
In this PR, we're discussing builders, which assume that `Target.Builder` is an instance that was mapped from a source instance, but if we expanded to think of a DI world, we don't really care where that instance came from, as long as it could be resolved in the current context:
```
@MappedInstanceFactory(factoryMethod = "getSubject")
Subject toUserDetails(@Context SecurityContext context);
```
Anyhow, food for thought - I obviously don't mind scrapping and changing directions if you think this approach has any merit.  I ended up ditching it out of time, and because I couldn't think of many practical use cases.  If this were a valid approach, a `BuilderProvider` would simply connect the dots on intermediate mappings, and let mapstruct that the Builder types aren't to be mapped, but rather you should invoke method `xxx()` on the builder X to produce an instance of type Y

[End tangent.]

I'll make a few changes and push them up.  Thx for the conversation.

filiphr:
@smartytime regarding the type check, why don't you implement an `AstModifyingAnnotationProcessor`? In there you can do the checks that you need to do if the type has `@lombok.Builder` for example. There can be any number of `AstModifyingAnnotationProcessor`(s) present. 

You only need this for Lombok right? It should work ok with the other APTs (such as AutoValue I think). Maybe even Lombok will need to improve their implementation of the processor (we could contribute it to them). I would implement one in the integration test though now within MapStruct.

>This works for me. Do you want me to create a repo? Or do you want to create one under the mapstruct umbrella and I can set the project up? I'll look at AstModifyingAnnotationProcessor as a guide.

Let's wait for this a bit. I think that we can make the `BuilderProvider` much simpler (see my comments below). 

I don't think you need to have all that logic in the `BuilderProvider`. The BuilderProvider should only return the `TypeMirror` of the builder, the discovery of the method for creation of the builder and building the type would be later on done within MapStruct.

During the creation of the `Type` only the builder `Type` would be passed the rest of the methods will follow.

Then when we need to construct the target references we are going to use the Builder type (I think you are doing that already in the PR).

The instantiation of the Builder could then be done in the `MappingResolverImpl#getFactoryMethod` in the case when there are no other found matching methods. The method for creation can look like:

```java
private FactoryMethod findBuilderFactoryMethod(MappingMethod mappingMethod, Type targetType) {
    if ( !targetType.isBuilder() ) {
        return null;
    }

    Type returnType = mappingMethod.getReturnType();
    List<ExecutableElement> builderCreators = new ArrayList<ExecutableElement>();
    for ( ExecutableElement executableElement : executables.getAllEnclosedExecutableElements( returnType
        .getTypeElement() ) ) {
        if ( !executableElement.getModifiers().contains( Arrays.asList( Modifier.PUBLIC, Modifier.STATIC ) )
            || !executableElement.getParameters().isEmpty()
            || !typeUtils.isSameType( executableElement.getReturnType(), targetType.getTypeMirror() )) {
            continue;
        }
        builderCreators.add( executableElement );
    }

     if ( builderCreators.size() == 1 ) {
        FactoryMethod factoryMethod; //construct it
        return factoryMethod;
    }
    else if ( builderCreators.size() > 1 ) {
        //error
        return null;
    }

    // Find the default constructor, if it exists, and construct the FactoryMethod
    // We could also live with assuming it exists
    return null;
}
```

This part is entirely agnostic to the type of the builder. It can be anything. First it looks for a `public static` no arg method that returns the Builder type (the targetType). If one was found we construct a `FactoryMethod`. Otherwise we go and look for a public empty constructor in the target type (heck we can even assume that it exists and just generate code that won't compile, there will be a compile error in any case).

For finding the builder method creator you can do something like:

```java
private String findTargetCreator(Type targetType, Type builderType) {
    if ( !targetType.hasBuilder() || !builderType.isBuilder() ) {
        return null;
    }

    for ( ExecutableElement executableElement : ctx.getExecutables()
        .getAllEnclosedExecutableElements( builderType.getTypeElement() ) ) {
        if ( !executableElement.getModifiers().contains( Modifier.PUBLIC )
            || !executableElement.getParameters().isEmpty()
            || !ctx.getTypeUtils().isSameType( executableElement.getReturnType(), targetType.getTypeMirror() ) ) {
            continue;
        }

        return executableElement.getSimpleName();
    }
//Maybe we need to check that there aren't more...
    return null;
}
```

Then in the `BeanMappingMethod` instead of needing to create the complex `SourceMethod` you can just do `return ${resultName}.${finalizeMethod}();`

The builder provider will then be implemented somehow like: 


```java
protected TypeMirror getLombokBuilder(TypeElement typeElement, Elements elements) {
    lombok.Builder lmbokBuilder = typeElement.getAnnotation( lombok.Builder.class );
    //there would be more checks, but
    if ( lmbokBuilder != null ) {
        TypeElement builder = elements.getTypeElement( typeElement.getQualifiedName() + "Builder" );
        return builder == null ? null : builder.asType();
    }
    return null;
}

protected TypeMirror getAutoValueBuilder(TypeElement typeElement, Elements elements) {
    AutoValue autoValue = typeElement.getAnnotation( AutoValue.class );
    if ( autoValue == null ) {
        return null;
    }
    for ( Element element : typeElement.getEnclosedElements() ) {
        if ( !element.getKind().isInterface()
            || !element.getKind().isClass() ) {
            continue;
        }
        AutoValue.Builder autoValueBuilder = element.getAnnotation( AutoValue.Builder.class );
        if ( autoValueBuilder != null ) {
            return element.asType();
        }
    }
    return null;
}

protected TypeMirror getValueImmutableBuilder(TypeElement typeElement, Elements elements) {
    Value.Immutable valueImmutable = typeElement.getAnnotation( Value.Immutable.class );
    if ( valueImmutable == null || !valueImmutable.builder()) {
        return null;
    }
    TypeElement builder = elements.getTypeElement( name );
    return builder == null ? null : builder.asType();
}
```
Each library has it's own way of getting the types.

*** 

Regarding your tangent, I was also thinking something in the direction of that. I think that my proposals above are showing something which can be simpler...

Additional food for thought: theoretically we can send the source and `@Context` parameter to the builder as well. However, I would propose that we don't go that deep and make it as simple as possible. Then we can iterate in new PRs. Once we have a working support for the most simple cases the rest will adjust easily :)

One important thing the `FactoryMethod` I created in the example method does not exist. I tried creating an interface that extends `Assignment` and replacing all places where we use `MethodReference` for the `FactoryMethod` and it compiles, so it means that we don't need anything from the `MappingMethod` we can pull everything that we need into the `MethodReference`. Therefore (if we go with such an approach), I would propose that we first add such interface in a separate PR and then continue working from there (I can do that if you want me to).

filiphr:
Regarding the `BuilderNamingStrategy`, I think that we can assume that the passed type will always be a builder type, so you don't need to look for it

smartytime:
> You only need this for Lombok right? It should work ok with the other APTs (such as AutoValue I think). Maybe even Lombok will need to improve their implementation of the processor (we could contribute it to them). I would implement one in the integration test though now within MapStruct.

I really appreciate your help in understanding how to create the simplest and easiest-to-maintain solution.  All things the same, simpler is better.

If we make the builder configuration values a string instead of `Method` or `Assignment`, it will make a few things easier/smaller,  and could actually solve the lombok problem completely.  

The more I play with this, the less comfortable I am with having the code making decisions about what is a builder and what's not.  IMO, any heuristic we follow by inspecting the code should be strict and deterministic.  False positives could have really bad side effects.

Using library-specific implementations is a big first step, but I would take it one step further and say that each `BuilderProvider` SHOULD make a best effort to return as much deterministic builder configuration as possible (no guessing), and should provide _at least_ the TypeMirror for the builder type.  The other values that are important for making correct mappings:

* The name of the static method, if any, that produces a new empty instance of the builder.  If none is provided, a no-arg constructor is favored, followed by a static method on the result type class
* The name of the instance method on the builder class that finalizes that builder and produces a new instance of the result type
* (Optionally), the name of a thaw method.  Mapstruct provides the ability to remap values, so I could see dropping support for thaw, but it's possible that there's some additional logic or initialization to `thaw()` that mapstruct couldn't replicate. 

If the `BuilderProvider` can't provide a reasonable guarantee for these values, it should return null for that property, which signals to Mapstruct to defer resolution and let Mapstruct handle it.

Using this mechanism, lombok can provide exact configuration values for all these properties every time.  Immutables can in most instances, assuming you're not using some of the more crazy features like package-level style annotations.  For the edge cases, we defer resolution.  But if the annotation is sitting there telling us precisely what the configuration values are, we should use them.  I don't see any good in ignoring those values.

> I don't think you need to have all that logic in the BuilderProvider. The BuilderProvider should only return the TypeMirror of the builder, the discovery of the method for creation of the builder and building the type would be later on done within MapStruct.

I think I stated this above, but I agree with you _except_ in the case where we can easily and deterministically pull these values from annotations.  I believe we should still provide a container like `BuilderInfo` or `BuilderModel` for returning all values, even though certain implementations would only return a TypeMirror.

> The instantiation of the Builder could then be done in the MappingResolverImpl#getFactoryMethod in the case when there are no other found matching methods. The method for creation can look like:

This sounds good.  I'd vote for an accessible no-arg constructor first, static method second. 

> Then in the BeanMappingMethod instead of needing to create the complex SourceMethod you can just do return ${resultName}.${finalizeMethod}();

Yes, agreed.

Two last thoughts:

1. Finding a builder from a resultType is usually pretty straightforward, and depending on the library, going the other way might be simple.  However, there are cases, like homegrown builders or lombok, where it may not be deterministic for Mapstruct to identify that a class is a builder, and to work backwards to identify what type it builds.  In most cases, it will be a static inner class, and we'll have a method like build(), and the class will be named something obvious like `Builder`, but it's something to consider.  Like you said, worst case, we can just fail to compile...  I do think each `BuilderProvider` should also be responsible for that reverse lookup, because (at least for the case of lombok), it will need to inspect explicit annotations.

2.  For builderNamingStrategy - should I just add methods to the existing AccessorNamingStrategyInterface?  `getBuilderPropertyName`.  Does there really need to be an entirely new class?

filiphr:
@smartytime I haven't had the chance to have a look at your last comment. I have been swamped with other things and I haven't had the chance to have a better look. Just wanted to let you know that this is not forgotten :)

gunnarmorling:
This is for post-1.2, too, correct?

filiphr:
Yes post 1.2

filiphr:
@smartytime we've released `1.2.0.Final` and I finally found some time to reply back to you. We want to include this in the next 1.3 release, which we want to release it sooner than what we did for 1.2 :).

> The more I play with this, the less comfortable I am with having the code making decisions about what is a builder and what's not.

The benefit of MapStruct trying to find out what is a builder is in the fact that we would be able to work out of the box for a lot of frameworks (but also hand written builders). In a nutshell everyone that follows the builder pattern. I would really like to avoid the need for people to pull in more dependencies or needing to define which Builder providers they are going to use. 

I also doubt that the writers of the other libraries would want to take over the modules that would be for their annotations. If people adhere to the builder pattern, it is easy not to make false positive. We would have strict rules that we would use and if we follow them we should be safe.

You are right that the annotations from the other providers are there and we could use them. However, that ties us to their own implementation (which I would very much like to avoid).

>This sounds good. I'd vote for an accessible no-arg constructor first, static method second.

It is fine for me. I don't have a strong preference in any case.

For the `BuilderNamingStrategy`. Maybe adding it to the `AccessorNamingStrategy` is OK, but that would be a breaking change, in case people are doing their own implementation and we can't use Java 8 `default` yet. I would like to hear what the others from the team think about it. In any case using own interface for now, should be OK. We can always change that in the end.

dpolivaev:
I would like to ask how easy the feature could be used with builders generated by http://immutables.github.io/ . Have you considered it?

filiphr:
@dpolivaev the main goal is to support the most known / used frameworks out of the box. Immutables is one of them. Also the solution aims to be independent of a particular 3rd party implementation and support the builder pattern

anjo-swe:
@smartytime @filiphr any updates?

filiphr:
@anjo-swe I haven't heard from @smartytime about my last comments. It is up to him if he wants to continue working on this. If he is not able to, we would need to see who can takeover and wrap this up. We want to have this feature in 1.3.0, it is by far one of the most requested one :)

gunnarmorling:
@filiphr, so in terms of an annotation controlling the behavior of the default SPI implementation, I was thinking of something like this:

    @interface BuilderConfig {
        String builderCreationMethod() default "<CLASS_NAME>.builder";
        String buildMethod() default "build";
    }

If needed, the defaults could be overridden per method/mapper/mapper config:

    @Mapper
    @BuilderConfig(builderCreationMethod="<CLASS_NAME>Builder.newInstance", buildMethod="finalize")
    public interface CarMapper {
        ...
    }

Btw. what is the "thaw" method mentioned in the discussion above?

filiphr:
@gunnarmorling are missing the name of the builder class itself or that would acquired from the `<CLASS_NAME>.builder` return type?

>Btw. what is the "thaw" method mentioned in the discussion above?

I am not entirely sure. What I could find is [this](https://github.com/fge/btf/wiki/The-freeze-thaw-pattern), it is some kind of a reversible builder method, where you can access the builder prepopulated with the values of a type and then update it and return a new type back.

gunnarmorling:
> are missing the name of the builder class itself or that would acquired from the <CLASS_NAME>.builder return type?

Yes, exactly, it can be obtained from the return type of that method.

> What I could find is this ...

Ok, thanks. I'd say let's leave it out then for the first round and wait whether there's demand for it.

dpolivaev:
Folks, how are you doing? I am waiting for this feature like forever. Do you need any help in development?

gunnarmorling:
Sorry, seems everyone was doing a winter break :) We'll discuss this in next week's team hangout and hopefully get it merged and shipped soon.

filiphr:
Closing this as #1373 has been merged. @ericmartineau thanks a lot for helping out with this. Sorry that it took so long

Issue #744, opened 2016-02-08
Enable to use MapStruct with JDK 9 Jigsaw


gunnarmorling:

agudian:
Pasting a mail from Rory O'Donnell, I guess it's time we get things working in the integration-tests at least.

Hi,
 
* JDK 9 Developer Preview is now available on java.net [1]

Developer Preview milestone: - A reasonably stable build suitable for
broad testing by the developer community is available.
JDK 9 Builds 163 and higher include all planned features.


*Attention annotation processing users and authors - * Request for
feedback on annotation processing API changes made in JDK 9.

As has been done previously during Java SE 7 and Java SE 8, the JSR 269
annotation processing API is undergoing a maintenance review (MR) as
part of Java SE 9. Details of the changes in JDK 9 Early Access build
163 & build 164 available here [2]

Please report experiences running processors under JDK 9 and feedback on
the API changes to the compiler-dev mailing list.
(If you haven’t already subscribed to that list then please do so first,
otherwise your message will be discarded as spam.)

Rgds, Rory

[1] https://jdk9.java.net/download/
[2]
http://mail.openjdk.java.net/pipermail/compiler-dev/2017-April/010896.html

--
Rgds,Rory O'Donnell
Quality Engineering Manager

filiphr:
Here are some pointers about setting up with Travis if we want to have that:

* [JUnit 5 Testing with JDK-9](https://github.com/junit-team/junit5/issues/296) - There are a lot of links there that show how it can be done. Also their [Travis Configuration](https://github.com/junit-team/junit5/blob/master/.travis.yml)
* [Travis JDK-9 Support](https://github.com/travis-ci/travis-ci/issues/5520)

gunnarmorling:
+1 for testing this!

gunnarmorling:
One thing to address is that the @Generated annotation got a
replacement just recently, there is now
javax.annotation.processing.Generated.

netmikey:
JDK9 was released yesterday, so we tried to build our projects with it to see what happens.

First thing happening was that mapstruct generation failed. We're using the Gradle Plugin as described [here](http://mapstruct.org/documentation/installation/#gradle) with the latest Gradle 4.2 and the latest stable MapStruct 1.1.0.Final, but also tested with 1.2.0.CR2 - same result.

The reason is probably that JAXB isn't accessible directly anymore in the modularized JDK9.

Here's the Stacktrace from 1.2.0.CR2:

```
[...]

PR #1302, opened 2017-09-24
Java 9 compatibilities

filiphr:
The first commits makes sure that Mapstruct works correctly under Java 9 compilations and mix of Java 9 and source 1.8. I noticed that we area actually using types from 2 modules, `java.xml.bind` (`JAXBElement`) and `java.xml.datatype` (`XMLGregorianCalendard`). 

The main issue in #744 is that when Java 9 is used for compilation and the source / target version is 1.8 then `Elements#getTypeElement(CharSequence)` does not take the modules into consideration. Which means that it will say that `JAXBElement` is there, but we can't really use it as the module has not been added. Therefore, I added the `ClassUtils` that checks explicitly if the class can be loaded.

I've also tried to add Java 9 to the toolchains, some tests will fail. Can you help me out for this @gunnarmorling and @agudian?

### DIFF
--- a/etc/toolchains-travis-jenkins.xml
+++ b/etc/toolchains-travis-jenkins.xml
@@ -33,16 +33,5 @@
             <jdkHome>/usr/lib/jvm/java-8-oracle/</jdkHome>
         </configuration>
     </toolchain>
-    <toolchain>
-        <type>jdk</type>
-        <provides>
-            <version>9</version>
-            <vendor>oracle</vendor>
-            <id>jdk9</id>
-        </provides>
-        <configuration>
-            <jdkHome>/usr/lib/jvm/java-9-oracle/</jdkHome>
-        </configuration>
-    </toolchain>
-
+    
 </toolchains>
--- a/etc/toolchains-travis-jenkins.xml
+++ b/etc/toolchains-travis-jenkins.xml
@@ -33,5 +33,16 @@
             <jdkHome>/usr/lib/jvm/java-8-oracle/</jdkHome>
         </configuration>
     </toolchain>
-    
+    <toolchain>
+        <type>jdk</type>
+        <provides>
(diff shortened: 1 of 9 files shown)


gunnarmorling:
Hey @filiphr, weird that there is no way to find out the available modules using the annotation processing API. It may be worth asking on Stack Overflow; I found [this question](https://stackoverflow.com/questions/46266935/how-to-get-list-of-all-modules-programmatically-in-java-9) which is related, but I think it's not yet the right answer for our case. Otherwise, your approach seems good.

Something is wrong with the Travis build, though. Seems like somehow Java 8 is picking up classes compiled to the bytecode level of Java 9 (53).

filiphr:
@gunnarmorling when the target version is Java 9 it works correctly. I think that the annotation processing API is working correctly as under Java 8 the `JAXBElement` is available. Maybe the fix should be not to use those classes like we use them and just use their FQN.

As for the build it is strange, I think it is something with the version of java that maven picks up when it is being run. I think that we should switch our build to use Travis matrices and run in parallel on both Java 8 and Java 9, instead of using the toolchains.

filiphr:
@gunnarmorling is it OK if we revert the Java 9 test setup and we merge this? We can improve our test infrastructure for the 1.3 release

gunnarmorling:
@filiphr sorry for the late reply, I was travelling in the last days. +1 for excluding the test bits and moving on.

filiphr:
@gunnarmorling yes I saw on Twitter (you've been pretty busy with conferences 😄). I myself have been pretty busy as well. I've reverted the Java 9 tests. Should I merge after the tests are done?

filiphr:
I think that the uncovered lines are OK

gunnarmorling:
> Should I merge after the tests are done?

Yes, please!

PR #1310, opened 2017-10-19
Add IntelliJ Formatter to CONTRIBUTING.md

### LINKED ISSUE #1297: Add somewhere that we have an IntelliJ formatter
The IntelliJ formatter for our codebase is actually more up to date then the eclipse one. Make sure that we write this somewhere, so contributors can use it.

The formatter can be found [here](https://github.com/mapstruct/mapstruct/blob/master/etc/mapstruct.xml)

kevcodez:
This commit fixes #1297


filiphr:
Merged. Thanks @kevcodez :)

PR #1316, opened 2017-10-22
Followup to #610 unmappedSourcePolicy, ignore=true with no target

spoerri:


filiphr:
I have opened #1317 for this PR.

By the way can you please rebase it @spoerri. I think it has some issues with the merge of your previous PR

spoerri:
Rebased, but not nicely, so I'll squash at some point

As the failed test shows, I'm doing something wrong, and I'm not sure what :)...

filiphr:
@spoerri thanks for your patience. I would suggest you too look into `TargetReference.BuilderFromTargetMapping#build()` for some reason the target name on the mapping is not `null` (which is strange).

I'll need to have a better look at the implementation as well. It looks a bit complex on first sight

spoerri:
Thanks for feedback @filiphr 

I simplified and fixed

Could you review again even before I fix the integration test and rebase?

filiphr:
@spoerri I had a look at the finished PR and we also had a discussion within the team. The only issue that we have is that the `Mapping#target` becomes optional. Which is something that we think should not happen. 

However, we have another suggestion, we can add a new property to the `@BeanMapping` annotation. Something like `ignoredSourceProperties` this way we won't need to change the `@Mapping` and it would be much easier, we would only need to remove the appropriate source methods at the beginning. In case some of the defined ignored properties are not part of the source object then a warning should be emitted saying that.

I hope you understand and this is OK for you.

core-jdk8/src/main/java/org/mapstruct/Mapping.java
```diff
@@ -58,7 +58,7 @@
      *
      * @return The target name of the configured property or enum constant
      */
-    String target();
+    String target() default "";
```
gunnarmorling: I don't think we should make `target()` optional. The idea of `@Mapping` has always been that it describes a target property - so its name is mandatory - and then the different ways for how it could get its value (simple source property, expression, constant etc.).

Thinking about it with @sjaakd and @filiphr, a better approach would be to add `@BeanMapping#ignoredSourceProperties()`. That way the semantics of `@Mapping` aren't changed and you still get the feature you're after.

WDYT?

processor/src/test/java/org/mapstruct/ap/test/unmappedsource/IgnoreMapper.java
```diff
@@ -0,0 +1,42 @@
...
+import org.mapstruct.ReportingPolicy;
+import org.mapstruct.factory.Mappers;
+
+@Mapper(unmappedSourcePolicy = ReportingPolicy.WARN,
+        unmappedTargetPolicy = ReportingPolicy.IGNORE)
+public interface IgnoreMapper {
+
+    IgnoreMapper INSTANCE = Mappers.getMapper( IgnoreMapper.class );
+
+    @org.mapstruct.Mappings({
```
gunnarmorling: Is there any reason for not importing these types? Also `@Mapping*s*` could be omitted.

core-jdk8/src/main/java/org/mapstruct/Mapping.java
spoerri: Grumble, grumble, could’ve saved me a throw-away implementation by having that discussion earlier. Ok, now that’s out of the way... :)

Is there anything to say for a consistent approach to “ignore” mappings, supporting both source & target via either @BeanMapping or via @Mapping?

Also, with @BeanMapping is there any ambiguity between an ignored source property and a forged mapping? With @Mapping, ignored source should clearly prevent forgery.

spoerri:
Before doing an alternate implementation, just want to confirm: is it worth it to keep ```target``` always required, if that means:
1) having an inconsistent API
2) giving up the natural ability to suppress mapping of source property
?

That is, ignoring a mapping has slightly different semantics than ignoring a warning. Ignoring a mapping is more powerful. For example, there is a field X on both source and target, and we'd like to prevent mapping. Now, of course we could implement that behavior using the ```@BeanMapping ``` annotation API, but it's more natural on ```@Mapping```.

Giving up the build-time safety of required ```target``` would be a shame, but the annotation API is already not build-time safe,  e.g. we can have ```source``` and ```expression``` specified at the same time.

gunnarmorling:
Hi @spoerri,

> if that means:
> 
> * having an inconsistent API

In which way is it inconsistent?

>  * giving up the natural ability to suppress mapping of source property

I'm not following, why would we give up this possibility? Source properties without a counterpart are default by ignored. If there's a corresponding property, `@Mapping#ignore()` can be used. And in case you want to suppress the warning about an unmapped source property if there's no corresponding property on the target side, `@BeanMapping#ignoredSourceProperties()` would let you do this. Which part is missing from your POV?

>  e.g. we can have source and expression specified at the same time.

Ah, that actually should be flagged as an error. Only one of `source()`, `expression()` and `constant()` may be given.

spoerri:
It’d be nice to be able to suppress *mapping* by source as well as by target, instead of only being able to suppress *warning* by source. And it need only be specified once to work with `@InheritInverseConfiguration`.

All illegal combinations can be flagged as errors, including missing both target and ignore. It is nice to target be listed as required in the annotation itself, but IMHO not essential because none of the other conditions are listed in the annotation itself.

gunnarmorling:
> It’d be nice to be able to suppress mapping by source as well as by target

I don't see the point of supporting this. If there is no target property corresponding to a given source property (same name), it will not be mapped. Nothing needs to be done for that. If there is a corresponding target property, it can be ignored via `@Mapping(target="x", ignore=true)`.

But maybe we should take a step back and discuss the use case for this first. You said:

> For example, we have a mapper configured to warn on unmapped source fields, and we’d like to prevent mapping a field X that is actually present on both source and target.

If it's present on both sides, why don't you simply use `@Mapping(target="x", ignore=true)`? The key thing is that the `@Mapping` annotations provide the info needed for populating the target bean of a mapping, hence `target()` is mandatory.

spoerri:
I hear... It would be possible to use a combination of `@BeanMapping#ignoredSourceProperties` and `@Mapping(target, ignore)`, just not so lovely.

In my app, when we map from our incoming message DTOs to our internal domain entities, we use unmappedTargetPolicy and `@Mapping(target, ignore)`. When we map from entities to outgoing message DTOs we use unmappedSourcePolicy, and it just feels natural to use the similar syntax, i.e. `@Mapping(source, ignore)`.

filiphr:
@spoerri sorry for not getting back to you for such a long time. We have discussed this PR again and we would like to go with the approach `@BeanMapping(ignoredSourceProperties)` where the user can provide a list of properties.

We would like to keep the consistency of `@Mapping` to be mainly for targets. The `unmappedSourcePolicy` is highly specialised case and we don't think that it is worth making `@Mapping(target)` optional for that

filiphr:
Closing in favour of #1448

PR #1322, opened 2017-10-23
#1320 Properly create additional options for unprocessed defined targets

### LINKED ISSUE #1320: Not generating setter methods for all constants
MapStruct not generating setter methods for all constants. For the below mapper, code was generated only for `myState`. This is happening only for third level parameters.
```java
public interface MyMapper {
	@Mapping(target = "address.country.cityName", constant="myCity")
	@Mapping(target = "address.country.stateName", constant ="myState")
    Target map(Integer dummy);	
}
```

filiphr:
Create mappings for each unprocessed defined target based on their name and the mapping.

Fixes #1320.

@sjaakd guess what, the `Map<String, List<Mapping>>` hitting us again :). Really looking forward to 2.0.


sjaakd:
looks good.. please merge 👍

filiphr:
Thanks for the review @sjaakd

Issue #1332, opened 2017-11-15
Exceptions in nested beanmapping is not declared in generated code


chigix:
I have meet a problem when my getters in bean declaration is with exceptions thrown. For a nested bean mapping for just only 2 layers, this problem is not appeared because all the possible exceptions have been declared explicitly by user. But for a deeper mapping, such exception is not generated with generated private methods by mapstruct. For describing this problem,  I have also pushed a branch in my fork: https://github.com/chigix/mapstruct/tree/Exception_in_nested_mapping

I'm not sure is it a design originally to not cover exceptions thrown around getters or setters in pure mapping. However, I think it would not establish negative effect.

filiphr:
@chigix I think it is a major antipattern to throw checked exceptions from getters / setters. What you are doing in your getter is something that should be done in a service. This will also allow you to test your code better.

The issue occurs when you have nested source mappings. Theoretically we could do something about it, but it would add some complexity to the code and I am not sure if it is really worthwhile

chigix:
@filiphr Well, exactly as you said, it breaks right design pattern, however, I'm not doing pattern design code, it is a part of implementation for a repository, where the bean is also not domain model, which is directly a rich model proxy from domain model,  for usage restricted in dao ( or repository ) implementation layer, rather than the pattern usage across the whole project or any consideration about reusability or visibility. 

Actually, I think it is no need for me to create a service only used in implementation layer or pass another service object with the model where nested information has been cached and included.

So I really hope this support could be provided from mapstruct build.

filiphr:
@chigix you are right maybe we need to take this into consideration as well. 

Would you like to submit a PR for this? I can of course guide you through the process.

In this particular example the `NestedPropertyMappingMethod` and NestedPropertyMappingMethod.ftl` need to be modified, so exceptions are taken into consideration.

chigix:
@filiphr  OK, I'll have a try in this weekend. Any problem, however, if appeared I'll email you directly for potential instruction and guide.

filiphr:
@chigix no problem. You can also reach me on our [gitter chat](https://gitter.im/mapstruct/mapstruct-users)

filiphr:
Fixed via #1335 thanks @chigix for the PR

PR #1335, opened 2017-11-19
#1332 Fix exceptions declaration missing in nested mapping methods

chigix:
Issue occurs when exceptions declared in nested source mapping, which would make compilation failures for unreported exceptions.

Reference to the testing case: https://github.com/chigix/mapstruct/tree/Exception_in_nested_mapping

Signed-off-by: RichardLea <chigix@zoho.com>

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
@@ -19,7 +19,7 @@
      limitations under the License.
 
 -->
-<#lt>private <@includeModel object=returnType/> ${name}(<#list parameters as param><@includeModel object=param/><#if param_has_next>, </#if></#list>) {
+<#lt>private <@includeModel object=returnType/> ${name}(<#list parameters as param><@includeModel object=param/><#if param_has_next>, </#if></#list>)<@throws/> {
     if ( ${sourceParameter.name} == null ) {
         return ${returnType.null};
     }
@@ -43,3 +43,11 @@
 </#list>
 }
 <#macro localVarName index><#if index == 0>${sourceParameter.name}<#else>${propertyEntries[index-1].name}</#if></#macro>
+<#macro throws>
+    <#if (thrownTypes?size > 0)><#lt> throws </#if><@compress single_line=true>
+        <#list thrownTypes as exceptionType>
+            <@includeModel object=exceptionType/>
+            <#if exceptionType_has_next>, </#if><#t>
+        </#list>
+    </@compress>
+</#macro>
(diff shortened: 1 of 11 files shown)


processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
```diff
@@ -19,7 +19,7 @@
      limitations under the License.
 
 -->
-<#lt>private <@includeModel object=returnType/> ${name}(<#list parameters as param><@includeModel object=param/><#if param_has_next>, </#if></#list>) {
+<#lt>private <@includeModel object=returnType/> ${name}(<#list parameters as param><@includeModel object=param/><#if param_has_next>, </#if></#list>) <@throws/> {
```
filiphr: The generated code fails because you have `) <@throws/> {` (mind the whitespace between `)` and `<@throws`. If you replace that with `)<@throws/> {` then those tests won't fail. Have a look how it is in the `BeanMappingMethod.ftl`

filiphr:
Thanks for the PR @chigix, I have merged it into master

levimdmiller:
This bug seems to still exist in 1.3.0.Final:
(I know it's an antipattern, but the getter is in an external library so I can't change it)

`
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /C:/Users/levi.miller/dev/flint/target/generated-sources/annotations/com/qhrtech/fhir/transformer/mapstruct/mapper/MedicationDispenseMapperImpl.java:[420,103] unreported exception org.hl7.fhir.exceptions.FHIRException; must be caught or declared to be thrown
`

```
protected AnnotationModel annotationToAnnotationModel(Annotation annotation) throws FlintException {
        if ( annotation == null ) {
            return null;
        }
        AnnotationModel annotationModel = implementationFactory.getDefaultImplementation( AnnotationModel.class );

        if ( annotation.hasAuthorReference() ) {
            annotationModel.setAuthorReference( referenceMapper.toFlint( annotation.getAuthorReference() ) );
        }
        if ( annotation.hasTime() ) {
            annotationModel.setTime( new DateTime( annotation.getTime() ) );
        }
        if ( annotation.hasText() ) {
            annotationModel.setText( annotation.getText() );
        }

        return annotationModel;
    }
```

    ```
    /**
     * @return {@link #author} (The individual responsible for making the annotation.)
     */
    public Reference getAuthorReference() throws FHIRException { 
      if (this.author == null)
        return null;
      if (!(this.author instanceof Reference))
        throw new FHIRException("Type mismatch: the type Reference was expected, but "+this.author.getClass().getName()+" was encountered");
      return (Reference) this.author;
    }
    ```

Issue #1336, opened 2017-11-19
Inconsistent ADDER_PREFERRED


wenerme:
```java
if ( builder.getBrandsList() != null ) {
    /* Expected output
    for (MyDTO term : s.getBrands()) {
        builder.addBrands(mapParseTerm(term));
    }
    */
    // But got this
    List<NeoMyDTO> list = parseTermListToParseTermList( s.getBrands() );
    if ( list != null ) {
        builder.getBrandsList().addAll( list );
    }
}
```

Why use `addAll` if there is `addBrands`, I also create a demo project, but always use `addBrands`, how can I control this ?

filiphr:
@wenerme are you using Protobuf? Your builder looks a lot like protobuf. Have you tried using something like we have in the [mapstruct-protobuf3](https://github.com/mapstruct/mapstruct-examples/tree/master/mapstruct-protobuf3) example?

The reason why your adder is taken into consideration is because MapStruct checks for `addBrandsList` method. As your target property is `brandsList`

Can you share a demo project, so I can help you a bit more?

wenerme:
Yes, it's pb3, may target property is `brands`, because I removed `List` from pb's property name, so, I don't have to add Mapping for every collections.

```
    private boolean isMessageOrBuilder(ExecutableElement element) {
        Element e = element.getEnclosingElement();
        if (e != null && e.getKind() == ElementKind.CLASS) {
            if (((TypeElement) e).getSuperclass().toString().startsWith(PROTOBUF_GENERATED_MESSAGE_V3)) {
                return true;
            }
        }
        return false;
    }

    @Override
    public String getPropertyName(ExecutableElement element) {
        String name = super.getPropertyName(element);
        if (isMessageOrBuilder(element)) {
            if (name.endsWith("List")) {
                name = name.substring(0, name.length() - 4);
            }
        }
        return name;
    }
```

wenerme:
I found why this failed, the example not failed, because there only one candidate for adders

```java
        if ( candidates.isEmpty() ) {
            return null;
        }
        else if ( candidates.size() == 1 ) {
            return candidates.get( 0 );
        }
        else {
            for ( Accessor candidate : candidates ) {
                String elementName = Executables.getElementNameForAdder( candidate );
                if ( elementName != null && elementName.equals( Nouns.singularize( pluralPropertyName ) ) ) {
                    return candidate;
                }
            }
        }
```

[...]

PR #1337, opened 2017-11-20
When adder has multi candidates, check element name match pluralPropertyName

wenerme:
Relate issues https://github.com/mapstruct/mapstruct/issues/1336

MapStruct should check adder name match "expected plural name" not only singularize property name.


filiphr:
@wenerme as I have mentioned in the issue, sorry but I am inclined to reject this PR. The reason is that, I think that MapStruct is doing a correct job when generating the code. 

If you want can contribute your protobuf ehnancement in our examples repository.

filiphr:
I have closed this as we have already resolved it as part of #1336

PR #1341, opened 2017-11-22
Correctly use value mapping source parameters

### LINKED ISSUE #1340: Enum mapper generates mapping method with wrong signature. Wrong amount of parameters
Considering that `Quote` and `QuoteDTO` are just enums

`@Mapper
public interface QuoteMapper {
    QuoteDTO mapToDto(@Context MappingContext context, Quote source);
}`

generates the following

`public class QuoteMapperImpl implements QuoteMapper {
    @Override
    public QuoteDTO mapToDto(MappingContext context) {
        ...
    }
}`

As you can see only `context` parameter is passed. `source` is missed.
If order of parameters in mapper interface is changed, so 

`QuoteDTO mapToDto(Quote source, @Context MappingContext context);`

then generated method looks like:

`public QuoteDTO mapToDto(Quote source) {...}`

So, in this case `context` is missing. 
Well, all the time the second argument.

With regular classes, not enums, everything works as expected

filiphr:
Fixes #1340 

* The Value / Enum Mappings should copy all the parameters from the method that is being overriden
* The source parameter should be the first source parameter

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/EnumMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/EnumMappingMethod.java
@@ -209,6 +209,6 @@ public class EnumMappingMethod extends MappingMethod {
     }
 
     public Parameter getSourceParameter() {
-        return first( getParameters() );
+        return first( getSourceParameters() );
     }
 }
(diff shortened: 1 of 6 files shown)


gunnarmorling:
LGTM, thanks, @filiphr!

PR #1342, opened 2017-11-22
Primitive Context parameters should be skipped when looking for lifecycle methods

### LINKED ISSUE #1339: Context annotation on primitive types with AfterMapping throws exception on compile
The `@Context` feature in 1.2.0 seems to generate a compile time NullPointerExceptions when used on primitives (for example boolean). I'm not sure if this is intended behavior, and I couldn't find anything in the documentation suggesting that it was intentional. Similar code seems to work in 1.1.0 without the Context annotation.

For example, I think the following use of `@Context` will reproduce the issue:

```
@Mapping(target = "field1", source = "src.field1")
@Mapping(target = "primitiveField1", source = "primitive1")
@Mapping(target = "someField", ignore = true)
public abstract TestTarget map(TestSource src, boolean primitive1, @Context boolean primitive2);

@AfterMapping
public void afterMap(@MappingTarget TestTarget target, @Context boolean primitive2)
{
    //do some complicated mapping logic involving primitive2
    int result = someComplicatedFunction(primitive2);
    target.setSomeField(result);
} 
```

When I try to compile the above example with map-struct I get the following exception:
```
[ERROR] diagnostic: <filename> error: Internal error in the mapping processor: java.lang.NullPointerException       at 
org.mapstruct.ap.internal.util.workarounds.SpecificCompilerWorkarounds.replaceTypeElementIfNecessary(SpecificCompilerWorkarounds.java:112)     at 
org.mapstruct.ap.internal.util.Executables.getAllEnclosedAccessors(Executables.java:208)     at 
org.mapstruct.ap.internal.util.Executables.getAllEnclosedExecutableElements(Executables.java:188)
...
```

Just removing the `@Context` parameter does not seem to work because mapstruct generates the following code (substituting in the wrong parameter)
```
afterMap(testTarget, primitive1);
```

Using 1.1.0 (without the `@Context` annotation) my example works fine, and generates the following mapping, as I would expect.
```
afterMap(testTarget, primitive2);
```

filiphr:
Fixes #1339 

Primitive context parameters should not be searched for lifecycle methods

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
@@ -276,6 +276,9 @@ public class MethodRetrievalProcessor implements ModelElementProcessor<Void, Lis
 
         ParameterProvidedMethods.Builder builder = ParameterProvidedMethods.builder();
         for ( Parameter contextParam : contextParameters ) {
+            if ( contextParam.getType().isPrimitive() ) {
+                continue;
+            }
             List<SourceMethod> contextParamMethods = retrieveMethods(
                 contextParam.getType().getTypeElement(),
                 mapperToImplement,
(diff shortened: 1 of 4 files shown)


gunnarmorling:
LGTM, thanks, @filiphr!

PR #1364, opened 2017-12-29
#1353 Add a trim for source and target mappings

### LINKED ISSUE #1353: @Mapping arguments should be trimmed
I've noticed that if you add a space in front of a mapping, you get an error. However, the error is not related to a property but to the complete method which makes it hard to debug (see #1354). Its questionable whether this should be seen as an error or just been ignored.

smythie86:
Add a trim for source and target mappings, but throw a warning message if trimmed.

Fixes #1353

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
@@ -108,8 +108,14 @@ public class SourceReference {
             boolean isValid = true;
             boolean foundEntryMatch;
 
+            String sourceNameTrimmed = sourceName.trim();
+            if ( !sourceName.equals( sourceNameTrimmed ) ) {
+                messager.printMessage( method.getExecutable(), mapping.getMirror(), mapping.getSourceAnnotationValue(),
+                                       Message.PROPERTYMAPPING_WHITESPACE_TRIMMED, mapping.getSourceName()
+                );
+            }
             String[] sourcePropertyNames = new String[0];
-            String[] segments = sourceName.split( "\\." );
+            String[] segments = sourceNameTrimmed.split( "\\." );
             Parameter parameter = null;
 
             List<PropertyEntry> entries = new ArrayList<PropertyEntry>();
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
@@ -110,8 +110,13 @@ public class SourceReference {
 
             String sourceNameTrimmed = sourceName.trim();
             if ( !sourceName.equals( sourceNameTrimmed ) ) {
-                messager.printMessage( method.getExecutable(), mapping.getMirror(), mapping.getSourceAnnotationValue(),
-                                       Message.PROPERTYMAPPING_WHITESPACE_TRIMMED, mapping.getSourceName()
+                messager.printMessage(
+                    method.getExecutable(),
+                    mapping.getMirror(),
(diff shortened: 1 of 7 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/TargetReference.java
```diff
@@ -138,8 +138,13 @@ public TargetReference build() {
                 return null;
             }
 
-
-            String[] segments = targetName.split( "\\." );
+            String targetNameTrimmed = targetName.trim();
+            if ( !targetName.equals( targetNameTrimmed ) ) {
+                messager.printMessage( method.getExecutable(), mapping.getMirror(), mapping.getSourceAnnotationValue(),
```
filiphr: This needs to be `mapping.getTargetAnnotationValue()` instead of `mapping.getSourceAnnotationValue()`. This property is actually not used in Java 8, but it is used correctly under Java 9, and not because of us, but because it was correctly implemented in the JDK only then.

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -61,6 +61,7 @@
     PROPERTYMAPPING_NO_PRESENCE_CHECKER_FOR_SOURCE_TYPE( "Using custom source value presence checking strategy, but no presence checker found for %s in source type." ),
     PROPERTYMAPPING_NO_READ_ACCESSOR_FOR_TARGET_TYPE( "No read accessor found for property \"%s\" in target type." ),
     PROPERTYMAPPING_NO_WRITE_ACCESSOR_FOR_TARGET_TYPE( "No write accessor found for property \"%s\" in target type." ),
+    PROPERTYMAPPING_WHITESPACE_TRIMMED( "The property name \"%s\" has whitespace that was trimmed.", Diagnostic.Kind.WARNING ),
```
filiphr: Maybe the message `The property named \"%s\" has whitespaces, using trimmed property \"%s\" instead.` is more appropriate (similar to `PROPERTYMAPPING_INVALID_PROPERTY_NAME`). And the users will get the actual property that is being used. Not sure if it too much

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
```diff
@@ -108,8 +108,14 @@ public SourceReference build() {
             boolean isValid = true;
             boolean foundEntryMatch;
 
+            String sourceNameTrimmed = sourceName.trim();
+            if ( !sourceName.equals( sourceNameTrimmed ) ) {
+                messager.printMessage( method.getExecutable(), mapping.getMirror(), mapping.getSourceAnnotationValue(),
```
filiphr: Can you format this and the one in `TargetReference` to:

```
messager.printMessage( 
    method.getExecutable(), 
    mapping.getMirror(), 
    mapping.getSourceAnnotationValue(),
    Message.PROPERTYMAPPING_WHITESPACE_TRIMMED, 
    mapping.getSourceName()
);
```

That would be inline with our code style. However, I can also do this during the merge ;)

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceReference.java
```diff
@@ -108,8 +108,14 @@ public SourceReference build() {
...
+                                       Message.PROPERTYMAPPING_WHITESPACE_TRIMMED, mapping.getSourceName()
```
filiphr: I would pass the `sourceName` (`targetName` in the `TargetReference`). I know that it is the same, but it reads better I think.

filiphr:
Thanks for the PR @smythie86 I have merged it into master

Issue #1365, opened 2018-01-02
Allow package-private mapper


gervaisb:
To limit visibility, we have a package-private mapper interface. While the compilation pass without any problems, the tests reveals that the mapper impl cannot be created:
```
Caused by: java.lang.IllegalAccessException: Class org.mapstruct.factory.Mappers can not access a member of class com.acme.PackagePrivateMapperImpl with modifiers ""
  at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
  at java.lang.Class.newInstance(Class.java:436)
  at org.mapstruct.factory.Mappers.doGetMapper(Mappers.java:99)
```

Having package-private classes (and mappers) help to expose clean interfaces outside a package without exposing the plumbing.

filiphr:
If you are using the `Mappers` factory in your tests it means you are exposing the plumbing. How are you using the mappers in your production code? If you are not using the default `componentModel`, I would suggest that you instantiate the mappers in the same way in your tests, otherwise you are risking to have invalid mappers during your tests.

What you could do is to create your own `Mappers` factory that would allow you to instantiate package private mappers. One would need to get the constructor, make it accessible and then instantiate the class. In theory we can bring this into the Factory, but I am not sure if it is really needed...

I also have a similar setup, where the public interface has no knowledge of MapStruct, but then the mapping definition extends this interface and provides the mappings and some extra information for the MapStruct mapping. However, I am using the spring `componentModel`, for the tests I am using spring so I can get the mappers as well.

gervaisb:
I'm testing the mappers but from a test within the same package. So I test the plumbing but keep it package-private.

Of course I can create my own mapper factory. But that's code that we have to duplicate while we use MapStruct to reduce the duplication and boilerplate. This was just a proposal.

gunnarmorling:
@filiphr We could add something like this to the `Mappers` factory:

```
if ( !isAccessible( mapperImplTypeConstructor ) ) {
    mapperImplTypeConstructor.setAccessible( true );
}
```

The only caveat is that on Java 9 with named modules a module containing the mapper implementation must open (at least that package) towards the MapStruct module.

filiphr:
@gunnarmorling there is already an open PR (#1366) from @gervaisb for this. Do you have time to review it?

[...]

PR #1366, opened 2018-01-04
Allow package-private mapper

gervaisb:
Use and make the default constructor accessible to create the mapper
instance. Solve #1365


core-common/src/main/java/org/mapstruct/factory/Mappers.java
```diff
@@ -93,11 +97,13 @@ private Mappers() {
...
     }
 
-    private static <T> T doGetMapper(Class<T> clazz, ClassLoader classLoader) {
+    private static <T> T doGetMapper(Class<T> clazz, ClassLoader classLoader) throws NoSuchMethodException {
         try {
             @SuppressWarnings("unchecked")
-            T mapper = (T) classLoader.loadClass( clazz.getName() + IMPLEMENTATION_SUFFIX ).newInstance();
-            return mapper;
+            Class<T> implementation = (Class<T>) classLoader.loadClass( clazz.getName() + IMPLEMENTATION_SUFFIX );
+            Constructor<T> constructor = implementation.getDeclaredConstructor((Class<?>[]) null);
```
gunnarmorling: Why is `(Class<?>[]) null` passed? Shouldn't be necessary.

core-common/src/main/java/org/mapstruct/factory/Mappers.java
```diff
@@ -93,11 +97,13 @@ private Mappers() {
...
+            constructor.setAccessible(true);
```
gunnarmorling: I think on Java 9 and beyon we should use `trySetAccessible()` instead. This will allow to raise a meaningful exception in case we can't make it accessible due to lack of `opens`.

core-common/src/main/java/org/mapstruct/factory/Mappers.java
```diff
@@ -93,11 +97,13 @@ private Mappers() {
...
+            return constructor.newInstance((Object[]) null);
```
gunnarmorling: Is the parameter value actually needed?

core-common/src/test/java/org/mapstruct/test/PackagePrivateMapperTest.java
```diff
@@ -0,0 +1,33 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.test;
+
+import org.junit.Test;
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+public class PackagePrivateMapperTest {
```
gunnarmorling: Could you omit the doubled new lines, add a line break at EOF and a `@IssueKey` annotation?

core-common/src/main/java/org/mapstruct/factory/Mappers.java
gervaisb: I did not have a good knowledge of Java9 but I agree.  Do you already move your code to Java9 ?

gunnarmorling: We should _support_ Java 9 as well as possible, but not _require_ it. I.e. I'd suggest to do it like this:

```
if ( isJava9 ) {
     trySetAccessible;
}
else {
    setAccessible;
}

File: core-common/src/test/java/org/mapstruct/test/PackagePrivateMapperTest.java (hunk 4 above)
Comment (gervaisb) [reply core-common/src/test/java/org/mapstruct/test/PackagePrivateMapperTest.java]: I don't know this `@IssueKey` annotation. Can you point me to the doc ?

Comment (gervaisb) [reply core-common/src/test/java/org/mapstruct/test/PackagePrivateMapperTest.java]: The problem is that `@IssueKey` is defined into the _mapstruct-processor_ module which is not accessible from _mapstruct-common_. Should I add a dependency with the _test_ scope ?

File: core-common/src/main/java/org/mapstruct/factory/Mappers.java (hunk 3 above)
Comment (gervaisb) [reply core-common/src/main/java/org/mapstruct/factory/Mappers.java]: No. I will remove it

File: core-common/src/main/java/org/mapstruct/factory/Mappers.java (hunk 1 above)
Comment (gervaisb) [reply core-common/src/main/java/org/mapstruct/factory/Mappers.java]: No. I will remove it

File: core-common/src/test/java/org/mapstruct/test/PackagePrivateMapperTest.java (hunk 4 above)
Comment (gervaisb) [reply core-common/src/test/java/org/mapstruct/test/PackagePrivateMapperTest.java]: I have tried to package the test classes from _mapstruct-processor_ and add a _test_ dependency on it from _mapstruct-common_. However I have the feeling that this is a strange dependency. In my opinion the `@IssueKey` must be moved into _mapstruct-common\src\test_  and used in other modules by adding a dependency to _mapstruct-common_ instead.
What do you think/should I do ?

Comment (gunnarmorling) [reply core-common/src/test/java/org/mapstruct/test/PackagePrivateMapperTest.java]: Hi, agreed, this dependency isn't good. When adding the original comment, I didn't notice that your change is in the _common_ module and thus can't access the `@IssueKey` annotation. I'd suggest to just omit it then, seems not worth to move this. Sorry for the confusion.

**Comment by gervaisb** [comment]
Okay fine. So is my PR clean for you or should I still do some changes ?
Le mer. 10 janv. 2018 à 21:58, Gunnar Morling <notifications@github.com> a
écrit :

**Comment by gervaisb** [comment]
I have pushed the last bits so that everything would be fine
Le mer. 10 janv. 2018 à 22:29, Blaise Gervais <gervais.b@gmail.com> a
écrit :

File: core-common/src/test/java/org/mapstruct/test/PackagePrivateMapperTest.java [hunk 5]
```diff
@@ -0,0 +1,31 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.test;
+
+import org.junit.Test;
+import static org.assertj.core.api.Assertions.assertThat;
```
filiphr: There should be an empty line above this line

core-common/src/test/java/org/mapstruct/test/PackagePrivateMapper.java
```diff
@@ -0,0 +1,27 @@
...
+ *  limitations under the License.
+ */
+package org.mapstruct.test;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+interface PackagePrivateMapper {
+    PackagePrivateMapper INSTANCE = Mappers.getMapper( PackagePrivateMapper.class );
```
filiphr: This is causing checkstyle to fail. You can remove this `INSTANCE` here and it should work. See the `MappersTest`. By the way you don't have to annotate the class with `@Mapper`, the `Mappers` factory does no checks for that

gervaisb:
@filiphr for the Java9 stuff, that would be really nice. 

I solved the latest checkstyle issues and moved all package-private tests in `MappersTest` and his package.

Thanks

gunnarmorling:
Yes, I'll push a commit with the Java 9 adjustments I had in mind.

sjaakd:
@gervaisb , @filiphr : what is the status on this one?

filiphr:
@sjaakd everything is done from the side of @gervaisb. The only thing left is for us to add the Java 9 specifics. @gunnarmorling said that he would be able to add the Java 9 specifics.

filiphr:
Thanks a lot for the PR @gervaisb. We have merged it and it is available on master. Sorry that it took so long.

Issue #1345, opened 2017-11-23
Ignore doesn't works for readonly properties


lujop:
If we map a readonly property(only getter) there's no way to use inverse configuration. 
The problem is that mapstruct give an error that there isn't a write method for the inverse and you can't use @Mapping(ignore) to solve it

sjaakd:
@lujop can you add an example (e.g. as gist)?

lujop:
Of course. Here is the [Gist](https://gist.github.com/lujop/e53d46eeaca432c444ec1c7fa4110827).
When you compile you get this error: 
`a Property "readOnlyProperty" has no write accessor in cat.joanpujol.A.`

And there is no way to get rid of it. Although you use a target ignore you get the error.
And this is very tedious because if you have a big mapping where you can use the InverseConfig and you have only one read-only property you have to manually do all the inverse configuration.

sjaakd:
@lujop : Thanks.. I'm able to reproduce your problem. Now finding a solution.

lujop:
Thanks, I think the same, it's better if Mapstruct detects it's a readonly property.
But a least ignore should work ;)

sjaakd:
@lujop : took a while.. But there's a fix PR #1369 pending. If you want, you can check it out and build it locally.

sjaakd:
fixed.

Issue #838, opened 2016-07-21
MapStruct should check for suitable constructors when reversing


sjaakd:
...and not leave it up to the compiler.

PR #1369, opened 2018-01-11
#1345 ignoring reversed mappings with no target accessor silently

sjaakd:
There seems to be functionality to deliberately do something with the fact that MapStruct cannot perform a reverse mapping when there's no target write accessor.

However there are no testcases for it.

There is however a test case failing when the solution is implemented, see #838. However, this is funny because now the fault message is actually covering #838 in stead of "unknown property". 

So: what should we do: 
1. should we indeed automatically ignore a reverse mapping if there's no write accessor in the reverse situation as is introduced with this PR? I personally think so.
2. if we agree on that, it most likely makes no sense to pass this `isReverse` boolean  to ` errorMessage.report( isReverse );` on line 175 in `TargetReference` and remove subsequent functionality in the message implementations.

@gunnarmorling : WDYT about point 1.?
@filiphr : I saw you put this error messages in (also handling the reverse case as error) together with @navpil . What was the reasoning behind this? Do I miss anything ?

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
@@ -37,7 +37,6 @@ import javax.lang.model.util.Types;
 import org.mapstruct.ap.internal.model.common.FormattingParameters;
 import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.common.TypeFactory;
-import org.mapstruct.ap.internal.prism.CollectionMappingStrategyPrism;
 import org.mapstruct.ap.internal.prism.MappingPrism;
 import org.mapstruct.ap.internal.prism.MappingsPrism;
 import org.mapstruct.ap.internal.util.FormattingMessager;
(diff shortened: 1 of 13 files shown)


sjaakd:
> The reason why isReverse was added was to provide a more detailed error location. However, why a test is missing for something like this I don't know. I think I didn't add a test case, as I was just doing a refactoring and no test broke, so I thought everything is fine.

There was a test case there testing whether there was a suitable constructor on the target side. This has been reinstated with this PR.



> I think that the more appropriate solution would be indeed to add support for explicitly ignoring read only properties with @Mapping(ignore = true)

I'm not sure. I think we should MapStruct as non-verbose as possible. So if you don't need an annotation, you should not have to add it. Reversing is convenience. If there's a mapping method that can be reversed we should do it. If not, we should not enforce the user explicitly to exclude it. Note: its not necessarily an error either. Perhaps you don't control the target bean (you can't edit it because its not yours, or generated).

gunnarmorling:
> should we indeed automatically ignore a reverse mapping if there's no write accessor in the reverse situation as is introduced with this PR? I personally think so.

@sjaakd, to be sure, I don't think the _entire_ `@InheritInverseMapping` should be ignored. While determining the inverse mappings, it should apply those `@Mapping` annotations which can be applied, whereas those `@Mapping` annotations which can't be applied (because there's no write accessor for the inverse case), should be ignored. I believe that's what you meant?

I wouldn't raise a warning in that case, it seems comparable to just a regular mapping method where we just populate those target properties which are writeable.

PR #1376, opened 2018-02-10
#1375 Allow using target references when there is no intermediary read accessor

### LINKED ISSUE #1375: Invalid mapper is created when there is no read accessor for nested target mappings
Example:

```
public class Target {

    private Nested nested;

    public setNested(Nested nested) {
        this.nested = nested;
    }
}

@Mapper
public interface MyMapper {

    @Mapping(target = "nested.value", source = "value")
    Target map(Source source);
}
```

When we have such mapper then we generate an empty mapper and report no errors.

The generated mapper:

```
}
```

filiphr:
Fixes #1375 

We can actually create a valid mapper when an intermediary read accessor is missing. Therefore I have removed that check (which actually was not reporting anything before)

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/TargetReference.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/TargetReference.java
@@ -195,8 +195,7 @@ public class TargetReference {
                 boolean isLast = i == entryNames.length - 1;
                 boolean isNotLast = i < entryNames.length - 1;
                 if ( isWriteAccessorNotValidWhenNotLast( targetWriteAccessor, isNotLast )
-                    || isWriteAccessorNotValidWhenLast( targetWriteAccessor, targetReadAccessor, mapping, isLast )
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/TargetReference.java
```diff
@@ -195,8 +195,7 @@ public TargetReference build() {
                 boolean isLast = i == entryNames.length - 1;
                 boolean isNotLast = i < entryNames.length - 1;
                 if ( isWriteAccessorNotValidWhenNotLast( targetWriteAccessor, isNotLast )
-                    || isWriteAccessorNotValidWhenLast( targetWriteAccessor, targetReadAccessor, mapping, isLast )
-                    || ( isNotLast && targetReadAccessor == null ) ) {
```
sjaakd: hmm. I'm thinking why we (I? :) ) put this here.. Don't have a clear idea anymore..

filiphr:
Thanks for the review @sjaakd. I have merged this

PR #1396, opened 2018-03-12
#1395 Unused mappers should not be used in constructor injection

### LINKED ISSUE #1395: Injection strategy Constructor does not generate compilable code
Given the following mapper interface using Mapstruct 1.3.0-SNAPSHOT

```
@Mapper(config = SpringMappingConfig.class)
public interface MoneyMapper {

	default MonetaryAmount toMonetaryAmount(a.b.c.MonetaryAmount monetaryAmount) {
		return monetaryAmount != null ? monetaryAmount.toJavaMonetaryAmount() : null;
	}
}
```
and config 

```
@MapperConfig(componentModel = "spring", unmappedTargetPolicy = ReportingPolicy.ERROR, uses = {SimpleTypeMapper.class}, injectionStrategy = InjectionStrategy.CONSTRUCTOR)
public interface SpringMappingConfig {
}
```

the  result is:

```
@Component
public class MoneyMapperImpl implements MoneyMapper {

    @Autowired
    public MoneyMapperImpl(SimpleTypeMapper simpleTypeMapper) {
        this.simpleTypeMapper = simpleTypeMapper;
    }
}
```

The constructor parameter `simpleTypeMapper` is correctly assigned to `this.simpleTypeMapper` but there is no local `simpleTypeMapper`. 

*Update*
I have noticed that this feature works if all of the mappers are declared in the same module. In my case I compile a multi module project with Gradle and the `SimpleTypeMapper` is located in a different module. 

I will provide you an example project for that behaviour...

filiphr:
Fixes #1395.

Unused mappers should not be used in constructors of constructor injection

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
@@ -134,10 +134,9 @@ public abstract class AnnotationBasedComponentModelProcessor implements ModelEle
             new ArrayList<AnnotationMapperReference>( mapper.getReferencedMappers().size() );
 
         for ( MapperReference mapperReference : mapper.getReferencedMappers() ) {
-            if ( !mapperReference.isUsed() ) {
-                continue;
+            if ( mapperReference.isUsed() ) {
+                mapperReferencesForConstructor.add( (AnnotationMapperReference) mapperReference );
             }
-            mapperReferencesForConstructor.add( (AnnotationMapperReference) mapperReference );
         }
 
         List<Annotation> mapperReferenceAnnotations = getMapperReferenceAnnotations();
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
@@ -134,6 +134,9 @@ public abstract class AnnotationBasedComponentModelProcessor implements ModelEle
             new ArrayList<AnnotationMapperReference>( mapper.getReferencedMappers().size() );
 
         for ( MapperReference mapperReference : mapper.getReferencedMappers() ) {
+            if ( !mapperReference.isUsed() ) {
+                continue;
+            }
             mapperReferencesForConstructor.add( (AnnotationMapperReference) mapperReference );
         }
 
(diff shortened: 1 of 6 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_1395/NotUsedService.java
```diff
@@ -0,0 +1,25 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.bugs._1395;
+
+/**
+ * @author Filip Hrisafov
+ */
+public interface NotUsedService {
```
sjaakd: Shouldn't there be an `@Mapper`?

processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java
```diff
@@ -134,6 +134,9 @@ private AnnotatedConstructor buildAnnotatedConstructorForMapper(Mapper mapper) {
             new ArrayList<AnnotationMapperReference>( mapper.getReferencedMappers().size() );
 
         for ( MapperReference mapperReference : mapper.getReferencedMappers() ) {
+            if ( !mapperReference.isUsed() ) {
```
sjaakd:
```java
if (mapperReference.isUsed() ) {
   mapperReferencesForConstructor.add( (AnnotationMapperReference) mapperReference )
}
```

processor/src/test/java/org/mapstruct/ap/test/bugs/_1395/NotUsedService.java
filiphr: Not really. One doesn't have to use mappers in `uses`. You can use any class / service in there

filiphr:
@sjaakd thanks for the review. The changes have been done

processor/src/test/java/org/mapstruct/ap/test/bugs/_1395/NotUsedService.java
sjaakd: I know.. But the behaviour is different. For the 'normal' component mapper, MapStruct initialises the instance variable with `Mappers.getMapper(... )` instead of `new ...`. Filtering is carried out if the mapper is indeed used. When not, there's not even going to be a field. 

So, to make a long story short it could be that the code above is not even triggered because MapStruct already decided there's no need. Or perhaps.. I'm just seeing stuff.  :smile:

filiphr: > But the behaviour is different. For the 'normal' component mapper, MapStruct initialises the instance variable with Mappers.getMapper(... ) instead of new .... 

Not when an injection `componentModel` is used (`spring`, `jsr330`, `cdi`). When one of the non defaults is used, then the uses mappers or services are always injected.

>So, to make a long story short it could be that the code above is not even triggered because MapStruct already decided there's no need. Or perhaps.. I'm just seeing stuff.

That is the entire idea of the PR. To show that the `NotUsedService` should not be injected in the mapper 😄.

PR #1397, opened 2018-03-13
#1367 aligning @InheritConfiguration and @InheritReverseConfiguration with AUTO_INHERIT_ALL_FROM_CONFIG.

### LINKED ISSUE #1367: Shared configuration not applied when using InheritInverseConfiguration
I am trying to make both features "Shared Configuration" and "Inverse Mapping" work together, without success. ( Version used:  1.2.0.Final )
I use MapStruct in my project in order to handle mapping between domain entities layer (JPA entities) and REST resources layer (simple DTOs used by REST/JSON API).
- shared configuration:
Both entities and dtos layers have a root base class that handle common properties, so I have implemented a MapperConfig to handle mapping rules for these fields : 
```
@MapperConfig(unmappedTargetPolicy = ReportingPolicy.WARN,
        mappingInheritanceStrategy = MappingInheritanceStrategy.AUTO_INHERIT_ALL_FROM_CONFIG)
public interface EntityToDtoMappingConfig {
    
    @Mapping(target = "dbId", source = "id")
    @Mapping(target = "links", ignore = true)
    BaseDto entityToDto(final BaseEntity entity);
    
    @Mapping(target = "id", source = "dbId")
    @Mapping(target = "createdBy", ignore = true)
    @Mapping(target = "creationDate", ignore = true)
    @Mapping(target = "lastModifiedBy", constant = "restApiUser") // force modifiedBy with restApiUser constant
    @Mapping(target = "lastModifiedDate", ignore = true)
    public BaseEntity dtoToEntity(BaseDto dto);
    
}
```
- Inverse Mapping
I use @InheritInverseConfiguration in order to simplify "mapFrom" methods definition: generally all mapping rules are defined on the "mapTo" methods, so "mapFrom" methods can be defined simply with "inverse mapping" :
```
@Mapper(config = EntityToDtoMappingConfig.class)
public abstract class CarMapper {

    @Mapping(source = "manufacturer", target = "maker")
    @Mapping(source = "numberOfSeats", target = "seatCount")
    abstract public CarDto mapTo(Car car);
    
    @InheritInverseConfiguration(name = "mapTo")
    // inheritInvers configuration should map both maker and seatCount properties.
    // additional mapping should also be inherited from   EntityToDtoMappingConfig.dtoToEntity method, BUT IT DOES NOT SEEM TO WORK
    abstract Car mapFrom(CarDto carDto);
    
}
```

Mapping from Entity to DTO is working fine: inheritance is handled correctly and mapping rules from EntityToDtoMappingConfig  are appended to mapping rules defined on CarMapper.mapTo method.

But mapping from DTO to Entity is not working as expected :  mapping rules defined in the EntityToDtoMappingConfig  are not applied.

Question : is there a limitation when using InheritInverseConfiguration that prevent us using shared configuration mechanisme?  or is this a bug ?

I have shared a concrete and simple project on GitHub to demonstrate this issue : see  https://github.com/mricciuti/sandbox-mapstruct

Thanks by advance.

sjaakd:
Fixes #1367 

Whereas its not possible to combine forward and reverse explicit configurations (you get a warning:  `Method cannot be annotated with both a @InheritConfiguration and @InheritInverseConfiguration`) it seems to be possible with the current auto-inheritance options. Here you have the possibilities: `AUTO_INHERIT_FROM_CONFIG`, `AUTO_INHERIT_REVERSE_FROM_CONFIG`, and **`AUTO_INHERIT_ALL_FROM_CONFIG`** .

This PR removes that restriction and aligns the strategy for auto inheritance. So it will be possible to define an `@InheritConfiguration` together with `@InheritReverseConfiguration` on  the same method.

Forward mappings (`@InheritConfiguration`) takes precedence over reverse mappings (`@InheritInverseConfiguration`)

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/inheritfromconfig/multiple/CarMapperTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/inheritfromconfig/multiple/CarMapperTest.java
@@ -1,0 +1,101 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.inheritfromconfig.multiple;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Date;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+@RunWith(AnnotationProcessorTestRunner.class)
+@WithClasses({
+    BaseDto.class,
+    BaseEntity.class,
+    CarDto.class,
+    CarEntity.class,
+    CarMapper.class,
+    EntityToDtoMappingConfig.class,
+    Car2Dto.class,
+    Car2Entity.class,
+    CarMapper2.class
+})
+@IssueKey("1367")
+public class CarMapperTest {
+
+    @Test
+    public void testMapEntityToDto() {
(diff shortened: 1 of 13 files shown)


filiphr:
@sjaakd I'll review this ASAP

processor/src/test/java/org/mapstruct/ap/test/inheritfromconfig/multiple/CarMapperTest.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Test
+    public void testMapDtoToEntity() {
+        CarEntity car = CarMapper.MAPPER.mapFrom( newCarDto() );
+        assertThat( car.getId() ).isEqualTo( 9L );
+        assertThat( car.getManufacturer() ).isEqualTo( "Nissan" );
+        assertThat( car.getNumberOfSeats() ).isEqualTo( 5 );
+
+        // following assertion fails :  createdBy property is not set.
```
filiphr: Do we still need this line and the one below?

sjaakd: I'll remove that one.. You're right..

sjaakd:
@filiphr .. thanks for the review. I'll fix the issue an wait a couple of days for the response of @mricciuti. I'll proceed to merge on sunday evening or monday.

sjaakd:
squashed

Issue #1392, opened 2018-03-08
Ignore mappings by default


mowcixo:
As **Augusto Breno** described at Google Groups (https://groups.google.com/forum/?fromgroups#!searchin/mapstruct-users/ignore%7Csort:date/mapstruct-users/BDaK45b3rAg/V1amaHESBwAJ) it would be great to have the possibility of ignore mappings by default.

And yes, as @agudian said at the same topic, it can result a little odd based on the main purpose of MapStruct, but it can be very useful in cases like I'll try to describe:

In some cases you have a class that has copied values from another for historical purposes, for example, an invoice and a customer. And at the same time, both classes have some fields that have the same name.

```
class Invoice {

    private Long id;

    private String code;

    // Customer historical fields

    private String customerNoun;

    private String customerSurname;

    // getters and setters ....

}

class Customer {

   private Long id;

   private String code;

   private String noun;

   private String surname;

    // getters and setters ....

}
```

Then, you may have following mapping definition to copy values from Customer to Invoice:

```
// ...
    @Mappings({
            @Mapping(source = "noun", target = "customerNoun"),
            @Mapping(source = "surname", target = "customerSurname"),
            @Mapping(target = "id", ignore = true),
            @Mapping(target = "code", ignore = true)
    })
    public abstract void fillInvoice(Customer customer, @MappingTarget Invoice invoice);
// ...
```

If you have just two properties to ignore, it's not a big deal, but in more complex structs, it can be a large number of properties that match it's name, so it would be great to have something as **Augusto** described in the ```@Mapper``` annotation, or even in the ```@Mappings``` annotation such as:

```
// ...
    @Mapper(ignoreByDefault=true)
// ...
```
or...

```
// ...
    @Mappings(ignoreByDefault=true, {
            @Mapping(source = "noun", target = "customerNoun"),
// ...
```

Regards!

sjaakd:
@mowcixo if you want something like this, I would prefer the `@BeanMapping` iso `@Mappings`. In java8 `@Mappings` is marked by `@Mapping` as `@Repeatable`. `@BeanMapping` is there as holder specific for bean mapping methods.

[...]

PR #1403, opened 2018-03-19
#1392 add option to default ignoreAll mappings in a bean mapping method

sjaakd:
The user can now control the mapping on `@BeanMapping` level. Especially when there only a few source properties and the target object has a large number of properties this comes in handy.

In stead of manually entering all the `@Mapping( target = "?", ignore=true ), this is now the default behaviour for this one method.

Note that this is different than disabling warnings (unmapped target policy). This leaves still implicit mappings to take care of.

See #1392 for more details.

### DIFF
--- a/core-common/src/main/java/org/mapstruct/BeanMapping.java
+++ b/core-common/src/main/java/org/mapstruct/BeanMapping.java
@@ -73,4 +73,14 @@ public @interface BeanMapping {
      * @return The strategy to be applied when {@code null} is passed as source value to the methods of this mapping.
      */
     NullValueMappingStrategy nullValueMappingStrategy() default NullValueMappingStrategy.RETURN_NULL;
+
+    /**
+     * Default ignore all mappings. All mappings have to be defined manually. No automatic mapping will take place. No
+     * warning will be issued on missing target properties.
+     *
+     * @return The ignore strategy (default false).
+     *
+     * @since 1.3
+     */
+    boolean ignoreByDefault() default false;
 }
--- a/processor/src/test/java/org/mapstruct/ap/test/ignore/expand/IgnorePropertyTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/ignore/expand/IgnorePropertyTest.java
@@ -1,0 +1,64 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.ignore.expand;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+/**
+ * Test for ignoring properties during the mapping.
+ *
+ * @author Sjaak Derksen
+ */
+@WithClasses({
+    ExpandedToolbox.class,
+    Hammer.class,
+    Wrench.class,
+    FlattenedToolBox.class,
+    ToolBoxMapper.class
+})
+@RunWith(AnnotationProcessorTestRunner.class)
+public class IgnorePropertyTest {
+
+    @Test
+    @IssueKey("1392")
+    public void shouldIgnoreAll() {
+        FlattenedToolBox toolboxSource = new FlattenedToolBox();
+        toolboxSource.setBrand( "Stanley" );
+        toolboxSource.setHammerDescription( "heavy" );
+        toolboxSource.setHammerSize( 5 );
+        toolboxSource.setWrenchIsBahco( Boolean.TRUE );
+        toolboxSource.setWrenchDescription( "generic use" );
+
+        ExpandedToolbox toolboxTarget = ToolBoxMapper.INSTANCE.expand( toolboxSource );
+
+        assertThat( toolboxTarget ).isNotNull();
+        assertThat( toolboxTarget.getBrand() ).isNull();
+        assertThat( toolboxTarget.getHammer() ).isNotNull();
+        assertThat( toolboxTarget.getHammer().getDescription() ).isEqualTo( "heavy" );
+        assertThat( toolboxTarget.getHammer().getSize() ).isNull();
+        assertThat( toolboxTarget.getWrench() ).isNull();
+    }
+
+}
--- a/processor/src/test/java/org/mapstruct/ap/test/ignore/inherit/IgnorePropertyTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/ignore/inherit/IgnorePropertyTest.java
@@ -1,0 +1,90 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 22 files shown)


sjaakd:
@gunnarmorling .. this is an api addition. Its good to have your opinion on that one as well.

sjaakd:
I'm not sure.. can you can inherit from ignored mappings. In my implementation all the mappings are ignored template mapping, but not in the overriding method (if you desire). So, `restrictToDefinedMappings`, I presume something different. Perhaps align in 2.0?

filiphr:
I am not talking about inheritance I am talking about the new API. We already have support for our internal nested targets to only perform mappings on defined `@Mapping`. Exactly what you are solving here. Right now you are adding "dummy" `@Mapping(target = "brand", ignore = true)`.

What I am proposing is to just reuse what we have already there. And move the `restrictToDefinedMappings` from the `MappingOptions` in the `BeanMapping`. We can then expose this field in `@BeanMapping` and your tests would pass as the mapping would be only for the defined mappings.

sjaakd:
Have you checked out `processor/src/test/java/org/mapstruct/ap/test/ignore/inherit/ToolMapper.java`? 

You can inherit a (reverse)configuration from a prototype method. If the prototype method has all properties ignored, and the mapping method not, you'll still need to implement all the properties of the parent object in the mapping method. As opposed to reusing the structure of 'restrictToDefinedMappings' where its all or nothing.

So, fine grained control + inheritance  is probably what I would expect looking at the use cases of entities. I would like to have the possibility to say: all properties in the base-entity are not relevant, however in the entity itself -'its business'-, and I need to map it.

sjaakd:
`restrictToDefinedMappings` seems disable name based mapping (see `BeanMapping`).
```java
            if ( !method.getMappingOptions().isRestrictToDefinedMappings() ) {

                // map properties without a mapping
                applyPropertyNameBasedMapping();

                // map parameters without a mapping
                applyParameterNameBasedMapping();
            }
```
However it does not disable error handling for unmapped targets. That probably means that you have to specify all mappings by means of the `@Mapping` annotation. Hence, it feels `rather` the opposite of what we want, which is consider all mappings as ignored .

filiphr:
That is true. However, we internally we have reporting policy on the bean mapping (see `BeanMapping#forForgedMethods`. We can set the reporting policy to ignore when the user sets `restrictToDefinedMappings`

filiphr:
This ignore policy would be only for this particular method

sjaakd:
> That is true. However, we internally we have reporting policy on the bean mapping (see BeanMapping#forForgedMethods. We can set the reporting policy to ignore when the user sets restrictToDefinedMappings

That feels a bit of a hack :smile: : your solution solves stuff in two places. And.. the there's more to report than only missing target properties, right..? 

The current proposed solution gives you a more fine-grained control. It offers the possibility to inherit all properties from `@InheritConfiguration` and let those be ignored whilst name-based-mapping / defining the non ignored. For this scenario I'll add a meaningful use case, one I from my own context encountered as well.

filiphr:
>And.. the there's more to report than only missing target properties, right..?

Well not really the name of the set property is `unmappedTargetPolicy` so reporting unmapped target properties :).

I'll wait for the scenario, thanks

sjaakd:
@filiphr scenario has been added as we discussed. That shows how inheritance looks like.

Lets discuss the user defined mapping from the constant (literal) to the target type. 
1. The proposed solution asserts whether a string (between "") can be mapped to the target type if the target type is primitive or boxed. 
1a. If this is the case, it considered the source type as if it were the same type as the target type.
1b. If this is not the case it assumes the constant is a `String` and should be treated like that.
1c. Matching continues based on these assumptions.

Advantages:
1. Optimised code generation
2. Actually more formats are supported (as for instance the `Integer.getValue` or `Integer.parseValue` do not support hex/octal/binary numbers.
3. TypeConversion (automatic) is not applied.

Disadvantages:
1. The user might have the case were a constant matches the format of the target type. If he/she defined a mapping method that took `String` and delivered `TargetType` this would not be called anymore. Note: there also might have been undesired 2 step conversions.

Now there are some improvements possible. 
1. We could 'mark' the 'asumed' source-type as literal and treat it different when we try to discover a mapping method. In particular: we might want to disable type-conversion and not allow 2 step mapping methods.
2. We could then also give a decent error (triggered by the combo 'mark' AND 'no mapping method found') that the provided constant does not match the target type.

Last but not least: I think that constants should be consistent with the way the compiler handles literal assignments. We should not have to rely on expressions for doing this. I personally was puzzled by compilation errors occurring after parsing due to the use of expressions. You simply do not expect that. Hence, were-ever possible we should provide the user with the means to handle things via a defined interface and not via the **emergency break mechanism: expression**).

----
A final thought on this: perhaps we should handle string literals as well in a clear way: `"\"test\""` for 2.0

filiphr:
@sjaakd can we move the discussion about the constant (literal) to target type in #1409 instead of this one?

I will have a look at your commit about the clarification for this one

filiphr:
@sjaakd Btw your last commit messages references a wrong issue

sjaakd:
> @sjaakd can we move the discussion about the constant (literal) to target type in #1409 instead of this one?

of course.. my mistake.. I'm a bit feverish - not thinking to clear - having a severe cold. :(..

sjaakd:
> @sjaakd Btw your last commit messages references a wrong issue

fixed..

filiphr:
Thanks for the fixes.

>Of course.. my mistake.. I'm a bit feverish - not thinking to clear - having a severe cold. :(..

Hope you get well fast and soon :)

processor/src/test/java/org/mapstruct/ap/test/ignore/expand/IgnorePropertyTest.java
```diff
@@ -0,0 +1,64 @@
...
+    Wrench.class,
+    FlattenedToolBox.class,
+    ToolBoxMapper.class
+})
+@RunWith(AnnotationProcessorTestRunner.class)
+public class IgnorePropertyTest {
+
+    @Test
+    @IssueKey("1392")
+    public void shouldIgnoreAll() {
```
filiphr: What exactly is this test showing? 

That there are no warnings for the "unmapped" ignored properties?

processor/src/test/java/org/mapstruct/ap/test/ignore/inherit/IgnorePropertyTest.java
```diff
@@ -0,0 +1,80 @@
...
+        assertThat( toolTarget.getSize() ).isNull();
+        assertThat( toolTarget.getType() ).isEqualTo( "smash" );
+
+    }
+
+    @Test
+    @IssueKey("1392")
+    public void shouldIgnoreBase() {
+
+        WorkBenchDto workBenchDto = new WorkBenchDto();
```
filiphr: The properties from the base are `null` even before the mapping. To show what you want I presume they need to be non `null` before the mapping?

processor/src/test/java/org/mapstruct/ap/test/ignore/inherit/IgnorePropertyTest.java
```diff
@@ -0,0 +1,80 @@
...
+    WorkBenchEntity.class,
+    ToolMapper.class
+})
+@RunWith(AnnotationProcessorTestRunner.class)
+public class IgnorePropertyTest {
+
+    @Test
+    @IssueKey("1392")
+    public void shouldIgnoreAll() {
+        HammerDto hammer = new HammerDto();
```
filiphr: Same here, I think we need to fill in all the properties

processor/src/test/java/org/mapstruct/ap/test/ignore/inherit/IgnorePropertyTest.java
sjaakd: hould not issue warnings on unmapped target properties. There are no more source props.

processor/src/test/java/org/mapstruct/ap/test/ignore/inherit/IgnorePropertyTest.java
sjaakd: should not issue warnings on unmapped target properties

processor/src/test/java/org/mapstruct/ap/test/ignore/inherit/IgnorePropertyTest.java
sjaakd: ok

processor/src/test/java/org/mapstruct/ap/test/ignore/expand/IgnorePropertyTest.java
sjaakd: yes.. Changed the title.

core-common/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -73,4 +73,12 @@
...
      */
     NullValueMappingStrategy nullValueMappingStrategy() default NullValueMappingStrategy.RETURN_NULL;
+
+    /**
+     * Default ignore all mappings. All mappings have to be defined manually. No automatic mapping will take place. No
+     * warning will be issued on missing target properties.
+     *
+     * @return The ignore strategy (default false).
+     */
+    boolean defaultIgnore() default false;
```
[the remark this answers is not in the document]
sjaakd: done

sjaakd:
@filiphr .. did the changes.. have a look.

core-common/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -73,4 +73,12 @@
...
      */
     NullValueMappingStrategy nullValueMappingStrategy() default NullValueMappingStrategy.RETURN_NULL;
+
+    /**
+     * Default ignore all mappings. All mappings have to be defined manually. No automatic mapping will take place. No
+     * warning will be issued on missing target properties.
+     *
+     * @return The ignore strategy (default false).
+     */
+    boolean ignoreByDefault() default false;
```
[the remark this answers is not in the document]
sjaakd: ok

sjaakd:
@filiphr .. thanks for the review

Issue #1370, opened 2018-01-17
Mapping String to org.joda.time.LocalDate issues compiler error


MichaelRisthausAtos:
Hi,

lets say I have a source class containing a String for a credit card expiration date (in the form of MM/yyyy) and a target class containing a joda LocalDate for the same.

So I write a mapper with a mapping:

`@Mappings({
             @Mapping(source = "creditCardExpirationDate", target = "creditCardExpirationDate", dateFormat = "MM/yyyy")
     })
     public abstract Target mapSourceToTargetWithDateTime(Source source);`

This will fail during compilation:

`Given date format "MM/yyyy" is invalid. Message: "org.joda.time.format.DateTimeFormat".`

The fancy thing is: in another project with exactly the same joda and mapstruct versions and exactly the same source (String) and target (LocalDate), Mapstruct generates the mapper just fine:

`if ( source.getCreditCardExpirationDate() != null ) {
                target.setCreditCardExpirationDate( DateTimeFormat.forPattern( "MM/yyyy" ).parseLocalDate( source.getCreditCardExpirationDate() ) );
            }`

What am I missing here? :(

The used versions are 1.2.0.Final for Mapstruct and 2.9.9 for joda time.

I created a little repository with wich you should be able to reproduce the problem:
[https://github.com/lazydays79/mapstructissue](https://github.com/lazydays79/mapstructissue)

Best regards,
Michael

MichaelRisthausAtos:
After having a deeper look into the code of mapstruct itself, I found out, that this error will occur if mapstruct doesn't find joda time (or esp. the DateTimeFormat class) in its classpath.

If I integrate mapstruct using the "old" way by attaching it to the generate-sources phase, everything works fine.

You can find it in the [generate-sources](https://github.com/lazydays79/mapstructissue/tree/generate-sources) branch in the repository mentioned above.

You can find the difference [here](https://github.com/lazydays79/mapstructissue/compare/generate-sources).

So, the issue is: With the integration of mapstruct in the pom as given in the example in the documentation, mapstruct can't find joda in its classpath (or maybe even other dependencies?).

Regards,
Michael

MichaelRisthausAtos:
Adding joda to the annotationProcessorPaths seems to solve the problem:

`<annotationProcessorPaths>
                        <path>
                            <groupId>joda-time</groupId>
                            <artifactId>joda-time</artifactId>
                            <version>${joda-time.version}</version>
                        </path>
                        <path>
[...]

PR #1407, opened 2018-03-30
#1370 avoid errors when Joda is not on classpath

sjaakd:


filiphr:
Maybe adding a more descriptive message and telling the user what needs to be done would be better. Something like:

Cannot validate Joda dateformat, no Joda on classpath. Consider adding Joda to the annotation processorpath (annotationProcessorPaths in Maven, apt in Gradle)

Something along those lines so users know how to fix it. Maybe @MichaelRisthausAtos has a better idea for the message. What would have helped you in this case?

sjaakd:
Hmm. I noticed that `NOTE` is not printed at all currently.. Perhaps it should be `WARN`.. But that felt to heavy... I'm not sure whether MapStruct should do this validation..

filiphr:
I think adding on `WARN` level is OK for us. We can't do anything about this, but we can tell the user what he / she needs to do to get this working correctly and we will assume that the date format is valid

sjaakd:
Hmm. perhaps you knew, but to show warnings during maven compilation you have to add `showWarnings`..  Took me a while to figure that out..

```
           <plugin>
                <version>3.6.2</version>
                <configuration>
                    <source>1.8</source>
                    <target>1.8</target>
                    <showWarnings>true</showWarnings>                 
                    <annotationProcessorPaths>
                        <path>
                            <version>${org.mapstruct.version}</version>
                        </path>
                    </annotationProcessorPaths>
                </configuration>
            </plugin>
```

filiphr:
You can also configure the compiler to fail on warnings with `<failOnWarning>true</failOnWarning>` see [this](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#failOnWarning)

Issue #1401, opened 2018-03-19
Possible performance improvement of constant/defaultValue primitive to String mappings


twentylemon:
When a constant primitive to string mapping is specified (such as `Mapping(constant = "1")` or `@Mapping(defaultValue = "1")`), the code generated is:

```java
bean.setIntAttribute(Integer.parseInt("1"));
```
But it is possible to instead write this as
```java
bean.setIntAttribute(1);
```
Since `Integer.parseInt("1")` is known at compile time. A similar improvement can be made for the reverse mapping, which currently uses `String.valueOf(<primitive>)` at runtime.

sjaakd:
MapStruct only knows its a String, since that's what the `defaultValue` in essence is. It does not do any analysis on the value. Of course this could be done (interpreting that the string is only comprised of digits, or digits with a designator (e.g. L for long). I think the code is more or less the same as for constants.  Be careful though, we should think what we do if the constant is too large for its target type. Follow the same rules als JDK? 

I'm not sure when one of us has time to implement this. But feel free to contribute.

filiphr:
@twentylemon This can be achieved (partially) with the upcoming 1.3 release and the new `defaultValueExpression` added in #1372. 

You would need to write `@Mapping(defaultValueExpression = "java(1)")`. Is this sufficient for you?

If a `constantExpression` is also needed can we rename this issue for that, or close this and open a new one?

In any case as @sjaakd mentioned we won't have time to implement this. However, if someone wants to give it a shot we can help out.

twentylemon:
> If a `constantExpression` is also needed can we rename this issue for that, or close this and open a new one?

Does `defaultValue` work for primitives? I thought it required a `source`, which for a primitive can never be null. A `constantExpression` may be required then.
Let's assume I'm wrong though, and `defaultValue` works without a `source`.

> You would need to write `@Mapping(defaultValueExpression = "java(1)")`. Is this sufficient for you?

From a consumer standpoint, *I* would be ok with it.

In general though, I find it odd that there would be two ways to achieve the exact same mapping from an interface standpoint.

```java
@Mapper interface MyMapper {
    @Mapping(target = "one", defaultValue = "1")
    Model fromDto(Dto dto);

    @Mapping(target = "one", defaultValueExpression = "java(1)")
    Model fromDtoButUseThisOneAsItIsMarginallyFaster(Dto dto);
}
```

[...]

PR #1409, opened 2018-04-01
#1401 performance improvements when assigning constants

sjaakd:
I think I got all checks covered. The examples from the JLS work. Minimum, maximum works. Added a bunch of tests. So, apart from the discussion on whether there are problems to be foreseen that we cannot handle in the release notes, this should be ready to go.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -23,6 +23,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
+import javax.lang.model.element.AnnotationMirror;
 
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.type.DeclaredType;
@@ -116,6 +117,7 @@ public class PropertyMapping extends ModelElement {
 
         protected List<String> dependsOn;
         protected Set<String> existingVariableNames;
+        protected AnnotationMirror mirror;
 
         MappingBuilderBase(Class<T> selfType) {
             super( selfType );
@@ -143,6 +145,11 @@ public class PropertyMapping extends ModelElement {
             this.targetWriteAccessorType = TargetWriteAccessorType.of( targetWriteAccessor );
             this.targetType = determineTargetType();
 
+            return (T) this;
+        }
+
+        T mirror(AnnotationMirror mirror) {
+            this.mirror = mirror;
             return (T) this;
         }
 
@@ -359,7 +366,7 @@ public class PropertyMapping extends ModelElement {
                 &&  ( !rhs.getSourceType().isPrimitive() || rhs.getSourcePresenceCheckerReference() != null) ) {
                 // cannot check on null source if source is primitive unless it has a presence checker
                 PropertyMapping build = new ConstantMappingBuilder()
-                    .constantExpression( '"' + defaultValue + '"' )
+                    .constantExpression( defaultValue )
                     .formattingParameters( formattingParameters )
                     .selectionParameters( selectionParameters )
                     .dependsOn( dependsOn )
@@ -749,7 +756,12 @@ public class PropertyMapping extends ModelElement {
         public PropertyMapping build() {
             // source
             String sourceErrorMessagePart = "constant '" + constantExpression + "'";
-            Type sourceType = ctx.getTypeFactory().getType( String.class );
+
+            Type sourceType = ctx.getTypeFactory().getTypeForConstant( targetType, constantExpression );
+            if ( String.class.getCanonicalName().equals( sourceType.getFullyQualifiedName() ) ) {
+                // convert to string
+                constantExpression = "\"" + constantExpression + "\"";
+            }
 
             Assignment assignment = null;
             if ( !targetType.isEnumType() ) {
@@ -808,6 +820,7 @@ public class PropertyMapping extends ModelElement {
             else {
                 ctx.getMessager().printMessage(
                     method.getExecutable(),
+                    mirror,
                     Message.CONSTANTMAPPING_MAPPING_NOT_FOUND,
                     sourceType,
                     constantExpression,
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
@@ -253,6 +253,18 @@ public class MappingResolverImpl implements MappingResolver {
                 return simpleAssignment;
             }
 
+            // At this point the SourceType will either
+            // 1. be a String
+            // 2. or when its a primitive / wrapped type and analysis successful equal to its TargetType. But in that
+            //    case it should have been direct assignable.
+            // In case of 1. and the target type is still a wrapped or primitive type we must assume that the check
+            // in NativeType is not successful. We don't want to go through type conversion, double mappings etc.
+            // with something that we already know to be wrong.
+            if ( sourceType.hasOriginatedFromConstant() && ( targetType.isPrimitive() || targetType.isBoxed() ) ) {
+                // TODO: convey some error message
+                return null;
+            }
+
             // then type conversion
             Assignment conversion = resolveViaConversion( sourceType, targetType );
             if ( conversion != null ) {
--- a/processor/src/test/java/org/mapstruct/ap/test/source/constants/ConstantOptimizingTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/source/constants/ConstantOptimizingTest.java
@@ -1,0 +1,342 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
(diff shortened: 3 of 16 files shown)


processor/src/main/java/org/mapstruct/ap/internal/util/PrimitiveUtils.java
```diff
@@ -0,0 +1,236 @@
...
+            // octal
+            return new BaseRepresentation( 8, withoutSign.substring( 1 ), isNegative );
+        }
+        else {
+            // decimal
+            return new BaseRepresentation( 10, withoutSign, isNegative );
+        }
+    }
+
+    private static class BaseRepresentation {
```
sjaakd: I shall put the class-def on top of the file according jls convention

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -749,7 +750,16 @@ public ConstantMappingBuilder selectionParameters(SelectionParameters selectionP
         public PropertyMapping build() {
             // source
             String sourceErrorMessagePart = "constant '" + constantExpression + "'";
-            Type sourceType = ctx.getTypeFactory().getType( String.class );
+
+            Type sourceType;
+            if ( !isConstantDirectAssignable( targetType, constantExpression ) ) {
+                // convert to string
+                sourceType = ctx.getTypeFactory().getType( String.class );
```
sjaakd: knowing what I know now.. If the target is a primitive or boxed type and we use the default conversion fall-back, (e.g. Integer.toValue) that we used to, you will get runtime exceptions.. But it could be that there's a user defined mapping method that takes this into consideration.. We simply don't know.

sjaakd: I'll do another update and use patternmatchig for hex,oct,bin .. that's far more easy and we can change when going to java8 only.

filiphr: Won't this break backwards compatibility? It might be directly assignable. However, up to now we would have called a user method for converting `String` into `int`. However, now we would directly assign it.

sjaakd: don't think it breaks anything.. Suppose you would have a hex/oct/bin pattern, and try to run it through the parse function it fails runtime. That's what I found out. So.. you first need to remove all `0b` `0x` `0b` prefixes, underscores and `L,l,F,f,D,d` suffixes before you can call parse

sjaakd: in short: next to having more efficient code, we have more possibilities (e.g use of bit/octal/byte-mask-patterns).

sjaakd:
> @sjaakd I meant in more general. If a user has defined a custom method that converts from String to Integer. Are we going to consider this method or assign this directly?

Good point... In the current solution probably not. We change the source type to the target type if its directly assignable. Selection takes place in the property mapping. I think that:
1. We should discuss if this is acceptable for 1.3. IMHO it probably is, since the proposed solution is more what you would expect.
2. If not, take precautions. But then we might need to consider 2-step mapping as well (also with mapping and the dreaded type conversion).
3. delay to 2.0?

I'll also think about the consequences a bit more.

sjaakd:
@filiphr scenario has been added as we discussed. That shows how inheritance looks like.

Lets discuss the user defined mapping from the constant (literal) to the target type. 
1. The proposed solution asserts whether a string (between "") can be mapped to the target type if the target type is primitive or boxed. 
1a. If this is the case, it considered the source type as if it were the same type as the target type.
1b. If this is not the case it assumes the constant is a `String` and should be treated like that.
1c. Matching continues based on these assumptions.

Advantages:
1. Optimised code generation
2. Actually more formats are supported (as for instance the `Integer.getValue` or `Integer.parseValue` do not support hex/octal/binary numbers.
3. TypeConversion (automatic) is not applied.

Disadvantages:
1. The user might have the case were a constant matches the format of the target type. If he/she defined a mapping method that took `String` and delivered `TargetType` this would not be called anymore. Note: there also might have been undesired 2 step conversions.

Now there are some improvements possible. 
1. We could 'mark' the 'asumed' source-type as literal and treat it different when we try to discover a mapping method. In particular: we might want to disable type-conversion and not allow 2 step mapping methods.
2. We could then also give a decent error (triggered by the combo 'mark' AND 'no mapping method found') that the provided constant does not match the target type.

Last but not least: I think that constants should be consistent with the way the compiler handles literal assignments. We should not have to rely on expressions for doing this. I personally was puzzled by compilation errors occurring after parsing due to the use of expressions. You simply do not expect that. Hence, were-ever possible we should provide the user with the means to handle things via a defined interface and not via the **emergency break mechanism: expression**).

----
A final thought on this: perhaps we should handle string literals as well in a clear way: `"\"test\""` for 2.0

filiphr:
@sjaakd this seems quite complex to me. I am still not sure that this is the way that we should be doing this. It really looks like we are trying to emulate what the compiler would do.

Adding the `hasOriginatedFromConstant` to the `Type` doesn't really look right to me.

I really think that the exact same thing can be done by the use of `defaultValueExpression` and if we add `constantExpression`. The users will write the exact stuff that needs to go in the generated code. And the compiler will then compile and give a proper compilation error if something is not OK.

I disagree that the `defaultValueExpression` is an _emergency mechanism_. It provides a more powerful way to set a default value. However, I agree that the `expression` is an _emergency mechanism_.

The advantages from 2 to 5 can be done by the using of `defaultValueExpression`. And the last 6th is there from the compiler itself. Which I would argue is more reliable than ours, as it is the compiler itself.

sjaakd:
@filiphr . Lets discuss this. Tuesday. Some background:  The `hasOriginatedFromConstant` belongs ihmo to `Type` because it tells something about how its created.

The manipulation you see is mostly striping of `_` and detection of type `0x, 0b, 0`.. The actual work is delegated to parse. I used material form JLS to test extensively for the unit test.

What bothers me in particular about the current solution is that you can get runtime exceptions. So, detection not during code-generation, not during compilation but even later. 


Wrt to complexity.

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
```diff
@@ -253,6 +253,14 @@ private Assignment getTargetAssignment(Type sourceType, Type targetType) {
                 return simpleAssignment;
             }
 
+            // For constants as source, stop here because there is no direct match between provided literal and target
+            // type and no mapping method has been provided by the user either.
+            // TypeConversion will not give any sensible result for primitive or boxed type because all relevant checks
+            // have been carried out in {@link TypeFactory#getTypeForConstant)
+            if ( sourceType.hasOriginatedFromConstant() && ( targetType.isPrimitive() || targetType.isBoxed() ) ) {
```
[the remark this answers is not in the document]
sjaakd: Good question. At hind sight I wonder whether we need this. At this point the`SourceType` will either
1. be a `String`,
2. or when its a primitive / wrapped type and analysis succesful  equal to its `TargetType`.
However, in the case of 2. (actually the check we carry out) the direct assignment would have been successful, so this possibly renders this check obsolete..

processor/src/test/java/org/mapstruct/ap/test/source/constants/ConstantOptimizingTest.java
```diff
@@ -0,0 +1,342 @@
...
+        assertThat( isStringAssignable( TypeKind.BYTE, true, "0" ) ).isTrue();
+        assertThat( isStringAssignable( TypeKind.BYTE, true, "2" ) ).isTrue();
+        assertThat( isStringAssignable( TypeKind.BYTE, true, "127" ) ).isTrue();
+        assertThat( isStringAssignable( TypeKind.BYTE, true, "-128" ) ).isTrue();
+        assertThat( isStringAssignable( TypeKind.SHORT, true, "1996" ) ).isTrue();
+        assertThat( isStringAssignable( TypeKind.SHORT, true, "-1996" ) ).isTrue();
+    }
+
+    @Test
+    public void testPtrns() {
```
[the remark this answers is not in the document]
sjaakd: ok will fix..

sjaakd:
> @sjaakd I have to say great job on this 😄. It looks really thorough.

thanks :)


> I have some comments inline. Apart from those, shouldn't we do the same for defaultValue?

we should..

> On top of that don't you think that it is better to give the reason why a constant cannot be mapped. For example when the it doesn't fit, or something like that. Basically to take the message from the NumberFormatException.

That might be tricky. The `NumberFormatException` is caught at line 121 in `NativeTypes`.. I just assume it can't be mapped to a `Number` type, so we assume `String` and fall back on everything the user has defined on mapping methods and type conversion, etc. There are several places in where we call the `MappingResolver` and the only outcome is an `Assignment` or `null`.

> Can we also add a some processor tests with the different variants? Where we have boxed, primitive etc. I don't think we have such tests at the moment.

We have some in the same `SourceConstantTests`. But I agree, there could be more. At the other hand.. all possible combinations are checked on `NativeTypes` directly.

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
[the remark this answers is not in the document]
filiphr: Maybe we can write that does in there. I had some trouble understanding it

filiphr:
>we should.. And it already works..

Nice 😄 

>That might be tricky. 

Then it's fine for me.

Looks good to me thanks for the final touches. Can you just add that description from the comment in the code and go for it and merge this 😄

sjaakd:
@filiphr thanks for reviewing.. I opened #1458.. I probably need that one too for my upcoming PR on #5

PR #1412, opened 2018-04-02
#1359 Collection target should be considered as immutable if there is no read accessor

### LINKED ISSUE #1359: MapStruct throws an exception during collection update mapping when target is immutable and there is no target read accessor
The error happens when we have something like:

```
public class Source {

    public Set<String> getProperty() {
    }
}

public class Target {

    public void setProperty(Set<String> property) {
    }
}
```

A workaround is to set the `Mapper#collectionMappingStrategy` to `CollectionMappingStrategy.TARGET_IMMUTABLE`

filiphr:
Fixes #1359.

We need to consider a collection target as immutable if there is no read accessor for it


filiphr:
>was this issue introduced due to the builder?

Not really it was always there. However, I noticed it at work when I was using builders for updates

filiphr:
Thanks for the review @sjaakd

PR #1413, opened 2018-04-02
#1338 Always determine collection argument type when searching for adder

### LINKED ISSUE #1338: In protobuf3, how to mapping a ProtocolStringList
`ProtocolStringList` is `List<String>` in protobuf, mapstruct will not think this is a collection, because no type parameter.

```java
// this is a collection, so this can be done always
if ( !collectionProperty.getTypeParameters().isEmpty() ) {}
```
I need to tell mapstruct to call `addPropertyName` method, not `getPropertyNameList().addAll()`, so , add a mapping from `ProtocolStringList` to `List<String>` will not work.

Maybe mapstruct should support property based addAll method, like `addProperty`, so, mapstruct should use `addAllPropertyName( java.lang.Iterable<java.lang.String> values)`. Then all list in protobuf3 will just works.

filiphr:
If the Collection type is not actually generic it has no type parameters.
However, it's type argument can be determined.
One such list exists in protobuf (ProtocolStringList)

Fixes #1338

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -50,6 +50,8 @@ import org.mapstruct.ap.internal.util.Nouns;
 import org.mapstruct.ap.internal.util.accessor.Accessor;
 import org.mapstruct.ap.internal.util.accessor.ExecutableElementAccessor;
 
+import static org.mapstruct.ap.internal.util.Collections.first;
+
 /**
  * Represents (a reference to) the type of a bean property, parameter etc. Types are managed per generated source file.
  * Each type corresponds to a {@link TypeMirror}, i.e. there are different instances for e.g. {@code Set<String>} and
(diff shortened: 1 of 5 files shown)


sjaakd:
I'll have a look asp.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -560,23 +562,21 @@ private Accessor getAdderForType(Type collectionProperty, String pluralPropertyN
...
-                    ExecutableElement executable = adder.getExecutable();
-                    if ( executable == null ) {
-                        // it should not be null, but to be safe
-                        continue;
-                    }
-                    VariableElement arg = executable.getParameters().get( 0 );
-                    if ( typeUtils.isSameType( arg.asType(), typeArg ) ) {
-                        candidates.add( adder );
-                    }
+            TypeMirror typeArg = first( collectionProperty.determineTypeArguments( Iterable.class ) ).getTypeBound()
```
sjaakd: nice.. I think that the original solution predates the `determineTypeArguments`

filiphr: I think so too 😄

filiphr:
Thanks for the review @sjaakd

PR #1421, opened 2018-04-06
#1417 Add documentation about the builder support

### LINKED ISSUE #1417: Add documentation about the Builders support
How a builder is detect?
How are builder create method and build method detected?

filiphr:
Fixes #1417

### DIFF
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -627,8 +627,13 @@ So for example `Person` has a public static method that returns `PersonBuilder`.
 In our example `PersonBuilder` has a method returning `Person`.
 
 If such type is found then MapStruct will use that type to perform the mapping to (i.e. it will look for setters into that type).
-If no factory method exists for instantiating the builder type the builder creation method would be used.
 To finish the mapping MapStruct generates code that will invoke the build method of the builder.
+
+[NOTE]
+======
+The <<object-factories>> lifecylce is also considered for the builder type. 
+E.g. If an object factory exists for our `PersonBuilder` than this factory would be used instead of the builder creation method.
+======
 
 .Person with Builder example
 ====
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -5,6 +5,8 @@
 :Author: Gunnar Morling, Andreas Gudian, Sjaak Derksen, Filip Hrisafov and the MapStruct community
 :processor-dir: ../../../../processor
 :processor-ap-test: {processor-dir}/src/test/java/org/mapstruct/ap/test
+:integration-tests-dir: ../../../../integrationtest
+:immutables-builder-test: {integration-tests-dir}/src/test/resources/immutablesBuilderTest
 
 [[Preface]]
 == Preface
@@ -312,6 +314,21 @@ The property name as defined in the http://www.oracle.com/technetwork/java/javas
 
 [TIP]
 ====
+Fluent setters are also supported.
+Fluent setters are setters that return the same type as the type being modified.
+
+E.g.
+
+```
+public Builder seatCount(int seatCount) {
+    this.seatCount = seatCount;
+    return this;
+}
+```
+====
+
+[TIP]
+====
 When using Java 8 or later, you can omit the `@Mappings` wrapper annotation and directly specify several `@Mapping` annotations on one method.
 ====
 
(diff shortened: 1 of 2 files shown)


documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -594,6 +611,106 @@ You can find the complete example in the
...
 
+[[mapping-with-builders]]
+=== Using builders
+
+MapStruct also supports mapping of immutable types via builders.
+When performing a mapping MapStruct checks if there is a builder for the type being mapped.
+This is done via the `BuilderProvider` SPI.
+If a Builder exists for a certain type, than that builder will be used for the mappings.
+
+The default implementation of the `BuilderProvider` requires the following for a type to have a builder:
```
sjaakd: This line does not run smoothly. .

sjaakd: So: `The default implementation of the `BuilderProvider` assumes the following:

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -594,6 +611,106 @@ You can find the complete example in the
...
+
+* The type (`Person`) has a parameterless public static method (builder creation method) that returns a type (`PersonBuilder`) different than the one that needs building
```
sjaakd: Split in line + example: The type has a parameterless public static builder creation method that returns a builder of the type. So `Person` has a for example a method `createPerson` that returns `PersonBuilder`.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -594,6 +611,106 @@ You can find the complete example in the
...
+* The found builder type (`PersonBuilder`) has a parameterless public method (build method) that returns the type being build (`Person`)
```
sjaakd: split: The builder type has a parameterless public method (build method) that returns the type being build. In our example: `PersonBuilder` has a method returning `Person`.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -594,6 +611,106 @@ You can find the complete example in the
...
+
+If such type is found then MapStruct will use that type to perform the mapping to (i.e. it will look for setters into that type).
+If no factory method exists for instantiating the builder type the builder creation method would be used.
```
sjaakd: factory method is new.. you just called it a method before. Align.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -594,6 +611,106 @@ You can find the complete example in the
...
+To finish the mapping MapStruct will invoke the build method of the builder.
```
sjaakd: To finish the mapping MapStruct generates code that will invoke the build method of the builder.

sjaakd:
> As if there is a factory method (which we use in the documentation) for the PersonBuilder. Then that method would be used instead of the builder creation method

Ok. Didn't understand it like this.. Perhaps good to refer to chapter on factory methods. Btw; is this valid for automatically selected factory methods and/or annotated ones?

filiphr:
>Btw; is this valid for automatically selected factory methods and/or annotated ones?

We basically delegate the builder type to the resolver. I'll write a test for it as well.

filiphr:
@sjaakd I've updated the documentation. I noticed that Factories don't really work for Builders. I will do the fix in a new PR

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: refer to chapter 9...(ObjectFactoryies)

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -627,9 +627,14 @@ So for example `Person` has a public static method that returns `PersonBuilder`.
 In our example `PersonBuilder` has a method returning `Person`.
 
 If such type is found then MapStruct will use that type to perform the mapping to (i.e. it will look for setters into that type).
-If no factory method exists for instantiating the builder type the builder creation method would be used.
 To finish the mapping MapStruct generates code that will invoke the build method of the builder.
 
+[NOTE]
+======
+The <<object-factories>> lifecylce is also considered for the builder type.
```
sjaakd: the term " lifecycle"  is confusing..

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
filiphr: I've removed that line actually

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
filiphr: That's true. What about just:

>The <<object-factories>> are also considered...

sjaakd: 👍

PR #1422, opened 2018-04-06
#1418 Add support for NoOp BuilderProvider for turning off using of builders

### LINKED ISSUE #1418: Add a NoOp BuilderProvider
A no operation `BuilderProvider` that users can use to disable the builder support. The provider would just return `null` when requested for builders

filiphr:
Fixes #1418


sjaakd:
@filiphr .. I understood from the issue that the builder approach is causing problems with existing users..  I'm thinking: should we explicitly enable use of builders somehow? (default off)?

filiphr:
I am not entirely sure what the best approach is. The problems the user faced was that he / she couldn't do update mappings (I have created #1423 for that).

This PR allows the users to explicitly opt out of the builders. I would propose that we go with the opt out approach at least in the first beta and get feedback from our users. We can then decide what to do.

filiphr:
>I'm ok, as long as we mention the consequences in the migration notes.

Like not being able to use update mappings?

filiphr:
Thanks for the review @sjaakd

PR #1424, opened 2018-04-08
#1414 Pass originating mapper element to the Filer API

### LINKED ISSUE #1414: Pass originating elements to the Filer API
When creating sources and resources tools can wrap the FIller API and listen on the originating elements to perform certain caching. One example is Gradle see https://github.com/gradle/gradle/issues/1320.

filiphr:
Fixes #1414 

I was thinking to add a test with Mockito. However, that would require to perform a lot of mocking just to validate that we are passing the mapper element to the API. 

If you think it is worth it, I can add it


filiphr:
Thanks for the review @sjaakd 

You are right about Mockito. I also think it is not quite worth it for us and for this test

Issue #1383, opened 2018-02-19
Report error on the @Mapping annotation when using invalid values


filiphr:
In `Mapping#fromMappingPrism` we do some validation checks and report an error on the method instead of the `@Mapping` annotation.

example:

```
@Mapper
public interface MyMapper {

    @Mapping( target = "id", expression = "java(String.valueOf( \"123\" )), constant = "1234") // error should be reported here
    Target map(Source source); // error report on this line
}
```

MatteoJoliveau:
Hi @filiphr, I was looking to start contributing to MapStruct and stumbled upon this issue.
If you could explain a little what are the requirements I would love to tackle it and get a grasp on the project :)

filiphr:
Hey @MatteoJoliveau it's really nice that you want to contribute to the project. We are always looking for contributors 😄.


Regarding this issue. The idea is to change the location of the reported error. Currently it is reported on the method, but it would be better if it is reported on the `@Mapping` annotation itself. The location for this changes are in [`Mappings#fromPrism`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java#L99). The idea is in `message.printMessage(...)` to pass in the `AnnotationMirror` of the `MappingPrism` and the `AnnotationValue` from the appropriate location (`expression`, `defaultValue`, etc).

There are already tests that cover all those error messages:
* [`SourceConstantsTests#L115`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/test/java/org/mapstruct/ap/test/source/constants/SourceConstantsTest.java#L115) -> In this test the first error should be on line 37 and the second one is OK.

You can find the others by searching for the error message in the code base.

If you need more help feel free to post more questions 😄. You can also find us in or gitter [mapstruct/mapstruct-users](https://gitter.im/mapstruct/mapstruct-users) chat. Usually we respond in the evenings there (CET time)

MatteoJoliveau:
Hi @filiphr, thank you! I'll start digging in the codebase tomorrow and get in touch if needed :)

dvfeinblum:
Hey there @filiphr.

I'm also looking to get my feet wet on Mapstruct and saw that this issue had quieted down. Thought I'd take a crack at it.

[...]

PR #1429, opened 2018-04-15
#1383 Report error on the @Mapping annotation when using invalid values

dvfeinblum:
This PR will change the line that is reported in an error message when an `@Mapping` annotation contains improper values. An example of this is:
```
@Mapper
public interface MyMapper {

    @Mapping( target = "id", expression = "java(String.valueOf( \"123\" )), constant = "1234") // error should be reported here
    Target map(Source source); // error report on this line
}
```

Before this PR, the error message would report an error on the Target. Now, the error message will report the line that the @Mapping annotation is on.


dvfeinblum:
Realized I missed a few tests. Sorry about that! I'll reopen this once I fix them.

Issue #1427, opened 2018-04-13
@Mapper(componentModel = "spring") doesn't support custom name for Spring Service Annotation


zurbergram:
Can't specify a name for the Spring Service annotation
If the interface definition something like

    @Mapper(componentModel = "spring", componentModelName = "cakeMapperV2")
     public interface CakeMapper{
         Cake map(Cupcake cupcake);
     }

     @Service("cakeMapperV2")
     public class CakeMapperImpl {
     }

filiphr:
@zurbergram yes this is not possible in the moment out of the box.

However, there is a way for you to achieve what you are looking for with some internal customizations. We have an internal SPI [ModelElementProcessor](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/processor/ModelElementProcessor.java) that adds the spring `@Component` annotation. The one for Spring is the [SpringComponentProcessor](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/processor/SpringComponentProcessor.java).

What you can do is to define your own `ModelElementProcessor` that would be of a custom component model, provide your own marker annotation for the name of the component and you would be able to do what you are looking for.

Just keep in mind that this is an internal SPI, which means that we cannot guarantee backwards compatibility between versions. However, it is rarely changed.

chris922:
Oups, sorry! I wasn't aware about the fact that changes in my forked mapstruct repo will directly be linked to this issue. Hopefully I haven't spammed you with too much notification mails.

@filiphr as my changes are now linked here maybe you could already take a look or should I create a PR? _(this is my more or less first 'real' OS contribution; right now I am waiting that the built finishs locally including all tests)_

I tried to add the possibility to have an extra annotation (`@MapperSpringConfig` / `@MapperJsr330Config`) to be able to have more component model specific configurations, e. g. if `@Service` or `@Component` should be used and - as requested in this ticket - to define an own bean name.

Maybe its better to extend the `@Mapper` configuration with a new property like `componentName`, but my idea was that there are maybe also other configuration things that only belongs to a special `componentModel` and having a special annotation for this could be helpful for those usecases.

i-telligence-af:
> @zurbergram yes this is not possible in the moment out of the box.
> 
[...]

PR #1437, opened 2018-04-18
 #1427 Add additional configuration for Jsr330 and Spring component models

chris922:
Added two new annotations that allows to configure the component models, e. g. the bean name

* `@MapperSpringConfig`
* `@MapperJsr330Config`

This adds the functionality requested in ticket #1427. A few more information from my side are available there.
I added `@Since 1.3` to new classes, so in case the PR will be accepted after 1.3 was released I/someone must update this.


sjaakd:
@chris922 First of all, nice idea. But I also need to think about it a bit more.. 

- I think we need to discuss this with the MapStruct team. 
- Are the configurations mutual exclusive..? What happens if you specify them both?
- Is there a need for the component model parameter in `@Mapping` when we introduce such configuration annotations? Or should we cut the suffix @..Config?
- I'm not a big fan of using the 'value' in the annotations.. There's an implicit meaning to be attached.. I'm not sure whether in a `@..Config` the name is the proper candidate for `value`. 
- What about CDI?

Anyway.. If we are going to swap out the `componentName` for annotations, that should be 2.0.. (which is upcoming :smile:. ).

chris922:
Hi @sjaakd,

* Let me know the outcome of your discussion! Would be happy if this PR will be merged and if not then at least I learned more internal stuff from mapstruct (what was my main intention)
* Only the configuration that matches the configured `componentModel` in `@Mapper` will be used. So in case you define `@MapperJsr330Config` but set your `componentModel` to `spring` it will be ignored. Maybe a warning should be printed in this case?
* Right now it is required. Maybe we can make it optional but then we've got issues in case both new annotations will be used (I would propose to stop the processing with an error)
* I could remove the `value` attribute and use `componentName` or something similar if you prefer
* No special configuration for CDI available right now, any ideas what could be configured? I never used CDI myself.. always Spring ;)

To what `componentName` attribute do you refer? In 1.2.0 I am not aware about something like this. Already something new that I missed?

Thanks for your feedback!

Best regards
Christian

sjaakd:
> I learned more internal stuff from mapstruct (what was my main intention)

I really appreciate that. Its good to have more folks on board. :smile:.

> To what componentName attribute do you refer? In 1.2.0 I am not aware about something like this. Already something new that I missed?

My wrong... Doing too much at the same time.. I meant `componentModel`. I could envision that you only need to specify (a form of) your annotation in stead of a `componentModel` in the `@Mapper` annotation. But that would be downward incompatible and a `natural` candidate for a 2.0 version.

chris922:
> I could envision that you only need to specify (a form of) your annotation in stead of a componentModel in the @Mapper annotation. But that would be downward incompatible and a natural candidate for a 2.0 version.

I think we could already implement that `componentModel` is not required without removing it - just mark it as deprecated and don't remove the functionality.
And in 2.0 it could be removed. So that both things are valid.. maybe with the `componentModel` as the leading value.

filiphr:
This is indeed an interesting PR. Replacing the `Mapper#componentModel` with a new annotation sounds like a good idea. We would only need to see how the inheritance of the component model would work. As currently one can set it via `@MapperConfig` as well. Maybe in 2.0 we can even add support for meta annotations, so we might not need the config anymore 😃.

Using a custom annotation would be easier to add injection specific parameters, like the `delegateQualifier` and `componentType` in the `MapperSpringConfig`.

@chris922 I had a quick glance over it and there are some places that I think we can improve one:

* Instead of specifying our own enum for `SpringComponentType` maybe we can have a string or a class that would be the annotation that needs to be used. In that case the users can even pick their own custom meta annotations
* The names, one of the most difficult thing in programming 😃. Maybe something like `@SpringMapper`, `@Jsr330Mapper`, etc, or add `Config` in the end.
* There are some other general changes in the way some internals are used, but we can go over that later

In any case our plan is to release 1.3.0.Beta1 as soon as possible. If you don't mind I would put this for the release after that 1.3.0.Beta2.

chris922:
Hi @filiphr ,

* I started with the same idea to allow specifying the annotation directly, but I choosed the enum so that it is easier to allow different generations based on the `componentType` and also to restrict it to the 4 possible component annotations from Spring. Otherwise it could be misused for other stuff. Of course it could be restricted to those annotations and it is still possible to do different things during generation, but I always find enumerations easier. Better overview about possibilities and built-in restriction to those values
* "There are only two hard things in computer science: [...] and _naming things_"... you know ;)
  Just `@SpringMapper`/`@Jsr330Mapper` sounds fine for me. Short and meaningful.
* Okay, just let me know.. e. g. add comments in the code :smiley:


I try to summarize the current status:

1. Make `componentModel` deprecated in `@Mapper` and make it possible to just use the new annotations
1.1. `componentModel` is leading, print `WARN` in case `componentModel` doesnt match used annotation / multiple annotations used
1.2. `WARN` as soon as `componentModel` will be used as it is "deprecated and will be removed in 2.0"
1.3. `ERROR` in case multiple component model annotations are used in parallel
2. Add `@CdiMapper` annotation (without special configuration right now, or do you have any ideas what could be configured?)
3. No annotation + no `componentModel` = default mapping (as is)
4. Rename `value` in new annotations to `componentName`
5. Rename annotations to `@SpringMapper` and `@Jsr330Mapper`

If those things fit your expectation I would already start with them. Please let me know.

I will wait for feedback if the `componentType` should be switched to `Class<? extends Annotation>` or `String`.
And of course if you have some other ideas or comments.

sjaakd:
Hmm. its a pity java does not allow you to etend annotations. It would make perfect sense to let `Jsr330Mapper` inherit from `Mapper` 😞

core-common/src/main/java/org/mapstruct/util/ComponentModelMapper.java
```diff
@@ -0,0 +1,44 @@
...
+ * <p>It is never allowed to have more than one component model annotation at one
+ * {@link org.mapstruct.Mapper @Mapper}.<br>
+ * In case there is no annotation of this kind the {@code default} component model will be used.
+ *
+ * @author Christian Bandowski
+ */
+@Documented
+@Target(ElementType.ANNOTATION_TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ComponentModelMapper {
```
chris922: I had to add this meta-annotation that allows to recognize the `@SpringMapper`/.. annotations, otherwise it is not possible to detect that no such annotation exist and the `default` component model should be used - or if multiple component model annotations exists (-> ERROR).

processor/pom.xml
```diff
@@ -98,6 +98,11 @@
             <artifactId>javax.inject</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>javax.enterprise</groupId>
+            <artifactId>cdi-api</artifactId>
```
chris922: Seems that I added the first CDI unit tests ... hopefully in a correct way

processor/src/main/java/org/mapstruct/ap/internal/util/MapperConfiguration.java
```diff
@@ -49,16 +51,41 @@
...
-    public static MapperConfiguration getInstanceOn(Element e) {
-        return new MapperConfiguration( MapperPrism.getInstanceOn( e ) );
+    private final FormattingMessager messager;
+
+    public static MapperConfiguration getInstanceOn(Element e, FormattingMessager messager) {
+        // store configurations per Element instance in a cache map
+        // this prevents creating many instances of this configuration for the same Element
+        // it will be assumed that this method will never be called with different messagers for the same Element
+        // one reason to cache instances is that the the MapperConfiguration validates the Element and multiple
+        // instantiations would lead to multiple identical warnings/errors
```
chris922: Is this okay? Maybe a `WeakHashMap` is too much as a processor is only a short-running process. Maybe you've got better ideas

processor/src/main/java/org/mapstruct/ap/internal/util/MapperConfiguration.java
```diff
@@ -70,6 +97,36 @@ private MapperConfiguration(MapperPrism mapperPrism) {
...
+            ).value();
+            if ( !mapperPrism.componentModel().equalsIgnoreCase( annotationValue ) ) {
+                messager.printMessage(
+                    e,
+                    componentModelAnnotation,
+                    Message.COMPONENT_MODEL_PROPERTY_NOT_MATCHING_ANNOTATION,
+                    mapperPrism.componentModel()
+                );
+            }
+        }
```
chris922: I added the validation of the mapper here, there is also a TODO a few lines earlier that some more validations are required here. Removing the `printMessage` calls here would allow to remove the `configurations` cache.

processor/src/main/java/org/mapstruct/ap/internal/util/MapperConfiguration.java
```diff
@@ -49,16 +51,41 @@
...
+        AnnotationMirror componentModelAnnotation = null;
+
+
+        return componentModelAnnotation;
     }
 
-    private MapperConfiguration(MapperPrism mapperPrism) {
-        this.mapperPrism = mapperPrism;
+    private MapperConfiguration(Element e, FormattingMessager messager) {
+        this.mapperPrism = MapperPrism.getInstanceOn( e );
```
chris922: Moved the `MapperPrism.getInstanceOn` to the constructor as the `Element` is required a few lines later. I also added the `FormattingMessager` to be able to print the validation messages.

filiphr:
@chris922 just to let you know that we haven't forgotten this one.  We are focused on the 1.3.0.Beta1 release and we want to get it out as soon as possible (and issues just keep popping up 😟). Once we get that out I am going to have a look at your comments and update in this one. I hope this is fine for you

filiphr:
Closing this since it is now possible to achieve this with the new `@AnnotateWith` annotation.

PR #1438, opened 2018-04-18
Remove cloudbees CI links and add travis-ci

chris922:
Just a small PR that replaces the cloudbees links I found with travis-ci links.

I haven't found such a nice "built on" image that is shown for cloudbees for travis-ci, but I guess this is not so important as the build status is already available at the top of `readme.md`.


filiphr:
Thanks a lot for this PR @chris922. It has been integratd

Issue #1425, opened 2018-04-10
org.joda.time.LocalDate to java.time.LocalDate generated code has compilation error


abhijitkottur:
There is a compilation error on highlighted line as the object **dt** is being referred as a java.timeLocalDate whereas it should have been a org.joda.time.LocalDate. The import statement is incorrectly generated. 

Please let me know if there is another way of mapping between these dates or if I am missing something.

The generated code:

```
dto.setRequestedDeliveryDate( xmlGregorianCalendarToLocalDate( jodaLocalDateToXmlGregorianCalendar( entity.getRequestedDeliveryDate() ) ) );



private XMLGregorianCalendar jodaLocalDateToXmlGregorianCalendar( LocalDate dt ) {
        if ( dt == null ) {
            return null;
        }

        try {
            return DatatypeFactory.newInstance().newXMLGregorianCalendarDate(
                dt.getYear(),
                **dt.getMonthOfYear(),**
                dt.getDayOfMonth(),
                DatatypeConstants.FIELD_UNDEFINED );
        }
        catch ( DatatypeConfigurationException ex ) {
            throw new RuntimeException( ex );
        }
    }

private static java.time.LocalDate xmlGregorianCalendarToLocalDate( XMLGregorianCalendar xcal ) {
        if ( xcal == null ) {
            return null;
        }

        return java.time.LocalDate.of( xcal.getYear(), xcal.getMonth(), xcal.getDay() );
    }
```

filiphr:
This has been fixed in PR #1442 on top of that all our templates now would generate a FQN or a simple name based on the type themselve (at the moment they were handwritten by us). Issues like this should no longer occur.

Thanks a lot to @chris922 for doing the fix and in general improving our entire built in template support

PR #1439, opened 2018-04-18
 #1425 Update builtin templates and add package name to all classes

chris922:
Importing the classes, as it was before, makes troubles as soon as a
class with the same name in a different package is used. It is not
possible to have two imports for classes with the same simple-name and
thus it is safer to always include the package name and don't use
imports for the builtin templates

Hopefully I don't missed a package in the templates, but all tests were fine. So in case for every builtin template there is at least one testcase my changes should be okay. Otherwise there should be a compile issue which would break the test.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/BuiltInMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/BuiltInMethod.java
@@ -18,14 +18,11 @@
  */
 package org.mapstruct.ap.internal.model.source.builtin;
 
-import static org.mapstruct.ap.internal.util.Collections.first;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
-
 import javax.lang.model.element.ExecutableElement;
 
 import org.mapstruct.ap.internal.conversion.SimpleConversion;
@@ -38,6 +35,8 @@ import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.model.source.ParameterProvidedMethods;
 import org.mapstruct.ap.internal.util.MapperConfiguration;
 import org.mapstruct.ap.internal.util.Strings;
+
+import static org.mapstruct.ap.internal.util.Collections.first;
 
 /**
  * Represents a "built-in" mapping method which will be added as private method to the generated mapper. Built-in
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/CalendarToXmlGregorianCalendar.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/CalendarToXmlGregorianCalendar.java
@@ -18,14 +18,7 @@
  */
 package org.mapstruct.ap.internal.model.source.builtin;
 
-import static org.mapstruct.ap.internal.util.Collections.asSet;
-
 import java.util.Calendar;
-import java.util.GregorianCalendar;
-import java.util.Set;
-
-import javax.xml.datatype.DatatypeConfigurationException;
-import javax.xml.datatype.DatatypeFactory;
 import javax.xml.datatype.XMLGregorianCalendar;
 
 import org.mapstruct.ap.internal.model.common.Parameter;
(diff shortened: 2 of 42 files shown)


chris922:
travis-ci failed because it tried to run my test (which requires java 8) with java 7. Is there a special annotation or something like this to tell travis/maven that this test requires java >= 8?

gunnarmorling:
Hey @chris922, first of all, thanks a lot for your efforts. Great to see you being interested in contributing to MapStruct!

That being said, I'm having some doubts about this change. The usage of imports in generated code was a conscious decision, the idea being that generated code should be equally well readable as hand-written code. So I'd prefer to stick to this approach.

> It is not possible to have two imports for classes with the same simple-name

That's true, but MapStruct should take care of this case already. I.e. it adds an import when seing a type with a given simple name for the first time and then if another type with the same simple name is used, it'll add the FQN for that in generated code (as you'd do it in hand-written code).

If this doesn't work in a given case, it should be considered as a bug. Could you provide an example where this happens so we could explore why import handling is broken in this situation? Thanks!

chris922:
Hi @gunnarmorling ,

Thanks for the feedback! I personally also like it much more having imports than the FQNs.

I analyzed issue #1425 and tried to built up a test-case that fails -> `Issue1425Test`
And I think there is really a bug in the way the imports will be chosen, because the `java.time.LocalDate` will be imported and not the `org.joda.time.LocalDate`. So if this will be changed it might work.

Nevertheless I haven't tried to find the solution for this bug, because I think even if it is fixed there _could_ be an issue in the future. 

What if two templates will be used that are using the same class name, but from different packages?
Looks like it is right now not possible, because `JodaLocalDateToXmlGregorianCalendar.ftl` is using the simple-name and `LocalDateToXmlGregorianCalendar.ftl` the FQN. I don't see other templates where we could have such a collision.
But this means you must know what other templates exist to be able to determine if you have to use the FQN or if you can use the SimpleName.

As those are only stored in the internal API this could be totally fine, but imho it is also a bit error-prone and thus always using FQN will prevent those things.

But yes, I also agree with you that the generated source code should be easily readable. So this is totally reasonable if this is more important. Feel free to close the PR in this case :)

Maybe I will also find some more time to analyze the root cause of #1425 a bit more. 
It looked like the ordering in which the `typeFactory` for `java.time.LocalDate` and `org.joda.time.LocalDate` will be called is important. As the `Issue1425Mapper` is using `java.time.LocalDate` maybe this type will be created first, with an import, and then the template will be generated and the `org.joda.time.LocalDate` type will be created, without an import. But this definitly requires it.

filiphr:
@chris922 first thanks a lot for the efforts and taking the time to make this contribution.

Regarding the PR I also agree with @gunnarmorling about this. I don't think that we need to add the FQN to all the template.

I think that the reason for #1425 is that the `XmlGregorianCalendarToLocalDate` imports the return type which is `java.time.LocalDate`. however it's template `XmlGregorianCalendarToLocalDate.ftl` uses the FQN. On top of that you also have `XmlGregorianCalendarToJodaLocalDate` that also imports the return type and it's template does not use the FQN. 

The fix for this would be to use the FQN for `org.joda.LocalDate` in the `XmlGregorianCalendarToJodaLocalDate.ftl` and not import the java and joda returns types in the appropriate classes. Do you want to update this PR with the fixes, or should we close this one and you can do another one?

On long term the best might be to use our `<@includeModel object=someType/>` that handles whether the type is already imported or not and uses an FQN if needed.

chris922:
@filiphr I also noticed that there are sometimes imports given that aren't required because the FQN is used.

I also had a similar idea like you with the `<@includeModel object=someType/>`, but my idea was to create an own directive to have something like `<@type name="org.joda.LocalDate"/>`
As far as I understood the `includeModel` directive it is required to set the correct `Type` instance as `object` value and it will then be written to the file. I guess it is probably a bit more complex to get the correct instance.

Having an own directive would simplify this and I already analyzed this and thus I think it is also possible to implement this. When I got it correctly the `Type` instance could be found within the `importTypes` from the `Environment` and when it was found it will be written to the file.

What do you think about such an approach? I could try to implement this and update a few (maybe all) templates.

chris922:
I was able to get a PoC working in the way I described .. so in case you like this and don't have other ideas I would clean up & enhance the code, integrate it into more templates and check if everything still works.

Maybe you've got a better idea to get the `Type` instance instead of retrieving it from the `importTypes` (but it looked like this is really often available?).. maybe there is a way to get the `TypeFactory` instance from the `Environment`. This would allow to define any type.. but when it is nowhere in the `importTypes` the import will probably never created.. so maybe the way I did it is already good to prevent errors.

filiphr:
@chris922 the approach sound interesting. Your idea gave me another idea. We can do something like:
`<@includeModel object=findType('javax.xml.datatype.XMLGregorianCalendar')/> ` and then in `VirtualMappingMethod` we can have a method:

```
public Type findType(String fqn) {
    //do the find
}
```

The `BuiltInMethod` and the `HelperMethod` can provide a map between the fqn and the type that they are using so we can do a lookup there. 

This is similar to your directive.

chris922:
Sounds like it's simpler to implement and would probably not have any package-cycle issues.

I will give it a try. 

The only drawback I could see for this approach is that you have to type more code in the templates compared to using an own directive. But I would say the most important thing is the generated code and not the code within the template.

filiphr:
Actually `includeModel` is our own directive. It's `ModelIncludeDirective` and it will actually trigger the template `Type.ftl` that handles the FQN stuff properly.

To make it simpler you might want to use `<@includeMode object=returnType />` in the template return types and `<@includeModel object=parameters[0]/>`. Although that would make the template more difficult to read.

chris922:
I just created #1442 as the solution there is completly different like the one in this PR.

Therefore I will close this here.

PR #1442, opened 2018-04-20
#1425 Added findType to VirtualMappingMethod and use it in all builtin templates

chris922:
_(new PR, discussion started in #1439)_

Together with the includeModel directive this will ensure that the type will be written to the file as a FQN if required, otherwise as a simple name.

The PR allows extending templates like this:
```java
<@includeModel name=findType("LocalDate")/>
```
or
```java
<@includeModel name=findType("org.joda.time.LocalDate")/>
```

The given fully-qualified/simple name must be available in the `importTypes` of the builtin method. Unlike it was proposed in #1439 I have not extended `BuiltinMethod` and `HelperMethod` to provide a `Map` with possible `Types` as the included `Types` must be returned in the `importTypes` property, otherwise the import statement will maybe not be added. So it should fail-fast in case someone tries to include a `Type` that might not be added to imports which would generate not compilable code.

There is just one thing I am not completly happy with. In case the `Type` has generic parameters those will be written to the template as well, therefore - to ensure the correct `Type` name will be used - the `JaxbElemToValue.ftl` template looks like this:

```java
<#assign jaxbElement = findType("JAXBElement")>
<#assign typeElement = jaxbElement.typeParameters.get(0)>

private <<@includeModel object=typeElement/>> <@includeModel object=typeElement/> ${name}( <@includeModel object=jaxbElement/> element ) {
```
A bit hard to read, but possible. Maybe someone has a good idea to enhance this to make it easier readible.

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/source/builtin/JaxbElemToValue.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/source/builtin/JaxbElemToValue.ftl
@@ -19,7 +19,10 @@
      limitations under the License.
 
 -->
-private <T> T ${name}( JAXBElement<T> element ) {
+<#assign jaxbElement = findType("JAXBElement")>
+<#assign typeElement = jaxbElement.typeParameters.get(0)>
+
+private <<@includeModel object=typeElement/>> <@includeModel object=typeElement/> ${name}( <@includeModel object=jaxbElement/> element ) {
     if ( element == null ) {
         return null;
     }
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/source/builtin/JaxbElemToValue.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/source/builtin/JaxbElemToValue.ftl
@@ -19,7 +19,6 @@
      limitations under the License.
 
 -->
-
 private <T> T ${name}( <@includeModel object=findType("JAXBElement") raw=true/><T> element ) {
     if ( element == null ) {
         return null;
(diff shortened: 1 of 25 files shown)


processor/src/main/resources/org/mapstruct/ap/internal/model/source/builtin/JaxbElemToValue.ftl
```diff
@@ -19,7 +19,10 @@
      limitations under the License.
 
 -->
-private <T> T ${name}( JAXBElement<T> element ) {
+<#assign jaxbElement = findType("JAXBElement")>
+<#assign typeElement = jaxbElement.typeParameters.get(0)>
+
+private <<@includeModel object=typeElement/>> <@includeModel object=typeElement/> ${name}( <@includeModel object=jaxbElement/> element ) {
```
filiphr: This method is actually a generic method with it looks like:

```
private <T> T something( JAXBElement<T> element)
```

which is correct. In order to get what you are looking for there is a property that you can pass. The template would then look like:

```
private <T> T ${name}( <@includeModel object=findType("JAXBElement") raw=true /> ) {
```

The `Type.ftl` doesn't use the generic parameters then

chris922:
Built failed using JDK 7.. the test requires JDK 8 and I don't know how to disable it for JDK < 8.

I checked other test cases using new JDK 8 stuff and I am not able too see any settings to disable those tests for JDK < 8. 

How could I do this?

chris922:
@filiphr Thanks again for your feedback, all required changes are hopefully integrated and travis also reported that everything is fine :)

filiphr:
Thanks a lot for the PR @chris922 I have integrated into master. I did a last commit to remove an extra line and remove a not needed test method.

PR #1445, opened 2018-04-21
#1423 Updating types that have a builder should be allowed

### LINKED ISSUE #1423: Allow using update mappings for types with builders
As mentioned in https://github.com/mapstruct/mapstruct/issues/1417#issuecomment-378971085 some users can use both builders and setters to perform their mappings. This means that we could still do update mappings.

i.e. Not all types that have builders are immutable

filiphr:
It is possible that a type has both a builder and accessors.
In such case doing an update to this type should be allowed

Fixes #1423


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -114,8 +114,12 @@ private Builder setupMethodWithMapping(Method sourceMethod) {
             this.method = sourceMethod;
             this.methodMappings = sourceMethod.getMappingOptions().getMappings();
             CollectionMappingStrategyPrism cms = sourceMethod.getMapperConfiguration().getCollectionMappingStrategy();
-            Map<String, Accessor> accessors = method.getResultType()
-                .getEffectiveType()
+            Type mappingType = method.getResultType();
+            if ( !method.isUpdateMethod() ) {
```
sjaakd: formulate positive: `if(method.isUpdateMethod).. else ..`

processor/src/test/java/org/mapstruct/ap/test/builder/mappingTarget/simple/BuilderInfoTargetTest.java
```diff
@@ -54,17 +52,48 @@ public void testSimpleImmutableBuilderHappyPath() {
...
-            @Diagnostic(type = ErroneousSimpleBuilderMapper.class,
-                kind = javax.tools.Diagnostic.Kind.ERROR,
-                line = 26,
-                messageRegExp = "^Can't generate mapping method when @MappingTarget is supposed to be immutable "
-                + "\\(has a builder\\)\\.$")
-        })
-    public void shouldFailCannotModifyImmutable() {
+    public void testMutableTargetWithBuilder() {
+        SimpleMutableSource source = new SimpleMutableSource();
+        source.setAge( 20 );
```
sjaakd: you are kidding 😃

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
sjaakd: don't think its that important .. its only readability..

PR #1446, opened 2018-04-22
Out of the box support for Immutables

### LINKED ISSUE #1415: Decide how to provide support for Immutables
The support for Immutables requires custom `AccessorNamingStrategy` and `BuilderProvider` implementations. We already have one in our integration tests. How do we want to provide this support to our users:

* Add a new module that sets the SPIs
* Add the class to the `mapstruct-processor` module and users will have to opt in by changing the `AccesorNamingStrategy` and `BuilderProvider`
* Part of the defaults and doing the required checks in them
* Add the SPIs in the examples repo and users would need to copy them

filiphr:
This PR adds out of the box support for the Immutables builders by using the Immutables specific `AccessorNamingStrategy` and `BuilderProvider` when `@Value.Immutable` is present.

I had to move the static usage of the `AccessorNamingStrategy` into a different class, because we need that to be dynamic and we need the `Elements` to check if the Immutables elements are present in the annotation round.

Fixes #1415


sjaakd:
@filiphr .. started reviewing.. need some more time.

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
```diff
@@ -343,11 +345,13 @@ public void init(SourceMethod method, FormattingMessager messager, TypeFactory t
...
      * @param messager the messager that can be used for outputting messages
      * @param typeFactory the type factory
+     * @param namingUtils the accessor naming utils
      * @param isReverse whether the init is for a reverse mapping
      * @param reverseSourceParameter the source parameter from the revers mapping
      */
-    private void init(SourceMethod method, FormattingMessager messager, TypeFactory typeFactory, boolean isReverse,
-        Parameter reverseSourceParameter) {
+    private void init(SourceMethod method, FormattingMessager messager, TypeFactory typeFactory,
+                      AccessorNamingUtils namingUtils, boolean isReverse,
```
sjaakd: Now this is an instance.. should we somehow combine this with the `CollectionMappingStrategy`? Feels like this is very much related and in the past I kept that one dragging around.. Anyway.. perhaps solve that separately.

processor/src/main/java/org/mapstruct/ap/internal/util/Filters.java
```diff
@@ -36,11 +36,11 @@
     private Filters() {
     }
 
-    public static List<Accessor> getterMethodsIn(List<Accessor> elements) {
+    public static List<Accessor> getterMethodsIn(AccessorNamingUtils namingUtils, List<Accessor> elements) {
         List<Accessor> getterMethods = new LinkedList<Accessor>();
 
         for ( Accessor method : elements ) {
-            if ( Executables.isGetterMethod( method ) ) {
+            if ( namingUtils.isGetterMethod( method ) ) {
```
sjaakd: readability use full name at these places.: accessorNamingUtil or just  accessorNaming (expressing the function better)

processor/src/main/java/org/mapstruct/ap/internal/util/Filters.java
```diff
@@ -60,34 +60,35 @@ private Filters() {
...
     }
 
-    public static List<ExecutableElementAccessor> presenceCheckMethodsIn(List<Accessor> elements) {
+    public static List<ExecutableElementAccessor> presenceCheckMethodsIn(AccessorNamingUtils namingUtils,
+                                                                         List<Accessor> elements) {
         List<ExecutableElementAccessor> presenceCheckMethods = new LinkedList<ExecutableElementAccessor>();
 
         for ( Accessor method : elements ) {
-            if ( Executables.isPresenceCheckMethod( method ) ) {
+            if ( namingUtils.isPresenceCheckMethod( method ) ) {
```
sjaakd: same here

processor/src/main/java/org/mapstruct/ap/internal/util/Filters.java
```diff
@@ -60,34 +60,35 @@ private Filters() {
...
         return presenceCheckMethods;
     }
 
-    public static List<Accessor> setterMethodsIn(List<Accessor> elements) {
+    public static List<Accessor> setterMethodsIn(AccessorNamingUtils namingUtils, List<Accessor> elements) {
         List<Accessor> setterMethods = new LinkedList<Accessor>();
 
         for ( Accessor method : elements ) {
-            if ( Executables.isSetterMethod( method ) ) {
+            if ( namingUtils.isSetterMethod( method ) ) {
```
sjaakd: accessorNaming..

processor/src/main/java/org/mapstruct/ap/internal/util/Filters.java
```diff
@@ -60,34 +60,35 @@ private Filters() {
...
         return setterMethods;
     }
 
-    public static List<Accessor> adderMethodsIn(List<Accessor> elements) {
+    public static List<Accessor> adderMethodsIn(AccessorNamingUtils namingUtils, List<Accessor> elements) {
         List<Accessor> adderMethods = new LinkedList<Accessor>();
 
         for ( Accessor method : elements ) {
-            if ( Executables.isAdderMethod( method ) ) {
+            if ( namingUtils.isAdderMethod( method ) ) {
```
sjaakd: here too

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
filiphr: What do you mean combine it with the `CollectionMappingStrategy`? I don't get it 😄

processor/src/main/java/org/mapstruct/ap/internal/util/Filters.java
filiphr: `accessorNaming` it is then

filiphr:
Not sure why something broke now. I have to leave I will have a look at it tonight

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
sjaakd: Well, for collections which method to select in the end comes down to both: `CollectionMappingStrategy` and `AccessorNamingUtils`. Now they have to be combined on the call site.

filiphr: They were always combined. Sorry if I don't quite follow

PR #1448, opened 2018-04-22
#1317 Add support for ignoring unmapped source properties

### LINKED ISSUE #1317: Add support for ignoring source properties follow up on #610
With the integration of #1282 we have support for reporting on unmapped source properties. Additionally we need to add support for ignoring certain source properties.

We need to do the following:

1. Make the `target()` parameter of `@Mapping` optional, so that we can simply use `@Mapping(source="foo", ignore=true)`
1. `target()` should only be allowed to be empty when `unmappedSourcePolicy != IGNORE`, `ignore = true` and `source` has some value, otherwise it should not be allowed.
1. Ignoring `source` should only be allowed hen `unmappedSourcePolicy != IGNORE`

filiphr:
This property has only effect on the unmapped source properties report

Fixes #1317

### DIFF
--- a/core-common/src/main/java/org/mapstruct/BeanMapping.java
+++ b/core-common/src/main/java/org/mapstruct/BeanMapping.java
@@ -85,12 +85,12 @@ public @interface BeanMapping {
     boolean ignoreByDefault() default false;
 
     /**
-     * The unmapped source properties that need to be ignored. This should be used when {@link ReportingPolicy#WARN}
+     * Unmapped source properties to be ignored. This could be used when {@link ReportingPolicy#WARN}
      * or {@link ReportingPolicy#ERROR} is used for {@link Mapper#unmappedSourcePolicy()} or
-     * {@link MapperConfig#unmappedSourcePolicy()}. When properties are set they won't be part of the unmapped source
-     * properties report. This property has no effect on choosing whether a source property needs to be mapped.
+     * {@link MapperConfig#unmappedSourcePolicy()}. Listed properties will be ignored when composing the unmapped
+     * source properties report.
      * <p>
-     * NB: This does not support ignoring nested source properties
+     * <b>NOTE</b>: This does not support ignoring nested source properties
      *
      * @return The source properties that should be ignored when performing a report
      *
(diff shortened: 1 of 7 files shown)


core-common/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -83,4 +83,18 @@
      * @since 1.3
      */
     boolean ignoreByDefault() default false;
+
+    /**
+     * The unmapped source properties that need to be ignored. This should be used when {@link ReportingPolicy#WARN}
```
sjaakd: should == could?

core-common/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -83,4 +83,18 @@
...
+     * or {@link ReportingPolicy#ERROR} is used for {@link Mapper#unmappedSourcePolicy()} or
+     * {@link MapperConfig#unmappedSourcePolicy()}. When properties are set they won't be part of the unmapped source
+     * properties report. This property has no effect on choosing whether a source property needs to be mapped.
+     * <p>
+     * NB: This does not support ignoring nested source properties
```
sjaakd: NB = NOTE?

core-common/src/main/java/org/mapstruct/BeanMapping.java
sjaakd: Unmapped source properties to be ignored.

core-common/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -83,4 +83,18 @@
      * @since 1.3
      */
     boolean ignoreByDefault() default false;
+
+    /**
+     * The unmapped source properties that need to be ignored. This should be used when {@link ReportingPolicy#WARN}
+     * or {@link ReportingPolicy#ERROR} is used for {@link Mapper#unmappedSourcePolicy()} or
+     * {@link MapperConfig#unmappedSourcePolicy()}. When properties are set they won't be part of the unmapped source
```
sjaakd: Listed properties will be ignored when composing the unmapped source properties report.

core-common/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -83,4 +83,18 @@
...
+     * properties report. This property has no effect on choosing whether a source property needs to be mapped.
```
sjaakd: I don't understand.. `This properties has.. mapped.`

filiphr:
Thanks for the review @sjaakd I've pushed a modified wording.

filiphr:
Thanks for the review @sjaakd

PR #1450, opened 2018-04-22
#1433 Support for lifecycle methods for builders

### LINKED ISSUE #1433: Access to Builder type in @BeforeMapping and @AfterMapping
We should consider `@BeforeMapping` and `@AfterMapping` for builders as well.

We should add some tests where there are lifecycle callbacks both for the type being built and it's builder to see what is invoked.

filiphr:
When doing mappings with builder the effective type should be considered for lifecycle callback methods. The actual target type won't be considered in the lifecycle callbacks

Fixes #1433

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/LifecycleCallbackFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/LifecycleCallbackFactory.java
@@ -106,11 +106,17 @@ public final class LifecycleCallbackFactory {
         MethodSelectors selectors =
             new MethodSelectors( ctx.getTypeUtils(), ctx.getElementUtils(), ctx.getTypeFactory() );
 
+        Type targetType = method.getResultType();
+
+        if ( !method.isUpdateMethod() ) {
+            targetType = targetType.getEffectiveType();
+        }
+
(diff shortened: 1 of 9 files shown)


agudian:
The code change looks good and it makes sense to support that. 

However, as a user, I would expect that the before methods without the target bean, but the non-builder target type would also be called, and that after methods with the build result (and the result type) are also invoked. That would then allow for using the cycle detection again... that could be tracked ina separate issue, though, but I’m sure people will ask for it at sime point 🙂

filiphr:
Thanks for the review @agudian. Yes I agree that that would be expected. I've created #1454 so we can track that separately.

Can I merge this?

As for lifecycle, it will never be possible. Imagine the following:

```
public class OrderBuilder {

    public OrderBuilder addItem(Item item) {}
    public Order create() { }
}

public class ItemBuilder {

    public ItemBuilder order(Order order) {}
    public Item create() {}
}
```

In order to get ahold of the `Order` you need to have it's reference before `Item` is built. However, in order to build `Item` you need to have an `Order`. Maybe I am not seeing something 😄

agudian:
Haha, you are right with the cycle! 🙃

Go ahead with merging this 👍

PR #1451, opened 2018-04-22
#1449 Adders should not be considered as builder setter methods

### LINKED ISSUE #1449: Adder method not working for builders
Due to the way the `DefaultAccessorNamingStrategy` is implemented adders are not working for builders.

Example:

```java
public class Builder {
    private List<String> names;

    public Builder addName(String name) {
        this.names.add( name );
        return this;
    }
}
```

The `addName` would be considered as a property and not an adder.

I think that the best would be to when checking if a builder is a setter, to check that it is not an adder first.

filiphr:
Fixed #1449

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
+++ b/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
@@ -102,9 +102,22 @@ public class DefaultAccessorNamingStrategy implements AccessorNamingStrategy {
     protected boolean isBuilderSetter(ExecutableElement method) {
         return method.getParameters().size() == 1 &&
             !JAVA_JAVAX_PACKAGE.matcher( method.getEnclosingElement().asType().toString() ).matches() &&
-            !isAdderMethod( method ) &&
+            !isAdderWithUpperCase4thCharacter( method ) &&
             //TODO The Types need to be compared with Types#isSameType(TypeMirror, TypeMirror)
             method.getReturnType().toString().equals( method.getEnclosingElement().asType().toString() );
+    }
+
+    /**
+     * Checks that the method is an adder with an upper case 4th character. The reason for this is that methods such
+     * as {@code address(String address)} are considered as setter and {@code addName(String name)} too. We need to
+     * make sure that {@code addName} is considered as an adder and {@code address} is considered as a setter.
+     *
+     * @param method the method that needs to be checked
+     *
+     * @return {@code true} if the method is an adder with an upper case 4h character, {@code false} otherwise
+     */
+    private boolean isAdderWithUpperCase4thCharacter(ExecutableElement method) {
+        return isAdderMethod( method ) && Character.isUpperCase( method.getSimpleName().toString().charAt( 3 ) );
     }
 
     /**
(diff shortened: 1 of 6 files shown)


filiphr:
This one is a bit more complex than I thought. The current implementation can mix with methods like `address`. Maybe we need to check that it starts with `add` and the 4th character is not a capital letter.

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
```diff
@@ -102,10 +102,24 @@ public boolean isSetterMethod(ExecutableElement method) {
...
         return method.getParameters().size() == 1 &&
             !JAVA_JAVAX_PACKAGE.matcher( method.getEnclosingElement().asType().toString() ).matches() &&
+            !isAdderWithUpperCase4thCharacter( method ) &&
             //TODO The Types need to be compared with Types#isSameType(TypeMirror, TypeMirror)
             method.getReturnType().toString().equals( method.getEnclosingElement().asType().toString() );
     }
 
+    /**
+     * Checks that the method is an adder with an upper case 4th character. The reason for this is that methods such
+     * as {@code address(String address)} are considered as setter and {@code addName(String name)} too. We need to
```
sjaakd: ahh... tricky..

sjaakd: Can't we derive the fact that its an adder by means of this Singular / Plural thing and by means of the fact that there 's also a getter returning a list of what the adder adds? Feels a bit more safe than relying on camelcasing..

processor/src/test/java/org/mapstruct/ap/test/builder/simple/SimpleImmutablePerson.java
```diff
@@ -80,5 +97,19 @@ public Builder city(String city) {
...
+        public Builder address(String address) {
+            this.address = address;
+            return this;
+        }
+
+        public List<String> getChildren() {
+            throw new UnsupportedOperationException( "This is just a marker method" );
+        }
+
+        public Builder addChild(String child) {
```
sjaakd: have a look at the class `Nouns`.. that should do the trick of preferring `addChild` over `address`..

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
filiphr: That is in the internal packages and this is in the public package. I don't think that it is a good idea to expose that here.

processor/src/test/java/org/mapstruct/ap/test/builder/simple/SimpleImmutablePerson.java
filiphr: I know about it, but that is from the internal packages.

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
sjaakd: True.. We should not expose it directly.. But we could expose perhaps expose it as util (hiding the implementation).. WDYT?

filiphr: I disagree. The accessor naming strategy decides based only on a method whether it can be a getter, setter, adder or something else. It doesn't know whether a getter (setter) for the given adder exists.

filiphr:
Thanks for the review @sjaakd

PR #1461, opened 2018-04-27
#1460 Ensures the FQN will be used for SimpleConversion if required

### LINKED ISSUE #1460: String to enum conversion: Compile error in generated mapper
Today we found a new bug in MapStruct 1.2.0:

We had a bean with a String property which was mapped to a different bean where the same property is an enum.
String to enum is totally fine, but: The enum (E1) had the same class name as a different enum (E2) that was referenced in the same mapper.

The generated class imported the E2, but is using the simple name of E1 in the String to E1 mapping.

A bit complicated to describe, but I already analyzed this and will provide a PR as soon as possible. 

I guess it is the correct way to first create an issue and then a PR with the fix or would it be okay to do this without an issue?

Best regards,
Christian

chris922:
* Added getReferenceName() to Type which returns simple or fully-qualified-name
* Use getReferenceName() in all SimpleConversions
* Added testcase that is not working without these changes

The issue itself occurs as soon as there is the same class name used for the classes in the `SimpleConversions` and in the mapper (e. g. an own `LocalDate` class and a `LocalDate` conversion). 
I added a testcase that shows this for a few types., but I changed a lot more `SimpleConversions` and checked my changes multiple times. 
Please also take a look..  hopefully I haven't missed things like a `(` or an import. These were the issues I noticed after I added the changes, but in case all conversions are covered by at least one test everything should still work.

I am also not very happy with the name `getReferenceName()`, maybe you've got a better idea. But what it returns is the name that could be used to reference to this type .. so maybe it's okay.

Looking forward to your feedback!

Best regards,
Christian

PS: I directly used the java8 time classes in some conversions. Don't know if this is legit, maybe I need to change this to using the FQN as a `String` instead of the `Class`. But if this is the case I expect that Travis complains this.. 

Fixes #1460

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractJavaTimeToStringConversion.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractJavaTimeToStringConversion.java
@@ -52,10 +52,12 @@ public abstract class AbstractJavaTimeToStringConversion extends SimpleConversio
 
     private String dateTimeFormatter(ConversionContext conversionContext) {
         if ( !Strings.isEmpty( conversionContext.getDateFormat() ) ) {
-            return "DateTimeFormatter.ofPattern( \"" + conversionContext.getDateFormat() + "\" )";
+            return dateTimeFormatterType( conversionContext )
+                + ".ofPattern( \"" + conversionContext.getDateFormat()
+                + "\" )";
         }
         else {
-            return "DateTimeFormatter." + defaultFormatterSuffix();
+            return dateTimeFormatterType( conversionContext ) + "." + defaultFormatterSuffix();
         }
     }
 
@@ -64,7 +66,7 @@ public abstract class AbstractJavaTimeToStringConversion extends SimpleConversio
     @Override
     protected String getFromExpression(ConversionContext conversionContext) {
         // See http://docs.oracle.com/javase/tutorial/datetime/iso/format.html for how to parse Dates
-        return new StringBuilder().append( conversionContext.getTargetType().getFullyQualifiedName() )
+        return new StringBuilder().append( conversionContext.getTargetType().getReferenceName() )
                                   .append( ".parse( " )
                                   .append( parametersListForParsing( conversionContext ) )
                                   .append( " )" ).toString();
@@ -74,9 +76,8 @@ public abstract class AbstractJavaTimeToStringConversion extends SimpleConversio
         // See http://docs.oracle.com/javase/tutorial/datetime/iso/format.html for how to format Dates
         StringBuilder parameterBuilder = new StringBuilder( "<SOURCE>" );
         if ( !Strings.isEmpty( conversionContext.getDateFormat() ) ) {
-            parameterBuilder.append( ", DateTimeFormatter.ofPattern( \"" )
-                            .append( conversionContext.getDateFormat() )
-                            .append( "\" )" );
+            parameterBuilder.append( ", " );
+            parameterBuilder.append( dateTimeFormatter( conversionContext ) );
         }
         return parameterBuilder.toString();
     }
@@ -84,14 +85,25 @@ public abstract class AbstractJavaTimeToStringConversion extends SimpleConversio
     @Override
     protected Set<Type> getToConversionImportTypes(ConversionContext conversionContext) {
         return Collections.asSet(
-                        conversionContext.getTypeFactory().getType( JavaTimeConstants.DATE_TIME_FORMATTER_FQN )
+            conversionContext.getTypeFactory().getType( JavaTimeConstants.DATE_TIME_FORMATTER_FQN )
         );
     }
 
     @Override
     protected Set<Type> getFromConversionImportTypes(ConversionContext conversionContext) {
-        return Collections.asSet(
-                        conversionContext.getTypeFactory().getType( JavaTimeConstants.DATE_TIME_FORMATTER_FQN )
-        );
+        if ( !Strings.isEmpty( conversionContext.getDateFormat() ) ) {
+            return Collections.asSet(
+                conversionContext.getTargetType(),
+                conversionContext.getTypeFactory().getType( JavaTimeConstants.DATE_TIME_FORMATTER_FQN )
+            );
+        }
+
+        return Collections.asSet( conversionContext.getTargetType() );
+    }
+
+    protected String dateTimeFormatterType(ConversionContext conversionContext) {
+        return conversionContext.getTypeFactory()
+            .getType( JavaTimeConstants.DATE_TIME_FORMATTER_FQN )
+            .getReferenceName();
     }
 }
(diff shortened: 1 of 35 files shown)


processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractJavaTimeToStringConversion.java
```diff
@@ -74,24 +76,34 @@ private String parametersListForParsing(ConversionContext conversionContext) {
         // See http://docs.oracle.com/javase/tutorial/datetime/iso/format.html for how to format Dates
         StringBuilder parameterBuilder = new StringBuilder( "<SOURCE>" );
         if ( !Strings.isEmpty( conversionContext.getDateFormat() ) ) {
-            parameterBuilder.append( ", DateTimeFormatter.ofPattern( \"" )
-                            .append( conversionContext.getDateFormat() )
-                            .append( "\" )" );
+            parameterBuilder.append( ", " );
+            parameterBuilder.append( dateTimeFormatter( conversionContext ) );
```
filiphr: Isn't

```
.append( conversionContext.getDateFormat() )
.append( "\" )" );
```

missing here?

processor/src/main/java/org/mapstruct/ap/internal/conversion/BigDecimalToPrimitiveConversion.java
```diff
@@ -50,11 +50,15 @@ public String getToExpression(ConversionContext conversionContext) {
...
-        return "BigDecimal.valueOf( <SOURCE> )";
+        return bigDecimal( conversionContext ) + ".valueOf( <SOURCE> )";
     }
 
     @Override
     protected Set<Type> getFromConversionImportTypes(ConversionContext conversionContext) {
         return asSet( conversionContext.getTypeFactory().getType( BigDecimal.class ) );
     }
+
+    private String bigDecimal(ConversionContext conversionContext) {
```
filiphr: This function is the same on 5 different places. Maybe can move it to the parent? Or have a util?

processor/src/main/java/org/mapstruct/ap/internal/conversion/BigIntegerToPrimitiveConversion.java
```diff
@@ -54,11 +54,15 @@ public String getFromExpression(ConversionContext conversionContext) {
...
-        return "BigInteger.valueOf( " + castString + "<SOURCE> )";
+        return bigInteger( conversionContext ) + ".valueOf( " + castString + "<SOURCE> )";
     }
 
     @Override
     protected Set<Type> getFromConversionImportTypes(ConversionContext conversionContext) {
         return asSet( conversionContext.getTypeFactory().getType( BigInteger.class ) );
     }
+
+    private String bigInteger(ConversionContext conversionContext) {
```
filiphr: same as the bigDecimal function. This one is in 3 places

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaLocalDateTimeToDateConversion.java
```diff
@@ -28,11 +36,49 @@
...
+
+    @Override
+    protected Set<Type> getFromConversionImportTypes(ConversionContext conversionContext) {
+        return Collections.asSet(
+            conversionContext.getTypeFactory().getType( LocalDateTime.class ),
+            conversionContext.getTypeFactory().getType( ZoneId.class )
+        );
+    }
+
+    private String date(ConversionContext conversionContext) {
```
filiphr: The functions here could also be reused

processor/src/main/java/org/mapstruct/ap/internal/conversion/BigDecimalToPrimitiveConversion.java
chris922: Yes I thought the same but wasn't sure where to put.. I found a util package, maybe create a util class with static methods?

processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractJavaTimeToStringConversion.java
chris922: Looks like it is missing, but the `dateTimeFormatter` method (already existing one) is adding this. Naming of the method is confusing as I always named the methods to get the reference name in this way.

processor/src/main/java/org/mapstruct/ap/internal/conversion/BigDecimalToPrimitiveConversion.java
filiphr: Yes you can put them there. `ConversionUtils` maybe?

chris922: Ok great, I will move all those helper methods that are spread over the `SimpleConversion` classes to such a new class.

processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractJavaTimeToStringConversion.java
filiphr: Yes you are right sorry. This should have used that method from before.

chris922:
I just pushed changes based on your feedback, @filiphr.

Just let me know what still needs to be changed.

processor/src/main/java/org/mapstruct/ap/internal/conversion/BigIntegerToStringConversion.java
```diff
@@ -57,22 +60,21 @@ public String getToExpression(ConversionContext conversionContext) {
...
-            return "new BigInteger( <SOURCE> )";
+            return "new " + bigInteger( conversionContext ) + "( <SOURCE> )";
         }
     }
 
     @Override
     protected Set<Type> getFromConversionImportTypes(ConversionContext conversionContext) {
         if ( requiresDecimalFormat( conversionContext ) ) {
-            // no imports are required when decimal format is used.
-            return super.getFromConversionImportTypes( conversionContext );
```
filiphr: Shouldn't we get the one from the parent too?

chris922: Hmm, you mean that `super.getFromConversionImportTypes` should be called?
I don't think this is required as `getFromExpression` isn't using the `DecimalFormat` that would be returned from `super.getFromConversionImportTypes`. The `appendDecimalFormatter` method is adding a call to a helper method that has it's own import types.

filiphr: And again you are right 😄. Didn't analyze it that deep 😄

chris922: But it was definitly a good point. I also had to go through the code again and was wondering why I removed the `super` call.

filiphr:
This has now been integrated. Thanks a lot for the PR @chris922

PR #1465, opened 2018-04-28
#1453 Make sure that we always create non bound collection

### LINKED ISSUE #1453: Compilation error on CollectionMapping. Wrong capture generated
Hi,
```
protected Collection<? extends PaymentDTO> paymentListToPaymentDTOCollection(List<Payment> list) {
        if ( list == null ) {
            return null;
        }
        // Capture cause compilation error, should be "Collection<PaymentDTO>"
        Collection<? extends PaymentDTO> collection = new ArrayList<PaymentDTO>( list.size() );
        for ( Payment payment : list ) {
            collection.add( toDto( payment ) ); // <== here
        }

        return collection;
    }

```

error
```
error: method add in interface Collection<E> cannot be applied to given types;
            collection.add( toDto( payment ) );
                      ^
  required: CAP#1
  found: PaymentDTO
  reason: argument mismatch; PaymentDTO cannot be converted to CAP#1
  where E is a type-variable:
    E extends Object declared in interface Collection
  where CAP#1 is a fresh type-variable:
    CAP#1 extends PaymentDTO from capture of ? extends PaymentDTO
1 error
```

toDto mapping method is just
```
    @Mapping(target = "term",                        ignore = true)
    PaymentDTO toDto(Payment payment);
```
which generates
```
    @Override
    public PaymentDTO toDto(Payment payment) {
        if ( payment == null ) {
            return null;
        }

        PaymentDTOBuilder paymentDTO = PaymentDTO.builder();

        paymentDTO.type( payment.getType() ); // I am using getters with getX for Intellij to advise, since plugin can't see fluent getters , later I just uncomment @Accessors(fluent=true)
        paymentDTO.amount( payment.getAmount() );
        paymentDTO.description( payment.getDescription() );

        return paymentDTO.build();
    }
```

The Solution for now is adder method and ADDER_PREFFERED which works fine.

filiphr:
When generating collection mapping methods make sure that the created variable is from the bound type

Fixes #1453

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_1453/Auction.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_1453/Auction.java
@@ -1,0 +1,62 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.bugs._1453;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author Filip Hrisafov
+ */
+public class Auction {
+
+    private final List<Payment> payments;
+    private final List<Payment> otherPayments;
+    private Map<Payment, Payment> mapPayments;
+    private Map<Payment, Payment> mapSuperPayments;
+
+    public Auction(List<Payment> payments, List<Payment> otherPayments) {
+        this.payments = payments;
+        this.otherPayments = otherPayments;
+    }
+
+    public List<Payment> getPayments() {
+        return payments;
+    }
+
+    public List<Payment> getOtherPayments() {
+        return otherPayments;
+    }
+
+    public Map<Payment, Payment> getMapPayments() {
+        return mapPayments;
+    }
+
+    public void setMapPayments(Map<Payment, Payment> mapPayments) {
+        this.mapPayments = mapPayments;
+    }
+
+    public Map<Payment, Payment> getMapSuperPayments() {
+        return mapSuperPayments;
+    }
+
+    public void setMapSuperPayments(Map<Payment, Payment> mapSuperPayments) {
+        this.mapSuperPayments = mapSuperPayments;
+    }
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -598,6 +598,7 @@ public class PropertyMapping extends ModelElement {
         private Assignment forgeWithElementMapping(Type sourceType, Type targetType, SourceRHS source,
             ExecutableElement element, ContainerMappingMethodBuilder<?, ? extends ContainerMappingMethod> builder) {
 
+            targetType = targetType.withoutBounds();
             ForgedMethod methodRef = prepareForgedMethod( sourceType, targetType, source, element, "[]" );
 
             ContainerMappingMethod iterableMappingMethod = builder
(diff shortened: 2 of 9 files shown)


processor/src/main/resources/org/mapstruct/ap/internal/model/IterableMappingMethod.ftl
```diff
@@ -121,7 +121,7 @@
         <#if resultType.fullyQualifiedName == "java.lang.Iterable">
             <@includeModel object=resultType.implementationType/>
         <#else>
-            <@includeModel object=resultType/>
+            <@includeModel object=resultType withoutBounds=true/>
```
sjaakd: I'm not sure this is a good idea.. We are generating the methods ourselves because of nesting right? Should we even generate a method with a wildcard return type? See [here](http://www.angelikalanger.com/GenericsFAQ/FAQSections/ProgrammingIdioms.html#Should%20I%20use%20wildcards%20in%20the%20return%20type%20of%20a%20method?) for some details.. Just google "avoid return type wildcard". Of course.. we probably need to do something like this if the user decides to define a mapping method returning a wildcard.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1453/Auction.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final List<Payment> payments;
+    private final List<Payment> otherPayments;
+
+    public Auction(List<Payment> payments, List<Payment> otherPayments) {
+        this.payments = payments;
+        this.otherPayments = otherPayments;
+    }
+
+    public List<Payment> getPayments() {
```
sjaakd: What about Maps? Would we have a similar case there? Or even other parameterised types..?

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableMappingMethod.ftl
filiphr: Don't have access to that. In any case we are delivering `Collection<? extends MyType>`. If you all at [`Collection#addAll(Collection)`](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#addAll-java.util.Collection-) it accepts `Collection<? extends E>` that is what the setter does actually too internally

sjaakd: Fixed the link.. Perhaps I'm wrong.. I'll try to make an example

sjaakd:
Hi Filip, 

Try this:
```java
public class NewClass {

        List<MyBase> baseCollection = new ArrayList<>();
        MyBase base = baseCollection.get( 0 );
        add1( baseCollection );
        add2( baseCollection );
        List<MyExt1> ext1Collection = new ArrayList<>();
        MyExt1 ext1 = ext1Collection.get( 0 );
        add1( ext1Collection );
        add2( ext1Collection ); // error.. MyBase only has two supertypes: Object and MyBase itself.

    }

    public static void add1(List<? extends MyBase> list ) {        
    }
    
    public static void add2(List<? super MyBase> list ) {
    }
    
    public class  MyBase {
    }
    
    public static class MyExt1 extends MyBase {        
    }
    
    public static class MyExt2 extends MyBase {     
    }
}
```
You should be fine with the `public setPayment(Collection<? extends Payment> payments)`, adding `Collection<Payment>` according to the above example. The same goes for `public setPayment(Collection<? super Payment> payments)`. `Payment` itself belongs both, to the 'family' of objects that is both a super of `Payment` and is a sub of `Payment`

The [FAC](http://www.angelikalanger.com/GenericsFAQ/FAQSections/ProgrammingIdioms.html) on generics is my bible when it comes to generics :smile:. 

I personally think this stuff is tricky. So when using wild card types defined types (so not calling method argument like above, but handling it in the body of this method), this [SO](https://stackoverflow.com/questions/4343202/difference-between-super-t-and-extends-t-in-java) answer is also really helpful. The *PECS* acronym helps me to remember.

filiphr:
@sjaakd you are correct. However, you are mixing the types. We would never be in such situation. We are always generating a method that would map to the one defined in the setter. Which means that our extension would be `? extends MyBase` or `? super MyBase`. This would lead to us creating an inner collection of the bound so `MyBase`.

What do you think I need to change in this PR?

filiphr:
>However, my example was meant to demonstrate that our return type for forged methods which does not need to follow the exact `? extends MyBase` or `? super MyBase`

Now it makes a lot more sense to me 😄. You are completely right, maybe we need to create a method which would return just the type bound. That way we don't even need the stuff I did in this PR. I will try that later this week in his PR

sjaakd:
> Now it makes a lot more sense to me 😄. You are completely right, maybe we need to create a method which would return just the type bound. That way we don't even need the stuff I did in this PR. I will try that later this week in his PR

Exactly.. but only for the forged method.. For implemented methods your solution will still be valid..

filiphr:
@sjaakd I've updated the PR.  I scrubbed my previous work. It seems doing it only for forged methods is enough. We actually don't allow implementing a method whose return type has an extends bound (and those are the ones that make problems). I've added a fixture as well so we can see how it looks like.

filiphr:
>We can't forbid a user to define a mapping method with extends and super bounds in the return type.

Well we are actually doing that for extends. See [this](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java#L116).

And for super there is no problem as this is valid:

```java
List<? super AuctionDto> list = new ArrayList<AuctionDto>( auctions.size() );
for ( Auction auction : auctions ) {
    list.add( map( auction ) );
}
```

See the fixture in the PR as well.

filiphr:
It seems has been done as part of #527 from you 😉

sjaakd:
> It seems has been done as part of #527 from you 😉

Hmm. my memory is failing me here.. must be my age 😢 . I have no further comments :smile:

filiphr:
Thanks for the review @sjaakd

Issue #1458, opened 2018-04-25
Introduce mechanism to report error from MappingResolver


sjaakd:
Is needed for

- [x] The solution for #1401
- [ ] The solution for #5

filiphr:
@sjaakd now that we merged #1466 should we close this?

sjaakd:
> @sjaakd now that we merged #1466 should we close this?

No.. #5 also requires such mechanism.. its not fully implemented yet.

filiphr:
>No.. #5 also requires such mechanism.. its not fully implemented yet.

Good was not sure. I updated the the original message with checkpoints so we can track it better 😄

sjaakd:
done.. position hint is available in `MapperResolverImpl`, partial message as well. Used in #5

PR #1466, opened 2018-04-29
#1458 adding detailed messages to constant mapping failures

sjaakd:
Although `NativeTypes` and `NativeTypesTest` are changed, the `Analysers` are more or less the same. I abandoned the concept of in-line class definition for inner classes so I can use them twice (once for boxed, once for primitive). This simplifies the solution greatly and made it possible to bring the exception outward.

Furthermore, the terminology is aligned. Not constants, but literals.


processor/src/test/java/org/mapstruct/ap/test/source/constants/ErroneousMapper7.java
```diff
@@ -0,0 +1,48 @@
...
+import org.mapstruct.Mapping;
+import org.mapstruct.Mappings;
+import org.mapstruct.factory.Mappers;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+@Mapper
+public interface ErroneousMapper7 {
```
twentylemon: `ErroneousMapper7` -> `ErroneonsConstantsMapper` ?

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -760,12 +761,22 @@ public ConstantMappingBuilder selectionParameters(SelectionParameters selectionP
...
-                // convert to string
+            Class<?> baseForLiteral = null;
+            try {
+                baseForLiteral = NativeTypes.getLiteral( targetType.getFullyQualifiedName(), constantExpression );
+            }
+            catch ( IllegalArgumentException ex ) {
+                errorMessageDetails = ex.getMessage();
+            }
+
+            // asume a String constant
```
twentylemon: typo, asume -> assume
Not sure of mapstructs standards, but when I write a comment, I try to explain stuff in some detail. The comment exists, it should justify itself. Here, I'd probably write something like:
`// the constant is not a primitive literal, assume it to be a String`

processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java
```diff
@@ -233,6 +221,185 @@ private boolean floatHasBecomeZero() {
...
+            NumberRepresentation br = new NumberRepresentation( s, false, false, false ) {
+
+                @Override
+                void parse(String val, int radix) {
+                    Double d = Double.parseDouble( radix == 16 ? "0x" + val : val );
+                    if ( doubleHasBecomeZero( d  ) ) {
+                        throw new NumberFormatException( "floating point number too small" );
+                    }
+                    if ( d.isInfinite() ) {
+                        throw new NumberFormatException( "infinitive is not allowed" );
```
twentylemon: This is odd. On one hand, the user needs to specify a number as a string, not `Double.POSITIVE_INFINITY`, so this is a logical restriction to have. But I can definitely imagine cases where "infinity" is a reasonable default/constant value to use (say, as a sentinel value).

Perhaps both `0` and infinite should be treated as warnings and not errors?

processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java
```diff
@@ -233,6 +221,185 @@ private boolean floatHasBecomeZero() {
...
+
+        @Override
+        public void validate(String s) {
+            NumberRepresentation br = new NumberRepresentation( s, false, false, false ) {
+
+                @Override
+                void parse(String val, int radix) {
+                    Double d = Double.parseDouble( radix == 16 ? "0x" + val : val );
+                    if ( doubleHasBecomeZero( d  ) ) {
+                        throw new NumberFormatException( "floating point number too small" );
```
twentylemon: Does this disallow `@Mapping(constant = "0.0")` ?

processor/src/test/java/org/mapstruct/ap/test/source/constants/ErroneousMapper7.java
sjaakd: Better.. I'll change.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: I'll adapt.

processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java
sjaakd: Since we have full freedom, we could introduce a magic string: e.g. 'INFINITIVE'. I'm not in favour of having unintentional infinitive. That's why I decided to have a mapping error.

processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java
sjaakd: It should. I'll add a test case for that one..

sjaakd:
@twentylemon : thanks for reviewing. If I find some time tis week, I'll update the PR.

processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java
twentylemon: I agree that unintentional infinity should be detected, along with unintentional `0`. It's easy to imagine that causing a lot of pain, especially when it's hidden in generated code. I just worry it's too restrictive in some cases; plus it is not be backward compatible with current behaviour.

Magic strings is a good idea. With good errors ("did you mean to use ZERO?"), you don't really need to go to the docs each time you want a zero or infinite value. These also aren't backward compatible, but good error messages make for a simple upgrade path at least.

sjaakd: I think we should handel zero the normal way 😄 . However, we might want to borrow some keywords from the `Double` and `Float` definition. MapStruct would assign them automatically. I need to think though if I solve it in this PR or in a separate one. We can refer to those in the javadoc. We could refer to those in (the proper) error messages as you suggest.

Natural candidates would be (from `Float` and `Double` definition):  `POSITIVE_INFINITY`,`NEGATIVE_INFINITY`,`NaN`,`MAX_VALUE `, `MIN_NORMAL`, `MIN_VALUE`

twentylemon: Re-read... `NEGATIVE_MAX_VALUE` is less ambiguous.

twentylemon: Sounds good to me.

processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java
```diff
@@ -233,6 +221,185 @@ private boolean floatHasBecomeZero() {
...
+        public void validate(String s) {
+            NumberRepresentation br = new NumberRepresentation( s, true, false, false ) {
+
+                @Override
+                void parse(String val, int radix) {
+                    if ( radix == 10 ) {
+                        // when decimal: treat like signed
+                        Integer.parseInt( val, radix );
+                    }
+                    else {
```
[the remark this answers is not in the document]
sjaakd: I'll fix it..

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -822,17 +833,30 @@ public PropertyMapping build() {
                 }
             }
             else {
```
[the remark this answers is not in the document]
sjaakd: ok

processor/src/test/java/org/mapstruct/ap/test/source/constants/ConstantsMapper.java
```diff
@@ -29,11 +29,15 @@
...
-public interface NumericMapper {
+public interface ConstantsMapper {
 
-    NumericMapper INSTANCE = Mappers.getMapper( NumericMapper.class );
+    ConstantsMapper INSTANCE = Mappers.getMapper( ConstantsMapper.class );
 
     @Mappings({
+        @Mapping(target = "booleanValue", constant = "true"),
+        @Mapping(target = "booleanBoxed", constant = "false"),
+        @Mapping(target = "charValue", constant = "'b'"),
```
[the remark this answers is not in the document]
sjaakd: well, I made it this way because, you can't confuse it with a string constant and its in line with integer/float assignments. After all, you can't do `setSomething( b )` or `something = b` as you can `setSomething( 5 )'`

processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java
```diff
@@ -233,6 +221,185 @@ private boolean floatHasBecomeZero() {
...
+        public void validate(String s) {
+            NumberRepresentation br = new NumberRepresentation( s, true, true, false ) {
+
+                @Override
+                void parse(String val, int radix) {
+                    if ( radix == 10 ) {
+                        // when decimal: treat like signed
+                        Long.parseLong( val, radix );
+                    }
+                    else {
```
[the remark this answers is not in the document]
sjaakd: wil fix.

processor/src/test/java/org/mapstruct/ap/test/source/constants/ConstantsMapper.java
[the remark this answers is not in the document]
filiphr: Makes sense

sjaakd:
@filiphr .. thanks for your review.. I processed the comments

PR #1475, opened 2018-05-03
#1338 Using an adder with non generic source collection should work

filiphr:
Another fix for #1338

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
@@ -19,6 +19,7 @@
 package org.mapstruct.ap.internal.model.assignment;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
```diff
@@ -44,6 +48,7 @@ public AdderWrapper( Assignment rhs,
         this.thrownTypesToExclude = thrownTypesToExclude;
         String desiredName = Nouns.singularize( targetPropertyName );
         rhs.setSourceLocalVarName( rhs.createLocalVarName( desiredName ) );
+        adderType = first( getSourceType().determineTypeArguments( Collection.class ) );
```
sjaakd: and `first` yields null if there is no type-bound..? Think that should be a comment here.

filiphr:
Yes we did. However, only in a single direction (when we were looking for an adder). This problem occurs when we try to use an adder wrapper that should use the element as a source type for matching.

filiphr:
Thanks for the review @sjaakd

PR #1480, opened 2018-05-05
Use constants instead of strings in ValueMapping javadoc

chris922:
I noticed that the String values were used in `@ValueMapping` javadoc and as there are constants available I think it is always better to just use those.


filiphr:
Good catch @chris922. Thanks a lot for the PR

Issue #537, opened 2015-04-23
"Ambiguous mapping methods found" when the same method is found twice


agudian:
We have a used ReferenceMapper defined in an `@MapperConfig` annotation and also in the `@Mapper` annotation itself.

This leads to the following error:

```
Ambiguous mapping methods found for mapping property "foo.bar.Source source" to foo.bar.Target:
foo.bar.Target foo.bar.ReferenceMapper.toTarget(foo.bar.Source source),
foo.bar.Target foo.bar.ReferenceMapper.toTarget(foo.bar.Source source).
```

gunnarmorling:
Interesting one :) I think we could do that in a 1.0.1 as there is a work-around (remove one of the references).

DarkmanMuMoo:
+1  
I really want this , at least provide some config that i can specified mapping method  

or pick it base on  method modifier  

thx

gunnarmorling:
Yes, we should eliminate the duplicates in this case. Maybe you'd be
willing to help out? Should be an easy fix.

You should be able to work around this issue easily by just removing the
referenced mapper from @Mapper#uses().

chris922:
Found this old issue and looks like it was solved in the meantime or my test-case is not correct.

I just pushed the test case. If you still know what exactly the issue was/is you may could check the test case I created.

filiphr:
@chris922 Looking at the test case it seems that it is correct and it seems it has been fixed somehow 😄. In any case you can provide a PR with the test, having more tests is always good 😉.

Just some small info. Classes in `Mapper#uses` can be non `@Mapper` classes. It can also be a normal class that (in which case MapStruct will instantiate by invoking its default constructor) or if it has static methods just invoke that.

And for the `@{jacocoArgLine}` if the reason is IntelliJ, I think that you can explicitly tell IntelliJ to not pick the surefire `argLine`

chris922:
Ok great, later on I will create a PR with this test-case.

It was not planned to commit the removal of the `@{jacocoArgLine}`. But good to know that there is somewhere in IntelliJ the setting to disable this. It was a bit annoying to remove it always and take care it will not be committed..
But I will clean this up before creating the PR.

filiphr:
Closing as we now have a test case for this and it seems it can been fixed somehow as well. I suspected that it was by https://github.com/mapstruct/mapstruct/issues/634.

Thanks @chris922 for providing the PR.

>But I will clean this up before creating the PR.

[...]

PR #1486, opened 2018-05-09
#537 ambiguous mapping when same method twice

chris922:
Issue #537 was most likely fixed in the meantime, but here is a test-case for the reported issue to improve the test-suite.


filiphr:
Thanks for the PR @chris922. I had a look at the code and I think the fix for https://github.com/mapstruct/mapstruct/issues/634 fixed this. In any case having a test case is always good :)

Issue #1454, opened 2018-04-23
Support for lifecycle methods on type being build with builders


filiphr:
Currently we only support lifecycle methods on for a single type.

This issue is to expand that so the lifecycles are invoked for the type being build.

Example. We have `Person` with `PersonBuilder`.

The following lifecycle methods are / should be invoked:

- [x] `@BeforeMapping` with no parameters
- [ ] `@BeforeMapping` with `@TargetType` `Person`
- [x] `@BeforeMapping` with `@TargetType` `PersonBuilder`
- [x] `@BeforeMapping` with `@MappingTarget` `PersonBuilder`
- [x] `@AfterMapping` with no parameters
- [ ] `@AfterMapping` with `@TargetType` `Person`
- [x] `@AfterMapping` with `@TargetType` `PersonBuilder`
- [ ] `@AfterMapping` with `@MappingTarget` `Person`
- [x] `@AfterMapping` with `@MappingTarget` `PersonBuilder`
- [x] `@AfterMapping` with `@MappingTarget` `PersonBuilder` and returning `Person`

andrei-ivanov:
Any chance to get this one completed? 😩

PR #1494, opened 2018-05-15
#1454 add more tests for builder lifecycle methods

chris922:
Added tests for following lifecycle methods mentioned in #1454

* `@BeforeMapping` with no parameters
* `@AfterMapping` with no parameters
* `@AfterMapping` with `@MappingTarget` `PersonBuilder` and returning `Person`

Last entry is not yet checked in #1454, but works

Generated code:

```java
Order target = context.afterWithBuilderTargetReturningTarget( order );
if ( target != null ) {
    return target;
}
```


filiphr:
Thanks for the tests @chris922

PR #1498, opened 2018-05-20
#1479 Add support for Builders with multiple build methods

### LINKED ISSUE #1479: Wrong builder method is selected when Protobuf is used
Protobuf has mutliple public parameterless methods that create the type in the builder:

* `getDefaultInstanceForType`
* `build`
* `buildPartial`

Currently we pick `getDefaultInstanceForType` without any error. This is clearly a mistake and we should either:

* not pick any and error out so the user can define a custom SPI that would pick the correct
* provide a custom SPI for Protobuf that would pick the `build` method.

An example code can be found [here](https://github.com/pascalwhoop/grpc-adapter). The `CompetitionMapper` can be adjusted to map into `PBCompetition` instead of `PBCompetition.Builder`

filiphr:
* When there are multiple build methods look for a method named `build`, if such method exists use it
* When a builder has multiple methods throw an exception and don't use the builder during the mapping
* When a type has multiple builder creation methods throw an exception and don't use the builder

Defaulting to a method named `build` will make sure that a correct method is slected for:
* FreeBuilder - it has two methods: `build` and `buildPartial`
* Protobuf - it has three methods: `getDefaultInstanceForType`, `build` and `buildPartial`

Fixes #1479

### DIFF
--- a/core-common/src/main/java/org/mapstruct/Builder.java
+++ b/core-common/src/main/java/org/mapstruct/Builder.java
@@ -1,0 +1,45 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.mapstruct.util.Experimental;
+
+/**
+ * Configuration of builders, e.g. the name of the final build method.
+ *
+ * @author Filip Hrisafov
+ *
+ * @since 1.3
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({})
+@Experimental
+public @interface Builder {
+
+    /**
+     * The name of the build method that needs to be invoked on the builder to create the type being build
+     *
+     * @return the method that needs to tbe invoked on the builder
+     */
+    String buildMethod() default "build";
+}
--- a/core-common/src/main/java/org/mapstruct/Mapper.java
+++ b/core-common/src/main/java/org/mapstruct/Mapper.java
@@ -18,14 +18,14 @@
  */
 package org.mapstruct;
 
-import static org.mapstruct.NullValueCheckStrategy.ON_IMPLICIT_CONVERSION;
-
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 import org.mapstruct.factory.Mappers;
+
+import static org.mapstruct.NullValueCheckStrategy.ON_IMPLICIT_CONVERSION;
 
 /**
  * Marks an interface or abstract class as a mapper and activates the generation of a implementation of that type via
@@ -200,4 +200,22 @@ public @interface Mapper {
      */
     boolean disableSubMappingMethodsGeneration() default false;
 
+    /**
+     * The information that should be used for the builder mappings. This can be used to define custom build methods
+     * for the builder strategy that one uses.
+     *
+     * If no builder is defined the builder given via {@link MapperConfig#builder()} will be applied.
+     *
+     * <p>
+     * NOTE: In case no builder is defined here, in {@link BeanMapping} or {@link MapperConfig} and there is a single
+     * build method, then that method would be used.
+     * <p>
+     * If the builder is defined and there is a single method that does not match the name of the finisher than
+     * a compile error will occurs
+     *
+     * @return the builder information
+     *
+     * @since 1.3
+     */
+    Builder builder() default @Builder;
 }
--- a/core-common/src/main/java/org/mapstruct/MapperConfig.java
+++ b/core-common/src/main/java/org/mapstruct/MapperConfig.java
@@ -18,14 +18,14 @@
  */
 package org.mapstruct;
 
-import static org.mapstruct.NullValueCheckStrategy.ON_IMPLICIT_CONVERSION;
-
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 import org.mapstruct.factory.Mappers;
+
+import static org.mapstruct.NullValueCheckStrategy.ON_IMPLICIT_CONVERSION;
 
 /**
  * Marks a class or interface as configuration source for generated mappers. This allows to share common configurations
@@ -186,4 +186,24 @@ public @interface MapperConfig {
      * @since 1.2
      */
     boolean disableSubMappingMethodsGeneration() default false;
+
+    /**
+     * The information that should be used for the builder mappings. This can be used to define custom build methods
+     * for the builder strategy that one uses.
+     *
+     * <p>
+     * Can be overridden by {@link MapperConfig#builder()}.
+     *
+     * <p>
+     * NOTE: In case no builder is defined here, in {@link BeanMapping} or {@link Mapper} and there is a single
+     * build method, then that method would be used.
+     * <p>
+     * If the builder is defined and there is a single method that does not match the name of the finisher than
+     * a compile error will occurs
+     *
+     * @return the builder information
+     *
+     * @since 1.3
+     */
+    Builder builder() default @Builder;
 }
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -639,6 +639,12 @@ The default implementation of the `BuilderProvider` assumes the following:
 So for example `Person` has a public static method that returns `PersonBuilder`.
 * The builder type has a parameterless public method (build method) that returns the type being build
 In our example `PersonBuilder` has a method returning `Person`.
+* In case there are multiple build methods, MapStruct will look for a method called `build` if such methods exists
+than this would be used, otherwise a compilation error would be created.
+* A specific build method can be defined by using `@Builder` within: `@BeanMapping`, `@Mapper` or `@MapperConfig`
+* In case there are multiple builder creation methods that satisfy the above conditions then a `MoreThanOneBuilderCreationMethodException`
+will be thrown from the `DefaultBuilderProvider` SPI.
+In case of a `MoreThanOneBuilderCreationMethodException` MapStruct will write a warning in the compilation and not use any builder.
 
 If such type is found then MapStruct will use that type to perform the mapping to (i.e. it will look for setters into that type).
 To finish the mapping MapStruct generates code that will invoke the build method of the builder.
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BuilderFinisherMethodResolver.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BuilderFinisherMethodResolver.java
@@ -1,0 +1,101 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model;
+
+import java.util.Collection;
+import javax.lang.model.element.ExecutableElement;
+
+import org.mapstruct.ap.internal.model.common.BuilderType;
+import org.mapstruct.ap.internal.model.source.BeanMapping;
+import org.mapstruct.ap.internal.model.source.Method;
+import org.mapstruct.ap.internal.prism.BuilderPrism;
+import org.mapstruct.ap.internal.util.MapperConfiguration;
+import org.mapstruct.ap.internal.util.Message;
+import org.mapstruct.ap.internal.util.Strings;
+
+import static org.mapstruct.ap.internal.util.Collections.first;
+
+/**
+ * @author Filip Hrisafov
+ */
+public class BuilderFinisherMethodResolver {
+
+    private static final String DEFAULT_BUILD_METHOD_NAME = "build";
+
+    private BuilderFinisherMethodResolver() {
+    }
+
+    public static MethodReference getBuilderFinisherMethod(Method method, BuilderType builderType,
+        MappingBuilderContext ctx) {
+        Collection<ExecutableElement> buildMethods = builderType.getBuildMethods();
+        if ( buildMethods.isEmpty() ) {
+            //If we reach this method this should never happen
+            return null;
+        }
+
+        BuilderPrism builderMapping = builderMappingPrism( method, ctx );
+        if ( builderMapping == null && buildMethods.size() == 1 ) {
+            return MethodReference.forMethodCall( first( buildMethods ).getSimpleName().toString() );
+        }
+        else {
+            String buildMethodPattern = DEFAULT_BUILD_METHOD_NAME;
+            if ( builderMapping != null ) {
+                buildMethodPattern = builderMapping.buildMethod();
+            }
+            for ( ExecutableElement buildMethod : buildMethods ) {
+                String methodName = buildMethod.getSimpleName().toString();
+                if ( methodName.matches( buildMethodPattern ) ) {
+                    return MethodReference.forMethodCall( methodName );
+                }
+            }
+
+            if ( builderMapping == null ) {
+                ctx.getMessager().printMessage(
+                    method.getExecutable(),
+                    Message.BUILDER_NO_BUILD_METHOD_FOUND_DEFAULT,
+                    buildMethodPattern,
+                    builderType.getBuilder(),
+                    builderType.getBuildingType(),
+                    Strings.join( buildMethods, ", " )
+                );
+            }
+            else {
+                ctx.getMessager().printMessage(
+                    method.getExecutable(),
+                    builderMapping.mirror,
+                    Message.BUILDER_NO_BUILD_METHOD_FOUND,
+                    buildMethodPattern,
+                    builderType.getBuilder(),
+                    builderType.getBuildingType(),
+                    Strings.join( buildMethods, ", " )
+                );
+            }
+        }
+
+        return null;
+    }
+
+    private static BuilderPrism builderMappingPrism(Method method, MappingBuilderContext ctx) {
+        BeanMapping beanMapping = method.getMappingOptions().getBeanMapping();
+        if ( beanMapping != null && beanMapping.getBuilder() != null ) {
+            return beanMapping.getBuilder();
+        }
+        return MapperConfiguration.getInstanceOn( ctx.getMapperTypeElement() ).getBuilderPrism();
+    }
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/BuilderType.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/BuilderType.java
@@ -18,6 +18,7 @@
  */
 package org.mapstruct.ap.internal.model.common;
 
+import java.util.Collection;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Types;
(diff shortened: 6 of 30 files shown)


filiphr:
@chris922 You also had a look at the issue, if you have time you can also review this PR. Thanks 😄

chris922:
@filiphr I just had a quick look:

1. In the reference documentation you are mentioning the new exceptions. As a user of mapstruct this would be a bit confusing for me, because - as far as I got it - those exceptions are used internally and are only exposed to the user as warnings/errors during processing and nothing I can catch. So is it really helpful to mention these exception names? Maybe for a custom SPI? (haven't yet checked the builder SPI)
2. This will prevent implementing an own `@AfterMapping` that is calling the correct builder method in case one of the new exceptions would be thrown, isn't it? I still think as a user writing a custom SPI is rather complex. You may use different builder patterns in case you are using different libraries and the SPI must support everything together. So using `@AfterMapping` is for me still a good and easy solution to select the correct `build` method.

filiphr:
Thanks a lot @chris922.

1. I mentioned the exceptions because they are in the public SPI package and they can be thrown by a custom SPI implementations. However, you are right, maybe I can remove them from the documentation and just mention what happens when that happens (no reference to exceptions).

1. You might be right. I think that we need to move that to a separate issue. What we can do is to allow setting the methods to `null` and just passing a type. In such case we can even use an `@ObjectFactory` to create a builder instance and we probably need something like `@ObjectCreator` or whatever that can be invoked to produce a target instance. In any case I think that we need to do that in a new issue and target it for Beta2.

chris922:
> MapStruct should alleviate the user from too complicated stuff like SPI's ..

I totally agree with this point. SPI should only be required for really complex stuff and not for basic stuff like deciding which `build` method should be called. 

Having a new parameter for `@Mapper`/`@MapperConfig` sounds good to me. 
Maybe this value should also be available in `@BeanMapping` to be able to decide this per mapping method and not just per mapping class?

filiphr:
@chris922 and @sjaakd I hear you 😄. I will try to find a better approach for this.

I agree that we need to make it easier for the users. The only issue is that the `Type` is created really early on and it's properties are required really early on, maybe even before we even touch the methods. Therefore it is a bit difficult too handle the more than one builder creation exception. Without an SPI there is no way for us to know the type for the builder and thus no way to know how to get the properties. I am open to ideas 😄 

As for the build method, I would propose that we introduce some new annotation like `@BuilderFinisher` or whatever and we can then filter in the methods that we have and if we find one we invoke that method instead.

For example instead of doing `return builder.build()` we would do `return builderFinisher(builder)`/

Btw, I am travelling until Sunday so won't be able to work on this

sjaakd:
> What is your suggestion about this? Is it OK to leave this for a custom SPI?

Is there a way to automatically detect what kind of builder pattern is present given the builder technology used? I think we did this recently as well, right?

filiphr:
>Is there a way to automatically detect what kind of builder pattern is present given the builder technology used? I think we did this recently as well, right?

By builder technology do you mean Immutables, FreeBuilder etc, or something else?

The current mechanism works in the following way:

* Find all public static methods on the type being build that have no parameters
* For each of those methods get the return type and in that type look for a public method without parameters that has a return type same as the type being build

What we can actually work with is when there is more that one builder creation method to not use builders. However, if there is a single builder creation method and multiple build methods than we would assume the user has defined a `BuilderFinisher` or `BuildFactory` and we would invoke that to finish the building. The `BuildFactory` can always be used, in the same way the `@ObjectFactory` is used.

sjaakd:
> By builder technology do you mean Immutables, FreeBuilder etc, or something else?

or even protobuf (believe it uses a builder pattern as well, right?



> What we can actually work with is when there is more that one builder creation method to not use builders. However, if there is a single builder creation method and multiple build methods than we would assume the user has defined a BuilderFinisher or BuildFactory and we would invoke that to finish the building. The BuildFactory can always be used, in the same way the @ObjectFactory is used.

So.. If i understand you correctly, in case of conflicts (multiple qualifying methods), you leave it to the user to point out the right "factory"  for creation, and the right "finisher" and delegate to the generated code, right? 

Could we also use the qualifier approach?  Because we normally use that to resolve conflicts somehow..?

filiphr:
>So.. If i understand you correctly, in case of conflicts (multiple qualifying methods), you leave it to the user to point out the right "factory" for creation, and the right "finisher" so the user could delegate to the generated code, right?


Factory for creation would be to implement an SPI as we need the builder really early, even before we work with the methods. The finisher can be done with an annotation.

>Could we also use the qualifier approach? Because we normally use that to resolve conflicts somehow..?

Not really, most of the time the builder code is generated, so there is no way for users to put the qualifiers. At least I don't see it, If you have some idea please do share it.

Example for the more than one builder creation method:

```
public Person {

    public static PersonBuilder builder() {}

    public static SecondPersonBuilder secondBuilder() {}
}
```

In the case of the builder creator for `Person` without an SPI we cannot figure out what is the type of the builder that we need, `PersonBuilder` or `SecondPersonBuilder`. If such case occurs the user will need to implement an SPI.

For the more than one build methods:

```
public class Vehicle {

    public static VehicleBuilder builder(){}

    public static class VehicleBuilder {

        public Vehicle create() {}
   
        public Vehicle otherCreate() {}
    }
}
```

In this case we can find the type of the builder (`VehicleBuilder`). However, the user will need to define something like:

```
@Mapper
public interface MyMapper {

    @BuilderFinisher
    default Vehicle build(VehicleBuilder builder) {
        return builder.otherCreate();
    }
}
```

With `@BuilderFinisher` the users would be able to define their own way of invoking the builder.

How does this sound?

sjaakd:
Ill check it out this evening..

sjaakd:
@filiphr .. I understand your approach.

However it would still take a considerable  amount of work to do this for each method. If you only get 3 or 4 properties to map in a bean, you now need to implement possibly a creater and a finisher method. Do you know if the builders predominantly use a consistent naming convention that we could use? E.g. if its always `createXYZ` this could be considered a pattern. You could also do this on the `@MappingConfig`, `@Mapping` and `@BeanMapping` as two properties. Like `builderCreatetor = "^createSecond.*$"`, `builderFinisher = "^buildSecond.*$"`.

filiphr:
@sjaakd, @chris922 I've changed the approach a bit. I took ideas from you, @sjaakd . There is now `@BuilderMapping` in which you can set the pattern for the finisher method.

I still need to update the Javadoc and the reference documentation, but wanted to get your opinion first. I went with a standalone annotation as that opens the path to meta annotations for 2.0 (perhaps). 

WDYT?

core-common/src/main/java/org/mapstruct/BuilderMapping.java
```diff
@@ -0,0 +1,38 @@
...
+ * @since 1.3
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({
+    ElementType.TYPE,
+    ElementType.METHOD
+})
+public @interface BuilderMapping {
+
+    String builderFinisher() default "build";
```
sjaakd: I thought the problem was 2 fold: `builderFinisher` and `builderCreator`.. where's the latter one?

processor/src/main/java/org/mapstruct/ap/internal/model/common/BuilderType.java
```diff
@@ -107,11 +107,6 @@ public static BuilderType create(BuilderInfo builderInfo, Type typeToBuild, Type
         if ( builderInfo == null ) {
             return null;
         }
-        ExecutableElement buildMethod = builderInfo.getBuildMethod();
```
sjaakd: is this check obsolete?

core-common/src/main/java/org/mapstruct/BuilderMapping.java
filiphr: The `builderCreator` one is a more complex one. We need the type of the builder at the time of the creation of the `Type` itself. As we need the properties really early on, when we are constructing the `Target` and `Source` references. That's why that is not hear, and I am having hard time pulling that here.

processor/src/main/java/org/mapstruct/ap/internal/model/BuilderFinisherMethodResolver.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+        else {
+            BuilderMappingPrism builderMapping = builderMappingPrism( method, ctx );
+            String buildMethodPattern = "build";
+            if ( builderMapping != null ) {
+                buildMethodPattern = builderMapping.builderFinisher();
+            }
+            for ( ExecutableElement buildMethod : buildMethods ) {
+                String methodName = buildMethod.getSimpleName().toString();
+                if ( methodName.matches( buildMethodPattern ) ) {
```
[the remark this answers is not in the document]
filiphr: Yes we could. However, currently we are making sure that this can't happen in the `DefaultBuilderProvider`. We can even make that as part of the contract of the `BuilderInfo`.

processor/src/main/java/org/mapstruct/ap/internal/model/BuilderFinisherMethodResolver.java
```diff
@@ -0,0 +1,109 @@
...
+        if ( buildMethods.isEmpty() ) {
+            //If we reach this method this should never happen
+            return null;
+        }
+        else if ( buildMethods.size() == 1 ) {
+            return MethodReference.forMethodCall( first( buildMethods ).getSimpleName().toString() );
+        }
+        else {
+            BuilderMappingPrism builderMapping = builderMappingPrism( method, ctx );
+            String buildMethodPattern = "build";
```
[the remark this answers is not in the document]
filiphr: Makes sense

core-common/src/main/java/org/mapstruct/BuilderMapping.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @author Filip Hrisafov
+ * @since 1.3
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({
+    ElementType.TYPE,
+    ElementType.METHOD
+})
+public @interface BuilderMapping {
```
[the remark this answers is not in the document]
filiphr: You are right. It might not make sense to `MapMapping`. However, it will make sense during the computation of the intermediate methods for the keys / values mappings. In my opinion we need to change the `MapperConfig` and allow use of meta annotations in 2.0. The mapper config might even not be needed.

processor/src/main/java/org/mapstruct/ap/internal/model/common/BuilderType.java
filiphr: Probably this check belongs to the `BuilderFinisherMethodResolver` now. As we can check the method return type there. Although this can't happen with the default builder provider as we have this check there.

filiphr:
@sjaakd it might actually make sense to apply it on `MapMapping` or `IterableMapping`. The reason is that we are constructing intermediary methods in there, so if they need builders we can use that from there (although I think it is not passed appropriately now).

2. Yes my idea was to go with meta annotations for the future. In any case I can change this and apply the `BuilderMapping` as a field in the `Mapper`, `MapperConfig` and `BeanMapping` in order to be consistent with the current paradigm.

How should I proceed with this one?

core-common/src/main/java/org/mapstruct/BuilderMapping.java
sjaakd: ok.. but then leave it open as a new issue / improvement.. From a user perspective, its strange, agreed?

sjaakd:
> How should I proceed with this one?

I would prefer to have a consists meta annotation interface in 2.0 and (a consistent one now) :smile:

processor/src/main/java/org/mapstruct/ap/internal/model/common/BuilderType.java
sjaakd: ok

processor/src/main/java/org/mapstruct/ap/internal/model/BuilderFinisherMethodResolver.java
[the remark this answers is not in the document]
sjaakd: ok

core-common/src/main/java/org/mapstruct/BuilderMapping.java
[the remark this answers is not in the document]
filiphr: Yes we do that in 2.0. I will apply the changes here.

core-common/src/main/java/org/mapstruct/BuilderMapping.java
filiphr: I'll create an issue for that. yes it might be a bit strange from a user perspective.

processor/src/main/java/org/mapstruct/ap/internal/model/BuilderFinisherMethodResolver.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    public static MethodReference getBuilderFinisherMethod(Method method, BuilderType builderType,
+        MappingBuilderContext ctx) {
+        Collection<ExecutableElement> buildMethods = builderType.getBuildMethods();
+        if ( buildMethods.isEmpty() ) {
+            //If we reach this method this should never happen
+            return null;
+        }
+        else if ( buildMethods.size() == 1 ) {
```
chris922: If a `@BuilderMapping` was defined with an explicit name for the `builderFinisher`, but only a different method was found, the wrong method will be used. The user would not notice this.

In case he entered a `builderFinisher` I would expect a warning or even better an error in case the only method found has a different name. The issue is most likely how we should know that he explicitly entered  a name, because you defined `build` as a default name in the annotation. So we will always have a `builderFinisher` as soon as the `@BuilderMapping` annotation _(or whatever it will be at the end)_ will be used.

The only way I see to solve this issue is to use an empty string as a default annotation attribute value and in this case use `build` as a constant value here in the code (the `DEFAULT_BUILD_METHOD_NAME`).
And switch the `if-else` ordering.. if a `builderFinisher` was defined try to use a method with exactly this name, if not possible -> error. If nothing was defined fallback to the way you are doing it right now (`size == 1`, then `methodName equals build`).

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -100,6 +100,10 @@
     GENERAL_NOT_ALL_FORGED_CREATED( "Internal Error in creation of Forged Methods, it was expected all Forged Methods to finished with creation, but %s did not" ),
     GENERAL_NO_SUITABLE_CONSTRUCTOR( "%s does not have an accessible parameterless constructor." ),
 
+    BUILDER_MORE_THAN_ONE_BUILDER_CREATION_METHOD( "More than one builder creation method for \"%s\". Found methods: \"%s\". Builder will not be used. Consider implementing a custom BuilderProvider SPI.", Diagnostic.Kind.WARNING ),
+    BUILDER_MORE_THAN_ONE_BUILD_METHOD_DEFAULT_PATTERN( "More than one build method in \"%s\" for \"%s\". Found methods: \"%s\". Using default \"build\" method pattern. Consider to add @BuilderMapping in order to select the correct build method.", Diagnostic.Kind.ERROR ),
```
chris922: This _"Using default .. method pattern"_ sounds confusing for me. Just after reading this message I thought this is an INFO that he found more than one possible build method, but uses at the end the one with the name `build`. Maybe something like:

`More than one build method in \"%s\" for \"%s\". Found methods: \"%s\", but no method is named \"build\". Consider to add @BuilderMapping in order to select the correct build method.`

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -100,6 +100,10 @@
...
+    BUILDER_MORE_THAN_ONE_BUILD_METHOD_DEFINED_PATTERN( "More than one build method in \"%s\" for \"%s\". Found methods: \"%s\". Using defined \"%s\" method pattern.", Diagnostic.Kind.ERROR ),
```
chris922: Same as above, maybe:

`More than one build method in \"%s\" for \"%s\". Found methods: \"%s\", but no method is named \"%s\" as defined in @BuilderMapping#builderFinisher.`

filiphr:
I've finally found the time to go over the changes in this PR 😄.

Thank you @chris922 and @sjaakd for the reviews.

I've applied both of your comments.

* We now have the `@BuilderMapping` as part of the `@BeanMapping`, `@Mapper` and/or `@MapperConfig`
* I've adapted the error messages a bit, they should be more descriptive now.

If you have time let me know what you think.

sjaakd:
Hey Filip.. Looks more consistent. 

I'm still a bit struggling whether we want to have annotations within annotations to solve this problem. I know you are considering some kind of meta annotation mechanism for 2.0. 

However, It would be more usable (and step-up to 2.0) if you could define an annotation as a user (a bit like we do with qualifiers), grouping specific `@BuilderMapping ` properties for reuse and mark that annotation with `@BuilderMapping`

So:
```java
@Target(ElementType.METHOD)
@BuilderMapping(builderFinisher = "create")
public @interface BuilderMappingForFrameworkX { }
```

```java
@Target(ElementType.METHOD)
@BuilderMapping(builderFinisher = "wrongCreate")
public @interface BuilderMappingForFrameworkY { }
```

You should then also be able to use the `@BuilderMappingForFrameworkX` and `@BuilderMappingForFrameworkY ` directly :

```java
@Mapper(builder = @BuilderMappingForFrameworkX)
public interface BuilderMappingDefinedMapper {

    BuilderMappingDefinedMapper INSTANCE = Mappers.getMapper( BuilderMappingDefinedMapper.class );

    Process map(Source source);

    @BeanMapping(builder = @BuilderMappingForFrameworkY)
    Process wrongMap(Source source);
}
```

If this is possible, we could:
1. Extend this mechanism to other places (e.g. reusable pieces of mappings instead of the `@InheritConfiguration` / `@InheritInverseConfiguration`
2. Define our own extra's... e.g. `@BuilderMappingForProtoBuf`, `@BuilderMappingForLombok`

Note: Having only one property in `@BuilderMapping` makes it a bit hard to see where you are heading for.  

Shall we discuss this in our Tuesday meeting?

core-common/src/main/java/org/mapstruct/BuilderMapping.java
```diff
@@ -0,0 +1,42 @@
...
+ *  limitations under the License.
+ */
+package org.mapstruct;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that can be used to provide a custom finisher for the builders in case of multiple build methods.
```
gunnarmorling: Let's not repeat that it's an annotation, that's clear from its definition:

> Configuration of builders, e.g. the name of the final build method.

core-common/src/main/java/org/mapstruct/BuilderMapping.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @author Filip Hrisafov
+ *
+ * @since 1.3
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({})
+public @interface BuilderMapping {
```
gunnarmorling: Should it be marked as `@Incubating` or are we sure that it won't be altered incompatibly down the road?

core-common/src/main/java/org/mapstruct/BuilderMapping.java
filiphr: Issue created https://github.com/mapstruct/mapstruct/issues/1547

filiphr:
I've squashed all commits, did the rename, update the commit message to:

> * Add new @builder annotation for defining a build method
>* When there are multiple build methods look for a method named `build` and if found use it
>* If @builder is defined than look for a build method with the defined method
>* When a type has multiple builder creation methods throw an exception and don't use the builder
>
>Defaulting to a method named `build` will make sure that a correct method is selected for:
>* FreeBuilder - it has two methods: `build` and `buildPartial`
>* Protobuf - it has three methods: `getDefaultInstanceForType`, `build` and `buildPartial`

I also updated the documentation. Once the build is done, I'll do the merge

core-common/src/main/java/org/mapstruct/Builder.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({})
+@Experimental
+public @interface Builder {
+
+    /**
+     * The name of the build method that needs to be invoked on the builder to create the type being build
+     *
+     * @return the method that needs to tbe invoked on the builder
```
chris922: `to be` :)

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -200,4 +200,22 @@
...
+     * The information that should be used for the builder mappings. This can be used to define custom build methods
+     * for the builder strategy that one uses.
+     *
+     * If no builder is defined the builder given via {@link MapperConfig#builder()} will be applied.
+     *
+     * <p>
+     * NOTE: In case no builder is defined here, in {@link BeanMapping} or {@link MapperConfig} and there is a single
+     * build method, then that method would be used.
+     * <p>
+     * If the builder is defined and there is a single method that does not match the name of the finisher than
```
chris922: `the name of the build method` as we don't use `finisher` anymore?

core-common/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -200,4 +200,22 @@
...
+     * a compile error will occurs
```
chris922: `occur.`

core-common/src/main/java/org/mapstruct/MapperConfig.java
```diff
@@ -186,4 +186,24 @@ MappingInheritanceStrategy mappingInheritanceStrategy()
...
+     * NOTE: In case no builder is defined here, in {@link BeanMapping} or {@link Mapper} and there is a single
+     * build method, then that method would be used.
+     * <p>
+     * If the builder is defined and there is a single method that does not match the name of the finisher than
+     * a compile error will occurs
+     *
+     * @return the builder information
+     *
+     * @since 1.3
+     */
```
chris922: same here

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -639,6 +639,12 @@ The default implementation of the `BuilderProvider` assumes the following:
 So for example `Person` has a public static method that returns `PersonBuilder`.
 * The builder type has a parameterless public method (build method) that returns the type being build
 In our example `PersonBuilder` has a method returning `Person`.
+* In case there are multiple build methods, MapStruct will look for a method called `build` if such methods exists
```
chris922: `method`

filiphr:
Thanks for the final checks @chris922. I'll commit those directly on master 😄

Issue #1398, opened 2018-03-17
@ObjectFactory not called when on an @Context object


sjaakd:
I was trying to solve:
https://github.com/mapstruct/mapstruct-examples/tree/master/mapstruct-lookup-entity-with-id

And wondered: why is @OjbectFactory not called (just as the lifecycle methods) on an context object? Feels inconsistent..

@agudian : was this done deliberately? I remember somewhere this was a discussion once. It now feels a bit inconsistent. At the call-site: A `@BeforeMapping` returning a result is followed by a null check which returns immediately.  So that's not useful as an object factory or a DB lookup. An `@ObjectFactory` works correctly (since its called where its needed), however it does not work in combination in `@Context`..

sjaakd:
After consulting we decided to add this to the context.

sjaakd:
fixed by pr

PR #1506, opened 2018-05-23
#1398 allowing `@ObjectFactory` annotated methods on the context

sjaakd:
Whereas its possible to have non `@ObjectFactory` annotated methods as object factory in the `uses` method of the mapper, this is not possible on the `Context` annotated parameter objects.

Reason is backwards compatibility: we cannot predict if a user has already parameterless methods on the context object. 

Non `@ObjectFactory` annotated methods are most likely generated factories, like `Jaxb object factories`. It makes sense to include those in the uses part. The context is most likely a user implemented object.

### DIFF
--- a/core-common/src/main/java/org/mapstruct/Context.java
+++ b/core-common/src/main/java/org/mapstruct/Context.java
@@ -32,8 +32,9 @@ import java.lang.annotation.Target;
  * {@code @}{@link BeforeMapping}/{@code @}{@link AfterMapping} methods, which are called on the provided context
  * parameter value if applicable.
  * <p>
- * <strong>Note:</strong> no {@code null} checks are performed before calling before/after mapping methods on context
- * parameters. The caller needs to make sure that no {@code null} are passed in that case.
+ * <strong>Note:</strong> no {@code null} checks are performed before calling before/after mapping methods or object
+ * factory methods on {@code @}{@link Context} annotated parameters. The caller needs to make sure that no {@code null}
+ * are passed in that case.
  * <p>
  * For generated code to call a method that is declared with {@code @Context} parameters, the declaration of the mapping
  * method being generated needs to contain at least those (or assignable) {@code @Context} parameters as well. MapStruct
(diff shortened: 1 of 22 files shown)


documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -1119,6 +1119,8 @@ public class CarMapperImpl implements CarMapper {
 
 Additional _context_ or _state_ information can be passed through generated mapping methods to custom methods with `@Context` parameters. Such parameters are passed to other mapping methods, `@ObjectFactory` methods (see <<object-factories>>) or `@BeforeMapping` / `@AfterMapping` methods (see <<customizing-mappings-with-before-and-after>>) when applicable and can thus be used in custom code.
 
+`@Context` parameters is searched for `@ObjectFactory` methods, which are called on the provided context parameter value if applicable.
```
agudian: “parameters are searched”

processor/src/main/java/org/mapstruct/ap/internal/model/LifecycleMethodResolver.java
```diff
@@ -91,7 +91,11 @@ private LifecycleMethodResolver() {
         List<SourceMethod> availableMethods =
             new ArrayList<SourceMethod>( methodsProvidedByParams.size() + sourceModelMethods.size() );
 
-        availableMethods.addAll( methodsProvidedByParams );
+        for ( SourceMethod methodProvidedByParams : methodsProvidedByParams ) {
+            if ( methodProvidedByParams.isLifecycleCallbackMethod() ) {
```
sjaakd: this looks like a task of the `MethodFamiliySelector`. Need to analyse whether to do it here or leave it up to this selector.

processor/src/main/java/org/mapstruct/ap/internal/model/ObjectFactoryMethodResolver.java
```diff
@@ -121,4 +135,27 @@ private static MethodReference findBuilderFactoryMethod(Type targetType) {
...
+
+        List<SourceMethod> methodsProvidedByParams = contextProvidedMethods
+            .getAllProvidedMethodsInParameterOrder( method.getContextParameters() );
+
+        List<SourceMethod> availableMethods =
+            new ArrayList<SourceMethod>( methodsProvidedByParams.size() + sourceModelMethods.size() );
+
+        // add only methods from context that do have the @ObjectFactory annotation
+        for ( SourceMethod methodProvidedByParams : methodsProvidedByParams ) {
+            if ( methodProvidedByParams.isObjectFactory() && methodProvidedByParams.hasObjectFactoryAnnotation() ) {
```
sjaakd: this looks like a task of the `MethodFamiliySelector`. Need to analyse whether to do it here or leave it up to this selector.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: shall we really do this..? Or treat the object factory methods the same for the context as for the 'uses' ? That would open up the option to handle Jaxb ObjectFactories directly as `@Context` as well.

processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
```diff
@@ -289,14 +289,14 @@ private ParameterProvidedMethods retrieveLifecycleMethodsFromContext(
                 mapperConfig,
                 Collections.<SourceMethod> emptyList() );
 
-            List<SourceMethod> lifecycleMethods = new ArrayList<SourceMethod>( contextParamMethods.size() );
+            List<SourceMethod> contextProvidedMethods = new ArrayList<SourceMethod>( contextParamMethods.size() );
             for ( SourceMethod sourceMethod : contextParamMethods ) {
-                if ( sourceMethod.isLifecycleCallbackMethod() ) {
-                    lifecycleMethods.add( sourceMethod );
+                if ( sourceMethod.isLifecycleCallbackMethod() || sourceMethod.isObjectFactory() ) {
```
sjaakd: perhaps we should simply all methods.. and not pre-select here.

sjaakd:
> Could you also extend the Javadoc of @ObjectFactory and @Context?

done.. I added an example and fixe my own comments.

sjaakd:
@agudian thanks for reviewing..

core-common/src/main/java/org/mapstruct/Context.java
```diff
@@ -135,7 +136,55 @@
...
+ *
+ * public class ContextWithObjectFactoryMapperImpl implements ContextWithObjectFactoryMapper {
+ *
+ *   &#64;Override
+ *   public Valve map(ValveDto dto, ContextObjectFactory factory) {
+ *       if ( dto == null ) {
+ *           return null;
+ *       }
+ *
+ *       Valve valve = factory.create();
```
agudian: In the declaration above, the create method has a String parameter 🙂

sjaakd: oops.. will fix.. :smile:

Issue #1482, opened 2018-05-07
Generic mapping method not selected


sjaakd:
Although tricky when in another jar file.. 

```java
public interface Test<T> {

     T getTest();

}
```

Handwritten method:
```java
public <T> T map( Test<T> in ) {
   return in.getTest();
}
```

Will not be selected when when
```java
public class Source {

     private Test<BigDecimal> bd;
     private Test<String> s;

}

public class Target {
   private BigDecimal bd;
   private String s;
}

```

filiphr:
Interesting. I think that being in another jar should not have an impact as during compilation that info should be there.

sjaakd:
I think that in a grey past I reported such an issue (in other jar though). But its also not working in the same jar..

sjaakd:
So.. the the problem is more subtile. I'll adapt the example above..

PR #1511, opened 2018-06-03
#1482 allowing generic self references in types when matching

sjaakd:
`Enum<MyEnum>` is currently not matched to `MyEnum` when looking for mapping methods. This is, because in a mapping method in general the LHS is type-var for type-var matched to the RHS. 

So, when having a mapping method: `protected String map(Enum<MyEnum> e)`, `MyEnum` will not be matched because `Enum<MyEnum>.` has one type var and  `MyEnum` has non.

For the solution, also assigning to a raw type has to be considered. So it should be possible to assign: `List<String>` to its raw type `List`.


processor/src/test/java/org/mapstruct/ap/test/bugs/_1482/SourceTargetMapper.java
```diff
@@ -0,0 +1,40 @@
...
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public abstract class SourceTargetMapper {
+
+    public static final SourceTargetMapper INSTANCE = Mappers.getMapper( SourceTargetMapper.class );
+
+    @Mapping( target = "bigDecimal", source = "wrapper" )
+    abstract Target map(Source source);
```
filiphr: What about the reverse case? From what I could see in the coverage report we are only testing when `assignability` is `Assignability.VISITED_ASSIGNABLE_TO`?

sjaakd: Good point. Can't fix it right now but will have a look tomorrow early in the evening

sjaakd: If it's blocking the release, you can merge it and I'll add a until test later tomorrow

sjaakd: If it's blocking the release, you can merge it and I'll add a until test later tomorrow

sjaakd: No problem. dunno if I manage this evening. There's a drinking event 🍻 . They're risky. 😄

sjaakd:
@filiphr  I'm not sure whether a direct reverse cases makes sense.. Lets simply reverse the test case:

```java
@Mapper
public abstract class TargetSourceMapper {

    public static final TargetSourceMapper INSTANCE = Mappers.getMapper( TargetSourceMapper.class );

    @Mapping( target = "wrapper", /* source = "bigDecimal" */ ignore = true )
    abstract Source map( Target target);

    protected Enum<SourceEnum> map(String in) {
        return SourceEnum.valueOf( in );
    }

    protected <T> ValueWrapper<T> map( T in, @TargetType  Class<T> clz) {
        if ( clz.isAssignableFrom( BigDecimalWrapper.class ) ) {
            return (ValueWrapper<T>) new BigDecimalWrapper( (BigDecimal) in );
        }
        return null;
    }
}
```

This would assume that we need to assign an abstract class or interface to an implementation:

```java 
        source2.setTest( TargetSourceMapper.INSTANCE.map( "test" ) );
        source2.setWrapper( TargetSourceMapper.INSTANCE.map( new BigDecimal( 5 ), BigDecimal.class ) );
```

Above, you can't assign an interface to an implementation. So I need to adapt the scenario a bit..

filiphr:
I mentioned it only because I saw that we have a condition based on the type of the assignability. However, if we can't find a good example it is fine. We can leave it as is. I've accepted the PR. Feel free to merge it

sjaakd:
I provided a good example.. The enum though caused some problems. You need to indicate which class you want to instantiate. Adding `raw` to the target type makes mapstruct want to create an enum whereas you want to create the enum realisation... I also had to loosen the check on the generic type a bit (subType).

sjaakd:
@filiphr If you could have a look. If your ok.. I already rebased. Stuff is ready for merge.. let me know if I need to do it or if you pick it up.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1482/TargetSourceMapper.java
```diff
@@ -0,0 +1,55 @@
...
+import org.mapstruct.TargetType;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public abstract class TargetSourceMapper {
+
+    public static final TargetSourceMapper INSTANCE = Mappers.getMapper( TargetSourceMapper.class );
+
+    @Mappings({
+        //@Mapping(target = "test", ignore = true ),
```
filiphr: We probably don't need this :)

filiphr:
Sorry I got carried away a bit. Yes the example is good and the code is good. There is only one lingering commented out code that we probably don't need.

You don't have to rebase. In the GitHub UI you can squash and merge 😄

sjaakd:
> Sorry I got carried away a bit.

No problem. I think its good to add the reverse test as well. 

The `BigDecimalWrapper` worked right out of the box. The enum took somewhat longer to troubleshoot and come up with a solution. Must say, I'm not entirely happy with the `inferTypeWhenEnum` (there are probably more cases were this is required in a more generic fashion). But this is what it is.. 



> You don't have to rebase. In the GitHub UI you can squash and merge

Hmm. all that new stuff..So squash and merge makes a fastforward merge or should I use "rebase and merge" for that one?

I'm still getting to terms with IntelliJ (commercial) :smile: . started using it 2 weeks ago, and must say.. its easier to depart from NB as I thought..

filiphr:
I had the same opinion about the `inferTypeWhenEnum`. Maybe We need something on the `Type` itself.

squash and merge - combines all your commits into one (and you can adapt the commit message) and commits on master. This will probably change the date of the commit as it creates a new one.
rebase and merge - just moves the commits on top of master keeping the authored date
create merge commit - we don't do that 😉 

>I'm still getting to terms with IntelliJ (commercial) 😄 . started using it 2 weeks ago, and must say.. its easier to depart from NB as I thought..

It's a beast on it's own 😄. I am a big fan of IntelliJ

sjaakd:
> I am a big fan of IntelliJ

I know.. :smile: . Like it to. In the case of MapStruct: especially the freemarker integration..

PR #1525, opened 2018-06-15
#1523 dont lose timezone mapping Calendar to XMLGregorianCalendar

### LINKED ISSUE #1523: ZonedDateTime to XmlGregorianCalendar losses time zone information
Mapstruct: **1.2.0.Final**
Java: **JDK 1.8.0-45**

# Issue

We unfortunately have an issue with the mapping from `ZonedDateTime` to `XMLGregorianCalendar` and we believe this problem comes from the generated MapStruct code.

During the mapping the time zone information that was set on the ZonedDateTime is lost and the resulting XmlGregorianCalendar has the default system timezone.

Entity PickupDate:
~~~java
public class PickupDate {

    private final LocalDate value;
    private final ZonedDateTime validUntil;

    public PickupDate(LocalDate value, ZonedDateTime validUntil) {
        Assert.notNull(value, "value must not be null");
        Assert.notNull(validUntil, "validUntil must not be null");

        this.value = value;
        this.validUntil = validUntil;
    }

    public LocalDate getValue() {
        return value;
    }

    public ZonedDateTime getValidUntil() {
        return validUntil;
    }
}
~~~

Generated XML Type PickupDateType:
~~~java
@XmlAccessorType(XmlAccessType.FIELD)
@XmlType(name = "PickupDateType", propOrder = {
    "value",
    "validUntil"
})
public class PickupDateType {

    @XmlElement(name = "Value", required = true)
    @XmlSchemaType(name = "date")
    protected XMLGregorianCalendar value;
    @XmlElement(name = "ValidUntil", required = true)
    @XmlSchemaType(name = "dateTime")
    protected XMLGregorianCalendar validUntil;

    public XMLGregorianCalendar getValue() {
        return value;
    }

    public void setValue(XMLGregorianCalendar value) {
        this.value = value;
    }

    public XMLGregorianCalendar getValidUntil() {
        return validUntil;
    }

    public void setValidUntil(XMLGregorianCalendar value) {
        this.validUntil = value;
    }
}
~~~

Mapper interface:

~~~java
@Mapper(disableSubMappingMethodsGeneration = true)
public interface PickupDateTypeMapper {

    PickupDateType toPickupDateType(PickupDate pickupDate);

    List<PickupDateType> toPickupDateTypeList(List<PickupDate> pickupDates);

}
~~~

In the mapper implementation the following line is generated to set the XMLGregorianCalendar:
~~~java
pickupDateType.setValidUntil( calendarToXmlGregorianCalendar( zonedDateTimeToCalendar( pickupDate.getValidUntil() ) ) );
~~~

The method `zonedDateTimeToCalendar` returns a GregorianCalendar with the correct time zone information:
~~~java
    private Calendar zonedDateTimeToCalendar(ZonedDateTime dateTime) {
        if ( dateTime == null ) {
            return null;
        }

        Calendar instance = Calendar.getInstance( TimeZone.getTimeZone( dateTime.getZone() ) );
        instance.setTimeInMillis( dateTime.toInstant().toEpochMilli() );
        return instance;
    }
~~~

But then the `calendarToXmlGregorianCalendar` method is used and it creates a `GregorianCalendar` from the passed `Calendar` without preserving the time zone information.
~~~java
    private XMLGregorianCalendar calendarToXmlGregorianCalendar( Calendar cal ) {
        if ( cal == null ) {
            return null;
        }

        try {
            GregorianCalendar gcal = new GregorianCalendar();
            gcal.setTimeInMillis( cal.getTimeInMillis() );
            return DatatypeFactory.newInstance().newXMLGregorianCalendar( gcal );
        }
        catch ( DatatypeConfigurationException ex ) {
            throw new RuntimeException( ex );
        }
    }
~~~

# Suggested Solution

A direct mapping between `ZonedDateTime` and `XMLGregorianCalendar` should be provided instead of the current two mapping methods:
~~~java
    private XMLGregorianCalendar zonedDateTimeToXmlGregorianCalendar(ZonedDateTime dateTime) {
        if ( dateTime == null ) {
            return null;
        }

        try {
            return DatatypeFactory.newInstance().newXMLGregorianCalendar(GregorianCalendar.from(zonedDateTime));
        } catch (DatatypeConfigurationException ex) {
            throw new RuntimeException( ex );
        }
    }
~~~

Thank you for your great software!
Christian Loose

chris922:
Fixes #1523.

I enhanced the `Calender` to `XMLGregorianCalendar` mapping to keep the timezone instead of creating a new `ZonedDateTime` to `XMLGregorianCalender` mapping as proposed in the issue.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_1523/java8/Issue1523Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_1523/java8/Issue1523Test.java
@@ -19,7 +19,11 @@
 package org.mapstruct.ap.test.bugs._1523.java8;
 
 import java.time.ZonedDateTime;
+import java.util.Calendar;
+import java.util.TimeZone;
 
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mapstruct.ap.testutil.IssueKey;
@@ -29,6 +33,11 @@ import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
+ * This test will evaluate if the conversion from {@code Calendar} to {@code XMLGregorianCalendar} works in case the
+ * default timezone was not used. Additionally a direct conversion between {@code ZonedDateTime} to
+ * {@code XMLGregorianCalendar} was added for this issue to improve readability / performance. This will be tested as
+ * well.
+ *
  * @author Christian Bandowski
  */
 @WithClasses({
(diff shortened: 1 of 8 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_1523/java8/Issue1523Test.java
```diff
@@ -0,0 +1,57 @@
...
+@IssueKey("1523")
+public class Issue1523Test {
+
+    @Test
+    public void testThatCorrectTimeZoneWillBeUsedInTarget() {
+        Source source = new Source();
+        // we want to test that the timezone will correctly be used in mapped XMLGregorianCalendar and not the
+        // default one, so we must ensure that the used timezone is never equal to the default one
+        // hopefully the fictive TZ +12:34 is nowhere the default one
+        source.setValue( ZonedDateTime.parse( "2018-06-15T00:00:00+12:34" ) );
```
gunnarmorling: Nice idea :) Wouldn't it be easier though to set the default TZ of the VM to a specific value before running this test?

chris922: Is a bit more complex, but definitly cleaner. Updated & pushed!

chris922:
Thanks a lot for your feedback, @gunnarmorling ! I added a direct mapping from `ZonedDateTime` to `XMLGregorianCalendar` as well to keep the code clean. Thus this PR fixes the issue between losing timezone mapping from `Calendar` to `XMLGregorianCalendar` and adds a new direct mapping.

sjaakd:
@chris922 .. thanks for tackling this one.. nice job

Issue #1470, opened 2018-05-02
Can user-defined annotations on a field be accessed during mapping ?


amitga-rnd:
Hello.
I started playing around with MapStruct, and think its an awesome framework!
But, I have a certain use case, that I haven't found a solution for yet:

Lets say I have some user-defined annotations on a field of a class. Can I access this annotation somehow during the mapping/conversion to another class, in order to affect the conversion of the value of the field ?

For example:
```
public class PersonEntity extends Entity {
    public String birthdate;

    @GeoConvert(format="WKT")
    public String addressGeo;

    public Name name;

    public String gender;
}
```
When I convert `PersonEntity` to `Person`, I want to read the annotations on `addressGeo` field (of source class), so that I can pass the field's value, and the annotation's value to a converter (reverse geocoding in this case)  and set the target's `address` field accordingly.

This is just an example, I have some other annotations, that work in a similar manner.

I guess I could do this manually, looking at each entity's fields, and mapping it accordingly, but I preffer not to do this, as I'm not the one writing the entities (I get them from a jar file written by another team), and I don't want to change the mapping, every time I receive an update of the entities jar.
Another option is to user `@AfterMapping`, and then use reflection to find annotations on the source object's class, but this probably wouldn't be very performant (and kinda breaks the fact the MapStruct doesn't user reflection in runtime, which I like :-))

Any other ideas? Could this be implemented somehow in the framework (future feature)?
Thanks in advance.

sjaakd:
Did you have a look at qualifiers? That might solve your use case.

amitga-rnd:
Yes I have, but it doesn't exactly solve my use case.
I need to know the value of a parameter in the annotation (i.e. `format="WKT"` in the  @GeoConvert annotation) in order to properly translate the values of that field.
It would be nice if this could be extracted and then sent to a conversion method that I have.

I have other annotations, where I need the of a parameter in them in order to map, or translate the value correctly.

sjaakd:
How many formats are you expecting? Would an annotation `@GeoConvertWKT` also be an option?

sjaakd:
ah.. esri stuff. are you dealing with object graphs? or single mapping methods?

sjaakd:
I was thinking, we have context objects.. (annotated `@Context`).. perhaps they are somehow usable...

sjaakd:
So.. ideas start to form a bit.. And I must say.. I'm not entirely sure if this somehow achievable... But suppose you would have a method:

[...]

PR #1526, opened 2018-06-17
#1470 first version of handling annotations as info

sjaakd:
I still need to add more unit test. For instance, combination of `@SourceAnnotation` and `@TargetAnnotation`

I'm also thinking whether we should adapt our selection mechanism to select on matching .. But first, the idea itself.


processor/src/main/resources/org/mapstruct/ap/internal/model/MethodReference.ftl
```diff
@@ -51,7 +51,7 @@
             <@includeModel object=ext.targetType raw=true/>.class<#t>
         <#elseif param.mappingTarget>
             ${ext.targetBeanName}<#if ext.targetReadAccessorName??>.${ext.targetReadAccessorName}</#if><#t>
-        <#elseif param.mappingContext>
+        <#elseif param.mappingContext || param.sourceAnnotation || param.targetAnnotation>
```
sjaakd: param.contextual

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -940,16 +985,24 @@ public PropertyMapping build() {
...
+            new ArrayList<LocalVariableDefinition>()
         );
     }
 
     private PropertyMapping(String name, String sourceBeanName, String targetWriteAccessorName,
                             ValueProvider targetReadAccessorProvider, Type targetType,
                             Assignment assignment,
-        List<String> dependsOn, Assignment defaultValueAssignment) {
+                            List<String> dependsOn, Assignment defaultValueAssignment,
+                            List<LocalVariableDefinition> localVariabeleDefinitions) {
```
chris922: Typo: `localVariableDefinitions`

chris922:
Ah, and creating anonymous instances in the generated code for the annotations looks a bit awkward.. 

@gunnarmorling : Maybe you've got a better idea? I just found on stackoverflow that you worked on a similar topic for hibernate-validator (https://stackoverflow.com/a/16326389/2982468)

chris922:
> You mean the property (instance variable) on which the getter / setter is acting?

Oh yes, of course. Property and not parameter .. 😄 

> The mapping method has scope: one getter to one setter. So its considering only the annotations on that particular getter an setter.

Ah ok, I think I got it. It is meant to get the property value + the annotations of exactly this property and not to add the possibility to have one mapping method that will get the annotation values of several properties.

> A problem are Repeatable annotations (just like Mapping / Mappings. However, in the future we might provide a List<Annotation> as well.

Repeatable annotations always have one "container" annotation and imho you don't need to support `List<Annotation>` - it should be the container annotation instead of such a list. But this is definitly a bit more complex as you will not find this annotation directly on the element and must check if any of the element annotations are annotated with `@Repeatable(ContainerAnnotation.class)` or similar.

---

Just a random thought: In case one of these new methods that accepts annotations it would be hard to call this method manually (as you might have to instantiate the annotation like you are doing it right now in the generated code). I could image that you often just need the value of just one annotation attribute, e. g. the `value`.. what about a possibility to just get one (or several) annotation attribute values? 

```java
public Target toTarget(Source source, @SourceAnnotation(type = MyAnnotation.class, attribute = "value") String myAnnotationValue) { ... }
```

... maybe an idea as a follow-up. Maybe a new annotation should be defined for this purpose.. but having thousands of annotations makes it also more complex.
We shouldn't discuss this idea really in detail in this PR, but in case you also think this is a good idea someone could create a new issue for this. I just wanted to mentioned this idea.

sjaakd:
> Ah ok, I think I got it. It is meant to get the property value + the annotations of exactly this property and not to add the possibility to have one mapping method that will get the annotation values of several properties.

👍 exactly


> Repeatable annotations always have one "container" annotation and imho you don't need to support

Perhaps it works out of the box. There's a handy value in the annotation mirror. Enums also work that way.



> Just a random thought: In case one of these new methods that accepts annotations it would be hard to call this method manually

The idea was the method was only called by generated code. Checkout the #1470 issue. But once a solution is in place it can always be extended. I was struggling with overload methods though. I think that mapstruct should do a best match. So if a method without the new `Source/TargetAnnotation` is there and the annotations are available, MapStruct should not give an error (multiple candidates available).

agudian:
As I don’t fully understand the use case, I womder if you couldn’t get the same results by passing such information that is meant to be interpreted in manually created mappers down through simple `@Context` parameters... 
Or we provide a type that can be passed as Context and that we fill with the properly typed information? Because the installation of the annotations sure look funny and I feel we’re misusing annotations there.

sjaakd:
> As I don’t fully understand the use case, I womder if you couldn’t get the same results by passing such information that is meant to be interpreted in manually created mappers down through simple @Context parameters...

The original use case is in #1470 .. its all about passing meta information from source / target DTO to the mapping method. This opens new possibilities by annotating your DTO's to hint handling..

sjaakd:
@filiphr : I don't think we should push this one through.. What about you?  I think that generating a usable runtime version of an annotation (without reflection) has definitely added value, but perhaps as part of another project. 

Now: there's an example pending that uses reflection that works with the current versions of MapStruct. I know that MapStruct is all about code generation. But not all problems related to generating code belong 'naturally' to MapStruct. So should we merge the example and close this one?

Perhaps, in the future, if there's more demand for this we could re-open it. For that purpose we could push it on an experimental branch within MapStruct.

filiphr:
To be honest I haven't had a look at this PR @sjaakd. However, it seems like it is not something that should be handled by MapStruct. 

>But not all problems related to generating code belong 'naturally' to MapStruct. So should we merge the example and close this one?

Indeed not everything fits MapStruct. Yes let's close this and merge the example (I approved that PR btw).

>Perhaps, in the future, if there's more demand for this we could re-open it. For that purpose we could push it on an experimental branch within MapStruct.

Feel free to push it to an experimental branch if you want to. The PR itself will never go away as well 😄

sjaakd:
closed and parked as experiment.

PR #1530, opened 2018-06-28
Document of sample code is broken

yokotaso:
At part of `default expression`,  Document of sample code is broken.


sjaakd:
@yokotaso thanks for your contribution. Issue merged to master

Issue #938, opened 2016-10-25
Classes declared with 'uses' not being autowired


brettanomyces:
Classes declared with 'uses' will only be autowired if one of their methods is used as a source->target mapping, so if their only use is in an expression they will not get autowired

http://stackoverflow.com/q/38807415/1427295

nowakprojects:
I have this siutation in many cases. When we should expect this enhancement to be implemented?

filiphr:
Currently this is not high on our priorities. However, we would gladly accept a PR for this enhancement, if you are willing to contribute.

chris922:
What do you think is the best way for this?

The easiest way would probably be to just include all mapper references by default in the generated mapper, but imho this is not the best option..  I guess there are reasons why it will be checked which mapper references will be used and only those will be included.

Checking the expressions if such a reference will be used would probably the best, but also most complex, way. As far as I got it the expressions are just inserted as raw values and they will not be evaluated.

sjaakd:
This can be solved now, by means of `@Context` and a handwritten method.

```java
@Mapper(componentModel="spring")
public interface MyMapper { 

   @Mapping(target="x",ignore = true)
   // other mappings
   Target map( Source source, @Context MyService service);

    @AfterMapping
   default void map( @MappingTarget Target.X target, Source.ID source, @Context MyService service) {
        target.set( service.findById( source.getId() ) );
   }
 }
```

I think this is by far preferable over the proposed solution (expressions are intended to fill in stuff that MapStruct can't do). Also: relying on a generated field name is probably not the safest thing to do with future extensions in mind.

Therefore I would propose to close this issue.

yelhouti:
@sjaakd thsi doesn't solve the problem as now, if you want to call `map` you need to pass the context as a parameter, clearly not very handy...

yelhouti:
@filiphr I will gladly add a PR for this if I know what you guys expect, one choice could be:
if a method contains `@context` (ex: `MyService` in the first map method below) and the same service is present in `uses`, have the method superchaged using the Autowired/injected `this.MyService`
the resulting implementation would look like
```
Target map( Source source, MyService myService) {...}

Target map( Source source) {
    return map(source, this.myService)
}
```
We can di the same for the afterMapping method but I don't know if it would be usefull.

filiphr:
@yelhouti you know that you can make the mapper an `abstract` class and autowire what you need in that

mschielmann:
[...]

PR #1533, opened 2018-07-01
Issue#938 Fix Classes declared with 'uses' not being autowired

yokotaso:
Hi, Thank you for developing wonderful tools!
I try to fix https://github.com/mapstruct/mapstruct/issues/938.

If you have a free time, cloud you review it?

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_938/Issue938Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_938/Issue938Mapper.java
@@ -1,0 +1,32 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
+ *  and/or other contributors as indicated by the @authors tag. See the
+ *  copyright.txt file in the distribution for a full listing of all
+ *  contributors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.test.bugs._938;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+
+/**
+ * @author yokotaso <yokotaso.t@gmail.com>
+ */
+@Mapper(componentModel = "spring", uses = Issue938Service.class)
+public interface Issue938Mapper {
+
+    @Mapping(target = "value", expression = "java(issue938Service.convert( source ))")
+    Target map(Source source);
+}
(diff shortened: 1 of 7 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_938/Issue938Mapper.java
```diff
@@ -0,0 +1,32 @@
...
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+
+/**
+ * @author yokotaso <yokotaso.t@gmail.com>
+ */
+@Mapper(componentModel = "spring", uses = Issue938Service.class)
+public interface Issue938Mapper {
+
+    @Mapping(target = "value", expression = "java(issue938Service.convert( source ))")
```
sjaakd: I'm not in favour of directly referring to an generated name in an expression. For one thing: the internal workings of MapStruct (and naming of fields) are up for change when we add more functionality. However, we could create some syntax to do this more logically by referring to the used mapper type. E.g. " java( ${Issue938Service.class}.convert( source ) )" 

More general: I'm doubting whether this kind of functionality is still required now we have the `@Context` annotation. You can bring a mapper to a handwritten method by means of context. 

I will also put this up as a comment in the issue.

@gunnarmorling : WYDT?

gunnarmorling: So personally I'm not too concerned about referring to the fields from within expressions. It's not that we'd arbitrarily change it, but more importantly, the generated code wouldn't compile if there was a mismatch, so one would find out about it very quickly.

The bigger problem I see is how to detect the usage within an expression reliably. The proposed way of simply checking `contains()` may lead to superfluous fields being added, when e.g. having the name of a field within a string literal within the expression. OTOH it's not a realy problem I suppose. Ultimately, it leads back to the idea to having a proper expression language within `expression`, which we'd then fully parse and thus understand what are actual field references. But that's not something right around the corner :)

The context solution seems workable, too. It might actually be the better way to do it.

yokotaso:
> In general: I see the code is reformatted. Can you try to avoid this? There is a template for the eclipse formatter available in '/etc'

Sorry, But I configured 'etc/mapstruct.xml' with IntelliJ IDEA which is announced at CONTRIBUTING.md.

yokotaso:
@seanjob @gunnarmorling 
Thank you for comments and review.

> More general: I'm doubting whether this kind of functionality is still required now we have the @Context annotation. You can bring a mapper to a handwritten method by means of context.

Oh, I don't know `@Context` feature. It is very good idea. workaround seems to work fine.

So, It is good to add test code of #938 and close #938, I think.

I will try to resend PR which add test code for workaround #938, is it needed?

yokotaso:
938 will be closed.
https://github.com/mapstruct/mapstruct/issues/938#issuecomment-404278795

sjaakd:
> Sorry, But I configured 'etc/mapstruct.xml' with IntelliJ IDEA which is announced at CONTRIBUTING.md.

No reason to feel sorry. Your contributions are very much appreciated 👍 .. What you can do is just format a selection that you are working on. Makes reviewing a bit more easy.


> is it needed

Lets wait with this one how the issues react on the tag / SO answer. 

In general: if you want to pick up an issue let us know inside the issue. We can first discuss how / if we should address the issue. I'll try to be a bit alert on this as well.

sjaakd:
@yokotaso : thanks for the effort you put into this one.. However, given the absence of responses on the original issue, I'm going to close this one.

PR #1534, opened 2018-07-01
Fix Method name typo

yokotaso:
I found method name typo and then fix it. Please review PR?


yokotaso:
Thank you! I use so cool this library in my project!

Issue #1531, opened 2018-06-28
defaultValue not handled in InheritInverseConfiguration


sgrekov:

sjaakd:
behaviour is intentional. Only: `source`, `target` and if both are specified `ignore` is mapped. The reason is twofold:
1. A conversion to a type from and to `String` must exist (as of 1.3 primitive types and native types are treated as such) must be present.
2. There's no guarantee that `source` and `target` are of the same type.
3. There's no guarantee that the user wants the same for both `source` and `target`.

Having said that.. 3. could possibly be overcome in the same way as is specified above: specifying `source` and `target` in combination with `defaultValue`

sgrekov:
> 2. There's no guarantee that source and target are of the same type.
Well, defaultValue() already assumes that 1. and 2. is true

By 
>  specifying source and target in combination with defaultValue
you mean drop using @InheritInverseConfiguration and do twoway mapping manually?

sjaakd:
> Well, defaultValue() already assumes that 1. and 2. is true

Why? The problem is from a MapStruct perspective we cannot assume that a user always want a `defaultValue` to be applied when you inherit inverse mappings. The same goes for other fields you can specify in `@Mapping` besides `source` and `target` (like `qualifiedBy`, `dateFormat`, etc.) with the exception of `ignore` (explained below)


> you mean drop using @inheritinverseconfiguration and do twoway mapping manually?
 
No. That is not what I mean. However, you always have to option to inherit all inverse mappings but override individual mappings. So like this:
```java
@InheritInverseConfiguration // inherits all inverse mappings
@Mapping( source = "x", target = "y", defaultValue = "blah" ) // with exception of the mapping for target "y"
B mapTo( A dto );
```

That should solve your case. Let me know.

Forget about my remark: What I meant  was  a similar approach that we have to `ignore=true`. Only when you specify `source`, `target` and `ignore` the `ignore` is automatically applied to the `source`. So why is `ignore` treated differently than other fields in `@Mapping?:  `ignore` does apply to the `target`. Therefore: `source` only has meaning in the reverse mapping so in (`@InheritInverseConfiguration`). So, my mistake: This scenario is not comparable to what you want.

sgrekov:
> you always have to option to inherit all inverse mappings but override individual mappings

Ah! This is what I want, thanks! It will be nice if you provide this example somewhere in docs more clearly, cause I didn't even assumed I can do it that way.

sjaakd:
[...]

PR #1535, opened 2018-07-01
Add sample of @Mapping annotation to @InheritInverseConfiguration

sgrekov:
Add sample of @mapping annotation to @InheritInverseConfiguration in example in order to make it more clear that we can these annotation together.

Fixes mapstruct/mapstruct#1531


sjaakd:
merged.. @sgrekov thanks for your contribution..

Issue #1532, opened 2018-06-28
Cache DatatypeFactory


gunnarmorling:
I came across this (German) post which shares some experiences with MapStruct: https://blog.doubleslash.de/welcher-javabean-mapper-die-qual-der-wahl/

It is positive in general, but there's one performance issue being pointed out: when converting a `Date` into an `XmlGregorianCalendar`, the mapping method calls `DatatypeFactory#newInstance()` each single time (see [template](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/resources/org/mapstruct/ap/internal/model/source/builtin/DateToXmlGregorianCalendar.ftl#L30), might affect other built-in mappings, too). This call is expansive (service loader usage), so we should do it only once.

The easiest would be, if the template contained the field as well (it'd result in somewhat unclean code as it'd add the field not at the beginning of the class but somewhere in the middle, which I think is alright). We'd have to take care though to avoid naming collisions.

sjaakd:
I guess its not to difficult to extend the current solution to include also a new mechanism to declare instance variable definitions along the same lines the template for this conversion is added. That could not be to difficult to make.

sjaakd:
Done.. need to add unit test for CDI though.

filiphr:
@sjaakd should we close this one and open a new one for the CDI unit test?

sjaakd:
Most certainly. Initially I thought that this could be resolved quick. Now I realize that we need more time to fix UT first. Although it does not feel comfortable CDI is not properly tested.

PR #1540, opened 2018-07-08
#1532 adding a final field for DatatypeFactory constructor initialized.

sjaakd:
* Final field is constructor initialised
* VirtualMappingMethod is renamed to SupportingMappingMethod
* Solution is reusable.. Next to SupportingMappingMethod there are SupportingField and SupportingConstructorFragment
* Optimised solution (base class for all to -xmlCalendar builtins).
* Name clashes with object factory with coincidentally the same name avoided
* Extra test added on field duplication and avoiding name conflicts
* Fixed templates
* Fixed constructor based injection.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractMappingMethodBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractMappingMethodBuilder.java
@@ -37,7 +37,7 @@ public abstract class AbstractMappingMethodBuilder<B extends AbstractMappingMeth
         }
 
         String name = getName( sourceType, targetType );
-        name = Strings.getSaveVariableName( name, ctx.getNamesOfMappingsToGenerate() );
+        name = Strings.getSafeVariableName( name, ctx.getNamesOfMappingsToGenerate() );
         ForgedMethodHistory history = null;
         if ( method instanceof ForgedMethod ) {
             history = ( (ForgedMethod) method ).getHistory();
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/Field.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/Field.java
@@ -5,7 +5,9 @@
  */
 package org.mapstruct.ap.internal.model;
 
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 
 import org.mapstruct.ap.internal.model.common.ModelElement;
@@ -113,6 +115,14 @@ public class Field extends ModelElement {
         final Field other = (Field) obj;
         return !( (this.variableName == null) ?
             (other.variableName != null) : !this.variableName.equals( other.variableName ) );
+    }
+
+    public static List<String> getFieldNames(Set<Field> fields) {
+        List<String> names = new ArrayList<String>( fields.size() );
+        for ( Field field : fields ) {
+            names.add( field.getVariableName() );
+        }
+        return names;
     }
 
 }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/NoArgumentConstructor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/NoArgumentConstructor.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.internal.model;
+
+import java.util.Collections;
+import java.util.Set;
+
+import org.mapstruct.ap.internal.model.common.ModelElement;
+import org.mapstruct.ap.internal.model.common.Type;
+
+/**
+ * Represents a constructor that is used for constructor injection.
+ *
+ * @author Sjaak Derksen
+ */
+public class NoArgumentConstructor extends ModelElement implements Constructor {
+
+    private final String name;
+    private final Set<SupportingConstructorFragment> fragments;
+
+    public NoArgumentConstructor(String name, Set<SupportingConstructorFragment> fragments) {
+        this.name = name;
+        this.fragments = fragments;
+    }
+
+    @Override
+    public Set<Type> getImportTypes() {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    public Set<SupportingConstructorFragment> getFragments() {
+        return fragments;
+    }
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/StringToXmlGregorianCalendar.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/StringToXmlGregorianCalendar.java
@@ -5,59 +5,48 @@
  */
 package org.mapstruct.ap.internal.model.source.builtin;
 
-import static org.mapstruct.ap.internal.util.Collections.asSet;
-
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.GregorianCalendar;
 import java.util.Set;
 
-import javax.xml.datatype.DatatypeConfigurationException;
-import javax.xml.datatype.DatatypeFactory;
-import javax.xml.datatype.XMLGregorianCalendar;
-
 import org.mapstruct.ap.internal.model.common.ConversionContext;
 import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.common.TypeFactory;
 
+import static org.mapstruct.ap.internal.util.Collections.asSet;
+
 /**
  * @author Sjaak Derksen
  */
-public class StringToXmlGregorianCalendar extends BuiltInMethod {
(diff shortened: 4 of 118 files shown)


chris922:
Looks nice! Just one possible improvement: What if more then one supporting field is required? Maybe change the `getSupportingField()` in a way that it returns a `String` to `Field` map so that the fields could be accessed by key/name?

processor/src/main/java/org/mapstruct/ap/internal/model/VirtualField.java
```diff
@@ -0,0 +1,42 @@
+/**
+ *  Copyright 2012-2017 Gunnar Morling (http://www.gunnarmorling.de/)
```
gunnarmorling: We should update the header template, more than 50% of 2018 are over :)

processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/StringToXmlGregorianCalendar.java
```diff
@@ -77,4 +80,9 @@ public Type getReturnType() {
     public String getContextParameter(ConversionContext conversionContext) {
         return conversionContext.getDateFormat() != null ? "\"" + conversionContext.getDateFormat() + "\"" : "null";
     }
+
+    @Override
+    public BuildInFieldReference getFieldReference() {
```
gunnarmorling: s/BuildInFieldReference/BuiltInFieldReference/

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractMappingMethodBuilder.java
```diff
@@ -51,7 +51,7 @@ Assignment forgeMapping(SourceRHS sourceRHS, Type sourceType, Type targetType) {
         }
 
         String name = getName( sourceType, targetType );
-        name = Strings.getSaveVariableName( name, ctx.getNamesOfMappingsToGenerate() );
+        name = Strings.getSafeVariableName( name, ctx.getNamesOfMappingsToGenerate() );
```
gunnarmorling: :)

processor/src/main/java/org/mapstruct/ap/internal/model/Field.java
```diff
@@ -128,4 +130,12 @@ public boolean equals(Object obj) {
             (other.variableName != null) : !this.variableName.equals( other.variableName ) );
     }
 
+    public static List<String> getFieldNames(List<Field> fields) {
+        List<String> names = new ArrayList<String>( fields.size() );
```
gunnarmorling: Are we on Java 8 already? If so, it could be nicely done using streams. No strong need to change of course.

processor/src/main/java/org/mapstruct/ap/internal/model/VirtualField.java
```diff
@@ -0,0 +1,42 @@
...
+package org.mapstruct.ap.internal.model;
+
+import org.mapstruct.ap.internal.model.source.builtin.BuildInFieldReference;
+
+/**
+ * A mapper instance field, initialized as null
+ *
+ * @author Sjaak Derksen
+ */
+public class VirtualField extends Field {
```
gunnarmorling: Why is it "virtual"?

processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/BuildInFieldReference.java
```diff
@@ -0,0 +1,29 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.mapstruct.ap.internal.model.source.builtin;
+
+import org.mapstruct.ap.internal.model.common.Type;
+
+public interface BuildInFieldReference {
```
gunnarmorling: A comment would be great.

processor/src/main/resources/org/mapstruct/ap/internal/model/source/builtin/CalendarToXmlGregorianCalendar.ftl
```diff
@@ -27,7 +27,10 @@ private <@includeModel object=findType("XMLGregorianCalendar")/> ${name}( <@incl
     try {
         <@includeModel object=findType("GregorianCalendar")/> gcal = new <@includeModel object=findType("GregorianCalendar")/>( cal.getTimeZone() );
         gcal.setTimeInMillis( cal.getTimeInMillis() );
-        return <@includeModel object=findType("DatatypeFactory")/>.newInstance().newXMLGregorianCalendar( gcal );
+        if ( ${supportingField.variableName} == null ) {
```
gunnarmorling: Should go into the constructor as dicussed and be made final.

processor/src/main/java/org/mapstruct/ap/internal/model/Field.java
sjaakd: no.. we aren't unfortunately..

processor/src/main/java/org/mapstruct/ap/internal/model/VirtualField.java
sjaakd: Because of the same reason the method that uses the field is called `Virtual`. :smile: . The `BuiltinMethod` and `HelperMethod` are wrapped in a `VirtualMappingMethod ` that binds the template to a method. I've no idea who came up with the name `Virtual`, but the blame perhaps offers some clues :smile:

processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/StringToXmlGregorianCalendar.java
sjaakd: ok

processor/src/main/java/org/mapstruct/ap/internal/model/VirtualField.java
sjaakd: yep.. there should be an issue for it.. time's progressing fast... too fast..

sjaakd:
> Looks nice! Just one possible improvement: What if more then one supporting field is required? Maybe change the getSupportingField() in a way that it returns a String to Field map so that the fields could be accessed by key/name?

Lets take that one as  the demand comes along.

sjaakd:
As agreed. I'm going to investigate whether we can initialise the (final) field holding the `DatatypeFactory` in the constructor. Possibly requires a new concept (`VirtualConstructorFragment`) or something along that line.

processor/src/main/resources/org/mapstruct/ap/internal/model/source/builtin/CalendarToXmlGregorianCalendar.ftl
sjaakd: yep

processor/src/main/java/org/mapstruct/ap/internal/model/VirtualField.java
sjaakd: renamed it to `SupportingMappingMethod`

sjaakd:
@filiphr : about constructor based injection: I assume we only inject mappers (not datatypefactories). This means, whenever there is constructor based injection, we should also include a default (no arg constructor) that is called from the constructor with injection.. Agreed?

I need to check why there's a drop in code coverage.. I'm a bit surprised, since I improved current functionality. Will checkout later.\

Perhaps the above remarks need to be handled separately, in order to keep the PR manageable.

sjaakd:
There's no test infrastructure for CDI in unit test. I noticed it is added for Spring.. Should we add a test dependency on Weld to achieve this as well?

sjaakd:
please test this

sjaakd:
@filiphr travis seems to have a problem with downloading java8..

processor/src/main/java/org/mapstruct/ap/internal/model/NoArgumentConstructor.java
```diff
@@ -0,0 +1,42 @@
...
+    private final Set<SupportingConstructorFragment> fragments;
+
+    public NoArgumentConstructor(String name, Set<SupportingConstructorFragment> fragments) {
+        this.name = name;
+        this.fragments = fragments;
+    }
+
+    @Override
+    public Set<Type> getImportTypes() {
+        return Collections.emptySet();
```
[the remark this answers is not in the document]
sjaakd: Thought about that. You could claim the same for the instance variable definition. In the end I considered both fragment as instance variable definition as supportive to the built-in mapping method that uses them. So that defines all. It's perhaps not entirely perfect from a separation of concerns perspective but it'll do 😀

filiphr:
@sjaakd yes the integration tests are not included in the coverage. However, there is a way that you might be able to achieve this. You can in theory provide a custom `AnnotationBasedComponentModelProcessor` that would use Spring and would return `true` for the `additionalPublicEmptyConstructor` that way you would be able to test that easily :). 

In case you have a way to do the CDI unit tests feel free, I don't know how. In my opinion having a test dependency for CDI tests is OK

sjaakd:
Oops.. now that is a dirty trick 😉. Perhaps we can include weld for certain unit test or otherwise use Mockito.. I'll think about it and when I'm back Ive a solution 

Verstuurd vanaf mijn iPhone

Issue #1384, opened 2018-02-24
Extract AstModifyingAnnotationProcessor


twentylemon:
I am working on my own annotation processor which will use getters/setters in generated code, and ran into the same issues mapstruct once had with lombok. I started writing a solution which utilizes #510 #1042 (and it works! clever solution!) but I don't like using what seems like internal code. Plus, my AP now newly depends on mapstruct but doesn't use it for any other reason.

Have you guys considered extracting the `AstModifyingAnnotationProcessor` portion of this project to another library?

gunnarmorling:
`AstModifyingAnnotationProcessor` isn't really internal code, but I can relate do your point of not depending on MapStruct just for this. We haven't considered extracting this interface so far, but it's an interesting idea.

filiphr:
I also think that it is an interesting idea to extract this interface into a different project / library. I marked if for team discussion so we can talk about it in our next meeting. In any case if anyone has a suggestion for the library.

I don't think it would be a problem for us to move this piece of code to a new repository and under a different artifact so it can be reused in other annotation processors

sjaakd:
@filiphr , @gunnarmorling ; in your opinion, would this lib than only contain the `AstModifyingAnnotationProcessor` or all offered MapStruct SPI?

filiphr:
In my opinion only this SPI (`AstModifyingAnnotationProcessor`) should go there. This processor is a special one and should be independent of MapStruct.

twentylemon:
Hey guys, this popped into my head so I decided to check on the status (every AP I think of always uses pojos...). I see that #1545 was closed without a reason. A new [repo](https://github.com/mapstruct/ast-modifying-spi) was made to house the interface, though the artifacts don't seem to be available in maven.

The pr looked reasonable to me, backward compatible even. Was the pr closed just due to priorities?
Thanks!  
<sub>I **don't** intend this to be a "do work for me pls." Just asking what's up.</sub>

PR #1545, opened 2018-07-11
#1384 Put AstModifyingAnnotationProcessor in a separate module for reuse by Lombok and the likes

sjaakd:
@filiphr @gunnarmorling ..trying ot isolate the `AstModifyingProcessor` into a separate maven module. Like requested in #1384 

Questions.
1. what kind of dependency scope should I use? Provided / Compile?
2. I only did the `AstModifyingProcessor`.. I reckon the remainder stays in the processor.
3. I'm struggling with the unit test. Do not understand why there's a difference between the annotations which seems to be on the class path directly in contrast to the new `astmodifying-spi'  module, which seems to be there only as maven dependency.. 
4. Is this what you think should be done?

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -28,6 +28,7 @@
         <org.springframework.version>4.0.3.RELEASE</org.springframework.version>
         <org.eclipse.tycho.compiler-jdt.version>0.26.0</org.eclipse.tycho.compiler-jdt.version>
         <com.puppycrawl.tools.checkstyle.version>7.2</com.puppycrawl.tools.checkstyle.version>
+        <org.aptintegration.tools.ast-modifying-spi.version>1.0.0-SNAPSHOT</org.aptintegration.tools.ast-modifying-spi.version>
         <add.release.arguments />
         <forkCount>1</forkCount>
         <assertj.version>3.7.0</assertj.version>
@@ -256,6 +257,11 @@
                 <groupId>org.apache.maven.shared</groupId>
                 <artifactId>maven-verifier</artifactId>
                 <version>1.5</version>
+            </dependency>
+            <dependency>
+                <groupId>org.aptintegration.tools</groupId>
+                <artifactId>ast-modifying-spi</artifactId>
+                <version>${org.aptintegration.tools.ast-modifying-spi.version}</version>
             </dependency>
 
             <!-- Not directly used but required for dependency convergence -->
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -116,6 +116,12 @@
             <artifactId>joda-time</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <!-- Integration with Lombok -->
+        <dependency>
+            <groupId>org.aptintegration.tools</groupId>
+            <artifactId>ast-modifying-spi</artifactId>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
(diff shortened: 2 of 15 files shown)


sjaakd:
@agudian .. You are familiar with the test helper classes.. How do I get this new MapStruct submodule on the class path directly (not as maven dependency) like the MapStruct core (api) packages? It is not in the list that's filtered.

filiphr:
@sjaakd I think that the magic happens in [`CompilingStatement#buildProcessorClasspath`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java#L149)

As for the other questions:

1. It should be compile as we depend on it non optionally
1. I would say that only the `AstModifyingProcessor` should be moved
1. The tests are failing because the `AstModifyingProcessor` is not on the processor path. See my reply above
1. The `AstModifyingProcessor` can't stay in the same package, as that is not allowed by JPMS. We need to think what the best approach for this would be. Maybe it needs to go into a repo of it's own

sjaakd:
> @sjaakd I think that the magic happens in CompilingStatement#buildProcessorClasspath

Tried that.. but its only white-listing.. See my question to @agudian . So an already existing class path is filtered on your referred lines. But how does it get on the class path in the first place? Its not the maven dependency mechanism, that gives the current result..  Just place a breakpoint and see how the core listed in contrast to the API.

> It should be compile as we depend on it non optionally

OK. 



> The AstModifyingProcessor can't stay in the same package, as that is not allowed by JPMS. We need to think what the best approach for this would be. Maybe it needs to go into a repo of it's own

Did not know that.. I should get familiar with java9.

sjaakd:
> @sjaakd I think that the magic happens in CompilingStatement#buildProcessorClasspath

Now it works.. Tried it several times before.. strange..

filiphr:
@sjaakd I am not sure what the correct way of doing this is. If we move the package people will have to update to a new Lombok that would have this update. On top of that, I don't think that we need to have this module in the mapstruct repository. We should extract it out if possible.

I am not entirely sure about the package name as well. As this is not really an integration with MapStruct, but it is more about APT that are not according to the spec, so we just invoke implementations of it. 

This change has to be done in sync with the Lombok team.

There is one way to achieve a good way for this and that is to make the current `AstModifyingProcessor` deprecated and extend from the new one. This way we can get all implementations of the new SPI (not sure how this actually works as SPIs are defined in META-INF, so we might need to get both the old and the new SPI

sjaakd:
> We should extract it out if possible.

I'm not sure whether its worthwhile to call a new organisation into existence in GitHub. But we could use a personal GitHub account to do so (believe, if I recollect correctly, Lombok is [on a personal GitHub account](https://github.com/rzwitserloot/lombok). So we might move it out to yours or mine user account.

filiphr:
We don't need a new organization. It can be part of the mapstruct organization, but new repository. Just like the plugins and examples we have

sjaakd:
> We don't need a new organization. It can be part of the mapstruct organization, but new repository. Just like the plugins and examples we have

i was thinking: if its true integration functionality... its perhaps better to entirely decouple it from MapStruct. Anyway .. that's more like a psychological thing..

sjaakd:
> This way we can get all implementations of the new SPI (not sure how this actually works as SPIs are defined in META-INF, so we might need to get both the old and the new SPI

I reckon you have to load both of them.. the deprecated one and the new one. The extension only takes care of being able to write the handling code efficiently. But then again.. its a couple of lines where the `AstModifyingAnnotationProcessor ` is actually used

sjaakd:
I leave this one for the time being....

sjaakd:
Ok.. still need to add some integration test to see whether this flies.

Issue #1499, opened 2018-05-20
Add Integration tests with Protobuf


filiphr:
In 1.3.0.Beta1 we are adding support for builders. It would be good to have some integration tests with Protobuf just to make sure that they work there properly

filiphr:
Test added in https://github.com/mapstruct/mapstruct/pull/1549 thanks @chris922

PR #1549, opened 2018-07-14
#1499 Add Protobuf builder integration test

### LINKED ISSUE #1449: Adder method not working for builders

chris922:
Added a protobuf builder integration test. 

I generated the protobuf Java files by using `protoc` locally and committed it. Not the best to have generated code on git, but the `protobuf-maven-plugin` is just using the `protoc` compiler which must be available on the system that is running this test.

Unfortunately locally with the eclipse_jdt_java_8 compiler I am getting an error that looks like this compiler can't compile the generated protobuf Java class... haven't yet disabled this compiler as I would like to see if Travis is able to built it.

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project protobufIntegrationTest: Compilation failure: Compilation failure:
[ERROR] C:\development\git-repos\mapstruct\integrationtest\target\tmp\org.mapstruct.itest.tests.ProtobufBuilderTest\eclipse_jdt_java_8\src\main\java\org\mapstruct\itest\protobuf\PersonProtos.java:[60]
[ERROR] public  static final class Person extends
[ERROR] ^^^^^^
[ERROR] The return types are incompatible for the inherited methods MessageOrBuilder.getAllFields(), GeneratedMessageV3.getAllFields()
[ERROR] C:\development\git-repos\mapstruct\integrationtest\target\tmp\org.mapstruct.itest.tests.ProtobufBuilderTest\eclipse_jdt_java_8\src\main\java\org\mapstruct\itest\protobuf\PersonProtos.java:[175]
[ERROR] public  static final class Address extends
[ERROR] ^^^^^^^
[ERROR] The return types are incompatible for the inherited methods MessageOrBuilder.getAllFields(), GeneratedMessageV3.getAllFields()
[ERROR] C:\development\git-repos\mapstruct\integrationtest\target\tmp\org.mapstruct.itest.tests.ProtobufBuilderTest\eclipse_jdt_java_8\src\main\java\org\mapstruct\itest\protobuf\PersonProtos.java:[460]
[ERROR] public static final class Builder extends
[ERROR] ^^^^^^^
[ERROR] The return types are incompatible for the inherited methods MessageOrBuilder.getAllFields(), GeneratedMessageV3.Builder<PersonProtos.Person.Address.Builder>.getAllFields()
[ERROR] C:\development\git-repos\mapstruct\integrationtest\target\tmp\org.mapstruct.itest.tests.ProtobufBuilderTest\eclipse_jdt_java_8\src\main\java\org\mapstruct\itest\protobuf\PersonProtos.java:[1028]
[ERROR] public static final class Builder extends
[ERROR] ^^^^^^^
[ERROR] The return types are incompatible for the inherited methods MessageOrBuilder.getAllFields(), GeneratedMessageV3.Builder<PersonProtos.Person.Builder>.getAllFields()
[ERROR] 4 problems (4 errors)
```

Fixes #1449


chris922:
Same issue, I disabled the `eclipse_jdt_java_8` compiler.

Looks really like a bug that prevents using protobuf together with the jdt compiler, see: https://bugs.eclipse.org/bugs/show_bug.cgi?id=496591

The bug is closed, so I tried to update the tycho-jdt-compiler version which seems to fix this issue, but this will break other tests .. :/

In case it is preferred to update the tycho-jdt-compiler instead of disabling the jdt 8 processing option I can try to analyze why the other tests are now failing (looks like there are some issues with generics and inheritance)

chris922:
Thanks for the example! I saw this plugin but thought I need to have the `protoc` locally available, good to know that there are some configuration things to fetch them dynamically. I updated the PR .. let's see if travis likes my changes as well.

I will create an issue for the tycho compiler topic.

filiphr:
Thanks for the PR @chris922 I've integrated it into master

Issue #1541, opened 2018-07-09
NPE when using varargs in mapping method


filiphr:
From [this](https://stackoverflow.com/questions/51241842/mapstruct-using-varags-not-working) SO question.

The mapping:

```
@Mappings({
    @Mapping(source = "code", target = "code")
})
MyResponse mapInformationToErrorData(String code, String... parameters);
```

Stacktrace:

```
ErrorMapper.java:[20,8] Internal error in the mapping processor: java.lang.NullPointerException at 
org.mapstruct.ap.internal.util.workarounds.SpecificCompilerWorkarounds.replaceTypeElementIfNecessary(SpecificCompilerWorkarounds.java:112) at 
org.mapstruct.ap.internal.util.Executables.getAllEnclosedAccessors(Executables.java:208) at 
org.mapstruct.ap.internal.model.common.Type.getAllAccessors(Type.java:534) at 
org.mapstruct.ap.internal.model.common.Type.getPropertyReadAccessors(Type.java:397) at 
org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.applyPropertyNameBasedMapping(BeanMappingMethod.java:529) at 
org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.build(BeanMappingMethod.java:145) at 
org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMappingMethods(MapperCreationProcessor.java:373) at 
org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMapper(MapperCreationProcessor.java:152) at 
org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:123) at 
org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:80) at 
org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:280) at 
org.mapstruct.ap.MappingProcessor.processMapperTypeElement(MappingProcessor.java:260) at 
org.mapstruct.ap.MappingProcessor.processMapperElements(MappingProcessor.java:226) at 
org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:162) at 
com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:794) at 
com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:705) at 
[...]

PR #1550, opened 2018-07-15
#1541 Fix NPE when using varargs in mapping methods

chris922:
* Fixed NPE reported in ticket, but afterwards struggled around because the generated method must use varargs, otherwise the eclipse jdt compiler fails -> please check if this fix was correct (two added `isArrayType()` checks in `BeanMappingMethod`)
* Added `isVarArgs` to a `Type`
* In case `isVarArgs` is `true` and the code for the last parameter will be generated the last `[]` will be replaced by `...` - this means in other cases where the `Type` will be used it will still be written in array-notation (this was important when the `Type` will be stored in a local variable)


processor/src/test/java/org/mapstruct/ap/test/bugs/_1541/Issue1541Mapper.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @AfterMapping
+    @Named( "afterMappingParametersAsArray" )
+    protected void afterMappingParametersAsArray(@MappingTarget Target target, String[] parameters) {
+        target.setAfterMappingWithArrayCalled( true );
+        target.setParameters( parameters );
+    }
+
+    @AfterMapping
+    @Named( "afterMappingParametersAsVarArgs" )
```
filiphr: Why do we need the `@AfterMapping` to be `@Named`?

processor/src/test/java/org/mapstruct/ap/test/bugs/_1541/Issue1541Test.java
```diff
@@ -0,0 +1,112 @@
...
+public class Issue1541Test {
+
+    @Test
+    public void testMappingWithVarArgs() {
+        Target target = Issue1541Mapper.INSTANCE.mapWithVarArgs( "code", "1", "2" );
+
+        assertThat( target ).isNotNull();
+        assertThat( target.getCode() ).isEqualTo( "code" );
+        assertThat( target.getParameters() ).contains( "1", "2" );
+        assertThat( target.isAfterMappingWithArrayCalled() ).isFalse();
```
filiphr: Why are these false? By looking into the `@AfterMapping` I would assume that they are called. OK I see the `@BeanMapping` and `qualifiedByName` now. However, it seems a bit strange and not obvious when looking at the code.

processor/src/main/resources/org/mapstruct/ap/internal/model/common/Parameter.ftl
```diff
@@ -19,4 +19,4 @@
      limitations under the License.
 
 -->
-<@includeModel object=type/> ${name}
\ No newline at end of file
+<@includeModel object=type allowVarArgs=ext.allowVarArgs!false/> ${name}
```
filiphr: Does the `!` in `ext.allowVarArgs!false` mean if it is not set use `false`? Like default?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -129,7 +128,7 @@ private Builder setupMethodWithMapping(Method sourceMethod) {
             for ( Parameter sourceParameter : method.getSourceParameters() ) {
                 unprocessedSourceParameters.add( sourceParameter );
 
-                if ( sourceParameter.getType().isPrimitive() ) {
+                if ( sourceParameter.getType().isPrimitive() || sourceParameter.getType().isArrayType() ) {
```
filiphr: This is and line 573 are the actual fixes for the NPE right?

sjaakd:
Oops. Saw @filiphr made the same remark.. :smile:

processor/src/test/java/org/mapstruct/ap/test/bugs/_1541/Issue1541Mapper.java
chris922: It's not required, I just wanted to be sure to not have both after mappings automatically in all my mapping methods available - only in the two where I explicitly want to test this case.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1541/Issue1541Test.java
chris922: Yes that's the reason. The `@AfterMapping`s are only used when they are `qualifiedByName` (at first I also thought that they will be used automatically in case no `qualifiedByName` value is given.. but looks like an `@AfterMapping` is not automatically used when it is `@Named`)

processor/src/main/resources/org/mapstruct/ap/internal/model/common/Parameter.ftl
chris922: Exactly, otherwise Freemarker is complaining when `allowVarArgs` is not available

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
chris922: Exactly! Looks for me like it was tried to find getter/setters even if it is an array. This will be skipped there for primitives and thus I also added the skip for arrays. I am not so deep in this internal stuff, that's the reason I asked that someone else should check this as well. At least I haven't destroyed a different testcase. So hopefully I analyzed and fixed this correctly.

chris922:
Thanks for the feedback! Good idea with moving it to the `Parameter`, there it definitly makes more sense.

And @sjaakd: Thanks for the other suggestions for test cases. I will add some more. Using a varargs `@Context` sounds interesting, I will try it.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1541/Issue1541Test.java
filiphr: Yes when a qualifier is used they are explicitly ignored

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
filiphr: The analysis was great. This is exact place where the fix should come. I was just curious whether there was something additional, but it seems not

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
```diff
@@ -119,12 +126,18 @@ public boolean equals(Object o) {
     }
 
     public static Parameter forElementAndType(VariableElement element, Type parameterType) {
```
filiphr: This method is not used anymore. I think that there was a single usage of it and now that usage uses the overloaded method with `isVarArgs`

chris922:
Thanks @sjaakd for the idea with the varargs  `@Context` - there was indeed also a NPE.

I updated the test cases. What do you meant with the following?

> you might want to exclude @TargetType and @MappingTarget.

sjaakd:
> What do you meant with the following?

It would be strange to mark a `vararg` as `@TargetType` or `@MappingTarget`. What would that mean? For:
1. `@TargetType` you need to specify a class for which the mapping method is supposed to generate an instance as return value (singular).
2. `@MappingTarget` is used in case of so-called-merge-methods, to indicate a pre-existing target which should be populated by MapStruct. Also singular.

So we should possibly give an error message when you try to use such combination: "`@TargetType` not possible in combination with a vararg"  or something along those lines.

chris922:
Aaah understood! Should we raise an own issue for this or do we see this PR as adding support for varargs and it should be added here ?
Maybe I find tomorrow a few minutes to check if I can find the right place for such an error message quickly. But imho this could be handled independently.

sjaakd:
@filiphr race condition in answering.. :smile:. If you return an array its not a bean mapping I guess.  updates of iterable types is still not supported (#487)..

sjaakd:
@chris922 : i merged this one and opened a new issue #1562.

Issue #1456, opened 2018-04-24
Apply source presence tracking recursively


birdfriend:
Hi,

We want to apply source presence tracking recursively. For example:
//source object
```
public class SourceA {
  private String fieldA;
  private SourceB fieldB;

//getters & setters & hasXXX methods
...
}

public class SourceB {
  private String fieldC;
//getters & setters & hasXXX methods
...
}

//target
public class TargetObj {
  private String fieldA;
  private String fieldC;
//getters & setters
...
}
```
//mapper
```
@Mapper(nullValueCheckStrategy =  NullValueCheckStrategy.ALWAYS)
public interface TestMapper {

  @Mapping(source = "fieldB.fieldC", target = "fieldC")
  TargetObj testMap(SourceA source);
}
```

//The auto generated code:
```
public TargetObj testMap(SourceA source) {
    if(source == null) {
      return null;
    } else {
      TargetObj targetObj = new TargetObj();
      String fieldC = this.sourceFieldBFieldC(source);
      if(source.hasFieldB()) {
        targetObj.setFieldC(fieldC);
      }

      if(source.hasFieldA()) {
        targetObj.setFieldA(source.getFieldA());
      }

      return targetObj;
    }
  }
...
```

When mapping `fieldC`, source presence tracking has only applied to it's parent object,
```
if(source.hasFieldB()) {
        targetObj.setFieldC(fieldC);
      }
```
However we want to apply source presence tracking recursively:
```
if(source.hasFieldB() && source.getFieldB().hasFieldC()) {
        targetObj.setFieldC(fieldC);
      }
```

Is this some feature you can support? Thanks!

filiphr:
How does the `sourceFieldBFieldC` method looks like? 

What you are asking is quite complex to achieve as that can blow up. 

I think that in this case we should not do the `hasFieldB` check, but just check if the result  `field` is not `null`. As the has checks should have already been done in the nested source mapping method.

birdfriend:
auto-generated `sourceFieldBFieldC` looks like:
```
  private String sourceFieldBFieldC(SourceA sourceA) {
    if(sourceA == null) {
      return null;
    } else if(!sourceA.hasFieldB()) {
      return null;
    } else {
      SourceB fieldB = sourceA.getFieldB();
      if(!fieldB.hasFieldC()) {
        return null;
      } else {
        String fieldC = fieldB.getFieldC();
        return fieldC;
      }
    }
  }
```

[...]

PR #1555, opened 2018-07-17
#1456, apply source presence tracking recursively

birdfriend:
When mapping nested objects with source presence tracking enabled, source presence tracking is only enabled for the top level objects. It should be expected that source presence tracking can be applied to all nested levels.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -563,6 +563,20 @@ public class PropertyMapping extends ModelElement {
                 if ( propertyEntry.getPresenceChecker() != null ) {
                     sourcePresenceChecker = sourceParam.getName()
                         + "." + propertyEntry.getPresenceChecker().getSimpleName() + "()";
+
+                    String variableName = sourceParam.getName() + "."
+                        + propertyEntry.getReadAccessor().getSimpleName() + "()";
+                    for (int i = 1; i < sourceReference.getPropertyEntries().size(); i++) {
+                        PropertyEntry entry = sourceReference.getPropertyEntries().get( i );
+                        if (entry.getPresenceChecker() != null && entry.getReadAccessor() != null) {
+                            sourcePresenceChecker += " && " + variableName + " != null && "
+                                + variableName + "." + entry.getPresenceChecker().getSimpleName() + "()";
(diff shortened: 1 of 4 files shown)


birdfriend:
@filiphr - can you help us out with some feedback? Thanks!

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -563,6 +563,20 @@ private String getSourcePresenceCheckerRef( SourceReference sourceReference ) {
                 if ( propertyEntry.getPresenceChecker() != null ) {
                     sourcePresenceChecker = sourceParam.getName()
                         + "." + propertyEntry.getPresenceChecker().getSimpleName() + "()";
+
```
sjaakd: I think it would be good to extract this code and put it into its own method. Something like: `constructSourcePresenceCheck` don't you think?

processor/src/test/java/org/mapstruct/ap/test/source/presencecheck/spi/SoccerTeamMapperNestedObjects.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * @author Cindy Wang
+ */
+@Mapper
+public interface SoccerTeamMapperNestedObjects {
+    SoccerTeamMapperNestedObjects INSTANCE = Mappers.getMapper( SoccerTeamMapperNestedObjects.class );
+
+    @Mappings({
+        @Mapping(target = "players", ignore = true),
+        @Mapping(target = "goalKeeperName", source = "goalKeeper.name")
```
sjaakd: just out of curiosity: what would happen if the target is also nested? Should we add an `@Mapping` for this as well just to be sure it works? So: `@Mapping( target = "referee.name", source = "refereeName" )`

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: I saw its context now.. Its already a small method. So.. lets leave it like you propose.

sjaakd:
Hi @birdfriend : I did add the additional test case (and a small extra check to the original one). . I created a new PR. When its ready, I'll merge it.. Could you check if I got your name correctly in the copyright.txt file?

Thanks for your contribution!

sjaakd:
@thanks @birdfriend

birdfriend:
thanks @sjaakd !

PR #1563, opened 2018-07-24
#1558 Annotations `ElementType.TYPE_USE` not handled correctly 

### LINKED ISSUE #1558: Generated code fails to build when the source type is a byte array annotated with @NotNull
### Source class
```java
public class Car2 {
    private byte[] data;

    @NotNull
    public byte[] getData() {
        return data;
    }
    public void setData(byte[] data) {
        this.data = data;
    }
}
```

### Target class
```java
public class Car {
    byte[] data;

    public byte[] getData() {
        return data;
    }

    public void setData(byte[] data) {
        this.data = data;
    }
}
```

### Mapper
```java
@Mapper
public interface CarMapper {
    Car toCar(Car2 car2);
}
```

### Result
Generated source:
```java
        Car car = new Car();

            (@javax.validation.constraints.NotNull :: byte)[] data = car2.getData();
        if ( data != null ) {
            car.setData( Arrays.copyOf( data, data.length ) );
        }

        return car;
```

The generated source code fails to build with the following error
```
[ERROR] target/generated-sources/annotations/test/CarMapperImpl.java:[21,52] illegal start of expression
[ERROR] target/generated-sources/annotations/test/CarMapperImpl.java:[21,54] ')' expected
[ERROR] target/generated-sources/annotations/test/CarMapperImpl.java:[21,13] not a statement
[ERROR] target/generated-sources/annotations/test/CarMapperImpl.java:[21,59] ';' expected
[ERROR] target/generated-sources/annotations/test/CarMapperImpl.java:[21,60] illegal start of expression
[ERROR] target/generated-sources/annotations/test/CarMapperImpl.java:[21,61] ';' expected
```

This happens with both version 1.2.0.Final and 1.3.0.Beta1

Reproducer source  code: [buildfailurereproducer.zip](https://github.com/mapstruct/mapstruct/files/2219566/buildfailurereproducer.zip)

sjaakd:
There's something strange ongoing in JDK (not in Eclipse) when there's an annotation on a method. The return type seems to be influenced by this annotation. A reproducer is added. The reproduce is a java 8 reproducer, but that's probably not necessary. I created a copy of the `@NotNull` locally. That works as well as far is reproducing is concerned.

Feel free to continue on this one and get to the bottom. 

Fixes #1558

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
@@ -11,7 +11,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.NavigableMap;
 import java.util.NavigableSet;
@@ -46,6 +45,7 @@ import org.mapstruct.ap.internal.util.Extractor;
 import org.mapstruct.ap.internal.util.FormattingMessager;
 import org.mapstruct.ap.internal.util.JavaStreamConstants;
 import org.mapstruct.ap.internal.util.Message;
+import org.mapstruct.ap.internal.util.NativeTypes;
 import org.mapstruct.ap.internal.util.RoundContext;
 import org.mapstruct.ap.internal.util.Strings;
 import org.mapstruct.ap.internal.util.accessor.Accessor;
@@ -242,8 +242,9 @@ public class TypeFactory {
             else if (componentTypeMirror.getKind().isPrimitive()) {
                 // When the component type is primitive and is annotated with ElementType.TYPE_USE then
                 // the typeMirror#toString returns (@CustomAnnotation :: byte) for the javac compiler
-                name = componentTypeMirror.getKind().name().toLowerCase( Locale.ROOT ) + builder.toString();
+                name = NativeTypes.getName( componentTypeMirror.getKind() ) + builder.toString();
                 packageName = null;
+                // for primitive types only name (e.g. byte, short..) required as qualified name
                 qualifiedName = name;
                 isImported = false;
             }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
@@ -242,7 +242,7 @@ public class TypeFactory {
             else if (componentTypeMirror.getKind().isPrimitive()) {
                 // When the component type is primitive and is annotated with ElementType.TYPE_USE then
                 // the typeMirror#toString returns (@CustomAnnotation :: byte) for the javac compiler
-                name = componentTypeMirror.getKind().name().toLowerCase( Locale.ROOT) + builder.toString();
+                name = componentTypeMirror.getKind().name().toLowerCase( Locale.ROOT ) + builder.toString();
                 packageName = null;
                 qualifiedName = name;
                 isImported = false;
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
@@ -11,6 +11,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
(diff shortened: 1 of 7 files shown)


filiphr:
I think this should fix the problem. Java 8 was indeed necessary it only happens when the annotation is for `ElementType.TYPE_VAR`, in this case javac uses something else and has a different `toString` method. The fix should be enough

filiphr:
This was not a bug per se in the javac compiler. Have a look at the fix. In our `TypeFactory`, when we are looking for the name of the component type we were using `TypeMirror#toString() `, and now we use something else when the type is primitive.

sjaakd:
I saw that. By but Tycho behaves according expected behaviour. It's oracle that deviates

filiphr:
That's just a coincidence in my opinion. I don't think that we should rely on `toString`, there is no specification foe that (I think)

processor/src/test/java/org/mapstruct/ap/test/bugs/_1558/java8/NotNull.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({
+    ElementType.METHOD,
+    ElementType.TYPE_USE
```
sjaakd: Is this problem only occurring with 'TYPE_USE'?

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -238,6 +239,14 @@ else if ( mirror.getKind() == TypeKind.ARRAY ) {
...
+            else if (componentTypeMirror.getKind().isPrimitive()) {
+                // When the component type is primitive and is annotated with ElementType.TYPE_USE then
+                // the typeMirror#toString returns (@CustomAnnotation :: byte) for the javac compiler
+                name = componentTypeMirror.getKind().name().toLowerCase( Locale.ROOT ) + builder.toString();
+                packageName = null;
+                qualifiedName = name;
+                isImported = false;
+            }
             else {
                 name = mirror.toString();
```
sjaakd: so.. why only primitive types.. ?

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -238,6 +239,14 @@ else if ( mirror.getKind() == TypeKind.ARRAY ) {
                 qualifiedName = componentTypeElement.getQualifiedName().toString() + arraySuffix;
                 isImported = isImported( name, qualifiedName );
             }
+            else if (componentTypeMirror.getKind().isPrimitive()) {
+                // When the component type is primitive and is annotated with ElementType.TYPE_USE then
```
sjaakd: wouldn't it be better to check for the annotation target type then?

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -238,6 +239,14 @@ else if ( mirror.getKind() == TypeKind.ARRAY ) {
...
+                // the typeMirror#toString returns (@CustomAnnotation :: byte) for the javac compiler
+                name = componentTypeMirror.getKind().name().toLowerCase( Locale.ROOT ) + builder.toString();
```
sjaakd: consider adding a map to `org.mapstruct.ap.internal.util.NativeTypes`. Mapping `TypeKind` to native type instead of doing string conversion.

filiphr:
Thanks for the review @sjaakd. I have applied your comments. Can you please rebase the PR and merge my 3 commits into one and change the message of the commit to start with `#1558` I did a mistake and I can't force push to your branch 😄. If it is easier, I can open a new PR and we can close this one.

sjaakd:
> Thanks for the review @sjaakd. I have applied your comments. Can you please rebase the PR and merge my 3 commits into one and change the message of the commit to start with #1558 I did a mistake and I can't force push to your branch 😄.

No problem..  I'll wait until travis is finished..

sjaakd:
Hmm. If you use the web interface, the authors are not correctly reflected.. Sorry @filiphr ..

filiphr:
No worries @sjaakd. If you do squash and merge the author of the PR would be the author :)

PR #1567, opened 2018-07-29
Add support for initializing the AccessorNamingStrategy with Elements and Types and use Types for determining fluent setters

### LINKED ISSUE #1253: Access to Types and Elements in AccessorNamingStrategy
Looking at the implementation of the example for the Protobuf `AccessorNamingStrategy` I think that it could benefit if it could have used `javax.lang.model.util.Types` and `javax.lang.model.util.Elements`.

If we had those than instead of:

```java
@Override
public String getElementName(ExecutableElement adderMethod) {

    String methodName = super.getElementName(adderMethod);
    Element receiver = adderMethod.getEnclosingElement();
    if (receiver != null && receiver.getKind() == ElementKind.CLASS) {
        TypeElement type = (TypeElement) receiver;
        TypeMirror superType = type.getSuperclass();
        if (superType != null && PROTOBUF_GENERATED_MESSAGE_V3.equals(superType.toString())) {
            methodName += LIST_SUFFIX;
        }
    }
    return methodName;
}
```

We could have had:
```java
@Override
public String getElementName(ExecutableElement adderMethod) {

    String methodName = super.getElementName(adderMethod);
    Element receiver = adderMethod.getEnclosingElement();
    TypeElement protoGenMessageV3 = elements.getTypeElement(PROTOBUF_GENERATED_MESSAGE_V3);
    if (receiver != null && types.isAssignable(receiver.asType(), protoGenMessage.asType())) {
        methodName += LIST_SUFFIX;        
    }
    return methodName;
}
```

Of course the implementation can be even more optimized by storing the `protoGenMessageV3` in the class.

I see two ways of achieving this:

* Add a new method to the `AccessorNamingStrategy` with which we can set the `Types` and `Elements` 
* Pass the `Types` and `Elements` to the current methods

### LINKED ISSUE #1566: Setters on a Builder super class are not called
As per the following example if my Builder extends AbstractBuilder then the 'id' setter will not be called

```java
    static abstract class AbstractBuilder<T extends AbstractBuilder<T>> {
        String id;

        public T id(final String id) {
            this.id = id;
            return (T) this;
        }
    }
```

filiphr:
This fix for this actually depends on #1253. However, I am afraid that we might need to do something more involved like the [`MethodMatcher`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/source/MethodMatcher.java). I am not 100% certain whether the `isAssignable` hint added would be enough. Maybe @agudian and @sjaakd can chime in on this as well.

Fixes #1566 
Fixes #1253

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/spi/AccessorNamingStrategy.java
+++ b/processor/src/main/java/org/mapstruct/ap/spi/AccessorNamingStrategy.java
@@ -16,6 +16,15 @@ import javax.lang.model.element.ExecutableElement;
 public interface AccessorNamingStrategy {
 
     /**
+     * Initializes the accessor naming strategy with the MapStruct processing environment.
+     *
+     * @param processingEnvironment environment for facilities
+     */
+    default void init(MapStructProcessingEnvironment processingEnvironment) {
+
+    }
+
+    /**
      * Returns the type of the given method.
      *
      * @param method to be analyzed.
--- a/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
+++ b/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
@@ -12,8 +12,10 @@ import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Elements;
 import javax.lang.model.util.SimpleElementVisitor6;
 import javax.lang.model.util.SimpleTypeVisitor6;
+import javax.lang.model.util.Types;
 
 import org.mapstruct.ap.spi.util.IntrospectorUtils;
 
@@ -25,6 +27,15 @@ import org.mapstruct.ap.spi.util.IntrospectorUtils;
 public class DefaultAccessorNamingStrategy implements AccessorNamingStrategy {
 
     private static final Pattern JAVA_JAVAX_PACKAGE = Pattern.compile( "^javax?\\..*" );
+
+    protected Elements elementUtils;
+    protected Types typeUtils;
+
+    @Override
+    public void init(MapStructProcessingEnvironment processingEnvironment) {
+        this.elementUtils = processingEnvironment.getElementUtils();
+        this.typeUtils = processingEnvironment.getTypeUtils();
+    }
 
     @Override
     public MethodType getMethodType(ExecutableElement method) {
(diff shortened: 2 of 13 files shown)


sjaakd:
> Maybe @agudian and @sjaakd can chime in on this as well.

I'll get back on this one..

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
```diff
@@ -90,6 +90,7 @@ protected boolean isBuilderSetter(ExecutableElement method) {
         return method.getParameters().size() == 1 &&
             !JAVA_JAVAX_PACKAGE.matcher( method.getEnclosingElement().asType().toString() ).matches() &&
             !isAdderWithUpperCase4thCharacter( method ) &&
+            //typeUtils.isAssignable( method.getReturnType(), method.getEnclosingElement().asType() )
```
sjaakd: have you tried `typeUtils#asMemberOf`? If you have a return type or method parameter part of a containing executable-type or containing class you can evaluate the generic arguments as member of this containing type...

sjaakd: Or.. we should not give a generic `ExecutableElement` but a fully resolved one to this method. Don't know if that's possible as well..

filiphr: That might actually work. Although I am not sure what it might mean from a performance point of view. As we would need to resolve all methods, even though they might never be used.

sjaakd: I added some comments to #1253

filiphr:
Seems like FreeBuilder generates bunch of different methods, similar to Immutables. The difference is that Immutables generates only `from`, whereas `FreeBuilder` generates much more. We probably need a custom strategy for `FreeBuilder`

processor/src/main/java/org/mapstruct/ap/spi/AccessorNamingStrategy.java
```diff
@@ -15,6 +17,16 @@
  */
 public interface AccessorNamingStrategy {
 
+    /**
+     * Initializes the accessor naming strategy with the utility methods for operating on elements and types.
+     *
+     * @param elementUtils utility methods for operating on elements
+     * @param typeUtils utility methods for operating on types
+     */
+    default void init(Elements elementUtils, Types typeUtils) {
```
sjaakd: Should we wrap both `Elements` and `Types` in a bean so that we can extend it without changing the signature with other utils when needed in the future? I'm not sure myself here...

filiphr: I was actually thinking the same and was not sure what the best approach is. I'll wrap in in some bean so we are more flexible in the future. By the way we might need to do something similar for the `BuilderProvider` (just so we are consistent)

filiphr: @sjaakd just updated this. Can you have another look? Not 100% sure about the name of the new interface though

sjaakd: you mean `  public void init(MapStructProcessingEnvironment processingEnvironment) `.. LGTM.. What are you doubting about?

filiphr: The name of `MapStructProcessingEnvironment ` of the interface. Just pushed new changes it should be OK to add a new default method to an interface now (`japicmp-maven-plugin` was marking that as binary incompatible)

sjaakd:
LGTM.. go ahead and merge..

PR #1587, opened 2018-08-18
Update tycho-compiler-jdt to latest 1.2.0 version

### LINKED ISSUE #1553: Update tycho-compiler-jdt version
During development of #1549 I noticed that there is a new version 1.2.0 for the `tycho-compiler-jdt` dependency, because the version that MapStruct uses right now (0.26.0) has a bug which prevents running the integration tests with the jdt_java8 compiler.

Upgrading to 1.2.0 wasn't done with just changing the version number. For a few other dependencies I had to stick the version number, because the `maven-enforcer-plugin` complained mixed version numbers.
Nevertheless some test cases were broken afterwards. It looked like issues with generics (`GenericsHierarchyTest` failed), maybe some more issues are there as well.

MapStruct should update to the newest version. Afterwards also the test for #1549 could be enabled for jdt_java8.

filiphr:
Add forced dependencies to force correct resolution.

The `GenericsHierarchyTest#determinesAnimalKeyGetter` fails with the Eclipse compiler. The problem is that for `AbstractAnimal` the `getKey` method is resolved to the one from `Identifiable<KeyOfAllBeings>`. However, in javac it is resolved to the one from `IAnimal<AnimalKey>`. Not sure if this is a bug in javac or ecj. Maybe @agudian can weigh on his experience 😄. 

The problem occurs due in:

https://github.com/mapstruct/mapstruct/blob/04576de1d1689aac439c092164c1fef7175e57c3/processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java#L227-L244

On line 234 javac returns `true`, but ecj returns `false`

Fixes #1553

### DIFF
--- a/integrationtest/src/test/java/org/mapstruct/itest/tests/MavenIntegrationTest.java
+++ b/integrationtest/src/test/java/org/mapstruct/itest/tests/MavenIntegrationTest.java
@@ -36,8 +36,7 @@ public class MavenIntegrationTest {
     }
 
     @ProcessorTest(baseDir = "freeBuilderBuilderTest", processorTypes = {
-        ProcessorTest.ProcessorType.JAVAC,
-        ProcessorTest.ProcessorType.ECLIPSE_JDT
+        ProcessorTest.ProcessorType.JAVAC
     })
     void freeBuilderBuilderTest() {
     }
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -27,7 +27,7 @@
         <org.apache.maven.plugins.surefire.version>3.0.0-M3</org.apache.maven.plugins.surefire.version>
         <org.apache.maven.plugins.javadoc.version>3.1.0</org.apache.maven.plugins.javadoc.version>
         <org.springframework.version>4.0.3.RELEASE</org.springframework.version>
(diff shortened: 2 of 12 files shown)


filiphr:
I think that I have located the problem. I have opened https://bugs.eclipse.org/bugs/show_bug.cgi?id=540101 for the tycho-compiler-jdt. However, I think that the issue is in the JDT APT.

sjaakd:
@filiphr What is the status on this one?

filiphr:
Well it stills fails with upgraded JDT. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=540101. Perhaps we should disable that test on JDK 8 with the Eclipse plugin. WDYT?

sjaakd:
> Well it stills fails with upgraded JDT. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=540101. Perhaps we should disable that test on JDK 8 with the Eclipse plugin. WDYT?

I think.. given the time that this one has been open.. this is the best idea. Just document it and refer to this PR (closed)

filiphr:
I had to do some changes and adapt certain tests.

Now the builds are passing. Can you have a look before I merge @sjaakd?

filiphr:
Thanks for the review @sjaakd

Issue #1590, opened 2018-08-22
[Compilation Error] class  'java.util.ArrayList'  not found. MappersImpl cannot be compiled.


euclideanrn:
Hi
I think I have encountered an exception here:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:compile (default-compile) on project lst-targetad-service: Compilation failure
[ERROR] ***/target/generated-sources/annotations/***/***/***/***/beans/mapping/MappersImpl.java:[102,55] 找不到符号
[ERROR]   符号:   类 ArrayList
[ERROR]   位置: 类 ***.***.***.***.beans.mapping.MappersImpl
```

And here is my definition interface 'Mappers.java':
```java
@Mapper(componentModel = "spring",
        nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS,
        nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT,
        uses = MappingFactories.class
)
public interface Mappers {

    @Mappings({
            @Mapping(target = "gmtCreate"  , expression = "java( new java.util.Date() )"),
            @Mapping(target = "gmtModified", expression = "java( new java.util.Date() )")
    })
    ResourceTemplateDO mapToResourceTemplateDO(TemplateCommandParam source);

    @Mappings({
            @Mapping(target = "gmtCreate"  , expression = "java( new java.util.Date() )"),
            @Mapping(target = "gmtModified", expression = "java( new java.util.Date() )"),
            @Mapping(target = "templateId" , source     = "id"                          )
    })
    ResourceTemplateVersionDO mapToResourceTemplateVersionDO(TemplateCommandParam source);

    ResourceTemplateDO mapToResourceTemplateDO(TemplateQueryParam source);

    TemplateModel mapToTemplateModel(ResourceTemplateDO source);
    TemplateVersionModel mapToTemplateVersionModel(ResourceTemplateVersionDO source);

    default java.lang.String map(java.util.List<java.lang.String> value) {
        return JSON.toJSONString(value);
    }

    default java.util.List<java.lang.String> map(java.lang.String value) {
        return JSON.parseObject(value, new TypeReference<List<String>>(){}.getType());
    }
}
```

the generated class 'MappersImpl.class':
```java
package com.alibaba.lst.targetad.beans.mapping;

import *.dal.dataobj.ResourceTemplateDO;
import *.dal.dataobj.ResourceTemplateVersionDO;
[...]

PR #1598, opened 2018-08-27
#1590 Missing ArrayList import in case of NVMS.RETURN_DEFAULT

sjaakd:
Just trying to reproduce the problem..

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMapsWithNullCheck.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMapsWithNullCheck.java
@@ -63,6 +63,9 @@ public class SetterWrapperForCollectionsAndMapsWithNullCheck extends WrapperForC
         if (isDirectAssignment() || getSourcePresenceCheckerReference() == null ) {
             imported.addAll( getNullCheckLocalVarType().getImportTypes() );
         }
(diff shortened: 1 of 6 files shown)


sjaakd:
@chris922 : if you have time.. if you could do a review?

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMapsWithNullCheck.java
```diff
@@ -63,6 +63,9 @@ public SetterWrapperForCollectionsAndMapsWithNullCheck(Assignment decoratedAssig
         if (isDirectAssignment() || getSourcePresenceCheckerReference() == null ) {
             imported.addAll( getNullCheckLocalVarType().getImportTypes() );
         }
+        if ( isMapNullToDefault() ) {
+            imported.add( targetType.getImplementationType() );
```
chris922: In line 52 is a check if `targetType.getImplementationType()` is not `null` - so could it also be the case that at this point it is `null`? In this case we also need the check here to prevent adding `null` to this `Set`.

sjaakd: Good point. Will fix. Thanks

sjaakd:
Thanks for reviewing @filiphr and @chris922 !

PR #1604, opened 2018-09-06
1561 add support for adders in combination with streams

sebastianhaberey:


filiphr:
@sebastianhaberey the actual problem is the new `Type#isStream` method. When `isCollectionOrMapOrStream` is called then the `isStream` is also called and on Java 7 and 6 this class does not exist.

I would suggest that you use something like

https://github.com/mapstruct/mapstruct/blob/13aa94742183fc5eedb3c991b55336d1b63a68dc/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java#L99-L100

Apart from that everything else looks good

sebastianhaberey:
@filiphr thanks for looking over this. I fixed Type#isStream() like you proposed. I also had to find a way to refer to Stream.class in Type#getAdderForType() without importing Java 8 stream. Please have a look at that too - if you can think of a more elegant way, I'll be happy to change it. The build seems fine now :slightly_smiling_face: :+1:

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -625,46 +626,85 @@ else if ( accessorNaming.isGetterMethod( candidate ) || Executables.isFieldAcces
...
-                    candidates.add( adder );
-                }
-            }
+        List<Accessor> candidates;
+
+        if ( collectionProperty.isCollectionType() ) {
+            candidates = getAccessorCandidates( collectionProperty, Iterable.class );
+        }
+        else if ( collectionProperty.isStreamType() ) {
+            candidates = getAccessorCandidates( collectionProperty, getClassForName( JavaStreamConstants.STREAM_FQN ) );
```
filiphr: Here you can use `Stream.class` as the type is a `Stream` which means that the `Stream` is present. And It is safe to have that as on Java 6 and 7 the condition will never evaluate to `true` and thus it will never throw the `ClassNotFoundException`. You can then remove the `getClassFormName` method

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -745,14 +785,21 @@ private boolean correspondingSetterMethodExists(Accessor getterMethod,
...
+    private boolean isCollectionOrMapOrStream(Accessor getterMethod) {
+        return isCollection( getterMethod.getAccessedType() ) || isMap( getterMethod.getAccessedType() ) ||
+            isStream( getterMethod.getAccessedType() );
     }
 
     private boolean isCollection(TypeMirror candidate) {
         return isSubType( candidate, Collection.class );
     }
 
+    private boolean isStream(TypeMirror candidate) {
```
filiphr: The change you did in this method is enough to fix the problem on Java 6 and 7

sjaakd:
@sebastianhaberey , @filiphr .. I'll take a look and check if I can help out to get this merged..

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
sebastianhaberey: There is currently no import of ```java.util.stream.Stream``` (or any other Java 8 class) in the ```Type``` class. Using ```Stream.class``` here would introduce a Java 8 import. That was my reasoning behind this change. If you think the Java 8 import would not cause a problem, I am happy to revert this (including the ```getClassForName()``` method).

sebastianhaberey:
@filiphr @sjaakd sorry, I didn't get to have a look at this over the weekend. I added a comment above; if there's anything else I might do, just say the word. Thanks for your help.

sebastianhaberey:
@filiphr great! Thanks for the help. Happy to contribute to this project!

Issue #1571, opened 2018-07-31
Allow nullValueCheckStrategy on @BeanMapping t not have twice more classes/interfaces


soberich:
This one I could raise the very first simply because this limitation once brought me to Mapstruct tracker and it is quite big limitation. Not only you have to declare aka "Creators" and "Updaters" to do both operations, but also you need to inject those additionally into you services. Updaters with `nullValueCheckStrategy = ALWAYS` does magic with `patch`ing persistent entities. There is a wonderfully suitable "@BeanMapping" annotation - that would be great if able to specify it there.

sjaakd:
Should not be too difficult. I thought it was already in '@BeanMapping'.  Or perhaps there is already an issue for it.

sjaakd:
strangest thing.. the javadoc states that its also possible to specify on mapping level.. however that has never been implemented.

sjaakd:
@soberich .. give it a shot.. Its on the master. The option is now on `BeanMapping` level and on `Mapping` level.

Issue #1306, opened 2017-10-10
When source is null target should  NOT be set to null 


msundari:
Hi, I am trying to map nested properties using mapstruct 1.2.0.CR2. (Example map customer.address.houseNumber to userDTO.homeDTO.addressDTO.houseNo ).

Expectation : I do not want to set the addressDTO to null when customer.address is null. Since addressDTO contains "countyname" and other properties which are already set from other different sources

Please advice if there is property/setting that I could set so that the target it not set to null when source is null.

This is the mapping  I am using.

@Mapper( nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS )
public interface CustomerUserMapperNullCheck {

    @Mapping(source="address", target="homeDTO.addressDTO" )
    void mapCustomer(Customer customer, @MappingTarget  UserDTO userDTO)  ;
    
    @Mapping(source="houseNumber", target="houseNo" )
    void mapCustomerHouse(Address address, @MappingTarget  AddressDTO addrDTO)  ;
    
}

I initially tried in single mapping like below
 @Mapping(target="homeDTO.addressDTO.houseNo", source="address.houseNumber")
   abstract void mapCustomerHouse(Customer customer, @MappingTarget  UserDTO userDTO)  ; Then tried splitting up the mapping,  based on https://github.com/mapstruct/mapstruct/issues/649. 

Both  approaches does not produce the expected generated code.

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
**The generated code**

public class CustomerUserMapperNullCheckImpl implements CustomerUserMapperNullCheck {

    @Override
    public void mapCustomer(Customer customer, UserDTO userDTO) {
        if ( customer == null ) {
            return;
        }

        if ( userDTO.getHomeDTO() == null ) {
            userDTO.setHomeDTO( new HomeDTO() );
        }
        customerToHomeDTO( customer, userDTO.getHomeDTO() );
    }

    @Override
    public void mapCustomerHouse(Address address, AddressDTO addrDTO) {
        if ( address == null ) {
            return;
        }

        if ( address.getHouseNumber() != null ) {
            addrDTO.setHouseNo( address.getHouseNumber() );
        }
    }

    protected void customerToHomeDTO(Customer customer, HomeDTO mappingTarget) {
        if ( customer == null ) {
            return;
        }

        if ( customer.getAddress() != null ) {
[...]

Issue #879, opened 2016-09-01
Merge target ignore null or empty


wenerme:
I want a patch(very common in REST) update, which means ignore null(or empty collection/string) field from source, only map non-null field to target.

User{name,age,children}

```
@BeanMapping(ignoreNull=true, ignoreEmptyCollection=true, ingoreEmptyString=true)
User patch(User s,@MappingTarget User t){
  if(s.getName() != null && !s.getName().isEmpty()){
    t.setName(s.getName());
  }
  if(s.getAge() != null){
    t.setAge(s.getAge());
  }
  if(s.getChildren() != null && !s.getName().isEmpty()){
    t.getChildren().clear()
    ...
  }
  return t;
}
```

This is very helpful.

Currently we use [jodd BeanCopy](http://jodd.org/doc/beanutil.html#beancopy), extend to support ignore empty collection to avoid copy empty collection(due to this https://github.com/mapstruct/mapstruct/issues/865 ,we add empty collection to all collection field).

agudian:
If you have your bean classes under your own domain, you can add `hasX() : boolean` methods for each string or collection property `x` that should not be mapped if not empty. Together with `Mapper#nullValueCheckStrategy`, this could get you pretty far.

Could you check what would still be missing for your use case?

wenerme:
We also have a merge operation in mapper, which not ignore null for PUT method in REST.By now,we have about 100 DTP, so have to add about 1000 hasX method by hand(getter and setter is generated), we can't do this.

filiphr:
I have a similar use case for this issues. The only difference is that I need to not set `null` if the source property is `null`.

From my point of view there are two sub-issues here:

1. Do not set `null` in the update methods. More precisely from our `UpdateWrapper.ftl`:

```
  <#if includeSourceNullCheck>
    if ( <#if sourcePresenceCheckerReference?? >${sourcePresenceCheckerReference}<#else>${sourceReference} != null</#if> ) {
      <@assignToExistingTarget/>
      <@lib.handleAssignment/>;
    }
    else {
      ${ext.targetBeanName}.${ext.targetWriteAccessorName}<@lib.handleWrite>null</@lib.handleWrite>;
    }
  <#else>
    <@assignToExistingTarget/>
    <@lib.handleAssignment/>;
  </#if>
```

This will lead to a method generated like:

```java
@BeanMapping(ignoreNull=true) //How we decide to ignoreNull in update mappings is up for discussion
[...]

PR #1613, opened 2018-09-24
#1571 apply nullvaluecheck strategy on all relevant levels, cleanup

sjaakd:
Some remarks:
The original request was to include it on bean level. However, I see a definite use for it on `@Mapping` level as well. I'm also looking at some more issues. 

One which bothers me as well is #1306. These are merge mappings and especially here its important that null checks work. However, we break some existing behaviour as well if we fixed this. By having fine-grained control, we at least give the user an option to fully specify where to apply this option.

When implementing this I realised that our builders are sometimes very verbose (giving each individual property from `BeanMapping` to the `BeanMappingMethod`. This is a pattern which is repeated. Look at the `ContainerMappingMethod`, `PropertyMappingMethod`. I think we should hand the 'Annotation abstraction, e.g. `BeanMapping` or `Mapping` directly to the `BeanMappingMethod` respectively `PropertyMappingMethod`. Lets gradually change this

### DIFF
--- a/core-common/src/main/java/org/mapstruct/BeanMapping.java
+++ b/core-common/src/main/java/org/mapstruct/BeanMapping.java
@@ -11,6 +11,8 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import static org.mapstruct.NullValueCheckStrategy.ON_IMPLICIT_CONVERSION;
+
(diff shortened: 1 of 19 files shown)


sjaakd:
> Have you had a look at #879 it is similar to #1306. How will this work for update methods?

I need to dig in to it a bit more.. But first this one.. Thanks for the fast review. :smile:

PR #1616, opened 2018-09-25
Adding an issue template to streamline answers.

sjaakd:
I rather had a dialogue before the user writes a new issue, but this is the best next...

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,9 @@
+- [ ] Is this an issue (and hence not a question)? 
+
+If this is a question about a how to use MapStruct there are several resources available.
+- Our reference- and api [documentation](http://mapstruct.org/documentation/reference-guide/).
+- Our [examples](https://github.com/mapstruct/mapstruct-examples) repository (contributions always welcome)
+- Our [FAQ](http://mapstruct.org/faq/)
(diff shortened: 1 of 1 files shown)


chris922:
Good idea! As far as I know templates like this could be stored in a `.github` subfolder. Would this be better as this is more or less a "configuration" just for GitHub?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+- [ ] Is this an issue (and hence not a question)? 
+
+If this is a question about a how to use MapStruct there are several resources available.
```
chris922: without "about a" ? -> "If this is a question how to use MapStruct ..."?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
...
+- Our reference- and api [documentation](http://mapstruct.org/documentation/reference-guide/).
```
chris922: I personally always like to write API in uppercase =)

sjaakd:
> Good idea! As far as I know templates like this could be stored in a `.github` subfolder. Would this be better as this is more or less a "configuration" just for GitHub?

Thanks! I saw other templates as well in the root (`CONTRIBUTING.MD`) so I decided to leave it there.

PR #1618, opened 2018-09-27
#1306 Add new NullValuePropertyMappingStrategy which leaves @MappingTarget untouched

sjaakd:
I was doubting here.. Whether the `ALWAYS` option should leave the mapping target as is for all cases. 

Before the PR: 
1. For nested targets, when source property is null ,the `@MappingTarget` wil also be set to null (follows the source property). 
2. For collections, when source property is null,  the `@MappingTarget` will leave the target untouched  (looking at the code, line 51 at `ExistingInstanceSetterWrapperForCollectionsAndMaps.java` ).

However, we break a lot of testcases (downward compatibility) if I change the behaviour. So I added a new option 'NullValueCheckStrategy.ALWAYS_BUT_LEAVE_TARGET_AS_IS` that does so.

What do you guys think?


documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -2122,9 +2122,11 @@ MapStruct offers control over when to generate a `null` check. By default (`null
 
 First calling a mapping method on the source property is not protected by a null check. Therefor generated mapping methods will do a null check prior to carrying out mapping on a source property. Handwritten mapping methods must take care of null value checking. They have the possibility to add 'meaning' to `null`. For instance: mapping `null` to a default value.
 
-The option `nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS` will always include a null check when source is non primitive, unless a source presence checker is defined on the source bean.
+The option `nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS` will always include a null check when source is non primitive, unless a source presence checker is defined on the source bean. For an existing target (`@MappingTarget` annotated), the value will folllow the source object and set the target to null
```
chris922: typo: `follow`
enclose `null` in \`
end the sentence with a `.`

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -2122,9 +2122,11 @@ MapStruct offers control over when to generate a `null` check. By default (`null
...
 
-The strategy works in a hierarchical fashion. `@Mapper#nullValueCheckStrategy` will override `@MappingConfig#nullValueCheckStrategy`.
+The option `nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS_BUT_LEAVE_TARGET_AS_IS` does the same as `NullValueCheckStrategy.ALWAYS` but leaves an existing target untouched.
```
chris922: "but leaves an existing target untouched when the source is \`null\`" maybe?

core-common/src/main/java/org/mapstruct/NullValueCheckStrategy.java
```diff
@@ -27,12 +27,24 @@
      * <b>NOTE:</b> mapping methods (generated or hand written) are excluded from this null check. They are intended to
      * handle a null source value as 'valid' input.
      *
+     * <b>NOTE:</b> a pre-existing {@link MappingTarget} annotated (nested) target property will end up null when te
+     * source property is null.
```
chris922: `{@code null}` ? (also for the next change)
typo: `te` -> `the`

sjaakd:
> I don't know if I like the name `ALWAYS_BUT_LEAVE_TARGET_AS_IS`..
> But I also don't have a better idea that I like... I was thinking about things like `ALWAYS_WITHOUT_SETTING_TARGET_TO_NULL` because "leave target as is" sounds like the target will never be changed, but this is also very long.

I find this one tricky as well. Thought a couple of days on that, but don't have a better alternative either..

chris922:
Ok, so @filiphr... it's up to you! Give us a good name! 😆

sjaakd:
one of the things that we could do is this:
* Rename `ALWAYS` to `ALWAYS_NULL_WHEN_SOURCE_IS_NULL` or even `ALWAYS_DEPRECATED`
* Rename `ALWAYS_BUT_LEAVE_TARGET_AS_IS` to `ALWAYS`
* write a good migration note.

processor/src/test/java/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNvmsDefaultMapper.java
```diff
@@ -11,13 +11,17 @@
...
+import org.mapstruct.NullValuePropertyMappingStrategy;
 import org.mapstruct.factory.Mappers;
 
 /**
  *
  * @author Sjaak Derksen
  */
-@Mapper( nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT, uses = Helper.class )
+@Mapper(
+    nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT, // This is wrong level.. should be on beanlevel
```
sjaakd: this one worries me. Apparently the `NullValueMappingStrategy.RETURN_DEFAULT` has been (misused) to handle property mappings. To solve this I propose to have a `NullValuePropertyMappingStrategy.SET_TO_DEFAULT`. Note: this one also applies to non-update (regular) mappings to make it consistent.

sjaakd:
@filiphr : some food for thought. See my comments on L21 on `DomainDtoWithNvmsDefaultMapper`. 

I think I am reconsidering my first example: the `NullValuePropertyMappingStrategy` proposal with `SET_TO_DEFAULT`. 

The problem is that `NullValueMappingStrategy` is a bit vague. It applies to the arguments of the `BeanMapping`, `MapMapping` and `IterableMapping`. Hence its a mistake to apply it to properties as has happened above in the `DomainDtoWithNvmsDefaultMapper`. 

------

However: I think that the distinction between  in-line iterable mapping and forged iterable mapping is not that clear for a user. Note (for all clarity): a method is forged when the elements are not the same. 

To the in-line method the `NullValuePropertyMappingStrategy` applies
To the forged method, the  `NullValueMappingStrategy` applies

It is explainable but tricky. Perhaps we should consider (at one point in time) to drop the in-line collection mapping and always apply forged methods..

processor/src/test/java/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNvmsDefaultMapper.java
filiphr: Yes you are right, it has been misused and it is actually my mistake 😄(I merged that PR). Using `NullValuePropertyMappingStrategy.SET_TO_DEFAULT` would make more sense here. The fact that it applies to regular mappings for collections is a mistake in my opinion. We should not be setting it then. The users should provide defaults in the `@ObjectFactory` or object constructor instead.

processor/src/test/java/org/mapstruct/ap/test/nestedtargetproperties/ChartEntryToArtist.java
```diff
@@ -13,14 +13,15 @@
...
+import org.mapstruct.NullValuePropertyMappingStrategy;
 import org.mapstruct.ap.test.nestedsourceproperties._target.ChartEntry;
 import org.mapstruct.ap.test.nestedsourceproperties.source.Chart;
 import org.mapstruct.factory.Mappers;
 
 /**
  * @author Sjaak Derksen
  */
-@Mapper
+@Mapper( nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_NULL)
```
sjaakd: This is strange as well. I expected only update methods were adapted.. Need to checkout why I need a setting here..

sjaakd: Explanation: update methods are generated to cope with the nested mappings. One of such methods is `chartEntryToSong2`. The setting is needed to get exactly the same behaviour as before. So.. Perhaps `NullValuePropertyMappingStrategy.SET_TO_NULL` would make a better default. Not because of logical reasoning, but just to remain backward compatible.

sjaakd:
@filiphr  I think we're quite far with the API on this one.. But few things remain to be resolved. Perhaps you can still comment on these:

1. Have `NullValuePropertyMappingStrategy` only update mappings
> Agreed. I will update the descriptions accordingly. Perhaps we should also update the name of the strategy. With `@ObjectFactory` you will indeed be able to set the default for regular (non update) mapping methods.  But that does not solve what I describe [here](https://github.com/mapstruct/mapstruct/pull/1618#issuecomment-429656897). So we would stil would require a `NullValuePropertyMapping.SET_TO_DEFAULT` to be able to handle the `DomainDtoWithNvmsDefaultMapper`

2. Default value for `NullValuePropertyMappingStrategy`.
> Now it is `NullValuePropertyMappingStrategy.IGNORE` but given the impact (see the updates to the unit test in https://github.com/mapstruct/mapstruct/pull/1618/commits/b95f3bfe929b61a4b67b607fc49d4834b6487106) and especially given the way that nested mappings seem to rely on generating update methods I'm not sure whether the "logical" choice is to prefer over the "backward compatible" choice. WYDT? Are you able to provide a "good / understandable" migration note?

3. Currently I use special values (`MappingConstants.IGNORE` and `MappingConstants.SET_TO_DEFAULT`) for `defaultValues`. 
>However this feels like an odd hierarchical "override" behaviour (`MappingConfig#nullValuePropertyMapping`, `Mapper#nullValuePropertyMapping`, `BeanMapping#nullValuePropertyMapping` and then suddenly magic values in `Mapping#defaultValue`). Perhaps its better to just define a new property `nullValuePropertyMapping` on `@Mapping` as well, get rid of the special `MappingContstants` and check whether its mutual exclusive to `defaultValue`. This feels like a more consistent API. 

--------
Once resolved and agreed, I'll add some more testcases and then I think we're ready.

processor/src/test/java/org/mapstruct/ap/test/nestedtargetproperties/ChartEntryToArtist.java
filiphr: In this case a force set to `NullValuePropertyMappingStrategy.SET_TO_NULL` will actually be the most correct solution and would work nicely for the generated method

processor/src/test/java/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNvmsDefaultMapper.java
filiphr: I don't think we need to add anything for that. Honestly, I consider that behaviour as a bug and I am forced to do those mappings within `@AfterMapping`

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -2109,6 +2111,16 @@ However, by specifying `nullValueMappingStrategy = NullValueMappingStrategy.RETU
 The strategy works in a hierarchical fashion. Setting `nullValueMappingStrategy` on mapping method level will override `@Mapper#nullValueMappingStrategy`, and `@Mapper#nullValueMappingStrategy` will override `@MappingConfig#nullValueMappingStrategy`.
 
 
+[[mapping-result-for-null-properties]]
```
filiphr: I don't think that this is only for `null`, it should also be valid for presence check methods

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -2124,7 +2136,7 @@ First calling a mapping method on the source property is not protected by a null
 
 The option `nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS` will always include a null check when source is non primitive, unless a source presence checker is defined on the source bean.
 
-The strategy works in a hierarchical fashion. `@Mapper#nullValueCheckStrategy` will override `@MappingConfig#nullValueCheckStrategy`.
+The strategy works in a hierarchical fashion. `@Mapping#nullValueCheckStrategy` will override `@BeanMapping#nullValueCheckStrategy`, `@BeanMapping#nullValueCheckStrategy` will override `@Mapper#nullValueCheckStrategy` and `@Mapper#nullValueCheckStrategy` will override `@MappingConfig#nullValueCheckStrategy`.
```
filiphr: This is no longer the case right?

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/ExistingInstanceSetterWrapperForCollectionsAndMaps.java
```diff
@@ -47,7 +50,7 @@ public ExistingInstanceSetterWrapperForCollectionsAndMaps(Assignment decoratedAs
             fieldAssignment,
             mapNullToDefault
         );
-        this.includeSourceNullCheck = ALWAYS == nvms;
+        this.includeSourceNullCheck = ALWAYS == nvms || IGNORE == nvpms;
```
filiphr: Shouldn't the source null check be independent of the `NullValuePropertyMappingStrategy`. We should not do a mistake again and have usages of some properties affecting others

processor/src/test/java/org/mapstruct/ap/test/nestedtargetproperties/ChartEntryToArtist.java
sjaakd: That means that I need to adapt the testcase..

filiphr:
btw I forgot to mention I am pleasantly surprised how "easy" it looks to implement this 😄. I thought it would be a bit more invasive before looking at the code

processor/src/test/java/org/mapstruct/ap/test/nestedtargetproperties/ChartEntryToArtist.java
filiphr: I think that this is acceptable

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: ?  this is about `nullValueCheckStrategy`.. not about `nullValueMappingStrategy` :smile:.

sjaakd:
> I think we should be consistent and just not use `NullValueMappingStrategy` for the forget. One can configure the `NullValuePropertyMappingStrategy` on a mapper level, so this is good enough. In worst case one needs to define an explicit method and configure it.
> 
> >
Oof.. this is tricky.
-  i would say that we have to apply `NullValueMappingStrategy` for 'normal' forged-update-methods (just the same as we would put it into an `@IterableMapping`. But
- we have to distinguish this particular case then, where a nested mapping leads to forging an update method. Then and only then we should apply the `Mapper#nullValuePropertyMapping`.. I guess that's too complicated / difficult to explain.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/ExistingInstanceSetterWrapperForCollectionsAndMaps.java
sjaakd: I don't think the naming of this is very appropriate. The property takes care of including the else branch. I'll check it out.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
filiphr: So this is how it is currently, just the documentation was not correct then?

sjaakd: yep. I forgot to update it after my last PR I think..

sjaakd:
> btw I forgot to mention I am pleasantly surprised how "easy" it looks to implement this 😄. I thought it would be a bit more invasive before looking at the code

It's pretty much ok.. However, I get lost myself now and then in all these normalised templates..

sjaakd:
Still working on this.. Need to add some more testcases and `SET_TO_DEFAULT` option

processor/src/test/java/org/mapstruct/ap/test/nullvaluepropertymapping/NullValuePropertyMappingTest.java
```diff
@@ -0,0 +1,143 @@
...
+    CustomerDTO.class,
+    AddressDTO.class,
+    HomeDTO.class,
+    UserDTO.class
+})
+public class NullValuePropertyMappingTest {
+
+    @Test
+    @WithClasses(CustomerMapper.class)
+    @IssueKey("1306")
```
filiphr: Isn't it the entire class for 1306?

processor/src/test/java/org/mapstruct/ap/test/nullvaluepropertymapping/NullValuePropertyMappingTest.java
```diff
@@ -0,0 +1,143 @@
...
+    )
+    public void testBothDefaultValueANdNullValuePropertyMappingStrategyDefined() {
+    }
+
+    @FunctionalInterface
+    interface CustomerFI {
+        void apply(Customer source, CustomerDTO target);
+    }
+
+    private void testConfig(CustomerFI customerMapper) {
```
filiphr: just FYI you can use `java.util.BiConsumer` as well

processor/src/test/java/org/mapstruct/ap/test/nullvaluepropertymapping/NullValuePropertyMappingTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+        assertThat( userDTO.getHomeDTO() ).isNotNull();
+        assertThat( userDTO.getHomeDTO().getAddressDTO() ).isNotNull();
+        assertThat( userDTO.getHomeDTO().getAddressDTO().getHouseNo() ).isNull();
+        assertThat( userDTO.getDetails() ).isNotNull();
+        assertThat( userDTO.getDetails() ).isEmpty();
+    }
+
+    @Test
+    @Ignore // test gives different results for JDK and JDT
```
filiphr: Still need to investigate this

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
```diff
@@ -155,7 +157,14 @@ else if ( mappingPrism.values.defaultValue() != null && mappingPrism.values.defa
                     Message.PROPERTYMAPPING_DEFAULT_VALUE_AND_DEFAULT_EXPRESSION_BOTH_DEFINED );
             return null;
         }
-
+        else if ( mappingPrism.values.nullValuePropertyMappingStrategy() != null
+            && mappingPrism.values.defaultValue() != null ) {
```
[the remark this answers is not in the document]
sjaakd: good point

processor/src/test/java/org/mapstruct/ap/test/nullvaluepropertymapping/NullValuePropertyMappingTest.java
sjaakd: I got the reproducer form #1306. But in essence the answer is a bit larger than the original issue covered. So I think it deserved a package of its own. I can stick it to the class.

processor/src/test/java/org/mapstruct/ap/test/nullvaluepropertymapping/NullValuePropertyMappingTest.java
sjaakd: ok.. we take it a s a separate issue.

sjaakd:
> I also notices that you are using the abbreviations in the fields (`nvcs`, `nvms`, `nvpms`)

I thought the full names were a bit too long (don't like multiline assignments). But in the end its a matter of taste I think.. What's your pref? I can adapt if you want?

filiphr:
>I thought the full names were a bit too long (don't like multiline assignments). But in the end its a matter of taste I think.. What's your pref? I can adapt if you want?

Yes they are really long 😄. No problem, was just curious. Let's leave it like it is

processor/src/test/java/org/mapstruct/ap/test/nullvaluepropertymapping/NullValuePropertyMappingTest.java
sjaakd: JDK reports on method, Eclipse (correctly) on `@Mapping`..

filiphr: Can you create an issue (if you haven't done already). It would be interesting to see what happens in plain javac with this

sjaakd:
@filiphr .. thanks for reviewing

sahareh:
thanks! when this feature will available in maven central?

PR #1620, opened 2018-10-11
#1608 Make sure that property names for fluent setters starting with is are handled properly

### LINKED ISSUE #1608: Lombok Builders and fields that begin with 'is'
I am having an issue with MapStruct 1.3.0-Beta1 when used with Lombok Builders and fields that begin with **'is'**. 

I get an error like
`
BookMapper.java:[12,10] Unknown property "isbn" in result type api.Book. Did you mean "null"?
`

I recreated the issue in a smaller example. Here are the 2 bean classes that I am mapping between.
```java
@Value
@Builder
public class BookDto {
    String bookIsbn;
    String title;
}
```

```java
@Value
@Builder
public class Book {
    String isbn;
    String title;
}
```

The Lombok generated code, has `getIsbn()`, but the Builder uses fluent style setters `isbn("123")`

```java
@Mapper
public interface BookMapper {
    @Mapping(source = "bookIsbn", target = "isbn")
    Book dtoToBook(BookDto bookDto);
}
```

It does work if I change the mapping line to remove the leading **is** from the target.
```java
@Mapping(source = "bookIsbn", target = "bn")
```

In real code, I also had this mapping, which seems to imply a very different meaning.
issueDate → sueDate

filiphr:
Fixes #1608

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
+++ b/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
@@ -156,8 +156,8 @@ public class DefaultAccessorNamingStrategy implements AccessorNamingStrategy {
     @Override
     public String getPropertyName(ExecutableElement getterOrSetterMethod) {
         String methodName = getterOrSetterMethod.getSimpleName().toString();
-        if ( methodName.startsWith( "is" ) || methodName.startsWith( "get" ) || methodName.startsWith( "set" ) ) {
-            return IntrospectorUtils.decapitalize( methodName.substring( methodName.startsWith( "is" ) ? 2 : 3 ) );
+        if ( methodName.startsWith( "get" ) || methodName.startsWith( "set" ) ) {
+            return IntrospectorUtils.decapitalize( methodName.substring( 3 ) );
         }
         else if ( isFluentSetter( getterOrSetterMethod ) ) {
             return methodName;
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_1608/Book.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_1608/Book.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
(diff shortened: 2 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
```diff
@@ -156,8 +156,8 @@ public boolean isPresenceCheckMethod(ExecutableElement method) {
     @Override
     public String getPropertyName(ExecutableElement getterOrSetterMethod) {
         String methodName = getterOrSetterMethod.getSimpleName().toString();
-        if ( methodName.startsWith( "is" ) || methodName.startsWith( "get" ) || methodName.startsWith( "set" ) ) {
-            return IntrospectorUtils.decapitalize( methodName.substring( methodName.startsWith( "is" ) ? 2 : 3 ) );
+        if ( methodName.startsWith( "get" ) || methodName.startsWith( "set" ) ) {
```
sjaakd: perhaps you told me before, but what if we have a boolean target accessor? e.g. `boolean isAlive()`? Will MapStruct recognise `alive` as property name?

processor/src/test/java/org/mapstruct/ap/test/bugs/_1608/Book.java
```diff
@@ -0,0 +1,32 @@
...
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._1608;
+
+/**
+ * @author Filip Hrisafov
+ */
+public class Book {
+
+    private String isbn;
```
sjaakd: should you add a boolean target accessor?

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
filiphr: When we have accessor it would work. As it is not a setter. However, if you use `isAlive` for your setter then the name of the property would be `isAlive`, which I think is fine, you won't use such name for a setter.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1608/Book.java
filiphr: Not sure if it is really needed, we have other tests for that

Issue #1309, opened 2017-10-19
should be possible to use gradle default annotation processor


mozinrat:
With gradle apt support inbuild, should be possible to get rid of plugin  'net.ltgt.apt' version '0.8'
and use gradle's default apt for annotation processing.

filiphr:
Is this something new? Can you send us some links that we can have a look at?

Searching gives me the `net.ltgt.apt` plugin. And you can use the latest version of it 😀

mozinrat:
With Gradle 4, it's recommended that a plugin use compiler classpath (annotation processors, in general, do this) https://blog.gradle.org/incremental-compiler-avoidance, there is a way to clear specification. I didn't use the source code separation as mentioned, still explicitly stating the apt is still the good way.  Here is my working build.gradle without using apt plugin

apply plugin: 'java'
configurations {
    apt
}
dependencies {
    apt "org.projectlombok:lombok:${lombokVersion}"
    compileOnly "org.projectlombok:lombok:${lombokVersion}"
    apt "org.mapstruct:mapstruct-processor:${mapstructVersion}"
    compileOnly "org.mapstruct:mapstruct-jdk8:${mapstructVersion}"
}
tasks.withType(JavaCompile) {
    options.compilerArgs = ['-Amapstruct.suppressGeneratorTimestamp=true', '-Amapstruct.defaultComponentModel=spring']
}

It works fine with gradle bootrun or delegate to ide in IntelliJ. But to make it work in ide I have to add generated java to src. By default, it will be generated in the classes directory. 

As I mentioned in above forum, you can generate an src path, and this will support incremental build also. But yeah, using above configuration you no longer need the apt plugin.

Until I am using Lombok it's not so advantageous as separated annotation source will not work with lombok, but projects without lombok can go further and create configuration aptCompile for generated java and add them to sourcePath. This way they have full incremental build supported.
I really love that you have created the separate process and lib modules.
Cheers!!

kevcodez:
@mozinrat Do you want the current README section regarding gradle integration to be changed?

filiphr:
@mozinrat I think what you are saying is great. Improving the APT support and so on. However, I think that the apt-plugin already supports the compiler avoidance and works OK with gradle in that regard. Your snippet is actually what the plugin does (I think). Have a look at https://github.com/tbroyer/gradle-apt-plugin/issues/45#issuecomment-301676326. I would more go in the direction of improving the plugin instead of doing things by hand.

[...]

PR #1621, opened 2018-10-11
#1551 Use javax.annotation.processing.Generated if it is present

### LINKED ISSUE #1551: Use java.annotation.processing.Generated if available
See https://github.com/mapstruct/mapstruct/issues/1382#issuecomment-382122018.

Small note. If there is no `java.annotation.Generated` or `java.annotation.processing.Generated` then MapStruct generates something like:

```
/*
@Generated(
    date = "2018-07-14T18:14:00+0200",
    comments = "version: 1.3.0-SNAPSHOT, compiler: javac, environment: Java 10 (Oracle Corporation)"
)
*/
```

filiphr:
Fixes #1551 

There are no tests because we still don't have tests running on Java 9.  Once #1309 is implemented we should add tests for this.

I tested this with the examples and it works properly.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
@@ -74,12 +74,17 @@ public abstract class GeneratedType extends ModelElement {
         this.versionInformation = versionInformation;
         this.accessibility = accessibility;
 
-        this.generatedTypeAvailable = typeFactory.isTypeAvailable( "javax.annotation.Generated" );
-        if ( generatedTypeAvailable ) {
+        if ( typeFactory.isTypeAvailable( "javax.annotation.Generated" ) ) {
             this.generatedType = typeFactory.getType( "javax.annotation.Generated" );
+            this.generatedTypeAvailable = true;
+        }
+        else if (typeFactory.isTypeAvailable( "javax.annotation.processing.Generated" )) {
+            this.generatedType = typeFactory.getType( "javax.annotation.processing.Generated" );
+            this.generatedTypeAvailable = true;
         }
         else {
             this.generatedType = null;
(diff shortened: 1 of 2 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
```diff
@@ -74,12 +74,17 @@ protected GeneratedType(TypeFactory typeFactory, String packageName, String name
         this.versionInformation = versionInformation;
         this.accessibility = accessibility;
 
-        this.generatedTypeAvailable = typeFactory.isTypeAvailable( "javax.annotation.Generated" );
-        if ( generatedTypeAvailable ) {
+        if ( typeFactory.isTypeAvailable( "javax.annotation.Generated" ) ) {
```
sjaakd: Is this an 'or' situation? So when `java.annotation.Generated` is present than that should be used regardless of the presence of `java.annotation.processing.Generated`? Otherwise, the `if`.. `else if` should be flipped around.

filiphr: I had a better look at this. If we switch this then when you compile with Java 10 with Java 8 source then both types are present. Therefore I would propose to do something else. If `SourceVersion` is bigger than Java 8 we can use the `javax.annotation.processing.Generated`. Otherwise we use the `javax.annotation.Generated`.

sjaakd: Ok.. that feels more logical.. Please go ahead when done with merge

filiphr: I will do a new PR as that would involve a bit more changes 😉

filiphr:
Closing PR in order to resolve it by using `SourceVersion`

Issue #1301, opened 2017-09-23
Drop support for Java 6 and Java 7


agudian:
Up until MapStruct 1.2.x, we fully support using MapStruct with JDK 6, JDK 7 and JDK 8 (at compile time), and have it working on JDK 9 out of the box (with maybe an `-addmods` parameter required).

We talked about ditching JDK 6 and 7 for a long time and I think with the official release of JDK 9, it's a good time to do so now.

Here's what we can do:

- [x] Decide on a good time frame for the release and the release version to be used for that change.
- [x] Remove `mapstruct-jdk8` artifact and only use `mapstruct` going forward (using multi-release jar format in case we need different annotations/classes for future Java versions). We can also simplify our build structure again and remove the internal core-common Maven module. And while we're at it, we could rename the directory `core` to `mapstruct` to make it match the Maven artifact name.
- [x] Remove the ITs flavors that target Java 6 or Java 7, maybe also remove the `maven-processor-plugin` flavors (it's only delegating to the default compiler anyway).
- [ ] generate error when 6 or java 7 is used
- [x] adapt our test cases and make effective use of `@Repeatable` (perhaps also the documentation). Maybe we could do a regex replace.. (see https://github.com/mapstruct/mapstruct/issues/1628)
- [ ] look very careful which SpecificCompilerWorkarounds are still needed. Its mostly Eclipse tycho (but probably related to old java versions).
- [x] I think we go through lengths so java 8 types are not required during annotation processor run if java8 is not on path (e.g. in conversions and built-in mappings). We can remove that.
- [x] use diamond operator were applicable
- [ ] renaming directory `core` to `mapstruct`
😁

filiphr:
I would add that we change the way we do our integration tests. Instead of using toolchains let's use  the Travis [Matrix Expansion](https://docs.travis-ci.com/user/build-stages/matrix-expansion/). This would reduce the complexity of our integration tests as well. We can even do the entire build once with Java 8 and once with Java 9.

I am already using it for the IntelliJ plugin https://travis-ci.org/mapstruct/mapstruct-idea and it is pretty easy.

sjaakd:
> Decide on a good time frame for the release and the release version to be used for that change.

[...]

PR #1623, opened 2018-10-12
#1301 Drop support for Java 6 and Java 7

neoXfire:
#1301 
A first draft for a full java 8 build with core-jdk8 dropped.

> https://github.com/mapstruct/mapstruct/issues/1301#issuecomment-429433702


filiphr:
Thanks for the initial draft @neoXfire. What I would suggest is the following:

* Move everything from `core-common` (`mapstruct-common`) into `core`
* Remove the `core-common` (we don't publish this artifact). 
* Keep the pom for `mapstruct-jdk8` add add [maven relocation](https://maven.apache.org/guides/mini/guide-relocation.html) to it to `mapstruct`. This way users will get a warning in their maven build, and will still be able to use `mapstruct-jdk8` we would remove the `mapstruct-jdk8` module in 2.0.

Is that OK for you?

neoXfire:
@filiphr Yes, it sounds good.
I have commit some changes for the last bullet point as @gunnarmorling [mentionned](https://github.com/mapstruct/mapstruct/issues/1301#issuecomment-429433702) in the issue.
I would tackle the two other points that seems easier to me probably tommorow.
Cheers,

filiphr:
Awesome, thanks a lot @neoXfire. 

One other thing that I noticed is that you changed the 1.2.0.Final into 1.3.0.Final in the Readme. For people seeing the readme it would be a bit confusing when they can't find 1.3.0.Final since it is not released. Maybe we need to do that change once 1.3.0.Final is out or a 1.3.0.Beta2 is out. Otherwise people will get confused.

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
```diff
@@ -5,16 +5,17 @@
...
-import java.io.StringWriter;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.ServiceLoader;
-import java.util.Set;
+import org.mapstruct.ap.internal.model.Mapper;
```
sjaakd: Something strange is happening with the imports here..

neoXfire: Yes sorry.  I think I have inadvertently reorganized imports for that class. I will undo it.
However, I paid attention to import your checkstyle rules in my IntelliJ. The import order layout is not defined inside and it looks like you have been using Eclipse default settings. Maybe it should be a good idea to add a chekstyle rule for that to keep cross IDE usage more consistent.

readme.md
```diff
@@ -64,13 +64,13 @@ For Maven-based projects, add the following to your POM file in order to use Map
 ```xml
 ...
 <properties>
-    <org.mapstruct.version>1.2.0.Final</org.mapstruct.version>
```
neoXfire: > One other thing that I noticed is that you changed the 1.2.0.Final into 1.3.0.Final in the Readme. For people seeing the readme it would be a bit confusing when they can't find 1.3.0.Final since it is not released. Maybe we need to do that change once 1.3.0.Final is out or a 1.3.0.Beta2 is out. Otherwise people will get confused.

@filiphr I initially thought this draft would be merged in a dedicated branch, not in master before release. I'm going to revert the changes of this file totally for 1.2.0.Final and keep the Java 1.6 requirement

filiphr: Your initial assumption makes complete sense. Thanks for the revert

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
filiphr: There are still some classes that don't have the proper formatting completely applied to them. The Eclipse and IntelliJ formatter should be aligned already (unless we have missed something). Adding checkstyle might help, but we would first need to fix all classes 😄

filiphr:
If one other from @sjaakd, @gunnarmorling, @agudian or @chris922 agree I would go on and merge this PR into master

sjaakd:
> If one other from @sjaakd, @gunnarmorling, @agudian or @chris922 agree I would go on and merge this PR into master

I agree. Can we make issues out of the other points I mentioned above?

neoXfire:
@sjaakd I tried to do a refactoring to use repeated annotations in the test classes but it is going to be harder that I thought. We have to change all line numbers in error reports.

filiphr:
@neoXfire let's do the test migration gradually.

Why the deprecation of `@Mappings` and `@ValueMappings`. If people prefer to use them they should. Also `@Repeatable` ia just syntactic sugar

sjaakd:
> Why the deprecation of `@Mappings` and `@ValueMappings`. If people prefer to use them they should. Also `@Repeatable` ia just syntactic sugar

Agreed. Deprecation is not required. 

Although I do feel its a good idea to adapt the unit test since MapStruct in itself is all about minimising (mapping) work and we should showcase that. Just as the documentation & examples. But like @filiphr said: we can do this in a separate PR. (don't throw away your commit yet :smile:).

neoXfire:
Ok

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
neoXfire: I had missed the `/etc` folder with formaters, I thought all style rules were in the checkstyle.xml in `/build-config` folder. 🔎🔎🔎
Thank you for rising the topic of formaters.

neoXfire:
> Why the deprecation of `@Mappings` and `@ValueMappings`. If people prefer to use them they should. Also `@Repeatable` ia just syntactic sugar

I disagree with you on that for these reasons : 
- I personally don't think that a lot of people prefer the `@Mappings` and `@ValueMappings` API. One goal of MapStruct is to avoid writing boilerplate. `@Mappings` and `@ValueMappings` declare absolutely no valuable information about mapping. They are just workarounds around a Java 6 missing feature : repeatable annotations.
- If you adapt the unit tests to use repeatable annotations, `@Mappings` and `@ValueMappings` won't be well covered. Is it a good idea to expose an non-deprecated API with so few test cases ?
- Some developers don't have time to read all changelogs of all librairies they use in their project. They just upgrade regularly and check if their tests still pass and if some warnings appears in the IDE. By deprecating `@Mappings` and `@ValueMappings` you can tell them : "By the way, now there is a shorter syntax available if you missed that"  
- Deprecated don't necessarily mean deleted soon. You can have  `@Mappings` and `@ValueMappings` for people who really like the wrapped syntax.

neoXfire:
Should I rename core directory to mapstruct to match artefact name?

sjaakd:
> Should I rename core directory to mapstruct to match artefact name?

@filiphr : I'm in favour... What about you?

sjaakd:
> I disagree with you on that for these reasons :

I agree with you on the first point. 

But on a more fundamental level: `@Mapping` refers to `@Mappings` by means of its repeatable annotation `@Repeatable(Mappings.class)`. Now you could decide to have an inner annotation definition (I think that Hibernate Validation does such a thing) in stead of an explicit file to hide this fact a bit. But nonetheless, A (similar) thing like `@Mappings` is still needed because the repeatable annotation simple would not work otherwise. That in itself is enough reason not to deprecate it :smile:. 

----
In general, I don't like the term "syntactic sugar" because its subjective (and time bound). A couple of years ago, we did not have Lamda expressions. They are (according to some I know) also syntactic sugar for inner classes. Exaggerating: you could claim that Java itself is syntactic sugar for byte code language of the VM (and so on). IMHO everything that is there to make us write concise efficient (non redundant) code is as far as I am concerned needed to make us productive. So.. Thats enough rant for today :smile: 

Thanks for your work.. You sure put a lot of effort in it and its appreciated! 👍

filiphr:
@neoXfire as Sjaak said thnks a lot for your work here.

For the directoty rename, if you want to rename it then feel free to do so as part of this PR. 

The other parts can go in other tickets. I understand your reasoning about the repeatable, but I still think we should not deprecate them.

neoXfire:
I'm going to stick to the old name as I can't get a maven build working by renaming. The annotations can be resolved in processor tests. 
I'll have a look at your commits if you decide to do the changes.

filiphr:
>I'm going to stick to the old name as I can't get a maven build working by renaming. The annotations can be resolved in processor tests.

No problem.

filiphr:
Just merged this PR. Big thank you @neoXfire for the work on this.

Issue #1591, opened 2018-08-22
Map a java.sql.Date to a java.time.LocalDate


pioto:
When attempting to map a field which, on the source object has type `java.util.Date` but which, when loaded via JPA ends up with a `java.sql.Date` object instance, it seems that MapStruct tries to use the `.toInstant()` method to perform the mapping.

However, this will not work, as `java.sql.Date` always throws an `UnsupportedOperationException`.

The root cause is probably just the poor design of the java.sql.Date class and its friends from an inheritance perspective, but it would be nice if there were some "magic" in MapStruct to handle this conversion, as I expect it isn't that unusual when working with a JPA entity.

A workaround seems to be using an `expression` like:

```java
    @Mapping(
            target = "createdDate",
            expression = "java(((java.sql.Date)(source.getCreated())).toLocalDate())"
    )
```

filiphr:
The problem is that MapStruct is an annotation processor which means that it knows only the compilation types and not the runtime types. If you define your type as `java.sql.Date` instead of `java.util.Date` then it should work properly (it will do double conversion, but we accept a PR for `java.sql.Date` into `java.time.LocalDate` 😄).

Another approach would be instead of using an expression to use a [qualifier](http://mapstruct.org/documentation/dev/reference/html/#selection-based-on-qualifiers).

pioto:
Yeah, I came across this approach instead, using `@Mapper(uses=SqlDateMapper.class)`, and this bit of code:

```java
public class SqlDateMapper {

    public java.util.Date asDate(LocalDate date) {
        return java.util.Date.from(date.atStartOfDay(ZoneId.systemDefault()).toInstant());
    }

    public LocalDate asLocalDate(java.util.Date date) {
        if (date instanceof java.sql.Date) {
            return ((java.sql.Date) date).toLocalDate();
        } else {
            return LocalDate.from(date.toInstant());
        }
    }

}
```

That's probably the "right" fix for my use case. (Your default conversions also use UTC instead of the local zone, which probably won't quite match up with my business rules' expectations)

filiphr:
I've marked this as up-for-grabs in case someone wants to provide a contribution.

PR #1626, opened 2018-10-13
#1591 Fix LocalDateTime conversion from UTC to use the local timezone

sivviswa22:


sivviswa22:
@filiphr Please review.

filiphr:
@sivviswa22 this PR is not linked with the #1591. What exactly are you trying to achieve? Using the UTC zone is by design it is not a bug

sivviswa22:
But should we be not just be mapping the date time to LocalDateTime and vice-versa to exactly the same timezone values instead of formatting/converting it to UTC?

> @sivviswa22 this PR is not linked with the #1591. What exactly are you trying to achieve? Using the UTC zone is by design it is not a bug

But should we be not just be mapping the date time to LocalDateTime and vice-versa to exactly the same timezone values instead of formatting/converting it to UTC?

sivviswa22:
Closing this pull request will create a new one with the correct changes.

PR #1627, opened 2018-10-14
Add support for mapping java.time.LocalDate to java.sql.Date

### LINKED ISSUE #1591: Map a java.sql.Date to a java.time.LocalDate

sivviswa22:
Fixes #1591

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
@@ -5,8 +5,6 @@
  */
 package org.mapstruct.ap.internal.conversion;
 
-import static org.mapstruct.ap.internal.conversion.ReverseConversion.reverse;
-
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.sql.Time;
@@ -16,13 +14,14 @@ import java.util.Currency;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
-
 import javax.lang.model.util.Elements;
 
 import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.common.TypeFactory;
 import org.mapstruct.ap.internal.util.JavaTimeConstants;
 import org.mapstruct.ap.internal.util.JodaTimeConstants;
+
+import static org.mapstruct.ap.internal.conversion.ReverseConversion.reverse;
 
 /**
  * Holds built-in {@link ConversionProvider}s such as from {@code int} to {@code String}.
@@ -168,8 +167,8 @@ public class Conversions {
         registerToStringConversion( Boolean.class );
         register( char.class, String.class, new CharToStringConversion() );
         register( Character.class, String.class, new CharWrapperToStringConversion() );
-        register( BigInteger.class, String.class, new BigIntegerToStringConversion() );
-        register( BigDecimal.class, String.class, new BigDecimalToStringConversion() );
+        register( BigInteger.class, String.class, new BigIntegerToStringConversion( ) );
+        register( BigDecimal.class, String.class, new BigDecimalToStringConversion( ) );
 
         registerJodaConversions();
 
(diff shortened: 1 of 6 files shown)


processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
```diff
@@ -168,8 +168,8 @@ public Conversions(Elements elementUtils, TypeFactory typeFactory) {
         registerToStringConversion( Boolean.class );
         register( char.class, String.class, new CharToStringConversion() );
         register( Character.class, String.class, new CharWrapperToStringConversion() );
-        register( BigInteger.class, String.class, new BigIntegerToStringConversion( ) );
-        register( BigDecimal.class, String.class, new BigDecimalToStringConversion( ) );
+        register( BigInteger.class, String.class, new BigIntegerToStringConversion() );
```
filiphr: This change is not really necessary

processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
```diff
@@ -222,6 +222,7 @@ private void registerJava8TimeConversions() {
         register( JavaTimeConstants.ZONED_DATE_TIME_FQN, Date.class, new JavaZonedDateTimeToDateConversion() );
         register( JavaTimeConstants.LOCAL_DATE_TIME_FQN, Date.class, new JavaLocalDateTimeToDateConversion() );
         register( JavaTimeConstants.LOCAL_DATE_FQN, Date.class, new JavaLocalDateToDateConversion() );
+        register( java.sql.Date.class, JavaTimeConstants.LOCAL_DATE_FQN, new JavaSqlDateToLocalDateConversion() );
```
filiphr: Why not do the reverser registration (first the `LOCAL_DATE_FQN` and then `java.sql.Date`)? Then you won't need the new register method. Also the name of the conversion would need to be `JavaLocalDateToSqlDateConversion`

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaSqlDateToLocalDateConversion.java
```diff
@@ -0,0 +1,56 @@
...
+    protected Set<Type> getToConversionImportTypes(ConversionContext conversionContext) {
+        return Collections.asSet(
+            conversionContext.getTypeFactory().getType( ZONE_OFFSET_FQN )
+        );
+    }
+
+    @Override
+    protected String getFromExpression(ConversionContext conversionContext) {
+        return "new " + sqlDate( conversionContext ) + "(" +
+            date( conversionContext )
```
filiphr: Why not use `Instant#toEpochMilli` instead of creating `java.util.Date`?

sivviswa22:
@filiphr  Please review and merge.  I have taken care of all of the comments provided.Thanks, Siva

processor/src/main/java/org/mapstruct/ap/internal/conversion/JavaLocalDateToSqlDateConversion.java
```diff
@@ -0,0 +1,53 @@
...
+
+import static org.mapstruct.ap.internal.conversion.ConversionUtils.sqlDate;
+import static org.mapstruct.ap.internal.conversion.ConversionUtils.zoneOffset;
+import static org.mapstruct.ap.internal.util.JavaTimeConstants.ZONE_OFFSET_FQN;
+
+/**
+ * SimpleConversion for mapping {@link java.time.LocalDate} to
+ * {@link Date} and vice versa.
+ */
+public class JavaLocalDateToSqlDateConversion extends SimpleConversion {
```
filiphr: You can add your name as `@author` if you want to

sivviswa22: I don't want to add my name as an author yet. This is my first PR contribution to open source. Thanks a lot for helping me out.

I hope to do more so I can add my name as an author at some point later 😀.

filiphr: Thanks for letting us know. It's your choice. Really glad that you decided to contribute to MapStruct as your first contribution.

Would it be OK for you if we add your name to https://github.com/mapstruct/mapstruct/blob/master/copyright.txt?

neoXfire:
Strange, i can't have integration tests passing on my computer with that commit. Can you ?

PR #1634, opened 2018-10-26
#1478 Use source property name for the adder iteration

### LINKED ISSUE #1478: Adder wrapper should use singular from the source for the iterable element
Currently generated code:

```

 if ( ptacObject.getCustomers() != null ) {
    for ( CustomerInfo customersList : ptacObject.getCustomers() ) {
        builder.addCustomers( map( customersList ) );
    }
}

if ( ptacObject.getBrokers() != null ) {
    for ( String brokersList : ptacObject.getBrokers() ) {
        builder.addBrokers( brokersList );
    }
}
```
Example code can be found [here](https://github.com/pascalwhoop/grpc-adapter)

filiphr:
A WIP for using the source property name for the adder iteration.

One thing that I noticed is that we are doing a lot of copying around when we can just pass the `Mapping` to the appropriate builders and then use whatever we need in there.

Fixes #1478

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -105,6 +105,7 @@ public class PropertyMapping extends ModelElement {
         protected Type targetType;
         protected Accessor targetReadAccessor;
         protected String targetPropertyName;
+        protected String sourcePropertyName;
 
         protected List<String> dependsOn;
         protected Set<String> existingVariableNames;
@@ -168,6 +169,11 @@ public class PropertyMapping extends ModelElement {
 
         public T targetPropertyName(String targetPropertyName) {
             this.targetPropertyName = targetPropertyName;
+            return (T) this;
+        }
+
+        public T sourcePropertyName(String sourcePropertyName) {
+            this.sourcePropertyName = sourcePropertyName;
             return (T) this;
         }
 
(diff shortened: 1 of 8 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -468,12 +474,13 @@ private Assignment assignToPlainViaAdder( Assignment rightHandSide) {
 
             Assignment result = rightHandSide;
 
+            String adderWrapperName = sourcePropertyName == null ? targetPropertyName : sourcePropertyName;
```
sjaakd: `adderIteratorName` ?

processor/src/test/java/org/mapstruct/ap/test/collection/adder/AdderTest.java
```diff
@@ -255,4 +260,17 @@ public void testMissingImport() throws DogException {
         assertThat( target ).isNotNull();
         assertThat( target.getAttributes().size() ).isEqualTo( 1 );
     }
+
+    @IssueKey("1478")
```
sjaakd: I wish I never created this zoo of animals :smile:

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
filiphr: Thank you very much. I had troubles coming up with a nice name, your suggestion is perfect

filiphr:
Thanks for the review @sjaakd

Issue #1633, opened 2018-10-25
Fix wrong line number in JDK for @RepeatableMappings


sjaakd:
checkout:  processor/src/test/java/org/mapstruct/ap/test/nullvaluepropertymapping/NullValuePropertyMappingTest.java:

filiphr:
@sjaakd I am closing this and we can track the `@Repeatable` problem with javac in #1628

PR #1635, opened 2018-10-27
#1633 Add support for ignoring tests on a certain compiler

filiphr:
* This should be used as a last resort when a bug in a compiler cannot be fixed by us
* The NullValuePropertyMappingTest uses repeatable annotations that reports errors on wrong lines in javac JDK-8042710

Related to #1633.

This is a temporary solution so that we can at least have some tests (Eclipse) running properly


filiphr:
That might actually be better. Maybe we even allow different lines for different compilers so we at least notice when the bug gets fixed.

filiphr:
@sjaakd I added a new approach. However, I couldn't really do it per compiler as we would have had cyclic dependencies in the testing between our `@Diagnostic` and `Compiler`.

Not sure whether I like this approach though. WDYT?

sjaakd:
> Not sure whether I like this approach though. WDYT?

I think its good enough.. Some remarks: we should stick to not using `@Repeatable` as long as this is not fixed (for new test cases). These are allright. 

Do you know if its fixed in JDK9 (and later)?

Go ahead and merge..

filiphr:
>Do you know if its fixed in JDK9 (and later)?

I tried with JDK 9, 10, 11 it was all the same. The Oracle bug is still open.

>Some remarks: we should stick to not using `@Repeatable` as long as this is not fixed (for new test cases). These are allright.

I agree for sticking to not using `@Repeatable`

PR #1636, opened 2018-10-28
#1301 Update build with latest dependencies and Java 8 baseline

filiphr:
* Do not use maven-processor-plugin and use the maven-compiler-plugin annotationProcessors instead
* Update NoPackageCyclesRule and exclude tests
* Update checkstyle to 8.14 and remove FIleContentsHolder module and move SuppressionCommentFilter to the TreeWalker
* Update assertj to 3.11.1 and replace usage of RuntimeIOException (from AssertJ) with UncheckedIOException (from java 8)
* Add how to import MapStruct into IntelliJ and Eclipse into the readme

Needed for further support for Java 9+

### DIFF
--- a/build-config/src/main/resources/build-config/checkstyle.xml
+++ b/build-config/src/main/resources/build-config/checkstyle.xml
@@ -62,8 +62,6 @@
 
         <property name="cacheFile" value="${checkstyle.cache.file}"/>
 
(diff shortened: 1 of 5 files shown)


filiphr:
Thanks for the review @sjaakd

Issue #1504, opened 2018-05-22
Property error logging on @Mapping annotation


sjaakd:
Currently errors in property mapping are recorded on method level instead of `@Mapping` annotation level. If such annotation is available , its, with the advent of #1503 quite easy to report on this level.

sjaakd:
don't think so.. the annotation mirror is now in rare occasions handed to the `PropertyMapping` class. There used to be no distinction between name-based-property mapping (implicit) and explicit (`@Mapping`). I introduced to make this distinction as part of the constant mapping. The PR #1503 takes that just a bit further in the errors / warnings required for type conversion loss. This makes it a rather small step to solve more general. Hence this issue.

sjaakd:
There are 2 more occurences were we report in the `MapperResolverImpl` were we could benifit from the position hint. Small frie, but still needs to be addressed.

sjaakd:
done

PR #1639, opened 2018-11-02
#1504 adding position hints for ambiguous mapping methods

sjaakd:


sjaakd:
> Is this the last one for #1504?

Yep. should be fine now..

sjaakd:
@filiphr thanks for the review 👍

PR #1641, opened 2018-11-03
#1640 Add init method with MapStructProcessingEnvironment to the BuilderProvider

### LINKED ISSUE #1640: Use MapStructProcessingEnvironment for initialising the BuilderProvider
In https://github.com/mapstruct/mapstruct/commit/6f19d5615500d66193f34244381db1e31b6856d9 we added `MapStructProcessingEnvironment` that can be used to initialize the `AccessorNamingStrategy` and pass some extra utilities to it. We should do the same in the `BuilderProvider`

filiphr:
Used for initializing the Elements and Types

Fixes #1640

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
@@ -522,7 +522,7 @@ public class TypeFactory {
         try {
             return roundContext.getAnnotationProcessorContext()
                 .getBuilderProvider()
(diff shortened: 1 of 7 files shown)


filiphr:
>However there's a possibility that people started to use this.. so I guess it should be in the release notes.. and is there documentation that should be adapted?

Yes I will mention this breaking change in the `BuilderProvider` and the `DefaultBuilderProvider`. I think it is fine for the beta release. As for the documentation there is nothing that needs adapting. We are directly pulling the `NoOpBuilderProvider` so it would automatically be updated

PR #1642, opened 2018-11-03
#1576 Delay determining whether a Type needs to be imported

### LINKED ISSUE #1576: FQN of LocalDateTime is used when a simple name is possible
When using the `JavaLocalDateToDateConversion` we generate a code that looks like:

```
if ( as.getPromotionDate() != null ) {
    article.setPromotionDate( java.time.LocalDateTime.ofInstant( as.getPromotionDate().toInstant(), ZoneOffset.UTC ).toLocalDate() );
}
```

The code should look like:

```
if ( as.getPromotionDate() != null ) {
    article.setPromotionDate( LocalDateTime.ofInstant( as.getPromotionDate().toInstant(), ZoneOffset.UTC ).toLocalDate() );
}
```

The `java.time.LocalDateTime` is not necessary to be FQN, as it is the only `LocalDateTime`. The actual problem is that when Joda time is present than the Joda types are loaded in the `Conversion` which leads to them being part of the `importedQualifiedTypesBySimpleName` in our `TypeFactory`. Seems that we use the `Type` in the Conversion [`Key`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java#L305). However, looking into that code it seems that we can easily use the FQN of the types (without the need to load the `Type`).

I see 2 ways to solve this:

* Add a flag to the `TypeFactory` to load a type without marking it as imported
* Use `String` in the Conversion `Key`

sjaakd:
Fixes #1576


processor/src/main/java/org/mapstruct/ap/internal/model/source/builtin/BuiltInMappingMethods.java
```diff
@@ -39,15 +38,12 @@ public BuiltInMappingMethods(TypeFactory typeFactory) {
...
-        if ( isJava8TimeAvailable( typeFactory ) ) {
-            builtInMethods.add( new ZonedDateTimeToCalendar( typeFactory ) );
-            builtInMethods.add( new ZonedDateTimeToXmlGregorianCalendar( typeFactory ) );
-            builtInMethods.add( new CalendarToZonedDateTime( typeFactory ) );
-            if ( isXmlGregorianCalendarPresent ) {
-                builtInMethods.add( new XmlGregorianCalendarToLocalDate( typeFactory ) );
-                builtInMethods.add( new LocalDateToXmlGregorianCalendar( typeFactory ) );
-            }
+        builtInMethods.add( new ZonedDateTimeToCalendar( typeFactory ) );
+        builtInMethods.add( new ZonedDateTimeToXmlGregorianCalendar( typeFactory ) );
```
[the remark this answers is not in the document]
sjaakd: ok.. I'll fix that. I know that the whole XML stuff is not in the standard module.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1576/java8/Source.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._1576.java8;
+
+import java.util.Date;
+
+public class Source {
```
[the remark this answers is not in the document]
sjaakd: > The problem is indeed that calling `TypeFactory#getType` marks the entries as imported. Maybe we need to distinguish between types that are used in the generation and types that aren't

I thought about that. But i still think its to brittle. In principle all types created by the `getType` can be used later on. We simply don't know at forehand. We need the functionality in type, because we check on `assignment` than on the FQN (need to check this still).

The problem is that determining whether a type is imported belongs to the `GeneratedType` and not as such to `Type`. We need it earlier, during model creation (checkout `Type#getReferenceName`) and we only 'know' which types we already added during writing the model. Its the famous catch22. 

------------

The method name `TypeFactory#getType` is misleading.. Its actually `createType`, which is a bit strange at the same time, because it might actually already have been created by another mapping method (perhaps we should cache already created types in the future, but that's a different optimisation and return a cached one in the `getType` in the future).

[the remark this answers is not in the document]
sjaakd: Thought a little bit more about it.. There are 2 moments when we need to know if a type is imported:
1. When its referenced (during creation: `Type#getReferenceName`)
2. When we generate the imports for the `GeneratedType`.

What if we move the administration whether a type is imported or not to this `Type#getReferenceName`? We could rebrand it as `Type#createReference` returning the same `String` but additionally doing the administration that is now in the `TypeFactory#isImported`..

Than we don't have to be so picky when to call the `TypeFactory#getType`. WDYT?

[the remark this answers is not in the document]
filiphr: Regarding the caching it has come to my mind as well, but I would say it should be done some other time. We should create some benchmarking to see if we gain some speed from this.

Regarding the reference and the import. I agree with you. I think that we can even leave the `getReferenceName` and always delegate to `TypeFactory#isImported` to check if a type is imported and not have a boolean in the `Type`. This would mean that we need to have the `TypeFactory` in the `Type`, but I think this is OK. This would mean that the last commit is obsolete, right?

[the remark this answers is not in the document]
sjaakd: Apart from the test case, yes. I probably only delegate v the hash table

sjaakd:
@chris922 , @filiphr : I hit a problem here. Its called #1460 .. 

@chris922 : you worked on this problem in the past. 

The problem I'm facing is that we cannot determine the order in which `Type`s are created. Whether or not a `Type` needs to be imported in the `GeneratedType` is governed by 2 factors:
1. Whether the simple name has of the `Type` already been used before.
2. Whether the `Type` to be imported is declared in the same package as the generated class or is even declared in the super of the generated class. 

Now: since types are created in an arbitrary way, it could be that a type is already be marked as `toBeImported` (and the model calling element is created) according to factor 1. 

 If a next type is a `Type` from the same package or defined in the super class (e.g. in the Mapper) you actually want to revert that decision and `award` the type from the same package the short name. 

But you can't come back on the choice made.

So: I committed some code.. But I'm not sure how to continue.. Any ideas?

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -169,12 +174,15 @@ public String getName() {
     }
 
     /**
-     * String that could be used in generated code to reference to this {@link Type}.
+     * String that could be used in generated code to reference to this {@link Type}. The first time a name
+     * is used (e.g. LocalDateTime has java.time.LocalDate and the org.joda.LocalDateTime) it will be marked to be
+     * imported. So if the java.time variant is used first, this name will be used and java.time.LocalDateTime
+     * will be imported, and the org.joda variant will use the FQN (and will not be imported).
```
chris922: Maybe the following?

```
Returns a String that could be used in generated code to reference to this {@link Type}.<br>
<p>
The first time a name is used (e. g. for {@code LocalDateTime} there is {@code java.time.LocalDateTime} and {@code org.joda.LocalDateTime}) it will be marked as to be imported. So if the {@code java.time} variant is used first, the {@code java.time.LocalDateTime} will be imported and the {@code org.joda} variant will be used with the FQN (and will not be imported).
```

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -1004,4 +1029,28 @@ public boolean isLiteral() {
         return isLiteral;
     }
 
+    /**
+     * It strips the all the {@code []} from the {@code className}.
```
chris922: `It strips all the ...`

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -1004,4 +1029,28 @@ public boolean isLiteral() {
...
+     * <pre>
+     *     trimSimpleClassName("String[][][]") -> "String"
+     *     trimSimpleClassName("String[]") -> "String"
+     * </pre>
+     *
+     * @param className that needs to be trimmed
+     *
+     * @return the trimmed {@code className}, or {@code null} if the {@code className} was {@code null}
+     */
+    private String trimSimpleClassName(String className) {
```
chris922: Maybe move this to a utility class as it might be useful also somewhere else in the future?

chris922:
Coming back to your issue I am not sure if I understood it correctly, maybe I try to summarize it with an example:

Your change is that as soon as `createReferenceName` is called the first time it will be decided if the `Type` need to be imported or not.

Example:
* Mapper: `example.FooMapper`
* Two classes: `example.Foo` and `mapstruct.Foo`

The `createReferenceName` of type `mapstruct.Foo` will be called first, thus this type will be imported and referenced by it's simple name `Foo`. But additionally the other type `example.Foo` will be used later and as `mapstruct.Foo` is already marked as `toBeImported` it will be referenced by its FQN.
But better would be that `example.Foo` will be referenced with it's simple name (without any import as it is in the same package like the mapper) and `mapstruct.Foo` will be used by its FQN. Correct?

sjaakd:
@filiphr .. Think I solved it.. Still have to add some more testcases as you proposed.. But can you have a look in the mean time?

sjaakd:
oh.. and I'm for some reason not allowed to use streams.. Can you give me a hint?

sjaakd:
@chris922 .. just saw your comments.. The above was of course also intended for you...

sjaakd:
> Example:
> 
> * Mapper: `example.FooMapper`
> * Two classes: `example.Foo` and `mapstruct.Foo`

Almost.. Try for an example

xample:

Mapper: `example.FooMapper`
Two classes: `example.Foo` and `mapstruct.FooMapper.Foo`

As soon as `example.Foo` is marked from import, MapStruct cannot come back on this decision. However, `example.Foo` cannot be imported and referenced as `Foo`. That right is reserved for `mapstruct.FooMapper.Foo` (because the implementation inherits from `FooMapper`

filiphr:
>oh.. and I'm for some reason not allowed to use streams.. Can you give me a hint?

What do you mean you can't use streams? I see that your commit has them.

Regarding the changes, I am not entirely sure about the maps in the `Type`. Why not have all of that in the `TypeFactory`? You can always pass the factory to the `Type` and use it from there. 

It is also weird why doing the same stuff as before in a different order breaks things.

sjaakd:
> It is also weird why doing the same stuff as before in a different order breaks things.

You know what it is with maps.. orders are not guaranteed... But check the difference run-time.. The Map is now far smaller. It only contains the types that are actually used in the model. Not stuff that we try to match or try to build a conversion-set from.. Makes you wonder.. If we need the `getImportedTypes` at all. But that's a next phase..

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
sjaakd: Not sure... lets do that when the need arises..

filiphr:
>But some plugin fails.. Man.. we're far to restrictive. 😄

Remove the `animal-sniffer-plugin` it is not needed not I think.

>You know what it is with maps.. orders are not guaranteed... But check the difference run-time.. The Map is now far smaller. It only contains the types that are actually used in the model.

I can imagine that it is smaller, much less things to store 😄

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -341,14 +352,37 @@ public String getImportName() {
...
+    public boolean isToBeImported() {
+        if ( isToBeImported == null ) {
+            String trimmedName = trimSimpleClassName( name );
+            if ( notToBeImportedTypes.containsKey( trimmedName ) ) {
+                return false;
+            }
+            String trimmedQualifiedName = trimSimpleClassName( qualifiedName );
+            String importedType = toBeImportedTypes.get( trimmedName );
+
+            isToBeImported = false;
```
filiphr: Shouldn't we initialize this value when we return false on line 164 as well?

sjaakd: 164?

filiphr: 364. Sorry a typo

sjaakd:
@filiphr are you ok with this PR now?

Issue #1645, opened 2018-11-11
Cleanup dropping support for Java 6 and 7


filiphr:
With https://github.com/mapstruct/mapstruct/issues/1301 we dropped Java 6 and 7 support. This issue covers some of the topics mentioned in the previous issue.

- [ ] generate error when 6 or java 7 is used
- [ ] look very careful which SpecificCompilerWorkarounds are still needed. Its mostly Eclipse tycho (but probably related to old java versions).
- [ ] renaming directory `core` to `mapstruct`

PR #1646, opened 2018-11-11
#1645 Use Repeatable Mapping annotation in the reference documentation

filiphr:
Cleanups in the reference documentation to use Repeteable `@Mapping` annotation. 
Related to #1645


sjaakd:
LGTM.. One worry though.. The error messages are reported on the method.. Do we need to highlight that somewhere?

filiphr:
>Do we need to highlight that somewhere?

Yes that is annoying as hell. Not sure if we should put that somewhere. Maybe in our FAQ on the website?

PR #1649, opened 2018-11-12
Improvement: builder for Mapper/Decorator/GeneratedType

sjaakd:


processor/src/main/java/org/mapstruct/ap/internal/model/Mapper.java
```diff
@@ -177,6 +112,38 @@ public Mapper build() {
 
     }
 
+    private final boolean customPackage;
+    private final boolean customImplName;
+    private Decorator decorator;
+
+    @SuppressWarnings( "checkstyle:parameternumber" )
+    private Mapper(TypeFactory typeFactory, String packageName, String name, String superClassName,
```
filiphr: Why move the constructor down?

sjaakd:
Convention? I thought according to jls..

Verstuurd vanaf mijn iPhone

filiphr:
>Convention? I thought according to jls..

It's fine, I was just confused, because on other places it is above the builders (I think)

PR #1651, opened 2018-11-17
#1648 Source properties defined in Mapping should not be reported as unmapped

### LINKED ISSUE #1648: unmappedSourcePolicy ERROR may lead to false positives
When mapping properties with different names in source and target, `unmappedSourcePolicy = ReportingPolicy.ERROR` may lead to false positives. For this, see the following example:

```
@Mapping(target = "foo", source = "bar")
protected abstract TargetType fromSourceType(SourceType sourceType)
```

Even though the field `bar` is explicitly mapped, the source policy will produce a warning/error unless `@BeanMapping` is used to ignore the source property, such as

```
@Mapping(target = "foo", source = "bar")
@BeanMapping(ignoreUnmappedSourceProperties = { "bar" })
protected abstract TargetType fromSourceType(SourceType sourceType)
```

In fact this may lead to real errors once the line that contains the actual mapping were removed by accident. What I would expect to see instead is that the analysis of the `unmappedSourcePolicy` takes the actual `@Mapping`s into account.

filiphr:
Fixes #1648


filiphr:
Thanks for the review @sjaakd

PR #1664, opened 2018-12-02
#1656 Support for mapping between String and java.time Duration, Period, Instant

### LINKED ISSUE #1656: Builtin mapping of String <-> Duration/Period/Instant
A fair chunk of the java.time objects have builtin mappings to and from String. Duration and Period also has a standard string format, the docs describe it a bit.

* [Duration#parse](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-)
* [Duration#toString](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#toString--)
* [Period#parse](https://docs.oracle.com/javase/8/docs/api/java/time/Period.html#parse-java.lang.CharSequence-)
* [Period#toString](https://docs.oracle.com/javase/8/docs/api/java/time/Period.html#toString--)
* [Instant](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#toString--) is similar, using `DateTimeFormatter.ISO_INSTANT`

However, MapStruct doesn't support these conversions [implicitly](http://mapstruct.org/documentation/dev/reference/html/#implicit-type-conversions). Would be a nice to have, out of the box, to fit with the rest of java.time mappings.

twentylemon:
Fixes #1656 + some formatting that intellij wanted to do

### DIFF
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -944,6 +944,8 @@ public interface CarMapper {
 
 * Between `java.time.ZonedDateTime`, `java.time.LocalDateTime`, `java.time.LocalDate`, `java.time.LocalTime` from Java 8 Date-Time package and `String`. A format string as understood by `java.text.SimpleDateFormat` can be specified via the `dateFormat` option (see above).
 
+* Between `java.time.Instant`, `java.time.Duration`, `java.time.Period` from Java 8 Date-Time package and `String`, as per the `parse` method in each class.
+
 * Between `java.time.ZonedDateTime` from Java 8 Date-Time package and `java.util.Date` where, when mapping a `ZonedDateTime` from a given `Date`, the system default timezone is used.
 
 * Between `java.time.LocalDateTime` from Java 8 Date-Time package and `java.util.Date` where timezone UTC is used as the timezone.
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractStaticParseToStringConversion.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractStaticParseToStringConversion.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.internal.conversion;
+
+import org.mapstruct.ap.internal.model.common.ConversionContext;
+
+/**
+ * Base class for a conversion between a type <tt>T</tt> and {@link String},
+ * where {@code T#parse(String)} and {@code T#toString} are inverse operations.
+ */
+public abstract class AbstractStaticParseToStringConversion extends SimpleConversion {
+
+    @Override
+    protected String getToExpression(ConversionContext conversionContext) {
+        return "<SOURCE>.toString()";
+    }
+
+    @Override
+    protected String getFromExpression(ConversionContext conversionContext) {
+        return getParsingType().getCanonicalName() + ".parse( <SOURCE> )";
+    }
+
+    /**
+     * @return the type which contains the <tt>parse(String)</tt> method
+     */
+    protected abstract Class<?> getParsingType();
+
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
@@ -9,10 +9,12 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.sql.Time;
 import java.sql.Timestamp;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
+import java.time.Period;
 import java.time.ZonedDateTime;
 import java.util.Calendar;
 import java.util.Currency;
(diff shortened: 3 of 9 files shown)


processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractStaticParseToStringConversion.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.mapstruct.ap.internal.conversion;
+
+import org.mapstruct.ap.internal.model.common.ConversionContext;
+
+/**
+ * Base class for a conversion between a type <tt>T</tt> and {@link String},
+ * where {@code T#parse(String)} and {@code T#toString} are inverse operations.
+ */
+public abstract class AbstractStaticParseToStringConversion extends SimpleConversion {
```
filiphr: I would get rid of this class and just do the needed conversion in it's implementations. Or would only have a single class where you pass in the class that needs to be used in the `getFromExpression`

processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractStaticParseToStringConversion.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    protected String getToExpression(ConversionContext conversionContext) {
+        return "<SOURCE>.toString()";
+    }
+
+    @Override
+    protected String getFromExpression(ConversionContext conversionContext) {
+        return getParsingType().getCanonicalName() + ".parse( <SOURCE> )";
```
filiphr: We try to generate code that is human readable as much as possible. Instead of doing the FQN it would be better to use the short name if it is possible. Have a look at the `ConversionUtils`. `CurrencyToStringConversion` is also a nice example class for how this can look like.

processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractStaticParseToStringConversion.java
twentylemon: Imo, this class should stay -- or, at least the code sharing it offers. It's easy to imagine additional classes with the `parse`/`toString` property that mapstruct would want to support. Joda Duration is the easy example.
I'm open to moving it to a utility class instead of a base class. I was moreso following the [AbstractJavaTimeToStringConversion](https://github.com/mapstruct/mapstruct/blob/cf668bea77d5d48f34d32c229823a04dbd01f21d/processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractJavaTimeToStringConversion.java) example when making the base class.

processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractStaticParseToStringConversion.java
twentylemon: I did start with that, similar to `CurrencyToStringConversion` as you mention. Later, I added the base class as I noticed the common implementation between the three new conversions. The problem I ran into was [typeReferenceName](https://github.com/mapstruct/mapstruct/blob/cf668bea77d5d48f34d32c229823a04dbd01f21d/processor/src/main/java/org/mapstruct/ap/internal/conversion/ConversionUtils.java#L55) is `private` but I wanted to stay generic in the base class.
Solution for now was to fully qualify the class, though I did want to revisit.

twentylemon:
> We are trying to generate code that is human readable as much as possible. This means that we don't want to have FQN if we can avoid this.

I had a feeling this would be called out. I had planned on importing the type instead of fully qualifying, but I'm more or less out of time. At some point I meant to write that in the pr description, but it seems I forgot.

> Also noticed that you are not adding yourself as an `@author` in new classes, is this on purpose or you just don't want to do it.

It's a habit now, I avoid `@author` in my daily life as the entire team owns the code. I can add if need be.

> Let us know if you don't have time to do the changes, we can quickly do them during the merge.

I may get some free time next weekend (Dec 8) to revisit.

processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractStaticParseToStringConversion.java
filiphr: No problem for me

processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractStaticParseToStringConversion.java
filiphr: We can make the `typeReferenceName` `public` no problem. Thanks for looking into it

filiphr:
>It's a habit now, I avoid `@author` in my daily life as the entire team owns the code. I can add if need be.

It is completely up to you. For us it doesn't make a big difference.

>I may get some free time next weekend (Dec 8) to revisit.

No problem, I'll see if I can do the merge some time this week. Thanks for taking the time to create the PR

filiphr:
@twentylemon I did some slight modifications. We can actually use a single class and get the target type from the `ConversionContext` (I forgot about that before 😄). If you are OK with the changes I am going to merge it into master.

processor/src/main/java/org/mapstruct/ap/internal/conversion/StaticParseToStringConversion.java
```diff
@@ -0,0 +1,36 @@
...
+package org.mapstruct.ap.internal.conversion;
+
+import java.util.Set;
+
+import org.mapstruct.ap.internal.model.common.ConversionContext;
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.util.Collections;
+
+/**
+ * Base class for a conversion between a target type <tt>T</tt> and {@link String},
```
twentylemon: Minor, update doc  to remove "base class" now. Something like "Handles conversion between..."

filiphr: Indeed you are right. Will update. Thanks

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -944,6 +944,8 @@ public interface CarMapper {
 
 * Between `java.time.ZonedDateTime`, `java.time.LocalDateTime`, `java.time.LocalDate`, `java.time.LocalTime` from Java 8 Date-Time package and `String`. A format string as understood by `java.text.SimpleDateFormat` can be specified via the `dateFormat` option (see above).
 
+* Between `java.time.Instant`, `java.time.Duration`, `java.time.Period` from Java 8 Date-Time package and `String`, as per the `parse` method in each class.
```
sjaakd: as per .. is this correct english?.. Perhaps: "it uses the `parse` method in each class"

filiphr: I think you are right, changed it

Issue #1665, opened 2018-12-05
Adder with primitive type not being used


bgK:
I am trying to map protobuf generated classes from POJOs using mapstruct 1.3.0.Beta2.

### Protobuf schema
```protobuf
syntax = "proto3";

option java_package = "test";

message CarMessage {
    repeated int64 tyre = 1;
    repeated string seat = 2;
}
```

### Source class
```java
package test;

import java.util.List;

public class Car {
    private List<Long> tyres;
    private List<String> seats;

    public List<Long> getTyres() {
        return tyres;
    }

    public void setTyres(List<Long> tyres) {
        this.tyres = tyres;
    }

    public List<String> getSeats() {
        return seats;
    }

    public void setSeats(List<String> seats) {
        this.seats = seats;
    }
}
```

### Mapper
```java
@Mapper(collectionMappingStrategy = CollectionMappingStrategy.ADDER_PREFERRED,
        nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS)
public interface CarMapper {
    @Mapping(target = "tyreList", source = "tyres")
    @Mapping(target = "seatList", source = "seats")
    Test.CarMessage map(Car value);
}
```

### Protobuf generated builder
```java
      public Builder addTyre(long value) {
        tyre_.add(value);
        return this;
      }

      public Builder addSeat(java.lang.String value) {
        seat_.add(value);
        return this;
      }
```

### Mapstruct generated mapper
```java
        if ( carMessage.getTyreList() != null ) {
            List<Long> list = value.getTyres();
            if ( list != null ) {
                carMessage.getTyreList().addAll( list );
            }
        }
        if ( value.getSeats() != null ) {
            for ( String seat : value.getSeats() ) {
                carMessage.addSeat( seat );
            }
        }
```

In the generated mapper, the adder is properly used for the `String` list, but not for the `Long` one, resulting in a runtime exception. I believe this is because mapstruct does not match the adder because it uses a primitive type whereas the POJO uses a boxed type. Shouldn't mapstruct do implicit conversion?

This is a reproducer project:
[buildfailurereproducer.zip](https://github.com/mapstruct/mapstruct/files/2649814/buildfailurereproducer.zip)

filiphr:
Fixed by https://github.com/mapstruct/mapstruct/pull/1666, thanks @seime for the PR

PR #1666, opened 2018-12-06
Box any primitives before attempting type comparison

seime:
Possible fix for #1665 . Could be that this problem is more widespread than for the ADDER_PREFERRED  collection style.


filiphr:
The solution looks OK. Can you just add a testcase for it? You can have a look at the bugs test package to get an idea how go write one. Let us know if you need help with it

filiphr:
Thanks a lot for the PR @seime, I have integrated it into master

Issue #1650, opened 2018-11-15
Nested mapping & MappingTarget - "cannot find symbol"


bcole:
If you have nested mapping (`a.b.c` -> `aPrime.cPrime`) and want to use a `@MappingTarget`, you will run into a Java compilation error.

Consider the following example.

Source classes:
```
public class A {
    private B b;
}

public class B {
    private C c;
}

public class C {}
```

Target classes:
```
public class APrime {
    private CPrime cPrime;
}

public class CPrime {}
```

And Mapper:
```
@Mapper
public interface AMapper {
    @Mapping(source = "b.c", target = "cPrime")
    APrime toAPrime(A a, @MappingTarget APrime mappingTarget);

    CPrime toCPrime(C c, @MappingTarget CPrime mappingTarget);
}
```

Trying to generate the mapper implementation will cause a Java compilation failure!
```
[ERROR] .../mapping/test/AMapperImpl.java:[22,23] cannot find symbol
[ERROR] symbol:   variable c
[ERROR] location: class mapping.test.AMapperImpl
```

If you look at the generated mapping, it fails to define / instantiate the `c` variable that it later references:
```
public class AMapperImpl implements AMapper {

    @Override
    public APrime toAPrime(A a, APrime mappingTarget) {
        if ( a == null ) {
            return null;
        }

        if ( aBC( a ) != null ) {
            if ( mappingTarget.getcPrime() == null ) {
                mappingTarget.setcPrime( new CPrime() );
            }
            // ***** `c` was never defined!! *****
            // ***** Implementation should have done something like: *****
            // ***** `C c = aBC( a );` *****
            toCPrime( c, mappingTarget.getcPrime() );
        }
        else {
            mappingTarget.setcPrime( null );
        }

        return mappingTarget;
    }

    @Override
    public CPrime toCPrime(C c, CPrime mappingTarget) {
        if ( c == null ) {
            return null;
        }

        return mappingTarget;
    }

    private C aBC(A a) {
        if ( a == null ) {
            return null;
        }
        B b = a.getB();
        if ( b == null ) {
            return null;
        }
        C c = b.getC();
        if ( c == null ) {
            return null;
        }
        return c;
    }
}
```
Note: This has been tested with versions 1.2.0.Final and the latest beta 1.3.0.Beta2

sjaakd:
Are you sure you both need update and creation behaviour? I'm considering whether we need support for this in the future. What's your use case?

bcole:
[...]

PR #1670, opened 2018-12-20
#1650 cannot find symbol: nested mapping and mappingtarget

chris922:
Added the local-var assignment and ensure that this local-var will be used in the following `if`-clause. This logic is similar to the one in `SetterWrapper.ftl`

@sjaakd As you also worked on this please take a look if the solution is fine for you.


sjaakd:
Hey @chris922 .. Looks very nice.. I just added one more commit (extended the test a bit).. For the remainder its: LGTM.. 😄

sjaakd:
@chris922  thanks! Merged in PR #1671

PR #1671, opened 2018-12-20
Chris922 1650 cannot find symbol nested mapping mappingtarget

sjaakd:


chris922:
Oh, great idea to add some life to the unit-test! :)

PR #1676, opened 2019-01-03
#1675 Making mapstruct-processor compileable with OpenJDK 11

### LINKED ISSUE #1675: NullPointerException when building on OpenJDK 11
Just tried to build MapStruct on OpenJDK 11.0.1 and am getting this exception:

```
Caused by: java.lang.NullPointerException
	at net.java.dev.hickory.prism.internal.PrismGenerator.getPrismName(PrismGenerator.java:122)
	at net.java.dev.hickory.prism.internal.PrismGenerator.process(PrismGenerator.java:108)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:972)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:888)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1214)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1326)
	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.processAnnotations(JavaCompiler.java:1258)
	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:936)
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.lambda$doCall$0(JavacTaskImpl.java:104)
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.handleExceptions(JavacTaskImpl.java:147)
	... 28 more
```

Is this a known issue, @filiphr?

gunnarmorling:
@filiphr This makes _mapstruct-processor_ compile for me with OpenJDK 11. But tests fail as the test runner seems not to find any classes from the JDK (stuff like `Integer` etc.). Have you encountered this, too?

Fixes #1675

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/testutil/assertions/JavaFileAssert.java
+++ b/processor/src/test/java/org/mapstruct/ap/testutil/assertions/JavaFileAssert.java
@@ -38,6 +38,8 @@ public class JavaFileAssert extends FileAssert {
         "\"\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\+\\d{4}\",";
     private static final String GENERATED_COMMENTS_REGEX = "\\s+comments = \"version: , compiler: .*, environment: " +
         ".*\"";
+    private static final String IMPORT_GENERATED_ANNOTATION_REGEX = "import javax\\.annotation\\.(processing\\.)?" +
+        "Generated;";
 
     private Diff diff = new Diff();
 
@@ -135,7 +137,8 @@ public class JavaFileAssert extends FileAssert {
         else if ( delta.getType() == Delta.TYPE.CHANGE ) {
             List<String> lines = delta.getOriginal().getLines();
             if ( lines.size() == 1 ) {
-                return lines.get( 0 ).matches( GENERATED_DATE_REGEX );
+                return lines.get( 0 ).matches( GENERATED_DATE_REGEX ) ||
+                    lines.get( 0 ).matches( IMPORT_GENERATED_ANNOTATION_REGEX );
             }
             else if ( lines.size() == 2 ) {
                 return lines.get( 0 ).matches( GENERATED_DATE_REGEX ) &&
--- a/processor/src/test/java/org/mapstruct/ap/testutil/runner/AnnotationProcessorTestRunner.java
+++ b/processor/src/test/java/org/mapstruct/ap/testutil/runner/AnnotationProcessorTestRunner.java
@@ -37,6 +37,19 @@ import org.mapstruct.ap.testutil.compilation.annotation.ProcessorOption;
  * @author Andreas Gudian
  */
 public class AnnotationProcessorTestRunner extends ParentRunner<Runner> {
+
+    private static final boolean IS_AT_LEAST_JAVA_9 = isIsAtLeastJava9();
+
+    private static boolean isIsAtLeastJava9() {
+        try {
+            Runtime.class.getMethod( "version" );
+            return true;
+        }
+        catch ( NoSuchMethodException e ) {
+            return false;
+        }
+    }
+
     private final List<Runner> runners;
 
     /**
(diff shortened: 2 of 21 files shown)


gunnarmorling:
Force pushed an update, things compile including _integration-test_ now. I gave up in the docs module due to that:

```
[ERROR] Failed to execute goal org.asciidoctor:asciidoctor-maven-plugin:1.5.3:process-asciidoc (output-pdf) on project mapstruct-documentation: Execution output-pdf of goal org.asciidoctor:asciidoctor-maven-plugin:1.5.3:process-asciidoc failed: (OpenSSL::X509::StoreError) setting default path failed: No password supplied for PKCS#12 KeyStore.
```

filiphr:
@gunnarmorling I finally had the time to push my branch and run your branch locally. What I did for Java 9 is https://github.com/filiphr/mapstruct/commit/a484e042820b233c11c85cd23c3615973e2df56e.

I also tried to make the tests on your branch run green. I pushed two commits (they should be squashed) that takes us in that direction. However, when running on Java 11 we still have 2 failures and 2 errors.

Both are connected to default date parsing (`DateConversationTest` and `JodaConversationTest`).
Back in the day I did some analysis (see https://twitter.com/filiphr/status/982855913177518080). And it turns out that the default Locale Data provider has been changed from Java 9. See https://bugs.openjdk.java.net/browse/JDK-8198491, http://www.oracle.com/technetwork/java/javase/9-relnote-issues-3704069.html#JDK-8008577 … and http://openjdk.java.net/jeps/252 . This leads to dates from Java 8 being unparseable in java 9.

Not sure if we need to do something about it or try to use dates that would be parseable on all Java JDKs.

How do you think we should proceed with this?

gunnarmorling:
@filiphr, would it help to provide explicit date formats in the test, and/or setting a specific Locale as the default locale?

filiphr:
>would it help to provide explicit date formats in the test, and/or setting a specific Locale as the default locale?

Setting specific date formats in the mapper it would help. I am not sure about the Locale though, we should try with the English one I suppose. Locally I have an `en_CH` locale 😄.

Maybe we could set the `java.locale.providers` property for those tests so they can be backwards compatible with Java 8. Something that I also found as a suggestion in [this](https://stackoverflow.com/questions/50411064/simpledateformat-with-german-locale-java-8-vs-java-10) SO question

processor/src/test/java/org/mapstruct/ap/testutil/assertions/JavaFileAssert.java
```diff
@@ -38,6 +38,8 @@
         "\"\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\+\\d{4}\",";
     private static final String GENERATED_COMMENTS_REGEX = "\\s+comments = \"version: , compiler: .*, environment: " +
         ".*\"";
+    private static final String IMPORT_GENERATED_ANNOTATION_REGEX = "import javax\\.annotation\\.(processing\\.)" +
+        "?Generated;";
```
chris922: Really really minor enhancement in case you still need to change something else: 
Move the `?` in the previous line. I just read this line several times and was wondering when a class/package name could have `?` in it until I realized this `?` is the usual reg-exp quantifier... :rofl:

processor/src/test/java/org/mapstruct/ap/testutil/runner/AnnotationProcessorTestRunner.java
```diff
@@ -57,6 +70,12 @@ public AnnotationProcessorTestRunner(Class<?> klass) throws Exception {
         if (singleCompiler != null) {
             return Arrays.<Runner> asList( new InnerAnnotationProcessorRunner( klass, singleCompiler.value() ) );
         }
+        else if ( IS_AT_LEAST_JAVA_9 ) {
+            // Current tycho-compiler-jdt (0.26.0) is not compatible with Java 11
+            // Updating to latest version 1.3.0 fails some tests
+            // Once https://github.com/mapstruct/mapstruct/pull/1587 is resolved we can remove this line
+            return Arrays.asList( new InnerAnnotationProcessorRunner( klass, Compiler.JDK11 ) );
```
chris922: "is at least Java 9" and then providing JDK 11 looks a bit strange for me... if it is Java 9, 10, 12, ... then JDK11 compiler is correct as well?!

processor/src/test/java/org/mapstruct/ap/testutil/assertions/JavaFileAssert.java
filiphr: Changed it

processor/src/test/java/org/mapstruct/ap/testutil/runner/AnnotationProcessorTestRunner.java
filiphr: Not sure from which Java the change was done, but we are testing with Java 11 so I think it is fine

filiphr:
I changed the approach. @gunnarmorling all my commits can be merged into one. In case you don't have the time to do it, I can do it later today or tomorrow (will need to create a new PR though)

Issue #1406, opened 2018-03-27
Flat target object, nested source objects, target = "." not working


afdia:
I often come across situations where objects have fields on different levels (e.g. target has flat fields, source has multiple nested objects and fields are on different levels)

Let's take the following example (getters/setters removed for clarity):
```java
public class MapstructExample {
public static class FlatTarget {
	public String targetA;
	public String targetB;
}
public static class NestedOuter {
	public String sourceA;
	public NestedInner nestedInner;
}
public static class NestedInner {
	public String sourceB;
}
@Mapper
public static interface MyMapper {
	@Mapping(target = "targetA", source = "sourceA")
	@Mapping(target = "targetB", source = "nestedInner.sourceB") // redundant mapping info
	//@Mapping(target = ".", source = "nestedInner") // not working
	FlatTarget convert1(NestedOuter in);

	@Mapping(target = "targetB", source = "sourceB") // redundant mapping info
	FlatTarget convert2(NestedInner in);

	@Mapping(target = "sourceA", source = "targetA")
	@Mapping(target = "nestedInner", source = ".") // no redundancy because "." works
	NestedOuter convert3(FlatTarget in);

	@Mapping(target = "sourceB", source = "targetB")
	NestedInner convert4(FlatTarget in);
}
}
```

As you can see the last 2 mappings have no repetition, because I can declare `source = "."` and it passes the main object to the next mapping method.
Unfortunately `target = "."` doesn't seem to work (and applying `@InheritInverseConfiguration(name="convert3")` to the first method causes and error)
I have also tried to formulate all converter method using `@MappingTarget` but it doesn't change the behavior.

Is there a way to avoid repeating the mappings of all fields of the NestedInner object? `@InheritConfiguration` is also not an option because the mappings are declared on a different layer (every field is prefixed with `nestedInner` in the second method).
Typically I work around the issue by declaring `@AfterMapping` methods which call the other method to fill in the missing parts of the objects, but I hope there is a better way to declare these mappings (or perhaps there is a way to get `target = "."` to work which would allow automatic inversion of mappings using `source = "."`)

------

EDIT: I have reformatted the example in order to have a fully working example and named the convert methods from 1-4 for clarity

sjaakd:
Try to specify: `in.sourceA` and `in.nestedInner.sourceB`.. So include the parameter name as well.

afdia:
[...]

PR #1686, opened 2019-01-14
#1406 Targeting "." as current object allowing to map source properties to current target

dainiusfigoras:
It allows to map composition to inheritance and opposite way, for example "@Mapping(source = "dataObject.*", target = ".")" every property in dataObject is mapped to return type, this allows to have hierarchy target object and composition source object.


sjaakd:
@dainiusfigoras thanks for the contribution.. 👍  I'll try to have a look as soon as possible.

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
```diff
@@ -0,0 +1,36 @@
...
+import org.mapstruct.Mapping;
+import org.mapstruct.ReportingPolicy;
+import org.mapstruct.factory.Mappers;
+
+@Mapper(
+        unmappedSourcePolicy = ReportingPolicy.IGNORE,
+        unmappedTargetPolicy = ReportingPolicy.IGNORE,
+        collectionMappingStrategy = CollectionMappingStrategy.ADDER_PREFERRED
+)
+public interface CustomerEntityMapper1 {
```
sjaakd: What would be interesting to see are testing with:
1. existing targets (`@MappingTarget`)
2. reverse mappings `@InheritInverseConfiguration)

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
```diff
@@ -0,0 +1,36 @@
...
+    CustomerEntityMapper1 INSTANCE = Mappers.getMapper( CustomerEntityMapper1.class );
+
+    @Mapping(target = ".", source = "entity.*")
```
sjaakd: I'm not sure about the `*` notation. Suppose that entity has another nesting level, would all elements from that level be mapped as well?
so, would both `prop1` and `prop2` be mapped?
`entity.prop1`
`entity.subentity.prop2`

Would this also happen with nested targets: so not `.` but `target = "target.prop2", source ="*"`?

sjaakd: what would the latter mean if you have multiple source parameters? Like: `Target map(Source s1, Source2)`?

dainiusfigoras: > I'm not sure about the `*` notation. Suppose that entity has another nesting level, would all elements from that level be mapped as well?
> so, would both `prop1` and `prop2` be mapped?
> `entity.prop1`
> `entity.subentity.prop2`
> 
> Would this also happen with nested targets: so not `.` but `target = "target.prop2", source ="*"`?

In my head '*' means all attributes from that object, so 'entity.*' should try to map all properties from entity, 'entity.subentity.*' should map all properties from subentity, but not from entity.
'.' just way to say that mapping should be done to current target.
Ideally I would say that it would be good if it could target some other property as target 'target=entity.subentity source=dto.*', could try to map everything from 'dto' to 'entity.subentity', but it's not done in this PR.

parent/pom.xml
```diff
@@ -638,7 +638,7 @@
                                     <version>[1.8,)</version>
                                 </requireJavaVersion>
                                 <DependencyConvergence />
-                                <requirePluginVersions />
+                                <!--<requirePluginVersions />-->
```
sjaakd: why is this uncommented?

processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
```diff
@@ -507,19 +509,91 @@ private boolean checkParameterAndReturnType(ExecutableElement method, List<Param
             if ( !mappings.containsKey( mappingAnnotation.target() ) ) {
                 mappings.put( mappingAnnotation.target(), new ArrayList<>() );
             }
-            Mapping mapping = Mapping.fromMappingPrism( mappingAnnotation, method, messager, typeUtils );
-            if ( mapping != null ) {
-                mappings.get( mappingAnnotation.target() ).add( mapping );
+
+            if (mappingAnnotation.source().endsWith( ".*" )) {
```
sjaakd: The logic of analysing source reference / target reference is in the classes `SourceReference` and `TargetReference`.. Looking from a separation of concerns, I've the feeling this is not the proper place to put this kind of analysis.

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
sjaakd: I'm missing some enters :smile: 

So, to be clear (just to make sure I interpret your text above correctly):
* `entity.*' should map all properties from entity (but **not** nested properties from `entity.subentity.`)
* `entity.subentity.*` should map all properties from `subentity` but **not** from `entity`..
??

parent/pom.xml
dainiusfigoras: sorry, it's from local build. Because of some strange reasons it fails locally.

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
dainiusfigoras: lets say there's source:

Source {
id;
created;
updated;
object;
}

Object {
prop1;
prop2;
}

@mapping("target=.", "source=Source.*")

should generate:

this.setId(source.id);
this.setCreated(source.created);
this.setUpdated(source.updated);
this.setObject(source.object); // or this.setObject( map(source.object) ) if there is special mapping for object.

dainiusfigoras: > I'm still analysing.. perhaps its good that @filiphr also voices his opinion..
> 
> However, I've the feeling that this PR solves 2 problems..
> 
> 1. #1406 (the target =".")
> 2. grouping of source properties:
> 
> Is 2. required to solve 1.? Otherwise it is perhaps better to split this into 2 PR's. The first one (this one) could solve the problem #1406... It requires testing some more scenarios.
> 
> The second one (the '*' notation to group properties) i'm still doubting about.. What does it offer extra..? We have name based mapping, so if names are the same in target and source property they are mapped automatically. You normally only have to specify the mappings that differ in name. (That can be switched off by an ignore option on bean mapping level: `BeanMapping#ignoreByDefault`)

1. Yes, it't related to #1406, as it allows to point to target object without actually writing name of that target.
2. to solve this, it requires some way to name, where multiple source properties should be mapped, so 1. is kinda required (or some other way, for example skipping target at all).

motivation here is that some parent objects in Source class hierarchy have fields that is always mapped to target objects, so to avoid repetition of copying same lines, we map all attributes to Target class, that have exactly same fields, just flat (aka in same level). This lets significantly decrease Mapper class (6 attributes in 50 classes, mapping both ways) and it reduces possible errors of missing to copy some of attributes and even worst if in future someone would new base class property that should be mapped to every single object.

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
dainiusfigoras: I updated mapper in test to include @InheritInverseConfiguration and @MappingTarget (for update).

dainiusfigoras: It was nice to see that @InheritInverseConfiguration generated exactly same code as with Mapping annotation. Just had to add name after putting there methods for update, as compiler got confused which one to use.

sjaakd:
@dainiusfigoras .. thanks for adding the unit test 😄 !!.. I have some time in the weekend to review more in depth. I'll get back to you soon.

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
```diff
@@ -0,0 +1,45 @@
...
+    // @Mapping(target = "entity", source = ".")
+    CustomerDTO inverseMap(CustomerEntity customer);
+
+    @Mapping(target = "entity", source = ".")
+    void update(CustomerEntity customer, @MappingTarget CustomerDTO dto);
+
+    @Mapping(target = ".", source = "entity.*")
+    void update(CustomerDTO dto, @MappingTarget CustomerEntity customer);
+
+    @Mapping(target = ".", source = "customer.entity.*")
```
sjaakd: I don't think the `*` should be needed: `source = "customer.entity" ` should be sufficient. Although it might actually reflect better what you intent, it would also mean inconsistency with the current solution. Name base mapping is applied even without any `@Mapping` annotation. In the future we could even consider to reserve the `.*` for doing scanning on deeper nesting.

sjaakd: I played around with this.. I you leave out the source altogether you get an NPE in annotation processing.. That cannot be the intention.

dainiusfigoras:
> 1. Why the need for the `.` notation in `source`?
> 
> Instead of
> 
> ```
> @Mapping(target = "entity", source = ".")
> OrderDTO map( OrderEntity order );
> ```
> why not use what is now possible:
> 
> ```
> @Mapping(target = "entity", source = "order")
> OrderDTO map( OrderEntity order );
> ```
> I think I know why you went for `.`. Is it because you want to use it in inheritance? In this case why not use a reserved word like `<CURRENT>` or `<SOURCE>`

Is it possible to use 'Current' now? If not I can see at least 2 reasons not to use it: 'current' might be used as property in some class and then it will be impossible to distinct between actual property and reserved word. And if I'm not mistaken, it is possible to use opposite syntax today - @Mapping(source = ".", target = "entity") - is valid mapping.

> 
> 1. Why the need of the `.*` notation
> 
> Instead of
> 
> ```
> @Mapping(target = ".", source = "customer.entity.*")
> @Mapping(target = "name", source = "customer.name")
> CustomerEntity map(OrderDTO order);
> ```
> Why not
> 
> ```
> @Mapping(target = ".", source = "customer.entity")
> @Mapping(target = "name", source = "customer.name")
> CustomerEntity map(OrderDTO order);
> ```
> With `.` we clearly say that the `customer.entity` should be mapped into the current target (`.`). Maybe instead of going for `.` we can go for some reserved word like `<CURRENT>` or `<TARGET>`.

What code would "@Mapping(target = ".", source = "customer.entity")" generate?
In my mind it's something like:

CustomerEntity map(OrderDTO order) {
 ...
 this.setEntity( order.getCustomer().getEntity() );
 this.setName( order.getCustomer().getName() );
...
}

while desired code is more like this:

 ...
 this.setId( order.getCustomer().getEntity().getId() );
 this.setCreated( order.getCustomer().getEntity().getCreated() );
 this.setName( order.getCustomer().getName() );
...
}

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
dainiusfigoras: > I don't think the * should be needed: source = "customer.entity"

But if I have <source = "customer.entity"> then mapper will try to set entity object to current target <target,setEntity(entity)>, right? While desire is to map every property from [source] entity to current target without need of naming every property.

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
```diff
@@ -88,10 +96,51 @@
         return mappings;
     }
 
+    public static Mapping artificialMapping(String source, String target) {
```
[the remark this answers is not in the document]
dainiusfigoras: I used source ".*" as replacement for mapping all properties from source and because I don't know mapstruct so well it was easier for me to read all properties from source object, construct mapping relation to current object (kinda if I would manually write mapping for every property) and just reuse default functionality of library to do the rest. 
But if you think that more right way would be do it in BeanMappingMethod, I can try that as well.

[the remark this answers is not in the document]
dainiusfigoras: Thanks, I can try to do that myself, but if I stuck with some part then would be nice to have someone that could help with that part.

sjaakd:
> Why not
> 
> ```
> @Mapping(target = ".", source = "customer.entity")
> @Mapping(target = "name", source = "customer.name")
> CustomerEntity map(OrderDTO order);
> ```

👍  @dainiusfigoras .. That's what I tried to say as well. The concept of name based mapping should take care of mapping all properties inside `customer.entity` to the target (flatten). This is in line what we have now. In case of non-nested-mappings:  If `source` and `target` share properties with the same name, you don't have to specify anything (also not `.*`)



> With `.` we clearly say that the `customer.entity` should be mapped into the current target (`.`). Maybe instead of going for `.` we can go for some reserved word like `<CURRENT>` or `<TARGET>`.

I prefer the `.`.. Its intuitive (for me it relates to file/systems and directory structures, where `.` reflects `current` or `working dir`). Also, we use existing notation (the `.` is already used, introducing new gives rise to new complexity)

dainiusfigoras:
> 👍 @dainiusfigoras .. That's what I tried to say as well. The concept of name based mapping should take care of mapping all properties inside `customer.entity` to the target (flatten). This is in line what we have now. In case of non-nested-mappings: If `source` and `target` share properties with the same name, you don't have to specify anything (also not `.*`)
> 
> > With `.` we clearly say that the `customer.entity` should be mapped into the current target (`.`). Maybe instead of going for `.` we can go for some reserved word like `<CURRENT>` or `<TARGET>`.

I think one of main reasons why I like `property.*`, because it removes any double meaning for me what does `target="." source="entity"` could mean. 
Will it generate code `this.setEntity(source.entity)` or will it generate code `this.setId(source.entity.getId(); this.setStatus(source.entity.getStatus());`

But I realised that because it isn't possible to map to current target directly, no one could used that before, so probably `@Mapping(target=".", source="entity")` will work fine. And I guess if someone would need to map actual object instead of subproperties, he could map directly by name `@Mapping(target="entity", source="entity")`.

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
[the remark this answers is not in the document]
dainiusfigoras: I removed '.*' from source part, so if using target="." it will try to map everything from source property (and if source is missing, it will just do default thing and maps everything from source to target). Also moved actual mapping generation to BeanMappingMethod.

sjaakd:
> I think one of main reasons why I like `property.*`, because it removes any double meaning for me what does `target="." source="entity"` could mean

I agree it offers more clarity.. But then we should do this for existing functionality as well.

We could (at one point, not now that is) add is as a kind of syntactic sugar, both supporting the old way and the new way (so de-factor ignoring the `.*` at the end. We might even add deep nesting in the future (based on name), by allowing `**` (matching sub-entities).

seime:
@filiphr & @sjaakd what do you think - are we getting close to approving this PR? I think it solves (at least partially) a few other issues related to multilevel vs flat structures.

sjaakd:
@seime & @dainiusfigoras . Thanks for all your effort. It really appreciated. 

 I need some time to check this.. Will try to do this later this week. To manage expectations, I think 1.3 will come too soon to include this work.  If/when we include it we should test this in beta. @filiphr : WDYT?

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerDTO.java
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.targetthis;
+
+public class CustomerDTO {
+    private String name;
+    private int level;
+    private EntityDTO entity;
```
sjaakd: it is a bit confusing to stick an class EntityDTO inside a customerDTO.. Wouldn't `StatusDto` be better? (you normally map a DTO to an entity)

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
```diff
@@ -0,0 +1,45 @@
...
+    CustomerEntityMapper1 INSTANCE = Mappers.getMapper( CustomerEntityMapper1.class );
+
+    @Mapping(target = ".", source = "entity")
+    CustomerEntity map(CustomerDTO customer);
+
+    @InheritInverseConfiguration(name = "map")
+    // @Mapping(target = "entity", source = ".")
+    CustomerDTO inverseMap(CustomerEntity customer);
+
+    @Mapping(target = "entity", source = ".")
```
sjaakd: its currently (without this solution) possible to specify as source: `parameter`, `parameter.property` and `property` (omitting `parameter` altogether).  If you have multiple sources I think ~~(need to check)~~ the parameter designation is mandatory: so `parameter1.propX`, `parameter2.propY`. BTW: this is stuff inside `SourceReference`. How does `"."` relate to all the other options on source side? Will it help or confuse the user to have an additional `source=` option?

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
```diff
@@ -0,0 +1,45 @@
...
+    void update(CustomerEntity customer, @MappingTarget CustomerDTO dto);
+
+    @Mapping(target = ".", source = "entity")
```
sjaakd: could you also specify `customer` as target?

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
sjaakd: you probably need an additional test with multiple source arguments. So something like:
```java 
@Mapping( target = ".", source = "param1" ) // first set of properties, lets say "propX", "propY"
@Mapping( target = ".", source= "param2" ) // second set of properties, lets say "propU", "propW"
Target map( Source1 param1, Source2 param2 );
```

and

```java 
@Mapping( target = ".", source = "param1" ) // first set of properties, lets say "propX", "propY"
@Mapping( target = ".", source= "param2" ) // second set of properties, lets say "propU", "propW" 
void map( @MappingTarget Target target, Source1 param1, Source2 param2 );
```

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -361,6 +363,24 @@ public int compare(PropertyMapping o1, PropertyMapping o2) {
             }
         }
 
+        private Type getPropertyType(List<String> nameDetails, Type currentType, String parentName) {
+            if ( nameDetails.size() > 0 ) {
```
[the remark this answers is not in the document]
dainiusfigoras: Nice. That saves some code.

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerDTO.java
dainiusfigoras: Entity here is just generic name for a thing, for practice purposes it's just to mimic hierarchical/flat structure:
Customer extends Entity {}; // when you have inheritance
CustomerDTO { EntityDTO }; // when you don't have inheritance.

This is a bit of mirroring our data model and there Entity is basic class that every other inherits (directly or via another parent). In some other models I have seen "Record", "Item" and similar things. On it's own Entity would be like abstract class as it doesn't hold any "real world" value, but contains properties that every other object must have (typically it's something like "created, updated, status, deleted etc").
StatusDTO would be like proper attribute for CustomerDTO, but IMHO it would be wrong to inherit from it, aka "Customer extends Status {}".

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
dainiusfigoras: probably not at this point, there is explicit check for target "." to read all sub-properties from source, if you use something else it should work exactly as before, so nothing is broken for others.

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
sjaakd: Thinking of that a bit more.. Perhaps we should allow `"."`  ony in one specific case:  when we have no `@MappingTarget` to indicate the target (then the user could refer to the `@MappingTarget` annotated parameter.  This to avoid ambiguity.

We should not allow it in the source mapping attribute **at all** . So we probably have to do some checking (+ unit test) when parsing the mirror in the `[Mapping](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java)` class. Also, we should take care of the `@InheritReverseConfiguration` (`Mapping#invert")? to strip this..

So this is ok:

```java 
@Mapping( target = "." ) 
Target map( Source1 param );
```
But, this is **not ok**:
```java 
@Mapping( target = "." ) 
void map( Source1 param, @MappingTarget Target target );
```
```java 
@Mapping( target = "." ) 
void Target map( Source1 param, @MappingTarget Target target );
```
```java 
@Mapping( target = "target.property", source = "." ) 
Target map( Source1 param );
```

dainiusfigoras: I think most common case for us is
`
@Mapping( target = ".", source = "param1" )
@Mapping( target = ".", source= "param2" )
Target map( SourceObject source  );
`

but I can see that sometimes I would like to use something like 
`
@Mapping( target = ".", source = "param1" )
@Mapping( target = ".", source= "param2.innerproperty" )
Target map( SourceObject source );
`
and it's probably difficult to have "innerproperty" as parameter (just guessing).

Also what will happen if you have some property as target and target mapping?
`
@Mapping( target = "name" ) 
void map( Source1 param, @MappingTarget Target target );
`

In my mind information in @Mapping is guidance from direct mapping Source.property -> Target.property. And in such way it's Ok to have @Mapping and MappingTarget, as:

`
@Mapping( target = ".", source = "param1" )
@Mapping( target = ".", source= "param2.innerproperty" )
public void map( SourceObject source, @MappingTarget Target target );
`
To be honest I haven't tested that, but for me it doesn't create any ambiguity because I see it as "set all properties from Source.param1 to current target (target), set all properties from innerproperty that is nested property on param2 that is property of Source to current target (target) and map all other properties from source to target)". So in this case only difference I can see is that Target is moved from return parameter to update parameter (and that of course will force me to create object myself first).
When there are more than one source parameters it gets a bit ambiguous  in such way, that I might not be sure if param1 belongs to source1 and param to belongs to source2, or they can be in any of these objects, but I would have some thing with named target parameter anyway.

And just to be on same page, target="." without source makes no sense for me, because that's default behaviour anyway (mapping all properties from source to target).

dainiusfigoras: btw I wonder about :

```
void map( @MappingTarget Target target, Source1 param1, Source2 param2 );
```

what happens when param1 and param2 have property with same name, or they need to be mapped to different target properties (param1.name -> target.name, param2.name -> target.title)?

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerDTO.java
sjaakd: > Entity here is just generic name for a thing, for practice purposes it's just to mimic hierarchical/flat structure:
> Customer extends Entity {}; // when you have inheritance
> CustomerDTO { EntityDTO }; // when you don't have inheritance.

i understood as much.. However, here the purpose is to make clear that you map A to B. If the same "terminology"  is used in A and B this will be confusing. I guess that many developers automatically associate entity with a JPA entity (although that is not your intention) and DTO with an object that should be converted in an entity (and visa versa).




> this is a bit of mirroring our data model and there Entity is basic class that every other inherits (directly or via another parent).

Same here.. In my line of business we also talk about generic entities and attributes... So I can understand were you're coming from.. But its not about me only :smile:

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
sjaakd: @dainiusfigoras .. Its a lot of text to gauge.. But lets try to setup a small convention.. 

* With `param` I mean a parameter (argument) of the mapping method. 

So with respect to a mapping method you have:
1. Always one target parameter (`@MappingTarget`) and / or return
2. One or more source parameters
3. One or more (@Context) annotated context parameters

* With `property` I mean a property in as in `bean properties`, exposed via a readAccessor / writeAccessor (normally getter / setter)

We try to make use of that consistently, when working on the source code and on examples. Take for instance a look  `SourceReference` / `TargetReference` in the source code of MapStruct.

---

Now, I am concerned about complexity.. The API should be clear. That should always have our focus. It can't be one use case that we try to implement opens up a can of worms for all kinds of other use cases (believe me, this happens :smile: ) that we need to correct with patches.

 I'll try to explain by my examples what I meant case by case:

> But, this is **not ok**:

**Case 1:**

> 
> ```java
> @Mapping( target = "." ) 
> void map( Source1 param, @MappingTarget Target target );
> ```

Here the user already has a possibility to refer to the target  by means of `target="target"`. Opening up more possibilities increases complexity. 

- if we detect the setup above we should give an error (fail fast)
- if we encounter an `@InheritConfiguration` from a mapping method with a "return type only" like this:

```java
@InheritConfiguration( name = "map" ) // the method above
Target map( Source1 param );
```
we should handle this correctly in the source code (`Mapping#targetReference`) I guess. 

 **Case 2:**
> ```java
> @Mapping( target = "." ) 
> void Target map( Source1 param, @MappingTarget Target target );
> ```
> 
- if we detect the setup above we should give an error (fail fast)
- same reasoning as above. We have a way to indicate the target by a FQN (`target="target"`) because the `@MappingTarget` identical to the return type

**Case 3:**
> ```java
> @Mapping( target = "target.property", source = "." ) 
> Target map( Source1 param );
> ```
- if we detect the setup above we should give an error (fail fast)
- Allowing `source="."` opens a new possibility next the ones we already have (see my explanation above). So let's not do that.  In case of a method annotated with `@InheritInverseConfiguration` we should adapt in our source code (`Mapping#inverse`) to point to the write source parameter. So example:
```java
@Mapping( target = ".", source = "param" ) 
Target map( Source1 param );

@InheritInverseConfiguration // from the method above
Source map( Target param);
```

sjaakd: > I think most common case for us is
> 
> ```java
> @Mapping( target = ".", source = "param1" )
> @Mapping( target = ".", source= "param2" )
> Target map( SourceObject source  );
> ```

I guess you mean 
```java
 @Mapping( target = ".", source = "property1" )
@Mapping( target = ".", source= "property2" )
Target map( SourceObject source  );
```
Or identical to this (fully qualifying parameter as well): 

```java
 @Mapping( target = ".", source = "source.property1" )
@Mapping( target = ".", source= "source.property2" )
Target map( SourceObject source  );
```

> but I can see that sometimes I would like to use something like
> 
> ```java
> @Mapping( target = ".", source = "param1" )
> @Mapping( target = ".", source= "param2.innerproperty" )
> Target map( SourceObject source );
> ```
```java
 @Mapping( target = ".", source = "property1" )
@Mapping( target = ".", source= "property2.innerProperty" )
Target map( SourceObject source  );
```
?
Yes that should be supported as well, meaning that all properties (except `property1.innerproperty`  should be mapped from level `property1` combined with all properties from `property1.innerproperty` 

> and it's probably difficult to have "innerproperty" as parameter (just guessing).
> 
Not sure what you mean here. 

> Also what will happen if you have some property as target and target mapping?
> 
> ```java
> @Mapping( target = "name" ) 
> void map( Source1 param, @MappingTarget Target target );
> ```
Not sure what this would mean / or what the user would intend with this.. I guess MapStruct would give an error.  (Nice to try BTW). But cases with only a parameter are possible.. For instance:
 ```java
@Mapping( target = "name", qualfiedBy = "SomeQualfier.class"  ) 
void map( Source1 param, @MappingTarget Target target );
```
Meaning that there would be a `param.name` as present as well in the source that can be mapped based on its name. 
Meaning: the

Comment (sjaakd) [reply processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java]: > In my mind information in @mapping is guidance from direct mapping Source.property -> Target.property.

`@Mapping` is a guidance where MapStruct cannot make a decision on its own. For instance, when property names differ, or when there's a conflict in which method / conversion to select.

Comment (sjaakd) [reply processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java]: > ```java
> @Mapping( target = ".", source = "param1" )
> @Mapping( target = ".", source= "param2.innerproperty" )
> public void map( SourceObject source, @MappingTarget Target target );
> ```
> 
> To be honest I haven't tested that, but for me it doesn't create any ambiguity because I see it as "set all properties from Source.param1 to current target (target), set all properties from innerproperty that is nested property on param2 that is property of Source to current target (target) and map all other properties from source to target)". So in this case only difference I can see is that Target is moved from return parameter to update parameter (and that of course will force me to create object myself first).
> When there are more than one source parameters it gets a bit ambiguous in such way, that I might not be sure if param1 belongs to source1 and param to belongs to source2, or they can be in any of these objects, but I would have some thing with named target parameter anyway.

I hope I made my point clear above. It is not about the workings of such mappings.. The simple reason I don't like to have this is because you can also specify:
```java
@Mapping( target = "target", source = "param1" )
@Mapping( target = "target", source= "param2.innerproperty" )
public void map( SourceObject source, @MappingTarget Target target );
 ```
Or alternatively:
```java
@Mapping( target = "target", source = "source.param1" )
@Mapping( target = "target", source= "source.param2.innerproperty" )
public void map( SourceObject source, @MappingTarget Target target );
 ```

File: processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java (hunk 9 above)
Comment (sjaakd) [reply processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java]: > could you also specify `customer` as target?

I'll open up another issue for that.. I'm not sure if its possible already, but I think its not the in the scope of this PR.. (See #1705)

File: processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerDTO.java (hunk 7 above)
Comment (dainiusfigoras) [reply processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerDTO.java]: We aren't using database in that part of a project, so I totally forgot about database, but when tried to look with a bit of "fresh" eyes, first thing that came when reading entity was JPA. 
I changed everywhere to "item", hopefully it still logical (Customer extends Item) but with less confusion.

File: processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java (hunk 8 above)
Comment (sjaakd) [reply processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java]: > Case3 should not work and give compile error.
I would say: a MapStruct error.. It should be detected in: `Mapping#fromMappingPrism` (there are already many of such checks there). 

> It sounds a bit dangerous because that syntax is working in current release and removing it would mean changing existing behaviour.
How do you mean? I was not aware I could currently specify `source="."`?

File: processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java [hunk 11]
```diff
@@ -407,6 +412,120 @@ private boolean handleDefinedMappings() {
             return errorOccurred;
         }
 
+        /**
+         * Iterates over all source parameter properties and maps them to current target.
+         * When a property name match occurs, the remainder will be checked for duplicates. Matches will be removed from
+         * the set of remaining target properties.
+         */
+        private void applySourcePropertiesToCurrentTarget( Mapping mapping ) {
```
filiphr: Isn't this method more or less the same as `applyPropertyNameBasedMapping`? Can we somehow combine it so we don't need to do a lot of manual things in here? I don't think that the methods that were added to support this method belong here really.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -384,13 +404,17 @@ private boolean handleDefinedMappings() {
             for ( Map.Entry<String, List<Mapping>> entry : methodMappings.entrySet() ) {
                 for ( Mapping mapping : entry.getValue() ) {
                     TargetReference targetReference = mapping.getTargetReference();
-                    if ( targetReference.isValid() ) {
+                    if ( !mapping.getTargetName().equals( "." ) && targetReference.isValid() ) {
                         if ( !handledTargets.contains( first( targetReference.getPropertyEntries() ).getFullName() ) ) {
                             if ( handleDefinedMapping( mapping, handledTargets ) ) {
```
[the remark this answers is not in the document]
dainiusfigoras: It probably could, but eventually it's necessary somewhere to check that target is ".", so we don't map concrete property, but mapping multi properties.
In this case, `targetReference` is also set as valid, but it doesn't have any properties, so it fails before `handleDefinedMapping` (but when trying to get first property from property entries).

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
dainiusfigoras: applyPropertyNameBasedMapping was example for applySourcePropertiesToCurrentTarget (in other words I copied applyPropertyNameBasedMapping and changed it to do stuff with current target).
Main difference is that in `applyPropertyNameBasedMapping` it checks if there is property reader for method source parameter, while in `applySourcePropertiesToCurrentTarget` it uses source to get readers one level down (and use source from mapping not method). `applySourcePropertiesToCurrentTarget` also always maps to current current target object.
I would say 2 key difference are: 
source readers and SourceReference creation.
I guess mainly because `applySourcePropertiesToCurrentTarget` is called for mapping annotation, while `applyPropertyNameBasedMapping` is called for target properties without explicit mapping annotations.

dainiusfigoras: > Can we also have some more tests with second or third level nesting? I would like to see whether this works.

In CustomerMapper1 I have defined this:
```

    @Mapping( target = ".", source = "customer.item" )
    OrderItem map(OrderDTO order );

```

and code for it looks like this:
```

    public OrderItem map(OrderDTO order) {
        if (order == null) {
            return null;
        } else {
            OrderItem orderItem = new OrderItem();
            String id = this.orderCustomerItemId(order);
            if (id != null) {
                orderItem.setId(id);
            }

            int status = this.orderCustomerItemStatus(order);
            orderItem.setStatus(status);
            if (order.getOrderLines() != null) {
                Iterator var5 = order.getOrderLines().iterator();

                while(var5.hasNext()) {
                    OrderLineDTO orderLine = (OrderLineDTO)var5.next();
                    orderItem.addOrderLine(this.map(orderLine));
                }
            }

            orderItem.setCustomer(this.map(order.getCustomer()));
            return orderItem;
        }
    }

    private String orderCustomerItemId(OrderDTO orderDTO) {
        if (orderDTO == null) {
            return null;
        } else {
            CustomerDTO customer = orderDTO.getCustomer();
            if (customer == null) {
                return null;
            } else {
                ItemDTO item = customer.getItem();
                if (item == null) {
                    return null;
                } else {
                    String id = item.getId();
                    return id == null ? null : id;
                }
            }
        }
    }

    private int orderCustomerItemStatus(OrderDTO orderDTO) {
        if (orderDTO == null) {
            return 0;
        } else {
            CustomerDTO customer = orderDTO.getCustomer();
            if (customer == null) {
                return 0;
            } else {
                ItemDTO item = customer.getItem();
                if (item == null) {
                    return 0;
                } else {
                    int status = item.getStatus();
                    return status;
                }
            }
        }
    }


```

dainiusfigoras:
> Also I am not entirely sure how the generated code looks like. Is it possible to add a fixture for it in the PR?

Not sure, how to add fixture, but I can put some generated code for main use case (it's pretty simple) and data model definition.

```

class ItemDTO {
    private String id;
    private int status;
}

class CustomerDTO {
    private String name;
    private int level;
    private ItemDTO item;
}

class Item {
    private String id;
    private int status;
}

class CustomerItem extends Item {
    private String name;
    private int level;
}


// Mapping definition:

@Mapping( target = ".", source = "item" )
CustomerItem map(CustomerDTO customer );



// Generated code:

public CustomerItem map(CustomerDTO customer) {
        if (customer == null) {
            return null;
        } else {
            CustomerItem customerItem = new CustomerItem();
            String id = this.customerItemId(customer);
            if (id != null) {
                customerItem.setId(id);
            }

            int status = this.customerItemStatus(customer);
            customerItem.setStatus(status);
            customerItem.setName(customer.getName());
            customerItem.setLevel(customer.getLevel());
            return customerItem;
        }
}

private String customerItemId(CustomerDTO customerDTO) {
        if (customerDTO == null) {
            return null;
        } else {
            ItemDTO item = customerDTO.getItem();
            if (item == null) {
                return null;
            } else {
                String id = item.getId();
                return id == null ? null : id;
            }
        }
}

private int customerItemStatus(CustomerDTO customerDTO) {
        if (customerDTO == null) {
            return 0;
        } else {
            ItemDTO item = customerDTO.getItem();
            if (item == null) {
                return 0;
            } else {
                int status = item.getStatus();
                return status;
            }
        }
}

```

sjaakd:
@dainiusfigoras : for fixtures, look at one of the testcases in the `/src/test/resources/fixtures`.. A testcase should include:

```java
    @Rule
    public final GeneratedSource generatedSource = new GeneratedSource().addComparisonToFixtureFor(
        UserMapper.class
    );

```
See for instance `Issue1685Test`.

dainiusfigoras:
> @dainiusfigoras : for fixtures, look at one of the testcases in the `/src/test/resources/fixtures`.. A testcase should include:
> 
> ```java
>     @Rule
>     public final GeneratedSource generatedSource = new GeneratedSource().addComparisonToFixtureFor(
>         UserMapper.class
>     );
> ```
> 
> See for instance `Issue1685Test`.

That's how, thanks.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
sjaakd: What happens when the target is also nested? Perhaps there are already testcases covering this, but could it be that when stacking down, you ultimately hit `"."` or will it stop at an earlier level.

So:

`@Mapping( target = "store.shop", source = "store.shop.customer.item" )`

Will be stacked down to:

`@Mapping( target = "shop", source = "shop.customer.item" )`

and I guess it stops here (I guess we have testcase for this).

dainiusfigoras: when target is not `"."` it should be handled as before in `handleDefinedMapping`.

seime:
@sjaakd Are there still changes needed before merging this functionality?

processor/src/test/java/org/mapstruct/ap/test/targetthis/CustomerEntityMapper1.java
sjaakd: > How do you mean? I was not aware I could currently specify `source="."`?

I checked. and you're right.. Seems I don't know all capabilities of MapStruct :smile: 

`org.mapstruct.ap.test.nestedsource.parameter.LetterMapper` changed: 

```java
    @InheritInverseConfiguration
    @Mapping(target = "font", source = "entity")
    LetterDto deNormalizeLetter(LetterEntity entity);
```
into
```java
    @InheritInverseConfiguration
    @Mapping(target = "font", source = ".")
    LetterDto deNormalizeLetter(LetterEntity entity);
```
and it gives exactly the same code..

dainiusfigoras: > Seems I don't know all capabilities of MapStruct 😄

it's more powerful then you can imagine :D ;)

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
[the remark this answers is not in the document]
sjaakd: I leave this one open for @filiphr to have a look at..

sjaakd:
@dainiusfigoras I'm done from the API perspective.. I've got the feeling the implementation is still a bit big and could be done smarter. But perhaps @filiphr has an opinion on that.. Thanks.

dainiusfigoras:
> @dainiusfigoras I'm done from the API perspective.. I've got the feeling the implementation is still a bit big and could be done smarter. But perhaps @filiphr has an opinion on that.. Thanks.

Thanks @sjaakd 
Is there something else I could do for this PR? Would be nice to merge it to master, so I can start using it from there.

facboy:
i'm also interested in seeing this functionality added.

dainiusfigoras:
> @dainiusfigoras I'm done from the API perspective.. I've got the feeling the implementation is still a bit big and could be done smarter. But perhaps @filiphr has an opinion on that.. Thanks.

Hi,

just wanted to remind about this @sjaakd @filiphr 
just so we can plan a bit with mapstruct versions.

lizhecao:
@dainiusfigoras target . is supported now ？

polarlights:
> @dainiusfigoras target . is supported now ？@sjaakd @filiphr

Not released yet.

sjaakd:
@dainiusfigoras .. I'll pick this one up and drive it home

daniel-shuy:
Thanks so much for the PR @dainiusfigoras, can't wait for this feature.

@sjaakd Any update on the release?

sjaakd:
@dainiusfigoras I worked this PR and will finally merged into the master. Can I add your name to the copyright.txt file (root of our project). I took the liberty of adding your name as author to the (test) classes you created.

sjaakd:
closed, its superseded by  PR #1833 and PR #1920

PR #1692, opened 2019-01-19
Mark some Builder providers as Experimental

filiphr:
Mark the Immutables BuilderProvider and AccessorNamingStrategy and the FreeBuilderAccessorNamingStrategy as Experimental.

Was not entirely sure whether we should do this or not.


filiphr:
Thanks for the review @sjaakd

PR #1693, opened 2019-01-19
#1681 Do not use finalizer method when using update method and result type and return type match

### LINKED ISSUE #1681: Non-void update methods try calling build() on the @MappingTarget when a Builder is available
Given two Lombok POJOs

    @SuperBuilder
    @Data
    @NoArgsConstructor
    @AllArgsConstructor
    public class Car {
        private String make;
        private int numberOfSeats;
        private CarType type;
    }

and

    @Data
    @SuperBuilder
    @NoArgsConstructor
    @AllArgsConstructor
    public class CarDto {
        private String make;
        private int seatCount;
        private String type;
    }

and a Mapper

    @Mapper
    public interface CarMapper {
        @Mapping(source = "numberOfSeats", target = "seatCount")
        CarDto carToCarDto(Car car);

        @Mapping(source = "numberOfSeats", target = "seatCount")
        CarDto updateFromCar(Car car, @MappingTarget CarDto dto);
    }

the generated code for the update method looks like this:

    @Override
    public CarDto updateFromCar(Car car, CarDto dto) {
        if ( car == null ) {
            return null;
        }

        dto.setSeatCount( car.getNumberOfSeats() );
        dto.setMake( car.getMake() );
        if ( car.getType() != null ) {
            dto.setType( car.getType().name() );
        }

        return dto.build();
    }

The setter calls are correct, but the last line should simply read `return dto;`. If the update method is declared as `void`, the generated code compiles fine.

Noticed in version 1.3.0.Beta2.

A complete example project is available [here](https://github.com/Chessray/mapstruct-examples/tree/master/bug-demo).

filiphr:
Fixes #1681

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -245,8 +245,11 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 ( (ForgedMethod) method ).addThrownTypes( factoryMethod.getThrownTypes() );
             }
 
-            MethodReference finalizeMethod = getFinalizerMethod(
-                resultType == null ? method.getResultType() : resultType );
+            MethodReference finalizeMethod = null;
+
+            if ( shouldCallFinalizerMethod( resultType == null ? method.getResultType() : resultType ) ) {
+                finalizeMethod = getFinalizerMethod();
+            }
 
             return new BeanMappingMethod(
                 method,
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -262,13 +262,20 @@ else if ( !method.isUpdateMethod() &&
...
 
         private MethodReference getFinalizerMethod(Type resultType) {
-            if ( method.getReturnType().isVoid() ||
-                resultType.getEffectiveType().isAssignableTo( resultType ) ) {
+            Type returnType = method.getReturnType();
+            if ( returnType.isVoid()) {
                 return null;
             }
-            BuilderType builderType = resultType.getBuilderType();
+            Type mappingType = method.isUpdateMethod() ? resultType : resultType.getEffectiveType();
```
sjaakd: I tried to parse this code.. but I remain having problems understanding what this code does.. IIUC if this is an update method we can have: 1) either a builder (in which we should call the finaliser ) or 2) not a builder (and then we should not call the finaliser ).. 

The method implies that it returns a finaliser.. However that is not what it does. Proposal: split into 2 methods that can be used on the call site.. 

Proposal:
1. `getFinalizerMethod` returns a finalizer when it can.
2. `useFinalizer` or `callFinalizer` (boolean) takes all scenarios into account.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1681/Issue1681Mapper.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper
+public interface Issue1681Mapper {
+
+    Issue1681Mapper INSTANCE = Mappers.getMapper( Issue1681Mapper.class );
+
+    Target update(@MappingTarget Target target, Source source);
```
sjaakd: To be complete: I reckon you should also test `Target update(@MappingTarget Target.Builder targetBuilder, Source source)`.. right?

sjaakd:
Can you also change the commit name? Logically would be: `#1681` don't call finaliser on return when @MappingTarget not a builder. `

processor/src/test/java/org/mapstruct/ap/test/bugs/_1681/Issue1681Mapper.java
filiphr: We already have a test case for that. See `SimpleBuilderMapper`. However, I can add one here as well

filiphr:
Thanks for the review @sjaakd. I have applied the comments you mentioned. Can you have another look?

I will change the commit when I do the squash and merge

PR #1696, opened 2019-01-22
1695 update gradle installation

### LINKED ISSUE #1695: update gradle installation
It would be greate to update the documentation about installing mapstruct with gradle. With the latest version of net.ltgt.apt the installing proccess should be like this:

  ```
  implementation "org.mapstruct:mapstruct:${mapstruct}"
  annotationProcessor "org.mapstruct:mapstruct-processor:${mapstruct}"
```

If you do not mind I can help with that

### PR DESCRIPTION
**Pull request by stianrincon**
based on the fact that the command "compile" was deprecated in Gradle I was changing that and updating the version of the net.ltgt.apt by the latest

Fixes #1695

### DIFF
--- a/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
+++ b/documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
@@ -142,7 +142,7 @@ dependencies {
 ...
 ----
 ====
-Gradle (3.3 and older)
+.Gradle (3.3 and older)
 ====
 [source, groovy, linenums]
 [subs="verbatim,attributes"]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc [hunk 1]
```diff
@@ -133,9 +133,36 @@ apply plugin: 'net.ltgt.apt-eclipse'
...
+    implementation "org.mapstruct:mapstruct:${mapstructVersion}"
+    annotationProcessor "org.mapstruct:mapstruct-processor:${mapstructVersion}"
 
+    // If you are using mapstruct in test code
+    testAnnotationProcessor "org.mapstruct:mapstruct-processor:${mapstructVersion}"
+}
+...
+----
+====
+Gradle (3.3 and older)
```
filiphr: I think that a dot (.) is missing here in order for the text to look like a caption. Perhaps name it `.Gradle configuration (3.3 and older)`?

stianrincon: Yes, you are right, sorry for that :$
And about the mapstruct-on-gradle, of course, I will help you with that as well

filiphr:
Just merged the PR. Thanks a lot for the update @stianrincon

Issue #1707, opened 2019-02-08
Uncompilable code is generated when mapping Stream to List and using AfterMapping on result.


SimSonic:
- [x] Is this an issue (and hence not a question)? 

When mapping method accepts Stream and results to Collection/List/Set and there is an void @AfterMapping method accepting the result of mapping method (Collection/List/Set), the generated code cannot be compiled:

Source code:
```java
@Value(staticConstructor = "of")
public static final class Source {
    String text;
    int number;
}

@Data
public static class Target {
    private String text;
    private int number;
    private boolean custom;
}

@Mapper
public abstract static class Converter {

    public abstract Set<Target> convert(Stream<Source> source);

    @AfterMapping
    public void addCustomValue(@MappingTarget Set<Target> targetList) {
        targetList.forEach(t -> t.custom = true);
    }
}
```

The code produced by mapstruct:
```java
@Generated(
    value = "org.mapstruct.ap.MappingProcessor",
    date = "2019-02-08T15:56:38+0700",
    comments = "version: 1.3.0.Beta2, compiler: javac, environment: Java 1.8.0_191 (Oracle Corporation)"
)
public class MapStructBugExample$ConverterImpl extends Converter {

    @Override
    public Set<Target> convert(Stream<Source> source) {
        if ( source == null ) {
            return null;
        }
        
        set.addAll( source.map( source1 -> sourceToTarget( source1 ) )
            .collect( Collectors.toCollection( HashSet<Target>::new ) )
        );

        addCustomValue( set );

        return set;
    }

    protected Target sourceToTarget(Source source) {
        if ( source == null ) {
            return null;
        }

        Target target = new Target();

        target.setText( source.getText() );
        target.setNumber( source.getNumber() );

        return target;
    }
}
```

In real life I would expect the generated code to be like this:
```java
    @Override
    public Set<Target> convert(Stream<Source> source) {
        if (source == null) {
            return null;
        }

        Set<Target> set = source.map(source1 -> sourceToTarget(source1))
                .collect(Collectors.toCollection(HashSet<Target>::new));

        addCustomValue(set);

        return set;
    }
```

sjaakd:
Ok.. looks like a bug..

sjaakd:
closed by #1707.. Will end up in 1.3

PR #1708, opened 2019-02-09
#1707 fix for not defining local variable in stream-iterable mapping

sjaakd:
@filiphr  The solution is adapting the template.. But a lot happens in the template. I first tried to update the `needVarDefine` itself  but then a lot of tests start to fail. It feels not entirely right to do it only on this spot, so please feel free to adapt / add.

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/StreamMappingMethod.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/StreamMappingMethod.ftl
@@ -52,7 +52,7 @@
     <#assign needVarDefine = beforeMappingReferencesWithMappingTarget?has_content && !existingInstanceMapping />
 
     <#if resultType.arrayType>
-        <#if !existingInstanceMapping && needVarDefine>
+        <#if needVarDefine>
             <#assign needVarDefine = false />
             <#-- We create a null array which later will be directly assigned from the stream-->
             ${resultElementType}[] ${resultName} = null;
@@ -67,7 +67,7 @@
         </#if>
     <#else>
         <#-- Streams are immutable so we can't update them -->
-        <#if !existingInstanceMapping && needVarDefine>
+        <#if needVarDefine>
             <#assign needVarDefine = false />
             <@iterableLocalVarDef/> ${resultName} = Stream.empty();
         </#if>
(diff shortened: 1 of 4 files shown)


processor/src/main/resources/org/mapstruct/ap/internal/model/StreamMappingMethod.ftl
```diff
@@ -52,7 +52,7 @@
     <#assign needVarDefine = beforeMappingReferencesWithMappingTarget?has_content && !existingInstanceMapping />
 
     <#if resultType.arrayType>
-        <#if !existingInstanceMapping && needVarDefine>
```
sjaakd: the `!existinInstanceMapping` is already present in `needVarDefine`, hence it could be removed.

processor/src/main/resources/org/mapstruct/ap/internal/model/StreamMappingMethod.ftl
```diff
@@ -60,7 +60,7 @@
     <#elseif resultType.iterableType>
         <#if existingInstanceMapping>
             ${resultName}.clear();
-        <#elseif needVarDefine>
+        <#elseif needVarDefine || afterMappingReferences?has_content>
```
sjaakd: I rather did have this as part of the `needVarDefine` definition (with an OR on `beforeMappingWithTargetReferences`).. but then stuff starts to breakdown..

filiphr:
I don't know what I was thinking when I saw doing that template. It is really complex. In any case I simplified something. Your assumption was right we should use the `afterMappingReferences`  in the `needVarDefine`. See my last commit. I didn't build locally, so let's see how this goes, I also added a test case with an array.

If the tests do not pass then we just go the way you did it.

processor/src/main/resources/org/mapstruct/ap/internal/model/StreamMappingMethod.ftl
```diff
@@ -180,5 +180,5 @@
     </@compress>
 </#macro>
 <#macro returnLocalVarDefOrUpdate>
-    <#if canReturnImmediatelly><#if returnType.name != "void">return </#if><#elseif !needVarDefine><@iterableLocalVarDef/> ${resultName} = <#else>${resultName} = </#if><#nested />
+    <#if canReturnImmediatelly><#if returnType.name != "void">return </#if><#elseif needVarDefine><@iterableLocalVarDef/> ${resultName} = <#else>${resultName} = </#if><#nested />
```
sjaakd: ah.. I tried the stuff above, but apparently missed this one.. In my case, all kinds of stuff started to fail in the iterable-iterable mappings...

sjaakd:
one question:should we add a fixture?

sjaakd:
> one question:should we add a fixture?

answered that myself :smile:

sjaakd:
@filiphr thanks for your help 👍

Issue #1698, opened 2019-01-25
Unreadable error due ambigious mapping


sjaakd:
It seems that MapStruct tries to map on inheritance (`Object`). 

Below is example of an error trace when there are a lot of suposed qualifying methods:

[...]

PR #1712, opened 2019-02-11
#1698 Skip "java.lang.Object" as intermediate result in 2 step mappings

sjaakd:
When mapping A to C via B,  A -> B -> C the 2 step mechanism tries B. Some builtin methods have `java.lang.Object` as intermediate result (user methods might have the same). Most notably: `T map(JAXBElement<T> source)` in which the result `T` is in essence an `java.lang.Object`.

This results in the strange error messages.


sjaakd:
Thanks for your feedback

PR #1715, opened 2019-02-13
#1435 add import to MapperConfig

### LINKED ISSUE #1435: Add imports to @MapperConfig
I have a scenario where I am using an expression in a MapperConfig mapping and I am having to add an imports for the Mapper annotation of the interfaces using it (see below for an example). Without adding the imports to the actual mapper the import is not there and the code doesn't compile. This is despite the fact the expression is on the config.

I'd really like to see a feature added that means we can have an imports directly in the MapperConfig declaration. If there were positive feedback on the request I'd be happy to implement it.

Cheers, Andy.

Example

```java
@MapperConfig(mappingInheritanceStrategy = MappingInheritanceStrategy.AUTO_INHERIT_FROM_CONFIG)
public interface Config {
    @Mapping(expression = "java( StringUtil.left(source.getText, 5) )", target = "text")
    OutObject map(final InObject source);
}

@Mapper(config = Config.class, imports = StringUtil.class)
public interface TextMapper {
    @Mapping(constant = "sometext", target = "outText")
    OutObject map(final InObject source);
}

@Mapper(config = Config.class, imports = StringUtil.class)
public interface TextMapper2 {
    OutObject map(final InObject source);
}

@Mapper(config = Config.class, imports = StringUtil.class)
public interface TextMapper3 {
    OutObject map(final InObject source);
}
```

Zomzog:
It adds import field to @MapperConfig annotation.

Fixes #1435

### DIFF
--- a/core/src/main/java/org/mapstruct/MapperConfig.java
+++ b/core/src/main/java/org/mapstruct/MapperConfig.java
@@ -45,6 +45,16 @@ public @interface MapperConfig {
     Class<?>[] uses() default { };
 
     /**
+     * Additional types for which an import statement is to be added to the generated mapper implementation class.
+     * This allows to refer to those types from within mapping expressions given via {@link Mapping#expression()},
+     * {@link Mapping#defaultExpression()} or using
+     * their simple name rather than their fully-qualified name.
+     *
+     * @return classes to add in the imports of the generated implementation.
+     */
+    Class<?>[] imports() default { };
+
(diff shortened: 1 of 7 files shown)


filiphr:
Thank you for the PR @Zomzog and congratulations on your first MapStruct contribution 😄.

>I am just wondering why this attribute was not already available...
Most probably people didn't need it and it was overlooked

Issue #1719, opened 2019-02-14
Strange error message when mapping with adder update method


sjaakd:
```java
@Mapper(  collectionMappingStrategy = CollectionMappingStrategy.ADDER_PREFERRED )
public interface XyzMapper {

    XyzMapper INSTANCE = Mappers.getMapper( XyzMapper.class );

    @InheritConfiguration
    @Mapping( target = "targetElements", source = "sourceElements" )
    void map(Source source, @MappingTarget Target target);

    void map(SourceElement sourceElement, @MappingTarget TargetElement targetElement);
}
``` 

```java
public class Source {

    private List<SourceElement> sourceElements;

    public List<SourceElement> getSourceElements() {
        return sourceElements;
    }

    public void setSourceElements(List<SourceElement> sourceElements) {
        this.sourceElements = sourceElements;
    }

}
```

```java

public class Target {

    private List<TargetElement> targetElements = new ArrayList<>(  );

    public List<TargetElement> getTargetElements() {
        return targetElements;
    }

    public void setTargetElements(List<TargetElement> targetElements) {
        this.targetElements = targetElements;
    }

    public TargetElement addTargetElement( TargetElement elemnt) {
        getTargetElements().add( elemnt );
        return elemnt;
    }
}
```

`XyzMapper.java:16 Can't map property "java.util.List<org.mapstruct.ap.test.bugs._xyz.SourceElement> sourceElements" to "org.mapstruct.ap.test.bugs._xyz.TargetElement targetElements". Consider to declare/implement a mapping method: "org.mapstruct.ap.test.bugs._xyz.TargetElement map(java.util.List<org.mapstruct.ap.test.bugs._xyz.SourceElement> value)`

filiphr:
Done in #1724

Issue #489, opened 2015-03-03
Follow up of #160, select update element / key-value mapping method from iterable / map mapping methods


sjaakd:
This is tricky.. Iterables / Maps are basically CRUD. Sets are problematic altogether. No idea yet how to tackle this in an automated fashion.

sjaakd:
I have thought a bit more about this, since I recently stumbled on this issue. What if we add 2 more properties to `@IterableMapping`? So `IterableMapping#sortComparator` and `IterableMapping#mergeComparator`. The latter one could be used for update methods. The first one could be applied to both create- and update methods? We could do something similar for keying `MapMappingMethods`

agudian:
Sounds interesting! What exactly would we need the sortComparator for?

sjaakd:
> Sounds interesting! What exactly would we need the sortComparator for?

I think there are 2 distinct actions: merging setA with setB and returning the result of that merge. The merging requires some kind of comparison, and the sorting independently from the marging on the result.

PR #1724, opened 2019-02-17
#1719 strange error message for selecting collection update methods

sjaakd:
Currently there's the strange error message as described in the PR..

For adder methods MapStuct cannot generate an update method. MapStruct would cannot know how to remove objects from the child-parent relation. It cannot even assume that the the collection can be cleared at forehand. Therefore the only sensible choice is for MapStruct to create a create method for the target elements.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -724,7 +724,13 @@ public class PropertyMapping extends ModelElement {
         }
 
         private Assignment forgeMapping(SourceRHS sourceRHS) {
-            Type sourceType = sourceRHS.getSourceType();
+            Type sourceType;
+            if ( targetWriteAccessorType == TargetWriteAccessorType.ADDER ) {
+                sourceType = sourceRHS.getSourceTypeForMatching();
+            }
+            else {
+                 sourceType = sourceRHS.getSourceType();
+            }
             if ( forgedNamedBased && !canGenerateAutoSubMappingBetween( sourceType, targetType ) ) {
                 return null;
             }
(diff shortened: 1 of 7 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -397,6 +408,34 @@ private void reportCannotCreateMapping() {
...
+        void reportCannotCreateCollectionOrMapUpdateMapping(Method method, AnnotationMirror posHint,
+                                                            String sourceErrorMessagePart,
+                                                            Type sourceType, Type targetType,
+                                                            String targetPropertyName ) {
+            ctx.getMessager().printMessage(
+                method.getExecutable(),
+                posHint,
+                targetWriteAccessorType == TargetWriteAccessorType.ADDER ?
+                    Message.PROPERTYMAPPING_CANNOT_CREATE_UPDATE_ADDER
+                    : Message.PROPERTYMAPPING_CANNOT_CREATE_UPDATE_MAP_OR_COLLECTION,
```
sjaakd: @filiphr .. I'm confused.. It seems that if you have an update mapping, MapStruct forges a regular collection mapping to call. However, if you selected "ADDER"  it will not do so and give the vague error message.

It feels inconsistent. I think what we should do is  **always** forge a regular-collection-(non-update)-mapping method also in the case of adder.  Reasoning.. we can't revert this behaviour. This means this PR will possibly end up in the examples how to do update methods.. So.. than we can also close #487 and #489 (the unit test shows how tricky this is, and I only covered one case)..

Another thought is to include a util library.. I've some ideas here.. Perhaps we should go forward.. Shall I make a demonstrator?

processor/src/test/java/org/mapstruct/ap/test/bugs/_1719/Issue1719Test.java
```diff
@@ -0,0 +1,54 @@
...
+        Target target = new Target();
+        TargetElement bob = new TargetElement( 1, "bob" );
+        target.addTargetElement( bob );
+        TargetElement louise = new TargetElement( 3, "louise" );
+        target.addTargetElement( louise );
+
+        Issue1719Mapper.INSTANCE.map( source, target );
+
+        assertThat( target.getTargetElements() ).hasSize( 3 );
+        assertThat( target.getTargetElements() ).containsOnly(
```
filiphr: This test doesn't use field by field comparator. I would change the test to use 
```
assertThat( target.getTargetElements() )
    .extracting( TargetElement:getId(), TargetElement::getName() )
    .containsOnly(
        tuple(1, "bob"),
        tuple(2, "alice"),
        tuple(3, "louis")
    );
```

sjaakd:
cool another GitHub feature.. Re-review..

PR #1737, opened 2019-02-25
extra-test-jaxbelem

sjaakd:
I'll adapt the commit name later to an issue (to be written, or refer to the old issue for which I provided extra unit test)... But have a look..

There are things I forgot in the 2 step mapping which assure this goes right. It seems that rather than using conversion / built in directly, we only use the source-return types and than still resolve the method. That makes sure it works.. I probably need to add some additional comment to make this clear.


filiphr:
We don't really have to add an issue. Looks OK to me (tests are failing due to missing classes).

Why not just put everything in the top `@WithClasses`, like it was before?

sjaakd:
> Why not just put everything in the top `@WithClasses`, like it was before?

Hard do debug.. All mappers come by + all mapping methods in each test case.

filiphr:
That is completely valid, I agree with you

sjaakd:
thanks for reviewing..

sjaakd:
now to convince travis :smile:

Issue #1714, opened 2019-02-12
Qualifiers work dubiously in 2 step mappings


sjaakd:
See also [stack overflow](https://stackoverflow.com/questions/54623673/mapstruct-ambiguous-mapping-methods-found-for-mapping-property?noredirect=1#comment96073389_54623673) 

The issue here is that if MapStruct cannot make a match immediately, it tries to do a 2 step approach (mapping - type conversion). 

MapStruct seems to forget the qualifiers completely and that scenario and selects also methods with an entirely different qualifier. That's strange and unexpected.

See also [gist](https://gist.github.com/sjaakd/b925f2ecb6c5a8116debb97f3fb31c1d)

filiphr:
Closing since we have a commit for this. Feel free to reopen @sjaakd if you think we should

PR #1739, opened 2019-02-27
#1714 Qualifiers should not qualfiy when no qualfier is found

sjaakd:


sjaakd:
please test this

sjaakd:
@filiphr .. did some cleaning up.. In the end I did not include JSR305 (yet). Perhaps in another refactoring PR.

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/QualifierTest.java
```diff
@@ -96,14 +96,13 @@ public void shouldMatchClassAndMethod() {
...
-                        line = 28,
-                        messageRegExp = "Ambiguous mapping methods found for mapping property "
-                                + "\"java.lang.String title\" to java.lang.String.*" )
-            }
+        value = CompilationResult.FAILED,
+        diagnostics = {
+            @Diagnostic(type = ErroneousMapper.class,
+                kind = Kind.ERROR,
+                line = 28,
+                messageRegExp = "Can't map property \"java.lang.String title\" to \"java.lang.String title\".*")
```
filiphr: What is the full error message here? Asking since I think that this error message will be confusing to a user.

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/iterable/TopologyMapper.java
```diff
@@ -29,6 +32,14 @@
     @Mapping( target = "topologyFeatures", qualifiedBy = Cities.class )
     public abstract TopologyEntity mapTopologyAsCity(TopologyDto dto);
 
+    @Rivers
+    @IterableMapping( qualifiedBy = Rivers.class )
+    public abstract List<TopologyFeatureEntity> mapTopologiesAsRiver(List<TopologyFeatureDto> in);
```
filiphr: What was the reason for adding this 2 abstract methods? Are these methods failing without your changes?

sjaakd: the qualifier acts both on the element as on the `@BeforeMapping` now qualifiers are working as they should, MapStruct cannot find a qualifying method anymore for the element.

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/QualifierTest.java
sjaakd: I'll add some more detail to the message so its more obvious.

sjaakd:
@filiphr I'll do some rework on the message asap. Do you think its ok to merge when done? You already observed qualifiers are now more strict. If it can't find a method annotated with the proper qualifier it quits. (Which I think is right). So, we should probably warn in the release notes, since this might affect existing code.

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/iterable/TopologyMapper.java
filiphr: So this means that this is a backwards incompatible change? In this case I am not quite sure if it is OK to break it. Although you can say that if you are adding the qualifiedBy we should report thay we didn't find that

filiphr:
Yes I think that it is ok to merge if we output better error messages. It would be good to see the error messages on all places where you needed to add the new methods and if those aren't good we should improve them. Perhaps add test cases for those error messages?

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/QualifierTest.java
filiphr: Thanks

sjaakd:
There are more nasty side effects of the current solution.. : wrong method selection, generating stuff while but expected, no errors, not working 2 step mapping..and the list goes on

Verstuurd vanaf mijn iPhone

filiphr:
When you say current solution, do you mean this PR or master?

sjaakd:
The master 😃

Verstuurd vanaf mijn iPhone

filiphr:
Then I think it is fine as the change should actually be "more" correct. I would expect that I would get an error if I say that I want the mapping to happen with some qualification, but that doesn't exist

sjaakd:
> Then I think it is fine as the change should actually be "more" correct. I would expect that I would get an error if I say that I want the mapping to happen with some qualification, but that doesn't exist

I'll have a look and see how difficult this is to realise..

processor/src/test/java/org/mapstruct/ap/test/selection/qualifier/iterable/TopologyMapper.java
sjaakd: as discussed elsewhere..

filiphr:
>I'll have a look and see how difficult this is to realise..

That would be great 😄

sjaakd:
I'll fix this one in a separate PR.. Just want to fix the bug first. 

You'll find this check everywhere: `if ( assignment == null && !criteria.hasQualfiers() ) ` .. My thought is to rewrite this to a lamda expression and handle it in the resolver.  But first fix this one.. :smile:

sjaakd:
@filiphr .. thanks for the review..

Issue #1140, opened 2017-03-13
Warning when the target has no target properties


filiphr:
As discussed in #1137. We should find another way and another message to display a warning when the target has no target properties (usually that means that something is strange in the mapping).

augustojaba:
I will get this one.

augustojaba:
Can I get this one? Could you please give more context? This is my first issue on this project.
Have a good one!

zyberzebra:
@filiphr Could you provide an example and what behavior you expect? 
`if ( method instanceof ForgedMethod && targetProperties.isEmpty() ) {`
I guess you want an alternative approach to that check in `org.mapstruct.ap.internal.model.BeanMappingMethod.Builder#reportErrorForUnmappedTargetPropertiesIfRequired` right?

filiphr:
It's been a long time since I've looked into this.

I believe that it should be slightly sooner. Perhaps at

```java
this.targetProperties = new LinkedHashSet<>( accessors.keySet() );
```

in the `BeanMappingMethod`. Provide some kind of a warning at that point.

I'm not sure if changing `reportErrorForUnmappedTargetPropertiesIfRequired` is the right approach as that place is giving out a compile error right now.

zyberzebra:
Should this also lead to the warning:
```
public class FilledJar {
    private String label;
    private double weight;
    private Object content;

   // setter & getters ...
```

```java
public class AirplaneWithNoAccessors {
    private int flightNumber;
    private String airplaneName;
}
```

```java
public interface JarToAirplaneMapper {

    JarToAirplaneMapper INSTANCE = Mappers.getMapper( JarToAirplaneMapper.class );

    AirplaneWithNoAccessors mapToAirplaneWithNoAccessors(FilledJar jar);
}
``` 

```java

// Errormessage?: Target \"Airplane\" has no target properties, targeted by this mapping method: \"Airplane mapToAirplaneWithNoAccessors(FilledJar filledJar)\

    void targetHasNoAccessibleProperties() {
        AirplaneWithNoAccessors airplane = JarToAirplaneMapper.INSTANCE.mapToAirplaneWithNoAccessors( new FilledJar() );
        assertThat( airplane ).hasAllNullFieldsOrProperties();
    }

```

PR #1741, opened 2019-03-03
#37 Rudimentary logging in mapstruct

sjaakd:
You can activate MapStruct in verbose mode by Maven (need to update documentation)..

```xml
    <build>
        <plugins>
            <plugin>
                <version>3.6.2</version>
                <configuration>
                    <annotationProcessorPaths>
                        <path>
                            <version>${org.mapstruct.version}</version>
                        </path>
                    </annotationProcessorPaths>
                    <showWarnings>true</showWarnings>
                </configuration>
            </plugin>
        </plugins>
    </build>

```

This would be a typical output

```
[INFO] MapStruct: using default builder pattern.
[INFO]  MapStruct: processing: org.mapstruct.jpa.SourceTargetMapper
[INFO]  MapStruct: applying MapperConfiguration
[INFO] -- MapStruct: mapping: s.getName() to: setName( . )
[INFO] -- MapStruct: selecting property mapping: s.getName()
[INFO] -- MapStruct: mapping: s.getChildren() to: setChildren( . )
[INFO] -- MapStruct: selecting element mapping: ChildEntity #toEntity(childDto)
[INFO] -- MapStruct: creating property mapping: List #childDtoListToChildEntityList(s.getChildren())
[INFO] - MapStruct: creating bean mapping method implementation for: org.mapstruct.jpa.ParentEntity toEntity(org.mapstruct.jpa.ParentDto s, @Context org.mapstruct.jpa.JpaContext ctx)
[INFO] -- MapStruct: mapping: s.getName() to: setName( . )
[INFO] -- MapStruct: selecting property mapping: s.getName()
[INFO] - MapStruct: creating bean mapping method implementation for: org.mapstruct.jpa.ChildEntity toEntity(org.mapstruct.jpa.ChildDto s, @Context org.mapstruct.jpa.JpaContext ctx)
[INFO] 
```


lorenzleutgeb:
This pull request resolves #37.

I would like to use this, and in addition ask for documentation on how logging can be enabled using Gradle instead of Maven.

processor/src/main/java/org/mapstruct/ap/internal/util/FormattingMessager.java
```diff
@@ -67,4 +67,10 @@ void printMessage(Element e,
                       AnnotationValue v,
                       Message msg,
                       Object... args);
+
+    /**
+     * Just log as plain note
+     * @param log
+     */
+    void log(int level, String log);
```
filiphr: Perhaps name it `verbose`?

sjaakd:
> I am not that familiar with the 'NOTE`scope of the javac`Messager`. Did you think about putting this under a MapStruct specific compiler option? This way people will need to enable it on the MapStruct side. It is a bit weird needing to show warnings to display verbose information.

I thought about that.. But then you have to write to system.out.. I personally don't like that.. We should use the tool at hand (messager in annotation processor).

processor/src/main/java/org/mapstruct/ap/internal/util/AnnotationProcessorContext.java
```diff
@@ -64,19 +69,32 @@ private void initialize() {
...
         }
         this.accessorNamingStrategy = Services.get( AccessorNamingStrategy.class, defaultAccessorNamingStrategy );
         this.accessorNamingStrategy.init( this );
+        if ( this.accessorNamingStrategy != defaultAccessorNamingStrategy ) {
+            messager.printMessage( Diagnostic.Kind.NOTE,
+                "MapStruct: Overriding found accessor naming strategy with user SPI defined strategy" );
+        }
         this.builderProvider = Services.get( BuilderProvider.class, defaultBuilderProvider );
         this.builderProvider.init( this );
+        if ( this.accessorNamingStrategy != defaultAccessorNamingStrategy ) {
```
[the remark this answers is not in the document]
sjaakd: 😄 .. will fix that..

processor/src/main/java/org/mapstruct/ap/internal/util/AnnotationProcessorContext.java
```diff
@@ -64,19 +69,32 @@ private void initialize() {
...
         else {
             defaultAccessorNamingStrategy = new DefaultAccessorNamingStrategy();
             defaultBuilderProvider = new DefaultBuilderProvider();
+            messager.printMessage( Diagnostic.Kind.NOTE, "MapStruct: using default builder pattern." );
         }
         this.accessorNamingStrategy = Services.get( AccessorNamingStrategy.class, defaultAccessorNamingStrategy );
         this.accessorNamingStrategy.init( this );
+        if ( this.accessorNamingStrategy != defaultAccessorNamingStrategy ) {
+            messager.printMessage( Diagnostic.Kind.NOTE,
+                "MapStruct: Overriding found accessor naming strategy with user SPI defined strategy" );
```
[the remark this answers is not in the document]
sjaakd: good idea.. will look into that..

processor/src/main/java/org/mapstruct/ap/internal/util/AnnotationProcessorContext.java
```diff
@@ -64,19 +69,32 @@ private void initialize() {
...
+        }
         this.builderProvider = Services.get( BuilderProvider.class, defaultBuilderProvider );
         this.builderProvider.init( this );
+        if ( this.accessorNamingStrategy != defaultAccessorNamingStrategy ) {
+            messager.printMessage( Diagnostic.Kind.NOTE,
```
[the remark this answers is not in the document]
sjaakd: good idea.. will look into that..

processor/src/main/java/org/mapstruct/ap/internal/util/FormattingMessager.java
sjaakd: > Some other thing, all other places where we do logging with use our [`Message`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java). Why not do it the same? We can even keep everything as is and only output `NOTE` when the flag is enabled, otherwise it is a noop. Maybe even outputting the annotation value and element could have more output.

I can try to fit everything in a standard terminology (Message will enforce that).. I tried to use terms as create, select etc..

filiphr: The terms were really good. Forgot to mention that 😀

filiphr:
> thought about that.. But then you have to write to system.out.. I personally don't like that.. We should use the tool at hand (messager in annotation processor).

I would still print on note via the messager, just put it under an option. This way people opt in into using it and they won't need to use maven compiler specifics. Why would we need to put it on sys out?

sjaakd:
> I would still print on note via the messager, just put it under an option. This way people opt in into using it and they won't need to use maven compiler specifics. Why would we need to put it on sys out?


https://stackoverflow.com/questions/13151678/messages-with-level-diagnostic-kind-note-and-others-on-annotation-processors

processor/src/main/java/org/mapstruct/ap/internal/util/FormattingMessager.java
sjaakd: > The terms were really good. Forgot to mention that 😀

Thanks...

sjaakd:
> I would like to use this, and in addition ask for documentation on how logging can be enabled using Gradle instead of Maven.

I just ran my solution with Gradle (from our examples)  it seems to log `note` out of the box.

sjaakd:
@filiphr I did most of the work you requested, except my remark above. You need this maven option to show NOTE diagnostics.. Is more of a plugin issue than it has to do with Apt.. Also, I don't like to have 2 settings to let something work..

processor/src/main/java/org/mapstruct/ap/internal/util/accessor/ExecutableElementAccessor.java
```diff
@@ -28,4 +28,9 @@ public TypeMirror getAccessedType() {
     public ExecutableElement getExecutable() {
         return element;
     }
+
+    @Override
+    public String toString() {
+        return element.getSimpleName().toString() + "( . )";
```
filiphr: Why not just use the `element.toString()` instead? If we want to use something specific in the printing of the messager we should pass that in the arguments, and not do it in the `toString`. For debugging it would be much easier to just have `element.toString()` here.

processor/src/main/java/org/mapstruct/ap/internal/util/accessor/VariableElementAccessor.java
```diff
@@ -29,4 +29,9 @@ public TypeMirror getAccessedType() {
     public ExecutableElement getExecutable() {
         return null;
     }
+
+    @Override
+    public String toString() {
+        return element.getSimpleName().toString();
```
filiphr: Same comment as in the `ExecutableElementAccessor`

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
```diff
@@ -231,4 +231,8 @@ public String getCollectionGetterName(String property) {
             + "future versions." );
     }
 
+    @Override
+    public String toString() {
+        return getClass().getCanonicalName();
```
filiphr: Why do we need to use `getClass().getCanonicalName()`?  We could do that on the caller side, not sure if it is a good `toString` method

processor/src/main/java/org/mapstruct/ap/spi/DefaultBuilderProvider.java
```diff
@@ -280,4 +280,10 @@ protected boolean isBuildMethod(ExecutableElement buildMethod, TypeElement typeE
     protected boolean shouldIgnore(TypeElement typeElement) {
         return typeElement == null || JAVA_JAVAX_PACKAGE.matcher( typeElement.getQualifiedName() ).matches();
     }
+
+    @Override
+    public String toString() {
+        return getClass().getCanonicalName();
```
filiphr: Same comment as in the `DefaultAccessorNamingStrategy`

processor/src/main/java/org/mapstruct/ap/spi/FreeBuilderAccessorNamingStrategy.java
```diff
@@ -35,4 +35,10 @@ protected boolean isFluentSetter(ExecutableElement method) {
         // with set
         return false;
     }
+
+    @Override
+    public String toString() {
+        return getClass().getCanonicalName();
```
filiphr: Same comment as in the `DefaultAccessorNamingStrategy`

processor/src/main/java/org/mapstruct/ap/spi/ImmutablesAccessorNamingStrategy.java
```diff
@@ -23,4 +23,10 @@
     protected boolean isFluentSetter(ExecutableElement method) {
         return super.isFluentSetter( method ) && !method.getSimpleName().toString().equals( "from" );
     }
+
+    @Override
+    public String toString() {
+        return getClass().getCanonicalName();
```
filiphr: Same comment as in the `DefaultAccessorNamingStrategy`

processor/src/main/java/org/mapstruct/ap/spi/ImmutablesBuilderProvider.java
```diff
@@ -83,4 +83,10 @@ protected TypeElement asImmutableElement(TypeElement typeElement) {
         builderQualifiedName.append( "Immutable" ).append( typeElement.getSimpleName() );
         return elementUtils.getTypeElement( builderQualifiedName );
     }
+
+    @Override
+    public String toString() {
+        return getClass().getCanonicalName();
```
filiphr: Same comment as in the `DefaultAccessorNamingStrategy`

processor/src/main/java/org/mapstruct/ap/spi/NoOpBuilderProvider.java
```diff
@@ -23,5 +23,11 @@
     public BuilderInfo findBuilderInfo(TypeMirror type) {
         return null;
     }
+
+    @Override
+    public String toString() {
+        return getClass().getCanonicalName();
```
filiphr: Same comment as in the `DefaultAccessorNamingStrategy`

processor/src/main/java/org/mapstruct/ap/internal/model/MethodReference.java
```diff
@@ -348,4 +350,14 @@ public static MethodReference forMethodCall(String methodName) {
         return new MethodReference( methodName, null, false );
     }
 
+    @Override
+    public String toString() {
```
filiphr: I am not sure if the `toString` is always correct. Is `returnType` always non-null? Looking at the code it is not always like that.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -265,6 +271,10 @@ suppressGeneratorTimestamp`
 |If set to `true`, the creation of a time stamp in the `@Generated` annotation in the generated mapper classes is suppressed.
 |`false`
 
+|`mapstruct.verbose`
+|If set to `true`, MapStruct in which MapStruct logs its major decisions. Note: also `showWarnings` need to be added due to a problem in the maven-compiler-plugin configuration.
```
filiphr: I think we need to say that `showWarnings` is for use within maven.

processor/src/main/java/org/mapstruct/ap/internal/model/MethodReference.java
sjaakd: good catch!

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: added additional text

sjaakd:
@filiphr . Fixed your comments.. 

> Changes look much better thank you.
> 
Thanks

> I've left some more comments in the PR, I am not 100% sure all new `toString` are correct and whether we should use them like that. See my comments.
> 
I changed them.

> I also noticed some double usage of `verbose`. Why not just use the `printMessage` and if the `Message#getDiagnosticKind()` is `NOTE` we check for verbose and if it is we print it out if it isn't we don't. This way we don't need to use the verbose on different locations and we don't need the `note(int, Message, Object...)` on the `FormattingMessager`.
>
The FormattingMessager cannot is not available at the other spots.. Its part of the processor context.. It will require some more refactoring (which I did not want to do right now)..
 
> Why not just add the `-` and `Mapstruct:` into the new messages?
> 

What do you mean?

> Btw. I think that we can test this as well, the same way we are testing the error and warning prints.

I'll take a better look this way.. Perhaps we should do some more testing..

filiphr:
Thanks.

>The `FormattingMessager` cannot is not available at the other spots.. Its part of the processor context.. It will require some more refactoring (which I did not want to do right now).

You are completely right, I missed that.


>>Why not just add the - and Mapstruct: into the new messages?
>
>What do you mean?

For example for `Message#ITERABLEMAPPING_CREATE_ELEMENT_NOTE` instead of doing

```
ctx.getMessager().note( 2, Message.ITERABLEMAPPING_CREATE_ELEMENT_NOTE, assignment );
```

We can make `Message#ITERABLEMAPPING_CREATE_ELEMENT_NOTE` have the text: "-- MapStruct: creating element mapping: %s.". And we can just do:

```
ctx.getMessager().printMessage(Message.ITERABLEMAPPING_CREATE_ELEMENT_NOTE, assignment );
```

sjaakd:
>We can make Message#ITERABLEMAPPING_CREATE_ELEMENT_NOTE have the text: "-- MapStruct: creating element mapping: %s.". And we can just do

The idea is to take nesting into account in a next step.

filiphr:
>The idea is to take nesting into account in a next step.

Not sure that I am following, but I feel confident that you've thought it through.

Btw, just thought about https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/MappingProcessor.java#L224. Perhaps we can add a line saying that a mapper has been deferred and the reason for the deferral.

sjaakd:
> Not sure that I am following, but I feel confident that you've thought it through.
> 
Well.. If we have nested properties, we could add more dashes `-` by just specifying how many we'd like. So it will provide a visual indentation hint what MapStruct is doing on which nesting level.. That's actually not so difficult to implement (I believe we keep history also somewhere)..

> Btw, just thought about https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/MappingProcessor.java#L224. Perhaps we can add a line saying that a mapper has been deferred and the reason for the deferral.

Ok... I can add that (I think its mostly for if we are in an annotation processor run with more annotation processors right (e.g. Lombok).  So it would be for debugging purposes?

I'm a bit careful with adding too much information.. The logging must be comprehensive

sjaakd:
@filiphr .. seems that the plexus compiler does not make a distinction between WARNING and NOTE.. in fact.. NOTE is never set.. What a mess.

sjaakd:
> > Btw, just thought about https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/MappingProcessor.java#L224. Perhaps we can add a line saying that a mapper has been deferred and the reason for the deferral.
> 
> Ok... I can add that (I think its mostly for if we are in an annotation processor run with more annotation processors right (e.g. Lombok). So it would be for debugging purposes?

I did not add this yet..  Lets keep that in a separate PR.

processor/src/test/java/org/mapstruct/ap/testutil/compilation/annotation/ExpectedNote.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * The notes in the order they are expected
+     */
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.METHOD)
+    public @interface ExpectedNotes {
+
+        /**
+         *  Regexp voor the value.
```
chris922: In Germany we often say "denglisch" in case someone mixes English and German ("deutsch"). But maybe we can also say this for mixing Dutch and English? 🤔 

```suggestion
         *  Regexp for the expected note.
```

processor/src/test/java/org/mapstruct/ap/testutil/compilation/model/CompilationOutcomeDescriptor.java
```diff
@@ -25,21 +28,37 @@
  */
 public class CompilationOutcomeDescriptor {
 
+    private static final String LINE_SEPARATOR = System.getProperty( "line.separator" );
```
chris922:
```suggestion
    private static final String LINE_SEPARATOR = System.lineSeparator();
```

processor/src/test/java/org/mapstruct/ap/testutil/compilation/annotation/ExpectedNote.java
sjaakd: ha! well spotted @chris922 .. its the same here..

processor/src/test/java/org/mapstruct/ap/testutil/compilation/model/CompilationOutcomeDescriptor.java
sjaakd: hey.. did not know that was there.. I'll fix that one

sjaakd:
@chris922 : thanks for reviewing

processor/src/test/java/org/mapstruct/ap/test/verbose/VerboseTest.java
```diff
@@ -0,0 +1,113 @@
...
+    @WithClasses(CreateBeanMapping.class)
+    @ExpectedNote("^- MapStruct: creating bean mapping method implementation for.*$")
+    @ExpectedNote("^-- MapStruct: creating property mapping.*$")
+    public void testCreateBeanMapping() {
+    }
+
+    @Test
+    @ProcessorOption(name = "mapstruct.verbose", value = "true")
+    @WithClasses(SelectBeanMapping.class)
+    @ExpectedNote("^- MapStruct: creating bean mapping method implementation for.*$")
```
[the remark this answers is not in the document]
sjaakd: It's a balance between test readability and completeness.. so I want to know if I got the correct message in the correct order.. but the details are not so relevant and can be reviewed as well.. of course I checked then buy they span 4 lines sometimes.. all with the dots escaped

processor/src/test/java/org/mapstruct/ap/test/verbose/VerboseTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.WithServiceImplementation;
+import org.mapstruct.ap.testutil.compilation.annotation.ExpectedNote;
+import org.mapstruct.ap.testutil.compilation.annotation.ProcessorOption;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+import org.mapstruct.ap.testutil.runner.Compiler;
+import org.mapstruct.ap.testutil.runner.WithSingleCompiler;
+
+@IssueKey("37")
+@WithSingleCompiler(Compiler.JDK)
```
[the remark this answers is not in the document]
sjaakd: I'll have a look.. also want to add a test for the latest extra log I added

processor/src/test/java/org/mapstruct/ap/test/verbose/VerboseTest.java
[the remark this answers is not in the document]
filiphr: Makes sense.

processor/src/test/java/org/mapstruct/ap/test/verbose/VerboseTest.java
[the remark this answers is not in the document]
sjaakd: done.. I see it has no retention on class, so I chose to add it to the individual methods..

processor/src/test/java/org/mapstruct/ap/test/verbose/CreateMapMapping.java
```diff
@@ -0,0 +1,69 @@
...
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface CreateMapMapping {
+
+    CreateMapMapping INSTANCE = Mappers.getMapper( CreateMapMapping.class );
+
+    Map<TargetKey, TargetValue> map(Map<SourceKey, SourceValue> source);
+
+    // empty beans fail.. TODO check
```
filiphr: I saw some commits coming for this is this done now?

processor/src/test/java/org/mapstruct/ap/test/verbose/VerboseTest.java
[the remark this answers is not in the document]
filiphr: Yes I only needed it for a method. You could add it on a class if you want to

processor/src/test/java/org/mapstruct/ap/test/verbose/CreateMapMapping.java
sjaakd: I'll handle this in a separate one.. The error message is just very unclear..  Its correct that they fail, but  as a user: I wouldn't have a clue why..

sjaakd:
@filiphr .. thanks for reviewing..

processor/src/test/java/org/mapstruct/ap/test/verbose/CreateMapMapping.java
filiphr: Makes sense. Maybe related to https://github.com/mapstruct/mapstruct/issues/1140?

Issue #1075, opened 2017-02-15
Please add a feature to map from java.util.map to an object


twothe:
Refering to [Stack-Overflow](http://stackoverflow.com/questions/42226847/mapstruct-mapping-from-java-util-map-to-bean)

The situation is: I have a REST service that must be called with a huge amount of parameters (as by customer request) and those should be stored in a database. The easiest way to do that would be to take the parameter map and map that to an entity object. As it is about the same type of code to map from an object to and object or from a map to an object I assume it should not be too difficult to allow mapping like this:

    public MyEntity mapToEntity(final Map<String, String> parameters);

Essentially that should turn into code like this:

    myEntity.setName( parameters.get("name") );
    myEntity.setReferrer( stringToReferrer( parameters.get("referrer") ) );
    //...

It would be nice to have other variable types supported as well, like Map<String, Object>, but I am not sure at which point it becomes too complicated.

Raild:
I've tried to do it in the following way:
```java
@Mapping(expression = "java(parameters.get(\"name\"))", target = "name")
public MyEntity mapToEntity(final Map<String, String> parameters);
```
But is it possible to convert value if the MyEntity.name type is boolean (without calling Boolean.parse(...) manually)?

ft0907:
I have a similar requirement now, and I need to directly map the map to the bean. I don't know if there is any information on the advancement of this function or if there are other solutions?

markbanierink:
We have this requirement too. However, for us it is important that this Map to Object mapper would behave in a PATCH manner, updating only the object fields that are provided in the map and ignoring the rest. Is that also the intention of this feature?

filiphr:
@markbanierink the idea for this feature is to support all the other functionalities that MapStruct has. So if you are using `@MappingTarget` with `NullValuePropertyMappingStrategy#IGNORE` it would behave as you need.

We have not yet discussed the length of the implementation, but I envision that it would generate something like:

```
public interface CustomerMapper {

    public void updateCustomer(@MappingTarget Customer customer, Map<String, Object> parameters) {

        if (parameters.contains("name")) {
            target.setName(parameters.get("name"));
        }
    }

}
```

Of course with proper conversions.

braeluca:
Will it be possible to map a map with a nested map?
e.g. 
```
Map<String,Object> map = new HashMap<>();
Map<String,Object> nestedMap = new HashMap<>();
        
nestedMap.put("name","Name");
[...]

Issue #1731, opened 2019-02-21
Support of java.sql.ResultSet and spring-jdbc RowMapper


Lysergid:
What about having support of JDBC call results in mappers.

Example of how it may look like:

@Mapper
public interface CarMapper {

    CarMapper INSTANCE = Mappers.getMapper( CarMapper.class );
 
    @JdbcMapping(column = "NUMBER_OF_SEATS", target = "seatCount")
    CarEntity carRsToCarEntity(ResultSet rs)
}

It's not necessary to have new annotation, but I think it will improve readability.

CarMapper might also extend spring-jdbc RowMapper so that mapping implementation will be in CarMapperImpl.mapRow()

I just want to know if this can be in scope of MapStruct or no, thus I don't go into details.

sjaakd:
I'm not sure how the other MapStruct members feel about this one.. But I think its too much geared towards one purpose. @filiphr : WDYT?

filiphr:
I think that the purpose is not that bad honestly. `java.sql.ResultSet` is part of Java itself so it can be optionally used (due to modules).

The option with the spring-jdbc `RowMapper` will come for free, with the only exception that it would need to override the `mapRow` with

```
@Mapper
public interface CarMapper extends RowMapper {

    CarMapper INSTANCE = Mappers.getMapper( CarMapper.class );
 
    @JdbcMapping(column = "NUMBER_OF_SEATS", target = "seatCount")
    CarEntity carRsToCarEntity(ResultSet rs)

    @Override
    default T mapRow(ResultSet rs, int rowNum) {
        return carRsToCarEntity(rs);
    }
}
```

I think that this is really similar to https://github.com/mapstruct/mapstruct/issues/1075. Similar in the sense that it allows using the target field name in order to get a value for it. 

I would not add a new custom `@JdbcMapping` in our API as we can just reuse:

```
@Mapping(target = "seatCount", source = "NUMBER_OF_SEATS")
```

@sjaakd I think that if someone from the community wants to work on this we should help them out.

sjaakd:
> @sjaakd I think that if someone from the community wants to work on this we should help them out.

[...]

PR #1744, opened 2019-03-09
#1075 initial steps mapping from Map to object

sjaakd:
Outline:
* Make a copy `BeanMappingMethod` including backing template: call it: `BeanMapMappingMethod`
* Select this method on `java.util.Map` to `Bean` and vice versa.
* Start from that point onwards to refactor the method.

Bean to Map, Map to Bean should be a first class citizen (initially we thought about wrapping the bean and using an annotation). The PR below can be scrapped. 

Some findings:
Note 1: Can't do nesting on source side. From a source side perspective we have no clue which types are in a Map
Note 2: Can't handle a map as a parameter (direct mapping)
Note 3: We should not allow Map to Map

TODO
- [x]: refactoring source reference
- []: copy and adapt `BeanMappingMethod`
- []: a lot of unit testing (multiple source parameters, error conditions).
- []: How to handle a special PropertyEntry? We can only determine the Type when we have the target in our hand
- []: error handling

### DIFF
--- a/core/src/main/java/org/mapstruct/AsBean.java
+++ b/core/src/main/java/org/mapstruct/AsBean.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Wraps the source / target object in a bean
+ *
+ * @author Sjaak Derksen
+ */
+@Target( ElementType.PARAMETER )
+@Retention(RetentionPolicy.CLASS)
+public @interface AsBean {
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/gem/GemGenerator.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/gem/GemGenerator.java
@@ -9,6 +9,7 @@ import javax.xml.bind.annotation.XmlElementDecl;
 import javax.xml.bind.annotation.XmlElementRef;
 
 import org.mapstruct.AfterMapping;
+import org.mapstruct.AsBean;
 import org.mapstruct.BeanMapping;
 import org.mapstruct.BeforeMapping;
 import org.mapstruct.Builder;
@@ -61,6 +62,7 @@ import org.mapstruct.tools.gem.GemDefinition;
 @GemDefinition(ValueMappings.class)
 @GemDefinition(Context.class)
 @GemDefinition(Builder.class)
+@GemDefinition(AsBean.class)
 
 @GemDefinition(MappingControl.class)
 @GemDefinition(MappingControls.class)
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanToMap.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanToMap.java
@@ -1,0 +1,49 @@
+package org.mapstruct.ap.internal.model;
+
+import java.util.List;
+
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.util.Strings;
+
+public class BeanToMap {
+
+    private final List<MapEntry> mapEntries;
+    private final String className;
+
+
+    public BeanToMap(String className, List<MapEntry> types) {
+        this.className = className;
+        this.mapEntries = types;
+    }
+
+    public String getClassName() {
+        return className;
+    }
+
+    public List<MapEntry> getMapEntries() {
+        return mapEntries;
+    }
+
+    public String capitalize(String in) {
+        return Strings.capitalize( in );
+    }
+
+    public static class MapEntry {
+
+        private final String name;
+        private final Type type;
+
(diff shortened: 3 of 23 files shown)


sjaakd:
and still yet doing some more refactoring on sourcref..

sjaakd:
@filiphr .. I think that the `refactoring source reference` has value on its own right.. Shall I take that one in a separate PR?

filiphr:
>@filiphr .. I think that the refactoring source reference has value on its own right.. Shall I take that one in a separate PR?

Most definitely, always up for a little bit of refactoring 😀

sjaakd:
ok.. here we go then.. :smile:

chris922:
Good to hear that you are started with this feature! I would love to have it in place 🤩

I've got some questions regarding your notes/findings:

**Note 1:**

What is with this?

```java
@Mapping( target = "childFirstName", source = "child.firstName" )
Target map(Map<String, Source> source);
```

Wouldn't this be possible? I would expect that this will call `source.get("child").getFirstName()`

**Note 2:**

What do you mean with this point?

```java
class Target {
  Map<String, String> foo;
  // ... other properties
}

Target map(Map<String, Source> source, Map<String, String> foo);
```

`foo` can't be mapped in such a case to `target.foo`? If yes, why?

**Note 3:**

As soon as we also support mapping to a `Map`, why shouldn't we allow this? But as a first step mapping _from_ a `Map` would already be great. Mapping _to_ it could be a second step.


And some other questions:

1. How would you handle special characters or whitespaces in map keys?
2. Is it possible to map from a map that is a property in the source-type? 

    `class Source{Map<> map;} class Target{Foo object;}` -> map from `source.map` to `target.object`
3. What if the key is *not* a String (or maybe some other simple type)... 
    
    Maybe in such a case we can search for a mapper as well (String to type of the mapping key)?


I think we've got a lot of small tasks here, maybe we should try to separate them and create smaller issues that could be solved step-by-step.

sjaakd:
> **Note 1:**
> 
> What is with this?
> 
> ```java
> @Mapping( target = "childFirstName", source = "child.firstName" )
> Target map(Map<String, Source> source);
> ```

What is the type that you obtain from the `Map`. on which you call `.getFirstName()`?

sjaakd:
> Good to hear that you are started with this feature! I would love to have it in place 🤩

Me too. But I want to take it slow (step by step). I wrote the above a couple of weeks ago.. Stuff came in between. So my memory is a bit blurry..

sjaakd:
> **Note 2:**
> 
> What do you mean with this point?
> 
> ```java
> class Target {
>   Map<String, String> foo;
>   // ... other properties
> }
> 
> Target map(Map<String, Source> source, Map<String, String> foo);
> ```
> 
> `foo` can't be mapped in such a case to `target.foo`? If yes, why?

Need to think about this as well. Sure it will pop into my mind when I start working again on this.. :smile:

sjaakd:
> 1. How would you handle special characters or whitespaces in map keys?

Not support this initially.. We could always relax the rules for mappings at some point, but you need to have the fact that you are mapping a map in the right place.

sjaakd:
> 2\. Is it possible to map from a map that is a property in the source-type?
>     `class Source{Map<> map;} class Target{Foo object;}` -> map from `source.map` to `target.object`

I think that should be possible..

sjaakd:
> 3\. What if the key is _not_ a String (or maybe some other simple type)...
>     Maybe in such a case we can search for a mapper as well (String to type of the mapping key)?

How would you see this? We normally map source properties by definition (in `@Mapping` or automatically by name).  We could assume if there's no defined mapping, the key is equal to the target bean property name.

chris922:
> > **Note 1:**
> > What is with this?
> > ```java
> > @Mapping( target = "childFirstName", source = "child.firstName" )
> > Target map(Map<String, Source> source);
> > ```
> 
> What is the type that you obtain from the `Map`. on which you call `.getFirstName()`?

`map.get("child")` will return `Source` (defined via generic) and this contains a property `firstName`.
Obviously we can't do something like this when there are no generics defined.

sjaakd:
> `map.get("child")` will return `Source` (defined via generic) and this contains a property `firstName`.

Ah.. but then all values would be of the same type.. What use case would that full fill?

chris922:
> > `map.get("child")` will return `Source` (defined via generic) and this contains a property `firstName`.
> 
> Ah.. but then all values would be of the same type.. What use case would that full fill?

Good point! So it is probably a rare case .. like .. uuuh.. this one

```java
class Family {
  private String fatherName;
  private String motherName;
  private String childName;
}

@Mapping(target = "fatherName", source = "father.name")
// ...
Family map(Map<String, Person> familyAsMap);
```
https://github.com/mapstruct/mapstruct/issues/1731#issuecomment-466820538
Here I described that maybe a new property like `sourceType` in `@Mapping` could be helpful, then you could do something like this:

```java
@Mapping(target = "bar", source = "foo.bar", sourceType = Foo.class)
Target map(Map<String, Object> map);
```

But that's definitely also something for a next step. Hmm.. I am missing a place where we can collect all ideas, order them, draft solutions... maybe a wiki-page or an issue where we always edit the first post?

sjaakd:
> Here I described that maybe a new property like `sourceType` in `@Mapping` could be helpful, then you could do something like this:

Funny.. I did not look at this PR a long time.. But started thinking about it last weekend. Biking  to my work, I thought: we should have a `sourceType` and `targetType` in the `@Mapping`  to be able to do nesting.

When not present, you can only assume the source type is the same as (deepest) target type

PR #1747, opened 2019-03-10
#1738: Use typeBound for the return type of the nested property mapping method and for the definition of the properties within the method

### LINKED ISSUE #1738: MapStruct producing uncompilable code when using bounded wildcard
Hello there :wave: 

As a bit of context, I am using JAXB as a way to convert XML messages to Java POJOs, and then using MapStruct to map those objects onto domain objects that my application uses. I have come across an issue in the way that MapStruct generates code when I use a bounded wildcard (e.g. `B<? extends C>`) in a source field. Here's an example:

```java
public class A {
  private B<? extends C> someField;
}
```
Let's say that `B` has a field `anotherField`. If I try to access any fields of `B` in a MapStruct `source` parameter, like this:

```java
@Mapping(source = "b.anotherField")
```

Then the generated code doesn't compile. I get something like this:

```java
private ? extends C mappingMethod(A mappingInput) {
  ? extends C someField = b.getSomeField();
  ...
}
```

This Java code obviously does not compile. The reason that I mention JAXB is that some of the source POJO's fields have a type of `JAXBElement<? extends SomeType>` which we can't really change.

Sorry for the abstract example, I can't display the actual code I am working on as it's part of a work project. If needed I can produce a [minimal working example](https://en.wikipedia.org/wiki/Minimal_working_example).

I am using `mapstruct-jdk8` version **1.3.0.Final**

Many thanks :smile:

filiphr:
Fixes #1738

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
@@ -6,7 +6,7 @@
 
 -->
 <#-- @ftlvariable name="" type="org.mapstruct.ap.internal.model.NestedPropertyMappingMethod" -->
-<#lt>private <@includeModel object=returnType/> ${name}(<#list parameters as param><@includeModel object=param/><#if param_has_next>, </#if></#list>)<@throws/> {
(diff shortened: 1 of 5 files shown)


filiphr:
This has been merged into master. Thanks for the review @sjaakd

PR #1748, opened 2019-03-10
Update plugins to latest versions

filiphr:
With this MapStruct processor compiles with Open JDK 13 (EA)


sjaakd:
oh.. and perhaps add checkstyle as well 😄

filiphr:
>oh.. and perhaps add checkstyle as well 😄

I need to test that a bit better. I know that they've changed things from time to time and if we are using something older and deprecated we might have problems 😄. I'll do that separately.

filiphr:
Merged into master. Thanks for the review @sjaakd

Issue #1756, opened 2019-03-18
Vague message when handling nested property mapping.


sjaakd:
When creating an empty bean (nested), MapStruct cannot (in the absence of properties) create a mapping. Therefore it stops and suggest to implement a hand written method. The error message however does not really point to the case.

sjaakd:
solved by PR #1757

filiphr:
Reopening as the current changes need to be reworked. See my comment in the PR

filiphr:
I shouldn't be commenting when I am tired it seems 😄. I was looking in the wrong place.

sjaakd:
> I shouldn't be commenting when I am tired it seems

I'm in need of some sleept too :smile:.. Have a nice evening!

PR #1757, opened 2019-03-18
#1756 better forged method error based on empty target bean properties

sjaakd:


chris922:
What is now the difference? You just added "No target bean properties found" at the beginning of the message?

sjaakd:
> What is now the difference? You just added "No target bean properties found" at the beginning of the message?

exactly.. I was puzzled when I got this warning. I had 2 beans defined (empty albeit), but I had to go deep into tracing to figure out what the reason was.

sjaakd:
@chris922 : thanks for reviewing 👍

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -39,7 +39,8 @@
     PROPERTYMAPPING_CREATE_NOTE( "creating property mapping: %s.", Diagnostic.Kind.NOTE ),
     PROPERTYMAPPING_SELECT_NOTE( "selecting property mapping: %s.", Diagnostic.Kind.NOTE ),
     PROPERTYMAPPING_MAPPING_NOT_FOUND( "Can't map %s to \"%s %s\". Consider to declare/implement a mapping method: \"%s map(%s value)\"." ),
-    PROPERTYMAPPING_FORGED_MAPPING_NOT_FOUND( "Can't map %s to %s. Consider to implement a mapping method: \"%s map(%s value)\"." ),
+    PROPERTYMAPPING_FORGED_MAPPING_WITH_HISTORY_NOT_FOUND( "No target bean properties found: can't map %s to \"%s %s\". Consider to declare/implement a mapping method: \"%s map(%s value)\"." ),
+    PROPERTYMAPPING_FORGED_MAPPING_NOT_FOUND( "No target bean properties found: can't map %s to %s. Consider to implement a mapping method: \"%s map(%s value)\"." ),
```
filiphr: This change is not OK. The `PROPERTYMAPPING_FORGED_MAPPING_NOT_FOUND` is also used in other places where this error message is not valid.

processor/src/test/java/org/mapstruct/ap/test/collection/forged/CollectionMappingTest.java
```diff
@@ -87,8 +87,8 @@ public void shouldForgeNewMapMappingMethod() {
             @Diagnostic(type = ErroneousCollectionNonMappableSetMapper.class,
                 kind = Kind.ERROR,
                 line = 17,
-                messageRegExp = "Can't map Collection element \".* nonMappableSet\" to \".* nonMappableSet\". "
-                    + "Consider to declare/implement a mapping method: .*."),
+                messageRegExp = "No target bean properties found: can't map Collection element \".* nonMappableSet\" "
```
filiphr: This should not have been changed. See my previous comment

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
sjaakd: I'll be happy to open it up again.. But I can only find one reference of both messages.. Its in

```java
            if ( method instanceof ForgedMethod && targetProperties.isEmpty() ) {
                //TODO until we solve 1140 we report this error when the target properties are empty
                ForgedMethod forgedMethod = (ForgedMethod) method;
                if ( forgedMethod.getHistory() == null ) {
                    Type sourceType = this.method.getParameters().get( 0 ).getType();
                    Type targetType = this.method.getReturnType();
                    ctx.getMessager().printMessage(
                        this.method.getExecutable(),
                        Message.PROPERTYMAPPING_FORGED_MAPPING_NOT_FOUND,
                        sourceType,
                        targetType,
                        targetType,
                        sourceType
                    );
                }
                else {
                    ForgedMethodHistory history = forgedMethod.getHistory();
                    ctx.getMessager().printMessage(
                        this.method.getExecutable(),
                        Message.PROPERTYMAPPING_FORGED_MAPPING_WITH_HISTORY_NOT_FOUND,
                        history.createSourcePropertyErrorMessage(),
                        history.getTargetType(),
                        history.createTargetPropertyName(),
                        history.getTargetType(),
                        history.getSourceType()
                    );
                }
            }
```

Look at the line ` if ( method instanceof ForgedMethod && targetProperties.isEmpty() ) `.. that coincides with the new message, right?

processor/src/test/java/org/mapstruct/ap/test/collection/forged/CollectionMappingTest.java
sjaakd: I'm not so sure.. the full error message (removing the wild card matches is:

`No target bean properties found: can't map Collection element "org.mapstruct.ap.test.collection.forged.Foo nonMappableSet" to "org.mapstruct.ap.test.collection.forged.Bar nonMappableSet". Consider to declare/implement a mapping method: "org.mapstruct.ap.test.collection.forged.Bar map(org.mapstruct.ap.test.collection.forged.Foo value)`

`Bar` is an empty bean (No properties).

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
filiphr: Ah yes you are right. I searched for the wrong message. I was searching for `PROPERTYMAPPING_MAPPING_NOT_FOUND` and not for `PROPERTYMAPPING_FORGED_MAPPING_NOT_FOUND`. Sorry for the mistake

processor/src/test/java/org/mapstruct/ap/test/collection/forged/CollectionMappingTest.java
filiphr: You are right, sorry for my mistake

PR #1760, opened 2019-03-20
Typo in builder documentation

juliojgd:
It is "then" instead "than


filiphr:
Thanks for the fix @juliojgd

PR #1765, opened 2019-03-23
Make hickory an optional dependency as well

filiphr:
Using optional because IntelliJ is picking up hickory as a transitive
dependency and runs the hickory processor in projects using mapstruct-processor.
This happens only when the processor is defined in the maven-compiler annotationProcessorPaths.
This is related to https://youtrack.jetbrains.com/issue/IDEA-200481.

### DIFF
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -36,10 +36,14 @@
         <!-- Compile-only; Using "provided" scope as there is no such scope in Maven;
         these dependencies are not required at runtime, only for prism generation
         and tests -->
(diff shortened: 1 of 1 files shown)


sjaakd:
ah... that's causing the warning in Intellij?

filiphr:
Yes that is causing the warning and I got fed up of it and I decided to try to see if we can do something. Looking at the issue from IntelliJ the entire processor path that they compute is wrong 😄.

filiphr:
Thanks for the review @sjaakd

PR #1766, opened 2019-03-23
Remove use of prerequisites from parent pom.xml

filiphr:
According to https://maven.apache.org/pom.html#Prerequisites, prerequisites are only checked with Maven 2.
The maven-enforcer-plugin could be used for this check. However, Considering that Maven 2 is not really used I think that we don't really need such check in our build


filiphr:
Thanks for the review @sjaakd

Issue #1772, opened 2019-03-28
Bug with Nested Properties and ReportingPolicy.ERROR on Unmapped Source Properties


jengelhart0:
As requested by @sjaakd, filing the issue described at https://stackoverflow.com/questions/55365722/mapstruct-issue-with-nested-properties-and-reportingpolicy-error-on-unmapped-so. 

Text of the above link question copied below for convenience.

-------

Using MapStruct, we want to use ReportingPolicy.ERROR, and have code like the following:

`@Mapping(source = "nestedSource.doublyNestedSourceField", target = "nestedTarget.doublyNestedTargetField")
Target mapSourceToTarget(Source source);`

Where nestedSource is not the same type as nestedTarget, and both doublyNested*Field types are String.

There is no mapper declared for NestedSource -> NestedTarget. The String properties declared in the Mapping above are the only ones in those types.

The above causes an unmapped source error:
`Unmapped source property: "doublyNestedSourceField".`

That seems more-or-less reasonable, as we didn't declare a mapper for NestedSource -> NestedTarget.

However, here's the issue: If we change the ReportingPolicy for unmapped sources to warn/ignore, MapStruct figures out how to correctly map the doublyNestedSourceField in the mapper implementation, even though it claims there is no source mapping present. Just wondering what is going on here, and whether I'm missing something.

----Into the weeds a bit more (in the MapStruct code itself)----

I could be doing something wrong, but I did notice that in BeanMethodMapping.java MapStruct attempts to remove "nestedSource.doubleNestedSourceField" from unprocessedSourceProperties, even though the key for the appropriate property is just "nestedSource" in unprocessedSourceProperties. Thus "nestedSource" is left as an unprocessed source property and an error is thrown.

- [x] Is this an issue (and hence not a question)?

sjaakd:
merged into master. @jengelhart0 : you can try the master to see of the problem is fixed.

PR #1778, opened 2019-04-02
#1772 unmapped source prop remaining when target and source entry diff

sjaakd:


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -393,26 +393,35 @@ private boolean handleDefinedMappings() {
             }
 
             for ( Map.Entry<String, List<Mapping>> entry : methodMappings.entrySet() ) {
+                String target = null;
```
chris922: Any reasons you define `target` in this scope and not where it is used? (line 400ff)

sjaakd: oops.. that's a remanence of a previous iteration.. Spent a lot of time on this one.. (put System.out.println's to find out how this works exactly.. I'll fix this.

Issue #1457, opened 2018-04-24
@AfterMapping methods with non unique parameters


lazydays79:
Hi!

First of all: we are using MapStruct now for almost 2 years in a big project and it saved us a lot of time, headache and coding. So for this I want to give you a big thanks and thumbs up! :grin: 

But all of the sudden, MapStruct cost us a lot of analysis, research, nerves and coke... but I think we found some issue. Ok, not a real issue, it's more like an "undefined behaviour" which should somehow be taken care of.

Let's assume we have a mapping method (defined in an abstract class) includung an @AfterMapping method. The method signatures both of the mapping and the after mapping method include two "context parameters", both of type String.

I must admit, that the bug that now occurs is made by us: we forgot to add the @Context annotation to the String parameters. If we would have done it right, MapStruct would give up throwing an error:

_The types of @Context parameters must be unique._

But we missed to add the annotation, so MapStruct generated a mapper which is calling the AfterMapping method wrong: using the first String parameter twice.

So, is this a bug in MapStruct? No, not really. It's just undefined behaviour. No question about that.

Could MapStruct have saved us some hours, if it would complain about this? Yes, I guess so. ;-)

So, I suggest to either let MapStruct throw a compilation warning or even error, when @Context is missing. Or it should complain about non unique parameters as it would with @Context annotated parameters.

To make it more clear what I'm blathering about, I created a simple test project with might help you to understand what I mean:

[https://github.com/lazydays79/mapstructissue2](https://github.com/lazydays79/mapstructissue2)

Feel free to clone, compile, look at the generated mapper and enjoy the failing test. :laughing: 

If you won't fix this, I will totally accept it. But I guess it would be a lot more helpfull (and time saving) to let MapStruct fail on this undefined behaviour than to let it generale a wrong Mapper implementation in silence.

Best regards,
Michael

sjaakd:
HI Michael,

Thanks for your extensive analysis. I'll have a look soon if we can generate  a  warning in this case.

Best regards,
SJaak

filiphr:
I have to say reading such an issue is a joy to read. It's always nice to hear such nice words from users and we are glad that MapStruct has saved you time. Will try to do something so such pesky issues don't lose more time to you or other users.

RaymondDubovik:
I had the same issue when I tried to upgrade from mapstruct 1.1.0.Final to mapstruct 1.2.0.Final soon after it's release.
[...]

PR #1782, opened 2019-04-03
[#1457] Stricter matching for lifecycle methods / non-unique parameters

chris922:
In case a lifecycle method has multiple matching parameters (e. g. same type)
all parameter names must match exactly with the ones from the mapping method,
otherwise the lifecycle method will not be used and a warning will be shown.

As soon as the warning is displayed the lifecycle method will *not* be used, this might be a breaking change, but if the method will be used there is a high chance that it will be used in a wrong way (invalid parameter ordering, see #1457). The warning is not coming up for any of our tests.

I would prefer that only the names of the ambiguous parameters must match, but imho this would lead to bigger and really complex changes (see discussion in the ticket). As this here should already fix the issue and provide a possible solution (rename all parameters) it should be better to have it in mapstruct than waiting for a (more or less) complete rework of the "find-matching-method" algorithm.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
@@ -5,10 +5,12 @@
  */
 package org.mapstruct.ap.internal.model.source.selector;
 
-import static org.mapstruct.ap.internal.util.Collections.first;
-
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
 import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.common.ParameterBinding;
@@ -17,6 +19,10 @@ import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.common.TypeFactory;
 import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.model.source.MethodMatcher;
+import org.mapstruct.ap.internal.util.FormattingMessager;
+import org.mapstruct.ap.internal.util.Message;
+
+import static org.mapstruct.ap.internal.util.Collections.first;
 
 /**
  * Selects those methods from the given input set which match the given source and target types (via
@@ -27,9 +33,11 @@ import org.mapstruct.ap.internal.model.source.MethodMatcher;
 public class TypeSelector implements MethodSelector {
 
     private TypeFactory typeFactory;
+    private FormattingMessager messager;
 
-    public TypeSelector(TypeFactory typeFactory) {
+    public TypeSelector(TypeFactory typeFactory, FormattingMessager messager) {
         this.typeFactory = typeFactory;
+        this.messager = messager;
     }
 
     @Override
@@ -63,7 +71,7 @@ public class TypeSelector implements MethodSelector {
 
             if ( parameterBindingPermutations != null ) {
                 SelectedMethod<T> matchingMethod =
-                    getFirstMatchingParameterBinding( targetType, method, parameterBindingPermutations );
+                    getMatchingParameterBinding( targetType, mappingMethod, method, parameterBindingPermutations );
 
                 if ( matchingMethod != null ) {
                     result.add( matchingMethod );
@@ -77,7 +85,6 @@ public class TypeSelector implements MethodSelector {
         SourceRHS sourceRHS) {
         List<ParameterBinding> availableParams = new ArrayList<>( method.getParameters().size() + 3 );
 
-        addMappingTargetAndTargetTypeBindings( availableParams, targetType );
         if ( sourceRHS != null ) {
             availableParams.addAll( ParameterBinding.fromParameters( method.getContextParameters() ) );
             availableParams.add( ParameterBinding.fromSourceRHS( sourceRHS ) );
@@ -86,6 +93,8 @@ public class TypeSelector implements MethodSelector {
             availableParams.addAll( ParameterBinding.fromParameters( method.getParameters() ) );
         }
 
+        addMappingTargetAndTargetTypeBindings( availableParams, targetType );
+
         return availableParams;
     }
 
@@ -93,8 +102,6 @@ public class TypeSelector implements MethodSelector {
             Type targetType, Method mappingMethod) {
 
         List<ParameterBinding> availableParams = new ArrayList<>( sourceTypes.size() + 2 );
-
-        addMappingTargetAndTargetTypeBindings( availableParams, targetType );
 
         for ( Type sourceType : sourceTypes ) {
             availableParams.add( ParameterBinding.forSourceTypeBinding( sourceType ) );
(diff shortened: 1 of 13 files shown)


sjaakd:
Hi Chris,

I got some remaining questions:

1. Is this only for lifecyle methods? Or also for selecting mapping methods / factory methods?

2. Is the order of parameters still relevant? 

3. What if you have less parameters to choose from in the lifecycle method?

4. Looking a the solution, I see that `@MappingTarget` and `@TargetType` parameters are excluded, right?

Now: I do think we miss some unit test cases (or perhaps they are covered elsewhere)..

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -106,24 +111,101 @@ public TypeSelector(TypeFactory typeFactory) {
...
+        }
+        if ( !targetTypeAvailable ) {
+            availableParams.add( ParameterBinding.forTargetTypeBinding( typeFactory.classTypeOf( targetType ) ) );
+        }
     }
 
-    private <T extends Method> SelectedMethod<T> getFirstMatchingParameterBinding(Type targetType,
-            SelectedMethod<T> method, List<List<ParameterBinding>> parameterAssignmentVariants) {
+    private <T extends Method> SelectedMethod<T> getMatchingParameterBinding(Type targetType,
+            Method mappingMethod, SelectedMethod<T> method, List<List<ParameterBinding>> parameterAssignmentVariants) {
```
sjaakd: rename `method` to something like `selectedMethodInfo`

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -106,24 +111,101 @@ public TypeSelector(TypeFactory typeFactory) {
...
+
+        List<List<ParameterBinding>> matchingParameterAssignmentVariants = new ArrayList<>(
+            parameterAssignmentVariants
+        );
 
-        for ( List<ParameterBinding> parameterAssignments : parameterAssignmentVariants ) {
-            if ( method.getMethod().matches( extractTypes( parameterAssignments ), targetType ) ) {
-                method.setParameterBindings( parameterAssignments );
-                return method;
+        Method realMethod = method.getMethod();
```
sjaakd: rename `realMethod` to `method` or perhaps even `selectedMethod` (consistency)

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -106,24 +111,101 @@ public TypeSelector(TypeFactory typeFactory) {
...
+            // no matching variants found
+            return null;
+        }
+        else if ( matchingParameterAssignmentVariants.size() == 1 ) {
+            // we found exactly one set of variants, use this
+            method.setParameterBindings( first( matchingParameterAssignmentVariants ) );
+            return method;
+        }
+
+        // more than one variant matches, try to find one where even the parameter names are matching
```
sjaakd: in comment: "even"  == "also"

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -106,24 +111,101 @@ public TypeSelector(TypeFactory typeFactory) {
...
+        for ( Iterator<List<ParameterBinding>> it = matchingParameterAssignmentVariants.iterator(); it.hasNext(); ) {
```
sjaakd: why use 2 styles of for loops (inner / outer)? Perhaps better to use one. Makes code harder to understand.

Also: perhaps refactor inner loop into a method.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -106,24 +111,101 @@ public TypeSelector(TypeFactory typeFactory) {
...
+            List<ParameterBinding> parameterAssignments = it.next();
+
+            int i = 0;
+            for ( ParameterBinding parameterBinding : parameterAssignments ) {
+                Parameter parameter = realMethod.getParameters().get( i++ );
+
+                // if the parameterBinding contains a parameter name we must ensure that this matches the name from the
+                // method parameter -> remove all variants where this is not the case
+                if ( parameterBinding.getVariableName() != null &&
+                    !parameter.getName().contentEquals( parameterBinding.getVariableName() ) ) {
```
sjaakd: why `contentEquals`?

sjaakd:
One other thing that is on my mind: couldn't we keep the current solution and add an additional selector (after the type selector) that selects on names if there are still conflicts?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
chris922: I prefer the for-each style (inner loop), but for the outer-loop I had to use the iterator, because entries within the `matchingParameterAssignmentVariants` list should be removed.

But I moved the inner-loop to an own method and switched to `removeIf` like I do it a few lines before.

Fine for you?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
chris922: good question.. don't remember.. both are `String`s ... changed!

chris922:
Thanks for the review, @sjaakd !

> 1. Is this only for lifecyle methods? Or also for selecting mapping methods / factory methods?

Should be valid for all cases, but I will add some tests to verify this.

> 2. Is the order of parameters still relevant?

No :)

> 3. What if you have less parameters to choose from in the lifecycle method?

Should also be fine.

> 4. Looking a the solution, I see that `@MappingTarget` and `@TargetType` parameters are excluded, right?

No, should stay as it was. 
But I noticed that these parameters were always added as a default and afterwards maybe once again because they were defined as well. I changed this, so that the defaults will only be added if they weren't added already before.

> Now: I do think we miss some unit test cases (or perhaps they are covered elsewhere)..

I just made changes according to your inline comments and will continue with adding a few more tests.

> One other thing that is on my mind: couldn't we keep the current solution and add an additional selector (after the type selector) that selects on names if there are still conflicts?

Idea sounds nice, but I had a short look and it looks like it's not so easy.
The selectors are returning list of matching methods, we are analyzing here exactly one possible method and are trying to find the correct parameters and their ordering. Making a new selector would require that this `TypeSelector` returns the same method several times for all possible parameter variants, so we have to duplicate the `SelectedMethod` and the new selector will reduce this set afterwards.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
sjaakd: its fine 👍

sjaakd:
LGTM.. Nice work @chris922 .. Merge when you feel you're done..

lazydays79:
Thanks @chris922 ! :)

Issue #1776, opened 2019-04-01
Inverted presence checker (absence checker)


baurchanu:
Hello. We use your framework in our project and face some limitations of using a presence checker feature. Now  it's impossible to check absence of a property using a custom method (hasNoX / isNullX), isn't it? So we need write something like @Mapping(target="X", expression="java(src.isNullX() ? null : src.getX())"). If absence checker functionality is interesting not only for us, I would implement this enhancement and submit PR.

sjaakd:
Is already implemented. Checkout documentation

baurchanu:
@sjaakd Could you please specify in which article of the documentation I can find example with a custom absence checker? I mean a checker with logic different than simple not null check ( `return X == null` ). Sorry for wasting your time, if I missed something.

sjaakd:
http://mapstruct.org/documentation/stable/reference/html/#source-presence-check

sjaakd:
Custom via spi

sjaakd:
You can change the `hasX` into `isNull` in the SPI but you can't negate. So, the first thing that comes to my mind is to extend the SPI with some kind of boolean check (downward compatible), with default false; E.g. negate.

filiphr:
@baurchanu, you don't have control over the objects with the `isNullX()` method?

@sjaakd I am not certain whether it is worthwhile to introduce such complexity. To me currently it sounds really specific and would really be interested what is generating those `isNullX()` methods (if it is not in the control of the OP)

baurchanu:
@filiphr, yes, the objects are generated sources, and I have no opportunity to customize the process of the generation.

PR #1786, opened 2019-04-06
#1776 adding a message when no qualifiers are found

sjaakd:


sjaakd:
@filiphr / @chris922 .. does one of you guys have a possibility to review this one?

processor/src/main/java/org/mapstruct/ap/internal/model/MappingBuilderContext.java
```diff
@@ -93,7 +93,13 @@
         Assignment getTargetAssignment(Method mappingMethod, Type targetType,
                                        FormattingParameters formattingParameters,
                                        SelectionCriteria criteria, SourceRHS sourceRHS,
-                                       AnnotationMirror positionHint);
+                                       AnnotationMirror positionHint,
+                                       Forger forger);
+
+        @FunctionalInterface
+        interface Forger {
```
filiphr: Why not use `java.util.function.Supplier`? Also if we go with our own, maybe `AssignmentForger`?

sjaakd: I went for the `Supplier`.. Perhaps better than inventing our own types..

sjaakd:
@filiphr .. thanks for reviewing!

Issue #1742, opened 2019-03-06
MapStruct 1.3.0.Final produces uncompilable code when using Java 11 + Lombok 1.18.6


DmitriiSolovev:
My setup: Java 11.02, Gradle 5.2.1, MapStruct 1.3.0.Final, Lombok 1.18.6 and Spring Boot 2.3.1.

**build.gradle**

```groovy
plugins {
    id "net.ltgt.apt" version "0.21"
    id "net.ltgt.apt-idea" version "0.21"
}

...

/* Lombok */
compileOnly "org.projectlombok:lombok:${lombokVersion}"
annotationProcessor "org.projectlombok:lombok:${lombokVersion}"

/* Map Struct */
implementation "org.mapstruct:mapstruct-jdk8:${mapstructVersion}"
annotationProcessor "org.mapstruct:mapstruct-processor:${mapstructVersion}"
testAnnotationProcessor "org.mapstruct:mapstruct-processor:${mapstructVersion}"
```

**FirstEntity.java**
```java
@Getter
@Setter
@Builder
@NoArgsConstructor
@AllArgsConstructor
@EqualsAndHashCode(of = "id")
@Entity
@Table(name="first_entity")
public class FirstEntity {

  @Id
  private Long id;

  @ManyToOne
  @JoinColumn(name="second_entity_id")
  private SecondEntity secondEntity;

}
```

**FirstEntityDTO.java**
```java
@Data
@Builder
@NoArgsConstructor
@AllArgsConstructor
public class FirstEntityDTO {

  private Long id;
  private SecondEntityDTO secondEntity;

}
```

**SecondEntity.java**
```java
@Getter
@Setter
@Builder
@NoArgsConstructor
@AllArgsConstructor
@EqualsAndHashCode(of = "id")
@Entity
@Table(name="second_entity")
public class SecondEntity {

  @Id
  private Long id;

}
```

**SecondEntityDTO.java**
```java
@Data
@Builder
@NoArgsConstructor
@AllArgsConstructor
public class SecondEntityDTO {

  private Long id;

}
```

**EntityMapper.java**
```java
@Mapper(
        unmappedTargetPolicy = ReportingPolicy.IGNORE,
        componentModel = "spring"
)
public interface EntityMapper {

    FirstEntity updateEntity(FirstEntityDTO entityDto, @MappingTarget FirstEntity entity);
}
```

**Generated code, MapStruct 1.3.0.Final**
```java
[...]

PR #1805, opened 2019-05-03
#1742 refactor accessors (make consistent) to have accessed type alwa…

sjaakd:
…ys available

Currently the accessor type is defined in `PropertyMapper` while it should be an intrinsic property of `Accessor`.  Types are fetched and derived in context (again) all-over-the-place.. Sometimes that requires the parent type again and even the mapping method (to derive in context and resolve generics). 

This PR is the first in a number of refactoring with the goal to have a cleaner separation of concerns.


sjaakd:
followed up by #1811

PR #1806, opened 2019-05-03
#1797 Use EnumSet.noneOf when creating a new instance of EnumSet

### LINKED ISSUE #1797: [REGRESSION] Make MapStruct generate compilable mapping code for attributes of type EnumSet
See #732, does not work anymore

filiphr:
Fixes #1797

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/IterableCreation.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/IterableCreation.java
@@ -12,6 +12,8 @@ import org.mapstruct.ap.internal.model.common.ModelElement;
 import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.common.Type;
 
+import static org.mapstruct.ap.internal.util.Collections.first;
+
 /**
  * Model element that can be used to create a type of {@link Iterable} or {@link java.util.Map}. If an implementation
  * type is used and the target type has a constructor with {@code int} as parameter and the source parameter is of
@@ -69,6 +71,20 @@ public class IterableCreation extends ModelElement {
         if ( factoryMethod == null && resultType.getImplementationType() != null ) {
             types.addAll( resultType.getImplementationType().getImportTypes() );
         }
+
+        if ( isEnumSet() ) {
+            types.add( getEnumSetElementType() );
+            // The result type itself is an EnumSet
+            types.add( resultType );
+        }
         return types;
+    }
+
+    public Type getEnumSetElementType() {
+        return first( getResultType().determineTypeArguments( Iterable.class ) );
+    }
+
+    public boolean isEnumSet() {
+        return "java.util.EnumSet".equals( resultType.getFullyQualifiedName() );
     }
 }
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/IterableCreation.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/IterableCreation.ftl
@@ -9,6 +9,8 @@
 <@compress single_line=true>
     <#if factoryMethod??>
         <@includeModel object=factoryMethod targetType=resultType/>
+    <#elseif enumSet>
+        EnumSet.noneOf( <@includeModel object=enumSetElementType raw=true/>.class )
     <#else>
     new
         <#if resultType.implementationType??>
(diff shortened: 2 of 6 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/IterableCreation.java
```diff
@@ -69,6 +71,20 @@ public boolean isLoadFactorAdjustment() {
...
+        }
         return types;
     }
+
+    public Type getEnumSetElementType() {
+        return first( getResultType().determineTypeArguments( Iterable.class ) );
+    }
+
+    public boolean isEnumSet() {
+        return "java.util.EnumSet".equals( resultType.getFullyQualifiedName() );
```
chris922: What about this?

```suggestion
        return EnumSet.class.getName().equals( resultType.getFullyQualifiedName() );
```

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableCreation.ftl
```diff
@@ -9,6 +9,8 @@
 <@compress single_line=true>
     <#if factoryMethod??>
         <@includeModel object=factoryMethod targetType=resultType/>
+    <#elseif enumSet>
+        EnumSet.noneOf( <@includeModel object=enumSetElementType raw=true/>.class )
```
chris922: Is there somehow a way to do this more generic and don't hard-code the enum-set "instantiation" here in the template? Like the previous branch is there for all kind of factory methods.. or just keep it simple as we don't have any other use-case up to now?

filiphr: The reason why I did it like that was that it was already done in a similar way in [`ExistingInstanceSetterWrapperForCollectionsAndMaps.ftl`](https://github.com/mapstruct/mapstruct/blob/871353fccbf8ed05cfc7dee055f99eb81b0e69e0/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/ExistingInstanceSetterWrapperForCollectionsAndMaps.ftl#L39) and [`SetterWrapperForCollectionsAndMapsWithNullCheck.ftl`](https://github.com/mapstruct/mapstruct/blob/871353fccbf8ed05cfc7dee055f99eb81b0e69e0/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMapsWithNullCheck.ftl#L27).

I was thinking whether we can somehow create a fake factory method, but with the current way this was the most simple way to do it.

processor/src/main/java/org/mapstruct/ap/internal/model/IterableCreation.java
filiphr: I copied this from [`SetterWrapperForCollectionsAndMapsWithNullCheck`](https://github.com/mapstruct/mapstruct/blob/871353fccbf8ed05cfc7dee055f99eb81b0e69e0/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapperForCollectionsAndMapsWithNullCheck.java#L70-L72).

We could do it like your suggestion as well, not big difference I think. I doubt that the `EnumSet` will change it's package anytime soon 😄

sjaakd:
From the top of my head.. isn't there a list of interfaces and respective implementations in Typefactory?

Verstuurd vanaf mijn iPhone

processor/src/main/java/org/mapstruct/ap/internal/model/IterableCreation.java
chris922: Totally fine for me

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableCreation.ftl
chris922: Okay, then let us "keep it simple, stupid" as long as we don't need a more generic way suitable for similar cases :)

filiphr:
>From the top of my head.. isn't there a list of interfaces and respective implementations in Typefactory?

Yes, there is an `implementationType` for a certain `Type`. However, the `EnumSet` does not have an implementation type since it is a class. We then do:

```
    new
        <#if resultType.implementationType??>
            <@includeModel object=resultType.implementationType/><#if ext.useSizeIfPossible?? && ext.useSizeIfPossible && canUseSize>( <@sizeForCreation /> )<#else>()</#if>
        <#else>
            <@includeModel object=resultType/>()</#if>
        </#if>
```

filiphr:
@chris922 if you are OK with the PR I will merge it. No need to wait for @sjaakd I think :)

sjaakd:
No. No need 😉. 

Verstuurd vanaf mijn iPhone

filiphr:
Thanks for the review @chris922

PR #1810, opened 2019-05-08
#1608 Oracle's JavaBean getter/setter support

msymonov:
Originally Oracle declares what can be named getter/setter. And it's not only about using proper prefix "get", "set", "is", but also the character following the prefix has to be uppercase. This pull request fixes the naming pattern for bean properties. All existing tests and their test data are updated accordingly.

### DIFF
--- a/integrationtest/src/test/resources/namingStrategyTest/strategy/src/main/java/org/mapstruct/itest/naming/CustomAccessorNamingStrategy.java
+++ b/integrationtest/src/test/resources/namingStrategyTest/strategy/src/main/java/org/mapstruct/itest/naming/CustomAccessorNamingStrategy.java
@@ -29,20 +29,20 @@ public class CustomAccessorNamingStrategy  extends DefaultAccessorNamingStrategy
     public boolean isSetterMethod(ExecutableElement method) {
         String methodName = method.getSimpleName().toString();
 
-        return methodName.matches( "^with[A-Z].*" );
+        return IntrospectorUtils.isWither( methodName );
     }
 
     @Override
     public boolean isAdderMethod(ExecutableElement method) {
         String methodName = method.getSimpleName().toString();
-        return methodName.matches( "^add[A-Z].*" );
+        return IntrospectorUtils.isAdder( methodName );
     }
 
     @Override
     public String getPropertyName(ExecutableElement getterOrSetterMethod) {
         String methodName = getterOrSetterMethod.getSimpleName().toString();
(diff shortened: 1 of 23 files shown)


chris922:
Thank you very much for the PR!

I remember a discussion in some issue or PR where we insisted to follow the JavaBean specs and IIRC there were some cases where after `set`/`get`/... a lower-case letter could follow... maybe @sjaakd or @filiphr remember this discussion?

Besides that: In case we continue on your PR I prefer to not use regular expressions. Your expression expects a letter from A to Z, but there are even more upper-case letters possible in other languages (e. g. Ä, Ö and Ü in German).

What about staying with the checks like this: `methodName.startsWith( "get" ) && methodName.length() > 3`
But adding something like: `&& Character.isUpperCase( methodName.charAt( 3 ) )` ?

By the way: I like that you moved the checks to own methods, but why don't you use them in the `CustomAccessorNamingStrategy`? `IntrospectorUtils` belongs to the SPI and thus it is allowed to use it.

msymonov:
Christian, thx very much for your comments, 2 from 3 were taken into account:

- CustomAccessorNamingStrategy. Newly added methods are used here as well
- Not latin-only uppercase letters are supported now. Still via regexp though, see character class "p{javaUpperCase}"

So it seems that the only unresolved point by now is the main question, if we can consider as getter/setter the method which has lowercase letter after the prefix. Could anyone comment on this item?

msymonov:
@filiphr what's your view on this item?

msymonov:
@filiphr Thanks for reviewing the code. To keep the discussion moving in a constructive direction i can say that Lombok generates a getter "getAProperty()" for a property named "aProperty". So i agree some expectations could be broken, especially for those who use Intellij, however it's not correct as of me. I tried to analyze java.beans.Introspector(which the JavaBeans spec refers to) to understand how the creators of the spec declared the bean properties vs getters/setters in terms of conventional naming.

filiphr:
I saw that `Lombok` always uppercases the first letter of the property. 

I would like to avoid adding any breaking changes if possible. In this case I think it won't be that difficult to do it. We can do something similar as we did for fluent setters with "is" and something similar that we did with the adders. Where we explicitly check that the 4th character is upper case. That check would be for fluent java bean setters

filiphr:
Closing this in favour of https://github.com/mapstruct/mapstruct/pull/1917. Hope you are OK with it @msymonov. This one had a lot of breaking changes that I was not really happy with. Thanks for the PR in any case

Issue #1661, opened 2018-12-01
Opt-out  of using builders via a processor option


twentylemon:
Possibly related to #1547 

I've recently been playing with 1.3 (from 1.2), testing out builder/immutable object support. I'm trying to map between 2 classes, a service layer DTO and our internal model representation. Currently, our internal model is a lombok `@Data` class, as mapstruct requires mutable objects. Ideally, we'd move to `@Value` or otherwise make an immutable object, as part of the 1.3 upgrade.

The DTO is generated, mutable and comes with a builder class, the builder has setters named `withAttribute`. Mapstruct 1.3 doesn't know how to map these out of the box -- it tries to use the builder then complains about not mapping `withAttribute`. I started trying out the `AccessorNamingStrategy` [spi](http://mapstruct.org/documentation/dev/reference/html/#_custom_accessor_naming_strategy). It works, but I have to support all strategies with a single class (mapstruct throws if you provide more than 1 implementation).
Aside: I found this to be odd, as `ServiceLoader`s typically get multiple implementations of a service (like `Processor`). This would allow me to write a single class to support a `withAttribute` naming strategy, without having to worry about supporting `setAttribute` or `attribute` still. Mapstruct could in theory support multiple by trying all implementations before being unable to find a method.
While, strictly speaking, this universal naming strategy worked, I didn't feel confident that the strategy would cover all existing cases across all of our code, and future cases may be hard to integrate and be blocked.

So. Not overly confident in that, I searched for some way to simply turn off builder support. It seems this is only possible mapstruct-wide.
> In case you want to disable using builders then you can use the NoOpBuilderProvider by creating a org.mapstruct.ap.spi.BuilderProvider file in the META-INF/services directory with org.mapstruct.ap.spi.NoOpBuilderProvider as it’s content.

Not ideal, as I wanted to change our internal model to be immutable and don't want to support `withAttribute` through the spi. For my situation, I'd rather see a way to specify *this method, use the builder*; *this method, it's just a bean*. Specifying an arg as "don't use builders" would be a seamless way for me to upgrade, but mapstruct's default could be "autodetect builders" as it currently does. Overriding this config at a mapper or method level would allow me to opt-in to new features, so the new version is fully backward compatible.

filiphr:
>I started trying out the `AccessorNamingStrategy` spi. It works, but I have to support all strategies with a single class (mapstruct throws if you provide more than 1 implementation).

[...]

PR #1811, opened 2019-05-08
#1742 refactoring and making builder optional

### LINKED ISSUE #1742: MapStruct 1.3.0.Final produces uncompilable code when using Java 11 + Lombok 1.18.6

sjaakd:
Currently the accessor type is defined in PropertyMapper while it should be an intrinsic property of Accessor. Types are fetched and derived in context (again) all-over-the-place.. Sometimes that requires the parent type again and even the mapping method (to derive in context and resolve generics).

Next, builderType is part of type. One could say there is a relation from builderType to Type but not the other way around. A builderType could be in a completely different package. It's still not as I want.. But I managed to move the builderType (mostly) to the place were its relevant: the `BeanMappingMethod`. 

Done!

I'm still not completely happy with the number of places were I do need to consult the typeFactory for the `effectiveType` or the `builderType`. This indicates a bad separation of concerns. Probably stemming from the fact that we do analyse a nested target and try to find the associated types in `TargetReference`. I think that in the end `TargetReference` and `SourceReference` should be handled in the `BeanMappingMethod` and not in `Mapping` upfront but in the context of the `BeanMappingMethod`.

I experimented with making annotations and annotation fields `Optional`. I think we should apply this pattern, because we keep on doing null checks all over the place.

Fixes #1742

### DIFF
--- a/core/src/main/java/org/mapstruct/Builder.java
+++ b/core/src/main/java/org/mapstruct/Builder.java
@@ -29,4 +29,12 @@ public @interface Builder {
      * @return the method that needs to tbe invoked on the builder
      */
     String buildMethod() default "build";
+
+    /**
+     * Toggling builders on / off. Builders are sometimes used solely for unit testing (fluent testdata)
+     * MapStruct will need to use the regular getters /setters in that case.
+     *
+     * @return when true, no builder patterns will be applied
+     */
+    boolean noBuilder() default false;
 }
--- a/core/src/main/java/org/mapstruct/Builder.java
+++ b/core/src/main/java/org/mapstruct/Builder.java
@@ -36,5 +36,5 @@ public @interface Builder {
      *
      * @return when true, no builder patterns will be applied
      */
-    boolean noBuilder() default false;
+    boolean disableBuilder() default false;
 }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ObjectFactoryMethodResolver.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ObjectFactoryMethodResolver.java
@@ -38,7 +38,6 @@ public class ObjectFactoryMethodResolver {
      * returns a no arg factory method
      *
      * @param method target mapping method
-     * @param targetType return type to match
      * @param selectionParameters parameters used in the selection process
      * @param ctx
      *
@@ -46,9 +45,29 @@ public class ObjectFactoryMethodResolver {
      *
      */
     public static MethodReference getFactoryMethod( Method method,
-                                                    Type targetType,
                                                     SelectionParameters selectionParameters,
                                                     MappingBuilderContext ctx) {
+        return getFactoryMethod( method, method.getResultType(), selectionParameters, ctx );
+    }
+
+
+
+    /**
+     * returns a no arg factory method
+     *
+     * @param method target mapping method
+     * @param alternativeTarget alternative to {@link Method#getResultType()} e.g. when target is abstract
+     * @param selectionParameters parameters used in the selection process
+     * @param ctx
+     *
+     * @return a method reference to the factory method, or null if no suitable, or ambiguous method found
+     *
+     */
+    public static MethodReference getFactoryMethod( Method method,
+                                                    Type alternativeTarget,
+                                                    SelectionParameters selectionParameters,
+                                                    MappingBuilderContext ctx) {
+
 
         MethodSelectors selectors =
             new MethodSelectors( ctx.getTypeUtils(), ctx.getElementUtils(), ctx.getTypeFactory(), ctx.getMessager() );
@@ -58,18 +77,18 @@ public class ObjectFactoryMethodResolver {
                 method,
                 getAllAvailableMethods( method, ctx.getSourceModel() ),
                 java.util.Collections.<Type> emptyList(),
-                targetType.getEffectiveType(),
+                alternativeTarget,
                 SelectionCriteria.forFactoryMethods( selectionParameters ) );
 
         if (matchingFactoryMethods.isEmpty()) {
-            return findBuilderFactoryMethod( targetType );
+            return null;
         }
 
         if ( matchingFactoryMethods.size() > 1 ) {
             ctx.getMessager().printMessage(
                 method.getExecutable(),
                 Message.GENERAL_AMBIGIOUS_FACTORY_METHOD,
-                targetType.getEffectiveType(),
+                alternativeTarget,
                 Strings.join( matchingFactoryMethods, ", " ) );
 
             return null;
@@ -98,8 +117,7 @@ public class ObjectFactoryMethodResolver {
         }
     }
 
-    private static MethodReference findBuilderFactoryMethod(Type targetType) {
-        BuilderType builder = targetType.getBuilderType();
+    public static MethodReference getBuilderFactoryMethod(Method method, BuilderType builder ) {
         if ( builder == null ) {
             return null;
         }
@@ -110,7 +128,7 @@ public class ObjectFactoryMethodResolver {
             return null;
         }
 
-        if ( !builder.getBuildingType().isAssignableTo( targetType ) ) {
+        if ( !builder.getBuildingType().isAssignableTo( method.getReturnType() ) ) {
             //TODO print error message
             return null;
         }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
@@ -8,7 +8,6 @@ package org.mapstruct.ap.internal.model.source;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
-
 import javax.lang.model.element.ExecutableElement;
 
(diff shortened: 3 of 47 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/ObjectFactoryMethodResolver.java
```diff
@@ -58,18 +77,18 @@ public static MethodReference getFactoryMethod( Method method,
                 method,
                 getAllAvailableMethods( method, ctx.getSourceModel() ),
                 java.util.Collections.<Type> emptyList(),
-                targetType.getEffectiveType(),
+                alternativeTarget,
                 SelectionCriteria.forFactoryMethods( selectionParameters ) );
 
         if (matchingFactoryMethods.isEmpty()) {
-            return findBuilderFactoryMethod( targetType );
+            return null;
```
filiphr: Why return `null` here?

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
```diff
@@ -82,11 +81,11 @@ public ForgedMethod(String name, Type sourceType, Type returnType, MapperConfigu
      * @param mappingOptions the mapping options for this method
      * @param forgedNameBased forges a name based (matched) mapping method
      */
-    public ForgedMethod(String name, Type sourceType, Type returnType, MapperConfiguration mapperConfiguration,
-                        ExecutableElement positionHintElement, List<Parameter> additionalParameters,
-                        ParameterProvidedMethods parameterProvidedMethods, ForgedMethodHistory history,
-                        MappingOptions mappingOptions, boolean forgedNameBased) {
-        String sourceParamName = Strings.decapitalize( sourceType.getName() );
+     public ForgedMethod(String name, Type sourceType, Type returnType, MapperConfiguration mapperConfiguration,
```
filiphr: Are this formatting changes really needed?

processor/src/test/java/org/mapstruct/ap/test/builder/multiple/MultipleBuilderMapperTest.java
```diff
@@ -113,18 +113,7 @@ public void builderMappingMapperConfigDefined() {
         TooManyBuilderCreationMethodsMapper.class
     })
     @ExpectedCompilationOutcome(value = CompilationResult.SUCCEEDED,
-        // We have 2 diagnostics, as we don't do caching of the types, so a type is processed multiple types
         diagnostics = {
-            @Diagnostic(
```
filiphr: So we no longer report this warning?

core/src/main/java/org/mapstruct/Builder.java
```diff
@@ -29,4 +29,12 @@
...
      */
     String buildMethod() default "build";
+
+    /**
+     * Toggling builders on / off. Builders are sometimes used solely for unit testing (fluent testdata)
+     * MapStruct will need to use the regular getters /setters in that case.
+     *
+     * @return when true, no builder patterns will be applied
+     */
+    boolean noBuilder() default false;
```
filiphr: How about `disableBuilder`? For some reason `noBuilder` doesn't really stick. I am open to other suggestions as well.

filiphr:
I forgot to mention. This PR would also solve https://github.com/mapstruct/mapstruct/issues/1661, right?

sjaakd:
> I forgot to mention. This PR would also solve #1661, right?

it does..

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
sjaakd: No they aren't.. I'll remove them..

filiphr:
For #1661 let's also add an annotation processor option. So this would partially solve it.

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
sjaakd: I did a lot of trial and error here.. So I guess there was (in one of my commits) an extra param..

processor/src/main/java/org/mapstruct/ap/internal/model/ObjectFactoryMethodResolver.java
sjaakd: I changed the design here a bit.. If you look at the call site: 

```java
        private MethodReference getFactoryMethod(Type returnTypeImpl, SelectionParameters selectionParameters) {
            MethodReference factoryMethod = ObjectFactoryMethodResolver.getFactoryMethod( method,
                            returnTypeImpl,
                            selectionParameters,
                            ctx
            );
            if ( factoryMethod == null && returnTypeBuilder != null ) {
                factoryMethod = ObjectFactoryMethodResolver.getBuilderFactoryMethod( method, returnTypeBuilder );
            }

            return factoryMethod;
        }
```

You'll find that there's more to check.. Error handling for not-found methods takes place at the call sites of this method. At the other hand, when there's too many qualifying methods, we know its wrong anyway...

processor/src/test/java/org/mapstruct/ap/test/builder/multiple/MultipleBuilderMapperTest.java
sjaakd: correction: since we only resolve the builder once, we don't report the duplicate anymore :smile:

processor/src/main/java/org/mapstruct/ap/internal/model/source/ForgedMethod.java
filiphr: makes sense

processor/src/test/java/org/mapstruct/ap/test/builder/multiple/MultipleBuilderMapperTest.java
filiphr: I am not quite following. I think what the test is trying to show that if there are multiple builders (when we try to resolve) we are going to output a warning message. It is not about duplicates, there are 2 potential builder creation methods here.

filiphr: I see it now. There is another `@Diagnostic` with the exact same error

filiphr:
Thanks for the updates. LGTM for merging

chris922:
I just had a quick glance at your changes, looks awesome. Thanks @sjaakd !

PR #1814, opened 2019-05-13
Improve terms in qualifier docs

mattdrees:
This sentence is talking about `@Target`, not `@Retention`.
Also, let's use 'type' instead of 'class' to line up with `ElementType.TYPE`.


sjaakd:
@mattdrees .. thanks for correcting that error.

Issue #1819, opened 2019-05-16
Infinite Stack Trace Loop for Invocation Exception with Abstract Mapper


IDoCodingStuffs:
`java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
        at jdk.internal.reflect.GeneratedConstructorAccessor4.newInstance(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:488)
        at org.mapstruct.factory.Mappers.doGetMapper(Mappers.java:85)
        at org.mapstruct.factory.Mappers.getMapper(Mappers.java:69)
        at org.mapstruct.factory.Mappers.getMapper(Mappers.java:58)
        at com.____.OutboundRequestMapper.<init>(OutboundRequestMapper.java:19)
        at com.______.OutboundRequestMapperImpl.<init>(OutboundRequestMapperImpl.java:28)
        at jdk.internal.reflect.GeneratedConstructorAccessor4.newInstance(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:488)
        at org.mapstruct.factory.Mappers.doGetMapper(Mappers.java:85)
        at org.mapstruct.factory.Mappers.getMapper(Mappers.java:69)
        at org.mapstruct.factory.Mappers.getMapper(Mappers.java:58)
...
Same lines just keep repeating until memory is full
...
`

The line 19 for the OutboundRequestMapper is this:

`  OutboundRequestMapper INSTANCE = Mappers.getMapper(OutboundRequestMapper.class);
`
The line mentioned for the Impl class is just the class declaration itself. The rest are all from default libraries.

IDoCodingStuffs:
Removing the self referring INSTANCE takes care of the problem. So either those should not be allowed inside abstract mappers or whatever is causing the invocation exception should be fixed.

sjaakd:
So.. without any code your issue is a bit hard to grasp.. Can you add some reproducer?

IDoCodingStuffs:
`
@Mapper(componentModel = "spring")
public abstract class OutBoundRequestMapper implements .../*interfaces with some mappings*/... {
[...]

PR #1820, opened 2019-05-16
#1819

sjaakd:


chris922:
Changes looks fine for me!

Because you are working on this paragraph maybe you can add a hint that the instance must be statically defined after the following example ?

```java
@Mapper
public interface CarMapper {
    CarMapper INSTANCE = Mappers.getMapper( CarMapper.class );
    CarDto carToCarDto(Car car);
}
```

So that the users know they have to use `public static CarMapper INSTANCE = ...` when using a (abstract) class [maybe mention this as well]

sjaakd:
> Perfect, thanks!

Sure.. no problem. Thanks for reviewing..

PR #1827, opened 2019-05-22
#1826: fixed null pointer in nested property mapping when using prese…

### LINKED ISSUE #1826: Null pointer exception in nested property mapping method with presence check
Given the following object structures:
```
SourceParent                        Target
    |_ SourceChild                       |_Content
         |_ Content 
```
If you try to map `SourceParent->SourceChild->Content` to `Target->Content` and also are using presence tracking you will get a null pointer if `SourceChild` is present but null.  

The NestedPropertyMappingMethod.ftl is missing a null check before checking presence.  I have fixed this and will open a PR shortly.

jakraska:
Present but null fields could cause null pointer exceptions when mapping nested properties. 

Fixes #1826

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
@@ -12,7 +12,7 @@
     }
 <#list propertyEntries as entry>
     <#if entry.presenceCheckerName?? >
-    if ( !<@localVarName index=entry_index/>.${entry.presenceCheckerName}() ) {
+    if ( <#if entry_index != 0><@localVarName index=entry_index/> == null || </#if>!<@localVarName index=entry_index/>.${entry.presenceCheckerName}() ) {
         return ${returnType.null};
     }
     </#if>
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_1826/Issue1826Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_1826/Issue1826Mapper.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._1826;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface Issue1826Mapper {
+
+  Issue1826Mapper INSTANCE = Mappers.getMapper( Issue1826Mapper.class );
+
+  @Mapping(target = "content", source = "sourceChild.content")
+  Target sourceAToTarget(SourceParent sourceParent);
+
+}
--- a/copyright.txt
+++ b/copyright.txt
@@ -23,6 +23,7 @@ Gervais Blaise - https://github.com/gervaisb
 Gunnar Morling - https://github.com/gunnarmorling
 Ivo Smid - https://github.com/bedla
 Jeff Smyth - https://github.com/smythie86
+Jonathan Kraska - https://github.com/jakraska
 Joshua Spoerri - https://github.com/spoerri
 Kevin Grüneberg - https://github.com/kevcodez
 Michael Pardo - https://github.com/pardom
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_1826/Issue1826Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_1826/Issue1826Test.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._1826;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(AnnotationProcessorTestRunner.class)
+@IssueKey("1826")
+@WithClasses({
+        SourceParent.class,
+        SourceChild.class,
+        Target.class,
+        Issue1826Mapper.class
+})
+public class Issue1826Test {
+
+  @Test
+  public void testNestedPropertyMappingChecksForNull() {
+    SourceParent sourceParent = new SourceParent();
+    sourceParent.setSourceChild( null );
+
+    Target result = Issue1826Mapper.INSTANCE.sourceAToTarget( sourceParent );
+    assertThat( result.getContent() ).isNull();
+  }
+}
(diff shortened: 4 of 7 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_1826/Issue1826Mapper.java
```diff
@@ -0,0 +1,21 @@
...
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._1826;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.NullValueCheckStrategy;
+import org.mapstruct.factory.Mappers;
+
+@Mapper(nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS)
```
sjaakd: I think you need to add another test case (and possibly some additional coding) to test whether the `nullValueCheckStrategy` actually works.

processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
```diff
@@ -12,7 +12,7 @@
     }
 <#list propertyEntries as entry>
     <#if entry.presenceCheckerName?? >
-    if ( !<@localVarName index=entry_index/>.${entry.presenceCheckerName}() ) {
+    if ( <@localVarName index=entry_index/> == null || !<@localVarName index=entry_index/>.${entry.presenceCheckerName}() ) {
```
sjaakd: I guess the `||` should be `&&` in order to stay in line with other occurrences.

jakraska: If i change this to &&, it won't fix the null pointer.    This is the code that gets generated for the test case where `hasSourceChild` is `true` and `sourceChild` is `null`

```
 private String sourceParentSourceChildContent(SourceParent sourceParent) {
        if ( sourceParent == null ) {
            return null;
        }
        if ( sourceParent == null || !sourceParent.hasSourceChild() ) {
            return null;
        }
        SourceChild sourceChild = sourceParent.getSourceChild();
        if ( sourceChild == null || !sourceChild.hasContent() ) {
            return null;
        }
        String content = sourceChild.getContent();
        return content;
    }
```

processor/src/test/java/org/mapstruct/ap/test/bugs/_1826/Issue1826Mapper.java
jakraska: To be honest, I dont think the NullValueCheckStrategy property effects this code.  The source that gets generated here is identical without / without the strategy defined, so i can remove this from the test mapper definition if that helps.   I don't see any switches in the template that change the logic based on the  NullValueCheckStrategy setting.  What exactly would i be testing here?

processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
filiphr: Why `hasSourceChild` returns `true` when `sourceChild` is `null`?

filiphr: I missed the discussion in the issue before writing this. @jakraska explained everything in https://github.com/mapstruct/mapstruct/issues/1826#issuecomment-495920222. That is what I suspected as well.

sjaakd: > If i change this to &&, it won't fix the null pointer. This is the code that gets generated for the test case where `hasSourceChild` is `true` and `sourceChild` is `null`

Ok. You are right.. You break the flow (return) in stead of "setting some target".. Its de good all " de morgan"  again :smile:.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1826/Issue1826Mapper.java
sjaakd: I think that you should control the addition of the null check, having a possibility to opt-in. That would be  in line with our documentation. 

> I don't see any switches in the template that change the logic based on the NullValueCheckStrategy setting. What exactly would i be testing here?

I think that should be added to stay in line with the documentation and other places... It is present in the `PropertyMapping` and then `NullValueCheck` can be added to the creation of the backing bean of the template. BTW: this being a non update method it is the `NullValueCheckStrategy` you're after... not the `NullValuePropertyCheckStrategy`.

sjaakd: Hold on before you implement this.. Perhaps you're right.. I added also some comment to the issue on this.

filiphr: @sjaakd I am not sure that we need something special for this. This is only happening for extracting nested source value, if we add another option it might be overcomplicating it 😄

processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
filiphr: I have only one small remark about how the generated code looks like. The `sourceParent` is checked twice for different to `null`. Maybe we need to skip the second check. I think it can be done with an explicit check for `entry_index == 0`. WDYT @jakraska?

jakraska: @filiphr  - yes i agree with the concern about the duplicate null check.  I left it in there to keep the template logic / change as readable as possible, but i'm definitely happy to add that condition in if you prefer.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1826/Issue1826Mapper.java
sjaakd: 👍 ok.. Agreed.

sjaakd: @jakraska then my only remaining comment is to remove argument of the `@Mapper`..

jakraska: @sjaakd will do

jakraska: The extra parameters have been removed from the @mapper annotation.

processor/src/main/resources/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.ftl
jakraska: I updated the template.  the generated code now looks like this:
```
    private String sourceParentSourceChildContent(SourceParent sourceParent) {
        if ( sourceParent == null ) {
            return null;
        }
        if ( !sourceParent.hasSourceChild() ) {
            return null;
        }
        SourceChild sourceChild = sourceParent.getSourceChild();
        if ( sourceChild == null || !sourceChild.hasContent() ) {
            return null;
        }
        String content = sourceChild.getContent();
        return content;
    }
```

filiphr:
Latest changes look good to me. Thanks @jakraska. I would leave it to @sjaakd to do the merge (as he already did most of the review)

sjaakd:
@jakraska .. thanks for your contribution.

PR #1832, opened 2019-05-24
#1159 controlling errors thrown by the ServiceLoader iterator when loading AstModifyingAnnotationProcessors

jorjazo:


processor/src/main/java/org/mapstruct/ap/internal/util/AnnotationProcessorContext.java
```diff
@@ -106,20 +107,55 @@ else if ( elementUtils.getTypeElement( FreeBuilderConstants.FREE_BUILDER_FQN ) !
...
         return processors;
     }
 
+    protected boolean hasMoreAnnotationProcessors(Iterator<AstModifyingAnnotationProcessor> it) {
+        try {
+            return it.hasNext();
+        }
+        catch ( Throwable t ) {
+            if ( messager != null ) {
+                messager.printMessage( Kind.ERROR,
```
sjaakd: I guess this should be a note or warning, since you retry.. Right?

jorjazo: Hmm... I'm not retrying with the same element in the iterator. At this point there's at least one error with an element. I'm going to try to skip it, but it's not guaranteed. If trying to skip it fails, then there's another error message warning that some other elements might have failed.

filiphr:
Closing this since it was superseded by https://github.com/mapstruct/mapstruct/pull/2059

PR #1833, opened 2019-05-27
1406 targeting "." as current object

sjaakd:
Starting with a squashed version, rebased on top of the current master of https://github.com/mapstruct/mapstruct/pull/1686

Ready for review!


dainiusfigoras:
any plans for merge?

sjaakd:
> any plans for merge?

yes. But I want to simplify the solution a bit further to make it better maintainable (and make sure all scenarios are covered). It will take some more time.

dainiusfigoras:
thanks. I just wasn't sure if you planned to do it right away or take some more time for review.

sjaakd:
@dainiusfigoras I started working on this.. I spotted one problem where existing functionality does not work anymore. Try to add a property to `CustomerItem`.. e.g. a price.  I would expect that we get an unmapped target property. That does not happen.

From this,  I think it is a bit risky to make a 'special' method in `@BeanMapping`: `applySourcePropertiesToCurrentTarget`  covering this case. We run a risk on missing out on more functionality.  So I'll try to integrate it in the existing methods (as @filiphr also suggested). Any ideas on your side from on that?

I'll close your original PR..I don't think we documented this (but I think we should). We use the [rebase / fast forward model](https://medium.com/@mvuksano/git-tips-use-only-fast-forward-merges-with-rebase-c80c9d260a83) for commits (so no merge commits).  I spent some time on rebasing / getting the commit log clear in this one. 

Thanks in advance for any feedback you have.

sjaakd:
@daniel-shuy .. I saw your comment.. As you can see, I'm still working on this to make it fit to our current code. I'll continue soon, but before I'll do so, we must agree internally that this is the right approach.

Main problem is to keep all kinds of features compatible and working. For instance: ignoring source and source checking, consistent target ignore, etc. 

Also it needs to be embedded in a clear solution. Hence a lot of refactoring as well.

processor/src/test/java/org/mapstruct/ap/test/targetthis/ConflictingButIgnoredMapper.java
```diff
@@ -0,0 +1,23 @@
...
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface ConflictingButIgnoredMapper {
+
+    ConflictingButIgnoredMapper INSTANCE = Mappers.getMapper( ConflictingButIgnoredMapper.class );
+
+    @Mapping( target = ".", source = "customer.item" )
+    @Mapping( target = "id", source = "customer.item.id", ignore = true )
```
sjaakd: I'm really doubting whether we should do this.. have a way to ignore properties that occur in a next " target this " mapping.. To avoid conflicts the user should in such cases revert to fully mapping his intent. Lets consider target this as a convenience. Perhaps we should even allow it to be used only once in a set of mappings (at the other hand that might be too restricting).

filiphr: I am not really following. What shouldn't we allow? I don't see the need for the `source` here, since the ignore would apply on the target, so the source is obsolete. Or am I understanding something wrong?

sjaakd: The two sources are identical and contain both the same properties: id and status. You need to tell which one to ignore.

sjaakd: Or, alternatively you don't allow to create this situation by only allowing one "." Mapping

sjaakd: Or you do allow it, but if there's a conflict the user needs to resolve this with by fully specifying mappings the normal way

filiphr: We could allow the multiple `.`. That could be the same as multi source parameters. However, you should not define the `ignore`. You should define how you want to map the `id` and status. 

So if you want the status from the customer and the id from the customer then you do.

```
@Mapping( target = "status", source = "customer.item.status" )
@Mapping( target = "id", source = "item.id" )
```

If that is too complex then let's just allow one `.`

sjaakd: so.. that does not complete resolve the issue.. What if the user specifies:
```java
@Mapping( target = "status", source = "customer.item.status" )
@Mapping( target = ".", source = "item" )
```
?
Should I report an error (since the "." introduces a second mapping to status)?

Captain1653: Honestly, for me this point "." seems little strange. I don't know why, but... I'm not sure about my understanding this discussion, but I think to report error is better.

sjaakd: @Captain1653 .. what it in essence means is:
`Mapping( target = ".", source = "item.*" )`
so all the properties of item are mapped to the properties within target.

Captain1653: Honestly, for me this poit "." seems  little strange. I don't know why, but it is. I'm not sure about good understanding this situation, but I think to report an error is better.

Captain1653: @sjaakd  It's only my opinion, but it's not obvious for me.  I mean... I prefer explicit mapping. This approach is getting like SPEL from Spring. Too much magic  in the string, but maybe I'm wrong. Sure, it's only my opinion and I'll aprove it if you think we need it.

Captain1653: Next question about this  point ".". Should it work recursively? Which depth?

sjaakd: > t's only my opinion, but it's not obvious for me. I mean... I prefer explicit mapping. This approach is getting like SPEL from Spring. Too much magic in the string, but maybe I'm wrong. Sure, it's only my opinion and I'll aprove it if you think we need it.

It is a feature request and it does make kind of sense (it safes a lot of explicit mappings and is in essence name based mapping on nested level). So for instance: it is already possible to do this (if the target were to be nested as well)

```java
@Mapping( target = "item", source = "customer.item" ).
```

But I share your opinion that should limit its use. I guess it makes sense to only have one "." mapping since you cannot have multiple of the above like this..

```java
@Mapping( target = "item", source = "customer.item" ).
@Mapping( target = "item", source = "order.item" ).

```

Captain1653: Yes, you are right. Only one point. If we have more than one point then to throw error. It's seems good.

daniel-shuy: I think only allowing one `.` is too limiting. The use case of mapping a class with multiple nested classes to a fully flattened structure is not uncommon.

sjaakd: but it creates inconsistencies and stuff that is hard to explain in the api. So as always.. its a trade-off.

sjaakd:
I added a new test case myself, just to be sure that ignore also works. Unfortunately there's still an issue in the implementation. I'll add a commit for that.

processor/src/test/java/org/mapstruct/ap/test/targetthis/ConflictingButIgnoredMapper.java
filiphr: >so.. that does not complete resolve the issue.. What if the user specifies:
>
>```
>@Mapping( target = "status", source = "customer.item.status" )
>@Mapping( target = ".", source = "item" )
>```
>?
>Should I report an error (since the "." introduces a second mapping to status)?

If the user specifies this then status is already mapped and would be ignored when "." is used. I would say that "." should mean the rest of the unmapped properties. Since you might want to map certain properties in a different way.

All this means that there is no second mapping to status since "." means the rest (at least that's how I look at it conceptually).

Regarding multiple ".". Currently I see "." as a shortcut to saying mapping to the target. It doesn't have to be "." it can also be "<TARGET>", this is just an implementation detail.

In the end you could achieve the same with the current api by defining multiple source parameters.

sjaakd: The multiple source parameters is a compelling argument.

 Lets do it like this, first we merge this PR. 

Next we build upon the implementation and allow multiple ".". . It could be just as simple as changing the logic in mapper reference from one targetThis to a list of targetThis (and have some smart logic at the call site, but that's similar than the name based below.

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
```diff
@@ -55,18 +56,30 @@
      */
     public static class BuilderFromMapping {
 
-        private Mapping mapping;
+       // private Mapping mapping;
```
filiphr: Why is this commented it out? If we don't need it, why not remove it?

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
```diff
@@ -80,19 +93,23 @@ public BuilderFromMapping typeFactory(TypeFactory typeFactory) {
...
         public SourceReference build() {
 
-            String sourceName = mapping.getSourceName();
             if ( sourceName == null ) {
                 return null;
             }
 
             String sourceNameTrimmed = sourceName.trim();
-            if ( !sourceName.equals( sourceNameTrimmed ) ) {
+            if ( !sourceName.equals( sourceNameTrimmed ) && messager != null ) {
```
filiphr: How is it possible that the `messager` is `null`? We should always pass it, otherwise we can't print messages

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
```diff
@@ -206,12 +223,7 @@ private Parameter fetchMatchingParameterFromFirstSegment(String[] segments ) {
                 if ( parameter == null ) {
                     reportMappingError(
                         Message.PROPERTYMAPPING_INVALID_PARAMETER_NAME,
-                        parameterName,
-                        Strings.join(
-                            method.getSourceParameters(),
-                            ", ",
-                            Parameter::getName
-                        )
+                        parameterName, Strings.join( method.getSourceParameters(), ", ", Parameter::getName )
```
filiphr: The formatting is weird here. There are only whitespace changes here

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
```diff
@@ -384,4 +402,23 @@ public SourceReference pop() {
         }
     }
 
+    public List<SourceReference> push(TypeFactory typeFactory, Method method ) {
```
filiphr: Is `push` the right method name? I thought it changes the `SourceReference`, but it creates a new one it seems

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/TargetReference.java
```diff
@@ -56,55 +58,70 @@
...
+            Objects.requireNonNull( typeFactory );
 
-            String targetName = mapping.getTargetName();
             if ( targetName == null ) {
                 return null;
             }
 
             String targetNameTrimmed = targetName.trim();
-            if ( !targetName.equals( targetNameTrimmed ) ) {
+            if ( !targetName.equals( targetNameTrimmed ) && messager != null ) {
```
filiphr: Same here `messager` should never be `null`. Otherwise we can't report.

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/TargetReference.java
```diff
@@ -56,55 +58,70 @@
...
-            this.typeFactory = typeFactory;
+        public Builder method(Method method) {
+            this.method = method;
             return this;
         }
 
         public TargetReference build() {
 
-            boolean isInverse = mapping.getInheritContext() != null && mapping.getInheritContext().isReversed();
+            Objects.requireNonNull( method );
```
filiphr: Perhaps we can guarantee this by creating the builder with a constructor that takes `method`, `typeFactory` and `messager`?

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
```diff
@@ -475,9 +475,6 @@ public Mapping copyForForwardInheritance( SourceMethod method ) {
 
     @Override
     public boolean equals(Object o) {
-        if ( this == o ) {
```
filiphr: Why remove this? This makes is a shortcut when you check against the same instance.

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
```diff
@@ -586,11 +585,12 @@ public int compare(DiagnosticDescriptor o1, DiagnosticDescriptor o2) {
             if ( result != 0 ) {
                 return result;
             }
-            result = o1.getLine().compareTo( o2.getLine() );
-            if ( result != 0 ) {
-                return result;
+            if ( o1.getLine() != null && o2.getLine() != null ) {
```
filiphr: This is because the messager is `null`, right? So we don't output a line?

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
sjaakd: it does the opposite of pop.. So it adds an element. As a consequence it generates multiple new source references.

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
sjaakd: left over.. removed it.

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
sjaakd: When pushing the source reference I'm not interested in error messaging. It is part of processing not part of analysing the source model. I'm just interested if we have a valid result or not.

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
sjaakd: I adapter it and brought the Strings.join to the next line.. Better this way?

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/TargetReference.java
sjaakd: intentional.. See above

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
sjaakd: I keep on having problems with the eclipse lines / jdk lines because of this bug in javac. (annotations reported on the method). So for simple cases I thought we might enable to make line optional. I ran into NPE's, hence this fix.

it has nothing to do with the messager and I can revert if you want. In the end I did not use this, but I left it for the future.

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
sjaakd: Not sure whether this is needed still. There was a reason for removing this earlier.. I guess its a left over and I'll remove it again.

sjaakd:
> Looks really good @sjaakd. I added some comments for some null checks and some formatting.
> 
Thanks!

> Not sure whether we should add more tests where we explicitly define some mapping, not only just ignoring it.
> 

I tried to use as much as possible existing functionality. So if explicitly defining mappings works for name based mapping it should work here as well. And we know that code is triggered because of the ignore test case.

> I am also looking in the generated code and I am not 100% certain that I am happy with it. We generate a lot of nested source mappings. I know that you've spent a lot of time on this and perhaps we can do it as an enhancement, but we could generate a new method with nested source parameters.

I know, but that will probably open up a new can of worms: It has to interact with defined mappings. The moment we make something special for this feature, the complexity will increase considerably. Don't get me wrong: I'm in favour of more simple generated code. However, you cannot see it separate from other nested-source-mappings. We need to address that as an integral solution.

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
filiphr: I don't really like that. Even if you are not interested in the error messaging the conditions would always evaluate to `true`, right? Currently it is really easy to forget the messager and not log an error for that

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
filiphr: I know that bug, that's where there is the `alternativeLine` in the `@Diagnostic`. I've added it so we can workaround that bug

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
sjaakd: Duplicating all that code is also not an option

sjaakd: Duplicating all that code is also not an option

filiphr: Why not just pass the messager there as well? When doing the push there should be no errors in any way, right?

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/TargetReference.java
sjaakd: thought about that.. but abandoned that idea. It breaks the fluent api and we replace the construction of an object type problem (too many constructor args, often of the same type) to  the builder.

I don't see anything wrong with `Objects.requireNonNull` it was introduced in java7 and if you read the java doc especially for checking constructor and type args. 

The idea is to fail fast (and not somewhere inside a stream or fluent statement were you have to find which parameter actually caused the problem).

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
sjaakd: ah.. good point.. but still, a null check does not harm here..

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
sjaakd: ok.. resolved.

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/TargetReference.java
sjaakd: resolved

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingStatement.java
filiphr: Yes it doesn't harm, but it would hide the problem when you want to check something by just ignoring it

sjaakd: Ok.. I'l remove it..

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/TargetReference.java
filiphr: Fine for me, the reason why I don't like it is because it throw an NPE without any message. Although indeed you would know why that happened on that line. We should rethink the internals a bit and pass one object that would return all these utils everywhere instead of passing them all the time one by one :)

filiphr:
Thanks for the changes @sjaakd, LGTM

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/TargetReference.java
sjaakd: agreed..  I'll create a refactor issue for that..

filiphr: Thanks

PR #1840, opened 2019-06-18
Add support for InjectionStrategy.CONSTRUCTOR

reitzmichnicht:
When generating mappers with the setting InjectionStrategy.CONSTRUCTOR the `Mappers.getMapper()` call can not instantiate the mapper because it has no default constructor.
This adds recursive instantiation of mappers and its dependent used mappers.
This allows to instantiate mappers without relying on any DI framework.
This is very important for unit tests.

### DIFF
--- a/core/src/main/java/org/mapstruct/factory/Mappers.java
+++ b/core/src/main/java/org/mapstruct/factory/Mappers.java
@@ -82,20 +82,23 @@ public class Mappers {
             Class<T> implementation = (Class<T>) classLoader.loadClass( clazz.getName() + IMPLEMENTATION_SUFFIX );
             Constructor<?>[] constructors = implementation.getConstructors();
             Constructor<T> constructor;
-            if (constructors.length == 0) {
+            if ( constructors.length == 0 ) {
                 constructor = implementation.getDeclaredConstructor();
-                constructor.setAccessible(true);
-            } else {
+                constructor.setAccessible( true );
+            }
+            else {
                 constructor = (Constructor<T>) constructors[0];
             }
 
             Object[] params = Arrays.stream(constructor.getParameterTypes())
-                    .map(Mappers::getMapper)
-                    .toArray(Object[]::new);
-            return constructor.newInstance(params);
-        } catch (ClassNotFoundException e) {
+                    .map( Mappers::getMapper )
+                    .toArray( Object[]::new );
+            return constructor.newInstance( params );
+        }
+        catch ( ClassNotFoundException e ) {
             return getMapperFromServiceLoader( clazz, classLoader );
-        } catch (InstantiationException | InvocationTargetException | IllegalAccessException e) {
+        }
+        catch ( InstantiationException | InvocationTargetException | IllegalAccessException e ) {
             throw new RuntimeException( e );
         }
     }
--- a/core/src/main/java/org/mapstruct/factory/Mappers.java
+++ b/core/src/main/java/org/mapstruct/factory/Mappers.java
@@ -5,13 +5,14 @@
  */
 package org.mapstruct.factory;
 
+import org.mapstruct.Mapper;
+
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.ServiceLoader;
-
-import org.mapstruct.Mapper;
 
 /**
  * Factory for obtaining mapper instances if no explicit component model such as CDI is configured via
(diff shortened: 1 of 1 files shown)


filiphr:
Although this might be useful it might be dangerous to do this.

1. You can define a non MapStruct mapper in `Mapper#uses`. Try it and have a look at the generated code. If the used reference is a MapStruct mapper then `Mapper.getMapper()` is used. Otherwise the object is instantiated.
2. The `InjectionStrategy` is used only for DI `componentModel`, which means that we might be trying to instantiate classes that should not be instantiated by MapStruct.

>This is very important for unit tests.

If you really want to instantiate the mappers in tests and don't want to use your DI framework then I would say you need to do this manually by invoking the constructor in the implementation.

WDYT @sjaakd and @chris922

chris922:
Just for unit tests I would also prefer to instantiate the implementations itself or use the DI framework, for example within Spring there is out-of-the-box support for cases like this when you need a small injection/application context for running unit tests.

Besides that always using `getMapper` could lead to issues like you, @filiphr, already described with e. g. `uses` and not MapStruct-mappers
So if this approach will be continued than I would expect a few unit tests also including tests with `uses` and thus supporting non-MapStruct-mappers

ldziedziul:
Having this feature would make possible to instantiate mappers in unit tests without thinking about implementation details (i.e. generated classes).

What if you try to instantiate mapper using parameterized constructor only if all its parameters are annotated with `@Mapper`? This would be a compromise solution but still helping in many cases

reitzmichnicht:
If this PR has a chance to get merged I would update it so it can detect if a generated or provided mapper needs to be initiated.
Can you point me to the code where this detection is done during annotation progressing?

Starting a spring context is not an option for fast and lightweight unit tests.

filiphr:
>What if you try to instantiate mapper using parameterized constructor only if all its parameters are annotated with `@Mapper`? This would be a compromise solution but still helping in many cases

During annotation processing this is being done in the [`MapperCreationProcess#L137`](https://github.com/mapstruct/mapstruct/blob/119826982ab7a8d7d5a995ca80319992acca150f/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java#L137). However, you can't do the same during runtime since the `@Mapper` annotation has a `RetentionPolicy#CLASS`, which means that it is not present during runtime.

>Starting a spring context is not an option for fast and lightweight unit tests.

Have you tried starting a Spring Context with only 4 or 5 beans? It is extremely fast.

ldziedziul:
> During annotation processing this is being done in the [`MapperCreationProcess#L137`](https://github.com/mapstruct/mapstruct/blob/119826982ab7a8d7d5a995ca80319992acca150f/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java#L137). However, you can't do the same during runtime since the `@Mapper` annotation has a `RetentionPolicy#CLASS`, which means that it is not present during runtime.

Why not change it to `RetentionPolicy#RUNTIME` then? :)

> Have you tried starting a Spring Context with only 4 or 5 beans? It is extremely fast.

This would probably require some manual spring configuration instead of simple `Mappers.getMapper()`

sjaakd:
> If you really want to instantiate the mappers in tests and don't want to use your DI framework then I would say you need to do this manually by invoking the constructor in the implementation.
> 
> WDYT @sjaakd and @chris922

That's the way I do it.. Just write your own piece of reflection code / util.

ldziedziul:
I use something like the code below, but I imagine it could be supported by MapStruct itself

```java
public final class AdvancedMapperFactory {
    public enum InstantiationMode {ONLY_KNOWN_MAPPERS, DEFAULT}

    private AdvancedMapperFactory() {
    }

    public static <T> T getMapper(Class<T> clazz) {
        return getMapper(clazz, InstantiationMode.DEFAULT);
    }

    public static <T> T getMapper(Class<T> clazz, InstantiationMode instantiationMode) {
        Constructor<T>[] constructors = (Constructor<T>[]) Mappers.getMapperClass(clazz).getDeclaredConstructors();

        if (containsDefaultConstructor(constructors)) {
            return Mappers.getMapper(clazz);
        }

        assertSingleParameterizedConstructorPresent(constructors);

        Constructor<T> constructor = constructors[0];
        List<Object> args = new ArrayList<>();

        for (Parameter parameter : constructor.getParameters()) {
            if (allowedForInstantiation(parameter.getType(), instantiationMode)) {
                Object param = getMapper(parameter.getType(), instantiationMode);
                args.add(param);
            } else {
                throw new IllegalArgumentException("Constructor parameter " + parameter + " of " + clazz + " can not be instantiated");
            }
        }
        constructor.setAccessible(true);
        try {
            return constructor.newInstance(args.toArray());
        } catch (InstantiationException | InvocationTargetException | IllegalAccessException e) {
            throw new RuntimeException(e);
        }
    }

    private static boolean allowedForInstantiation(Class<?> clazz, InstantiationMode instantiationMode) {
        if (instantiationMode == InstantiationMode.ONLY_KNOWN_MAPPERS) {
            return MapStructMapper.class.isAssignableFrom(clazz);
        } else {
            return true;
        }
    }

    private static void assertSingleParameterizedConstructorPresent(Constructor<?>[] parameterizedConstructors) {
        long parameterizedConstructorsCount = Arrays.stream(parameterizedConstructors).filter(c -> c.getParameterCount() > 0).count();
        if (parameterizedConstructorsCount != 1) {
            throw new IllegalArgumentException("Cannot find a single parameterized constructor. Found: " + Arrays.toString(parameterizedConstructors));
        }
    }

    private static boolean containsDefaultConstructor(Constructor<?>[] parameterizedConstructors) {
        return Arrays.stream(parameterizedConstructors).anyMatch(c -> c.getParameterCount() == 0);
    }
}
```

sjaakd:
Perhaps we could support a unit test helper library or something along that line. I did write some code in the past that automatically populated dto's in an object graph, called a round-trip mapping (so forward-reverse) and asserted the result. It does not prevent you from making property wiring errors but covers the generated and handwritten code very well. That could be a candidate as well

filiphr:
>Why not change it to `RetentionPolicy#RUNTIME` then? :)

Since there is no need for the annotation to be retained during runtime. If it is kept there it would mean that all other runtime checks would introspect the `@Mapper` annotation and this would cause some kind of a performance penalty.

>This would probably require some manual spring configuration instead of simple `Mappers.getMapper()`

Since 1.3.0.Final there is the `Mappers#getMapperClass` method that allows you to retrieve the class for the mapper. This would make it easy to register custom classes in the custom Spring Context. Especially with the newer Spring 5 functional bean registration.

e.g.

```
GenericApplicationContext context = new GenericApplicationContext()
```

@ldziedziul your example is exactly what I said about you knowing your own code and being able to write your own factory. I see that in `allowedForInstantiation` you have `MapStructMapper`, which I presume is your own interface. Since MapStruct does not need to force people to implement some interface, there is no way for us to do something like you are doing.

filiphr:
Sorry for closing the PR, I pressed the wrong button. It is now reopened

filiphr:
After some long thinking we are not going to accept this PR.

One of the main reasons is that we don't want to change the retention policy on the `@Mapper` annotation (which is one of the only ways to achieve this).

One other reason is that it is not that difficult to do this within your own project based on the knowledge that you have about your mappers (E.g. all implement a common interface).

I am sorry about this @reitzmichnicht, I hope that you understand us and that we would work on some other contributions together in the future.

PR #1848, opened 2019-07-13
Cleanup

sjaakd:
work in progress


processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
```diff
@@ -61,25 +56,51 @@
...
-                List<Mapping> mappingsOfProperty = mappings.get( mappingPrism.target() );
-                if ( mappingsOfProperty == null ) {
-                    mappingsOfProperty = new ArrayList<>();
-                    mappings.put( mappingPrism.target(), mappingsOfProperty );
-                }
-
-                mappingsOfProperty.add( mapping );
-
-                if ( mappingsOfProperty.size() > 1 && !isEnumType( method.getReturnType() ) ) {
+                if ( !mappings.add( mapping ) ) {
```
filiphr: Maybe for making it more clear we should do

```
if ( mappings.contains(mapping) ) {
...
} else {
    mappings.add( mapping );
}
```

Otherwise you always need to read what the return of the `add` means. I think that it is easier to read it like this

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
```diff
@@ -166,7 +189,8 @@ public static Mapping forIgnore( String targetName) {
             null,
             null,
             null,
-            new ArrayList(),
+            new HashSet<>(),
```
filiphr: Since we are changing I think it is better to use `Collections.emptySet()`. Like in `fromMappingPrism`

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
```diff
@@ -556,20 +545,19 @@ public Mapping popSourceReference() {
...
 
-    public List<String> getDependsOn() {
-        return dependsOn;
+    /**
+     *  mapping can only be reversed if the source was not a constant nor an expression nor a nested property
+     *  and the mapping is not a 'target-source-ignore' mapping
+     *
+     * @return true when the above applies
+     */
+    public boolean canInverse() {
```
filiphr: Maybe we should name the method `canReverse`? The Javadoc already uses the `reverse` verb to explain what it means 😄.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
```diff
@@ -376,12 +349,6 @@ public boolean isMapMapping() {
         return isMapMapping;
     }
 
-    public boolean isEnumMapping() {
```
filiphr: Why not keep the method?

filiphr: I see that we are still doing a `isEnumMapping` check. So maybe we should remove everywhere and if someone is using `@Mapping` instead of `@Value` throw an error?

sjaakd: Good idea. I did not search actively for the check. I'll do so and throw an error at top level.

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
sjaakd: Agreed

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
```diff
@@ -61,25 +56,51 @@
...
+            return isForwarded;
+        }
+
+        public Method getInheritedFromMethod() {
+            return inheritedFromMethod;
+        }
+    }
+
+    public static Set<String> getMappingTargetNamesBy(Predicate<Mapping> predicate, Set<Mapping> mappings) {
+        return mappings.stream().filter( predicate ).map( Mapping::getTargetName ).collect( Collectors.toSet() );
```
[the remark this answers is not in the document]
sjaakd: Good point.. did not know that was possible

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
```diff
@@ -134,55 +134,45 @@ public static Parameter forForgedMappingTarget(Type parameterType) {
         );
     }
 
-    /**
+    /**ø
```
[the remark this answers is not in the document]
sjaakd: Will remove

processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java
sjaakd: I had long discussions on terminology in the past.. we settled on `inverse` lets make consistent with annotation..

sjaakd: I skipped the unit test from this action.. source code is consistent now.

sjaakd:
@filiphr I processed your comments.. Do you want to take another look or can I merge?

filiphr:
Looks good @sjaakd. I left one comment on the PR. The new message should say that it is removed I presume? Not that it is going to be.

Feel free to merge when you can

sjaakd:
@filiphr .. thanks for reviewing.

chris922:
Wow great work, thanks @sjaakd !

Issue #1851, opened 2019-07-23
Expression does not work with qualifiedBy


dekelintelematics:
According to the javadoc on `expression()`:
>This attribute can not be used together with source(), defaultValue(), defaultExpression() or constant().

Either the behaviour should be updated to allow use with `qualifiedBy()`, or the doc should be updated to show that it doesn't.

filiphr:
The reason for not mentioning that is because those methods `source()`, `defaultValue()`, `defaultExpression()`, `constant()` and `expression()` are used to control the source for the target.

The reason why it is not mentioned because `expression()` will take the value as is and everything else is ignored. It will not make sense to use `qualifiedBy()` here. Perhaps we can update the documentation on `qualifiedBy()` that it would only matter for `source()`. Would you like to prepare a PR for updating the Javadoc.

dekelintelematics:
Sure, I'll probably do that on the weekend, thanks for the communication @filiphr 

@dekelpilli

PR #1855, opened 2019-08-03
Clarifying qualifiedBy javadoc

dekelpilli:
Fixes https://github.com/mapstruct/mapstruct/issues/1851


filiphr:
@dekelpilli I just had a look at the places where the `qualifiers` are used and it seems that it can be used with `constant` as well. However, that is used when you a complex type as a target for the constant. For example 

```
public class Target {

    private String value;
    private List<String> names;
    private List<String> numbers;

    //getters and setters
}

public class Source {

    private String value;

    //getters and setters
}

@Mapper
public interface ConstantMapper {

    @Mapping(target = "names", constant = "Peter-Tom", qualifiedByName = "dashSeparated")
    @Mapping(target = "numbers", constant = "123-456,456-789", qualifiedByName = "commaSeparated")
    Target map(Source source);
    
    @Qualifier("dashSeparated")
    default List<String> fromDashSeparated(String value) {
        return Arrays.asList(value.split("-"));
    }

    @Qualifier("commaSeparated")
    default List<String> fromCommaSeparated(String value) {
        return Arrays.asList(value.split(","));
    }
}
```

Not sure how to word this in the javadoc

dekelpilli:
So, to clarify, are you saying this wouldn't work?

```
public class Target {

    private String targetValue;

    //getters and setters
}

public class Source {
...
}

@Mapper
public interface ConstantMapper {

    @Mapping(target = "targetValue", constant = "some string", qualifiedByName = "toUpper")
    Target map(Source source);
    
    @Qualifier("toUpper")
    default String fromDashSeparated(String value) {
        return value.toUpperCase();
    }
}
```

filiphr:
@dekelpilli that should work. The `target value` should then have a value set to `SOME STRING`

sjaakd:
@dekelpilli what is the status on this? can I close this one?

dekelpilli:
@sjaakd I think it's probably still worth clarifying the expression's interaction with qualifiers (or lack thereof). Maybe instead of being part of the `qualifiedBy` javadoc, it would be part of the `expression` javadoc. Maybe something like changing the current description of 

>This attribute can not be used together with {@link #source()}, {@link #defaultValue()},
     * {@link #defaultExpression()} or {@link #constant()}.

to

```
This attribute can not be used together with {@link #source()}, {@link #defaultValue()},
* {@link #defaultExpression()}, {@link #constant()}, {@link #qualifiedBy()}, or {@link #qualifiedByName()} 
```

filiphr:
I think what you are proposing would work @dekelpilli. Perhaps we can even have a compile error if you use `qualifiedBy` and `qualifiedByName` with `expression`.

Have a look at [here](https://github.com/mapstruct/mapstruct/blob/7e0327767f1a375a59b76ef2458a6394d8a5786e/processor/src/main/java/org/mapstruct/ap/internal/model/source/Mapping.java#L221) for an example. 

@dekelpilli are you interested in adapting this PR with this change?

dekelpilli:
@filiphr yup, doesn't seem too complex, I'll try to get it done during the week

dekelpilli:
Huh, didn't expect the CI to pass, was getting strange issues locally. Would you like me to squash the commits?

core/src/main/java/org/mapstruct/Mapping.java
```diff
@@ -186,6 +186,9 @@
      * A qualifier can be specified to aid the selection process of a suitable mapper. This is useful in case multiple
      * mapping methods (hand written or generated) qualify and thus would result in an 'Ambiguous mapping methods found'
      * error. A qualifier is a custom annotation and can be placed on a hand written mapper class or a method.
+     * <p>
+     * Qualifiers only apply to the {@link #source()}, and cannot be used to qualify {@link #constant()}
+     * and {@link #expression()} values.
```
filiphr: Didn't we say that we would remove this and adapt the `expression()` with the new text?

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -57,6 +57,7 @@
     PROPERTYMAPPING_CONSTANT_VALUE_AND_NVPMS( "Constant and nullValuePropertyMappingStrategy are both defined in @Mapping, either define a constant or an nullValuePropertyMappingStrategy." ),
     PROPERTYMAPPING_DEFAULT_EXPERSSION_AND_NVPMS( "DefaultExpression and nullValuePropertyMappingStrategy are both defined in @Mapping, either define a defaultExpression or an nullValuePropertyMappingStrategy." ),
     PROPERTYMAPPING_IGNORE_AND_NVPMS( "Ignore and nullValuePropertyMappingStrategy are both defined in @Mapping, either define ignore or an nullValuePropertyMappingStrategy." ),
+    PROPERTYMAPPING_EXPRESSION_AND_QUALIFIER_BOTH_DEFINED("Expression and a qualifier both defined in @Mapping, either defined an expression or a qualifier."),
```
filiphr:
```suggestion
    PROPERTYMAPPING_EXPRESSION_AND_QUALIFIER_BOTH_DEFINED("Expression and a qualifier both defined in @Mapping, either define an expression or a qualifier."),
```

filiphr:
>Huh, didn't expect the CI to pass, was getting strange issues locally. Would you like me to squash the commits?

What kind if issues did you have locally? Don't worry for the commits. We will squash them when we merge

dekelpilli:
> What kind if issues did you have locally? Don't worry for the commits. We will squash them when we merge

There were a few:
1. I had to comment out the argline option in the pom when running tests from intellij 
(this one: `<argLine>@{jacocoArgLine} -Xms1024m -Xmx3072m</argLine>`)
2. `FullFeatureCompilationTest` always failed for me, both from the command line (using `mvn clean install -DprocessorIntegrationTest.toolchainsFile=etc/toolchains-travis-jenkins.xml -B -V` as per the .travis file) and intellij. When I couldn't find any existing mappers that violate the new rule I pushed in case it was just local, but I thought I probably missed something.

core/src/main/java/org/mapstruct/Mapping.java
dekelpilli: Yup, you're right, I did it then stashed it and forgot to put it back. My latest commit should fix this and the typo below.

Captain1653:
@dekelpilli For runnning tests locally I use: mvn clean install  -Dmaven.multiModuleProjectDirectory=.
in the root folder with project. Maybe it will help you. And I also comment parameter jacoco in pom :) but you always can add additional commit in your PR before merge and then use "Squash and commit" :) .

filiphr:
For the `argLine` you can disable IntelliJ using it by unchecking `argLine` in Build, Execution, Deployment -> Build Tools -> Maven -> Running Tests

filiphr:
Thanks for the PR @dekelpilli I integrated it.

Would you mind sharing your name so we can add it to the [copyright.txt](https://github.com/mapstruct/mapstruct/blob/master/copyright.txt)

dekelpilli:
@filiphr No problem. My name is Dekel Pilli

PR #1859, opened 2019-08-11
Fix minor warnings

Captain1653:
Remove unnecessary generic type in Collections.empty*,
replace "for" on Stream API,
replace anonymous classes on lambda

I'm sorry, but I have added "dist: trusty" in travis.yml . I can't pass checks without it. I'll remove it if you will help me to pass checks without it.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -17,9 +17,11 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import javax.lang.model.type.DeclaredType;
 import javax.tools.Diagnostic;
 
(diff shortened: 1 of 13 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1051,24 +1050,19 @@ public MethodReference getFinalizerMethod() {
     }
 
     public List<Parameter> getSourceParametersExcludingPrimitives() {
-        List<Parameter> sourceParameters = new ArrayList<>();
-        for ( Parameter sourceParam : getSourceParameters() ) {
-            if ( !sourceParam.getType().isPrimitive() ) {
-                sourceParameters.add( sourceParam );
-            }
-        }
+        Predicate<Parameter> excludePrimitivePredicate = parameter -> !parameter.getType().isPrimitive();
```
filiphr: I am not sure how I feel about adding this predicate like this here and using `filterSourceParameters`. Perhaps if we remove the `filterSourceParameters` and do `return getSourceParameters().filter( param -> !param.getType().isPrimitive ).collect( Collectors.toList() );` instead. It would be more readable.

Same comments is valid for the other places where we pass a `Predicate`

Captain1653: @filiphr You are right. I have done it

filiphr:
Thanks for the PR @Captain1653. I've merged the PR

PR #1860, opened 2019-08-11
Simplify conditions in org.mapstruct.ap.internal.model.source

Captain1653:


processor/src/main/java/org/mapstruct/ap/internal/model/source/MethodMatcher.java
```diff
@@ -128,9 +128,7 @@ private boolean matchSourceType(Type sourceType,
                 if ( sourceType.isPrimitive() ) {
                     // the candidate source is primitive, so promote to its boxed type and check again (autobox)
                     TypeMirror boxedType = typeUtils.boxedClass( (PrimitiveType) sourceType.getTypeMirror() ).asType();
-                    if ( !parameterMatcher.visit( candidateSourceType.getTypeMirror(), boxedType ) ) {
-                        return false;
-                    }
+                    return parameterMatcher.visit( candidateSourceType.getTypeMirror(), boxedType );
```
filiphr: I know that logically the changes in this method are correct due to the last `return true`. However, the method is already quite complex and big and it might make it more complex to follow certain changes to it. Or in case the default or some logic before it changes.

Captain1653: @filiphr There are three similar changes in this class. Should I revert all of them or only one? All of them, right?)))

filiphr: Yup I was thinking about all of them.

Captain1653: Done

filiphr:
Thanks again @Captain1653 😄

PR #1863, opened 2019-08-12
Fix minor warnings

Captain1653:
Fix minor warnings:
remove unnecessary generic type for collections,
replace Charset.forName on StandartCharsets


Captain1653:
@filiphr 
I think you are right about size of ArrayList and expansions. I can replace it with custom supplier or revert this changes.

P.S. From javadoc ArrayList java 8:

> An empty ArrayList with elementData == DEFAULTCAPACITY_EMPTY_ELEMENTDATA will be expanded to DEFAULT_CAPACITY when the first element is added.

filiphr:
I think that for the time being it is better to leave the for loops as is. 

Not following what you mean about the javadoc.

Issue #1867, opened 2019-08-17
model.source should reflect source only and not (intermediate) creation state


sjaakd:
currently the `Mapping` contains source / target references. Also new `Mapping` is generated (with source / target) while not actually present in the code.

The intention of the source model was (one day) to abstract the annotations present in the source model. Currently it has moved beyond that and (particularly) `Mapping` but also `ForgedMethod` reflects something that is not present in the source code itself but generated in the `MapperCreationProcessor`.

Particularly prior to solving #1406 this needs to be straighten out.

PR #1868, opened 2019-08-17
#1867 refactor make the source model only a reflection of the source

sjaakd:


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -844,20 +794,22 @@ private void applyParameterNameBasedMapping() {
...
-            return additionalOptions;
+            return null;
         }
 
         private Accessor getTargetPropertyReadAccessor(String propertyName) {
             return method.getResultType().getPropertyReadAccessors().get( propertyName );
         }
 
         private ReportingPolicyPrism getUnmappedTargetPolicy() {
+            if ( mappingReferences.isForForgedMethods() ) {
```
sjaakd: Apparently `isRestrictToDefinedMethods` is not the same as `isForForgedMethods`..

This part was copied form the `MappingOptions` (effectively setting the reporting policy to IGNORE via an empty `@BeanMapping` and the solution did not work when I did not introduce `isForForgedMethods`. 

I would expect `isForForgedMethods` and `isRestrictToDefinedMethods` to reflect the same state from a usage perspective: when restricting to defined methods, we should not report on remaining properties that could have been handled based on their name.

In short: I do not understand this piece of logic. @filiphr : can you shed some light on this?

filiphr: The reason why you need to do this is because now `method.getMappingOptions()` would return the mapping options on from the `basedOn` method.

sjaakd: > The reason why you need to do this is because now `method.getMappingOptions()` would return the mapping options on from the `basedOn` method.

I understand that part.. But why (on a more fundamental level) is this different than isRestrictedToDefinedMethods.? Logically, you would not do error handling (override IGNORE in `method.getMappingOptions()`)  you would expect to do this when you `restrictToDefinedMethods` because you don't want to report errors on 'missed' name based mappings.

filiphr: For the `ForgedMethod` we are crafting new `MappingOptions`. It doesn't make sense to reuse the `MappingOptions` from the `basedOn` method. I think that `restrictToDefinedMethods` was added to support the complex target nested mappings. This is not exposed over annotations.

filiphr:
@sjaakd I am not sure about this one. More precisely I don't know if it would make it easier to understand.  

I get the idea of having the classes in the `model.source` package to reflect the actual source. The newly crafted `Mapping` for the nested target in a way it does represent the source. Since in the source you define `@Mapping( target = "nested.target", source = "source" )`. 

We create a forged method for the nested target instead of the user doing that on their own.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
sjaakd: The `MappingOptions` filter information in hierarchy. I can agree that `@BeanMapping` is not relevant, because generated methods can potentially be shared between defined mapping methods. But stuff in `@Mapper`, `@MappingConfig` is relevant..

You could even make a case for trickling down `@BeanMapping` stuff to generated methods. Think about qualifiers for before/after mappings. It is a bit tricky, because that would mean that a forged method cannot be shared if the `@MeanMapping` is different on some level.

sjaakd: `@MeanMapping` == of course `@BeanMapping`.. but it can be mean :smile:

filiphr: We do share the stuff from `@Mapper`, `@MappingConfig`, since we get those from the `MapperType`. For trickling down `@BeanMapping`, yes you can make a case (and I think that we have an issue to do that for `ForgedMethods`. However, you can't trickle down everything. `BeanMapping#resultType` for example. 

You can even argue that if someone wants to define some things specific for some forged methods that this should be either done via `@Mapper` or maybe they should define this method in their mapper.

sjaakd:
> @sjaakd I am not sure about this one. More precisely I don't know if it would make it easier to understand.

It is a separation of concerns thing. I did not realise this at first (when we made the nested mappings in the past). But the `model.source` package classes to reflect stuff that is in the actual annotations and should always do (in fact be read-only for the creation processor). So, for instance, the target / source in the method always reflect the original target source and is not adapted to the mapping level.

We should, I think make in our packaging structure, make that clear. Don't blur the reflection of the original source code, with stuff that we are going to generate. That's what happened.

When I wrote comments on #1406, that was one of the more fundamental remark only to find out that we violated these rules ourselves in the past.

I do think that clear rules on separation of concerns  is going to make a difference when  it comes to people that contribute.

---
Having said that.. I think that naming deserves attention. I'm not so happy with the whole `XYZReference` stuff.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
sjaakd: > However, you can't trickle down everything. `BeanMapping#resultType` for example.

Correct. And to be honest, I don't think we should do this.. (hand down `@BeanMapping`).. If there's an issue on that, we should answer it with this argument that forged methods can be shared at some level, and its hard to see which method originally created it.

filiphr:
>It is a separation of concerns thing. I did not realise this at first (when we made the nested mappings in the past). But the `model.source` package classes to reflect stuff that is in the actual annotations and should always do (in fact be read-only for the creation processor). So, for instance, the target / source in the method always reflect the original target source and is not adapted to the mapping level.

I understand that. I actually like the current `Mapping` in the `model.source`. Perhaps we need to move that to model and allow using it in other places. `MappingOptions` is not really a reflection of the source, it is more like a bean containing a grouping of some parameters that are used for the mapping then. We also have the different `XXXPrism` which also represent the actual annotations from the source code.

I always saw the `Mapping` as an internal representation for defining the mappings and the nested target mappings are doing just that, reusing a mechanism that we already have to create them. I imagined this as creating a new mapper with those `@Mapping` and then from that creating the final one 😄

sjaakd:
> I always saw the `Mapping` as an internal representation for defining the mappings and the nested target mappings are doing just that, reusing a mechanism that we already have to create them. I imagined this as creating a new mapper with those `@Mapping` and then from that creating the final one

I realised that `@Mapping` is not altogether different than `@BeanMapping` or `@Mapper`. You are right.. there are prisms, but they do not do any internal coherence checking (what we do in the classes `Mapping`, `BeanMapping` and `Mapper`.   The `MappingOptions` group the applicable annotations together for a certain method (it also does this for map-mappings, interable-mappings).

PR #1871, opened 2019-08-17
Fix minor warnings in test package bugs

Captain1653:
Fix minor warnings in test package bugs:

remove unnecessary generic type for collections,
remove unnecessary exceptions from signature test methods


processor/src/test/java/org/mapstruct/ap/test/bugs/_865/Issue865Test.java
```diff
@@ -41,7 +41,8 @@ public void shouldGenerateNpeCheckBeforCallingAddAllWhenInUpdateMethods() {
         assertThat( entity.getCoreUsers() ).isNull();
     }
 
-    public void shouldGenerateNpeCheckBeforCallingAddAllWhenInUpdateMethodsAndTargetWithoutSetter() {
+    @Test
```
filiphr: Nice spot

PR #1872, opened 2019-08-18
Simplify consistent checks in class Mapping

Captain1653:
Simplify consistent checks in class Mapping

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/dependency/GraphAnalyzer.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/dependency/GraphAnalyzer.java
@@ -116,10 +116,10 @@ public class GraphAnalyzer {
         private final Map<String, Node> nodes = new LinkedHashMap<>();
 
         public GraphAnalyzerBuilder withNode(String name, Set<String> descendants) {
-            Node node = getNode( name );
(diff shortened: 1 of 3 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -66,14 +67,12 @@ public TypeSelector(TypeFactory typeFactory, FormattingMessager messager) {
             List<List<ParameterBinding>> parameterBindingPermutations =
                 getCandidateParameterBindingPermutations( availableBindings, method.getMethod().getParameters() );
 
-            if ( parameterBindingPermutations != null ) {
```
filiphr: This makes sense from a programming perspective. However, in our case `getMatchingParameterBinding` is quite a complex method that would execute complex logic based on `parameterBindingPermutations`. Therefore I would revert the changes to the `TypeSelector` class.

Captain1653: Ok

Captain1653:
> Thanks for the PR @Captain1653. I left a comment about the change in the `TypeSelector`. In my opinion we shouldn't be changing that one

I have chanched

Issue #1877, opened 2019-08-24
Replace @Mapper.componentModel from string to enum


Captain1653:
@sjaakd @filiphr When I use Mapstruct on my work sometimes I can make mistake, when I write @Mapper.componentModel as string. Maybe it will be better to replace it on enum? I have prepared draft PR. If you think we need it, then I will fix documentation and ext.

filiphr:
@Captain1653 the reason why it is a `String` is because we want to allow people to provide their own [`AnnotationBasedComponentModelProcessor`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/processor/AnnotationBasedComponentModelProcessor.java) and use their own strings.

The IntelliJ plugin (not sure about the eclipse one) allows autocomplete for the known `componentModel`(s).

Last thing is that if we replace it with an enum this would be a breaking change and this is something that we don't really want to do.

Captain1653:
Ok, I agree with it.

PR #1878, opened 2019-08-24
#1877 Replace @Mapper.componentModel from string to enum

Captain1653:
I'll fix documentation, if you think this PR may be useful. I should comment some plugin in pom.xml, because it looks like breaking change. What dou you think about it?


filiphr:
See my comment in https://github.com/mapstruct/mapstruct/issues/1877#issuecomment-524543310. I would not add this functionality

PR #1879, opened 2019-08-24
#991 Split reference guide source into an adoc file per chapter

Captain1653:


Captain1653:
> @Captain1653 : did you try to generate the PDF / HTML? Is there a diff? Might be worthwhile to compare the results of old-and-new on PDF .. I think its even possible in acrobat reader.

I have compared it manually, but you are right. I'll attach comparasion's report

sjaakd:
> I have compared it manually, but you are right. I'll attach comparasion's report

Adding a report  is not really necessary.. If you verified it, I believe you :smile:

Captain1653:
Unfortunately, I don't believe myself ).
So, I have verified it with services: [textcompare](https://www.textcompare.org/pdf/)  and  [diffchecker](https://www.diffchecker.com/diff). 

I also attach two files. Maybe somebody know some good tool or service for verify.

[original.pdf](https://github.com/mapstruct/mapstruct/files/3542526/original.pdf)
[refactor-guide.pdf](https://github.com/mapstruct/mapstruct/files/3542527/refactor-guide.pdf)

sjaakd:
I tried this one: https://draftable.com/compare/tEHXpvwhylSi

Seems they are identical .. Please go ahead and merge.

Issue #1792, opened 2019-04-12
Annotation processor option for default injection strategy


reitzmichnicht:
- [x] Is this an issue (and hence not a question)? 

Please provide a general annotation processor option to configure the default injection strategy globally.
There is already
mapstruct.defaultComponentModel
please add
mapstruct.defaultInjectionStrategy
which is not documented as default option yet.

Kind regards,
Michael

PR #1880, opened 2019-08-24
#1792 Annotation processor option for default injection strategy

Captain1653:


documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -293,6 +293,17 @@ Supported values are:
 If a component model is given for a specific mapper via `@Mapper#componentModel()`, the value from the annotation takes precedence.
 |`default`
 
+|`mapstruct.defaultInjectionStrategy`
+| The type of the injection in mapper via parameter `uses`.
+
+Supported values are:
+
+* `field`: dependencies will be injected in fields
+* `constructor`: dependencies will be injected via constructor.
```
sjaakd: perhaps state that a constructor will be generated for this. I was thinking last week though: it should also generate an empty constructor for CDI (does it?), otherwise the object cannot be proxied.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
```diff
@@ -293,6 +293,17 @@ Supported values are:
...
+
+If a injection strategy is given for a specific mapper via `@Mapper#injectionStrategy()`, the value from the annotation takes precedence.
```
sjaakd: what do you mean here?

Captain1653: I wrote it like with component model from our documentation:
"If a component model is given for a specific mapper via @Mapper#componentModel(), the value from the annotation takes precedence." . Was I wrong?

Captain1653:
> Probably add a CDI example.

Yes, you are right. I'll do it, it's my fault.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
Captain1653: You're right. componentModel=cdi will generate two constructors: with inject value and default constructor without parameters. I'll add it in documentation.

Captain1653: @sjaakd I'm trying to create tests with CDI, but I can't. Maven (I don't know why) doesn't run my simple 'bad" test. But it compiled it (because I had compile errors with bad imports). I have attached picture. Maybe you know the reason? I use: mvn clean install . Tets 
cdiTest

sjaakd: It is more tricky than that. I tried to the same with `weld-unit` a while ago. Point is: we use a test runner `@RunWith` to trigger annotation processing and we control which classes we want on the class path (we can extend that with `@WithClasses`. Checkout the `testutil`  package for the details.

So, to use another `@RunWith` will not trigger annotation processing. Hence the implementation classes you need will never be generated. 

To solve this, you probably need to use [junit5](https://junit.org/junit5/docs/current/user-guide/#extensions). I believe that @filiphr has tried that. But that's already a tricky job. We probably need the ` @ExtendWith` to do that and use `Order` to first trigger annotation processing with our own runner and then have weld discover the generated implementations.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
sjaakd: No you are not wrong.. But perhaps add over what it takes precedence. So: `...takes precedence over the option`. Although it should probably be clear from the context as well.

documentation/src/main/asciidoc/mapstruct-reference-guide.asciidoc
Captain1653: Ow... I understand. So, about this PR. I think it's better to wait and don't merge it without tests CDI. Should I close?

sjaakd: by the way: I gave up on writing a unit test for that one for now. It is too tricky. But if you want a unit test, you could also go for a fixture. Checkout the way `@Rule` is used in `ConstantsTest` [here](https://github.com/mapstruct/mapstruct/blob/master/processor/src/test/java/org/mapstruct/ap/test/source/constants/ConstantsTest.java).

sjaakd: > Should I close?

No.. I think your solution is sound. Lets wait for @filiphr's review.

documentation/src/main/asciidoc/chapter-2-set-up.asciidoc
```diff
@@ -254,6 +254,18 @@ Supported values are:
...
 
+|`mapstruct.defaultInjectionStrategy`
+| The type of the injection in mapper via parameter `uses`.
+
+Supported values are:
+
+* `field`: dependencies will be injected in fields
+* `constructor`: will be generated constructor. Dependencies will be injected via constructor.
+
+For `componentModel()=cdi` also will be generated default constructor.
```
filiphr: Perhaps reword this to "When cdi `componentModel` a default constructor will also be generated"

documentation/src/main/asciidoc/chapter-2-set-up.asciidoc
```diff
@@ -254,6 +254,18 @@ Supported values are:
 If a component model is given for a specific mapper via `@Mapper#componentModel()`, the value from the annotation takes precedence.
 |`default`
 
+|`mapstruct.defaultInjectionStrategy`
+| The type of the injection in mapper via parameter `uses`.
```
filiphr: I think that we should mention somewhere that this is only is only used on annotated based component models such as CDI, Spring and JSR 330. Similar as we don in the Javadoc of `Mapper#injectionStrategy`

documentation/src/main/asciidoc/chapter-2-set-up.asciidoc
Captain1653: Done

documentation/src/main/asciidoc/chapter-2-set-up.asciidoc
Captain1653: Done

filiphr:
Thanks for the changes @Captain1653. Just squashed and merged it

Issue #1883, opened 2019-09-01
Remove not used method "asCollectionOrMap" from TypeFactory


Captain1653:
I'm not sure. Do we really need public method  TypeFactory.asCollectionOrMap ? I mean, I have deleted it, but tests are "green".   Or are we going to use this method in the future? Or should I  increate  our test coverage for this method?

PR #1884, opened 2019-09-01
#1883 Remove not used method "asCollectionOrMap" from TypeFactory

Captain1653:


Captain1653:
@sjaakd Thank you for review

Issue #724, opened 2016-01-17
Remove JaxbMapper from integration tests


gunnarmorling:
Hey @sjaakd, there is a comment in this class about removing it. Seems that's doable now?

PR #1892, opened 2019-09-10
#724 Remove JaxbMapper from integration test

Captain1653:


Captain1653:
@sjaakd Thank you for review

Issue #1897, opened 2019-09-11
Remove unused methods from class SourceMethod


Captain1653:
There are some unused methods in class org.mapstruct.ap.internal.model.source.SourceMethod. I've deleted it, but tests are green. Do we need these methods?

PR #1898, opened 2019-09-11
#1897 Remove unused methods from class SourceMethod

Captain1653:


Captain1653:
@sjaakd Thank you for review

PR #1900, opened 2019-09-12
Fix mapstruct adder finder (#1899)

### LINKED ISSUE #1899: Mapstruct ADDER_PREFERRED with more than one field that can use adder not working properly

If there is only one adder method, the name of the adder is not will be compared with property.

Below is a snippet with **two collections** and **ADDER_PREFERRED** option on the mapper and only **one adder method**

```
class Source {
    List<String> elements;
    List<String> secondElements;

    // ... standard getters and setters
}

class Target {
    List<String> elements;
    List<String> secondElements;

    public void addSecondElement(String element) {
        secondElements.add( element );
    }

    // ... standard getters and setters 
}
```
Elements from both collections will be in a secondElements collection after mapping.

**Also when I fixed this, I found that the previously written functionality was broken** (testShouldFallBackToDaliSingularInAbsenseOfHumanSingular test failed). The bug I fixed hid the problem.
So I fixed this problem also in accordance with the test.

DNAlchemist:
Fixes #1899

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/Nouns.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/Nouns.java
@@ -88,8 +88,8 @@ public class Nouns {
     }
 
     /**
-     * Converts given pluralized noun into the singular human form. If no singular form could be determined, the given word
-     * itself is returned.
+     * Converts given pluralized noun into the singular human form. If no singular form could be determined,
+     * the given word itself is returned.
      *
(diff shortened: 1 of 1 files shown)


filiphr:
@DNAlchemist as I mentioned in https://github.com/mapstruct/mapstruct/issues/1899#issuecomment-531592900. Your change makes sense. However, perhaps we should rethink whether we should even fallback to the single adder if it doesn't match the singular form.

PR #1902, opened 2019-09-12
Allow suffix "list" for properties

### LINKED ISSUE #1901: List mapping conversion with ADDER_PREFERRED strategy should be compatible with Protocol Buffers

Mapstruct can't find adder when working with generated protobuf classes 

*option ADDER_PREFERRED is enabled in mapper*

This proto file
```
message Request {
    repeated string book;
}
```
after protobuf processing having field *bookList*, and adder *addBook*.

**We need to use addBook when converting to protobuf**

Conversion is working fine, while you have only one adder in class.
In this case, it takes the first one, this is another issue #1899 .

If you have more than one repeated field in the proto file, it starts trying to match adder by name, instead of taking the first one.

DNAlchemist:
Fix #1901 
Depends on #1900 PR

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_1901/Issue1901Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_1901/Issue1901Mapper.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright MapStruct Authors.
+ *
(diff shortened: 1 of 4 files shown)


filiphr:
@DNAlchemist if we are going to support protobuf style adders out of the box I would say that we do it with a custom `AccessorNamingStrategy`. 

In this custom strategy we can easily detect if we are in a protobuf class and have a custom implementation for `getPropertyName` that would return the property name without `List`

PR #1903, opened 2019-09-12
Refactoring of BeanMapping and Source/TargetReferences (common base class)

sjaakd:
This PR is some general cleanup and refactoring I did as part of #1406, but is not really part of the business of 1406. So I split it off to make it more manageable.


sjaakd:
@Captain1653 thank for reviewing..

PR #1910, opened 2019-09-21
Compile error when deferred mapper was not created

### LINKED ISSUE #1904: Mappers using @Context not yielding Impl classes in Java11
While transitioning from Java 8 to Java 11 I found that any mappers that were using the `@Context` annotation stopped generating Impl classes, leading to "Cannot find implementation" errors at startup or test time. I've been able to reproduce using the CarDto mapper outlined in the reference guide:

```java
@Mapper
public interface CarMapper {
    default CarManualDto translateManual(CarManual manual, @Context Locale locale) {
        return new CarManualDto(manual.getContent());
    }

    CarDto toCar(Car car, @Context Locale locale);
}
```

When compiled using Java 8, we get the following Impl.java:
```java
@Generated(
    date = "2019-09-13T13:02:01-0500",
    comments = "version: 1.3.0.Final, compiler: javac, environment: Java 1.8.0_121 (Oracle Corporation)"
)

    @Override
    public CarDto toCar(Car car, Locale locale) {
        if ( car == null ) {
            return null;
        }

        CarDto carDto = new CarDto();

        carDto.setManual( translateManual( car.getManual(), locale ) );

        return carDto;
    }
}
```

When compiled using Java 11+ (with a source and target of 1.8), this Impl.java file is no longer being generated (using the same minimal maven pom):

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>

  <artifactId>context-bug</artifactId>
  <version>1.0-SNAPSHOT</version>

  <dependencies>
    <dependency>
      <version>1.3.0.Final</version>
    </dependency>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.12</version>
      <scope>test</scope>
    </dependency>
  </dependencies>

  <build>
    <plugins>
      <plugin>
        <version>3.8.0</version>
        <configuration>
          <source>1.8</source>
          <target>1.8</target>
          <annotationProcessorPaths>
            <path>
              <version>1.3.0.Final</version>
            </path>
          </annotationProcessorPaths>
        </configuration>
      </plugin>
    </plugins>
  </build>
</project
```

Note that this issue *may* have started earlier than Java 11, but I've only tested with Java 11 and Java 12.

filiphr:
Additionally exclude private methods when fetching enclosed executable elements

Fixes #1904


processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
```diff
@@ -85,8 +85,8 @@ private static TypeElement asTypeElement(TypeMirror mirror) {
 
     /**
      * Finds all executable elements within the given type element, including executable elements defined in super
-     * classes and implemented interfaces. Methods defined in {@link java.lang.Object} are ignored, as well as
-     * implementations of {@link java.lang.Object#equals(Object)}.
+     * classes and implemented interfaces. Methods defined in {@link java.lang.Object},
+     * implementations of {@link java.lang.Object#equals(Object)} and private methods are ignroed
```
Captain1653: small typo "ignored" :)

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
```diff
@@ -168,6 +168,15 @@ private static boolean isNotObjectEquals(ExecutableElement executable) {
         return true;
     }
 
+    /**
+     * @param executable the executable to check
+     *
+     * @return {@code true}, iff the executable does not have a private modifier
```
Captain1653: small typo "if"

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
```diff
@@ -347,4 +378,15 @@ public int compare(ModelElementProcessor<?, ?> o1, ModelElementProcessor<?, ?> o
             return Integer.compare( o1.getPriority(), o2.getPriority() );
         }
     }
+
+    private static class DeferredMapper {
```
Captain1653: I'm not sure. Should we write some javadoc about the purpose of this private static class? Or is it obvious?

processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
filiphr: iff is shortened for if and only if 😄

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
filiphr: I thought that it was obvious from the context. If it isn't I can add some javadoc

Captain1653: No, you are right

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
```diff
@@ -163,6 +164,32 @@ public boolean process(final Set<? extends TypeElement> annotations, final Round
...
+                String erroneousElementName;
+
+                if ( erroneousElement instanceof QualifiedNameable ) {
+                    erroneousElementName = ( (QualifiedNameable) erroneousElement ).getQualifiedName().toString();
+                }
+                else {
+                    erroneousElementName = erroneousElement.getSimpleName().toString();
+                }
+
+                processingEnv.getMessager()
```
sjaakd: is there any way to figure out what annotatiotion processors are in the compilation cycle? If so, it we should list them. If not: it seems that the error message needs to point more clearly to the underlying cause in a more generic way. something along the lines what you have now, but a hint that another ap (not generating classes) might be the problem.

filiphr: I don't think that we get can the current annotation processors. But if you have some better idea for the message I would gladly hear it. 

I can add the hint that maybe some another annotation processor didn't create the class and that you could enable verbose mode for mapstruct so you can see what is happening.

filiphr: I changed the text a bit let me know what you think

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
```diff
@@ -163,6 +164,40 @@ public boolean process(final Set<? extends TypeElement> annotations, final Round
...
+                // Otherwise the reporting will not work properly
+                deferredMapperElement = annotationProcessorContext.getElementUtils()
+                    .getTypeElement( deferredMapperElement.getQualifiedName() );
+
+                processingEnv.getMessager()
+                    .printMessage(
+                        Kind.ERROR,
+                        "No implementation was created for " + deferredMapperElement.getSimpleName() +
+                            " due to having a problem in the erroneous element " + erroneousElementName + "." +
+                            " Hint: this usually means that some other annotation processor was supposed to" +
```
sjaakd: usually == often?

filiphr:
Thanks for the review @sjaakd and @Captain1653

PR #1912, opened 2019-09-21
 #1911 Change return type MapperConfiguration.getBuilderPrism from Optional<BuilderPrism> to BuilderPrism

### LINKED ISSUE #1911: Change return type MapperConfiguration.getBuilderPrism from Optional<BuilderPrism> to BuilderPrism
Method MapperConfiguration.getBuilderPrism return Optional<BuilderPrism>, but our usage looks like:
`optional.orElse( null )`
or
```
if (optional.isPresent()) {
    optional.get()
}
```
  It seems there are not any benefits from Optional. Maybe is it better to remove it?

Captain1653:
Fixes #1911

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractMappingMethodBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractMappingMethodBuilder.java
@@ -58,7 +58,7 @@ public abstract class AbstractMappingMethodBuilder<B extends AbstractMappingMeth
         return createForgedAssignment(
                         sourceRHS,
                         ctx.getTypeFactory()
-                           .builderTypeFor( targetType, BeanMapping.builderPrismFor( method ).orElse( null ) ),
+                           .builderTypeFor( targetType, BeanMapping.builderPrismFor( method ) ),
                         forgedMethod
         );
     }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BuilderFinisherMethodResolver.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BuilderFinisherMethodResolver.java
@@ -6,7 +6,6 @@
 package org.mapstruct.ap.internal.model;
 
 import java.util.Collection;
-import java.util.Optional;
 import javax.lang.model.element.ExecutableElement;
 
 import org.mapstruct.ap.internal.model.common.BuilderType;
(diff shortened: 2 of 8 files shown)


sjaakd:
> LGTM, but I would leave it up to @sjaakd. He is the one that added that part :)

It was I guess what @chris922 also meant during our hangout. My preference would be to go about the other way around. 

So, we have classes in `org.mapstruct.ap.internal.model.source` these classes reflect our source model. And, also the annotations in our source model. Where the prism is just a convenient wrapper, the classes in  `org.mapstruct.ap.internal.model.source` are used to check validity of annotations during the `MethodRetrieveal` phase.

Annotations have mandatory parameters and optional parameters. This is exactly where `Optional` comes in handy. For instance, in `@Mapping` only `targetName` is mandatory. All other props should be `Optional` to express that you cannot rely on their presence.

So to make a long story short.. I was going the other way around.. This was the first one I did. But I also know that @filiphr is not so fond of the `Optional`, so the verdict is still out.

Captain1653:
So, what are we going to do? Are we going to change some others parameters with optional and to see how it will look? If we need it I can do it. I mean for some class, or two classes like little experiment.

sjaakd:
> So, what are we going to do? Are we going to change some others parameters with optional and to see how it will look? If we need it I can do it. I mean for some class, or two classes like little experiment.

In the discussion we had we also [discussed](https://stackoverflow.com/questions/13484202/how-to-use-nullable-and-nonnull-annotations-more-effectively) `@Nonnull` and `@Nullable`. That's an annotation that is checked by your IDE (and less evasive). 

As a suggestion: 
* leave this PR open.
* experiment with the `@Nonnull` and `@Nullable` instead in a new PR.

if the latter works out, we could merge this one..

filiphr:
Just to make it clear I am not fond of optional where we do `orElse(null)` and we do a `isPresent()` or the reverse check since that is the same as using `null`. I would indeed prefer using the JSR305 annotations in the codebase instead.

sjaakd:
> Just to make it clear I am not fond of optional where we do `orElse(null)` and we do a `isPresent()` or the reverse check since that is the same as using `null`. I would indeed prefer using the JSR305 annotations in the codebase instead.

There's a matter of taste involved here as well 😉  . But I agree: we should not have 2 mechanisms in parallel. So, I'm ok with removing the optional.

sjaakd:
@Captain1653 LGTM.. Please go ahead and merge. 

However, its still worthwhile to investigate point 2
*  experiment with the `@Nonnull` and `@Nullable` instead in a new PR.

PR #1915, opened 2019-09-21
 #1914 Remove unused method Type.isAnnotatedWith

### LINKED ISSUE #1914: Refactoring class Type
There is unused method Type.isAnnotatedWith and some big methods in the class Type. Should we remove this method and to do some refactoring?

Captain1653:
Fixes #1914

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -14,12 +14,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Stream;
-import javax.lang.model.element.AnnotationMirror;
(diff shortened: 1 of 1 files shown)


Captain1653:
> The removal of `isAnnotatedWith` is ok for me. However, the split for the other methods not really. Why is it bad to have a bigger method? Was it complex to follow it what is happening?

Honestly, It was little complex for me. I thought it can be little more readable, but I can revert splitted method and replaced with Stream API if it necessary.

filiphr:
I would prefer the previous non split method. As for replacing it with Stream API, I doubt that it would make it more readable. Stream is not always better :)

Captain1653:
@filiphr Ok, I have fixed it.

filiphr:
Thanks, feel free to merge when you can

Captain1653:
Thanks for review @sjaakd and @filiphr

PR #1917, opened 2019-09-22
#1799 Fluent setters starting with set should work properly

### LINKED ISSUE #1799: Fluent setters on fields with names starting with set are not working
It seems in version MapStruct 1.3.0-Final theres an issue with builders and fields that begin with the word set such as settlement or setup.

Example (with lombok)

Source: 
```
@Builder
public class Foo {
    private String setup;
    private LocalDate settlement;

}
```

Target: 
```
@Builder
public class Bar {
    private String setup;
    private LocalDate settlement;

}
```

Neither of these fields are mapped. But I did get it to work by adding

```
@Mapping(source = "settlement", target = "tlement")
@Mapping(source = "setup", target = "up")
```

filiphr:
Fixes #1799

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
+++ b/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
@@ -172,10 +172,10 @@ public class DefaultAccessorNamingStrategy implements AccessorNamingStrategy {
     public String getPropertyName(ExecutableElement getterOrSetterMethod) {
         String methodName = getterOrSetterMethod.getSimpleName().toString();
         if ( isFluentSetter( getterOrSetterMethod ) ) {
-            // If this is a fluent setter that starts with get or set and the 4th character is an uppercase one
+            // If this is a fluent setter that starts with set and the 4th character is an uppercase one
             // then we treat it as a Java Bean style method (we get the property starting from the 4th character).
             // Otherwise we treat it as a fluent setter
-            if ( ( methodName.startsWith( "get" ) || methodName.startsWith( "set" ) )
+            if ( methodName.startsWith( "set" )
                 && methodName.length() > 3
                 && Character.isUpperCase( methodName.charAt( 3 ) ) ) {
                 return IntrospectorUtils.decapitalize( methodName.substring( 3 ) );
--- a/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
+++ b/processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
@@ -57,7 +57,8 @@ public class DefaultAccessorNamingStrategy implements AccessorNamingStrategy {
     }
 
     /**
-     * Returns {@code true} when the {@link ExecutableElement} is a getter method. A method is a getter when it starts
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
```diff
@@ -166,11 +171,18 @@ public boolean isPresenceCheckMethod(ExecutableElement method) {
     @Override
     public String getPropertyName(ExecutableElement getterOrSetterMethod) {
         String methodName = getterOrSetterMethod.getSimpleName().toString();
-        if ( methodName.startsWith( "get" ) || methodName.startsWith( "set" ) ) {
-            return IntrospectorUtils.decapitalize( methodName.substring( 3 ) );
-        }
-        else if ( isFluentSetter( getterOrSetterMethod ) ) {
-            return methodName;
+        if ( isFluentSetter( getterOrSetterMethod ) ) {
```
sjaakd: feels a bit weird.. just below here you are asserting whether the method starts with "get".. Perhaps move `isFluentSetter` round or make it just `isFluent`

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
```diff
@@ -166,11 +171,18 @@ public boolean isPresenceCheckMethod(ExecutableElement method) {
...
+            // If this is a fluent setter that starts with get or set and the 4th character is an uppercase one
+            // then we treat it as a Java Bean style method (we get the property starting from the 4th character).
+            // Otherwise we treat it as a fluent setter
+            if ( ( methodName.startsWith( "get" ) || methodName.startsWith( "set" ) )
+                && methodName.length() > 3
+                && Character.isUpperCase( methodName.charAt( 3 ) ) ) {
```
sjaakd: I remember this was topic of discussion right? I mean, whether the third letter must be capital or not?

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
filiphr: You are right, would only make sense to check for the "get". Will change that

processor/src/main/java/org/mapstruct/ap/spi/DefaultAccessorNamingStrategy.java
filiphr: Yes it was a topic and to me the most sense was to treat strictly uppercase 4th character for fluent setter. This are special kind of setters that return the same instance back.

For example:

```
public Builder setSettlementDate(String settlementDate);
public Builder settlementDate(String settlementDate);
```

If someone uses something else for fluent setters then we don't support that.

sjaakd: would be good to add this particular example to the comment.

filiphr: Updated. Does it make more sense now?

sjaakd: yup..

sjaakd:
LGTM.. please go ahead and merge..

PR #1923, opened 2019-09-24
#737 Ensure types given for @Mapper#uses() and config() are valid

Captain1653:
@sjaakd  and @filiphr I need you help with issue #737 .
1. Do we still need to do it? :) (It's a little old issue)
2. I have some implementation (sure, it isn't so good), but there are some strange behavior. 
Tests for module processor are green. Integration tests are failed in class FullFeatureCompilationTest. My tests in idea failed with eclipse. (Is it some specific thing with eclipse?)
3. In class CompilationStatement I have 9 identical errors (honestly, I don't know. that's why I use discinct. How can I avoid it? like `filterExpectedDiagnostics` or something like that ? Am I doing something wrong?
4. Can you say me right direction or give some advice? Should I leave this issue, because it isn't my level and too complicated for me?

full build
tests module processor
failed test and filter

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
@@ -60,7 +60,8 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
     }
 
     private boolean isDisableSubMappingMethodsGeneration() {
-        MapperConfiguration configuration = MapperConfiguration.getInstanceOn( ctx.getMapperTypeElement() );
+        MapperConfiguration configuration = MapperConfiguration.getInstanceOn(
+                                                            ctx.getMapperTypeElement(), ctx.getMessager() );
         return configuration.isDisableSubMappingMethodsGeneration();
     }
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/MapperConfiguration.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/MapperConfiguration.java
@@ -12,6 +12,7 @@ import java.util.Optional;
 import java.util.Set;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.Element;
+import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeMirror;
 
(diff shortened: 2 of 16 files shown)


processor/src/main/java/org/mapstruct/ap/internal/util/MapperConfiguration.java
```diff
@@ -42,19 +43,29 @@
...
+    private MapperConfiguration(MapperPrism mapperPrism, FormattingMessager messager) {
         this.mapperPrism = mapperPrism;
+        this.messager = messager;
 
         if ( mapperPrism.values.config() != null ) {
-            // TODO #737 Only a declared type makes sense here; Validate and raise graceful error;
-            // Also validate that @MapperConfig is present
-            this.config = (DeclaredType) mapperPrism.config();
-            this.mapperConfigPrism = MapperConfigPrism.getInstanceOn( config.asElement() );
+            if ( mapperPrism.config() instanceof DeclaredType ) {
```
sjaakd: I think the config itself even needs to be an interface.. so you can check that as well

processor/src/test/java/org/mapstruct/ap/test/mypackage/CheckDeclaredTypeInMapperUses.java
```diff
@@ -0,0 +1,65 @@
...
+                kind = javax.tools.Diagnostic.Kind.ERROR,
+                line = 16,
+                messageRegExp = "Type \"int\" cannot be used in \"@Mapper\\.uses\\(\\)\""
+            )
+        }
+    )
+    public void shouldCompilationFailedForMapperUsesNotDeclaredType() { }
+
+    @Test
+    @WithClasses( { MapperWithNotDeclaredTypeInMapperConfigUses.class, MapperConfigWithNotDecladerTypeInUses.class} )
```
sjaakd: There's a convention: mappers that give errors should start with `ErroneousXYZ`..  Perhaps that's the problem in your full feature integration test

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
```diff
@@ -60,7 +60,8 @@ boolean canGenerateAutoSubMappingBetween(Type sourceType, Type targetType) {
     }
 
     private boolean isDisableSubMappingMethodsGeneration() {
-        MapperConfiguration configuration = MapperConfiguration.getInstanceOn( ctx.getMapperTypeElement() );
+        MapperConfiguration configuration = MapperConfiguration.getInstanceOn(
```
sjaakd: I'm not so happy with the amount of `getInstanceOn` that we do and the distributed checking of the config. Shouldn't we treat it more like we treat other annotations: have a class doing the checking in our `org.mapstruct.ap.internal.model.source` package?

sjaakd:
> 1. Do we still need to do it? :) (It's a little old issue)

I think so :).

sjaakd:
> 2\. I have some implementation (sure, it isn't so good), but there are some strange behavior.
>     Tests for module processor are green. Integration tests are failed in class FullFeatureCompilationTest. My tests in idea failed with eclipse. (Is it some specific thing with eclipse?)

Don't worry.. 

Could it be the erroneous mapper name? just guessing here.

sjaakd:
> 3\. n class CompilationStatement I have 9 identical errors (honestly, I don't know. that's why I use discinct. How can I avoid it? like `filterExpectedDiagnostics` or something like that ? Am I doing something wrong?

I'm not sure here .. are these coming from the same test case?

Captain1653:
> > 3. n class CompilationStatement I have 9 identical errors (honestly, I don't know. that's why I use discinct. How can I avoid it? like `filterExpectedDiagnostics` or something like that ? Am I doing something wrong?
> 
> I'm not sure here .. are these coming from the same test case?
Yes, from one test case. As you can see we use `getInstanceOn` many times that's why we receive so many identical warnings/errors. When I'll check config in one place this problem with be solved.

Captain1653:
> > 3. n class CompilationStatement I have 9 identical errors (honestly, I don't know. that's why I use discinct. How can I avoid it? like `filterExpectedDiagnostics` or something like that ? Am I doing something wrong?
> 
> I'm not sure here .. are these coming from the same test case?

Yes, from one test case. As you can see we use getInstanceOn many times that's why we receive so many identical warnings/errors. When I'll check config in one place this problem with be solved.

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
Captain1653: Maybe I'm wrong, but unfortunatelly no. We haven't class in org.mapstruct.ap.internal.model.source for mapper configuration. We always use MapperConfiguration.getInstanceOn eight times. Sometimes I think, it's too early to do this task with checking. Firstly we should refactor, maybe. What do you think?

Captain1653:
I think right now it's too earlt for this PR. Firstly we should refactor our code and create MapperConfig in source package, like said @sjaakd . I close it.

sjaakd:
@Captain1653 .. thanks.

PR #1927, opened 2019-09-26
#1742 Add test case

filiphr:
Mostly to make the backport of the fix easier


filiphr:
Must have missed something. Need to have a look  what

PR #1930, opened 2019-09-29
#1828 Use update method when nested property of target is populated with multiple nested properties of source

### LINKED ISSUE #1828: MapStruct 1.3.0.Final generates invalid implementation when nested property of target is populated with multiple nested properties of source
MapStruct v1.3.0.Final generates invalid mapper implementation when nested object property of target is populated from multiple nested object propertiese of source. MapStruct v1.1.0.Final works fine in this scenario.

```
@Mapper
public interface FirstMapper {

    @Mapping(target = "name", source = "name")
    @Mapping(target = "completeAddress.lineOne", source = "specialAddress.line1")
    @Mapping(target = "completeAddress.lineTwo", source = "specialAddress.line2")
    @Mapping(target = "completeAddress.city", source = "generalAddress.city")
    @Mapping(target = "completeAddress.country", source = "generalAddress.country")
    public Person mapPerson(Employee employee);
}
```
Where structure of target (Person) and person(Employee) is as below
```
class Person { String name; CompleteAddress completeAddress; }

class Employee { String name; GeneralAddress generalAddress,  SpecialAddress specialAddress; }
```

Generated Mapper is as below
```
public class FirstMapperImpl implements FirstMapper {

    @Override
    public Person mapPerson(Employee employee) {
        if ( employee == null ) {
            return null;
        }

        Person person = new Person();

        person.setCompleteAddress( generalAddressToCompleteAddress( employee.getGeneralAddress() ) );
        person.setCompleteAddress( specialAddressToCompleteAddress( employee.getSpecialAddress() ) );
        person.setName( employee.getName() );

        return person;
    }

    protected CompleteAddress generalAddressToCompleteAddress(GeneralAddress generalAddress) {
        if ( generalAddress == null ) {
            return null;
        }

        CompleteAddress completeAddress = new CompleteAddress();

        completeAddress.setCountry( generalAddress.getCountry() );
        completeAddress.setCity( generalAddress.getCity() );

        return completeAddress;
    }

    protected CompleteAddress specialAddressToCompleteAddress(SpecialAddress specialAddress) {
        if ( specialAddress == null ) {
            return null;
        }

        CompleteAddress completeAddress = new CompleteAddress();

        completeAddress.setLineOne( specialAddress.getLine1() );
        completeAddress.setLineTwo( specialAddress.getLine2() );

        return completeAddress;
    }
}
```

Here generated methods `generalAddressToCompleteAddress` and `specialAddressToCompleteAddress` are both returning new instance of `CompleteAddress`. Overridden method `mapPerson` is calling `person.setCompleteAddress` with output of these two generated methods so one instance is thrown away in the process.So, with v1.3.0.Final, mapper is not correctly implemented as per defined mapping.

Please see [this repo](https://github.com/mabilalmirza/mapstruct-multiple-nested-objects) created to demostrate this issue.
- Branch [mapstruct-1.1.0.Final](https://github.com/mabilalmirza/mapstruct-multiple-nested-objects/tree/mapstruct-1.1.0.Final) is passing all unit tests.
- Branch [mapstruct-1.3.0.Final](https://github.com/mabilalmirza/mapstruct-multiple-nested-objects/tree/mapstruct-1.3.0.Final) fails to pass the tests.

filiphr:
Fixes #1828 for the 1.3.1 release

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
@@ -160,6 +160,15 @@ public class NestedTargetPropertyMappingHolder {
                         groupedByTP.singleTargetReferences.get( targetProperty )
                     );
 
+                    // If there are multi source parameters that are mapped to the target reference or
+                    // there are multiple source references to the target reference then we need to force an
+                    // update method. The reason for this is that multiple sources have effect on the target.
+                    // See Issue18282Test for more info.
+                    boolean forceUpdateMethod =
+                        multipleSourceParametersForTP || groupedSourceReferences.groupedBySourceReferences.size() > 1;
+
+                    boolean forceUpdateMethodOrNonNestedReferencesPresent =
+                        forceUpdateMethod || !groupedSourceReferences.nonNested.isEmpty();
                     // For all the groupedBySourceReferences we need to create property mappings
                     // from the Mappings and not restrict on the defined mappings (allow to forge name based mapping)
                     // if we have composite methods i.e. more then 2 parameters then we have to force a creation
@@ -168,8 +177,6 @@ public class NestedTargetPropertyMappingHolder {
                         .groupedBySourceReferences
                         .entrySet() ) {
                         PropertyEntry sourceEntry = entryBySP.getKey();
-                        boolean forceUpdateMethodOrNonNestedReferencesPresent =
-                            multipleSourceParametersForTP || !groupedSourceReferences.nonNested.isEmpty();
                         // If there are multiple source parameters that are mapped to the target reference
                         // then we restrict the mapping only to the defined mappings. And we create MappingOptions
                         // for forged methods (which means that any unmapped target properties are ignored)
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
@@ -160,10 +160,11 @@ public class NestedTargetPropertyMappingHolder {
                         groupedByTP.singleTargetReferences.get( targetProperty )
                     );
 
-                    // If there are multi source parameters that are mapped to the target reference or
-                    // there are multiple source references to the target reference then we need to force an
-                    // update method. The reason for this is that multiple sources have effect on the target.
-                    // See Issue18282Test for more info.
+                    // We need an update method in the when one of the following is satisfied:
+                    // 1) Multiple source parameters for the target reference
+                    // 2) Multiple source references for the target reference
+                    // The reason for this is that multiple sources have effect on the target.
+                    // See Issue1828Test for more info.
                     boolean forceUpdateMethod =
                         multipleSourceParametersForTP || groupedSourceReferences.groupedBySourceReferences.size() > 1;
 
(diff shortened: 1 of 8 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -160,6 +160,15 @@ public NestedTargetPropertyMappingHolder build() {
                         groupedByTP.singleTargetReferences.get( targetProperty )
                     );
 
+                    // If there are multi source parameters that are mapped to the target reference or
+                    // there are multiple source references to the target reference then we need to force an
+                    // update method. The reason for this is that multiple sources have effect on the target.
+                    // See Issue18282Test for more info.
```
sjaakd: nr. 18282.. Did not know we have so many issues :)

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -160,6 +160,15 @@ public NestedTargetPropertyMappingHolder build() {
                         groupedByTP.singleTargetReferences.get( targetProperty )
                     );
 
+                    // If there are multi source parameters that are mapped to the target reference or
```
sjaakd: Start active phrasing: , so `We need an update method when: 1) condition 1, 2) condition 2).` I can't understand the current sentence

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
```diff
@@ -160,6 +160,15 @@ public NestedTargetPropertyMappingHolder build() {
...
+                    // there are multiple source references to the target reference then we need to force an
+                    // update method. The reason for this is that multiple sources have effect on the target.
+                    // See Issue18282Test for more info.
+                    boolean forceUpdateMethod =
```
sjaakd: are there really 2 booleans required to fix this?

processor/src/test/java/org/mapstruct/ap/test/bugs/_1828/FirstMapper.java
```diff
@@ -0,0 +1,24 @@
...
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface FirstMapper {
+
+    FirstMapper INSTANCE = Mappers.getMapper( FirstMapper.class );
+
+    @Mapping(target = "name", source = "name")
```
sjaakd: is this required? name based mapping?

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: 😂, good catch

processor/src/main/java/org/mapstruct/ap/internal/model/NestedTargetPropertyMappingHolder.java
filiphr: It is one boolean, but I wanted to extract one that forces an update method and the other one with the nested reference

filiphr:
>Are you going to map this as well to the current master?

Already done in PR https://github.com/mapstruct/mapstruct/pull/1931

processor/src/test/java/org/mapstruct/ap/test/bugs/_1828/FirstMapper.java
filiphr: Nope it isn't I copied it from the example

filiphr:
Applied the review comments

Issue #1933, opened 2019-10-03
@BeanMapping(ignoreByDefault=true) in @MapperConfig not passed to mapping method 


steirer00:
I have an Issue with the ignoreByDefault parameter of @BeanMapping:
Using the @BeanMapping on the mapping method works but if i define  in an @MapperConfig it is ignored by the mapping method. 
 ->Meaning all fields are mapped which are not ignored.

Do i use this feature the wrong way or is there some bug?

```java
@MapperConfig(mappingInheritanceStrategy = MappingInheritanceStrategy.AUTO_INHERIT_FROM_CONFIG)
public interface BasicMapperConfig{
    @Mapping(target = "id", ignore = true)
    @Mapping(target = "updateCount", ignore = true)
    @Mapping(target = "metaInfo", ignore = true)
    @BeanMapping(nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT, ignoreByDefault = true)
	DefaultDbo updateEntity(AbstractDto dto, @MappingTarget DefaultDbo entity);
}

@Mapper(config=BasicMapperConfig)
public abstract class TestMapper
{
    public abstract DefaultDbo updateDefaultEntity(AbstractDto dto, @MappingTarget DefaultDbo dbo);
}
```

sjaakd:
@steirer00 this kind of inheritance is explicitly disabled. I don't now why. Can't find any lead in the documentation nor in the original PR #1403 .. 

@filiphr .. was this us being cautious when implementing #1403? It feels like inconsistent now (@hindsight) .

@steirer00 .. some small remarks about your design: your prototype mapping should be in an interface. We still need to enforce this. Its also only about the mapping. You don't need to define a mapping target here. 

to solve your problem, just stick the `@BeanMapping(ignoreByDefault=true)` on the method in the mapper, although this is not ideal I realise.

filiphr:
@sjaakd I think that it was us being cautious or we didn't notice it. I would say that we should make it consistent and inherit it as well

sjaakd:
@steirer00 thanks for reporting.. Issue is fixed. Will be available in 1.4.0

PR #1935, opened 2019-10-05
#1933 ignoreByDefault not inherited from config

sjaakd:


sjaakd:
@filiphr .. there are failing test cases (see also issue #1933) that are deliberately designed to catch this situation. I can't remember why I did it like this.

filiphr:
@sjaakd looks good to me. I think it was an oversight. So I would say fix the tests and then we can merge it

sjaakd:
@filiphr thanks for reviewing..

Issue #1557, opened 2018-07-20
relieve rules for String->enum mapping


tomaszalusky:
Suppose we convert object of type `Foo` to `FooDto`, both classes have `fruit` field of enum type, respectively: (reasoning: `Fruit` might change and `FooDto` is used in frontend whose development is not tightly coupled with backend, so we want frontend not to crash when say `Fruit.ORANGE` constant is added to backend.)

    public enum Fruit {APPLE, BANANA, COCONUT}
    public enum FruitDto {APPLE, BANANA, COCONUT, FALLBACK}

This works great thanks to abstract conversion method `FruitDto m(Fruit)` annotated with `@ValueMapping(source = MappingConstants.ANY_REMAINING, target = "FALLBACK")`, generated method contains switch with common constants and default branch for `FALLBACK`. 

What's an issue: in some cases the `Foo.fruit` property is of type `String` (not `Fruit`) with literally same content. In such case conversion method `Fruit m(String)` either must not exist (which renders automatic conversion without fallback) or must be written manually with body. Abstract method with annotation above leads to `Can't generate mapping method from non-enum type to enum type` error.

Proposal: allow `@ValueMapping` on String->enum methods. Instead switching on `Fruit`s, generated code could switch on Strings. Case statements' constant values would be taken from target enum. IMHO it wouldn't mind since currently generator can render only intersection of two enum sets anyway (in `ANY_REMAINING` mode; `ANY_UNMAPPED` mode would behave same way).

sjaakd:
What about a regular mapping here? `@ValueMapping` was introduced for mapping enum-to-enum. There was discussion for extending this to other types. However it was not deemed use full at the time, since immutable maps are just as efficient.

tomaszalusky:
Could you please provide an example for my case or more elaborate on what you mean by "regular mapping"?
I cannot think out how would @Mapping annotation apply, multiple repeating annotations AFAIK serve for more fields of single mapping rather then more values of single field. I want to avoid enumerating constants manually, one at a time.

sjaakd:
> Could you please provide an example for my case or more elaborate on what you mean by "regular mapping"?

Implement a method (so manually) with uses a map to do the mapping. If fields can be mapped based on name, you could possibly shortcut by only mapping the (non matching names) in the map.

filiphr:
@tomaszalusky what @sjaakd  is referring to looks something like:

```
@Mapper
public interface MyMapper {

    BasketDto map(Basket source);

[...]

PR #1938, opened 2019-10-06
#1557 String-to-enum and enum-to-String

sjaakd:
working principle all source mappings must be mapped.

name based mapping is the default for all

**enum to enum**

1. ALL_REMAINING means, continue with name based (same name in target and source), what remains will be mapped to what is defined as ALL_REMAINING (default in case)
2. ALL_UNMAPPED means, stops after defined mapping, skips name based, map what remains to ALL_UNMAPPED (default in case)

**enum to string**, 

1. all not defined mappings will result in the source enum constant value being mapped to the string
2. ALL_REMAINING does not make sense there is no sense. We could claim all constants in source are present in the target, but then there would not be a default remaining. Or we could claim no constants in source are by default present, but then ALL_REMAINING would be the same as ALL_UNMAPPED. Better to avoid such ambiguous situation and give error.
3. ALL_UNMAPPED stops after handling defined mapping and proceeds to the default clause value.

**string to enum**

1. all not defined mappings will result in the target enum constant mapped from the string value
2. ALL_REMAINING does not make any sense on the same premise as above.
3. ALL_UNMAPPED stops after handling defined mapping and proceeds to the default clause value.


processor/src/test/java/org/mapstruct/ap/test/value/string2enum/StringToEnumMappingTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+@IssueKey( "1557" )
+@WithClasses({ OrderType.class, OrderMapper.class })
+@RunWith(AnnotationProcessorTestRunner.class)
+public class StringToEnumMappingTest {
+
+    @Test
+    public void testTheNormalStuff() {
```
filiphr: I think that for this we can also test that the exception is thrown for an unknown `String`

sjaakd: Hmm.. this is actually an interesting case. In the past we added this one for the possibility that a library containing the enum is updated runtime (and a new enum value has been added), but not yet compile time.

But what should we do for when the input is a `String`? I guess an `IllegalArgument` is justified. Or perhaps we should demand an `ANY_UNMAPPED` and even give an error when it's absent.. ?

filiphr: I would say it is up to the user to define what they want. Throwing the exception is completely OK.

sjaakd: > But what should we do for when the input is a `String`? I guess an `IllegalArgument` is justified. Or perhaps we should demand an `ANY_UNMAPPED` and even give an error when it's absent.. ?

thinking more about this.. This is also not the way forward. `ANY_UNMAPPED` stops after the defined mappings. Making it mandatory means the user always have to specify the mappings. 

Back to `ANY_REMAINING` and `ANY_UNMAPPED` for the string-to-enum case. And making one of these mandatory in this scenario.

sjaakd: done.

sjaakd:
@filiphr . thanks for reviewing.. I changed the behaviour for string to enum mapping. I don't think our code should throw `illegalArgument` if we can resolve things by making the api more consistent. So, that's what I did.

Have a look and see if you agree.

processor/src/test/java/org/mapstruct/ap/test/value/string2enum/StringToEnumMappingTest.java
sjaakd: > I would say it is up to the user to define what they want. Throwing the exception is completely OK.

Generating an `IllegalArgumentException` for this case might surprise the user when running the code. A String not matching the enum constant will be pretty common. 

I prefer to make the user aware that he should make a choice here at the expense of a little bit more verbosity.

filiphr: I would say that having a compile error is too strict. Perhaps only a warning so they are aware about it. Perhaps they want to do it on purpose like that.

documentation/src/main/asciidoc/chapter-8-mapping-values.asciidoc
```diff
@@ -67,7 +67,7 @@ public class OrderMapperImpl implements OrderMapper {
 By default an error will be raised by MapStruct in case a constant of the source enum type does not have a corresponding constant with the same name in the target type and also is not mapped to another constant via `@ValueMapping`. This ensures that all constants are mapped in a safe and predictable manner. The generated
 mapping method will throw an IllegalStateException if for some reason an unrecognized source value occurs.
 
-MapStruct also has a mechanism for mapping any remaining (unspecified) mappings to a default. This can be used only once in a set of value mappings. It comes in two flavors: `<ANY_REMAINING>` and `<ANY_UNMAPPED>`.
+MapStruct also has a mechanism for mapping any remaining (unspecified) mappings to a default. This can be used only once in a set of value mappings and only applies to the source. It comes in two flavors: `<ANY_REMAINING>` and `<ANY_UNMAPPED>`. The cannot be used at the same time.
```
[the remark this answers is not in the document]
sjaakd: done.

documentation/src/main/asciidoc/chapter-8-mapping-values.asciidoc
```diff
@@ -1,7 +1,7 @@
 [[mapping-enum-types]]
 == Mapping Values
 
-=== Mapping enum types
+=== Mapping enum to enum types
```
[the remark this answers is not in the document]
sjaakd: added a link name

sjaakd:
@filiphr thanks for reviewing.

Issue #1940, opened 2019-10-08
Confirm by annotation it is a value mapping


sjaakd:
currently, as #1788 shows, we can sometimes select a @BeanMapping method in stead of an enum mapping method. Especially in case of parameter mapping this can be the case.

There is no way to force this currently, whereas the chance that we did break something did increase by allowing string-to-enum and enum-to-string.

Hence we give the user a possibility to force a `@BeanMapping` where normally a `@ValueMapping` would be selected.

PR #1941, opened 2019-10-08
#1940 use annotations as additional criterion to select valuemapping

sjaakd:


filiphr:
I don't think that this is needed. You can't really do bean mapping when you do Enum to Enum, String to Enum or Enum to String.

sjaakd:
Unless you map parameters. But having said that.. the target is a bean in that case. So you are probably right

Verstuurd vanaf mijn iPhone

filiphr:
So should we close this and the issue?

sjaakd:
Done

PR #1950, opened 2019-10-19
#1645 Generate error when java 6 or java 7 is used

### LINKED ISSUE #1645: Cleanup dropping support for Java 6 and 7

Captain1653:
I have tried to generate error when we use  java 6 or java 7, but I need some help. 
1. Should I write some tests for it and how can I do it?
2. I use java.specification.version . Is it right? I have tried SourceVersion.compare, but it doesn't work with Eclipse correctly (I have received RELEASE_8 on Oracle JDK and RELEASE_7 on Eclipse)
3. I saw two points with SpecificCompikerWorkarounds and rename folder, but workarounds it's too hard for me. About rename folder I didn't understand (honestly :))

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/DefaultVersionInformation.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/DefaultVersionInformation.java
@@ -12,8 +12,11 @@ import java.util.jar.Manifest;
 
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.SourceVersion;
+import javax.tools.Diagnostic;
 
 import org.mapstruct.ap.internal.version.VersionInformation;
+
+import static org.mapstruct.ap.internal.util.Message.GENERAL_JAVA_VERSION_IS_NOT_SUPPORTED;
 
 /**
  * Provides information about the processor version and the processor context implementation version.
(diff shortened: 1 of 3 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/DefaultVersionInformation.java
```diff
@@ -92,6 +95,14 @@ static DefaultVersionInformation fromProcessingEnvironment(ProcessingEnvironment
         String runtimeVersion = System.getProperty( "java.version" );
         String runtimeVendor = System.getProperty( "java.vendor" );
 
+        String javaSpecificationVersion = System.getProperty( "java.specification.version" );
+
+        if ( javaSpecificationVersion.equals( "1.6" ) || javaSpecificationVersion.equals( "1.7" ) ) {
```
sjaakd: what about even earlier versions, 1.5 introduced annotation processing.  possibly not needed..

filiphr: I am not 100% convinced with using the java.specification property for this. I think that there are better ways for this and also the place for doing this check is not OK. Thinking more about this I think it is not that crucial to have it

Captain1653: >  I think it is not that crucial to have it
Maybe you are right. We haven't bugs about it, so... I can continue to work with it. I mean to fiend better way and better place if we need it. I'm waiting for your decision :)

Captain1653:
We have decided to close this PR without merge. Our users didn't complain about this things, so, I suppose, we can add this error message in the future if we will need it.

PR #1959, opened 2019-10-27
Updated docs to add compilerArgs instead of replacing them.

marceloverdijk:
When configuring MapStruct compilerArgs *add* them, not *replace* them


filiphr:
Thanks for the PR @marceloverdijk and sorry for the late merge :)

Issue #1689, opened 2019-01-18
Field with default visibility not found


gunnarmorling:
If there is a field with default visibility ("package private") on the target side, this is not found:

```
public class ZooDto {

	String mainAnimal;
}
```

```
Unknown property "mainAnimal" in result type org.mapstruct.ap.test.ignore.ZooDto. Did you mean "null"?
```

It works as expected when making that field public.

filiphr:
I think that this is the same for getters / setters. We only take public fields and public methods into consideration.

Do you think that we should report another error?

gunnarmorling:
> Do you think that we should report another error?

We should make use of the property if the mapper is in the same package as that type. In terms of the message itself, "Did you mean "null"?" seems a bit suspicious, too :)

filiphr:
>We should make use of the property if the mapper is in the same package as that type

I thought I've seen a request for this before as well. I've looked into our code and it would need some refactorings to make this work, but it is doable 😄.

I thought that we 
>In terms of the message itself, "Did you mean "null"?" seems a bit suspicious, too :)

Yes we need to change this 😄.

tokazio:
>  I've looked into our code 

Where can i start to search about that please ?

I need to generate mapping from package private fields.

tokazio:
Found Executables in 1.2.0 and Fields in 1.3+ with tests in FieldsMappingTest

tokazio:
https://github.com/mapstruct/mapstruct/pull/1975 for 1.2.0

JimTyrrell:
Hey all.  Any progress on this one?  I would love to use this capability in my summit talk that I have to record at the end of the month, between now and then can I contribute to making this issue go away?  Any pointers as to where to inject myself into resolving this.  Thank you1!!

Zegveld:
Here is where the source properties are found for the mapping:
https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java#L263
and here is where the target properties are found for the mapping:
https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java#L226

I'm not sure though where exactly the fields are found and how the non-public fields are removed from the collection of fields. But the above should be useful as starting points for this.

Also be aware that you will need to check if the mapper and the source/target share the same package. Otherwise you cannot access package scoped (package/protected) fields/setters/getters.

[...]

PR #1975, opened 2019-11-28
#1689 1.2.0 - Map accessors with default (package private) visibility

tokazio:
Map accessors with default (package private) visibility with MapStruct 1.2.0

I need this 1.2.0 version because i'm working on a legacy java7 project.

How to change the version to a 1.2.1 please ?

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/Executables.java
@@ -77,7 +77,7 @@ public class Executables {
 
     public static boolean isGetterMethod(Accessor method) {
         ExecutableElement executable = method.getExecutable();
-        return executable != null && isPublic( method ) &&
+        return executable != null && isMappableAccessor( method ) &&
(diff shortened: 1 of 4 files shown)


tokazio:
I don't know why but the whole processor does not work (nothing works) in eclipse in place of a 1.2.0.Final

tokazio:
Yes i was thinking about that. Maybe the processor can show an error when you try to map a 'package' field from a mapper outside... But it's adding complexity... And it needs more work...
My use cases are generally to map hibernate entities to dtos, entities allready needs getters/setters and dtos can be used from outside the domain to gives data informations then they should have getters/setters.

Protected access could be great when you want to work with 'immutable' data. I think a great tool like mapstruct should have options to work with this kind of accessors. Giving the choice to users.

My legacy project uses a 1.7 compiler with ant and a lot of xml config... With Eclipse and dozen of subprojects... 
Any idea why this doesn't work in jdt apt? Maybe a special config to build mapstruct for it? Any hint to debug the processing in eclipse?

I've tried from a maven and a gradle project, it's working...

Issue #807, opened 2016-05-02
Mapping composition


ebelanger:
It would be great if MapStruct could support mapping composition by defining a new annotation.  It could work similar to how [Hibernate Validator](http://docs.jboss.org/hibernate/validator/5.2/reference/en-US/html/ch06.html#section-constraint-composition) constraint composition does.  This would allow reuse of mapping configurations by duck-typing instead of inheritance, which is useful when you do not have total control over the entities.

gunnarmorling:
Hey Eric, could you provide an example of how such mapping composition
would look like and how it would be used? I'm aware of Bean Validation
constraint composition, but I'd like to know how you envision such
composition to apply to MapStruct.

ebelanger:
`
@Mapping(target = "id", ignore = true)
@Mapping(target = "deleted", ignore = true)
@Mapping(target = "account", ignore = true)
@Mapping(target = "options", ignore = true)
@Mapping(target = "promotions", ignore = true)
public abstract Plan update(PlanView view, @MappingTarget Plan plan);

@Mapping(target = "id", ignore = true)
@Mapping(target = "deleted", ignore = true)
@Mapping(target = "account", ignore = true)
@Mapping(target = "pricings", ignore = true)
@Mapping(target = "promotions", ignore = true)
public abstract Option update(OptionView optionView, @MappingTarget Option option);
`

In this scenario, Plan and Option do not share an interface explicitly but we can clearly see one is defined implicitely.  What I would like to do in this case is:

`
@Mapping(target = "id", ignore = true)
@Mapping(target = "deleted", ignore = true)
@Mapping(target = "account", ignore = true)
@Target(METHOD)
@Retention(RUNTIME)
@Documented
public @interface EntityMapping { }

@EntityMapping
@Mapping(target = "options", ignore = true)
@Mapping(target = "promotions", ignore = true)
public abstract Plan update(PlanView view, @MappingTarget Plan plan);

@EntityMapping
@Mapping(target = "pricings", ignore = true)
@Mapping(target = "promotions", ignore = true)
public abstract Option update(OptionView optionView, @MappingTarget Option option);
`

And that would pretty much be it.  It is pretty much the same as defining mappings for a base class, but for when you do not have one.

sjaakd:
👍 . I like the idea. Should also not be too difficult to realize.

ebelanger:
Great.  I will prepare a PR soon then :)

agudian:
[...]

PR #1979, opened 2019-12-03
#807 meta annotations and duck typing

sjaakd:
As much as possible used existing functionality. Recur though nested meta-annotations, excluding already found ones. Exclude also our own and java annotations (`@Documented` and the likes).

### DIFF
--- a/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
@@ -126,7 +126,7 @@ MapStruct takes all public properties of the source and target types into accoun
 
 [[mapping-composition]]
 === Mapping Composition (experimental)
-MapStruct supports the use of meta annotations. The `@Mapping` annotation can be defined with `@Target` `ElementType#ANNOTATION_TYPE`. This will allow `@Mapping` to be used on other (user defined) annotations for re-use purposes. For example:
+MapStruct supports the use of meta annotations. The `@Mapping` annotation supports now `@Target` with `ElementType#ANNOTATION_TYPE` in addition to `ElementType#METHOD`. This allows `@Mapping` to be used on other (user defined) annotations for re-use purposes. For example:
 
 ====
 [source, java, linenums]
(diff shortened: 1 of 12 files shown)


sjaakd:
> I had one question what would happen if you add a duplicated `@Mapping` (one which is defined in the meta annotation and another one on the method`?

it builds op a list of mappings, just as normal. If there are duplicates, they will be detected when adding them to the set.

sjaakd:
> Should we add some documentation about it perhaps? I think we should say that this is experimental and that we want some feedback.

We should.. I'll add an entry..

sjaakd:
> it builds op a list of mappings, just as normal. If there are duplicates, they will be detected when adding them to the set.

That might actually go wrong now...  I'll do some update

documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
```diff
@@ -124,6 +124,50 @@ Collection-typed attributes with the same element type will be copied by creatin
 
 MapStruct takes all public properties of the source and target types into account. This includes properties declared on super-types.
 
+[[mapping-composition]]
+=== Mapping Composition (experimental)
+MapStruct supports the use of meta annotations. The `@Mapping` annotation can be defined with `@Target` `ElementType#ANNOTATION_TYPE`. This will allow `@Mapping` to be used on other (user defined) annotations for re-use purposes. For example:
```
chris922: _can be defined_? Sounds like the user could do this.. but it actually _was defined with ..._? Following then in the next sentence: _will allow_ -> _allows_?

sjaakd: true! I'll fix this.. thanks @chris922

sjaakd: I rephrased it to: _The `@Mapping` annotation supports now `@Target` with `ElementType#ANNOTATION_TYPE` in addition to `ElementType#METHOD`. This allows `@Mapping` to be used ..._

sjaakd:
> Wow, really great @sjaakd !

Thanks 😄

Issue #1983, opened 2019-12-08
@InheritInverseConfiguration doesn't pickup ignore mapping


imtiazShakil:
@InheritInverseConfiguration doesn't apply ignore mapping 

Example:
`Student.java`
```java
@Getter
@Setter
public class Student {
  private String name;
  private String roll;
  private int age;
}
```

`StudentDto.java`
```java
@Getter
@Setter
public class StudentDto {
  private String name;
  private String roll;
  private int age;
}
```

My Mapper interface is:
```java
@Mapper()
public interface Mapper {
  @Mapping(target = "age", ignore = true)
  StudentDto studentToStudentDto(Student source);
  
  @InheritInverseConfiguration
  Student studentDtoToStudent(StudentDto source);
}
```

The expected behavior should be ignoring `setAge()` method from both implementation. but it's not working. This is what I get:

```java
@Generated(
    value = "org.mapstruct.ap.MappingProcessor",
    date = "2019-12-08T20:18:06+0600",
    comments = "version: 1.3.1.Final, compiler: javac, environment: Java 11.0.5 (Amazon.com Inc.)"
)
public class MapperImpl implements Mapper {

    @Override
    public StudentDto studentToStudentDto(Student source) {
        if ( source == null ) {
            return null;
        }

        StudentDto studentDto = new StudentDto();

        studentDto.setName( source.getName() );
        studentDto.setRoll( source.getRoll() );

        return studentDto;
    }

    @Override
    public Student studentDtoToStudent(StudentDto source) {
        if ( source == null ) {
            return null;
        }

        Student student = new Student();

        student.setName( source.getName() );
        student.setRoll( source.getRoll() );
        student.setAge( source.getAge() );

        return student;
    }
}
```

As you can see in the `studentDtoToStudent()` method `age` property is  not being ignored.

sjaakd:
Please, specify source as well.. remember: source becomes target when inversion takes place

imtiazShakil:
Thanks!

After specifying source parameter it's now working as expected. 

> remember: source becomes target when inversion takes place

I think you should add an example, so it becomes easy to understand and not miss this important detail. Maybe you can modify the `@InheritInverseConfiguration` example a bit like this:

```java
@Mapper
public interface CarMapper {

    @Mapping(source = "numberOfSeats", target = "seatCount")
    @Mapping(source = "engineClass", target = "enginePower", ignore=true)  
    CarDto carToDto(Car car);

[...]

PR #1986, opened 2019-12-12
#1983 javadoc and doc on @ignore in @InheritInverseConfituration

sjaakd:


sjaakd:
> Not sure if the example code is OK in an IDE. I think that we should not have an empty line after starting `<code>` and we should not have new lines before closing `

removed the empty lines.. Thx

Issue #1998, opened 2020-01-07
simplify usage of inheritance of annotation fields in @MapperConfig, @Mapper, @BeanMapping, @Mapping, @IterableMapping and @MapMapping


sjaakd:
* Use a chain of responsibility
* Make approach consistent
* Make sure that their (default content) is always present
* MappingConfig has its own representing class

sjaakd:
refactoring merged into master

PR #2000, opened 2020-01-09
#1998 simplify usage of all mapping annotations

sjaakd:
By making sure that their java counterpart class (standardised as xyzOptions) is always there and takes care on providing the proper value by taking care of the hierarchy via a `Chain Of Responsibility Pattern`

Its a lot of code impact, but it makes the handling easier. Also new hierarchic properties (bean present in config, mapper, beanmapping and mapping/iterablemapping/mapmapping easier). The chain of responsibility takes care on giving back the proper value and you don't need to do this on the place you use it.

It also prepares for replacing the hickory stuff by a home-brew solution and prepares for upcoming PR #1982. All annotations now work in a similar way and can be found at the same place (method, options as `MappingMethodOptions`) 

@Captain1653 .. I guess that also solves a problem you had with #1923 since `@MapperConfiguration` now has a representation as `@MapperConfigOptions`

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -20,6 +20,9 @@ import org.mapstruct.ap.internal.model.assignment.GetterWrapperForCollectionsAnd
 import org.mapstruct.ap.internal.model.assignment.SetterWrapper;
 import org.mapstruct.ap.internal.model.assignment.StreamAdderWrapper;
 import org.mapstruct.ap.internal.model.assignment.UpdateWrapper;
+import org.mapstruct.ap.internal.model.beanmapping.MappingReferences;
+import org.mapstruct.ap.internal.model.beanmapping.PropertyEntry;
+import org.mapstruct.ap.internal.model.beanmapping.SourceReference;
 import org.mapstruct.ap.internal.model.common.Assignment;
 import org.mapstruct.ap.internal.model.common.BuilderType;
 import org.mapstruct.ap.internal.model.common.FormattingParameters;
@@ -27,18 +30,13 @@ import org.mapstruct.ap.internal.model.common.ModelElement;
 import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.common.SourceRHS;
 import org.mapstruct.ap.internal.model.common.Type;
-import org.mapstruct.ap.internal.model.source.BeanMappingOptions;
-import org.mapstruct.ap.internal.model.beanmapping.MappingReferences;
+import org.mapstruct.ap.internal.model.source.DelegatingOptions;
 import org.mapstruct.ap.internal.model.source.Method;
-import org.mapstruct.ap.internal.model.beanmapping.PropertyEntry;
 import org.mapstruct.ap.internal.model.source.SelectionParameters;
-import org.mapstruct.ap.internal.model.beanmapping.SourceReference;
 import org.mapstruct.ap.internal.model.source.selector.SelectionCriteria;
 import org.mapstruct.ap.internal.prism.BuilderPrism;
 import org.mapstruct.ap.internal.prism.NullValueCheckStrategyPrism;
-import org.mapstruct.ap.internal.prism.NullValueMappingStrategyPrism;
 import org.mapstruct.ap.internal.prism.NullValuePropertyMappingStrategyPrism;
-import org.mapstruct.ap.internal.util.MapperOptions;
 import org.mapstruct.ap.internal.util.Message;
 import org.mapstruct.ap.internal.util.NativeTypes;
 import org.mapstruct.ap.internal.util.Strings;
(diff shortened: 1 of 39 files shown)


sjaakd:
See coverage dropped a bit.. Not sure whether I should invest in boosting it.

gunnarmorling:
Wow, PR #2000! Congrats, great milestone 🚀 🎉 🍾 !

sjaakd:
> Wow, PR #2000! Congrats, great milestone 🚀 🎉 🍾 !

Thanks.. I did not even notice it myself initially :smile:

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -218,35 +215,16 @@ public PropertyMappingBuilder forgedNamedBased(boolean forgedNamedBased) {
...
-        public PropertyMappingBuilder nullValueCheckStrategy(NullValueCheckStrategyPrism nvcs ) {
-            this.nvcs = nvcs;
-            return this;
-        }
-
-        public PropertyMappingBuilder nullValuePropertyMappingStrategy( NullValuePropertyMappingStrategyPrism nvpms ) {
-            this.nvpms = nvpms;
+        public PropertyMappingBuilder options(DelegatingOptions options) {
+            this.nvcs = options.getNullValueCheckStrategy();
+            if ( method.isUpdateMethod() ) {
```
filiphr: Why not store the `DelegatingOptions`? Right now this forces us to pass `method` before passing the options and can lead to an NPE.

processor/src/main/java/org/mapstruct/ap/internal/model/source/DefaultOptions.java
```diff
@@ -0,0 +1,128 @@
...
+import org.mapstruct.ap.internal.prism.CollectionMappingStrategyPrism;
+import org.mapstruct.ap.internal.prism.InjectionStrategyPrism;
+import org.mapstruct.ap.internal.prism.MapperPrism;
+import org.mapstruct.ap.internal.prism.MappingInheritanceStrategyPrism;
+import org.mapstruct.ap.internal.prism.NullValueCheckStrategyPrism;
+import org.mapstruct.ap.internal.prism.NullValueMappingStrategyPrism;
+import org.mapstruct.ap.internal.prism.NullValuePropertyMappingStrategyPrism;
+import org.mapstruct.ap.internal.prism.ReportingPolicyPrism;
+
+public class DefaultOptions extends DelegatingOptions {
```
filiphr: What do `DefaultOptions` mean?

processor/src/test/java/org/mapstruct/ap/test/verbose/CreateBeanMapping.java
```diff
@@ -8,7 +8,7 @@
 import org.mapstruct.Mapper;
 import org.mapstruct.factory.Mappers;
 
-@Mapper
+@Mapper( config = CreateBeanMappingConfig.class )
```
filiphr: Why do we need this test change?

sjaakd:
> This looks amazing @sjaakd, really good work on it.

Thanks!

sjaakd:
> One other question. Some of the deleted files have a lot of history. Is it somehow possible to keep that during the git changes? I didn't look at the separate commits, only the whole PR

Thats a bit of a tricky one.. Some classes got their history a bit mixed up.`MappingOptions` is now related to `Mapping` instead to a container which role has been taken by `MappingMethodOptions`.. Don't know how to restore the relations in git to line this up correctly.. Do you?

processor/src/test/java/org/mapstruct/ap/test/verbose/CreateBeanMapping.java
sjaakd: Ah.. Thats an interesting one..It printed the prism in stead of the config value :smile:

processor/src/main/java/org/mapstruct/ap/internal/model/source/DefaultOptions.java
sjaakd: When there's no Mapping, BeanMapping, Mapper, MapperConfig defined, in the end it falls back to what we defined as "default". 

We designed an hierarchy in annotations. Programmatically the question is whether this is correct on a more fundamental level. For instance, we define a `default` in the annotation for the `nullValuePropertyMappingStrategy` in  `@Mapping`, being `NullValuePropertyMappingStrategy nullValuePropertyMappingStrategy()
        default NullValuePropertyMappingStrategy.SET_TO_NULL`

We don't actually use this default, other than for convenience: the user does not have to specify this `nullValuePropertyMappingStrategy` in `@Mapping`. Instead of this, we check whether it was set explicitly or not. If not, we fall "through" the `@Mapping` to the next in line, the `@BeanMapping` were we repeat the same trick. Then to `@Mapper`, `@MappingConfig` `Options` (for some) and in the end.. `DefaultOptions`.. Its a chain-of-responsibility.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: I guess there's room for improvement (luckily). For instance, we only use the `NullValuePropertyMapping` when its an update method. There's another issue, that will align this in 2.0.  This PR was big enough on its own right. Hence I left this out.

sjaakd:
@filiphr .. I answered your questions. Did not do any updates. Restoring the history, I'm not actually sure if that can be done / how to do this. 

Let me know if you agree. We can always take on additional refactoring in the near future. Like to drive this one home.

sjaakd:
> One other question. Some of the deleted files have a lot of history. Is it somehow possible to keep that during the git changes? I didn't look at the separate commits, only the whole PR

I think I managed.. On the whole it still looks like files have been lost. But individual commits show the move actions. So, if we don't squash the PR I guess history will be kept

processor/src/main/java/org/mapstruct/ap/internal/model/source/DefaultOptions.java
filiphr: Makes sense now, thanks

sjaakd:
merged to master

sjaakd:
@filiphr .. thanks!

PR #2007, opened 2020-01-20
Annotation helper

sjaakd:
This is how I would like to integrate the annotation helper on top of #2000 

TODO:
* Remove (code) references to prism (local var's, package name, etc)
* Fix check style
* Include gem-api in minimal package (shade plugin)
* Include released version of annotation helper


processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
```diff
@@ -31,7 +31,7 @@
 public class BeanMappingOptions extends DelegatingOptions {
 
     private final SelectionParameters selectionParameters;
-    private final Optional<BeanMappingGem> beanMapping;
```
sjaakd: Ok.. At the time of writing it felt more or less ok to do this (we had a discussion on whether or not to pass optional as parameter). Now I found some more post and I agree. `Optional` should not be a member. Our 'chain of responsibility' ensures the result of most methods is actually not-null, so there's also no need to return it. Remains, its fluent use.

filiphr: Sorry, I really don't like using `Optional` as class member or passed as parameters. Thanks for understanding

sjaakd: but, that's what I meant.. You're right..

filiphr: I know, just wanted to comment on your comment :)

sjaakd:
> LGTM, nice collaboration on this @sjaakd

It sure is.. Thanks.. enjoyed it.

PR #2015, opened 2020-02-01
695b enum alternative 

sjaakd:


documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc
```diff
@@ -145,15 +145,29 @@ That way it is possible to map arbitrary deep object graphs. When mapping from e
...
-* If source and target attribute type differ, check whether there is another mapping method which has the type of the source attribute as parameter type and the type of the target attribute as return type. If such a method exists it will be invoked in the generated mapping implementation.
-* If no such method exists MapStruct will look whether a built-in conversion for the source and target type of the attribute exists. If this is the case, the generated mapping code will apply this conversion.
-* If no such method was found MapStruct will try to generate an automatic sub-mapping method that will do the mapping between the source and target attributes.
-* If MapStruct could not create a name based mapping method an error will be raised at build time, indicating the non-mappable attribute and its path.
+. If source and target attribute have the same type, the value will be simply copied *direct* from source to target. If the attribute is a collection (e.g. a `List`) a copy of the collection will be set into the target attribute.
```
filiphr: Maybe it would be good to reference the enums here. That way it would be clear from the documentation which one does what. What do you think?

core/src/main/java/org/mapstruct/control/MappingControl.java
```diff
@@ -0,0 +1,73 @@
...
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.CLASS)
+@Repeatable(MappingControls.class)
+@Target( ElementType.ANNOTATION_TYPE )
+@MappingControl( MappingControl.Use.DIRECT )
+@MappingControl( MappingControl.Use.BUILT_IN_CONVERSION )
+@MappingControl( MappingControl.Use.MAPPING_METHOD )
+@MappingControl( MappingControl.Use.COMPLEX_MAPPING )
+public @interface MappingControl {
```
filiphr: Maybe add some javadoc here?

sjaakd: example and java doc added

documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc
sjaakd: done..

sjaakd:
> I like it a lot. Really good job @sjaakd

Thanks.. 

I'll ponder a bit more on meta annotations coming weeks.. :smile:. I have an idea how to do this in a generic way. But first I want to continue on the Map <> Bean thing..

PR #2022, opened 2020-02-06
Updated versions on readme.md to 1.3.1.Final

vegemite4me:
Not the most productive of PRs, but hope it prevents someone bumping into the bug in 1.3.0 to do with field names that start with the three letters 'set'. Thanks for building a great tool.


filiphr:
Thanks a lot @vegemite4me.

>Not the most productive of PRs

Every single PR and contribution is welcomed, it doesn't matter how big or small is.

PR #2031, opened 2020-02-22
#2014 Add support for mapping from Java 14 records

### LINKED ISSUE #2014: Support mapping from Java 14 records (preview feature)

filiphr:
Fixes #2014

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ matrix:
   - jdk: openjdk13
     # There is an issue with the documentation so skip it
     script: mvn -B -V clean install -DskipDistribution=true
+  - jdk: openjdk14
+    # There is an issue with the documentation so skip it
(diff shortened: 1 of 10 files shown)


chris922:
Really cool, @filiphr ! Mapping to records is probably not yet possible, because MapStruct doesn't yet support instantiation via constructor, correct?

filiphr:
Thanks for the review @chris922.

>Mapping to records is probably not yet possible, because MapStruct doesn't yet support instantiation via constructor, correct?

Yes mapping into records is not possible due to the constructor limitation. Once we get that in it would be trivial to add support for mapping into records.

filiphr:
The coverage dropped due to not testing the processor module with Java 14. However, there is an integration test which does the testing of the mapping into records.

PR #2040, opened 2020-03-07
#2021 Generate compilable code when Decorator is nested within the Mapper

### LINKED ISSUE #2021: Wrong decorator class name used when Decorator class is nested inside mapper interface
Example:

```java
@Mapper
@DecoratedWith(MyMapper.Decorator.class)
public interface MyMapper {
    abstract class Decorator implements MyMapper {}
}
```

Generated implementation example:

```java
public class MyMapperImpl extends Decorator implements MyMapper {
...
}
```

Here class `Decorator` is not imported and not qualified by it's outer class name (like `MyMapper.Decorator`), so implementation just doesn't compile.

The right implementation should be:

```java
public class MyMapperImpl extends MyMapper.Decorator implements MyMapper {
...
}
```

filiphr:
We need to treat the import of the decoratorType specially when it is nested.
Calling addIfImportRequired is not the most correct approach since it would
lead to checking if the type is to be imported and that would be false
since the Decorator is a nested class within the Mapper.
However, when generating the Decorator this is not needed, because the Decorator is a top level class itself

In a nutshell creating the Decorator should have its own ProcessorContext, but it doesn't

Fixes #2021 

This was a tricky one. It was affected by the fix for #1460 and #1576

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/Decorator.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/Decorator.java
@@ -125,7 +125,22 @@ public class Decorator extends GeneratedType {
     @Override
     public SortedSet<Type> getImportTypes() {
         SortedSet<Type> importTypes = super.getImportTypes();
-        addIfImportRequired( importTypes, decoratorType );
+        // DecoratorType needs special handling in case it is nested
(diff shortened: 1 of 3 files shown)


filiphr:
Thanks for the review @sjaakd

Issue #1789, opened 2019-04-10
New functionality: Expression support in @ValueMapping


seime:
Think of it as doing search/replace in your text editor with regex support.

The order of processing would be like

- exact matches between source and target
- `@ValueMapping` with exact source and target (inc `NULL`)
- `@ValueMappings` with expressions
- at the end `ANY_REMAINING` or `ANY_UNMAPPED`

Other considerations could be

- expression marker regex () could be omitted - but makes it difficult to later on include a different rule syntax?
- if either source or target is an expression then both should be treated as expressions?

A more or less complete example:

```
package org.mapstruct.ap.test.value;

import org.mapstruct.Mapper;
import org.mapstruct.MappingConstants;
import org.mapstruct.ValueMapping;

@Mapper
public interface ValueExpressionMapper {

    @ValueMapping( source = "SPECIAL", target = "SPESIAL") // Exact mapping
    @ValueMapping( source = "regex( .* ) ", target = "regex ( PREFIX_$1 )") // Will match for RETAIL and map to PREFIX_RETAIL 
    @ValueMapping( source = "regex( .* ) ", target = "regex ( $1_POSTFIX )") // Will match for EXTRA and map to EXTRA_POSTFIX
    @ValueMapping( source = "regex( TRIM_THIS_AWAY_(.*) ) ", target = "regex ( $1 )") // Will match for TRIM_THIS_AWAY_EXPRESS and TRIM_THIS_AWAY_AIR and map to EXPRESS and AIR
    @ValueMapping( source = MappingConstants.NULL, target = "UNSPECIFIED") // Handle non existing source values
    @ValueMapping( source = MappingConstants.ANY_REMAINING, target = "UNSPECIFIED" ) // Handle non existing target values, source will match UNMAPPED
    ExternalExpressionOrderType map(OrderType source);

    // Source types
    public enum OrderType {

        RETAIL,  STANDARD, NORMAL, UNMAPPED, SPECIAL, TRIM_THIS_AWAY_EXPRESS, TRIM_THIS_AWAY_AIR
    }

    // Destination types
    public enum ExternalExpressionOrderType {

        PREFIX_RETAIL, STANDARD_POSTFIX, NORMAL, SPESIAL, UNSPECIFIED, EXPRESS, AIR
    }

}
```

- Initially only a SINGLE `@ValueMapping` with expression would be supported for each mapping method
- Initially no support for nested expressions
- Initially only support for REGEX expressions
[...]

Issue #1667, opened 2018-12-10
Avoid explicit Unrecognized mapping while working with Java Enum & proto buff


Kavyamanjanatha:
If the proto has enum then the generated file has Unregognized option added.However in the Java enum Dto we dont add this.
Due to this need to add explicit mapping of unrecoginzed while translation from java enum to proto enum.

@ValueMapping(source = "UNRECOGNIZED", target = MappingConstants.NULL)
Permission map(PermissionDTO permissionDTO);

Need a way to avoid this mapping either via config files.

chris922:
Do you have an example? Maybe a short extract from `Permission` and `PermissionDTO`?

I think MapStruct already supports what you are searching for or maybe I misunderstood your issue. Have you tried this?

```java
@ValueMapping(source = MappingConstants.ANY_REMAINING, target = MappingConstants.NULL)
Permission map(PermissionDTO permissionDTO);
```

Kavyamanjanatha:
Hi
In your approach developer has to explicitly map for every enum  conversion above lines.
Instead we should have an option to specify these configuration via config file so that  it can be globally applied without developer need to the specify @ValueMapping for every mapping conversion

chris922:
Ah ok, understood.

@filiphr / @sjaakd : WDYT? Useful and where should we add this? I think the right place would be `@Mapper` and `@MapperConfig` on the usual way.. any suggestions for the property name?
```java
@Mapper( defaultValueMapping = @ValueMapping(source = MappingConstants.ANY_REMAINING, target = MappingConstants.NULL) )

@Mapper( valueMappingPolicy = ValueMappingPolicy.ANY_REMAINING_TO_NULL )

...
```

Or is it already possible together with `@InheritConfiguration`? Something like:

```java
@ValueMapping( source = MappingConstants.ANY_REMAINING, target = MappingConstants.NULL )
Enum<?> defaultEnumMapping( Enum<?> source );

@InheritConfiguration
Permission map( PermissionDTO permissionDTO );
```

But in this case you still have to configure it for each enum.

filiphr:
Recently we added new `EnumNamingStrategy` and `EnumTransformationStrategy` that would enable the use case in the issue.

There is no out of the box support for this. However, it is possible to implement new custom `EnumNamingStrategy` and implement the `getDefaultNullEnumConstant` to handle the default value for protobuf enums.

Issue #1220, opened 2017-06-02
[Feature] Pattern to match enum constants


doctor-zoidberg:
I need to map various enums, manually written and generated from WSDL:
```java
    public enum DomAttribute {

        CONSTANT_1,

        CONSTANT_2,

        CONSTANT_3,

        CONSTANT_4,

        // ...
    }
```
```java
    @XmlEnum
    public enum ToAttribute {

        @XmlEnumValue("TO:CONSTANT_1")
        TO_CONSTANT_1("TO:CONSTANT_1"),

        @XmlEnumValue("TO:CONSTANT_2")
        TO_CONSTANT_2("TO:CONSTANT_2"),

        @XmlEnumValue("TO:CONSTANT_3")
        TO_CONSTANT_3("TO:CONSTANT_3"),

        @XmlEnumValue("TO:CONSTANT_4")
        TO_CONSTANT_4("TO:CONSTANT_4"),

        // ...
    }
```
The problem is that the generated enums always have a prefix on their constant names. My current workaround is a custom mapper. However, I don’t like this solution that much. Of course, I could use a large list of `@ValueMappings` / `@ValueMapping` but this is quite tedious.

So I would really appreciate a solution where I could specify some sort of pattern (regex!?) to generate a mapper automatically. Is it possible? Thank you.

filiphr:
There other issues similar to this one. Such as:

* #796 - Which is similar (if not the same as this one). As a solution it proposes using an SPI for matching `Enums`
* #1076 - Which proposes an SPI for general name matching strategies

Will a solution to one of those issues work for you? We are not big fans of wild card matching, but something in the direction of the 2 linked tickets is something interesting for us.

doctor-zoidberg:
I can understand your concerns about wildcard patterns. So, I will look at your proposal for using an SPI and then report back. Thank you.

doctor-zoidberg:
After reading issues #796 and #1076 I would say that an SPI as you prefer would do the trick for me.

Nonetheless, this solution feels a bit odd to me. You have to put a piece of your custom mapping in an external JAR module making it less understandable for another person reading the code. That is why I would prefer an annotation way like @twothe proposed in https://github.com/mapstruct/mapstruct/issues/1076#issuecomment-279988449.

However, as I said, a SPI would do the trick.

filiphr:
[...]

Issue #796, opened 2016-04-06
EnumNaming


maposchmi:
We had to map a lot of Beans to Jaxb-Objects, both containing enums.
Jaxb converts an underscore to 3 underscores:
Enum.class:
FOO_BAR
JaxbEnum.class
FOO___BAR

We didnt want to map all that enum-constants manually.

maposchmi:
So we introduced an EnumNamingStrategy similar to AccessorNamingStrategy to EnumMappingMethod.
Its not perfect, but it works für us.

[mapstruct-1.0.1-modified.zip](https://github.com/mapstruct/mapstruct/files/206232/mapstruct-1.0.1-modified.zip)

sjaakd:
@maposchmi: you are aware that `Mapping` of enums (EnumMappingMethod) is deprecated? Its replaced by `@ValueMapping`..

maposchmi:
We used 1.0.0.Final and i guess that there it is not deprecated.
On the quick i didnt find docu about @ValueMapping

DanielDressler:
So is a solution for this being considered? While newer, ValueMappingMethod still simply matches by exact names and needs to be configured with exact source/target pairs.

Support for JaxB-generated enums could be built-in. I'd probably prefer a "dumber" approach of just ignoring a few typical delimiters and maybe even case when matching enum constants by name. (Which project has a lot of enums containing both A_BC and aB_C?)

More generally, an SPI like the suggested one would help. It could be improved by providing a normalization of names for both the source and target values. The exact name match would then happen on these normalized sets of names. The custom normalization could just strip delimiters, or a prefix, or whatever a project needs.

Even more generally, if #798 results in a very generic name-mapping SPI, that might also work for enum-values.

filiphr:
Recently we added new `EnumNamingStrategy` and `EnumTransformationStrategy` that would enable the use case in the issue.

There are is no out of the box support for replacing 3 underscores with one. However, it is possible to register a custom `EnumTransformationStrategy` that would be able to do that.

PR #2045, opened 2020-03-19
@ValueMapping enum constant naming strategy SPI

seime:
Works for enum2enum, string2enum and enum2string.

Should mostly solve 
- #796 EnumNaming
- #1220 [Feature] Pattern to match enum constants
- #1789 New functionality: Expression support in @ValueMapping (my issue, this PR fully solves the underlying need)

It also allows for solving #1667 as an SPI implementation for which we have done here: https://github.com/entur/mapstruct-spi-protobuf/blob/valuemapping_accessor_spi/spi-impl/src/main/java/no/entur/mapstruct/example/protobuf/ProtobufEnumConstantNamingStrategy.java

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
@@ -9,15 +9,16 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Types;
 
+import org.mapstruct.ap.internal.gem.BeanMappingGem;
 import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.model.source.SelectionParameters;
 import org.mapstruct.ap.internal.model.source.ValueMappingOptions;
-import org.mapstruct.ap.internal.gem.BeanMappingGem;
 import org.mapstruct.ap.internal.util.Message;
 import org.mapstruct.ap.internal.util.Strings;
 
@@ -122,10 +123,23 @@ public class ValueMappingMethod extends MappingMethod {
 
                 // get all target constants
                 List<String> targetConstants = method.getReturnType().getEnumConstants();
+                List<String> mappedTargetConstants = targetConstants.stream()
+                    .map( e -> targetType.getMappedEnumValue( e ) )
+                    .collect( Collectors.toList() );
                 for ( String sourceConstant : new ArrayList<>( unmappedSourceConstants ) ) {
-                    if ( targetConstants.contains( sourceConstant ) ) {
-                        mappings.add( new MappingEntry( sourceConstant, sourceConstant ) );
-                        unmappedSourceConstants.remove( sourceConstant );
+                    String mappedSourceConstant = sourceType.getMappedEnumValue( sourceConstant );
+                    for ( int i = 0; i < mappedTargetConstants.size(); i++ ) {
+                        if ( mappedTargetConstants.get( i ).equals( mappedSourceConstant ) ) {
+                            if ( sourceConstant.equals( mappedSourceConstant ) ) {
+                                // The standard enum value
+                                mappings.add( new MappingEntry( sourceConstant, targetConstants.get( i ) ) );
+                            }
+                            else {
+                                // The mapped enum value
+                                mappings.add( new MappingEntry( sourceConstant, mappedTargetConstants.get( i ) ) );
+                            }
+                            unmappedSourceConstants.remove( sourceConstant );
+                        }
                     }
                 }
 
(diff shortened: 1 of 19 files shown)


seime:
Have you had any change to take a look at this @filiphr ? 
It seems like the last comment by @govindrk in https://github.com/mapstruct/mapstruct/issues/1492 is about exactly the same need?

filiphr:
Hey @seime,

I've had some time to think about this PR and I have some questions.

The protobuf enums, is the `UNKNOWN` enum automatically created or is that some kind of a pattern?

I've also been thinking about the SPI and I am not 100% convinced that it would be useful as we think.

My current thinking is the following:

* Perhaps we need an SPI for default null and any remaining mappings. That way you can define defaults globally for something like protobuf. However, I also liked the suggestion from @chris922 in https://github.com/mapstruct/mapstruct/issues/1667#issuecomment-448966276. That way a default value mapping can be registered for all your value mappings on a global level (no need for SPI). It can also be done per mapper for all your enums there.
* The other hook point is the name transformation (prefix, suffix, etc). For that I have more and more the feeling that the proposed SPI is not the right way. You would want to say on the method level what the prefix / suffix is. This can actually be a different kind of SPI. I'll explain below

For the transformation let's take your example

```
public enum CheeseType {

    BRIE,
    ROQUEFORT
}
public enum CheeseTypeSuffixed {

    DEFAULT_CHEESE_TYPE,
    BRIE_CHEESE_TYPE,
    ROQUEFORT_CHEESE_TYPE,
}
```

Then the mapper can look like:

```
public interface CheeseMapper {

    @EnumMapping(nameTransformFunction = "suffix", configuration = "CHEESE_TYPE")
    CheeseTypeSuffixed map(CheeseType cheese);

    //@EnumMapping(nameTransformFunction = "inverse-suffix", configuration = "CHEESE_TYPE")
    @InheritInverseConfiguration
    CheeseType map(CheeseTypeSuffixed cheese);

    @EnumMapping(nameTransformFunction = "suffix", configuration = "CHEESE_TYPE")
    String map(CheeseType cheese);

    @EnumMapping(nameTransformFunction = "inverse-suffix", configuration = "CHEESE_TYPE")
    CheeseType map(String cheese);
}
```

The generated class will then look like:

```
public class CheeseMapperImpl implements CheeseMapper {

    public CheeseTypeSuffixed map(CheeseType cheese) {
        // null check
        switch( cheese ) {
            case BRIE:
                return CheeseTypeSuffixed.BRIE_CHEESE_TYPE;
            //...
        }
    }

    public CheeseType map(CheeseTypeSuffixed cheese) {
        // null check
        switch( cheese ) {
            case BRIE_CHEESE_TYPE:
                return Cheese.BRIE;
            //...
        }
    }

    @EnumMapping(nameTransformPolicy = "suffix", configuration = "CHEESE_TYPE")
    public String map(CheeseType cheese) {
        // null check
        switch( cheese ) {
            case: BRIE:
                 return "BRIE_CHEESE_TYPE";
            //... 
        }
    }

    @EnumMapping(nameTransformPolicy = "inverse-suffix", configuration = "CHEESE_TYPE")
    public CheeseType map(String cheese) {
        // null check
        switch( cheese ) {
            case: "BRIE_CHEESE_TYPE":
                 return CheeseType.BRIE;
            //... 
        }
    }
}
```

There would be an SPI with out of the box implementation for some transformation policies. Such as `suffix`, `prefix`, not sure if we need something else.

```
public interface EnumNameTransformationFunction {

    String getType();

    String transform(String value, String configuration);

}
```

And some implementations that can look like:

```
public class SuffixEnumNameTransformationFunction implements EnumNameTransformationFunction {

    public String getType() {
        return "suffix";
    }


    public String transform(String value, String configuration) {
        return value + configuration;
    }
}
```

When using `@InheritInverseConfiguration` we are going to invoke the `EnumNameTransformationFunction#transform` with the value from the target enum and the configuration.

We will also treat the transform function prefixed with `inverse-` specially by looking for a function where the `inverse-` prefixed is stripped and applying it on the target enum instead of the source enum.

For `Enum` to `String` and `String` to `Enum` we are going to invoke the appropriate function for the enum only.

In the example above we will invoke `transform` on the `Enum` so we can get the value for the String.

This is just a dump of my ideas. 

Some additional functions can also be added like: `stripSuffix`, `stripPrefix`.

This kind of SPI would open the door to writing custom transformation policies and using them as you see fit in your mappings. Those policies are of course always applied during compilation time.

Would appreciate your feedback @seime and @sjaakd

filiphr:
I did a spike for the enum transformation strategy. Have a look at my branch https://github.com/filiphr/mapstruct/tree/enum-transformation-strategy

filiphr:
I created a PR with my spike as well. Have a look at https://github.com/mapstruct/mapstruct/pull/2089. Would appreciate any feedback.

seime:
@filiphr, thanks for taking your time. Since we now have 2 cases where discussions are taking place (here and #2089) I will answer your question regarding the default value here and continue the discussion in the other PR.

> The protobuf enums, is the UNKNOWN enum automatically created or is that some kind of a pattern?

Protobuf enums are stored internally as `(var)int` values. There are language requirements that a default enum constant with value 0 exists since protobuf does not have `null` values. https://developers.google.com/protocol-buffers/docs/proto3#enum

In addition there is a style guide recommending how to name enum constants; 
1. use the enum name as prefix, ie `CheeseType` constans shall have `CHEESE_TYPE_` as prefix
2. The zero value enum should have the suffix `UNSPECIFIED`
https://developers.google.com/protocol-buffers/docs/style#enums

So there cannot be any non-protobuf enum constant mapping to `null`, you must specify the corresponding `_UNSPECIFIED` constant. The other way around you must map the `_UNSPECIFIED` to `null`.

filiphr:
Closing this since the functionality was added in #2100. Thanks a lot @seime for your work on this it was extremely valuable

PR #2058, opened 2020-04-04
#2056: Handle null TypeMirror in uses and import gracefully

### LINKED ISSUE #2056: "uses" entries ignored when used class is generated during annotation processing
Tested with current master (1.4.0-SNAPSHOT): https://github.com/Chessray/mapstruct-spring-additions/tree/specifyNameForGeneratedBridgeClass

Version 1.3.1.Final throws a NullPointerExsception in this setup. The latest master just behaves as if the `uses` attribute weren't there:
- By default, it just generates a meaningless helper mapping method:

       protected SeatConfigurationDto seatConfigurationToSeatConfigurationDto(SeatConfiguration seatConfiguration) {
        if ( seatConfiguration == null ) {
            return null;
        }

        SeatConfigurationDto seatConfigurationDto = new SeatConfigurationDto();

        return seatConfigurationDto;
       }
- With `unmappedTargetPolicy = ERROR`, annotation processing fails, because there are no mappings for the child attributes in the parent mapper.

filiphr:
Due to a bug in javac (JDK-8229535) for an annotation with Class values, instead of returning a TypeMirror with TypeKind#ERROR the compiler returns the string "<error>". Eclipse doesn't have this problem currently.

Fixes #2056 

This depends on https://github.com/mapstruct/tools-gem/pull/4. We would need a new release of tools-gem to handle this properly.

### DIFF
--- a/integrationtest/src/test/java/org/mapstruct/itest/tests/MavenIntegrationTest.java
+++ b/integrationtest/src/test/java/org/mapstruct/itest/tests/MavenIntegrationTest.java
@@ -135,4 +135,25 @@ public class MavenIntegrationTest {
 
     }
 
+    /**
+     * Tests usage of MapStruct with another processor that generates the uses type of a mapper.
+     */
+    @ProcessorTest(baseDir = "usesTypeGenerationTest", processorTypes = {
+        ProcessorTest.ProcessorType.JAVAC
+    })
+    void usesTypeGenerationTest() {
+    }
+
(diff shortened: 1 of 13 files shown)


filiphr:
@sjaakd can you please also have a look at https://github.com/mapstruct/tools-gem/pull/4

filiphr:
Thanks for the review @sjaakd

PR #2067, opened 2020-04-12
Use exact message for Diagnostic test messages assertions

filiphr:
I've been working on the constructor creation which lead me to do some refactorings in the `TargetReference`, which lead to changing the `messageRegExp` in the `@Diagnostic` so we can make sure that I don't break something by accident. This is step 1.


filiphr:
It took me some time to write this down as well :). Honestly, I've never liked the regex, it is always difficult to see where the error is.

I actually kept the `messageRegExp` only because of 2 places where it was not possible to have a clear exact match .

Thanks a lot for the review

Issue #73, opened 2013-11-24
Make use of constructor arguments when instantiating mapping targets


gunnarmorling:
Some thoughts:
- Need a way to select the constructor to use if there are several ones
- Properties not covered by the chosen constructor should be populated by setters
- Optionally, the mechanism should take parameterized factory methods into account

pardom:
I would also like to see this, specifically for immutable objects. Any update?

gunnarmorling:
No update really, but it's one of the remaining things which I think should be done for 1.0. Would you be interested in giving this a try?

pardom:
I might be. Can you point me in the right direction?

gunnarmorling:
Some helpful docs to get started are:

* the JavaDocs of [MappingProcessor](https://mapstruct.ci.cloudbees.com/job/mapstruct-javadoc/javadoc/index.html?org/mapstruct/ap/MappingProcessor.html) which describe the overall architecture
* the JavaDocs of [org.mapstruct.ap.model](https://mapstruct.ci.cloudbees.com/job/mapstruct-javadoc/javadoc/index.html?org/mapstruct/ap/model/package-summary.html) and [org.mapstruct.ap.processor](https://mapstruct.ci.cloudbees.com/job/mapstruct-javadoc/javadoc/index.html?org/mapstruct/ap/processor/package-summary.html)

For this issue, you'd probably have to adapt `PropertyMapping` (describs the "link" between a source and target property, so this would somehow have to express that a property is propagated by a constructor argument rather than a setter call) and `BeanMappingMethod` (describes one mapping method between two bean types, so this would have to express which constructor of the target type to invoke).

Hope this helps for a start, let me know in case you run into any road blocks. Thanks!

pardom:
Hmm, not sure I have the time to spend on this right now, so I can't promise I'll get it done. This feature would satisfy my immutability requirement though, so kudos to whoever does it. :+1:

mozinrat:
I would love to use this feature if its there, can anyone point me about whats being done.

gunnarmorling:
@mozinrat no one found the time to implement it yet unfortunately. If you are interested and would like to help out with building it, let me know and I will help you to get running with Hacking on MapStruct.

mozinrat:
@gunnarmorling whats the plan for this, lets connect on some platform email etc and discuss the possibility. I can give it a week or so but its hard for me to give any more time. Forking the repo meanwhile...:)

[...]

PR #2068, opened 2020-04-12
Refactor target reference to report errors during bean mapping instead of creation of the target reference

### LINKED ISSUE #2069: TargetReference refactorings
In order to easy the implementation of #73 we need to do some refactorings in `TargetReference`. The main goal is to move the error reporting into the bean mapping instead of all target references in the `targetName`. The `TargetReference` should only contain the target property names as string, nothing else.

filiphr:
With this we can more easily go in the direction of using constructor to map into target beans.

This is a step towards #73 

In this PR all tests are passing green (had to change some, but I think that this is OK).

There is an alternative approach to this. That would be to have dedicated `ForgedMappingReferences` from which we would be able to create the new `MappingReferences` like we are doing for the `SourceMethod`. Basically the problem is that we first need to get the accessors and result type and then construct the `TargetReference` 

Fixes #2069


processor/src/test/java/org/mapstruct/ap/test/namesuggestion/SuggestMostSimilarNameTest.java
```diff
@@ -66,13 +66,21 @@ public void testNameSuggestion() {
             @Diagnostic(type = PersonGarageWrongTargetMapper.class,
                 kind = javax.tools.Diagnostic.Kind.ERROR,
                 line = 19,
-                message = "Unknown property \"garage.colour.rgb\" in result type org.mapstruct.ap.test.namesuggestion" +
-                    ".Person. Did you mean \"garage.color\"?"),
+                message = "Unknown property \"colour\" in type org.mapstruct.ap.test.namesuggestion.Garage for target" +
+                    " name \"garage.colour.rgb\". Did you mean \"garage.color\"?"),
+            @Diagnostic(type = PersonGarageWrongTargetMapper.class,
+                kind = javax.tools.Diagnostic.Kind.WARNING,
```
sjaakd, line 72: suddenly.. more WARNINGS?

build-config/src/main/resources/build-config/checkstyle.xml
```diff
@@ -122,7 +122,9 @@
 
         <!-- Checks for Size Violations.                    -->
         <!-- See http://checkstyle.sf.net/config_sizes.html -->
-        <module name="MethodLength"/>
+        <module name="MethodLength">
+            <property name="max" value="200"/>
```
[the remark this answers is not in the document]
filiphr: haha, maybe even completely remove it? I am all OK for checks and the likes, but it should not be blocking us

processor/src/test/java/org/mapstruct/ap/test/namesuggestion/SuggestMostSimilarNameTest.java
filiphr: Before the entire `TargetReference` for `garage.colour.rgb` was invalid which meant that handling the defined mappings led to an hour which stopped the creating of the `BeanMappingMethod` and this not reporting the unmapped target.

However, with the new change the target reference does not have `valid` property anymore, so the defined mappings would not report this error and thus we would reach the unmapped target properties reporting. 

IMO this is a good thing, since you have more warnings / errors that you can look at initially

filiphr:
Thanks a lot for the review @sjaakd.

Answers to your questions inline

> 1. The whole purpose is to delay checking of the stuff in `TargetReference` to the moment that we do `BeanMapping` validations right?

Yes the idea was that the do the checking of the stuff in `TargetReference` when we have all accessors (the constructor accessors will not be coming from the `Type` itself). Currently the `TargetReference` is actually created within the `BeanMappingMethod.Builder`, so perhaps I can still move the validation into it and create them once I have all builders. Would need to change the nested mapping a bit for that, but I think it is OK.

> 2. I see you dropped `TargetReference` but it had an abstract base class `AbstractReferenc` (shared with `SourceReference` .. is this still needed?

What do you mean I dropped? I am no longer inheriting from the `AbstractReference` since I think that it is no longer necessary. I also found it a bit confusing since there was a mix of accessors that that were depending on the `TargetReference` or `SourceReference`.


> 3. Do I understand you correctly and you replaced the `PropertyEntries` by a `String`.  should we do the same at the source side?

Yes that is exactly what I did replacing the property entries with a collection of `String`. I was thinking to do the same for the source side indeed. However, this was complex enough and I didn't want to make the PR and change even bigger.

sjaakd:
> What do you mean I dropped? I am no longer inheriting from the `AbstractReference` since I think that it is no longer necessary. I also found it a bit confusing since there was a mix of accessors that that were depending on the `TargetReference` or `SourceReference`

No.. I formulated it a bit crummy.. Sorry. You removed the abstract class `AbstractReference`.. There's no reason to keep this one around, since its sole purpose now is `SourceReference`. So, it can be integrated within this class. That was the geste of this remark.

filiphr:
OK Now it is clear. 

Yes you are completely right. However, I'd like to keep that separate. I created https://github.com/mapstruct/mapstruct/issues/2070 to track the `SourceReference` refactorings.

Issue #2077, opened 2020-04-25
NullPointer when accidentally using String source


sjaakd:
```java
@Mapper
public interface MyMapper {

    class Target {

        private String s1;

        public String getS1() {
            return s1;
        }

        public void setS1(String s1) {
            this.s1 = s1;
        }

    }

    @Mapping( target = "s1")
    Target map( String source);
}
```

filiphr:
Fixed in https://github.com/mapstruct/mapstruct/commit/7c62aec28195950ac6feafd4b2345a8c06ecf6c8

PR #2078, opened 2020-04-25
#2077 nullpointer due to no-getter source

sjaakd:


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -721,30 +721,33 @@ else if ( mapping.getJavaExpression() != null ) {
                     sourceRef = getSourceRef( method.getSourceParameters().get( 0 ), targetPropertyName );
                 }
 
-                if ( sourceRef.isValid() ) {
+                if ( sourceRef != null ) {
+                    // sourceRef == null is not considered an error here
```
filiphr, line 725: Maybe add an example when this can happen. For example for `@Mapping(target = "name", defaultValue = "test")`. Perhaps update the test since I was confused initially :).

sjaakd: target without source can happen in many ways :smile: . It's a hook. In this case to apply a qualifier on a name based mapping. An example is a good idea.. but not as such scope of this PR.. Its already there since ages..

I'll add a separate PR for that..

Need to check why it's not compiling..

PR #2089, opened 2020-05-03
Add EnumTransformationStrategy SPI 

filiphr:
Add a new custom `EnumTransformationStrategy` SPI which can be used for providing custom way of name based mapping for enums.

With this it would be possible to achieve what is needed in #796, #1220, #1789.

Documentation is still missing, but if we agree on the SPI then I can add that as well.

@seime would appreciate your feedback on this as well

### DIFF
--- a/core/src/main/java/org/mapstruct/EnumMapping.java
+++ b/core/src/main/java/org/mapstruct/EnumMapping.java
@@ -12,7 +12,7 @@ import java.lang.annotation.Target;
 
 /**
  * Configured the mapping between two value types.
- * <p><string>Example:</string> Using a suffix for enums</p>
+ * <p><strong>Example:</strong> Using a suffix for enums</p>
  * <pre><code class='java'>
  * public enum CheeseType {
  *     BRIE,
--- a/core/src/main/java/org/mapstruct/EnumMapping.java
+++ b/core/src/main/java/org/mapstruct/EnumMapping.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * @author Filip Hrisafov
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.CLASS)
+public @interface EnumMapping {
+
+    String nameTransformStrategy();
+
+    String configuration();
+}
--- a/core/src/main/java/org/mapstruct/EnumMapping.java
+++ b/core/src/main/java/org/mapstruct/EnumMapping.java
@@ -11,13 +11,113 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
+ * Configured the mapping between two value types.
+ * <p><string>Example:</string> Using a suffix for enums</p>
+ * <pre><code class='java'>
+ * public enum CheeseType {
+ *     BRIE,
+ *     ROQUEFORT
+ * }
+ *
+ * public enum CheeseTypeSuffixed {
+ *     BRIE_TYPE,
+ *     ROQUEFORT_TYPE
+ * }
+ *
+ * &#64;Mapper
+ * public interface CheeseMapper {
+ *
+ *     &#64;EnumMapping(nameTransformationStrategy = "suffix", configuration = "_TYPE")
+ *     CheeseTypeSuffixed map(Cheese cheese);
+ *
+ *     &#64;InheritInverseConfiguration
+ *     Cheese map(CheeseTypeSuffixed cheese);
+ *
+ * }
+ * </code></pre>
+ * <pre><code class='java'>
+ * // generates
+ * public class CheeseMapperImpl implements CheeseMapper {
+ *
+ *     &#64;Override
+ *     public CheeseTypeSuffixed map(Cheese cheese) {
+ *         if ( cheese == null ) {
+ *             return null;
+ *         }
+ *
+ *         CheeseTypeSuffixed cheeseTypeSuffixed;
+ *
+ *         switch ( cheese ) {
+ *             case BRIE:
+ *                 cheeseTypeSuffixed = CheeseTypeSuffixed.BRIE_TYPE;
+ *                 break;
+ *             case ROQUEFORT:
+ *                 cheeseTypeSuffixed = CheeseTypeSuffixed.ROQUEFORT_TYPE;
+ *                 break;
+ *             default:
+ *                 throw new IllegalArgumentException( "Unexpected enum constant: " + cheese );
+ *         }
+ *
+ *         return cheeseTypeSuffixed;
+ *     }
+ *
+ *     &#64;Override
+ *     public Cheese map(CheeseTypeSuffixed cheese) {
+ *         if ( cheese == null ) {
+ *             return null;
+ *         }
+ *
+ *         CheeseType cheeseType;
+ *
+ *         switch ( cheese ) {
+ *             case BRIE_TYPE:
+ *                 cheeseType = CheeseType.BRIE;
+ *                 break;
+ *             case ROQUEFORT_TYPE:
+ *                 cheeseType = CheeseType.ROQUEFORT;
+ *                 break;
+ *             default:
+ *                 throw new IllegalArgumentException( "Unexpected enum constant: " + cheese );
+ *         }
+ *
+ *         return cheeseType;
+ *     }
+ * }
+ * </code></pre>
+ *
  * @author Filip Hrisafov
+ * @since 1.4
  */
 @Target(ElementType.METHOD)
 @Retention(RetentionPolicy.CLASS)
 public @interface EnumMapping {
 
-    String nameTransformStrategy();
+    /**
+     * Specifies the name transformation strategy that should be used for implicit mapping between enums.
+     * Known strategies are:
+     * <ul>
+     *     <li>{@link MappingConstants#SUFFIX_TRANSFORMATION} - applies the given {@link #configuration()} as a
+     *     suffix to the source enum</li>
+     *     <li>{@link MappingConstants#STRIP_SUFFIX_TRANSFORMATION} - strips the the given {@link #configuration()}
+     *     from the end of the source enum</li>
+     *     <li>{@link MappingConstants#PREFIX_TRANSFORMATION} - applies the given {@link #configuration()} as a
+     *     prefix to the source enum</li>
+     *     <li>{@link MappingConstants#STRIP_PREFIX_TRANSFORMATION} - strips the given {@link #configuration()} from
+     *     the start of the source enum</li>
+     * </ul>
+     *
+     * It is possible to use custom name transformation strategies by implementing the {@code
+     * EnumTransformationStrategy} SPI.
+     *
+     * @return the name transformation strategy
+     */
+    String nameTransformationStrategy();
 
+    /**
+     * The configuration that should be passed on the appropriate name transformation strategy.
+     * e.g. a suffix that should be applied to the source enum when doing name based mapping.
+     *
+     * @return the configuration to use
+     */
     String configuration();
 }
--- a/documentation/src/main/asciidoc/chapter-8-mapping-values.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-8-mapping-values.asciidoc
@@ -159,3 +159,106 @@ MapStruct supports enum to a String mapping along the same lines as is described
 2. Similarity: `<ANY_UNMAPPED`> stops after handling defined mapping and proceeds to the switch/default clause value.
 3. Similarity: `<ANY_REMAINING>` will create a mapping for each target enum constant and proceed to the switch/default clause value.
 4. Difference: A switch/default value needs to be provided to have a determined outcome (enum has a limited set of values, `String` has unlimited options). Failing to specify `<ANY_REMAINING>` or `<ANY_UNMAPPED`> will result in a warning.
+
+=== Custom name transformation
+
+When no `@ValueMapping`(s) are defined then each constant from the source enum is mapped to a constant with the same name in the target enum type.
+However, there are cases where the source enum needs to be transformed before doing the mapping.
+E.g. a suffix needs to be applied to map from the source into the target enum.
+
+.Enum types
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+public enum CheeseType {
+
+    BRIE,
+    ROQUEFORT
+}
+
+public enum CheeseTypeSuffixed {
+
+    BRIE_TYPE,
+    ROQUEFORT_TYPE
+}
+----
+====
+
+.Enum mapping method with custom name transformation strategy
+====
+[source, java, linenums]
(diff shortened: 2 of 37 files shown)


processor/src/main/java/org/mapstruct/ap/spi/SuffixEnumTransformationStrategy.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+package org.mapstruct.ap.spi;
+
+/**
+ * @author Filip Hrisafov
+ */
+public class SuffixEnumTransformationStrategy implements EnumTransformationStrategy {
+
+    @Override
+    public String getType() {
```
chris922, line 14: Maybe a different name then `type`? As far as I understood this has to be the same value than `nameTransformStrategy` in the new `EnumMapping` annotation?

core/src/main/java/org/mapstruct/EnumMapping.java
```diff
@@ -0,0 +1,23 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * @author Filip Hrisafov
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.CLASS)
+public @interface EnumMapping {
```
chris922, line 18: What are the reasons to have a new annotation and not reusing `ValueMapping`? Isn't `ValueMapping` also meant for enumerations?

core/src/main/java/org/mapstruct/EnumMapping.java
```diff
@@ -0,0 +1,23 @@
...
+
+    String nameTransformStrategy();
```
chris922, line 20:
```suggestion
    String nameTransformationStrategy();
```

?

processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
```diff
@@ -61,7 +67,18 @@ public Builder valueMappings(List<ValueMappingOptions> valueMappings) {
...
 
-        public ValueMappingMethod build( ) {
+        public Builder enumMapping(EnumMappingOptions enumMapping) {
+            this.enumMapping = enumMapping;
+            return this;
+        }
+
+        public ValueMappingMethod build() {
+
+            if ( !enumMapping.isValid() ) {
```
chris922, line 77: Null check required or could it never be null when `build` is called? The `initializeEnumTransformationStrategy` that will be called after this if-clause includes a `null` check

filiphr:
Thanks a lot for the feedback @chris922.

You might be right about the `EnumContext`. However, I am not sure whether i'd like that door opened. It make your mapping really complex and in the end you can easily write your own implementation of the method without any annotations. I don't want to make this more complex that it should. I'd like to cover 80% of the cases, vs 100. 

In the end it is also possible for people to write their own Annotation Processor that would generate a MapStruct mapper, or an implementation of a value mapping.

I'll answer your other comments separately.

processor/src/main/java/org/mapstruct/ap/spi/SuffixEnumTransformationStrategy.java
filiphr: Yes this is the name / type of the `strategy` that you can reference in `nameTransformStrategy`. What would you propose? I of course need to add Javadoc

core/src/main/java/org/mapstruct/EnumMapping.java
filiphr: I was thinking the same thing. The thing with `@ValueMapping` is that it is closer to `@Mapping`. It has a `source` and `target`. The new `@EnumMapping` is closer to `@BeanMapping`. The idea is to be able to configure general mapping only once for value mappings

core/src/main/java/org/mapstruct/EnumMapping.java
filiphr: The suggestion makes sense. I'll do it

processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
filiphr: Good spot. I'll need to verify this again. In theory it can never be `null`, since we always create it (like the `BeanMappingOptions`). All tests are green as well, so the if condition is never hit it seems.

processor/src/main/java/org/mapstruct/ap/spi/SuffixEnumTransformationStrategy.java
chris922:
```suggestion
    public String getStrategyName() {
```

Maybe?

core/src/main/java/org/mapstruct/EnumMapping.java
chris922: Ahh, correct! `@ValueMapping` could be used multiple times

filiphr:
I've added javadoc, documentation and applied some of your comments @chris922

chris922:
> in the end you can easily write your own implementation of the method without any annotations.

Correct, but isn't the idea of this new `EnumMapping` to have a mapping approach that could be used by several mappings? Regarding the own mapping methods you have to create a new one for each enumeration (... or maybe map from `Enum` to `Enum`).

> I don't want to make this more complex that it should. I'd like to cover 80% of the cases, vs 100.

I am totally with you to cover the base cases .. as long as covering much more would be really complex, but I think in this case it isn't really complex to just pass-through more information to the transformation strategy?

But it's up to you to decide this, I am fine with both solutions.. just thought this might be a quick win :)

processor/src/main/java/org/mapstruct/ap/spi/SuffixEnumTransformationStrategy.java
filiphr: Good idea. Will apply it

chris922:
Just had a deeper look into the other PR regarding this topi and there you also had an example where `@InheritInverseConfiguration` will be used (https://github.com/mapstruct/mapstruct/pull/2045#issuecomment-622363585)

Should we support this here as well? In this case you probably not just need `transform`, but also something like `inverseTransform` or each strategy can define something like an inverse-strategy.. and in case it is not defined (or `inverseTransform` is empty) an `UnsupportedOperationException` could be thrown.. I could image there might be some transformations where it will not be possible to undo it

filiphr:
>Correct, but isn't the idea of this new `EnumMapping` to have a mapping approach that could be used by several mappings? Regarding the own mapping methods you have to create a new one for each enumeration (... or maybe map from `Enum` to `Enum`).

Currently you'll need to apply `@EnumMapping` to the places that you want to apply the strategy. Perhaps using it as a meta annotation would be the way to go since you would be able to define one annotation and use it in your methods. I am also not sure if all your enum mappings have the same strategy applied, maybe some need prefix, some others suffix etc. And most of the time not all have the same prefix / suffix.

>I am totally with you to cover the base cases .. as long as covering much more would be really complex, but I think in this case it isn't really complex to just pass-through more information to the transformation strategy?

This makes it more complex with inverse configuration. Since you'll need to know that and we'll need to give that information to the transformation. Not sure how I like that.

Regarding the `@InheritInverseConfiguration`. Initially I thought that we would need `inverseTransform`. However, during implementation I realised that we can just apply the strategy to the target enum instead of the source enum. Have a look at:

https://github.com/mapstruct/mapstruct/pull/2089/files#diff-281a6edb3173036b540a79288468f9ddR157-R186

This simplifies things quite a bit, since there is nothing to reverse, just checking against different target constants.

filiphr:
Just FYI, this PR does not 100% replace PR #2045. This only adds a transformation strategy. It doesn't affect the default any remaining, etc.

documentation/src/main/asciidoc/chapter-8-mapping-values.asciidoc
```diff
@@ -159,3 +159,106 @@ MapStruct supports enum to a String mapping along the same lines as is described
...
+        return cheeseType;
+    }
+}
+----
+====
+
+MapStruct provides the following out of the box enum name transformation strategies:
+
+* _suffix_ - Applies a suffix on the source enum
+* _stripSuffix_ - Strips a suffix from the source enum
```
sjaakd, line 259: is this the inverse? What happens with inverse mappings?

processor/src/main/java/org/mapstruct/ap/internal/model/source/EnumMappingOptions.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public String getNameTransformationConfiguration() {
+        return enumMapping.configuration().get();
+    }
+
+    public boolean isInverse() {
+        return inverse;
+    }
+
+    public EnumMappingOptions inverse() {
```
sjaakd, line 55: ah.. here it is..

seime:
@filiphr I cloned your code and made a testcase for protobuf mapping with a handcrafted protobuf-like enum; https://github.com/filiphr/mapstruct/compare/enum-transformation-strategy...entur:enum-transformation-strategy-protobuf?expand=1

The included prefix/postfix mappers do work fine, but IMHO they do not scale very well in the sense that one would need to define each and every enum mapping - both ways:

```
@EnumMapping(nameTransformationStrategy = MappingConstants.PREFIX_TRANSFORMATION,
        configuration = "CHEESE_TYPE_PROTOBUF_")
@ValueMapping(source =  MappingConstants.NULL, target = "CHEESE_TYPE_PROTOBUF_UNSPECIFIED")
CheeseTypeProtobuf map(CheeseType cheese);

@InheritInverseConfiguration
@ValueMapping(source = "UNRECOGNIZED", target = MappingConstants.NULL)
CheeseType mapInheritInverse(CheeseTypeProtobuf cheese);
```

This means 6 lines of code for each enum to enum mapping plus the fact that whenever a new enum is added the mapping interface must be updated (I have 300+ enums to be mapped).

The proposed solution in #2045 is able to handle these mappings automatically. By adding an SPI implementation like https://github.com/entur/mapstruct-spi-protobuf/blob/valuemapping_accessor_spi/spi-impl/src/main/java/no/entur/mapstruct/example/protobuf/ProtobufEnumConstantNamingStrategy.java _all_ protobuf enums following the style guide is handled automatically.

Similar implementations could be added for other candidates such as JAXB enums and other generated classes that follow certain patterns. 

My opinion is that both the `@EnumMapping` functionality as well as an SPI like the one proposed in #2045 should be available.

chris922:
@filiphr hmm.. for a SPI that also covers something like it was done here https://github.com/entur/mapstruct-spi-protobuf/blob/valuemapping_accessor_spi/spi-impl/src/main/java/no/entur/mapstruct/example/protobuf/ProtobufEnumConstantNamingStrategy.java you have to expose the source/target type to the `transform` method so that the implementation is able to do the "enum name to upper-camelcase prefix" stuff.

And if there is then also a possibility to define a default `@EnumMapper` @seime should be able to do his magic for all the 300+ enums.

Or do you have any better way?


@seime : Really cool that you verified this SPI approach with your requirements!

seime:
@chris922 yes if we could add a default value mapping with a richer SPI that also includes the null and default value handling (something similar to the one proposed here which is the interface implemented by the protobuf implementation; https://github.com/mapstruct/mapstruct/blob/9bffc88e4ee58b7a6810b57b5610ac1d0cad5137/processor/src/main/java/org/mapstruct/ap/spi/EnumConstantNamingStrategy.java) - I think we can solve several of the issues mentioned.

documentation/src/main/asciidoc/chapter-8-mapping-values.asciidoc
filiphr: Yes and no. I've implemented the reverse a tiny bit different. Have a look at 

https://github.com/mapstruct/mapstruct/pull/2089/files#diff-281a6edb3173036b540a79288468f9ddR157-R186

Basically, I am applying the normal strategy on the target enum instead of the source enum. That way no special handling is needed

filiphr:
Thanks for the review @seime.

As I said this does not cover the case for default `null`, any remaining etc. mappings. I wanted to separate the different ideas I had in my head.

>This means 6 lines of code for each enum to enum mapping plus the fact that whenever a new enum is added the mapping interface must be updated (I have 300+ enums to be mapped).

OK you are right about this. having 300+ enums with the same pattern would make the functionality here cumbersome. However, I'll have to say it is not 6 lines, but 1 line per enum (2 if you consider the `@InheritInverseConfiguration`. I am only talking about the name mapping, not the null, any remaining etc. mappings.

Having said this, I understand your point and yes I agree that having a something like your proposed SPI would also be beneficial.

Once @sjaakd addresses my last reply to his comment I'll merge this PR. Since I think it can be beneficial for some of the other use cases in the issues. 

However, I will have a look at your proposed SPI again and try to think about it from a different angle. Your explanation in https://github.com/mapstruct/mapstruct/pull/2045#issuecomment-623400680 for the Protobuf style guide makes sense to what you are saying. Most likely we won't be able to offer out of the box support for the style guide suggested mapping, but you would be able to do it yourself like in the example you provided.

filiphr:
@sjaakd I forgot to comment on 

>Anyways: a general comment. I remember we have a leftover in value mapping. There's the possibility to add a `@BeanMapping` annotation. This is a good moment to remove that one. It probably gives false expectations and is intended for other methods.

I think that we in generally look at `@MapMapping`, `@IterableMapping`, `@BeanMapping` on all methods. I don't think we do any special checks. I would like to do that separately

filiphr:
@seime I did an iteration of the needs you had in https://github.com/filiphr/mapstruct/commit/d4a881b4f822688c164084be743651f8c7fede4f. I'll create a PR with that once this one is merged. Would appreciate any feedback you have.

sjaakd:
> I think that we in generally look at `@MapMapping`, `@IterableMapping`, `@BeanMapping` on all methods. I don't think we do any special checks. I would like to do that separately

 I would like some changes there as well. One thought I had was in our GemMapping, to add a specific annotation to an Annotation.property for which the value is taken as the value of the Annotation.property and can be used stand-alone wherever the property is present. That means we don't have to change an leave it to GemGeneration to fix this for us.

E.g. `@GemDefinition( value = BeanMapping.class, properties = { @GemProperty( name = "resultType", annotation = `ResultType.class` )  }  )`

And then it would be usable at the same places as were you can apply `@BeanMapping`.. 

Anyway.. just a thought.

----

The reason I dislike `@BeanMapping` here is that we now have `@EnumMapping`. A value mapping is not a bean mapping. It does not make sense. 

particularly this piece of code: 
```java

        private SelectionParameters getSelectionParameters(Method method, Types typeUtils) {
            BeanMappingGem beanMapping = BeanMappingGem.instanceOn( method.getExecutable() );
            if ( beanMapping != null ) {
                List<TypeMirror> qualifiers = beanMapping.qualifiedBy().get();
                List<String> qualifyingNames = beanMapping.qualifiedByName().get();
                TypeMirror resultType = beanMapping.resultType().get();
                return new SelectionParameters( qualifiers, qualifyingNames, resultType, typeUtils );
            }
            return null;
        }
```

What's a result type doing here? And I guess `qualifiers` and `qualifyingnames`  are used seldomly / never with a value mapping methods and if so, they should be part of the new `@EnumMapping`

filiphr:
>The reason I dislike `@BeanMapping` here is that we now have `@EnumMapping`. A value mapping is not a bean mapping. It does not make sense.

That's 100% true and we should have a compile error if you do that since it is never used. We need to this separately. 

Can you please create issues from the things you think we need to go over separely?

I'll continue to merge this.

filiphr:
Thanks everyone for the reviews

Issue #2092, opened 2020-05-10
@InheritConfiguration and @BeanMapping: ignoreUnmappedSourceProperties are not "merged" as expected


drahkrub:
First of all: mapstruct is awesome! I use it to map a really complex hibernate domain model to simple DTOs and vice versa - works great!

I work with a strict MapperConfig regarding source and target properties to catch changes if the domain model changes:
```
@MapperConfig(
        unmappedSourcePolicy = ReportingPolicy.ERROR,
        unmappedTargetPolicy = ReportingPolicy.ERROR
)
public interface CentralConfig {}
```
For this reason, I have to work with `@BeanMapping(ignoreUnmappedSourceProperties = {...})` a lot. Unfortunately this does not work well if `@InheritConfiguration` is used: The second method `mapB` "forgets" that the source property "a2" should be ignored, as defined for the first method `mapA` which configuration is inherited by `mapB`.

Am I doing something wrong or am I expecting something that does not (yet) work?
```
@Mapper(
        config = CentralConfig.class
)
public interface MyMapper {

    @Mapping(source = "a1", target = "aa1")
    @BeanMapping(ignoreUnmappedSourceProperties = {"a2"})
    TargetBeanA mapA(BeanA beanA);

    @InheritConfiguration
    @Mapping(source = "b1", target = "bb1") // inheritance works as expected
    @BeanMapping(ignoreUnmappedSourceProperties = {"b2"}) // compiler error: Unmapped source property: "a2"
    TargetBeanB mapB(BeanB beanB);
}
```
Below all the beans:
```
@Getter
@Setter
public class BeanA {
    
    private int a1;
    private int a2;
}
```

```
@Getter
@Setter
public class BeanB extends BeanA {

    private int b1;
    private int b2;
}
```

```
@Getter
@Setter
public class TargetBeanA {
    
    private int aa1;
}
```

```
@Getter
@Setter
public class TargetBeanB extends TargetBeanA {
    
    private int bb1;
}
```

chris922:
Thanks @drahkrub ,

will `a2` be ignored if you don't define `ignoreUnmappedSourceProperties = {"b2"}` on `mapB`?

I would like to figure out if this is a merging issue or maybe an inheritance issue. `ignoreUnmappedSourceProperties` was added in 1.3 and maybe it was missed to consider it during inheritance.

drahkrub:
Yes, `a2` is ignored in that case:
```java
@Mapper(
[...]

Issue #1152, opened 2017-03-22
MapStruct should have means to mark getters / setters as going to be ignored


sjaakd:
Suppose a reusable source dto:

```java
public class MyDate {

    private Date date;

    public Date getDate() {
        return date;
    }

    public void setDate(Date date) {
        this.date = date;
    }

    public boolean isValid() {
        return date != null;
    }
}

```

Especially with the new nested mapping, I want to have some means that the `isValid()` is **not** considered a source accessor if by any chance somewhere in the structure of the target is a property `valid`. 

Something like this:

```java
public class MyDate {

    private Date date;

    public Date getDate() {
        return date;
    }

    public void setDate(Date date) {
        this.date = date;
    }

   @IgnoreAccessor
    public boolean isValid() {
        return date != null;
    }
}

```

filiphr:
I think that we can do this and #1151 together. We can do the implementation flexible so that it uses an SPI (maybe even a list of SPIs) that provides a way for the user to give multiple annotations that can be used to ignore getter / setter. This way they can use their own annotations and can for example in one go ignore `@Transient` getters. 

In the end `@Deprecated` is just another annotation that will be plugged into this SPI.

agudian:
> Especially with the new nested mapping, I want to have some means that the isValid() is not considered a source accessor if by any chance somewhere in the structure of the target is a property valid.

I don't see how this is different from not having the new auto mapping... If I don't want a property `valid` to be set from outside, I wouldn't create a public setter for it... Or I would use `@Mapping` annotations on the methods directly or for repeated stuff in `@MapperConfig`. That's much more in line of what we offered to do in the past.

But I like the idea of providing means to do it with an SPI, if there is really some special need for this.

gunnarmorling:
Hey @sjaakd any further insights on why @agudian's proposal above isn't good enought (using `@Mapping#ignore()`?

sjaakd:
> I don't see how this is different from not having the new auto mapping...

Well. when I wrote this I spent half an hour troubleshooting on a mapping that MapStruct tried to make on `java.util.date` to our own type which was created 4 years ago and used in a lot of other projects. It happened on a nesting level so it was not immediately obvious. 

[...]

PR #2096, opened 2020-05-14
 WIP: #2092 merge unmappedSourceProperties of BeanMappings

chris922:
I tackled #2092 and it was more complicated than I thought. Maybe it is even though not a bug, but now an enhancement.

Starting point are the changes in `MappingMethodOptions`.

I noticed that `@BeanMapping` will not be merged at all and in case a method is annotated with `@BeanMapping` and `@InheritConfiguration` the `@BeanMapping` will not be inherited and/or merged.

Therefore I implemented inheritance by following the pattern used for `DelegatingOptions`. The main idea is: If a method inherits the configuration and both methods are annotated with `@BeanMapping` I set the inherited `@BeanMapping` as the next `DelegatingOptions`.
This will ensure that all options of the other `@BeanMapping` will be inherited.

For the `ignoreUnmappedSourceProperties` stuff I had to implement a logic to join the lists.

I added the same logic for `@MapMapping` and `@IterableMapping` as well.

The following questions came up during the analyze and implementation:
1. Should we join `ignoreUnmappedSourceProperties` at all? Imho we should do this because I probably would also expect this. Nevertheless there might be some reasons that the method inheriting the configuration should have an own configuration of these properties
2. I am not sure if I understood the `SelectionParameters` correctly, right now I try to join the qualifiers, but maybe that is not the right way and they shouldn't be joined
3. Is it fine to change the `next` of the `BeanMappingOptions`? Could it be that there will then be a totally different inheritance flow? Looked for me that the delegating options works in the way that `next` in this case is the `@Mapper` config, then the `@MapperConfig` and so on. Now I change `next` to the other `@BeanMapping` and maybe this will lead to a different `@Mapper` config? Could this be?!

I marked this PR as WIP to get some first feedback from you, if everything is fine I would like to update the documentation to highlight that the properties will be merged

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
@@ -5,21 +5,22 @@
  */
 package org.mapstruct.ap.internal.model.source;
 
-import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
 
-import org.mapstruct.ap.internal.model.common.TypeFactory;
 import org.mapstruct.ap.internal.gem.BeanMappingGem;
 import org.mapstruct.ap.internal.gem.BuilderGem;
 import org.mapstruct.ap.internal.gem.NullValueCheckStrategyGem;
 import org.mapstruct.ap.internal.gem.NullValueMappingStrategyGem;
 import org.mapstruct.ap.internal.gem.NullValuePropertyMappingStrategyGem;
+import org.mapstruct.ap.internal.model.common.TypeFactory;
 import org.mapstruct.ap.internal.util.FormattingMessager;
 import org.mapstruct.ap.internal.util.Message;
 import org.mapstruct.tools.gem.GemValue;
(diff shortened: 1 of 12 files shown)


filiphr:
Thanks for taking over this one @chris922.

Let me try to answer the points one by one. I also haven't looked at the code yet. Talking conceptually right now.

>1. Should we join `ignoreUnmappedSourceProperties` at all? Imho we should do this because I probably would also expect this. Nevertheless there might be some reasons that the method inheriting the configuration should have an own configuration of these properties

I would say we shouldn't join. We have never joined anything up to now. Usually what you define in your current config overrides what is defined in the next config. For example you can override a `@Mapping` by defining it again for the same `target`. The use case here is that they want to merge unmapped source properties, but it is also completely valid not do it and to have to explicitly not ignore the inherited unmapped source properties.

>2. I am not sure if I understood the `SelectionParameters` correctly, right now I try to join the qualifiers, but maybe that is not the right way and they shouldn't be joined

 What would the backwards compatibility be here. Up to now we only took over if there was not `@BeanMapping` on the inherited method. This means that things might work differently after this PR.

> 3. Is it fine to change the `next` of the `BeanMappingOptions`? Could it be that there will then be a totally different inheritance flow? Looked for me that the delegating options works in the way that next in this case is the `@Mapper` config, then the `@MapperConfig` and so on. Now I change next to the other `@BeanMapping` and maybe this will lead to a different `@Mapper` config? Could this be?!

Yes it can be completely different mapper config. Perhaps this is even wrong now, because if there is not `@BeanMapping` and we inherit it then it would use the `@MapperConfig` from where that `@BeanMapping` is defined.

drahkrub:
Hi @chris922 @filiphr,

as I wrote in #2092 we have a fairly complex hibernate domain model containing multi-level inheritance, some beans implement several interfaces and there are quite a few `@Transient` getter methods around which I do not want to map with mapstruct.

On the other hand I don't want to miss changes in the domain model (I'm not the only one working on it), therefore I use a strict MapperConfig regarding source and target properties:

```
@MapperConfig(
        unmappedTargetPolicy = ReportingPolicy.ERROR
)
public interface CentralConfig {}
```

Because of this, I have to use `@BeanMapping(ignoreUnmappedSourceProperties = {...})` a lot.

And because of the afore mentioned complexity of the domain model I have to ignore the same source properties (which are often just `@Transient` getters) over and over and over again (I just found #1152 which looks related).

Since my DTOs are structured similar like the domain model (in terms of inheritance) this could be avoided if @BeanMapping would be inherited with unmappedSourceProperties being merged.

In my opinion this would feel natural. ;-)

Or how about an additional parameter to tackle this (`override`, `inheritanceOverride`,  `inheritUnmappedSourceProperties` or ...)?

filiphr:
@drahkrub just out of curiosity are your `@Transient` getters annotated with some annotation. If they are you can do something like proposed in #1152. You can have a custom `AccessorNamingStrategy` which will override `getMethodType` and return `MethodType.OTHER` for your transient getters. That way you won't need to ignore the properties explicitly in the mappers.

drahkrub:
@filiphr Good point. In my case the getters are indeed annotated (with [@Transient](https://docs.jboss.org/hibernate/jpa/2.2/api/javax/persistence/Transient.html)), *but* in Hibernate one can choose between annotating the getters (properties) _or_ the fields, see e.g. this epic discussion on [SO](https://stackoverflow.com/questions/594597/hibernate-annotations-which-is-better-field-or-property-access). In the latter case one does not have to use the transient annotation on "utility getters". So in my case the workaround you've proposed would probably work, on the other hand it would be anything but flexible, because maybe it could sometimes be desirable to include transient getters into some mapping...

drahkrub:
@filiphr First of all: Congratulations on the release of 1.4.0.Final! :-)

Any chance that #2092 / #2096 will be tackled in 1.4.x?

Issue #2101, opened 2020-05-20
nested constants do not work when the same type is concerned


sjaakd:
checkout [reproducer](https://github.com/mapstruct/mapstruct/pull/2103/commits/1866d90948906df7422c99cd0aac23400e41e18c).

sjaakd:
Problem is not reproducible on current version of MapStruct.

sjaakd:
Ah.. it is reproducible . Adding the `@ InheritInverseConfiguration` seems to trigger the problem

sjaakd:
Solved by PR

PR #2103, opened 2020-05-21
#2101 inherited properties need to be analysed against redefined properties when inheriting mappings

sjaakd:
Possibly need to add some unit test.. Although quite some testcases failed when I implemented it wrong 😄

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
@@ -214,11 +214,10 @@ public class MappingMethodOptions {
 
     private boolean elementsAreContainedIn( String redefinedName, String inheritedName ) {
         if ( redefinedName.startsWith( inheritedName ) ) {
-            if ( redefinedName.length() == inheritedName.length() ) {
-                // redefined == inherited
-                return true;
-            }
-            else {
+            // it is possible to redefine an exact matching source name, because the same source can be mapped to
+            // multiple targets. It is not possible for target, but caught by the Set and equals methoded in
+            // MappingOptions
+            if ( redefinedName.length() > inheritedName.length() ) {
                 // redefined.lenght() > inherited.length(), first following character should be separator
                 return '.' == redefinedName.charAt( inheritedName.length() );
             }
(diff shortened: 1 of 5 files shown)


sjaakd:
@filiphr .. shall we add the reproducer anyway..? Always good to have additional UT.

filiphr:
This is a tricky one. Basically what happens it the following mapper:

```
@Mapper
public interface Issue2101Mapper {
    @Mapping(target = "value1", source = "codeValue1")
    @Mapping(target = "value2", source = "codeValue2")
    Source map(Target target);

    //@InheritInverseConfiguration
    @Mapping(target = "codeValue1", source = "value1")
    @Mapping(target = "codeValue2", source = "value2")
    @Mapping(target = "codeValue1.code", constant = "c1")
    @Mapping(target = "codeValue1.value", source = "value1")
    @Mapping(target = "codeValue2.code", constant = "c2")
    @Mapping(target = "codeValue2.value", source = "value2")
    Target map(Source source);
}
```

Perhaps when doing the inverse inheritance if there is a target mapping which has the reversed segment we should not use it.

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
```diff
@@ -177,13 +177,54 @@ public void applyInheritedOptions(SourceMethod templateMethod, boolean isInverse
...
-            mappings.addAll( newMappings );
+            addAllNonRedefined( newMappings );
 
             // filter new mappings
             filterNestedTargetIgnores( mappings );
         }
     }
 
+    private void addAllNonRedefined(Set<MappingOptions> inheritedMappings) {
+        Set<String> redefinedSources = mappings.stream()
```
filiphr, line 188: Streams are all OK, but we are going twice through the same mappings.

Why not defined the sets, then use a for-loop to iterate the mappings and add to the sets?

sjaakd: will correct.. I'm all into this streams nowadays.. 😄

filiphr: I love streams as well. However, good old imperative style is better sometimes. Lately I am all into performance 😀

Issue #695, opened 2015-11-17
Including both shadow copy and  deep copy  may be better


wangtengchao:
Sometimes,developer need shadow copy  and somtimes deep copy may be a good choice .  If mapstruct support both of them, and developer can configure it ,i think it's better. I don't know whether this fuction has already been developed and haven't found it in document. 
Thank you!

gunnarmorling:
Hey @wangtengchao, could you describe a bit more what exactly you mean by "shadow copy" and "deep copy", just to make sure I get what you are after. Maybe you could add an example, specifically of the code you would like to see generated? Thanks!

wangtengchao:
class Product{
     private Attribute attribute;
}
class ProductMapper{
    ProductDTO productToProductDTO(Product product)
}
=>
1. Now the code is generated like this:
class ProductMapperImpl implments ProductMapper{
    ProductDTO productToProductDTO(Product product){
         ProductDTO productDTO = new ProductDTO();
         productDTO.setAttribute(product.getAttribute()); // shadow copy
   }
}
2. But sometimes, we need things below  
Attribute attribute = new Attribute(); // deep copy . generate a new object
attribute.setId(product.getAttribute().getId()); //copy data
 productDTO.setAttribute(attribute); // set the new object to productDTO

gunnarmorling:
Could you try and write a method yourself which does the cloning and add it via a mapper imported with `@Mapper#users()`:

    public Attribute clone(Attribute attribute) {
        return new Attribute(attribute.getId());
    }

The generated mapper then should pick it up when converting the `attribute` property in the `productToProductDTO()` method.

wangtengchao:
Yes, It's a good way  to realize it. But I think this is a common function for data mapping  , it may be better if i can configure it by annotation or else. 
Thanks!

ebelanger:
I would add my approval for such a change.  It would be very nice indeed if MapStruct could be used to create cloning methods since it would make sure that no properties are missed VS writing the method manually.

ebelanger:
I think I have found a way to use MapStruct to generate such cloning methods, but it will require 2 methods.

``
public Attribute clone(Attribute attribute) {
    return clone(source, new Attribute());
}

protected abstract Attribute clone(Attribute source, @MappingTarget Attribute target);
``

amidudu:
Not sure if something changed 2 years after...
We are using MapStruct for cloning and faced the same issue.
[...]

PR #2106, opened 2020-05-29
#695 Check for direct assignment for iterable, stream or map types

filiphr:
Should be done on their type parameters instead of the types themselves.

I was updating our deep clone example to use the new functionality and realized that it is not working properly for collections.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/mappingcontrol/MappingControlTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/mappingcontrol/MappingControlTest.java
@@ -5,6 +5,11 @@
  */
 package org.mapstruct.ap.test.mappingcontrol;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
 import org.junit.Test;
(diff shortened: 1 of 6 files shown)


filiphr:
I probably can take the example that we have and put it in a dedicated test as well. Might simplify some things

processor/src/test/java/org/mapstruct/ap/test/mappingcontrol/MappingControlTest.java
```diff
@@ -63,6 +72,49 @@ public void testDeepCloning() {
         assertThat( out.getShelve().getCoolBeer().getBeerCount() ).isEqualTo( "5" );
     }
 
+    /**
+     * Test the deep cloning annotation with lists
+     */
+    @Test
+    @WithClasses(CloningListMapper.class)
+    public void testDeepCloningListsAndMaps() {
```
sjaakd, line 80: should we test a stream as well?

filiphr: It wouldn't hurt, but let's leave it as is for now. I think both cases are covere

PR #2110, opened 2020-06-06
#2109 Generate correct array assignment for constructor parameter mapping

### LINKED ISSUE #2109: Assignment for array field is not generated properly
mapstruct version: 1.4.0.Beta1
java version: 14
kotlin version 1.3.72
build tool: gradle 6.5

Mapstruct processor doesn't generate mapping for not nullable array fields properly. When building project the following error occurs:
```
<path_to_project>/build/generated/source/kapt/main/com/example/ClassMapperImpl.java:28: error: variable image might not have been initialized
        ClassB classB = new ClassB( id, image );
```
Here is source code leading to the error:
com.example.ClassA.kt
```kotlin
package com.example

data class ClassA(
    val id: Long,
    val image: ByteArray
)
```
com.example.ClassB.kt
```kotlin
package com.example

data class ClassB(
    val id: Long,
    val image: ByteArray
)
```
com.example.ClassMapper.kt
```kotlin
package com.example

import org.mapstruct.Mapper

@Mapper
interface ClassMapper {
    fun classAtoClassB(classA: ClassA): ClassB
}
```
build.gradle.kts
```kotlin
val mapstruct_version = "1.4.0.Beta1"

plugins {
    val kotlinVersion = "1.3.72"
    kotlin("jvm") version kotlinVersion
    kotlin("kapt") version kotlinVersion
}

group = "org.example"
version = "1.0-SNAPSHOT"
java.sourceCompatibility = JavaVersion.VERSION_14

repositories {
    mavenCentral()
}

dependencies {
    implementation(kotlin("stdlib-jdk8"))

    implementation("org.mapstruct:mapstruct:$mapstruct_version")
    kapt("org.mapstruct:mapstruct-processor:$mapstruct_version")
}

tasks {
    compileKotlin {
        kotlinOptions.jvmTarget = "1.8"
    }
    compileTestKotlin {
        kotlinOptions.jvmTarget = "1.8"
    }
}
```
Actual generated mapping:
```kotlin
package com.example;

import java.util.Arrays;

@Generated(
    date = "2020-06-05T08:34:06+0300",
    comments = "version: 1.4.0.Beta1, compiler: IncrementalProcessingEnvironment from kotlin-annotation-processing-gradle-1.3.72.jar, environment: Java 14 (Oracle Corporation)"
)
public class ClassMapperImpl implements ClassMapper {

    @Override
    public ClassB classAtoClassB(ClassA classA) {
        if ( classA == null ) {
            return null;
        }

        long id;
        byte[] image;

        id = classA.getId();
        byte[] image1 = classA.getImage();
        if ( image1 != null ) {
            image = Arrays.copyOf( image1, image1.length );
        }

        ClassB classB = new ClassB( id, image );

        return classB;
    }
}
```
Expected generated mapping:
```kotlin
package com.example;

import java.util.Arrays;

@Generated(
    date = "2020-06-05T08:34:06+0300",
    comments = "version: 1.4.0.Beta1, compiler: IncrementalProcessingEnvironment from kotlin-annotation-processing-gradle-1.3.72.jar, environment: Java 14 (Oracle Corporation)"
)
public class ClassMapperImpl implements ClassMapper {

    @Override
    public ClassB classAtoClassB(ClassA classA) {
        if ( classA == null ) {
            return null;
        }

        long id;
        byte[] image;

        id = classA.getId();
        image = Arrays.copyOf( classA.getImage(), classA.getImage().length );

        ClassB classB = new ClassB( id, image );

        return classB;
    }
}
```
The same error with java classes:
com.example.ClazzA.java
```java
package com.example;

public class ClazzA {
    private final Long id;
    private final byte[] image;

    public ClazzA(Long id, byte[] image) {
        this.id = id;
        this.image = image;
    }
}
```
com.example.ClazzB.java
```java
package com.example;

public class ClazzB {
    private final Long id;
    private final byte[] image;

    public ClazzB(Long id, byte[] image) {
        this.id = id;
        this.image = image;
    }
}
```
com.example.ClazzMapper
```java
package com.example;

import org.mapstruct.Mapper;

@Mapper
public interface ClazzMapper {
    ClazzB classAtoClassB(final ClassA classA);
}
```
Actual genetared mapper:
```java
package com.example;

import java.util.Arrays;

@Generated(
    date = "2020-06-05T08:53:36+0300",
    comments = "version: 1.4.0.Beta1, compiler: IncrementalProcessingEnvironment from kotlin-annotation-processing-gradle-1.3.72.jar, environment: Java 14 (Oracle Corporation)"
)
public class ClazzMapperImpl implements ClazzMapper {

    @Override
    public ClazzB classAtoClassB(ClassA classA) {
        if ( classA == null ) {
            return null;
        }

        Long id;
        byte[] image;

        id = classA.getId();
        byte[] image1 = classA.getImage();
        if ( image1 != null ) {
            image = Arrays.copyOf( image1, image1.length );
        }

        ClazzB clazzB = new ClazzB( id, image );

        return clazzB;
    }
}
```
Expected generated mapper:
```java
package com.example;

import java.util.Arrays;

@Generated(
    date = "2020-06-05T08:53:36+0300",
    comments = "version: 1.4.0.Beta1, compiler: IncrementalProcessingEnvironment from kotlin-annotation-processing-gradle-1.3.72.jar, environment: Java 14 (Oracle Corporation)"
)
public class ClazzMapperImpl implements ClazzMapper {

    @Override
    public ClazzB classAtoClassB(ClassA classA) {
        if ( classA == null ) {
            return null;
        }

        Long id;
        byte[] image;

        id = classA.getId();
        byte[] image1 = classA.getImage();
        if ( image1 != null ) {
            image = Arrays.copyOf( image1, image1.length );
        } else {
            image = null;
        }

        ClazzB clazzB = new ClazzB( id, image );

        return clazzB;
    }
}
```
Probably for kotlin nullable fields assignments should be generated the same way as for fields in java classes.

filiphr:
Fixes #2109

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/ArrayCopyWrapper.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/ArrayCopyWrapper.ftl
@@ -12,8 +12,4 @@
     <@lib.handleSourceReferenceNullCheck>
         <#if ext.targetBeanName?has_content>${ext.targetBeanName}.</#if>${ext.targetWriteAccessorName}<@lib.handleWrite>Arrays.copyOf( ${sourceLocalVarName}, ${sourceLocalVarName}.length )</@lib.handleWrite>;
     </@lib.handleSourceReferenceNullCheck>
-    <#if !ext.defaultValueAssignment?? && !sourcePresenceCheckerReference?? && !ext.targetBeanName?has_content>else {<#-- the opposite (defaultValueAssignment) case is handeld inside lib.handleSourceReferenceNullCheck -->
-        ${ext.targetWriteAccessorName}<@lib.handleWrite>null</@lib.handleWrite>;
-        }
-    </#if>
 </@lib.handleExceptions>
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/ArrayCopyWrapper.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/ArrayCopyWrapper.ftl
@@ -12,4 +12,8 @@
     <@lib.handleSourceReferenceNullCheck>
         <#if ext.targetBeanName?has_content>${ext.targetBeanName}.</#if>${ext.targetWriteAccessorName}<@lib.handleWrite>Arrays.copyOf( ${sourceLocalVarName}, ${sourceLocalVarName}.length )</@lib.handleWrite>;
     </@lib.handleSourceReferenceNullCheck>
+    <#if !ext.defaultValueAssignment?? && !sourcePresenceCheckerReference?? && !ext.targetBeanName?has_content>else {<#-- the opposite (defaultValueAssignment) case is handeld inside lib.handleSourceReferenceNullCheck -->
+        ${ext.targetWriteAccessorName}<@lib.handleWrite>null</@lib.handleWrite>;
+        }
+    </#if>
 </@lib.handleExceptions>
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
@@ -30,7 +30,7 @@
                 <#list sourceParametersExcludingPrimitives as sourceParam>
                     <#if (constructorPropertyMappingsByParameter(sourceParam)?size > 0)>
                         <#list constructorPropertyMappingsByParameter(sourceParam) as propertyMapping>
-                            <@includeModel object=propertyMapping.targetType /> ${propertyMapping.targetWriteAccessorName};
+                            <@includeModel object=propertyMapping.targetType /> ${propertyMapping.targetWriteAccessorName} = ${propertyMapping.targetType.null};
                         </#list>
                         if ( ${sourceParam.name} != null ) {
                         <#list constructorPropertyMappingsByParameter(sourceParam) as propertyMapping>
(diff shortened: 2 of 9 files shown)


processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/ArrayCopyWrapper.ftl
```diff
@@ -12,4 +12,8 @@
     <@lib.handleSourceReferenceNullCheck>
         <#if ext.targetBeanName?has_content>${ext.targetBeanName}.</#if>${ext.targetWriteAccessorName}<@lib.handleWrite>Arrays.copyOf( ${sourceLocalVarName}, ${sourceLocalVarName}.length )</@lib.handleWrite>;
     </@lib.handleSourceReferenceNullCheck>
+    <#if !ext.defaultValueAssignment?? && !sourcePresenceCheckerReference?? && !ext.targetBeanName?has_content>else {<#-- the opposite (defaultValueAssignment) case is handeld inside lib.handleSourceReferenceNullCheck -->
```
sjaakd, line 15: we now add an else branch.. But I was thinking: why don't we initialise the variable with `null` (or do we have to handle primitive cases as well correctly then)? Are we going to encounter this problem more often?

filiphr: I think I tried that initially, but it was a bit complex. However, now that you mention I had a look and we can do it when we initialize the variables. I iterated so many times on the initial solution :).

filiphr:
@sjaakd I just pushed another commit with your idea. I think it makes more sense. Although the code is not like I would have written it by hand (not that it was like that before 😄).

sjaakd:
Hmm. I guess we entered the real m of fixtures now..:smile:

filiphr:
I've fixed the fixtures now as well. Let me know which one you prefer :)

sjaakd:
> I've fixed the fixtures now as well. Let me know which one you prefer :)

Personally, I often don't init variables before an (nested)  if-then-else construction, and make sure that I covered an assignment in all possible branches (let the compiler complain if this is not the case). This is more self protection.

The case in the generated code is a straightforward. I guess limiting the scope of a local variable is relevant. See: http://www.javapractices.com/topic/TopicAction.do?Id=126. Its also less verbose. Brevity over verbosity, which could come in as an advantage when you have many variables you need to initialise. Sonar will not like long methods. 

So.. are there potential cases have side-effects that we did not spot yet? Possibly. They will come to light in the beta.

I wrote down my reasoning. Not sure whether you agree, but I opt for first initialising the variable as outweighing  non-initialised branches (certainly in the generated case).

filiphr:
>Personally, I often don't init variables before an (nested) if-then-else construction, and make sure that I covered an assignment in all possible branches (let the compiler complain if this is not the case). This is more self protection.

I do the same when I write code myself as well.

In our current scenario ideally the initalization would have looked like:

```
       Long id = source.getId();

        byte[] data = null;
        byte[] data1 = source.getData();
        if ( data1 != null ) {
            data = Arrays.copyOf( data1, data1.length );
        } else {
            data = null;
        }
```

instead of:

```
        Long id = null;
        byte[] data = null;

        id = source.getId();
        byte[] data1 = source.getData();
        if ( data1 != null ) {
            data = Arrays.copyOf( data1, data1.length );
        }
```

However, due to our way of generating the code it was almost impossible to generate `Long id = source.getId();` 😄.

As far as I understood for the generated code you are more for the change I just did? Initialising the variables in advance, to avoid potential missed non initialised branches.

sjaakd:
> As far as I understood for the generated code you are more for the change I just did? Initialising the variables in advance, to avoid potential missed non initialised branches.

Yes. I think that's the better option.. Do you share that opinion ?

filiphr:
Yes that's what I had in mind as well

Issue #2122, opened 2020-06-19
Mapping a value wrapped in a list generates invalid code (stringToXmlGregorianCalender will be added)


chris922:
I really don't know how to title this issue.. but I found a strange error similar to a really old issue that we had in the past (#328).

Was testing with 1.4.0.Beta1, had a quick test and it also exists in 1.3.1.Final

Code:

```java
@Mapper
public interface TestMapper {
    @Mapping(target = "embeddedTarget", source = "value")
    Target toTarget(Source source);

    EmbeddedTarget toEmbeddedTarget(String value);
    
    default <T> List<T> singleEntry(T entry) {
        return List.of(entry);
    }

    @Data
    class Source {
        String value;
    }

    @Data
    class Target {
        List<EmbeddedTarget> embeddedTarget;
    }

    @Data
    class EmbeddedTarget {
        String value;
    }
}
```

The generated method looks like this:

```java
    @Override
    public Target toTarget(Source source) {
        if ( source == null ) {
            return null;
        }

        Target target = new Target();

        target.setEmbeddedTarget( singleEntry( stringToXmlGregorianCalendar( source.getValue(), null ) ) );

        return target;
    }
```

_(+ the usual `DatatypeFactory` stuff and `stringToXmlGregorianCalender` built-in method)_

In case such a (crazy) mapping is possible I would expect the `toEmbeddedTarget(String)` method will be used. At least this method was generated correctly... and if the mapping is not possible.. as always an appropriate error message and no invalid code :)

sjaakd:
This definitely does not seem right? @chris922, did you investigate the 2 step mapping mechanism?

sjaakd:
Perhaps you can make a reproducer en put that in a PR so that we can look together if you want..

chris922:
Good idea! PR for a reproducer is there.

 I haven't yet investigated where the issue starts... but interesting to see that the stringToXmlGregorianCalender was also used for #328, maybe somewhere is a list from which the first item will be used?!

sjaakd:
I'll try to have a good look at it tomorrow. I have included code to avoid A -> Object, Object -> B the last time (if I remember correctly) in 2 step mappings. Today I've got other appointments 😄

sjaakd:
@chris922 .. @filiphr did a review.. So I merged in our joint code (want to make a new release later and need to fix some other stuff)

sjaakd:
find by PR #2122 @filiphr .. thanks for reviewing

PR #2129, opened 2020-06-28
#2122 fix for 2 step mapping and generics

sjaakd:


sjaakd:
It would be nice to have some more tricky tests.. E.g nested stuff:

`List<SimpleElement<T>>` ... or something with a `Map<String, T>` possibly with an implementation type delivered by the used method as result to test the new method in the factory full (`isAssignable`)...

@chris922 perhaps you see even more options.. I'm not to sure about the visitor.. I always find those tricky things..

sjaakd:
and.. perhaps the method should not be part of type factory. but of type..

sjaakd:
inheritance does not work yet..Just replace `Map` with `HashMap` in the target..  and also need to cleanup..

sjaakd:
The test fails.. 

Somehow the wrong ref gets selected / returned on line 436 of the`MethodResolverImpl`. That is wrong and needs to be fixed. Probably deep in `MethodMatcher` (line 145)

However it also makes you think why the `resolveViaMethod` is called in the first place? 

We have the proper candidate in our hands, we just need to make a reference for it (methodY candidate) and even that we possibly should delay until we found a proper methodX.. (similar case for method-typeconversion). This is really a performance thing and needs to be addressed separately. 

I guess I'll make a separate issue out of that a s soon as I've got a solution.

sjaakd:
probably the same problem occurs for other 2 steps:
* method, type conversion
* type conversion, method

we need to add test cases for this and add the solution also there. @chris922 : do you have some time to do this? I can help with the implementation (that's rather straightforward). Making test case will help.

sjaakd:
Not sure why type conversion to method and vice versa was already working (in the master). Anyway, I guess its better to apply the same strategy everywhere. Added additional test as well.

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
```diff
@@ -426,29 +426,33 @@ private Assignment resolveViaMethodAndMethod(Type sourceType, Type targetType) {
...
+                if ( Object.class.getName().equals( ySourceType.getName() ) ) {
                     //  java.lang.Object as intermediate result
                     continue;
                 }
 
-                methodRefY =
-                    resolveViaMethod( methodYCandidate.getSourceParameters().get( 0 ).getType(), targetType, true );
+                if ( ySourceType.isTypeVar() ) {
+                    ySourceType = ySourceType.resolveToType( targetType, methodYCandidate.getResultType() );
+                }
```
filiphr, line 437: `resolveToType` already checks for `isTypeVar()`. You can directly do

```
ySourceType = ySourceType.resolveToType( targetType, methodYCandidate.getResultType() );
```

sjaakd: that's true..  but there's actually a reason for this:
1. I did not want to do the resolveToType unchecked in the `resolveToType` method itself.. Was a bit struggling with the method name.
2. I wanted to make the flow clear at the call site..

 I could have made it `resolveTypeVarToType` at hind site.  Will do this.

Issue #2133, opened 2020-07-03
Regression: @BeanMapping#resultType does not work icw nested mappings


sjaakd:
I could possibly make the reproducer more minimal. But this reflects the situation in my code.

Works in 1.3.1 but not in 1.4.0.Beta1

https://github.com/mapstruct/mapstruct/pull/2134

sjaakd:
Solved by PR #2134.. @filiphr thanks for reviewing..

PR #2134, opened 2020-07-03
#2133 @BeanMapping#resultType should not be applied to forged methods

sjaakd:


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -472,7 +472,10 @@ private void sortPropertyMappingsByDependencies() {
         }
 
         private Type getReturnTypeToConstructFromSelectionParameters(SelectionParameters selectionParams) {
-            if ( selectionParams != null && selectionParams.getResultType() != null ) {
+            // resultType only applies to method that actually has @BeanMapping annotation, never to forged methods
+            if ( !( method instanceof ForgedMethod )
```
filiphr, line 476: This makes sense. However, doesn't it make more sense to construct the appropriate `MappingMethodOptions` when we create a new `ForgedMethod` instead?

sjaakd: I (of course :smile: looked on that option). But
1. Its too much hassle
2. There might be actually a reason to carry down all options through the underlying forged methods.. You just have to be careful to use them and check whether they are really applicable

sjaakd: We had it like that some refactorings ago..

PR #2137, opened 2020-07-04
#2101 follow up - wip

sjaakd:


processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
```diff
@@ -369,4 +377,50 @@ public int hashCode() {
...
+
+    public static class WrappedBeanMappingOptions extends BeanMappingOptions {
+
+        protected WrappedBeanMappingOptions(BeanMappingOptions options) {
+            super( options );
+        }
+
+        @Override
+        public BuilderGem getBuilder() {
+            return null;
```
filiphr, line 407: We should use the delegating here. Since builder can be defined on `@Mapper`. We only need to skip the `beanMapping`.

sjaakd: done.. but I really need to debug this one..

sjaakd: let's park this one for the time being..

Issue #2136, opened 2020-07-04
Improve performance for 2 step mappings


sjaakd:
In a mapping `source ->A -> B -> target` I've noticed that once we determined that we are investigating method `A` we still go through all methods to resolve `A` while we are already have the method in our hands at that point in time.

This means if we have 
* `An` methods A
* `Bn` methods B
* Since it's the same set, `An` == `Bn` == `N` (number of available methods for using)
* Assuming we have a hit (on average) investigating half of the method. 

We have to investigate `(0.5 * N)^2`. But now we investigate `(0.5 * N)^3`

A practical example: 
An = 10,
* current solution 125
* optimised solution 25

sjaakd:
solved by #2138 . Thanks @filiphr for reviewing

PR #2138, opened 2020-07-04
#2136 performance improvement 2 step mappings

sjaakd:


processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
```diff
@@ -376,32 +377,56 @@ private Assignment resolveViaMethod(Type sourceType, Type targetType, boolean co
...
+         *
+         * @param sourceType the source type
+         * @param targetType the target type
+         * @param method     the method to match
+         * @return an assignment if a match, given the criteria could be found. When the method is a
+         * buildIn method, all the bookkeeping is applied.
+         */
+        private Assignment applyMatching(Type sourceType, Type targetType, Method method) {
+
+            if ( method instanceof SourceMethod ) {
```
filiphr, line 391: What if the method is not one of these types? It can be an intermediate `ForgedMethod` or an existing `HelperMethod` that we have created. Why shouldn't we reuse that one?

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
```diff
@@ -486,10 +511,9 @@ private Assignment resolveViaConversionAndMethod(Type sourceType, Type targetTyp
                 ySourceType = ySourceType.resolveTypeVarToType( targetType, methodYCandidate.getResultType() );
 
                 if ( ySourceType != null ) {
-                    methodRefY = resolveViaMethod( ySourceType, targetType, true );
+                    methodRefY = applyMatching( ySourceType, targetType, methodYCandidate );
                     if ( methodRefY != null ) {
-                        Type targetTypeX = methodYCandidate.getSourceParameters().get( 0 ).getType();
-                        ConversionAssignment conversionXRef = resolveViaConversion( sourceType, targetTypeX );
+                        ConversionAssignment conversionXRef = resolveViaConversion( sourceType, ySourceType );
```
filiphr, line 516: Why is this change needed?

`ySourceType` and `targetTypeX` don't have to be the same types I think. One can be more specific then the other.

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
```diff
@@ -376,32 +377,56 @@ private Assignment resolveViaMethod(Type sourceType, Type targetType, boolean co
...
+        /**
+         * Applies matching to given method only
+         *
+         * @param sourceType the source type
+         * @param targetType the target type
+         * @param method     the method to match
+         * @return an assignment if a match, given the criteria could be found. When the method is a
+         * buildIn method, all the bookkeeping is applied.
+         */
+        private Assignment applyMatching(Type sourceType, Type targetType, Method method) {
```
filiphr, line 389: Perhaps this should be named `resolveViaGivenMethod`? Or something in that direction. `applyMatching` differs from the other methods in my opinion

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
sjaakd: is not possible here.. We don't resolve forged methods.. If so, we should throw an unsupported.

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
sjaakd: the change was forgotten last time.. it's actually a bug.

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
sjaakd: not entirely.. We still have to apply all criteria to really figure out if it matches (qualifiers, xml-stuff, resultTypes, etc)

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
filiphr: You are right

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
filiphr: 👍

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
filiphr: OK

processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
sjaakd: resolved, I guess?

PR #2158, opened 2020-07-17
Strip leading underscores when sanitizing identifiers

### LINKED ISSUE #2142: Question: is it possible to change the naming of variables in the generated code?
Hi.

We currently try to map our data classes to some external generated classes.
The target classes look like this:
```java
@Builder //lombok builder
@Getter
class Contact {
   _PhoneNumber phone;

  @Builder
  @Getter
  class _PhoneNumber {
    // ....
  }
}
```

in a mapper like this
```java
abstract Contact map(InternalContact contact)
```
mapstruct will generate code like this
```java
_PhoneNumberBuilder _PhoneNumber = _PhoneNumber.builder();
```

We assume that the variable name "_PhoneNumber" is generated by lowercasing the first letter of the class name. Is there a way to configure this behavior somehow? We know `_` is not a good convention but thats what we have right now 🙃

Thanks!

We tried the following versions 1.3.1.Final and 1.4.0.Beta1

filiphr:
Fixes #2142

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
@@ -125,8 +125,7 @@ public class ForgedMethod implements Method {
                          boolean forgedNameBased) {
 
         // establish name
-        String sourceParamName = Strings.decapitalize( sourceType.getName() );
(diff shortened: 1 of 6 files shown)


sjaakd:
LGTM

PR #2160, opened 2020-07-18
#2150 Change the rules for how a constructor for mapping is picked

### LINKED ISSUE #2150: Allow @Default to overrule potentially existing parameterless default constructor.
Since 1.4.0.Beta2, MapStruct allows direct mapping to parameterized constructors as described [here](https://mapstruct.org/documentation/dev/reference/html/#mapping-with-constructors). The reference guide states:

> * If a parameterless constructor exists then it would be used to construct the object, and the other constructors will be ignored
> * If there are multiple constructors then the one annotated with annotation named @Default (from any package) will be used

I understand this as follows: If there is a parameterless default constructor, this one wins and @Default is only interpreted if there is no paramterless default constructor. However, this is problematic in a setup, where a default constructor is only made visible to the objects children, so being protected. Lets take the following example:

``` java
public class Vehicle {
    
    private String color;
    private int amountOfWheels;
    
    protected Vehicle() {
        // e.g. needed for serialization.
    }

    @Default
    public Vehicle(String color, int amountOfWheels) {
        this.color = color;
        this.amountOfWheels = amountOfWheels;
    }
...
}
```
and
``` java
public class Car extends Vehicle {

    private String brand;

    private Car() {
        // e.g. needed for serialization.
    }

    public Car(String brand, Vehicle vehicle) {
        super(vehicle.getColor(), vehicle.getAmountOfWheels());
        this.brand = brand;
    }
...
}
```
with a DTO as follows
``` java
public class VehicleDTO {

    private String color;
    private int amountOfWheels;

...
}
```
and the according mapper
``` java
@Mapper
public interface VehicleMapper {

    VehicleMapper INSTANCE = Mappers.getMapper(VehicleMapper.class);

    Vehicle map(VehicleDTO dto);
}
```

In this case I want the default constructor to be only visible to the child as it is needed e.g. for serialization and want MapStruct to ignore it and use the declared @Default constructor. However, MapStruct is picking up the protected default constructor and throws the following error:

> Compilation failure
> [ERROR] MapStructsExample/target/generated-sources/annotations/de/ralfhecktor/mapper/VehicleMapperImpl.java:[20,27] constructor Vehicle in class de.ralfhecktor.Vehicle cannot be applied to given types;
> [ERROR]   required: java.lang.String,int
> [ERROR]   found: no arguments
> [ERROR]   reason: actual and formal argument lists differ in length

Having a look at the generated implementation shows:

``` java
public class VehicleMapperImpl implements VehicleMapper {

    @Override
    public Vehicle map(VehicleDTO dto) {
        if ( dto == null ) {
            return null;
        }

        Vehicle vehicle = new Vehicle();

        return vehicle;
    }
}
```
instead of 

``` java
public class VehicleMapperImpl implements VehicleMapper {

    @Override
    public Vehicle map(VehicleDTO dto) {
        if ( dto == null ) {
            return null;
        }

        String color = null;
        int amountOfWheels = 0;

        color = dto.getColor();
        amountOfWheels = dto.getAmountOfWheels();

        Vehicle vehicle = new Vehicle( color, amountOfWheels );

        return vehicle;
    }
}
```

So what would be suggestion on solving this? 

For MapStruct I would see three possibilities:

1. Only consider `public` constructor to be preferred over the @Default.
1. Make @Default a definite override, so if somebody declares a constructor as default, it should also be picked up.
1. Give @Default a property e.g. `force` which can be set to `true` to force the usage.

Personally, I would find option 2 the most intuitive. 

**You can find a full working project [here](https://github.com/ralfhecktor/MapStructExample)**

I am looking forward to your thoughts and suggestions on this topic. Thanks.

filiphr:
New rules:

1. Constructor annotated with @Default (from any package) has highest precedence
2. If there is a single public constructor then it would be used to construct the object
3. If a parameterless constructor exists then it would be used to construct the object, and the other constructors will be ignored

Fixes #2150

### DIFF
--- a/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
@@ -537,8 +537,10 @@ When doing a mapping MapStruct checks if there is a builder for the type being m
 If there is no builder, then MapStruct looks for a single accessible constructor.
 When there are multiple constructors then the following is done to pick the one which should be used:
 
-* If a parameterless constructor exists then it would be used to construct the object, and the other constructors will be ignored
-* If there are multiple constructors then the one annotated with annotation named `@Default` (from any package) will be used
+* If a constructor is annotated with an annotation named `@Default` (from any package) it will be used.
+* If a single public constructor exists then it will be used to construct the object, and the other non public constructors will be ignored.
+* If a parameterless constructor exists then it will be used to construct the object, and the other constructors will be ignored.
+* If there are multiple eligible constructors then there will be a compilation error due to ambigious constructors. In order to break the ambiquity an annotation named `@Default` (from any package) can used.
 
 When using a constructor then the names of the parameters of the constructor will be used and matched to the target properties.
 When the constructor has an annotation named `@ConstructorProperties` (from any package) then this annotation will be used to get the names of the parameters.
--- a/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
@@ -542,6 +542,48 @@ When there are multiple constructors then the following is done to pick the one 
 * If a parameterless constructor exists then it will be used to construct the object, and the other constructors will be ignored.
 * If there are multiple eligible constructors then there will be a compilation error due to ambigious constructors. In order to break the ambiquity an annotation named `@Default` (from any package) can used.
 
+.Deciding which constructor to use
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+public class Vehicle {
+
+    protected Vehicle() { }
+
+    // MapStruct will use this constructor, because it is a single public constructor
+    public Vehicle(String color) { }
+}
+
+public class Car {
+
+    // MapStruct will use this constructor, because it is a parameterless empty constructor
+    public Car() { }
+
+    public Car(String make, String color) { }
+}
+
+public class Truck {
+
+    public Truck() { }
+
+    // MapStruct will use this constructor, because it is annotated with @Default
+    @Default
+    public Truck(String make, String color) { }
+}
+
+public class Van {
+
+    // There will be a compilation error when using this class because MapStruct cannot pick a constructor
+
+    public Van(String make) { }
+
+    public Van(String make, String color) { }
+
+}
+----
+====
+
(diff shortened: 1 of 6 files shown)


documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
```diff
@@ -537,8 +537,10 @@ When doing a mapping MapStruct checks if there is a builder for the type being m
 If there is no builder, then MapStruct looks for a single accessible constructor.
 When there are multiple constructors then the following is done to pick the one which should be used:
 
-* If a parameterless constructor exists then it would be used to construct the object, and the other constructors will be ignored
-* If there are multiple constructors then the one annotated with annotation named `@Default` (from any package) will be used
+* If a constructor is annotated with an annotation named `@Default` (from any package) it will be used.
+* If a single public constructor exists then it will be used to construct the object, and the other non public constructors will be ignored.
+* If a parameterless constructor exists then it will be used to construct the object, and the other constructors will be ignored.
+* If there are multiple eligible constructors then there will be a compilation error due to ambigious constructors. In order to break the ambiquity an annotation named `@Default` (from any package) can used.
```
sjaakd, line 543: Perhaps add an example?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -603,63 +603,84 @@ private ConstructorAccessor getConstructorAccessor(Type type) {
...
+
                 if ( constructor.getParameters().isEmpty() ) {
-                    defaultConstructor = constructor;
+                    parameterLessConstructor = constructor;
                 }
                 else {
                     accessibleConstructors.add( constructor );
                 }
+
+                if ( constructor.getModifiers().contains( Modifier.PUBLIC ) ) {
```
sjaakd, line 634: Should we also allow friendly and protected?  I guess we have some logic somewhere to detect these scenarios...

filiphr: We do allow protected ones. Those are part of the `accessibleConstructors`. This is to have public constructors with e precedence, so we can check them first.

documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
filiphr: That's a good point. I'll add one tonight

Issue #2161, opened 2020-07-19
Refactor messaging


sjaakd:
Messaging seems to be all over the place in different ways.

I've started working on this.. But its quite a lot of work. What needs to be done:

* When you come to look in detail at `ForgedMethodHistory` it is not as much the forged history, but a description of the mapping.  So rename `ForgedMethodHistory` to `MappingDescription`
* `MappingDescripition` should know how to format 1) the mapping that is failing, 2) the proposal what needs to be done.
* `MappingDescription` can now be applied to forged methods and regular methods. 
* `SourceRHS` contains a description as well. It's called `sourceErrorMessagePart`. Remove that.
* Printing types should be on simple name, only use FQN when verbose
* When ambigous methods are listed, keep the list to 5 entries. Only vebose mode should print the full list.
* 'isErroneous' on the messager feels like a bit of a hack. I could be mistaken, but it looks like its not always working. I needed to add 2 errors to my testcases (unmapped and ambigous) while it concerns the same property.

All in all this is a considerable redesign. I propose to break this up into pieces. Its to big to do at once. Perhaps we can finish it for the1.4 release otherwise 1.5

PR #2166, opened 2020-07-23
#2161 use short names instead of FQN's in error messages

sjaakd:
Possibly I need to add a verbose testcase as well. But I'm quite happy with the improved messages..


sjaakd:
Thanks. @filiphr Thanks for reviewing!

PR #2168, opened 2020-07-26
#2161 rename ForgedMethodHistory to MethodDescription

sjaakd:


processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
```diff
@@ -249,7 +249,7 @@ public Type getReturnType() {
         return thrownTypes;
     }
 
-    public ForgedMethodHistory getHistory() {
+    public MethodDescription getHistory() {
```
filiphr, line 252: maybe we should rename the method to `getDescription` as well?

sjaakd: I'll do so... This is all very much WIP.. The only thing I did was press refactor. I sometimes use CI to quickly check. 

I'll assign you as reviewer as soon as it becomes substantial 😄

PR #2173, opened 2020-08-02
#2169 Add support for defining a custom default exception for enum mapping

### LINKED ISSUE #2169: customize thrown Exception when Enum Mapping fails
It would be nice to be able to configure the Mapper to throw a custom Exception for an unknown enum value, instead of an `IllegalArgumentException`.

Possible solution:
@filiphr:
> We've added EnumNamingStrategy in 1.4. Perhaps we can rename that to EnumMapingStrategy and add a method that would return the TypeElement of the exception that it should be thrown.

For now we manipulated `ValueMappingMethod.ftl` in the classpath, but this is an internal file, hence not the way to go: mapstruct might switch away from using freemarker templates in the near future.

filiphr:
Fixes #2169

Perhaps we should rename `EnumNamingStrategy` into `EnumMappingStrategy`. Not sure if naming is still correct if we have the `getDefaultExceptionType()` in there.

What do you think @sjaakd

### DIFF
--- a/core/src/main/java/org/mapstruct/EnumMapping.java
+++ b/core/src/main/java/org/mapstruct/EnumMapping.java
@@ -123,8 +123,8 @@ public @interface EnumMapping {
 
     /**
      * Exception that should be thrown by the generated code if no mapping matches.
-     * If no exception is configured, the exception given via {@link MapperConfig#defaultEnumException()} or
-     * {@link Mapper#defaultEnumException()} will be used, using {@link IllegalArgumentException} by default.
+     * If no exception is configured, the exception given via {@link MapperConfig#unexpectedValueMappingException()} or
+     * {@link Mapper#unexpectedValueMappingException()} will be used, using {@link IllegalArgumentException} by default.
      *
      * <p>
      * Note:
@@ -140,5 +140,5 @@ public @interface EnumMapping {
      *
      * @return the exception that should be used in the generated code
      */
-    Class<? extends Exception> defaultException() default IllegalArgumentException.class;
+    Class<? extends Exception> unexpectedValueMappingException() default IllegalArgumentException.class;
 }
--- a/core/src/main/java/org/mapstruct/EnumMapping.java
+++ b/core/src/main/java/org/mapstruct/EnumMapping.java
@@ -123,6 +123,8 @@ public @interface EnumMapping {
 
     /**
      * Exception that should be thrown by the generated code if no mapping matches.
+     * If no exception is configured, the exception given via {@link MapperConfig#defaultEnumException()} or
+     * {@link Mapper#defaultEnumException()} will be used, using {@link IllegalArgumentException} by default.
      *
      * @return the exception that should be used in the generated code
      */
--- a/core/src/main/java/org/mapstruct/EnumMapping.java
+++ b/core/src/main/java/org/mapstruct/EnumMapping.java
@@ -111,7 +111,7 @@ public @interface EnumMapping {
      *
      * @return the name transformation strategy
      */
-    String nameTransformationStrategy();
+    String nameTransformationStrategy() default "";
 
     /**
      * The configuration that should be passed on the appropriate name transformation strategy.
@@ -119,5 +119,12 @@ public @interface EnumMapping {
      *
      * @return the configuration to use
      */
-    String configuration();
+    String configuration() default "";
+
+    /**
+     * Exception that should be thrown by the generated code if no mapping matches.
+     *
+     * @return the exception that should be used in the generated code
+     */
+    Class<? extends Exception> defaultException() default IllegalArgumentException.class;
 }
--- a/core/src/main/java/org/mapstruct/EnumMapping.java
+++ b/core/src/main/java/org/mapstruct/EnumMapping.java
@@ -126,6 +126,18 @@ public @interface EnumMapping {
      * If no exception is configured, the exception given via {@link MapperConfig#defaultEnumException()} or
      * {@link Mapper#defaultEnumException()} will be used, using {@link IllegalArgumentException} by default.
      *
+     * <p>
+     * Note:
+     * <ul>
+     *     <li>
+     *      The defined exception should at least have a constructor with a {@link String} parameter.
+     *     </li>
+     *     <li>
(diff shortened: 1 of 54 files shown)


sjaakd:
I guess the latter feels better..

filiphr:
>Does enummapping inherit from mapper? Iiuc you'll have to apply this for each method right? (Cannot see the full context from a mobile)

If you mean whether you can define the default exception on a `@Mapper` level or `@MapperConfig`. Then the answer is no. Should we expose it? It is exposed as an SPI globally

sjaakd:
I think we should. I thought we used the SPI for standard post and prefix stuff and the SPI for more tricky things.. that makes sense. But an exception and the standard stuff should follow the same rules as other settings. I'd prefer the SPI not to see as a 'straightforward' interface but to reserve it's use only for 'team' or 'organisation' policy stuff or for stuff that's too complex in doing otherwise like programmatic manipulation

sjaakd:
Does the new interface add checked exceptions to the signature of the value mapping method?

sjaakd:
Are exception  imports handled correctly?

sjaakd:
What about exceptions with a non empty constructor?

filiphr:
I haven't answered your questions it seems @sjaakd.

>I think we should. I thought we used the SPI for standard post and prefix stuff and the SPI for more tricky things.. that makes sense. But an exception and the standard stuff should follow the same rules as other settings. I'd prefer the SPI not to see as a 'straightforward' interface but to reserve it's use only for 'team' or 'organisation' policy stuff or for stuff that's too complex in doing otherwise like programmatic manipulation

I agree, we should expose this setting on `@Mapper` and `@MapeprConfig` as well.

>Does the new interface add checked exceptions to the signature of the value mapping method?

Not entirely following. We can't add a new checked exception to an overridden method.

>Are exception imports handled correctly?
Yes they are, have a look at the test fixture `CustomDefaultExceptionMapperImpl.java`.

>What about exceptions with a non empty constructor?

I was not sure about this one. We can do that check, and tell the user that we require the exception to have a constructor that takes a `String` for the message or we can leave it to the compilation. The generated code will not compile anyways.

sjaakd:
> I was not sure about this one. We can do that check, and tell the user that we require the exception to have a constructor that takes a `String` for the message or we can leave it to the compilation. The generated code will not compile anyways.

That's true.. And, perhaps.. using an `ObjectFactory` in these cases makes more sense...

filiphr:
>And, perhaps.. using an `ObjectFactory` in these cases makes more sense...

An `ObjectFactory` for creating an instance of the Exception? I would prefer to avoid that complexity.

processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
```diff
@@ -485,16 +500,28 @@ String getValue(ValueMappingOptions valueMapping) {
     }
 
     private ValueMappingMethod(Method method, List<MappingEntry> enumMappings, String nullTarget, String defaultTarget,
-        boolean throwIllegalArgumentException, List<LifecycleCallbackMethodReference> beforeMappingMethods,
+        Type defaultException,
+        List<LifecycleCallbackMethodReference> beforeMappingMethods,
```
sjaakd, line 504: I know it is not part of this solution.. but I'm not happy with those methods coming from `@BeanMapping`.. I think they should be provided by `@EnumMapping`

filiphr: the selectors are coming from `BeanMapping` indeed. I think that this is a good think for the 2.0 API overhauling. perhaps the qualifiers should go in their own annotation for this.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/value/exception/CustomDefaultExceptionDefinedInMapperConfigImpl.java
```diff
@@ -0,0 +1,101 @@
...
+            break;
+            case STANDARD: externalOrderType = ExternalOrderType.DEFAULT;
+            break;
+            case NORMAL: externalOrderType = ExternalOrderType.DEFAULT;
+            break;
+            case RETAIL: externalOrderType = ExternalOrderType.RETAIL;
+            break;
+            case B2B: externalOrderType = ExternalOrderType.B2B;
+            break;
+            default: throw new CustomIllegalArgumentException( "Unexpected enum constant: " + orderType );
```
sjaakd, line 73: when you throw a checked exception.. shouldn't we check whether the method signature has defined such exception in the throws clause or do we leave this up to the compiler?

core/src/main/java/org/mapstruct/EnumMapping.java
```diff
@@ -111,13 +111,22 @@
...
      * The configuration that should be passed on the appropriate name transformation strategy.
      * e.g. a suffix that should be applied to the source enum when doing name based mapping.
      *
      * @return the configuration to use
      */
-    String configuration();
+    String configuration() default "";
+
+    /**
+     * Exception that should be thrown by the generated code if no mapping matches.
```
sjaakd, line 125: we probably should mention that:
1. The exception should have an empty constructor
2. The exception (when checked) should be added to the method signature.

sjaakd (approved):
Still some minor comments.. I leave it up to you if it makes sense an how to handle those.. Otherwise LGTM

processor/src/test/resources/fixtures/org/mapstruct/ap/test/value/exception/CustomDefaultExceptionDefinedInMapperConfigImpl.java
filiphr: I say we leave it to the compiler. In the end you as a user defines the exception that you want to throw, so it is entirely in your hands.

core/src/main/java/org/mapstruct/EnumMapping.java
filiphr: I'll add that.

sjaakd: great.. After that, you could merge the PR..

PR #2190, opened 2020-08-28
#2177 Mapping into a generic class with a typed constructor argument should work

### LINKED ISSUE #2177: Mapping to generic class via constructor does not work (1.4.0.Beta3)
When using the current MapStruct 1.4.0.Beta3 and trying to map to a generic class via its constructor, I get an error message. See here for a code example that reproduces the error:

```
import org.mapstruct.Mapper;

@Mapper
interface MapstructGenericsTest {

  Generic<String> doesNotWork(NonGeneric nonGeneric);

  class NonGeneric {
    public String value;
    public String getValue() {
      return value;
    }
  }

  class Generic<T> {
    private final T value;
    public Generic(T value) {
      this.value = value;
    }
    public T getValue() {
      return value;
    }
  }
}
```

MapStruct raises the following error: 

> Error: Can't map property "java.lang.String value" to "T value". Consider to declare/implement a mapping method: "T map(java.lang.String value)".

Mapping to an equivalent class that has a setter method, works. So only the new "constructor mapping" feature seems to be affected, example:

```
@Mapper
interface MapstructGenericsTest {

  GenericBySetter<String> doesWork(NonGeneric nonGeneric);

  class NonGeneric {
    public String value;
    public String getValue() {
      return value;
    }
  }

  class GenericBySetter<T> {
    private T value;
    public void setValue(T value) {
      this.value = value;
    }
    public T getValue() {
      return value;
    }
  }
}
```

filiphr:
Fixes #2177

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2177/Issue2177Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2177/Issue2177Mapper.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2177;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper
+public interface Issue2177Mapper {
+
+    Issue2177Mapper INSTANCE = Mappers.getMapper( Issue2177Mapper.class );
+
+    Target<String> map(Source source);
+
+    class Source {
+        private final String value;
+
+        public Source(String value) {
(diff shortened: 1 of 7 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_2177/Issue2177Mapper.java
```diff
@@ -0,0 +1,44 @@
...
+        public Source(String value) {
+            this.value = value;
+        }
+
+        public String getValue() {
+            return value;
+        }
+    }
+
+    class Target<T> {
```
sjaakd, line 33: I saw that you also added code for java14 records.. should we have a testcase for that ?

filiphr: You are right it doesn't work for generic records. I'll add a test case and fix for that as well

sjaakd (approved):

PR #2191, opened 2020-08-28
#2174 Forged methods should inherit all thrown exceptions from their property mappings

### LINKED ISSUE #2174: Redundant exception wrapping in nested generated method
- [V] Is this an issue (and hence not a question)? 

Hi All,
I am using MapStruct v1.3.1.Final

The guide [says](https://mapstruct.org/documentation/stable/reference/html/#exceptions) that you can make the generated method to throw a checked exception by simply adding a `throws` declaration.

But in a nested generated method it doesn't seem to work.

For example, let say I have the following (All persistence and Spring annotations are just for the sake of the example):

```
public class User
{
    @Id
    UUID id;
    @OneToOne
    private Address address;
    
    // getters & setters
}

public class Address
{
    @Id
    UUID id;
    @OneToMany
    private City city;
    private String street;

   // getters & setters

}

public class City
{
    @Id
    UUID id;   
    String name;

   // getters & setters
}
```
And a DTO:
```
public class UserDto
{
    private String name;

    private String city;

    private String street;

    // getters & setters
}
```

The Mapper:
```
@Mapper
public abstract class UserMapper
{

    @Autowired
    private CityRepository cityRepository;

    @Mappings(
        @Mapping(source = "city", target = "address.city")
    )
    protected abstract User map(UserDto userDto) throws CityNotFoundException;

    protected City map(String city) throws CityNotFoundException {
        return cityRepository.findByName(city); // throws CityNotFoundException
    }
}
```

Now, the generated mapper implementation is:
```
public class UserMapperImpl extends UserMapper {

    @Override
    protected User map(UserDto userDto) throws CityNotFoundException {
        if ( userDto == null ) {
            return null;
        }

        User user = new User();

        user.setAddress( userDtoToAddress( userDto ) );

        return user;
    }

    protected Address userDtoToAddress(UserDto userDto) {
        if ( userDto == null ) {
            return null;
        }

        try {
            address.setCity( map( userDto.getCity() ) );
        }
        catch ( CityNotFoundException e ) {
            throw new RuntimeException( e );  // This warpping is redundant
        }

        return address;
    }
}
```

As you can see it wraps the `CityNotFoundException` with a `RuntimeException` even though I explicitly declared both of the mappers with `throws CityNotFoundException`

Expected behavior:
Add a `throws CityNotFoundException` declaration to the `userDtoToAddress` method, and don't wrap the exception.

filiphr:
Fixes #2174

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -296,8 +296,18 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                             existingVariableNames
             );
 
-            if (factoryMethod != null && method instanceof ForgedMethod ) {
-                ( (ForgedMethod) method ).addThrownTypes( factoryMethod.getThrownTypes() );
+            if ( method instanceof ForgedMethod ) {
+                ForgedMethod forgedMethod = (ForgedMethod) method;
+                if ( factoryMethod != null ) {
+                    forgedMethod.addThrownTypes( factoryMethod.getThrownTypes() );
+                }
+
+                for ( PropertyMapping propertyMapping : propertyMappings ) {
+                    if ( propertyMapping.getAssignment() != null ) {
+                        forgedMethod.addThrownTypes( propertyMapping.getAssignment().getThrownTypes() );
+                    }
+                }
+
             }
 
             MethodReference finalizeMethod = null;
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2174/UserMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2174/UserMapper.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2174;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper
+public interface UserMapper {
+
+    UserMapper INSTANCE = Mappers.getMapper( UserMapper.class );
+
+    @Mapping(target = "address.city", source = "city")
+    User map(UserDto dto) throws CityNotFoundException;
+
+    default City mapCity(String city) throws CityNotFoundException {
+        if ( "Test City".equals( city ) ) {
+            return new City( city );
+        }
+
+        throw new CityNotFoundException( "City with name '" + city + "' does not exist" );
+    }
+
+    class UserDto {
+        private final String city;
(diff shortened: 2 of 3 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_2174/UserMapper.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+@Mapper
+public interface UserMapper {
+
+    UserMapper INSTANCE = Mappers.getMapper( UserMapper.class );
+
+    @Mapping(target = "address.city", source = "city")
+    User map(UserDto dto) throws CityNotFoundException;
+
+    default City mapCity(String city) throws CityNotFoundException {
```
sjaakd, line 23: would be interesting to see if 2 forged methods are called from the same method. E.g. a postal code and a `PostalCodeNotFounException`. Both should be added to the thrown clause

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -296,8 +296,18 @@ else if ( !method.isUpdateMethod() ) {
                             existingVariableNames
             );
 
-            if (factoryMethod != null && method instanceof ForgedMethod ) {
-                ( (ForgedMethod) method ).addThrownTypes( factoryMethod.getThrownTypes() );
+            if ( method instanceof ForgedMethod ) {
+                ForgedMethod forgedMethod = (ForgedMethod) method;
+                if ( factoryMethod != null ) {
```
sjaakd, line 301: and what about factory methods?

sjaakd (changes requested):
Looks good in general.. 2 questions though..

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
filiphr: What about them? Before the if check was `if ( factoryMethod != null && method instanceOf ForgedMethod )`. I moved the `if ( factoryMethod != null )` inside.

processor/src/test/java/org/mapstruct/ap/test/bugs/_2174/UserMapper.java
filiphr: I'll add that as well

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
sjaakd: I meant: what if a factory method throws an exception while creating an object.. but I guess that's covered elsewhere ...

sjaakd (approved):
LGTM

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
filiphr: I didn't go down this rabbit hole :). I guess it is covered in some other test case

filiphr:
Thanks for the review @sjaakd

Issue #2164, opened 2020-07-20
Parameter no longer recognized in 1.4 when not explicitely mentioned


sjaakd:
compatibillity issue

In 1.3.1.Final:

```java
    @Mapping( target = "value", qualifiedBy = MappingUtil.Fraction3.class )
    @Mapping( target = "uom", constant = BroTransformConstants.UOM_M )
    abstract SFRCType.Offset toOffsetType(BigDecimal value);
```
Lead to implementation: 

```java
    @Override
    Offset toOffsetType(BigDecimal value) {
        if ( value == null ) {
            return null;
        }

        Offset offset = objectFactory.createSFRCTypeOffset();

        offset.setValue( measureTypeMapper.toBigDecimal3( value ) );

        offset.setUom( "m" );

        return offset;
    }
```
But in 1.4.0.Beta3 it is:

```java
    @Override
    Offset toOffsetType(BigDecimal value) {
        if ( value == null ) {
            return null;
        }

        Offset offset = objectFactory.createSFRCTypeOffset();

        offset.setValue( value ); // note: call to qualfied method missing

        offset.setUom( "m" );

        return offset;
    }
```

Note this can be corrected by specifying the source explicely, like this:

```java
    @Mapping( target = "value", source = "value", qualifiedBy = MappingUtil.Fraction3.class )
    @Mapping( target = "uom", constant = BroTransformConstants.UOM_M )
    abstract SFRCType.Offset toOffsetType(BigDecimal value);
```

sjaakd:
Solved by PR: #2193

PR #2193, opened 2020-08-29
#2164 parameter matching should be done based on name when source name is absent

sjaakd:


sjaakd:
@filiphr .. any idea why the integration tests fails.. ? I'm quite puzzled..

filiphr:
It is failing because the test is named incorrectly. It is named `Issue2164Tester.java` and it should be named `Issue2164Test.java`. I also suspect that it is not picked up by the maven surefire plugin for the normal tests as well

sjaakd:
> It is failing because the test is named incorrectly. It is named `Issue2164Tester.java` and it should be named `Issue2164Test.java`. I also suspect that it is not picked up by the maven surefire plugin for the normal tests as well

Ah.. I completely looked over that..

I think BTW that we should refactor this.. I guess that during the construction of `SourceReference` we should already cater for the fact that `source` has the same name as `target`.. This should not be solved in `beanmapping`

WDYT..?

filiphr:
I think that `beanMapping` is the correct place for this. However, I think that we need to do a solution like we are doing in `applyPropertyNameBasedMapping` and `applyParameterNameBasedMapping`, also with the error handling. Because when there are multiple source parameters and both have a property named in the same way then you have to defined the source parameter. Currently we will pick the property from the first source parameter, which is not the same if you don't define it.

I would say that the "// its a plain-old property mapping" `else` should be refactored to use the same logic as `applyPropertyNameBasedMapping` and `applyParameterNameBasedMapping`. You can even add a test with and without `@Mapping` to verify that the output is the same.

sjaakd:
> I think that `beanMapping` is the correct place for this. However, I think that we need to do a solution like we are doing in `applyPropertyNameBasedMapping` and `applyParameterNameBasedMapping`, also with the error handling. Because when there are multiple source parameters and both have a property named in the same way then you have to defined the source parameter. Currently we will pick the property from the first source parameter, which is not the same if you don't define it.
> 
> I would say that the "// its a plain-old property mapping" `else` should be refactored to use the same logic as `applyPropertyNameBasedMapping` and `applyParameterNameBasedMapping`. You can even add a test with and without `@Mapping` to verify that the output is the same

I guess we need to distinguish between:
1. Where
2. When

1. Where: I think that deriving all the segments in "SourceReference" and "TargetReferences" is in the right spot, from an separation of concerns perspective. We should not overload bean mapping with that kind of functionality. So I would assume that when we derive all the segments, we should also take into consideration that the name of target should be used (when the source name is absent). That's more what I meant.

2. when: I agree completely with you that deriving the correctness of the segments in relation to forged- or existing methods is done at the wrong moment.

I did look into merging name-based and defined parameters in one method but always gave up 😄 . But, I guess it would not harm to try again..

sjaakd:
@filiphr on a general note: I don't want to go into refactoring in this solution. It's just that I concluded the logic needs to be there, but can be done in refactored way, making it less brittle. Hence i referred to the refactoring.

Issue #2195, opened 2020-09-04
Possible regression: Wrong mapper code generation when using Builder 


juliojgd:
This is a issue that happens in **1.3.1.Final** and **1.4.0.CR1** but works OK in 1.3.0.Final. So I guess is some kind of regression.

I include a simple project [here](https://github.com/juliojgd/mapstruct_tests) using Java 8 and Lombok 1.18.12 that:
Builds **OK** when:
`$ mvn clean install -Dorg.mapstruct.version=1.3.0.Final`

But generated wrong mapper impl and output compilation failure when:
`$ mvn clean install -Dorg.mapstruct.version=1.3.1.Final`
and when:
`$ mvn clean install -Dorg.mapstruct.version=1.4.0.CR1`

The compilation failure is as follows:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project mapstructtest: Compilation failure: Compilation failure:
[ERROR] /C:/repos/mapstruct_test/juliojgd_mapstructtest/target/generated-sources/annotations/io/github/juliojgd/mapstruct/builder/mapper/BadCodeGeneratedMapperImpl.java:[52,57] incompatible types: io.github.juliojgd.mapstruct.builder.TenthRequest.Builder cannot be converted to io.github.juliojgd.mapstruct.builder.TenthRequest
```

Is this issue registered and is there any plan to fix it? In **1.3.2-SNAPSHOT** is failing also.
 Thanks in advance

sjaakd:
Ah.. builders were introduced as part of 1.3.0.Final. Your code contains builders (lombok)? In 1.3.1.Final and 1.4.0.CR1 you can switch them off. Add: `@Mapper( builder=builder = @Builder( disableBuilder = true ))` to your mapping definition.

juliojgd:
I provided a project that is building fine with 1.3.0.Final. But is failing
to build with 1.3.1.Final , 1.3.2-SNAPSOT and 1.4.0.CR1.

It seems to me like a regression bug.

Sjaak, we don't want to disable a feature that we use and that should work.
We'd better want to see it fixed by reporting the bug and helping to fix it
in the extent we can.

sjaakd:
@juliojgd sorry.. I made an assumption that you accidentally triggered the builder functionality (we had a number of reports on that one: people not knowing that they had builders).. 

[...]

PR #2196, opened 2020-09-05
#2195 @Beanmapping#resultType should be used to construct return type also if it's a builder

sjaakd:


sjaakd:
@filiphr I'm a but struggling with the reproducer A. Do we / or did we support a case were mapping methods return a builder?

The point is: the builder itself requires (sometimes) a factory method. For the target type, we analyse the builder type to check the creation method. But if we're analysing at the builder itself (as return type), we are not going to look for the a creation method in the target type. The issue #2195 seems to suggest something like that.

 It kind of makes sense (since we accept builders also as target type and we have methods that take a target and return type. I need to analyse that case, to see how that works and whether we can reuse that solution. 

What's your opinion here?

filiphr:
>Do we / or did we support a case were mapping methods return a builder?

Mapping method returning a builder is nothing more than mapping into some kind of arbitrary type. We supported that (with builder with `setX`) even before we had the builder support.

I am not quite sure that I follow the problem in #2195. I think I saw one place where the method return type differed than the BeanMapping result type, i.e. the BeanMapping result type was the builder. In my opinion that should be a compile error. 

The regression comes from https://github.com/mapstruct/mapstruct/commit/33710584d33856495992c7cab511b4a16643f53e, right? In my opinion we should get the result type that should be mapped (either via BeanMapping or the method) and then go and look for a builder. Right now we have the `returnTypeBuilder` in the `BeanMappingMethod`, but that can be the incorrect builder.

Not sure if it makes sense what I am saying.

sjaakd:
> Mapping method returning a builder is nothing more than mapping into some kind of arbitrary type. We supported that (with builder with `setX`) even before we had the builder support.

True.. But we only look for factory methods in used mappers and the mapper itself. If you return a builder with a private constructor and a create method, we should use the create mapping and that's something we don't do for normal types.

filiphr:
We don't look for factory methods in any type. We only do a special sequence to determine if a certain type has a builder. And for that we need the builder to have a factory in the type it is building.

filiphr (approved):
LGTM

sjaakd:
@filiphr thanks for the review

PR #2204, opened 2020-09-19
#2125 Parameter matching should be done based on name of all source parameters properties when source name is absent

### LINKED ISSUE #2125: Qualifier is ignored in case of name based mapping
As stated here: https://github.com/mapstruct/mapstruct/issues/2124#issuecomment-647184770 , MapStruct ignores the qualifier specification for a mapping if it is name based (i.e. source property is not specified explicitly).

Imho this behavior could be enhanced either:
- when a qualifier is specified, it can be assumed that the user would like to defer from the automatic (name based) mapping, so MapStruct should honor it, or
- at least the annotation processor should emit a warning that the qualifier is ignored, so the user immediately understands what is going on.

filiphr:
Fixes #2125

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -1068,8 +1068,31 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
                 SourceReference sourceRef = mappingRef.getSourceReference();
                 // sourceRef is not defined, check if a source property has the same name
-                if ( sourceRef == null && method.getSourceParameters().size() == 1 ) {
-                    sourceRef = getSourceRefByTargetName( method.getSourceParameters().get( 0 ), targetPropertyName );
+                if ( sourceRef == null ) {
+                    for ( Parameter sourceParameter : method.getSourceParameters() ) {
+                        SourceReference matchingSourceRef = getSourceRefByTargetName(
+                            sourceParameter,
+                            targetPropertyName
+                        );
+                        if ( matchingSourceRef != null ) {
+                            if ( sourceRef != null ) {
+                                errorOccured = true;
+                                // This can only happen when the target property matches multiple properties
+                                // within the different source parameters
+                                ctx.getMessager()
+                                    .printMessage(
+                                        method.getExecutable(),
+                                        mappingRef.getMapping().getMirror(),
+                                        Message.BEANMAPPING_SEVERAL_POSSIBLE_SOURCES,
+                                        targetPropertyName
+                                    );
+                                break;
+                            }
+                            // We can't break here since it is possible that the same property exists in multiple
+                            // source parameters
+                            sourceRef = matchingSourceRef;
+                        }
+                    }
                 }
 
                 if ( sourceRef == null ) {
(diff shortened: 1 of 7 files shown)


sjaakd (commented):
I'm not so sure whether doing name based selection on all source parameters if you've got more than one automatically (in absence) is a good idea.

We did exclude the option in the past.. Hence the check on the size.

Its also in the documentation: https://mapstruct.org/documentation/dev/reference/html/#mappings-with-several-source-parameters

Perhaps this behaviour doesn't make sense anymore. I do remember however that when I redesigned this years ago I added that check. Perhaps it is related to reuse of reverse configuration. Is that allowed when you have multiple source (so the reverse applies to the source that you indicated as target)?

filiphr:
We do explicit name based mapping if the same property exists in both sides. However, if the same property exists in multiple source parameters you have to define the `source` then.

>In case several source objects define a property with the same name, the source parameter from which to retrieve the property must be specified using the `@Mapping` annotation as shown for the `description` property in the example.

I am doing exactly that with this PR. This is actually aligning with the behaviour that we currently have. Otherwise, the mapping was successful, but we generated wrong code with `@Mapping` was ignored. In this case if there is a property named same in both source parameters we still have a compile error and you have to specify it.

As for the inverse, multi source parameter methods do not qualify as inverse methods. See:

https://github.com/mapstruct/mapstruct/blob/427f5023efec6f6a91fd8946bb2b6e86a1ed2a19/processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java#L304-L310

sjaakd:
I was more referring to the line: `Specifying the parameter in which the property resides is mandatory when using the @Mapping annotation`

filiphr:
OK I understand it now. However, isn't it a bit confusing that when you have a single source parameter `@Mapping(source = "issueId", qualifiedByName = "...")` works, and once you add a second parameter it stops working without any error?

Perhaps we should have a compile error if the parameter does not match the source parameters here:

https://github.com/mapstruct/mapstruct/blob/52ab22bbd87c89610e913ecf6b4a891475846d33/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java#L1088

imeszaros:
> I was more referring to the line: `Specifying the parameter in which the property resides is mandatory when using the @Mapping annotation`

What is the rationale behind this restriction? I mean if it works without a `@Mapping` annotation, why could it not work with an annotation just specifying a qualifier?

> Perhaps we should have a compile error if the parameter does not match the source parameters here:

I agree, anything is better than silent failure / misbehavior.

filiphr:
>What is the rationale behind this restriction? I mean if it works without a `@Mapping` annotation, why could it not work with an annotation just specifying a qualifier?

The rationale is that you if you do `@Mapping(target = "issueId", source = "issueId", qualifiedBy = ...)` it will have a compilation error, because when you have multiple source parameters and you defined `@Mapping` then the `source` has to start with referencing one of the source parameters. 

The idea of `@Mapping(target = "issueId", qualifiedBy = ...)` is to be equivalent to `@Mapping(target = "issueId", source = "issueId", qualifiedBy = ...)`.

We can perhaps relax this restriction and allow writing the source without a source parameter, but this is not for a final release (which we want to do asap for 1.4.0.Final 😉)

imeszaros:
I understand but why is `@Mapping(target = "issueId", source = "issueId", qualifiedBy = ...)`  a compilation error if only one of the arguments contain the `issueId` field?

If I omit the `@Mapping` annotation entirely MapStruct can work it out and takes the value from the argument having the field.

If I just want to qualify this mapping I would not expect that I need to explicitly specify the source.

filiphr:
I agree with you @imeszaros, and we are going to do that for 1.5, but since it is a small change in how things work right now we don't want to include it in the 1.4.0.Final release.

filiphr:
Closing this in favour of PR #2205

PR #2205, opened 2020-09-20
#2125 Report an error when source parameter could not be determined from target mapping

### LINKED ISSUE #2125: Qualifier is ignored in case of name based mapping

filiphr:
Fixes #2125 

This is a second iteration of a solution for #2125, in which we enforce the usage of `source` in `@Mapping` when there are multiple source parameters


sjaakd (approved):
Nice.. LGTM!

PR #2210, opened 2020-09-26
#607 Mapping Iterable<?> object to an object instead of collection

### LINKED ISSUE #607: Mapping Iterable<?> object to an object instead of collection
Mapping an java.lang.Iterable object (e.g. org.springframework.data.domain.Page) to an object it's not possible because the target object must implement the Iterable interface.
If the target object implement the Iterable interface, other fields are not mapped.

sahebpreet:
Fixes #607


core/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -18,7 +18,9 @@
 /**
  * Configures the mapping between two bean types.
  * <p>
- * Either {@link #resultType()}, {@link #qualifiedBy()} or {@link #nullValueMappingStrategy()} must be specified.
+ * When source or the target is of type {@code Iterable} or {@code Stream}, it treats the target as a simple pojo and
```
sjaakd, line 21: some textual stuff: I would use "handles" instead of "treats".. (a bit nitpicking, I agree)..

core/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -18,7 +18,9 @@
...
+ * tries to map the parameters. Otherwise, either {@link #resultType()}, {@link #qualifiedBy()} or
```
sjaakd, line 22: "tries" == attempts

core/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -18,7 +18,9 @@
 /**
  * Configures the mapping between two bean types.
```
sjaakd, line 19: Forces MapStruct to handle target and source as bean and configures the mapping.

sjaakd (changes requested):
First of all: I think you did a nice job @sahebpreet . Thanks for helping us out.

But I was trying to  find a good description now for the `@BeanMapping` annotation and I'm not so sure if the solution @filiphr proposed is such a  good one. (Sorry @filiphr  😄 ) 

If you want to use the `@BeanMapping` for forcing MapStruct to consider an iterable mapping method as a bean mapping we should not do that one sided (only target) and in one situation (iterable) only. That doesn't feel good.

I see two ways forward:
1. Use a dedicated annotation, on method as was proposed. You could extend my work i started in `https://github.com/mapstruct/mapstruct/pull/1744/files`, where I introduced the `@AsBean` annotation you could put on the return type (actually the method, but you could write it in front of the return type) and on parameters. We could gradually build up support and handle different kinds of types (Iterables, Maps)..

2. Go all the way.. But then you have to consider maps, values (as in value mapping), streams and enums. Possibly not advisable, because this one blows up in complexity.

Having said that.. I'm still looking for someone to pick up this Map to Bean stuff and vice versa together with me.. I feel a bit daunted to tackle this alone.. And this feels as a nice first iteration.

sahebpreet:
Thanks @sjaakd for reviewing the pull request and your comments.

I would be happy to help you with Map to Bean stuff. Let me know what subtasks can be picked by me.

filiphr:
Thanks a lot for the work on this PR @sahebpreet. I just had a look at the PR and it looks really good and well though out.

I talked to @sjaakd and I come up with some other ideas for it. I don't think that using `@BeanMapping` here is a bad thing 😄 and I think it would make sense to separate this from the map to bean mapping (which is great that you want to help @sahebpreet, will come up with ways that you can do this).

The current iterable to non-iterable and vice versa mapping can be treated a bit differently. I'll write some examples and ideas below.

## Iterable to Non-Iterable

There are 2 examples here:

* From Custom Domain Iterable, i.e. the Spring Data Page object
* From Iterable to a wrapped list

<details open>
<summary>Custom domain Iterable</summary>

    @Mapper 
    public interface CustomerMapper {

        PageDto<CustomerDto> map(Page<Customer> customers);
    }
</details>

<details open>
<summary>Wrapping a List</summary>

    @Mapper 
    public interface CustomerMapper {

        @Mapping(target = "list", source = "customers")
        ListWrapper<CustomerDto> map(List<Customer> customers);
    }
</details>

In both cases MapStruct will fail with a compile error. However, both cases are valid mappings and we can be smart and detect this. For this we don't even need to add the `@BeanMapping`. We can just go ahead and treat this as a valid bean mapping. This most likely means removing or moving the restriction for mapping from Iterable to Non-Iterable. 

What I mean by moving the restriction is not not allow doing Property Name Based Mapping for some types of iterables (as it doesn't make sense). Those iterables are the ones from the java package.

Actually both of the examples I shared are possible by adding a dummy parameter as a first parameter 😉.

## Non-Iterable to Iterable

I think that this one is a bit simpler. For this we can allow doing bean mapping when the method return type or `BeanMapping#resultType` is not  an Iterable from the java package. Thinking more about this it is possible to do it without `@BeanMapping` as well.

For both of this cases we will need to do some more checks in the build method of the `BeanMappingMethod` and  remove the iterable checks in the `MethodRetrievalProcessor`, but leave the stream checks. I don't see a use case for mapping Stream into bean yet.

----

We will of course need to update the documentation clarifying this restrictions as well.
What do you think about this @sjaakd?

By the way @sahebpreet, your test cases are really good. I would however suggest putting them in `org.mapstruct.ap.test.collection.iterabletononiterable` instead. You can have different variants there.

sjaakd:
> We will of course need to update the documentation clarifying this restrictions as well.
> What do you think about this @sjaakd?

I think it is in line with what we discussed. Some remarks though: 

1. Mapping a list to a wrapped list (your second example below *Iterable to Non-Iterable*) is a 'special case' I would say. Perhaps solve that in another PR to keep it clear what we are doing.

2. The other cases are `bean` - to `non-java-package-iterable-implementation` and vice versa. Right? I agree with the idea. But, what about `non-java-package-iterable-implementation` to `non-java-package-iterable-implementation`?

sjaakd:
> I would be happy to help you with Map to Bean stuff. Let me know what subtasks can be picked by me.

@sahebpreet I'm looking forward to that.. I discussed that with @filiphr as well. We are probably going to abandon the `@AsBean` annotation. Lets discuss that further in #1744 . So, we'll start from a clean slate there..

filiphr:
>1. Mapping a list to a wrapped list (your second example below _Iterable_ to _Non-Iterable_) is a 'special case' I would say. Perhaps solve that in another PR to keep it clear what we are doing.

I will leave this up to @sahebpreet. The changes for this are in the same place as the other changes, that's why I suggested it here.

>2.The other cases are `bean` - to `non-java-package-iterable-implementation` and vice versa. Right? I agree with the idea. But, what about `non-java-package-iterable-implementation` to `non-java-package-iterable-implementation`?

Yes the other cases are `bean` - to `non-java-package-iterable-implementation`.

As for `non-java-package-iterable-implementation` to `non-java-package-iterable-implementation`, I would leave that out for now. That is covered by the current iterable mapping. In case we need to do something, we can do it separately and maybe use an empty `@BeanMapping` to mark the method as a bean mapping instead of iterable mapping.

@sahebpreet will you have time to finish this PR with the last remarks from me and @sjaakd? Thanks a lot for your patience 😄

sahebpreet:
@filiphr & @sjaakd thank you so much for your replies 🙏 .


@filiphr for iterable to Non-iterable case, both the cases could be handled by simply removing the check for iterable in the `parameterType` in `MethodRetreivalProcessor`. Is my understanding correct on this? Would we need any other thing to handle this.

For the Non-iterable to iterable case, I think doing it without requirement of `@BeanMapping` could be a good idea. In that case, I can probably check the resultType package name should not be starting with` java`. If it is starting with `java`, then both source and target should be of type iterable otherwise we can allow them to map. Is my understanding correct on this one as well ?

Thanks

filiphr:
>@filiphr for iterable to Non-iterable case, both the cases could be handled by simply removing the check for iterable in the `parameterType` in `MethodRetreivalProcessor`. Is my understanding correct on this? Would we need any other thing to handle this.

Yes removing the check in the `MethodRetrievalProcessor` is enough. Something additional would be to not allow mapping iterable from `java` into Non-Iterable, because we know we can't do proper mapping there.

>For the Non-iterable to iterable case, I think doing it without requirement of `@BeanMapping` could be a good idea. In that case, I can probably check the resultType package name should not be starting with `java`. If it is starting with java, then both source and target should be of type iterable otherwise we can allow them to map. Is my understanding correct on this one as well ?

Yes your understanding on this one is correct.

sahebpreet:
Hi @filiphr ,

I have pushed the code with latest changes, allowing iterable to non iterable and non iterable to iterable mapping by default except when the result type is from java standard library.

Please help reviewing the PR and let me know about the changes required.

Sorry for taking time, I got a bit busy in something 🙏 .

Thanks

documentation/src/main/asciidoc/chapter-6-mapping-collections.asciidoc
```diff
@@ -90,7 +90,7 @@ carDto.getPassengers().addAll( personsToPersonDtos( car.getPassengers() ) );
 
 [WARNING]
 ====
-It is not allowed to declare mapping methods with an iterable source and a non-iterable target or the other way around. An error will be raised when detecting this situation.
+It is not allowed to declare mapping methods with an iterable source and a non-iterable target from java package or the other way around. An error will be raised when detecting this situation.
```
filiphr, line 93:
```suggestion
It is not allowed to declare mapping methods with an iterable source (from a java package) and a non-iterable target or the other way around. An error will be raised when detecting this situation.
```

processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
```diff
@@ -536,6 +537,11 @@ private boolean checkParameterAndReturnType(ExecutableElement method, List<Param
         return true;
     }
 
+    private boolean isStreamTypeOrIterableFromJavaStdLib(Type type) {
+        return type.isStreamType() || ( type.isIterableType() && type.getPackageName() != null
```
filiphr, line 541: `Type` already has a method called `isJavaLangType` we can use it here.

processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
```diff
@@ -485,7 +486,7 @@ private boolean checkParameterAndReturnType(ExecutableElement method, List<Param
 
         Type parameterType = sourceParameters.get( 0 ).getType();
 
-        if ( parameterType.isIterableOrStreamType() && !resultType.isIterableOrStreamType() ) {
+        if ( !resultType.isIterableOrStreamType() && isStreamTypeOrIterableFromJavaStdLib( parameterType ) ) {
```
filiphr, line 489: Let's keep the order of the checks same as before, first the parameter and then the result type. It makes it clearer when reading the message as well.

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -143,9 +143,9 @@
     RETRIEVAL_DUPLICATE_MAPPING_TARGETS( "Can't generate mapping method with more than one @MappingTarget parameter." ),
     RETRIEVAL_VOID_MAPPING_METHOD( "Can't generate mapping method with return type void." ),
     RETRIEVAL_NON_ASSIGNABLE_RESULTTYPE( "The result type is not assignable to the the return type." ),
-    RETRIEVAL_ITERABLE_TO_NON_ITERABLE( "Can't generate mapping method from iterable type to non-iterable type." ),
+    RETRIEVAL_ITERABLE_TO_NON_ITERABLE( "Can't generate mapping method from iterable type to non-iterable type from java stdlib." ),
```
filiphr, line 146:
```suggestion
    RETRIEVAL_ITERABLE_TO_NON_ITERABLE( "Can't generate mapping method from iterable type from java stdlib to non-iterable type." ),
```

processor/src/test/java/org/mapstruct/ap/test/collection/iterabletononiterable/FruitsMenu.java
```diff
@@ -0,0 +1,35 @@
...
+package org.mapstruct.ap.test.collection.iterabletononiterable;
+
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ *
+ * @author Saheb Preet Singh
+ */
+public class FruitsMenu implements Iterable<org.mapstruct.ap.test.collection.iterabletononiterable.Fruit> {
```
filiphr, line 15:
```suggestion
public class FruitsMenu implements Iterable<Fruit> {
```

processor/src/test/java/org/mapstruct/ap/test/collection/iterabletononiterable/IterableToNonIterableMappingTest.java
```diff
@@ -45,4 +47,34 @@ public void shouldReverseMapStringListToStringUsingCustomMapper() {
         assertThat( source.getNames() ).isEqualTo( Arrays.asList( "Alice", "Bob", "Jim" ) );
         assertThat( source.publicNames ).isEqualTo( Arrays.asList( "Alice", "Bob", "Jim" ) );
     }
+
+    @Test
+    @IssueKey("607")
+    @WithClasses({
+            FruitsMapper.class
```
filiphr, line 54: You can move this in the class annotated `WithClasses`. This might speed up the test a bit since it will do only one compilation per class instead of doing multiple in case the methods are ordered differently. Same comment for the other test methods.

processor/src/test/java/org/mapstruct/ap/test/collection/iterabletononiterable/IterableToNonIterableMappingTest.java
```diff
@@ -45,4 +47,34 @@ public void shouldReverseMapStringListToStringUsingCustomMapper() {
...
+    })
+    public void shouldMapIterableToNonIterable() {
+        List<Fruit> fruits =  Arrays.asList( new Fruit( "mango" ), new Fruit( "apple" ),
+                new Fruit( "banana" ) );
+        FruitsMenu menu = new FruitsMenu(fruits);
+        FruitSalad salad = FruitsMapper.INSTANCE.fruitsMenuToSalad( menu );
+        assertThat( salad.getFruits().get( 0 ).getType() ).isEqualTo( "mango" );
+        assertThat( salad.getFruits().get( 1 ).getType() ).isEqualTo( "apple" );
+        assertThat( salad.getFruits().get( 2 ).getType() ).isEqualTo( "banana" );
```
filiphr, line 63: You can do it in one assertion by using `assertThat( salad.getFruits() ).extracting( Fruit::getType ).containsExactly( "mango", "apple", "banana" )`. Same comment stands for the test method below.

filiphr (changes requested):
It looks really good @sahebpreet. Thanks a lot for your work on this.

I've added some comments and suggestions as well, mostly some polishing of the documentation. Please let us know if you are busy and we'll take care of it when we merge it.

sahebpreet:
Hi @filiphr,

Thank you so much for the review 👍.

Please have a look at the PR now, I have resolved all the concerns.

filiphr (approved):
Great work on this @sahebpreet. Thanks a lot for your changes.

If you are also OK with this @sjaakd, I will go ahead and merge it

sjaakd:
> If you are also OK with this @sjaakd, I will go ahead and merge it

Yes please 👍 Nice job @sahebpreet .... if you consider starting / working on #1075 let me know. Filip and I discussed it two weeks ago in the light of this PR.. We'll take a different approach than I originally considered.

filiphr:
This is now merged. Thanks a lot for your great work on this @sahebpreet.

As @sjaakd mentioned in case you are interested in tackling some other issues feel free to do it :). We are more than happy to help you out.

sahebpreet:
Thank you so much @filiphr and @sjaakd. I would be interested in working in #1075 .

PR #2217, opened 2020-10-03
#2213: primitive arrays should be directly mapped

### LINKED ISSUE #2213: Deep cloning creates intermediate methods for primitive arrays
As in [#1558]

`@Mapper(mappingControl = DeepClone.class)`

Seems to pull in the annotations on byte[] to byte[] copy. If I remove the DeepClone it maps fine, but I need DeepClone.

filiphr:
(we are cloning them anyways)

Additionally fix problem when annotations `ElementType.TYPE_USE` not handled correctly for javac

Fixes #2213

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
@@ -274,7 +274,9 @@ public class TypeFactory {
         else {
             isEnumType = false;
             isInterface = false;
-            name = mirror.toString();
+            // When the component type is primitive and is annotated with ElementType.TYPE_USE then
+            // the typeMirror#toString returns (@CustomAnnotation :: byte) for the javac compiler
+            name = mirror.getKind().isPrimitive() ? NativeTypes.getName( mirror.getKind() ) : mirror.toString();
             packageName = null;
             qualifiedName = name;
             typeElement = null;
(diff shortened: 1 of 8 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -274,7 +274,9 @@ else if (componentTypeMirror.getKind().isPrimitive()) {
         else {
             isEnumType = false;
             isInterface = false;
-            name = mirror.toString();
+            // When the component type is primitive and is annotated with ElementType.TYPE_USE then
+            // the typeMirror#toString returns (@CustomAnnotation :: byte) for the javac compiler
+            name = mirror.getKind().isPrimitive() ? NativeTypes.getName( mirror.getKind() ) : mirror.toString();
```
sjaakd, line 279: I'm not in favor of getting the name via the `mirror.toString()` at all. Can we go via the element somehow? I see somewhat up is the same construction.. (line 263).. 

If not.. go ahead and merge.. I think I even investigated this before in the exact same spot 😄

sjaakd (approved):
I'm a bit apprehensive wrt to the toString.. Have a look at my comment at 279

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
filiphr: We do that when the type is a declared type. Otherwise there is no element, there is only a `TypeKind` and we do use it then the type is primitive. Not sure if we can do more here. 

>I think I even investigated this before in the exact same spot 😄

We did this together for #1558

Issue #2220, opened 2020-10-05
@Mapping with defaultValue and method argument where is missing source does nothing


Saljack:
If you have an `@Mapping` with `defaultValue` and all method argument doesn't contains the source (by name of the target), MapStruct does nothing it simply ignores it (it fails in 1.4.0.Final but is not the root cause see #2214 ). You should use `constant` if there is no source. I think it will be better to fail during MapStruct processing because it is really easy to overlook it. Another solution of this problem can be use the `defaultValue` as `constant`.
```java
@Mapper
public interface MapperWithoutSource {

  class Tenant {
    public String id;
    public String type;
  }

  class TenantDTO {
    public String id;
  }
  
  /**
   *  There is no source "type" in TenantDTO so returned object has type = null. 
   *  I would expect failing during MapStruct processing or type = "STANDARD"
   */
  @Mapping(target="type", defaultValue = "STANDARD")
  Tenant map(TenantDTO tenant);

}
```

filiphr:
Thanks for the report @Saljack, this is actually the same problem as in #2214. When that regression is fixed there will be a compile error saying that there is no `type` in the source `TenantDTO`. 

I think that doing a fallback to constant is error prone and complex in this case. Having a compile error is enough and the user can fix that immediately in the annotation.

Saljack:
Just for clarification there is no error message in 1.3.1.Final too so I think it isn't the same as #2214 , that was the reason why I created this separete issue.

filiphr:
OK not having an error message is wrong then. We will fix that.

Thanks for opening the issue.

filiphr:
I was looking at this now and I realised that indeed it fails on 1.4.0.Final with the error message:

>The type of parameter "tenant" has no property named "type". Please define the source property explicitly.

This error message is correct, because when you are using `@Mapping(target = "type", defaultValue = "STANDARD")` it means that you want to defined the default value for the target. When `ignored`, `constant` or `javaExpression` are missing then `source` is implied. In this case the mapping will not be possible because there is no getter for `type` in the source.

Basically it is identical to having written `@Mapping(target = "type", source = "type", defaultValue = "STANDARD")`.

myucxy:
我感觉应该是 在没有写source 时, target 的 value = defaultvalue;
当写source时, source的getter = null时, target 的 value = defaultvalue;

Issue #2214, opened 2020-10-02
@Mapping needs to define source parameter with defaultValue if you have two arguments in 1.4.0.Final


Saljack:
I found a regression in the new version 1.4.0.Final. If you have a mapper method with two arguments and define a `@Mapping` annotation with a `defaultValue` parameter then you have to define `source` parameter. It is a regression. See the example which works in 1.3.1.Final but it doesn't work in the new 1.4.0.Final.
```java
@Mapper
public interface MapperWithoutSource {

  class Tenant {
    public String id;
    public String type;
    public String parentTenantId;
  }

  class TenantDTO {
    public String id;
    public String type;
  }

  @Mapping(target="type", defaultValue = "STANDARD")
  Tenant map(TenantDTO tenant, String parentTenantId);

}
```
The error is `No property named "type" exists in source parameter(s). Please define the source explicitly.`
As workaround you have to define `source` argument with the method parameter name as: `@Mapping(target="type", defaultValue = "STANDARD", source="tenant.type")`

filiphr:
Thanks for raising this issue @Saljack. We did some changes in this area and something slipped. I think that this issue is the same as #2206. At that time we were not aware about other possible regressions. We are going to look into it and fix this regression.

Saljack:
@filiphr Yes it is the same. These issues types are good because you can write a simply unit test and it will never happen again. Good thing is that thanks to this issue I found an error in my mappers with `defaultValue`. I think that you can close it in favor of #2206 .

filiphr:
Thanks for confirming @Saljack. I will keep this one open as it is slightly different. It will help us with writing the tests for the fix

Issue #2206, opened 2020-09-21
Mapping with qualifying and only target defined should implicitly look for a matching property in one of the parameters


filiphr:
This is basically the PR #2204.

With the following example:

```
class Comment {

    private Integer issueId;
}
class Repository {

    private String name;
}

class RepositoryComment {

    private Integer issueId;
    private String name;

}

@Mapper
interface RepoMapper {

    RepositoryComment clone(Comment comment, Repository repository);

    @Mapping(target = "issueId", defaultValue = "10") // This should match the issueId in Comment
    RepositoryComment clone(Comment comment, Repository repository);
}
```

PR #2222, opened 2020-10-07
#2206, #2214, #2220: Source property should be correctly determined when only target is defined

filiphr:
When having multiple source properties and only target is defined then the same rules should be applied as if there was no mapping:

* First we check for a matching property in any of the source type
* Second we check if the parameter name matches

Fixes: #2206
Fixes: #2214
Fixes: #2220

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -1070,8 +1070,36 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
                 SourceReference sourceRef = mappingRef.getSourceReference();
                 // sourceRef is not defined, check if a source property has the same name
-                if ( sourceRef == null && method.getSourceParameters().size() == 1 ) {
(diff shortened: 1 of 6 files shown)


sjaakd (approved):

filiphr:
Thanks for the reviews @sjaakd

myucxy:
Version :1.4.1.Final
@Mappings({
            @Mapping(source = "iCode", target = "securityId"),
            @Mapping(source = "settleType", target = "settlType"),
            @Mapping(target = "symbol" , defaultValue = "-"),
            @Mapping(target = "matchType" , defaultValue = "9"),
            @Mapping(target = "marketIndicator" , defaultValue = "4")
    })
    CfetsTradeXbondMdSubscribeBondIdBO xbondMdSubBondIdDoFromBondReadyDo(CfetsTradeXbondMdSubscribeBondReadyDO cfetsTradeXbondMdSubscribeBondReadyDO);

java: The type of parameter "cfetsTradeXbondMdSubscribeBondReadyDO" has no property named "symbol". Please define the source property explicitly.

java: The type of parameter "cfetsTradeXbondMdSubscribeBondReadyDO" has no property named "matchType". Please define the source property explicitly.
java: The type of parameter "cfetsTradeXbondMdSubscribeBondReadyDO" has no property named "marketIndicator". Please define the source property explicitly.

PR #2223, opened 2020-10-07
#2221: Fix configuration inheritance when there are multiple matching source parameters of the same type

### LINKED ISSUE #2221: @InheritConfiguration is broken when using multiple source parameters of the same type
- [X] Is this an issue (and hence not a question)? 

I have been trying to inherit a shared configuration for a multi-parameters mapper and I can't get it to work. My classes look like:

```java
@MapperConfig
public interface RestConfig {

  @Mapping(target = "tenantId", source = "tenantId")
  @Mapping(target = "siteId", source = "siteId")
  @Mapping(target = "ctiId", source = "source.cti", qualifiedBy = [...])
  SiteDto convert(RestSiteDto source, String tenantId, String siteId);

}

@Mapper(config = RestConfig.class)
public interface RestSiteMapper {

  @InheritConfiguration
  SiteDto convert(RestSiteDto source, String tenantId, String siteId);

}
```

However I keep getting the error : `Method has no source parameter named "tenantId". Method source parameters are: "source, tenantId, siteId".`

The weird thing is that if I only keep `@Mapping(target = "ctiId", source = "source.cti", qualifiedBy = [...])` the mapping works.

**Edit**: it seems to be a regression (it was working fine up until 1.3.1.Final), so I wonder if it is related to #2206 & #2214.

filiphr:
Fixes #2221

I also did some small refactoring as the `Collectors.reducing( (a, b) -> null )` was really confusing to me.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
@@ -10,7 +10,6 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.stream.Collectors;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.type.DeclaredType;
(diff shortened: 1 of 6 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
```diff
@@ -239,11 +238,24 @@ private Parameter getSourceParameterFromMethodOrTemplate(String parameterName )
...
+                    for ( Parameter sourceParameter : method.getSourceParameters() ) {
+                        if ( sourceParameter.getType().isAssignableTo( parameter.getType() ) ) {
+                            if ( result == null ) {
+                                result = sourceParameter;
+                            }
+                            else {
+                                // When we reach here it means that we found a second source parameter
+                                // that has the same type, then fallback to the matching source parameter
+                                // in the current method
+                                result = Parameter.getSourceParameter( method.getParameters(), parameterName );
```
sjaakd, line 254: is there a situation where the parameter you fetch here is not assignable? It seems like you fetch a parameter again only based on its name.

sjaakd (approved):
Only a question...

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
filiphr: No that is not possible. This is needed because the source reference is from the template method, so the name of the parameter is the one in that method. The name in the mapping method might not be the same. Therefore we need to find it by type. However, if there are multiple parameters of the same type we fallback to finding it by name

PR #2227, opened 2020-10-11
861 remove compiler specific workarounds

sjaakd:
Ok..have a look at whether you like this approach.. I can also bring stuff currently in filters to the new `TypeUtils` .. The nice thing is.. we don't have to pass `Types` and `Elements` as arguments on util methods..


filiphr (commented):
Looks good to me. Did you test it within Eclipse?

The workarounds for Eclipse were executing some of the method calls in different classloaders.

sjaakd:
> The workarounds for Eclipse were executing some of the method calls in different classloaders.

I don't have eclipse on my machines anymore. But fundamentally nothing changed.

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractTypeUtilsDecorator.java
```diff
@@ -52,12 +46,18 @@ public boolean isSameType(TypeMirror t1, TypeMirror t2) {
 
     @Override
     public boolean isSubtype(TypeMirror t1, TypeMirror t2) {
-        return SpecificCompilerWorkarounds.isSubtype( delegate, t1, t2 );
+        if ( isRegularType( t1 ) && isRegularType( t2 ) ) {
+            return delegate.isSubtype( erasure( t1 ), erasure(  t2 ) );
+        }
+        return false;
```
filiphr, line 52: This does not have the same logic as before. We not check if the types are regular types before delegating. Before we returned `false` if `t1` was of `TypeKind#VOID`. 

The current implementation is not forward compatible, as if there is a new `TypeKind` then we will always return `false`.

This comment holds for the other methods that are replaced from the `SpecificCompilerWorkarounds`

filiphr (changes requested):
I had a better look at the replacements and I would suggest to keep the same logic from the `SpecificCompilerWorkarounds` in the `AbstractTypeUtilsDecorator`

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractTypeUtilsDecorator.java
sjaakd: I actually thought about this a bit more.

the  previous logic was incomplete.. I guess that perhaps `module` has been added (I did not check java 11 and later), hence I choose to do the logical thing here.  Checkout the java doc:

```java
   /**
     * Tests whether one type is a subtype of another.
     * Any type is considered to be a subtype of itself.
     *
     * @param t1  the first type
     * @param t2  the second type
     * @return {@code true} if and only if the first type is a subtype
     *          of the second
     * @throws IllegalArgumentException if given an executable or package type
     * @jls 4.10 Subtyping
     */
    boolean isSubtype(TypeMirror t1, TypeMirror t2);
```

I also think that you should check both arguments.

So ` !isRegularType(  t1 ) || !isRegularType(  t2 ) -> false`.. 

Taking "the Morgan"  from boolean math over these:  `!(isRegularType( t1) && isRegularType(t2) )` or the "else branch"  of the current `isRegularType( t1) && isRegularType(t2)`..

The reason why I altered the code is our previous discussion (I can't find it).. But you mentioned other problems with incremental compilation.

filiphr: Due to ECJ throwing an exception in these methods we should check if they are of Type Void and return `false` if one of them is void or `true` if both of them are void. Otherwise, we delegate to the method. The `isRegularType` method is currently implemented based on what the java API does, which can change in the future and we shouldn't rely on current implementation logic.

The problems with the incremental compilation were in the `asMemberOf` I think and both of those issues are fixed, so I think that we are fine there. That is the one with the special classloader logic invocation.

sjaakd: I guess Eclipse does not throw an exception anymore for voids. See my latest commits.. So gradually EJC and Javac show similar behaviour.

filiphr: I think that eclipse does throw an exception, maybe we aren't calling the methods for `Void` types anymore 😉

PR #2249, opened 2020-10-24
#2245 Local variable should be created when using default value

### LINKED ISSUE #2245: Nested source null check calls method twice instead of using intermediate variable
The following mapper:

```
@Mapper
public interface TestMapper {
    class Tenant {
        public String id;
    }
    class Inner {
        public String id;
    }

    class TenantDTO {
        public Inner inner;
    }

    @Mapping(target="id", source = "inner.id", defaultValue = "test")
    Tenant map(TenantDTO tenant);

}
```

We generate the following code:

```
if ( tenantInnerId( tenant ) != null ) {
    tenant1.id = tenantInnerId( tenant );
}
else {
    tenant1.id = "";
}
```

We should instead generate:

```
String id = tenantInnerId( tenant );
if (  id != null ) {
    tenant1.id = id;
}
else {
    tenant1.id = "";
}
```

Note if we use `nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS` then the correct code is generated.

filiphr:
Fixes #2245

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -374,6 +374,11 @@ public class PropertyMapping extends ModelElement {
             return null;
         }
 
+        private boolean hasDefaultValueAssignment(Assignment rhs) {
+            return ( defaultValue != null || defaultJavaExpression != null ) &&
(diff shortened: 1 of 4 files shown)


sjaakd (approved):

filiphr:
Thanks for the review @sjaakd

PR #2261, opened 2020-10-31
#2251 Fix incorrect code generated for constructor mapping from implicit source parameter matching

### LINKED ISSUE #2251: Mapper create invalid implementation - variable is already defined in method

mapstruct version **1.4.1.Final**

Mapper implementation contains multiple definitions of one variable, therefore is not valid.
I will try to illustrate it on the example

```java
public class ModelFrom {
  private String value;
  public ModelFrom(String value1) {
    this.value = value1;
  }
  public String getValue() {
    return value;
  }
  public void setValue(String value) {
    this.value = value;
  }
}
```
```java
public class ModelTo {
  private String value1;
  private String value2;
  public ModelTo(String value1, String value2) {
    this.value1 = value1;
    this.value2 = value2;
  }
  public String getValue1() {
    return value1;
  }
  public void setValue1(String value1) {
    this.value1 = value1;
  }
  public String getValue2() {
    return value2;
  }
  public void setValue2(String value2) {
    this.value2 = value2;
  }
}
```
```java
@org.mapstruct.Mapper(unmappedTargetPolicy = ReportingPolicy.ERROR)
abstract class Mapper {
  @Mapping(target = "value1", source = "value.value")
  public abstract ModelTo mapModelNotWorking(ModelFrom value, String value2);

  @Mapping(target = "value1", source = "value.value")
  @Mapping(target = "value2", source = "value2")
  public abstract ModelTo mapModel(ModelFrom value, String value2);
}
```

Generated `MapperImpl.mapModelNotWorking` contains 2 definitions of variable `value2_1`, therefore is not valid..
```
    @Override
    public ModelTo mapModelNotWorking(ModelFrom value, String value2) {
        if ( value == null && value2 == null ) {
            return null;
        }

        String value1 = null;
        if ( value != null ) {
            value1 = value.getValue();
        }
        String value2_1 = null;
        if ( value2 != null ) {
            value2_1 = value2;
        }

        String value2_1 = null;

        ModelTo modelTo = new ModelTo( value1, value2_1 );

        return modelTo;
    }
```

However in second `mapModel()` if `@Mapping(target = "value2", source = "value2")` is specified, everything works well

filiphr:
Fixes #2251

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -1315,6 +1315,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                         sourceParameters.remove();
                         unprocessedDefinedTargets.remove( targetProperty.getKey() );
                         unprocessedSourceProperties.remove( targetProperty.getKey() );
+                        unprocessedConstructorProperties.remove( targetProperty.getKey() );
                     }
                 }
             }
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2251/Issue2251Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2251/Issue2251Mapper.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright MapStruct Authors.
+ *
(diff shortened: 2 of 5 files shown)


sjaakd (approved):

sjaakd:
LGTM

filiphr:
Thanks for the review @sjaakd

PR #2268, opened 2020-11-05
#2255 Add constants for componentModel

### LINKED ISSUE #2255: Add constants for componentModel
`@Mapper` has an attribute `componentModel` where you can define what component model will be generated but unfortunately there are no constants for it. I checked the `SpringComponentProcessor` and there is string too. So my proposal is to create a new class with constants which can be used there and then you can use it as `@Mapper(componentModel=ComponentModel.SPRING)`.
```java
public class ComponentModel {

  /**
   * the mapper uses no component model, instances are typically retrieved via
   * Mappers.getMapper(Class)
   */
  public static final String DEFAULT = "default";

  /**
   * The generated mapper is an application-scoped CDI bean and can be retrieved
   * via @Inject
   */
  public static final String CDI = "cdi";

  /**
   * The generated mapper is a Spring bean and can be retrieved via @Autowired
   */
  public static final String SPRING = "spring";

  /**
   * The generated mapper is annotated with @javax.inject.Named and @Singleton,
   * and can be retrieved via @Inject
   */
  public static final String JSR330 = "jsr330";

  private ComponentModel() {
  }

}
```

Saljack:
Add `ComponentModel` class with componentModel constants

Fixes #2255


core/src/main/java/org/mapstruct/ComponentModel.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct;
+
+/**
+ * Specifies the component model constants to which the generated mapper should adhere.
+ * It can be used with the annotation {@link Mapper#componentModel()}
```
filiphr, line 10: Or `MapperConfig#componentModel`

filiphr (changes requested):
Thanks for the PR @Saljack.

I am still not 100% convinced that we need a new class only for the component models. We have `MappingConstants`. What do you think @sjaakd?

Let's wait to decide whether we keep this or not. If we keep it then we need to add a `ComponentModelGem` and add tests in the `ConstantTest` in the same way the `MappingConstantsGem` is tested.

One other point, why do we need to change all the tests?

Saljack:
I can revert the tests but I think it is a good place where you can find an example. And if there are constants for componentModel it should be used instead of strings. 
I think that a separet class is better than mixing all constants in one class. One example if you use a code completion you get only valid constants with the separet class `ComponentModel`. Or it can be a inner class of `MappingConstants` but it is not my favorite. 
Otherwise I can rename `ComponentModel` to `ComponentModelConstants` but it is too long from my point of view.

core/src/main/java/org/mapstruct/ComponentModel.java
Saljack: Added

filiphr:
@Saljack we just had a discussion about this with the team. We would like to not expand the number of constants classes that MapStruct has. Therefore, we suggest moving the component model constants as an inner class in the `MappingConstants` class.

Will you have the time to finish this? If not, then please let us know then we can finish this one.

Saljack:
Yes I will prepare it.

Saljack:
@filiphr What is your preference how to use this inner class in code and documentation `ComponentModel.SPRING` or `MappingConstants.ComponentModel.SPRING`?

filiphr:
Thanks for the changes @Saljack. For the documentation I would prefer `MappingConstants.ComponentModel.SPRING`. It shows the location of the string. Everyone can decide on their own how they want to use this

Saljack:
@filiphr ok done. You can check it.

processor/src/main/java/org/mapstruct/ap/internal/gem/ComponentModelGem.java
```diff
@@ -0,0 +1,25 @@
...
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.internal.gem;
+
+/**
+ * Gem for the class {@link org.mapstruct.ComponentModel}
+ *
+ * @author Tomas Poledny
+ */
+public final class ComponentModelGem {
```
filiphr, line 13: This class is not needed anymore, right? Since you added a gem in the `MappingContsantsGem`.

filiphr (commented):
Thanks for your changes @Saljack. 

I have one last comment if you can address it and then this is ready to be merged 😄

processor/src/main/java/org/mapstruct/ap/internal/gem/ComponentModelGem.java
Saljack: Oh my bad. It is removed.

filiphr:
Thanks a lot for the contribution and your patience with us @Saljack. Just finished merging this.

PR #2270, opened 2020-11-06
#2253 remove unmapped source properties when source parameter is directly mapped

### LINKED ISSUE #2253: Wrong reporting for unmapped source properties when source source is implicitly mapped
This is an issue / feature request. I hope I didn't miss this but after reading the reference a few times I didn't find anything along those lines. I do hope that I can get across the point well, but I feel like my explanation isn't that good :| Let's try anyway.

Basically, I have the following situation:
```java
@Mapping(target = "string", source = "string")
Entity map(Dto dto, String string)
```
where `Entity` has a property `String string`. `dto` is my "primary" source for parameters (yes, "primary" is a bit vague), `string` is a property / mapped to a property of `Entity`. Nevertheless, mapstruct scans `String` for additional properties (even though the object itself is already mapped as a property). Moreover, generic `@ObjectFactories` and `@BeforeMapping`s / `@AfterMapping`s match both objects (which is desired in general but might not be in this case).

My suggestion: Similar to `@Context`, introduce a `@Parameter` annotation (or `@Property` / `@Mapped`), which excludes the marked argument from *any* consideration except as a source propery. Maybe one could even the "target" argument to directly assign the argument to a property. So, for example, the above then would be:
```java
Entity map(Dto dto, @Property("string") String string)
```

I guess that #1280 and #1315 are somewhat related.

Why I don't want to "work around" with `@Context`: The additional arguments (the `string` in the above example) typically are database ids which are resolved to database entities using a generic `T object(UUID id, @TargetType Class<T> clazz)` method. If I used `@Context` I would basically need to replicate this generated mapping by hand, except I am missing a trick of `@Context`.

filiphr:
Fixes #2253

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -1315,6 +1315,16 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                         sourceParameters.remove();
                         unprocessedDefinedTargets.remove( targetProperty.getKey() );
                         unprocessedSourceProperties.remove( targetProperty.getKey() );
+
+                        // The source parameter was directly mapped so ignore all of its source properties completely
+                        if ( !sourceParameter.getType().isPrimitive() && !sourceParameter.getType().isArrayType() ) {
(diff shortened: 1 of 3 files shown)


sjaakd (approved):
LGTM..

PR #2275, opened 2020-11-15
#2274, #2023 Fix problems with property mapping using source parameters

### LINKED ISSUE #2023: Multiple source properties with defaultExpression lead to unnecessary null checks and unexpected code
- [✓] Is this an issue (and hence not a question)? 

This mapper definition:
```
package com.turing;

import java.util.UUID;
import org.mapstruct.Mapper;
import org.mapstruct.Mapping;

@Mapper
public interface PersonDtoMapper {
  @Mapping(
      source = "correlationId",
      target = "correlationId",
      defaultExpression = "java(UUID.randomUUID())")
  CreateNewPersonRequest toRequest(final PersonDto person, final UUID correlationId);
}
```
Leads to the following mapper being generated:
```
package com.turing;

import com.turing.CreateNewPersonRequest.Builder;
import java.util.UUID;

@Generated(
    date = "2020-02-06T20:16:29+0100",
    comments = "version: 1.3.1.Final, compiler: javac, environment: Java 13 (AdoptOpenJDK)"
)
public class PersonDtoMapperImpl implements PersonDtoMapper {

    @Override
    public CreateNewPersonRequest toRequest(PersonDto person, UUID correlationId) {
        [...]
        if ( correlationId != null ) {
            if ( correlationId != null ) {
                createNewPersonRequest.setCorrelationId( correlationId );
            }
            else {
                createNewPersonRequest.setCorrelationId( UUID.randomUUID() );
            }
        }

        return createNewPersonRequest.build();
    }
}
```
There is a double `null`-check on `correlationId`. What is worse is that the `else` of the inner `if` can never be reached.

---
Expected behaviour:
The outer `if`-block should be absent:
```
package com.turing;

import com.turing.CreateNewPersonRequest.Builder;
import java.util.UUID;

@Generated(
    date = "2020-02-06T20:16:29+0100",
    comments = "version: 1.3.1.Final, compiler: javac, environment: Java 13 (AdoptOpenJDK)"
)
public class PersonDtoMapperImpl implements PersonDtoMapper {

    @Override
    public CreateNewPersonRequest toRequest(PersonDto person, UUID correlationId) {
        [...]
        if ( correlationId != null ) {
            createNewPersonRequest.setCorrelationId( correlationId );
        }
        else {
            createNewPersonRequest.setCorrelationId( UUID.randomUUID() );
        }

        return createNewPersonRequest.build();
    }
}
```
---
The full code to reproduce this behaviour can be found on [GitHub][GitHub].

[GitHub]: https://github.com/turing85/mapstruct-bug

### LINKED ISSUE #2274: MappingTarget member cannot be set to null
**Expected**
Given a mapper:
```java
@Mapper
public interface ExampleMapper {
  void update(@MappingTarget ExampleTarget target, ExampleSourceVO source, ExampleMember member);
}
```
...and calling it with a `null`-parameter for `member`:
```java
mapper.update(target, source, null);
```
will lead to `target.setMember(null)`.

**Actual**
The mapper implementation is generated as follows:
```java
    [...]
    if ( member != null ) {
      exampleTarget.setMember( member );
    }
    [...]
```
So the member will never be set to `null`.

I tried to work around this using `defaultExpression = "java(null)"`, but this leads to a completely confused implementation:
```java
    if ( member != null ) {
      if ( member != null ) {
        target.setMember( member );
      }
      else {
        target.setMember( null );
      }
    }
```
This weird double check without a chance to reach the else-branch was also mentioned in #2023

Demonstrator:
https://github.com/mickroll/demonstrator/tree/mapstruct-mappingtarget-set-member-to-null
1. `git clone -b mapstruct-mappingtarget-set-member-to-null https://github.com/mickroll/demonstrator.git`
2. `cd demonstrator`
3. `mvn clean install`

Feel free to integrate the included tests as wanted.

**Workaround**
```java
  @AfterMapping
  default void fixNullMember(@MappingTarget ExampleTarget target, ExampleMember member) {
    if (member == null) {
      target.setMember(null);
    }
  }
```
See demonstrator for reference. This workaround fixes the included tests.

filiphr:
Fixes problems when property mapping is using source parameter and has default value / expression or is doing an update

Fixes #2274 
Fixes #2023

I am not entirely happy with the solution, but I don't see currently how we can avoid this. Our `PropertyMapping` know too much about the `BeanMappingMethod` and vice versa

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -460,12 +460,12 @@ public class PropertyMapping extends ModelElement {
             }
 
             if ( rhs.getType().isConverted() ) {
-                // If the type is converted then check is needed
+                // A type conversion is applied, so a null check is required
                 return true;
             }
 
             if ( rhs.getType().isDirect() && targetType.isPrimitive() ) {
-                // If the type is direct and the target type is primtive then check is needed
+                // If the type is direct and the target type is primitive (i.e. we are unboxing) then check is needed
                 return true;
             }
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -46,6 +46,8 @@ import org.mapstruct.ap.internal.util.ValueProvider;
 import org.mapstruct.ap.internal.util.accessor.Accessor;
 import org.mapstruct.ap.internal.util.accessor.AccessorType;
 
+import static org.mapstruct.ap.internal.gem.NullValueCheckStrategyGem.ALWAYS;
+import static org.mapstruct.ap.internal.gem.NullValuePropertyMappingStrategyGem.IGNORE;
 import static org.mapstruct.ap.internal.model.ForgedMethod.forElementMapping;
 import static org.mapstruct.ap.internal.model.ForgedMethod.forParameterMapping;
 import static org.mapstruct.ap.internal.model.ForgedMethod.forPropertyMapping;
@@ -374,11 +376,6 @@ public class PropertyMapping extends ModelElement {
             return null;
         }
 
-        private boolean hasDefaultValueAssignment(Assignment rhs) {
-            return ( defaultValue != null || defaultJavaExpression != null ) &&
-                ( !rhs.getSourceType().isPrimitive() || rhs.getSourcePresenceCheckerReference() != null);
-        }
-
         private Assignment assignToPlain(Type targetType, AccessorType targetAccessorType,
                                          Assignment rightHandSide) {
 
@@ -421,8 +418,7 @@ public class PropertyMapping extends ModelElement {
             }
             else {
                 // If the property mapping has a default value assignment then we have to do a null value check
-                boolean includeSourceNullCheck = SetterWrapper.doSourceNullCheck( rhs, nvcs, nvpms, targetType )
-                    || hasDefaultValueAssignment( rhs );
+                boolean includeSourceNullCheck = setterWrapperNeedsSourceNullCheck( rhs, targetType );
                 if ( !includeSourceNullCheck ) {
                     // solution for #834 introduced a local var and null check for nested properties always.
                     // however, a local var is not needed if there's no need to check for null.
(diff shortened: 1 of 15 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -438,6 +434,49 @@ private Assignment assignToPlainViaSetter(Type targetType, Assignment rhs) {
...
+                return true;
+            }
+
+            if ( rhs.getType().isConverted() ) {
+                // If the type is converted then check is needed
+                return true;
+            }
+
+            if ( rhs.getType().isDirect() && targetType.isPrimitive() ) {
+                // If the type is direct and the target type is primtive then check is needed
```
sjaakd, line 468: add in the comment this is auto-unboxing..

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -438,6 +434,49 @@ private Assignment assignToPlainViaSetter(Type targetType, Assignment rhs) {
...
+         */
+        private boolean setterWrapperNeedsSourceNullCheck(Assignment rhs, Type targetType) {
+            if ( rhs.getSourceType().isPrimitive() && rhs.getSourcePresenceCheckerReference() == null ) {
+                // If the source type is primitive or it doesn't have a presence checker then
+                // we shouldn't do a null check
+                return false;
+            }
+
+            if ( nvcs == ALWAYS ) {
+                // NullValueCheckStrategy is ALWAYS -> do a null check
```
sjaakd, line 453: This is a strange one.. NullValueCheckStrategy does concern parameters, not bean properties.. Do you remember why this one is here? Perhaps it requires some more comment 😄

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -438,6 +434,49 @@ private Assignment assignToPlainViaSetter(Type targetType, Assignment rhs) {
...
+                return true;
+            }
+
+            if ( nvpms == SET_TO_DEFAULT || nvpms == IGNORE ) {
+                // NullValuePropertyMapping is SET_TO_DEFAULT or IGNORE -> do a null check
+                return true;
+            }
+
+            if ( rhs.getType().isConverted() ) {
+                // If the type is converted then check is needed
```
sjaakd, line 463: comment: 'a type conversion is applied' so null check is required.

sjaakd (approved):
I guess splitting this OR statement and documenting it better is a good idea. Perhaps enhance a bit on the comments a bit. Otherwise the solution looks ok to me.

I do think this whole set of rules around the null checks becomes a bit tricky. Especially how they interplay. We should really redesign the three null check settings and clarify those more in 2.0 😄

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
filiphr: I think that you are talking about `NullValueMappingStrategy`. The `NullValueCheckStrategy` is for properties

filiphr:
Thanks for the review @sjaakd

PR #2280, opened 2020-11-18
default component model: mapper reference use singleton INSTANCE if i…

### LINKED ISSUE #2277: Runtime dependency on mapstruct.jar with default component model
When using the default component model, other mapper types used by a mapper are instantiated using Mappers.getMapper(UsedMapper.class).

The unfortunate side effect is that mapsctruct.jar then becomes a mandatory runtime transitive dependency because of the Mappers class.

I haven't found a way to get the code generation use direct instantiation (private final UsedMapper usedMapper = new UsedMapperImpl()) or any other work around.

dmngb:
…t exists

This allows to easily avoid the runtime dependency on mapstruct.jar:
we can avoid Mappers.getMapper(...) for instantiating used mappers if
the code follows the conventionnal pattern for creating mapper singletons.

Fix #2277.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
@@ -6,6 +6,7 @@
 package org.mapstruct.ap.internal.processor;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
@@ -60,6 +61,9 @@ import org.mapstruct.ap.internal.util.Message;
 import org.mapstruct.ap.internal.util.Strings;
 import org.mapstruct.ap.internal.version.VersionInformation;
 
+import static javax.lang.model.element.Modifier.FINAL;
+import static javax.lang.model.element.Modifier.PUBLIC;
+import static javax.lang.model.element.Modifier.STATIC;
 import static org.mapstruct.ap.internal.model.SupportingConstructorFragment.addAllFragmentsIn;
 import static org.mapstruct.ap.internal.model.SupportingField.addAllFieldsIn;
 import static org.mapstruct.ap.internal.util.Collections.first;
@@ -133,9 +137,16 @@ public class MapperCreationProcessor implements ModelElementProcessor<List<Sourc
         List<String> variableNames = new LinkedList<>();
 
         for ( TypeMirror usedMapper : mapperAnnotation.uses() ) {
+            boolean isSingleton = typeUtils.asElement( usedMapper ).getEnclosedElements().stream()
+            .filter( a -> a.getKind().isField() )
+            .filter( a -> a.getModifiers().containsAll( Arrays.asList( PUBLIC, STATIC, FINAL ) ) )
+            .filter( a -> a.getSimpleName().contentEquals( "INSTANCE" ) )
+            .filter( a -> typeUtils.isSameType( a.asType(), usedMapper ) )
+            .anyMatch( a -> true );
             DefaultMapperReference mapperReference = DefaultMapperReference.getInstance(
                 typeFactory.getType( usedMapper ),
                 MapperGem.instanceOn( typeUtils.asElement( usedMapper ) ) != null,
+                isSingleton,
                 typeFactory,
                 variableNames
             );
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
@@ -14,7 +14,9 @@ import java.util.List;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import javax.lang.model.element.Element;
 import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
@@ -77,6 +79,9 @@ import static org.mapstruct.ap.internal.util.Collections.join;
  */
 public class MapperCreationProcessor implements ModelElementProcessor<List<SourceMethod>, Mapper> {
 
+    /** Modifiers for public "constant" e.g. "public static final" */
+    private static final List<Modifier> PUBLIC_CONSTANT_MODIFIERS = Arrays.asList( PUBLIC, STATIC, FINAL );
+
     private ElementUtils elementUtils;
     private TypeUtils typeUtils;
     private FormattingMessager messager;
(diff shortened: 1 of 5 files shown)


dmngb:
@filiphr 
Any feedback on this MR ?
Either on the proposed implementation or even on the idea behind ?

filiphr:
Hey @dmngb, we've been quite busy lately and not enough time to go through the PR.

Did a small glance on this and I think that I like the concept. It is also backwards compatible. Not sure if there are some edge cases that this can cause.

WDYT @sjaakd?

Btw @dmngb how does this fix #2277? You still need to instantiate the mappers in the public static final field. Is there a possibility for a circular class creation?

dmngb:
To avoid use of Mappers factory class at runtime ( #2277 ), this change allows to replace code like:
```
@Mapper interface UsedMapper {
   pubic static final UsedMapper INSTANCE = Mappers.getInstance(UserMapper.class);
}
```
by:
```
@Mapper interface UsedMapper {
   pubic static final UsedMapper INSTANCE = MyCustomMapperFactory.getInstance(UserMapper.class);
}
```
or even:
```
@Mapper interface UsedMapper {
   pubic static final UsedMapper INSTANCE = new UserMapperImpl();
}
```
I am not sure the later works in the general case, but it does in our codebase.

Anyway, I like the added flexiblity given by this change. I think it should be backward compatible in practice.
And it would even benefit those using the more conventional Mappers.getMapper(...) by avoiding to instantiate the mapper more than once!

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -5,14 +5,24 @@
  */
 package org.mapstruct.ap.internal.processor;
 
+import static javax.lang.model.element.Modifier.FINAL;
```
sjaakd, line 8: don't try to re-org the imports (nitty-witty thing.. but helps to review the code 😄

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -133,9 +138,16 @@ public int getPriority() {
         List<String> variableNames = new LinkedList<>();
 
         for ( TypeMirror usedMapper : mapperAnnotation.uses() ) {
+            boolean isSingleton = typeUtils.asElement( usedMapper ).getEnclosedElements().stream()
```
sjaakd, line 141: I don't know this from the top of my head, but does a field in an interface automatically get the modifiers "PUBLIC, STATIC, FINAL" ? 

I guess checkstyle does, but should we warn as well if you spot an intention to use this pattern but for instance left out STATIC?

sjaakd (commented):
I like the idea (a lot). Although I'm a bit cautious, since this is a rather fundamental concept. 

@filiphr is there something we can do in your Intellij plugin to automatically generate the `INSTANCE` field (add a template to create a mapper)?

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
dmngb: Do you want me to rebase and keep the order of imports ?

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
dmngb: Yes, a field on an interface is automatically "public static final".

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -133,9 +138,16 @@ public int getPriority() {
...
+            .filter( a -> a.getKind().isField() )
+            .filter( a -> a.getModifiers().containsAll( Arrays.asList( PUBLIC, STATIC, FINAL ) ) )
+            .filter( a -> a.getSimpleName().contentEquals( "INSTANCE" ) )
+            .filter( a -> typeUtils.isSameType( a.asType(), usedMapper ) )
+            .anyMatch( a -> true );
```
filiphr, line 146: I don't really like that we have so many different method calls here.

I would suggest the following. Extract a method named `hasInstanceMapper` that takes `Element` as a parameter and returns `true` if  these 4 conditions here are satisfied. Additionally instead of using filter you can directly use `anyMatch(this::hasInstanceMapper)`. 

I would also suggest extracting the `Arrays.asList` into a static field in the `MapperCreationProcessor` in order to avoid creating that list all the time.

filiphr (changes requested):
Thanks for your patience @dmngb.

I've done a pass of the PR and it is great. I've left a comment about the use of `filter` methods.

Apart from that, I think that it will be good to add a test case with different variations of the field. e.g. when different name than `INSTANCE` is used, or a different type is used etc.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
filiphr: I don't think it is our job to emit warnings for this. If it is really needed every project can define different rules and have such warnings through checkstyle / sonar

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
dmngb: I just pushed changes to extract  methods/constants (more or less) as suggested.

Or do you prefer a rebase/squash ?

dmngb:
> Apart from that, I think that it will be good to add a test case with different variations of the field. e.g. when different name than `INSTANCE` is used, or a different type is used etc.

I may need some guidance for that. It's  not obvious for me how to write tests in processor/src/test/java/org/mapstruct/ap.
How should I name the test (package/class name) ? Is there any existing test that I may get inspiration from ?

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
filiphr: There is some conflict now, if you can rebase your changes on top of master it would branch, it would be great. If you can't we can do it when merging

filiphr:
>I may need some guidance for that. It's not obvious for me how to write tests in processor/src/test/java/org/mapstruct/ap.
How should I name the test (package/class name) ? Is there any existing test that I may get inspiration from ?

For new features / enhancements we try to put the tests under `org.mapstruct.ap.test` and use a descriptive name for the feature. Maybe under `org.mapstruct.ap.test.defaultcomponentmodel`. As for an inspiration, have a look at `NestedProductPropertiesTest`. The idea is to use `GeneratedSource#addComparisonToFixtureFor` in order to create the fixture and verify that the generate code looks like it is expected.

You can create multiple mappers with different ways of using `Mappers#getMapper`.

e.g.

* `TestNonInstanceMapper MAPPER = Mappers.getMapper(TestNonInstanceMapper.class);
* `TestInstanceMapper INSTANCE = Mappers.getMapper(TestInstanceMapper.class);
* `static final TestInstanceNonPublicMapper INSTANCE = Mappers.getMapper(TestInstanceNonPublicMapper.class);

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
dmngb: I have rebased th MR on top os master.

filiphr:
Thanks for resolving the conflict. I've merged this and it would be part of our next 1.5 release

PR #2298, opened 2020-12-05
#2293 Use MapperConfig instead of MappingConfig in the documentation

### LINKED ISSUE #2293: Where is the @MappingConfig annotation
The documentation referes to an annotation @MappingConfig, but I don't find it in the code or any dependencies as I presumed it should be in the same package as the other annotations. Was this deleted and you forgot to update the documentation or it was deleted by mistake.

filiphr:
`MappingConfig` does not actually exists.

Fixes #2293


sjaakd (approved):

sjaakd:
Ay.. can't believe we missed htat one.. 👍

PR #2305, opened 2020-12-12
Add support for @InheritConfiguration method overloading

rj93:


filiphr:
Thanks for the PR @rj93. The idea is quite good, the only thing that I am not entirely sure is how backwards compatible this is. And additionally what will have if you don't have the source mapping in the simple `MapperConfig`.

i.e. The condition you added to the docs:

>It is also possible to overload the shared configuration's parameters. To do so requires that the `MapperConfig` mapping includes the source variable name.

is not satisfied.

rj93:
For the backwards compatibility, the thing which comes to my mind is that it could break a build if there is a method annotated with `@InheritConfiguration`, that is already overloaded. If that is the case, then the `@InheritConfiguration` isn't actually doing anything, and should be removed. One potential solution to not break builds because of this could be to add a `allowMethodOverloading` to the annotation, defaulted to `false`. Can you think of anything else I should be aware of?

If the source parameter name is not specified in configuration, then the build fails as the processor won't be able to find the variable name. For example, in the test case for this PR if I remove the update `source = "source.name"` to just be `source = "name"` then the test fails with:

```java
[ERROR] org.mapstruct.ap.test.bugs._1720.Issue1720Test.testShouldGiveNoErrorMessage[jdk]  Time elapsed: 1.216 s  <<< FAILURE!
org.junit.ComparisonFailure: [Compilation failed. Diagnostics: [DiagnosticDescriptor: ERROR Issue1720Mapper.java:18 Method has no source parameter named "name". Method source parameters are: "mySource, value".]] expected:<[SUCCEED]ED> but was:<[FAIL]ED>

[ERROR] org.mapstruct.ap.test.bugs._1720.Issue1720Test.testShouldGiveNoErrorMessage[eclipse]  Time elapsed: 1.013 s  <<< FAILURE!
org.junit.ComparisonFailure: [Compilation failed. Diagnostics: [DiagnosticDescriptor: ERROR Issue1720Mapper.java:18 Method has no source parameter named "name". Method source parameters are: "mySource, value".]] expected:<[SUCCEED]ED> but was:<[FAIL]ED>
```

documentation/src/main/asciidoc/chapter-11-reusing-mapping-configurations.asciidoc
```diff
@@ -162,4 +162,33 @@ The attributes `@Mapper#mappingInheritanceStrategy()` / `@MapperConfig#mappingIn
...
 * `AUTO_INHERIT_REVERSE_FROM_CONFIG`: the inverse configuration will be inherited automatically, if the source and target types of the target mapping method are assignable to the corresponding types of the prototype method. If multiple prototype methods match, the ambiguity must be resolved using `@InheritInverseConfiguration(name = ...)` which will cause ``AUTO_INHERIT_REVERSE_FROM_CONFIG` to be ignored.
-* `AUTO_INHERIT_ALL_FROM_CONFIG`: both the configuration and the inverse configuration will be inherited automatically. The same rules apply as for `AUTO_INHERIT_FROM_CONFIG` or `AUTO_INHERIT_REVERSE_FROM_CONFIG`.
\ No newline at end of file
+* `AUTO_INHERIT_ALL_FROM_CONFIG`: both the configuration and the inverse configuration will be inherited automatically. The same rules apply as for `AUTO_INHERIT_FROM_CONFIG` or `AUTO_INHERIT_REVERSE_FROM_CONFIG`.
+
+It is also possible to overload the shared configuration's parameters. To do so requires that the `MapperConfig` mapping includes the source variable name.
```
sjaakd, line 167: But, as I understand you correctly, the source variable name in the `@MapperConfig` does not have to be the same as the source variable name in the mapping method, right?

That's one of the things that I don't like currently.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1720/Issue1720Mapper.java
```diff
@@ -0,0 +1,20 @@
...
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+@Mapper(config = SharedConfig.class)
+public interface Issue1720Mapper {
+
+    Issue1720Mapper INSTANCE = Mappers.getMapper( Issue1720Mapper.class );
+
+    @InheritConfiguration
+    Target map(Source mySource, int value);
```
sjaakd, line 18: so.. what would happen if you have a method with signature:

`Target map( Source mySource1, Source mySource2, int value)` ?  How would you distinguish between `mySource1` and `mySource2`?

processor/src/test/java/org/mapstruct/ap/test/bugs/_1720/SharedConfig.java
```diff
@@ -0,0 +1,20 @@
...
+import org.mapstruct.Mapping;
+import org.mapstruct.Mappings;
+
+@MapperConfig
+public interface SharedConfig {
+
+    @Mappings(
+            @Mapping(target = "fullName", source = "source.name")
+    )
+    Target map(Source source);
```
sjaakd, line 18: The other way around.. what would it mean for the mapping annotated method if your configuration would look like:

```java

@Mapping( target = "fullName1", source = "souce1.name" )
@Mapping( target = "fullName2", source = "souce2.name" )
Target map( Source source1, Source source2, int value);

```
and your mapper

```java
 @InheritConfiguration
Target map( Source mySource)
```

sjaakd (commented):
To be honest: 

I would be in favour of removing the `@InheritConfiguration` and `@InheritReverseConfiguration` alltogether from MapStruct 2.0 in favour of the far more reusable composition mechanism we have introduced recently.

The latter one comes with the drawback its not type safe at the definition spot (the definition of the annotation). However, it is far more flexible and there's a lot you could do with decent errors in that case.

PR #2307, opened 2020-12-14
Example uses wrong maven plugin configuration

reitzmichnicht:
See official maven doc on this: https://maven.apache.org/plugins/maven-compiler-plugin/examples/pass-compiler-arguments.html
Otherwise for example intellij doesn't recognize the compiler options on maven import


filiphr:
Thanks a lot for the patch

PR #2310, opened 2020-12-19
#2301 Implicitly ignore forward inherited mappings from different method types

### LINKED ISSUE #2301: Getting either "Unknown property" or "Unmapped target property"
I updated mapstruct from 1.3.1 to 1.4.1 and now I'm having problems mapping one class:
Having:
```java
    @Override
    @Mapping(target = "dependantBuildRecords", ignore = true)
    /*
     * Builder that MapStruct uses when generating mapper has method dependantBuildRecord() which confuses MapStruct as
     * he thinks it is a new property
     */
    @Mapping(target = "dependantBuildRecord", ignore = true)
    Artifact toEntity(org.jboss.pnc.dto.Artifact dtoEntity);
```
I'm getting following error:
```
ArtifactMapper.java:[68,14] Unknown property "dependantBuildRecord" in result type Artifact. Did you mean "dependantBuildRecords"?
```
And after removing the `@Mapping`
```java
    @Override
    @Mapping(target = "dependantBuildRecords", ignore = true)
    /*
     * Builder that MapStruct uses when generating mapper has method dependantBuildRecord() which confuses MapStruct as
     * he thinks it is a new property
     */
    // @Mapping(target = "dependantBuildRecord", ignore = true)
    Artifact toEntity(org.jboss.pnc.dto.Artifact dtoEntity);
```
I'm getting this error instead:
```
ArtifactMapper.java:[68,14] Unmapped target property: "dependantBuildRecord".
```
It seems that mapstruct is now confused more then before.
This is the Artifact class (stripped of unrelated stuff):
```
public class Artifact implements GenericEntity<Integer> {

    @ManyToMany(mappedBy = "dependencies")
    private Set<BuildRecord> dependantBuildRecords;

    public Set<BuildRecord> getDependantBuildRecords() {
        return dependantBuildRecords;
    }

    public void setDependantBuildRecords(Set<BuildRecord> buildRecords) {
        this.dependantBuildRecords = buildRecords;
    }

    public void addDependantBuildRecord(BuildRecord buildRecord) {
        dependantBuildRecords.add(buildRecord);
        buildRecord.getDependencies().add(this);
    }

    public void removeDependantBuildRecord(BuildRecord buildRecord) {
        dependantBuildRecords.remove(buildRecord);
        buildRecord.getDependencies().remove(this);
    }

    public static class Builder {

        private Set<BuildRecord> dependantBuildRecords = new HashSet<>();

        public Builder dependantBuildRecord(BuildRecord dependantBuildRecord) {
            this.dependantBuildRecords.add(dependantBuildRecord);
            return this;
        }

        public Builder dependantBuildRecords(Set<BuildRecord> dependantBuildRecords) {
            this.dependantBuildRecords = dependantBuildRecords;
            return this;
        }

    }
}
```

filiphr:
Fixes #2301

This is a regression from 1.3. In 1.3 it worked correctly because we were checking the `@Mapping` only once, i.e. when they were defined, but not during the creation of the bean mapping method.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -938,6 +938,15 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
             if ( targetWriteAccessor == null ) {
                 if ( targetReadAccessor == null ) {
+                    if ( mapping.getInheritContext() != null && mapping.getInheritContext().isForwarded() &&
+                        mapping.getInheritContext().getTemplateMethod().isUpdateMethod() != method.isUpdateMethod() ) {
+                        // When a configuration is inherited and the template method is not same type as the current
+                        // method then we can safely ignore this mapping.
+                        // This means that a property which is inherited might be present for a direct mapping
+                        // via the Builder, but not for an update mapping (directly on the object itself),
+                        // or vice versa
+                        return false;
+                    }
                     Set<String> readAccessors = resultTypeToMap.getPropertyReadAccessors().keySet();
                     String mostSimilarProperty = Strings.getMostSimilarWord( targetPropertyName, readAccessors );
 
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2301/Artifact.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2301/Artifact.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2301;
+
+import java.util.Set;
+
+/**
+ * @author Filip Hrisafov
+ */
+public class Artifact {
+
+    private String name;
+    private Set<String> dependantBuildRecords;
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public Set<String> getDependantBuildRecords() {
+        return dependantBuildRecords;
+    }
+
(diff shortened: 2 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -938,6 +938,15 @@ private boolean handleDefinedMapping(MappingReference mappingRef, Type resultTyp
 
             if ( targetWriteAccessor == null ) {
                 if ( targetReadAccessor == null ) {
+                    if ( mapping.getInheritContext() != null && mapping.getInheritContext().isForwarded() &&
+                        mapping.getInheritContext().getTemplateMethod().isUpdateMethod() != method.isUpdateMethod() ) {
```
sjaakd, line 942: I don't get it.. Why would you not inherit a configuration in this case? The whole idea of inheritance is that you don't have to specify mappings for update and create mappings seperately. I'm using this all the time in my own code (defining one config and then using it for both create- and update mappings. That will break now, right?

And why only forward and not reverse? That seems to me like asymetry too..

filiphr: This is an extremely special case, which was handled like this in 1.3 as well. Basically it ignores explicitly defined `@Mapping` for properties that have no write and read accessors once they are inherited. 

Take the test case. The normal mapping has write accessors for `dependantBuildRecord` and `dependantBuildRecords` (due to the builder) and you explicitly ignore both. However, the update mapping does not have an accessor for `dependantBuildRecord` and thus the `@InheritConfiguration` right now leads to an error in compilation.

Before this checks were only done when the `TargetReference` was created and those were done only on the template method. Right now they are done when handling the mapping and thus we have to explicitly handle it here.

Note: this is only done for the error handling, i.e. no target read and write accessors are available and we only ignore the "error" in this special edge case.

sjaakd (approved):

Issue #2318, opened 2020-12-30
Mappings are ignored when @InheritConfiguration defines a mapping for a sub-field


tmonney:
While upgrading from 1.3.1.Final to 1.4.1.Final we stumbled upon what I think is a regression caused by #2103:

When
- a mapping inherits from another mapping using `@InheritConfiguration`
- the inherited mapping uses a source path `a.b` 
- the inheriting mapping uses a source path `a.b.c`, i.e., defines mapping for a sub-field

Then the property mapping for `a.b` in the parent is not generated at all in the mapper implementation class.

Here is an example (getters, setters and constructors omitted for brevity). The complete source code is also available [here](https://github.com/tmonney/mapstruct-inherited-mapping-issue).

### The model
```java
public class SourceParent {
    private Holder holder;

    public static class Holder {
        private String parentValue1;
        private Integer parentValue2;
    }
}

public class SourceChild extends SourceParent {
    private String value;
}

public class TargetParent {
    private String parentValue1;
    private Integer parentValue2;
}

public class TargetChild extends TargetParent {
    private String childValue;
}
```

### The mapping configuration
Please notice that the parent mapping uses `source = "holder"` and the inheriting mapping uses `source = "holder.parentValue2"`:

```java
@MapperConfig
public interface Config {

    @Mapping(target = "parentValue1", source = "holder")
    TargetParent mapParent(SourceParent parent);

    @InheritConfiguration(name = "mapParent")
    @Mapping(target = "childValue", source = "value")
    @Mapping(target = "parentValue2", source = "holder.parentValue2")
    TargetChild mapChild(SourceChild child);
}

@Mapper(config = Config.class)
public interface MyMapper {

    @InheritConfiguration(name = "mapChild")
    TargetChild mapChild(SourceChild child);

    default String parentValue1(SourceParent.Holder holder) {
        return holder.getParentValue1();
    }
}
```

### The generated mapper
Please notice that there is no code generated for `TargetChild.parentValue1`.
```java
public class MyMapperImpl implements MyMapper {
    @Override
    public TargetChild mapChild(SourceChild child) {
        if ( child == null ) {
            return null;
        }

        TargetChild targetChild = new TargetChild();

        targetChild.setChildValue( child.getValue() );
        targetChild.setParentValue2( childHolderParentValue2( child ) );

        return targetChild;
    }

    private Integer childHolderParentValue2(SourceChild sourceChild) {
        if ( sourceChild == null ) {
            return null;
        }
[...]

PR #2321, opened 2021-01-04
#2318 Fix missing mapping when source is redefined

tmonney:
Naive attempt at fixing #2318 by not considering redefined source mappings for exclusion. 
I am absolutely not sure if this is the right way to do it but at least it fixes the problem without breaking the unit tests of related issue #2101.


filiphr:
Closing this in favour of #2332

Issue #2278, opened 2020-11-16
@InheritInverseConfiguration and @Mapping(target:"bean.childProperty", ignore=true) do not play together


neowin35:
```java
public interface CarMapper {
    class Details {
        public String brand;
        public String model;
    }
    class DetailsDTO {
        public String brand;
    }
    class Car {
        public Details details;
    }
    class CarDTO {
       public DetailsDTO detailsDTO;
    }

    @Mapping(target = "detailsDTO", source = "details")
    CarDTO map(Car in);

    @InheritInverseConfiguration
    @Mapping(target = "details.model", ignore = true)
    Car map(CarDTO in);
}
```

This example mapper results at 1.3.1.Final to this code to be generatedfor the BrandDTO -> Brand mapping:
```java
protected Details detailsDTOToDetails(DetailsDTO detailsDTO) {
        if ( detailsDTO == null ) {
            return null;
        }
        Details details = new Details();
        details.brand = detailsDTO.brand;

        return details;
    }
```
However after upgrading to 1.4.1.Final this is what gets generated: 
```java
protected Details carDTOToDetails(CarDTO carDTO) {
        if ( carDTO == null ) {
            return null;
        }
        Details details = new Details();

        return details;
    }
```

For some strange reason mapstruct is now trying to map CarDTO to Details. In case I specify the full path to the property to ignore like `in.details.model` than it seems to work. Is that an intentional change or just a regression issue?

neowin35:
@filiphr any idea about this?

filiphr:
We'll need to investigate a bit more about this. Thanks for reporting it. I'll put it on 1.4.2 so we can have a look at it.

filiphr:
@sjaakd the cause of this issue is the fix of https://github.com/mapstruct/mapstruct/issues/2101. Maybe you can have a look at it, I can't seem to understand why exactly it is not working

filiphr:
Fixed in https://github.com/mapstruct/mapstruct/pull/2332

Issue #2322, opened 2021-01-05
Mapping with source="." ignored on inner objects when another mapping is specified


Tiller:
Hello,
I'm not entirely sure this is a bug as the use of `source="."` doesn't seem to be documented, but here it is anyway.

Considering the following example:
```java
public class A {
    public int field1A;
    public int field2;
    public int field3;
}

class B {
    public int field1B;
    public int field2;
    public int field3;
}

class Wrap {
    public B b;
}

@Mapper
interface MapInner {

    // @Mapping(source = "field1A", target = "b.field1B")
    @Mapping(source = ".", target = "b")
    Wrap map(A a);
}

@Mapper
interface MapDirect {

    // @Mapping(source = "field1A", target = "field1B")
    @Mapping(source = ".", target = ".")
    B map(A a);
}
```

The resulting mapping does bind `field2` and `field3` automatically, both for the direct A->B mapping and the A->Wrap(B)

```java
class MapInnerImpl implements MapInner {
    [...]
    protected B aToB(A a) {
        [...]
        B b = new B();

        b.field2 = a.field2;
        b.field3 = a.field3;

        return b;
    }
}

class MapDirectImpl implements MapDirect {

    @Override
    public B map(A a) {
        [...]
        B b = new B();

        b.field2 = a.field2;
        b.field3 = a.field3;

        return b;
    }
}
```

However, when uncommenting the annotation that gives special instructions for an unmapped field, mapstruct no longer maps `field2` and `field3` for the A->Wrap(B) mapping but does it perfectly for the A->B mapping

```java
class MapInnerImpl implements MapInner {
    [...]
    protected B aToB(A a) {
        [...]
        B b = new B();

        b.field1B = a.field1A;

        return b;
    }
}


class MapDirectImpl implements MapDirect {

    @Override
    public B map(A a) {
        [...]
        B b = new B();

        b.field1B = a.field1A;
        b.field2 = a.field2;
        b.field3 = a.field3;

        return b;
    }
}
```

I'm using version 1.4.1.Final

filiphr:
The fact that dot (`.`) works in `source` is an accident. I don't think that we support that, so anything that works with that is an accident.

Why do you have to use `.`? Why can't you use the name of the parameter?

Tiller:
> Why can't you use the name of the parameter?

It's not that I can't, it's just that as I have numerous parameters, using `source="."` is simplier and faster to map all fields

Tiller:
[...]

PR #2332, opened 2021-01-16
#2278 inherited property ignored entirely due to ignore on nested level

sjaakd:


filiphr:
The fix looks OK. However, I think that it fixes only the specific case in #2278. 

What if instead of ignoring it we use a `constant`?

I also have a feeling that #2318 should be fixed in the same time as #2278 as those are more or less similar.

sjaakd:
> What if instead of ignoring it we use a `constant`?

I thought about that as well. It comes down on how you see merging of properties. I'll check what the "normal" behavior is (so how we merge properties).

sjaakd:
I guess.. in the end I flipped just 2 arguments (not properly done when refactoring) and needed to add a null check for when the source is absent (I thought about making it equal to the target  but I remember somehow that's always right, for reversing you need to specify them both, but in case of ignoreByDefault you activate individual mappings)...

sjaakd:
Replaces #2321

filiphr:
Thanks for doing this @sjaakd, the fix looks really simple.

One last things. Can we also add the test case from #2322? I think that it is the same problem and that this fixes that as well.

sjaakd:
> One last things. Can we also add the test case from #2322? I think that it is the same problem and that this fixes that as well.

I'll have a look..

sjaakd:
> One last things. Can we also add the test case from #2322? I think that it is the same problem and that this fixes that as well.

I created a test. But it still fails. The current fix is fixing stacking of inherited(reverse) properties. This seems to be a problem in our beanmapping (the reproducer does not call the logic that is fixed).

Let treat that as a separate issue. I can push the reproducer after merging this one. Perhaps you can have a look as well?

filiphr:
Sounds good. Thanks for trying it out. Go ahead and merge this. I'll take a look at the reproducer and try to find a fix

sjaakd:
> Sounds good. Thanks for trying it out. Go ahead and merge this. I'll take a look at the reproducer and try to find a fix

See https://github.com/mapstruct/mapstruct/pull/2341

I'm not sure whether I've got time for it today.. It seems to be located somewhere in the `BeanMapping` - nested mapping handling.

PR #2333, opened 2021-01-16
#2250, #2324 Do not throw error when qualifier is missing on iterable mapping

### LINKED ISSUE #2250: Cannot map a collection with named mapping method
After upgrading to mapstruct 1.4.1 we are having some problems with mapping collections using named mapping methods. I suspect this is a bug b/c it worked fine for mapstruct 1.2.0.

The full example can be found here: https://gist.github.com/MaciejDobrowolski/6a9b9bedfd873e8ffd822301ab333be4

I am getting the following errors:

> Demo.java:36: error: Qualifier error. No method found annotated with `@Named#value`: [ fromBase ]. See https://mapstruct.org/faq/#qualifier for more info.
>     House toEntity(HouseDto houseDto);
>           ^
> Demo.java:36: error: Can't map property "List<aseDto> dogs" to "List<og> dogs". Consider to declare/implement a mapping method: "List<og> map(List<aseDto> value)".
>     House toEntity(HouseDto houseDto);
>           ^

+short in-ticket summary:

Having the following model:
```
class BaseEntity {
    private Long id;
}

class Dog extends BaseEntity {
    private String name;
}

class House extends BaseEntity {
    private List<Dog> dogs;
}

class BaseDto {
    private Long id;
}

class HouseDto extends BaseDto {
    private List<BaseDto> dogs;
}
```

And the following mapper:

```
@Named("fromBase")
public <T> T fromBase(BaseDto baseDto, @TargetType Class<T> entityClazz) {
    return Optional.ofNullable(baseDto)
            .map(BaseDto::getId)
             // somehow create instance of T from id
            .orElse(null);
}
```
The above mapper is not used by the one below:

```
@Mapping(target = "dogs", qualifiedByName = "fromBase")
House toEntity(HouseDto houseDto);
```

If I only change `dogs` to be a single `dog` (without `List`), it works as expected.

update: I have just removed generics and `@TargetType` from `fromBase` mapping method and it still returns the same error

### LINKED ISSUE #2324: Why did you make qualifiedByName more strict with collection mappings in 1.4.1.Final?
I'm talking about this comment here
https://github.com/mapstruct/mapstruct/issues/2250#issuecomment-719910576

It feels like this is a bit of a downgrade, because now we're forced to create a `@Named` method for the Collection mapping and in it, we have to iterate the collection and call the single object mapping method manually.

Consider the following example that worked in `1.3.1.Final`:
```java
@Mapper
public abstract class SalesRuleMapper {

    @Autowired
    private SalesRuleRepository salesRuleRepository;

    abstract void updateSalesRuleFromEventBusOrderSalesRuleDTO(@MappingTarget SalesRule salesRule, EventBusOrderSalesRuleDTO eventBusOrderSalesRuleDTO);

    abstract SalesRule mapFromEventBusOrderSalesRuleDTO(EventBusOrderSalesRuleDTO eventBusOrderSalesRuleDTO, @Context OmsContext omsContext);

    @Named("mapFromEventBusOrderSalesRuleDTOWithOmsContext")
    public SalesRule mapFromEventBusSalesRuleDTO(EventBusOrderSalesRuleDTO eventBusOrderSalesRuleDTO, @Context OmsContext omsContext) {
        SalesRule existingSalesRule = salesRuleRepository
                .findByNameAndCouponCodeAndFromDateAndToDateAndStoreId(eventBusOrderSalesRuleDTO.getName(),
                        eventBusOrderSalesRuleDTO.getCouponCode(),
                        eventBusOrderSalesRuleDTO.getFromDate(),
                        eventBusOrderSalesRuleDTO.getToDate(),
                        omsContext.getStore().getId());
        if (existingSalesRule != null) {
            updateSalesRuleFromEventBusOrderSalesRuleDTO(existingSalesRule, eventBusOrderSalesRuleDTO);
            return existingSalesRule;
        }
        SalesRule newSalesRule = mapFromEventBusOrderSalesRuleDTO(eventBusOrderSalesRuleDTO, omsContext);
        return newSalesRule;
    }
}
```

```java
@Mapper(uses = {
        SalesRuleMapper.class
}, injectionStrategy = InjectionStrategy.CONSTRUCTOR)
public abstract class OrderMapper {

    @Autowired
    protected OmsContext omsContext;

    @Mapping(source = "salesRules", target = "salesRules", qualifiedByName = "mapFromEventBusOrderSalesRuleDTOWithOmsContext")
    public abstract Order fromEventBusOrder(EventBusOrderDTO eventBusOrderDTO, @Context OmsContext omsContext);

    public Order fromEventBusOrder(EventBusOrderDTO eventBusOrderDTO) {
        return fromEventBusOrder(eventBusOrderDTO, omsContext);
    }
}
```
MapStruct would internally generate the method that'd iterate over the collections and map each item:
```java
    protected Set<SalesRule> eventBusOrderSalesRuleDTOArrayToSalesRuleSet(EventBusOrderSalesRuleDTO[] eventBusOrderSalesRuleDTOArray, OmsContext omsContext) {
        if ( eventBusOrderSalesRuleDTOArray == null ) {
            return null;
        }

        Set<SalesRule> set = new HashSet<SalesRule>( Math.max( (int) ( eventBusOrderSalesRuleDTOArray.length / .75f ) + 1, 16 ) );
        for ( EventBusOrderSalesRuleDTO eventBusOrderSalesRuleDTO : eventBusOrderSalesRuleDTOArray ) {
            set.add( salesRuleMapper.mapFromEventBusSalesRuleDTO( eventBusOrderSalesRuleDTO, omsContext ) );
        }

        return set;
    }
```

But now, in 1.4.1.Final, `eventBusOrderSalesRuleDTOArrayToSalesRuleSet` doesn't get generated automatically anymore. 
And since the `@Named` annotation now has to go on the Array to Collection method and not the individual object, this presents an inconvenience for the user, because we have to write the Collection mapping method and call the individual mapping method manually. Error prone code like this is something MapStruct intends to avoid.

Is there a reason why the decision has been made to add this limitation?

filiphr:
With this change we are relaxing the error handling when qualifiers are used in `@Mapping`
and we allow defining qualifiers for collections / array mapping to mean the same as if it was defined on `@IterableMapping`
i.e. apply the qualifier on the element mapping

Fixes #2250
Fixes #2324

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethodBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethodBuilder.java
@@ -22,6 +22,8 @@ import org.mapstruct.ap.internal.util.Strings;
 
 import static org.mapstruct.ap.internal.util.Collections.first;
 
+import javax.lang.model.element.AnnotationMirror;
+
 /**
  * Builder that can be used to build {@link ContainerMappingMethod}(s).
  *
@@ -37,6 +39,7 @@ public abstract class ContainerMappingMethodBuilder<B extends ContainerMappingMe
     private FormattingParameters formattingParameters;
     private String errorMessagePart;
     private String callingContextTargetPropertyName;
+    private AnnotationMirror positionHint;
 
     ContainerMappingMethodBuilder(Class<B> selfType, String errorMessagePart) {
         super( selfType );
@@ -55,6 +58,11 @@ public abstract class ContainerMappingMethodBuilder<B extends ContainerMappingMe
 
     public B callingContextTargetPropertyName(String callingContextTargetPropertyName) {
         this.callingContextTargetPropertyName = callingContextTargetPropertyName;
+        return myself;
+    }
+
+    public B positionHint(AnnotationMirror positionHint) {
+        this.positionHint = positionHint;
         return myself;
     }
 
@@ -88,7 +96,7 @@ public abstract class ContainerMappingMethodBuilder<B extends ContainerMappingMe
             formattingParameters,
             criteria,
             sourceRHS,
-            null,
+            positionHint,
             () -> forge( sourceRHS, sourceElementType, targetElementType )
         );
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -656,6 +656,7 @@ public class PropertyMapping extends ModelElement {
                 .method( methodRef )
                 .selectionParameters( selectionParameters )
                 .callingContextTargetPropertyName( targetPropertyName )
+                .positionHint( positionHint )
                 .build();
 
             return createForgedAssignment( source, methodRef, iterableMappingMethod );
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
@@ -298,7 +298,13 @@ public class MappingResolverImpl implements MappingResolver {
             }
 
             if ( hasQualfiers() ) {
-                printQualifierMessage( selectionCriteria );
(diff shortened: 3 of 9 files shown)


sjaakd (approved):
LGTM

filiphr:
Thanks for the review

PR #2337, opened 2021-01-20
Issue 1617: Add empty constructor with @inject annotation to mapper on  jsr330 componentmodel

### LINKED ISSUE #1617: Add empty constructor with @inject annotation to mapper on jsr330 componentmodel
When using dagger2, an object can only be injected if it has an `@Inject` constructor. When component model is set to jsr330, this constructor is added when `InjectionStrategy` is set to `CONSTRUCTOR` and if there is at least one dependency used. If there are no dependencies then no constructor is created and thus it can't be injected.

zlymeda:
Fixes #1617 

I had to upgrade spring to 4.3.5 to tolerate empty constructors (see https://github.com/spring-projects/spring-framework/issues/19572)


filiphr:
Thanks for the PR @zlymeda.

I would suggest adding a new componentModel named `dagger2` that would be responsible for creating the appropriate classes. Looking at your changes they are tailored a lot for dagger 2 already (the fact that dagger requires an empty constructor with `@Inject` or injected fields).

PR #2341, opened 2021-01-23
#2322 reproducer

sjaakd:


filiphr:
This is replaced by https://github.com/mapstruct/mapstruct/pull/2342. I couldn't push to your branch, so I pushed it to mapstruct repo.

PR #2353, opened 2021-02-05
#2352 Add source element type for Iterable mappings

### LINKED ISSUE #2352: Import statement not being generated
- [x ] Is this an issue (and hence not a question)? 

Hi,

I have encountered an issue where an import is not being generated correctly.  It is relatively simple to replicate.

1. Domain classes.
````java
package org.woody.mapstruct.bug.dto;
...
@Value
public class TheDto {
    @NonNull
    String id;
}

package org.woody.mapstruct.bug.model;
...
@Value
public class TheModel {
    @NonNull
    String id;
}

package org.woody.mapstruct.bug.model;
...
public class TheModels extends ArrayList<TheModel> {
    private static final long serialVersionUID = 1L;
}
````

2. The mappers
````java
package org.woody.mapstruct.bug.mapper;
...
@Mapper
public interface TheModelMapper  {
    TheDto convert(TheModel theModel);
}

package org.woody.mapstruct.bug.mapper;
...
@Mapper(uses = TheModelMapper.class)
public interface TheModelsMapper  {
    List<TheDto> convert(TheModels theModels);
}
```` 

This causes the following code to be generated for the TheModelsMapperImpl
````java
package org.woody.mapstruct.bug.mapper;

import java.util.ArrayList;
import java.util.List;
import javax.annotation.Generated;
import org.woody.mapstruct.bug.dto.TheDto;
import org.woody.mapstruct.bug.model.TheModels;

@Generated(
    date = "2021-02-03T21:41:48+0100",
    comments = "version: 1.4.2.Final, compiler: javac, environment: Java 1.8.0_251 (Oracle Corporation)"
)
public class TheModelsMapperImpl implements TheModelsMapper {

    private final TheModelMapper theModelMapper = Mappers.getMapper( TheModelMapper.class );

    @Override
    public List<TheDto> convert(TheModels theModels) {
        if ( theModels == null ) {
            return null;
        }

        List<TheDto> list = new ArrayList<TheDto>( theModels.size() );
        for ( TheModel theModel : theModels ) {  // <-- note here we reference TheModel but we have no import for it
            list.add( theModelMapper.convert( theModel ) );
        }

        return list;
    }
}
```` 

I've noticed this as I'm moving across to using the Spring conversion concept and as a result have had to break up some of my mappers.  The above mapper without the uses generates all correctly as it will generate the additional mapping method which will cause the import to be created.

Currently I just have a dummy mapping method to force the import to occur.

I have a sample project if necessary.

cheers,
Andrew

filiphr:
When the Iterable type we are mapping is not generic
(i.e. it is a custom type extending an Iterable) then the source element type
which is included in the loop was not imported.

Fixes #2352

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/IterableMappingMethod.java
@@ -9,6 +9,7 @@ import static org.mapstruct.ap.internal.util.Collections.first;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Set;
 
 import org.mapstruct.ap.internal.model.assignment.LocalVarWrapper;
 import org.mapstruct.ap.internal.model.assignment.SetterWrapper;
@@ -84,6 +85,14 @@ public class IterableMappingMethod extends ContainerMappingMethod {
             afterMappingReferences,
             selectionParameters
         );
+    }
+
+    @Override
+    public Set<Type> getImportTypes() {
+        Set<Type> types = super.getImportTypes();
+
+        types.add( getSourceElementType() );
+        return types;
     }
 
     public Type getSourceElementType() {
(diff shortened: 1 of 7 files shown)


sjaakd (approved):

filiphr:
Thanks for the review @sjaakd

Issue #2339, opened 2021-01-21
Enum mapping. Throw exception for unexpected input


Chaos-UA:
For example we have

```
class EnumA {
       VALUE1,
       VALUE2
}
```

and 

```
class EnumB {
       VALUE1,
       VALUE2,
       VALUE3
}
```

It would be very helpfull to add posibility to throw exception when "VALUE3" will be as an input value instead of failing compilation. Something like this:
```
@ValueMapping(target = MappingConstants.THROW_EXCEPTION, source = "VALUE3")
EnumA toEnumA(EnumB source)
```

At the moment it is only possible to set it to null or to another existing value, but not to throw some IllegalArgumentException etc.

filiphr:
This sounds like a reasonable improvement.

What would the exception message contain? Currently, for the default exception throwing it is "Unexpected enum constant: XXX". Does it make sense to throw this exception as well?

In addition for enum mappings this can also be used to remove the missing Any Unmapped or Any Remaining warning for String to Enum mappings.

I think that this issue is relatively easy to implement. @Chaos-UA would you me interested in providing a PR for it? Otherwise, is someone else from the community interested in adding this functionality? The MapStruct team is of course always here to help out.

filiphr:
Some discussions from PR https://github.com/mapstruct/mapstruct/pull/2354.

We need to think what we should do when inversing something like 

```
@ValueMapping(target = MappingConstants.THROW_EXCEPTION, source = "VALUE3")
```

most likely we should ignore them, as they don't make much sense. Having said that `MappingConstants.THROW_EXCEPTION` should also be only applicable to `ValueMapping#target()` and not `ValueMapping#source()` since we can't map from exceptions.

JudeNiroshan:
@filiphr  I'd like to give a try. This would be my first contribution to mapstruct.

filiphr:
Thanks for the initiative @JudeNiroshan. I've assigned this one to you. Please let us know if you have any questions

JudeNiroshan:
@filiphr can you help me with running the unit tests in IntelliJ IDEA?

Whenever I try to run a test class(entire class), I get this error. 

```
Process finished with exit code 1
Class not found: "org.mapstruct.ap.test.value.enum2enum.EnumToEnumMappingTest"
```

I believe there is something that needs to be changed in my run configurations. This is how it looks:
[...]

PR #2354, opened 2021-02-06
#2350 Generate core string to enum mapping when AnyRemaining or AnyUnmapped is not used

### LINKED ISSUE #2350: When using @ValueMapping string-to-enum mapping results in no cases in switch
My code:

```
@Mapper
public interface TransactionMapper {
    @Mapping(source = "fullName", target = "name")
    @Mapping(expression = "java(payload.getDonorFirstName() + \" \" + payload.getDonorLastName())", target = "donorName")
    @Mapping(source = "donorEmailAddress", target = "donorEmail")
    @Mapping(constant = "false", target = "processed")
    @Mapping(source = "schedule.givingFrequency", target = "recurringFrequency")
    @Mapping(source = "schedule.startDate", target = "recurringStartDate")
    @Mapping(source = "schedule.processImmediately", target = "processImmediately")
    TransactionEntity payloadToEntity(TransactionsPayload payload);

    @ValueMapping(source = "ONETIME", target = "ONE_TIME")
    TransactionType transactionTypeStringToTransactionType(String transactionTypeString);
}
```

The possible `String` values for `transactionTypeString` are `ONETIME`, `RECURRING` and `PLEDGE`. The `TransactionType` enum has values of `ONE_TIME`, `RECURRING` and `PLEDGE`. Not the underscore in `ONE_TIME`. That's why I have to specify the mapping.

The code generated has no `case` statements for each value and only the `default` that throws an exception:

```
@Override
    public TransactionType transactionTypeStringToTransactionType(String transactionTypeString) {
        if ( transactionTypeString == null ) {
            return null;
        }

        TransactionType transactionType;

        switch ( transactionTypeString ) {
            default: throw new IllegalArgumentException( "Unexpected enum constant: " + transactionTypeString );
        }
    }
```

If I change my mapper to:

```
@ValueMapping(source = "ONETIME", target = "ONE_TIME")
@ValueMapping( source = MappingConstants.ANY_REMAINING, target = "RECURRING" )
TransactionType transactionTypeStringToTransactionType(String transactionTypeString);
```

the generated code is:

```
@Override
    public TransactionType transactionTypeStringToTransactionType(String transactionTypeString) {
        if ( transactionTypeString == null ) {
            return null;
        }

        TransactionType transactionType;

        switch ( transactionTypeString ) {
            case "ONETIME": transactionType = TransactionType.ONE_TIME;
            break;
            case "ONE_TIME": transactionType = TransactionType.ONE_TIME;
            break;
            case "RECURRING": transactionType = TransactionType.RECURRING;
            break;
            case "PLEDGE": transactionType = TransactionType.PLEDGE;
            break;
            default: transactionType = TransactionType.RECURRING;
        }

        return transactionType;
    }
```

which is close. I need the `default` to still throw an exception.

filiphr:
Fixes #2350

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/value/string2enum/StringToEnumMappingTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/value/string2enum/StringToEnumMappingTest.java
@@ -16,6 +16,7 @@ import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutco
 import org.mapstruct.ap.testutil.runner.AnnotationProcessorTestRunner;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 @IssueKey( "1557" )
 @WithClasses({ OrderType.class, OrderMapper.class })
@@ -53,7 +54,20 @@ public class StringToEnumMappingTest {
                     "type String to an enum type." )
         }
     )
-    public void shouldRaiseErrorWhenUsingAnyRemaining() {
+    public void shouldRaiseWarningWhenNotUsingAnyRemainingOrAnyUnmapped() {
+
+        assertThatThrownBy( () -> ErroneousOrderMapperUsingNoAnyRemainingAndNoAnyUnmapped.INSTANCE.map( "unknown" ) )
+            .isInstanceOf( IllegalArgumentException.class )
+            .hasMessage( "Unexpected enum constant: unknown" );
+
+        assertThat( ErroneousOrderMapperUsingNoAnyRemainingAndNoAnyUnmapped.INSTANCE.map( null ) )
+            .isEqualTo( OrderType.STANDARD );
+
+        assertThat( ErroneousOrderMapperUsingNoAnyRemainingAndNoAnyUnmapped.INSTANCE.map( "STANDARD" ) )
+            .isNull();
+
+        assertThat( ErroneousOrderMapperUsingNoAnyRemainingAndNoAnyUnmapped.INSTANCE.map( "RETAIL" ) )
+            .isEqualTo( OrderType.RETAIL );
     }
 
 }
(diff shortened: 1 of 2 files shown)


processor/src/test/java/org/mapstruct/ap/test/value/string2enum/StringToEnumMappingTest.java
```diff
@@ -53,7 +54,20 @@ public void testRemainingAndNull() {
                     "type String to an enum type." )
         }
     )
-    public void shouldRaiseErrorWhenUsingAnyRemaining() {
+    public void shouldRaiseWarningWhenNotUsingAnyRemainingOrAnyUnmapped() {
```
sjaakd, line 57: Perhaps it is my misunderstanding.. Lets see if my assumption is correct:

* MapStruct will continue to do name based mapping for the remaining properties based on tartget
* And then still raise a warning (that cannot be suppressed)?

Or.. is this the default behavior and what should the user do to force MapStruct to continue with name-based mappings based on target without issuing a warning?

filiphr: Yes your assumptions are correct, this is what happens with this PR. Suppressing the warning could be done if we implement https://github.com/mapstruct/mapstruct/issues/2339.

Before, we generated a warning and didn't do name based mapping. This in my opinion is not correct and is misleading, since a warning should not stop the name based mapping.

sjaakd: The warning possibly stems from the enum-to-enum mapping, where it totally makes sense and is in line with the normal `@Mapping` (when you miss targets). Here however, it feels a bit awkward. Perhaps we should advance #2339 as part of 1.5 to fully finish this one.

sjaakd: But than it has to be something like: `@ValueMapping(target = MappingConstants.THROW_EXCEPTION, source =  MappingConstants.ANY_UNMAPPED)` (all unmapped) and and `@ValueMapping(target = MappingConstants.THROW_EXCEPTION, source =  MappingConstants.ANY_UNY_REMAINING)` (all non-name-based unmapped)

filiphr: I agree, and I also like #2339, so it will be part of 1.5 for sure :). Btw the `THROW_EXCEPTION` doesn't have to be only for the `ANY_UNMAPPED` or `ANY_REMAINING`, you can also define it for a specific mapping as well (although we need to see how the inverse works there (most likely it should be ignored).

sjaakd (approved):

filiphr:
Thanks for the review @sjaakd

PR #2355, opened 2021-02-06
#2347 Do not generate mapper implementation for private mappers

### LINKED ISSUE #2347: Compilation failure when mapper interface is a private nested interface
When a private nested interface is declared as a `@Mapper`, three compilation warnings are emitted in the generated class.  It's not completely communicated to the user _why_ this error is happening (though the private warnings are a big hint), nor have I encountered documentation stating that this usage should not be allowed.

It feels like this scenario should be handled better — either by allowing it; or by disallowing it, documenting it, and making the error messaging indicate the restriction against private nested interface mappers.

Note that changing the interface to non-private resolves the compilation issue, so a workaround is to have it be the default package private visibility.  That said, if the mapper is only used internally to a single class, it is natural to want to make it private in this manner.  

## Class:

```java
public class ClassWithPrivateMapper {
    public TargetType doSomething(SourceType source) {
        return Mappers.getMapper(PrivateMapper.class).map(source);
    }

    @Mapper
    private interface PrivateMapper {
        TargetType map(SourceType source);
    }
}
```

## Output:

```
/path/to/ClassWithPrivateMapper$PrivateMapperImpl.java:11: error: modifier private not allowed here
private class ClassWithPrivateMapper$PrivateMapperImpl implements PrivateMapper {
        ^
/path/to/ClassWithPrivateMapper$PrivateMapperImpl.java:3: error: PrivateMapper has private access in ClassWithPrivateMapper
import com.example.ClassWithPrivateMapper.PrivateMapper;
                                         ^
/path/to/ClassWithPrivateMapper$PrivateMapperImpl.java:11: error: PrivateMapper has private access in ClassWithPrivateMapper
private class ClassWithPrivateMapper$PrivateMapperImpl implements PrivateMapper {
                                                                  ^
3 errors
```

filiphr:
Provide a compiler error message instead of generating code that will not compile

Fixes #2347

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
@@ -15,6 +15,7 @@ import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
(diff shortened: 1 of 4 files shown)


sjaakd (approved):
LGTM

filiphr:
Thanks for the review @sjaakd

PR #2362, opened 2021-02-19
#2339 Enum mapping option for runtime exception

JudeNiroshan:
### Feature PR 🚀

Fixes: #2339 

### Changes

Introduced a new feature that can bypass the compilation error for certain Enum values. 

```
    @ValueMappings({
        @ValueMapping(source = "RETAIL", target = "RETAIL"),
        @ValueMapping(source = "EXTRA", target = MappingConstants.THROW_EXCEPTION)
    })
    ExternalOrderType orderTypeToExternalOrderTypeWithException(OrderType orderType);
```

This will generate the following code:

```
    @Override
    public ExternalOrderType orderTypeToExternalOrderTypeWithException(OrderType orderType) {
        if ( orderType == null ) {
            return null;
        }

        ExternalOrderType externalOrderType;

        switch ( orderType ) {
            case RETAIL: externalOrderType = ExternalOrderType.RETAIL;
            break;
            case EXTRA: throw new IllegalArgumentException( "Unexpected enum constant: " + orderType );
            default: throw new IllegalArgumentException( "Unexpected enum constant: " + orderType );
        }

        return externalOrderType;
    }
```

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
@@ -44,6 +44,7 @@ public class ValueMappingMethod extends MappingMethod {
     private final List<MappingEntry> valueMappings;
     private final String defaultTarget;
     private final String nullTarget;
+    private boolean nullAsException;
     private boolean defaultAsException;
 
     private final Type unexpectedValueMappingException;
@@ -121,6 +122,7 @@ public class ValueMappingMethod extends MappingMethod {
             return new ValueMappingMethod( method,
                 mappingEntries,
                 valueMappings.nullValueTarget,
+                valueMappings.hasNullValueAsException,
                 valueMappings.defaultTargetValue,
                 determineUnexpectedValueMappingException(),
                 beforeMappingMethods,
@@ -316,7 +318,7 @@ public class ValueMappingMethod extends MappingMethod {
 
             for ( ValueMappingOptions mappedConstant : valueMappings.regularValueMappings ) {
 
-                if ( THROW_EXCEPTION.equals( mappedConstant.getSource() ) ) {
+                if ( !enumMapping.isInverse() && THROW_EXCEPTION.equals( mappedConstant.getSource() ) ) {
                     ctx.getMessager().printMessage(
                         method.getExecutable(),
                         mappedConstant.getMirror(),
@@ -432,7 +434,8 @@ public class ValueMappingMethod extends MappingMethod {
 
         private Type determineUnexpectedValueMappingException() {
             boolean noDefaultValueForSwitchCase = !valueMappings.hasDefaultValue;
-            if ( noDefaultValueForSwitchCase || valueMappings.hasAtLeastOneExceptionValue ) {
+            if ( noDefaultValueForSwitchCase || valueMappings.hasAtLeastOneExceptionValue
+                || valueMappings.hasNullValueAsException ) {
                 TypeMirror unexpectedValueMappingException = enumMapping.getUnexpectedValueMappingException();
                 if ( unexpectedValueMappingException != null ) {
                     return ctx.getTypeFactory().getType( unexpectedValueMappingException );
@@ -490,7 +493,7 @@ public class ValueMappingMethod extends MappingMethod {
         boolean hasMapAnyUnmapped = false;
         boolean hasMapAnyRemaining = false;
         boolean hasDefaultValue = false;
-        boolean hasNullValue = false;
+        boolean hasNullValueAsException = false;
         boolean hasAtLeastOneExceptionValue = false;
 
         ValueMappings(List<ValueMappingOptions> valueMappings) {
@@ -511,7 +514,9 @@ public class ValueMappingMethod extends MappingMethod {
                 else if ( NULL.equals( valueMapping.getSource() ) ) {
                     nullTarget = valueMapping;
                     nullValueTarget = getValue( nullTarget );
-                    hasNullValue = true;
+                    if ( THROW_EXCEPTION.equals( nullValueTarget ) ) {
+                        hasNullValueAsException = true;
+                    }
                 }
                 else {
                     regularValueMappings.add( valueMapping );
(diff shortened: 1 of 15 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
```diff
@@ -361,6 +372,7 @@ private boolean reportErrorIfMappedTargetEnumConstantsDontExist(Method method, T
 
             for ( ValueMappingOptions mappedConstant : valueMappings.regularValueMappings ) {
                 if ( !NULL.equals( mappedConstant.getTarget() )
+                    && !THROW_EXCEPTION.equals( mappedConstant.getTarget() )
```
filiphr, line 375: I think that we need the same check on line 389 as well. It should be possible to map `ANY_UNMAPPED` or `ANY_REMAINING` into `<THROW_EXCEPTION>`. I know that the second one doesn't make sense as that is the default from MapStruct anyways :)

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -175,7 +175,9 @@
     VALUEMAPPING_ANY_REMAINING_FOR_NON_ENUM( "Source = \"<ANY_REMAINING>\" can only be used on targets of type enum and not for %s." ),
     VALUEMAPPING_ANY_REMAINING_OR_UNMAPPED_MISSING( "Source = \"<ANY_REMAINING>\" or \"<ANY_UNMAPPED>\" is advisable for mapping of type String to an enum type.", Diagnostic.Kind.WARNING  ),
     VALUEMAPPING_NON_EXISTING_CONSTANT_FROM_SPI( "Constant %s doesn't exist in enum type %s. Constant was returned from EnumMappingStrategy: %s"),
-    VALUEMAPPING_NON_EXISTING_CONSTANT( "Constant %s doesn't exist in enum type %s." );
+    VALUEMAPPING_NON_EXISTING_CONSTANT( "Constant %s doesn't exist in enum type %s." ),
+    VALUEMAPPING_THROW_EXCEPTION_SOURCE(
+        "Source = \"<THROW_EXCEPTION>\" is not allowed. Target = \"<THROW_EXCEPTION>\" can only be used." );
```
filiphr, line 180: Let's remove the new line here and leave the message on one line. No need to reformat here.

processor/src/main/resources/org/mapstruct/ap/internal/model/ValueMappingMethod.ftl
```diff
@@ -22,8 +22,8 @@
 
     switch ( ${sourceParameter.name} ) {
     <#list valueMappings as valueMapping>
-        case <@writeSource source=valueMapping.source/>: ${resultName} = <@writeTarget target=valueMapping.target/>;
-        break;
+        case <@writeSource source=valueMapping.source/>: <#if unexpectedValueMappingException?? && valueMapping.target?? && valueMapping.target="<THROW_EXCEPTION>">throw new <@includeModel object=unexpectedValueMappingException />( "Unexpected enum constant: " + ${sourceParameter.name} );<#else>${resultName} = <@writeTarget target=valueMapping.target/>;
```
filiphr, line 25: I think that this is not entirely correct. If someone uses `ANY_UNMAPPED` or `ANY_REMAINING` then `unexpectedValueMappingException` will be `null` which means that when someone explicitly wants to throw an exception for a mapping, but have a default for the rest it wouldn't work. 

What I would suggest is to enhance the `MappingEntry` to know whether an exception needs to be thrown or not and check that.

In addition to that I would change `defaultTarget` and `nullTarget` to use `MappingEntry` as well and then you will repeat the same logic as for the `valueMappings`.

processor/src/test/java/org/mapstruct/ap/test/value/enum2enum/SpecialOrderMapper.java
```diff
@@ -29,14 +29,24 @@
...
+        @ValueMapping(source = "STANDARD", target = MappingConstants.NULL),
+        @ValueMapping(source = MappingConstants.ANY_REMAINING, target = "SPECIAL")
     })
     ExternalOrderType orderTypeToExternalOrderType(OrderType orderType);
 
+    @Named("orderTypeToExternalOrderTypeWithException")
+    @ValueMappings({
+        @ValueMapping(source = "RETAIL", target = "RETAIL"),
+        @ValueMapping(source = "B2B", target = "B2B"),
+        @ValueMapping(source = "STANDARD", target = "SPECIAL"),
```
filiphr, line 42: These 3 are obsolete mappings I guess.

filiphr (changes requested):
This is a really good contribution @JudeNiroshan 🥳 . It is complete and there are nice tests.

I've left some comments, mostly cosmetic and one bigger conceptual problem in relation to using default values.

I would suggest adding a new mapper similar to the `DefaultOrderMapper`, `OrderMapper` and `SpecialOrderMapper` that will use the `MappingConstants#THROW_EXCEPTION`. In addition to that let's add a new test class `EnumToEnumThrowExceptionMappingTest`, lately we try to make our test mappers more isolated to have only specific logic to the bug / feature we are testing.

I'd say we need to add some more test variations as well:

* Add `DefaultOrderThrowExceptionMapper` and replace the target with throw exception
* Add `OrderThrowExceptionMapper` - This will verify that throw exception is ignored in the inverse configuration
* Add `SpecialThrowExceptionMapper` - This will test the different variations with any remaining and `NULL`. We should allow `@ValueMapping( source = MappingConstants.NULL, target = MappingConstants.THROW_EXCEPTION)`

I think that when you write these tests you'll see the errors I mentioned with the default mapping.

Thanks again for this great contribution, few more things to iron out and it will be ready for merging 😄 .

JudeNiroshan:
@filiphr Thanks for your detailed review. I'm currently addressing the logical problem you spotted. But whenever I try to run test cases, I get this error. 

`java.lang.AssertionError: No reference file could be found for Mapper org.mapstruct.ap.test.value.enum2enum.DefaultOrderMapper. You should create a file fixtures/org/mapstruct/ap/test/value/enum2enum/DefaultOrderMapperImpl.java`

According to this error, I should have the DefaultOrderMapperImpl inside the fixtures. In my setup, I do have the file but I still get this error. Do you have any idea why is that?

filiphr:
You'll need to maven resources once, so that the file is moved to the target folder. We have a small bug in our test harness there

filiphr:
@JudeNiroshan, let me know once you are done so I can do another pass on this PR

JudeNiroshan:
Hi @filiphr, I cannot understand the workflow for inversing(`@InheritInverseConfiguration`) of a mapping. 

For e.g: 

`@ValueMapping(source = "NORMAL", target = MappingConstants.THROW_EXCEPTION)`
when this inversed, it becomes invalid to have the `source = MappingConstants.THROW_EXCEPTION`. So, I tried to ignore it as follows: 
`if ( !enumMapping.isInverse() && THROW_EXCEPTION.equals( mappedConstant.getSource() ) ) {`

But it seems even in the process of inverse `enumMapping.isInverse()` will not become `true`. It is always `false`.

https://github.com/mapstruct/mapstruct/pull/2362/files?file-filters%5B%5D=.java#diff-37569bab982cd8f05076fbe281464cf2cc233d28e1ca3d93468585a73e48b439R319

Any help would be appreciated.

filiphr:
The inverse is done in the [`ValueMappingOptions#inverse`](https://github.com/mapstruct/mapstruct/blob/85af901ea7d7bc6cbdc52952211df76fc290f426/processor/src/main/java/org/mapstruct/ap/internal/model/source/ValueMappingOptions.java#L113-L127).

JudeNiroshan:
@filiphr you can review again now. 😁

filiphr:
Thanks for going through this again @JudeNiroshan. I will try to review it (and hopefully merge it 😄) this week.

JudeNiroshan:
@filiphr kind reminder 😸

filiphr:
Thanks for the reminder @JudeNiroshan. I thought that I finished the review and merged this. I'll check my local branches

filiphr:
I've manually merged this PR in https://github.com/mapstruct/mapstruct/commit/c4135e68ed8330fbc5180abd828e35b30f606ece and did some small polishing in https://github.com/mapstruct/mapstruct/commit/197dd4327afe1f9cf55a10579491aa3fe162fb0b

PR #2363, opened 2021-02-21
#2356 Implicitly ignore reverse inherited mappings that do not have read and write methods

### LINKED ISSUE #2356: Passing of mapping method argument to a property and using @InheritInverseConfiguration
```java
import org.mapstruct.*;

@Mapper
public interface CarMapper {
    class Car {
        public String brand;
        public String model;
        public String modelInternational;
    }
    class CarDTO {
        public String brand;
        public String modelName;
    }

    @Mapping(target = "modelName", source = "in")
    CarDTO map(Car in);

    default String mapToModel(Car in){
        return in.modelInternational==null? in.model : in.modelInternational;
    }

    @InheritInverseConfiguration
    @Mapping(target = "model", source = "modelName")
    @Mapping(target = "modelInternational", ignore = true)
    Car map(CarDTO in);
}
```

This code is compiling fine in 1.3.1.FINAL but in 1.4.2.FINAL gives following compile error: 
`Unknown property "in" in result type CarMapper.Car. Did you mean "brand"?`

Is that change of behavior on purpose?

filiphr:
Fixes #2356

This worked on 1.3 because we always ignored invalid target references after reversing them.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2356/Issue2356Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2356/Issue2356Mapper.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2356;
+
+import org.mapstruct.InheritInverseConfiguration;
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper
+public interface Issue2356Mapper {
+
+    Issue2356Mapper INSTANCE = Mappers.getMapper( Issue2356Mapper.class );
+
+    class Car {
+        //CHECKSTYLE:OFF
+        public String brand;
+        public String model;
+        public String modelInternational;
+        //CHECKSTYLE:ON
+    }
+
+    class CarDTO {
+        //CHECKSTYLE:OFF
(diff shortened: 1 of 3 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_2356/Issue2356Mapper.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    class CarDTO {
+        //CHECKSTYLE:OFF
+        public String brand;
+        public String modelName;
+        //CHECKSTYLE:ON
+    }
+
+    @Mapping(target = "modelName", source = "in")
```
sjaakd, line 36: Perhaps we have too many scenarios.. But why are we considering a parameter mapping at all for reversing? Or do I miss some cases? Or.. is your solution (no target write accessor) a more generic one?

filiphr: Just by looking at the mapping we can't know that it is a parameter mapping. This is being done [here](https://github.com/mapstruct/mapstruct/blob/85af901ea7d7bc6cbdc52952211df76fc290f426/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java#L194) and our `MappingOptions` have no knowledge of the types, only what is in the annotation.

Therefore, we need to do the check at the time we are doing here. Pre 1.4 it was a bit different and we had the target reference constructed earlier, this is no longer the case since we can't know about the constructors and the likes.

sjaakd: Ok.. go ahead. (add a comment perhaps).

sjaakd (approved):

Issue #2368, opened 2021-03-01
Make target the first parameter in all the uses of @Mapping


jvwilge:
There are some inconsistencies  in the documentation (md, asciidoc, javadoc) and the unit tests with the order of the parameters (`target` is not always before `source`) in the `@Mapping` annotations.

I asked a question about this : https://groups.google.com/d/msgid/mapstruct-users/977d0938-0781-4531-83c2-5c5832a88109n%40googlegroups.com?utm_medium=email&utm_source=footer

I already started on fixing all the docs, will start on the unit tests and create a PR afterwards.

filiphr:
Thanks for aligning everything @jvwilge. Closing this issue since we merged your PR (#2369)

PR #2369, opened 2021-03-01
#2368 Fix order of target param in mapping

jvwilge:
See https://groups.google.com/g/mapstruct-users/c/dM_ZGPJ_JlU/m/K8Gh0gybBQAJ

I fixed it in all the documentation and unit tests. Something to keep in mind when reviewing the code is that I blindly copy/pasted the target param everywhere and removed some spaces.


filiphr:
Thanks a lot for this @jvwilge. Everything seems good to me and our test suite is green 😄

PR #2380, opened 2021-03-12
2329 Optimization: Caching DateTimeFormatter instances

eforest:
* HelperMethod
  extended to supply additional template parameters to
  be more flexible in freemarker templates

* SupportingMappingMethod
** provide templateParameters to freemarker
** hashCode/equals base on name property instead of template name,
   as we use one specific template multiple times with different
   parameters generating multiple methods

* AbstractJavaTimeToStringConversion
  provides GetDateTimeFormatter as HelperMethod
  using cached DateTimeFormatter instance provided by
  GetDateTimeFormatter

* GetDateTimeFormatter
** a HelperMethod that creates cached DateTimeFormatters
   for given dateFormat or defaultFormatterSuffix
** method name and cache class name are created dynamically using
   given dateFormat creating a safe but unique name
** will be reused for multiple conversion fields with same
   dateFormat

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractJavaTimeToStringConversion.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/AbstractJavaTimeToStringConversion.java
@@ -6,8 +6,11 @@
 package org.mapstruct.ap.internal.conversion;
 
 import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
 
+import org.mapstruct.ap.internal.model.HelperMethod;
 import org.mapstruct.ap.internal.model.common.ConversionContext;
 import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.util.Collections;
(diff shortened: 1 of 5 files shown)


filiphr:
Thanks for the PR @eforest. 

I saw the code that we generate and I would like to avoid creating such a code by MapStruct. It is really difficult to read that code and we try to create code that is more readable.

I would suggest that we keep using the default way of accessing the formatter when using the known formats such as:

* `DateTimeFormatter.ISO_DATE_TIME`
* `DateTimeFormatter.ISO_LOCAL_DATE_TIME`
* `DateTimeFormatter.ISO_LOCAL_TIME`
* `DateTimeFormatter.ISO_LOCAL_DATE`

For the custom formatters I would suggest doing something similar like we are doing for the built in methods. Have a way to register built in fields or something like that that would allow us to register those fields, either as `static final` or just `final` fields in the mapper instance itself.

This code would be way more readable I think. As a first step we can even name those fields as `dateTimeFormatter` and add suffixes for that. We have a mechanism for this already I think.

I would also suggest looking into `AbstractToXmlGregorianCalendar` where we already use similar mechanism for caching `DataTypeFactory#newInstance` (done in #1532).

PR #2386, opened 2021-03-20
2329 Optimization: Caching DateTimeFormatter instances

eforest:
* HelperMethod
** extended to supply additional template parameters to
   be more flexible in freemarker templates
** extended to support mapper field reference / constructor fragment

* SupportingMappingMethod
** provide templateParameters to freemarker
** hashCode/equals base on name property instead of template name,
   as we use one specific template multiple times with different
   parameters generating multiple methods

* SupportingField / SupportingConstructorFragment
  added support for HelperFieldReference / HelperConstructorFragment

* AbstractJavaTimeToStringConversion
  provides GetDateTimeFormatter as HelperMethod
  using DateTimeFormatter instances provided as maper fields
  by GetDateTimeFormatter

* GetDateTimeFormatter
** a HelperMethod that creates DateTimeFormatters
   for given dateFormat as mapper fields
** variableName is created using given dateFormat

* MappingResolverImpl
  HelperMethods needs to know about existing fields

* HelperFinalField
  copy of FinalField (which could not be used due to package
restrictions)


filiphr:
I see that you pushed another commit @eforest. Is this ready for review now?

eforest:
Hi @filiphr, 
branch 2329-2 now contains refactorings regarding BuiltInMethod / HelperMethod, to dedup code where possible, but still uses pseudo helper method. 

--------

> I also saw that you added Pseudo-HelperMethod to be able to create the field. I would suggest to maybe expose another getter in the ConversionProvider that will return the helper fields.

Tried out this in branch 2329-3. Also works fine but requires some extensions to MappingResolverImpl / MapperCreationProcessor to carry on the additional fields to mapper template processing.
Actually with this approach I could also remove most changes to HelperMethod/BuiltInMethod/SupportingMappingMethod, but it's more symmetric now (so that decision is up to you).

Both could be reviewed.

Issue #2399, opened 2021-04-01
Support DS (Declarative Services) as Component Model 


ktor:
Hi Folks,

this issue is about using the OSGI-Style `@org.osgi.service.component.annotations.Component` and `@org.osgi.service.component.annotations.Reference` annotations.

I've got a working prototype (see attached merge request) but it is missing tests. I you are experienced with writing unit tests that could check if OSGi references are filled in correctly in similar fashion there are tests for Spring Component Model please join me!

- [x] Is this an issue (and hence not a question)?

dlaprova:
Hi ktor,

I am professionally involved in projects that rely on OSGI and Declarative Service as their technological stack.

I have some academic experience in writing JUnit and Mockito tests. I have never tried to mock OSGI components, but I am willing to contribute to find a way to implement them.

Can i be of some help? Do you suggest any starting point?

ktor:
Hi @dlaprova , I'm happy to hear from you!

Yes, you can certainly help! Please check the pull request: https://github.com/mapstruct/mapstruct/pull/2400. What we need is to write working implementation everywhere where there is **TODO osgi** comment, like here: https://github.com/mapstruct/mapstruct/pull/2400/files#diff-54dfa14bd76fc710acc684c990077482d0fc184f6b440d4e5b05b1d8d22666ccR56 It should spin up a real or mock osgi environment, activate generated osgi services according to generated annotations and test is just to check if service has been initialized- it'll confirm that annotations were generated correctly. Please let me know if you'll need more intro. We can arrange a call / screensharing and discuss the setup.

Torkin1:
Hi @ktor 

I'm trying to reach out @cschneider myself as well for assistance.

I'll report the mail here, which summarizes my documenting state on the matter so far:

> Dear Chris,
>
> My name is Daniele and I'm a junior professional involved in the development of modules targeting the OSGi framework.
>
> I'm currently try to collaborate to the MapStruct open source project, aiming at adding a OSGi-compatible implementation to it. In order to be merged, such implementation must be checked against automated tests, i.g. JUnit tests. The objective here is to test if a reference to Mapper instance as a OSGi service component is correctly instatiated.
>
> I found a recording of a [talk](https://www.youtube.com/watch?v=9WGbnfa2BaM) of yours where you showcast different methods to test OSGi applications. From my understanding, it seems that the library Pax Exams fits my requirements pretty well. I've also found this repository of yours with a MWE of a test suite.
>
[...]

PR #2400, opened 2021-04-01
#2399 Add OSGi Declarative Services as Componen Model

ktor:
It is work in progress. I've created the pull request to collaborate on the code. It is missing tests but otherwise generates OSGi DS annotations correctly, at least for my use case.


processor/src/main/java/org/mapstruct/ap/internal/processor/DsComponentProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * A {@link ModelElementProcessor} which converts the given {@link Mapper}
+ * object into a OSGi DS Component in case "ds" is configured as the
+ * target component model for this mapper.
+ */
+public class DsComponentProcessor extends AnnotationBasedComponentModelProcessor {
+    @Override
+    protected String getComponentModelIdentifier() {
+        return "ds";
```
filiphr, line 23: shall we call this `osgi-ds` perhaps?

processor/src/main/java/org/mapstruct/ap/internal/processor/DsComponentProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    protected List<Annotation> getTypeAnnotations(Mapper mapper) {
+        if ( mapper.getDecorator() == null ) {
+            return Arrays.asList( component() );
+        }
+        else {
+            return Arrays.asList( component() );
+        }
```
filiphr, line 33: There is no difference here between having a decorator and not having one. Are you sure that this would work correctly in the OSGi world?

filiphr (changes requested):
Thanks for the PR @ktor. I have extremely limited knowledge with OSGi, so I would trust your judgement here.

I've left some comments, which I think need addressing.

>I have a trouble creating OSGi container in the test so it could check if references are properly binded as it is done for Spring Component Model implementation. Maybe it is too much trouble anyways and there is a better / simpler way to test if it is correctly generated. Please advise.

Before merging this we should have tests for this. Otherwise, we won't be able to maintain this. I don't know what kind of troubles you had. I guess it is something with our custom test harness and our custom class loaders. I would suggest checking `CompilingStatement`, most likely you'll need to add the reference to the jar in the same way it is done for `spring-beans`.

processor/src/main/java/org/mapstruct/ap/internal/processor/DsComponentProcessor.java
ktor: done

processor/src/main/java/org/mapstruct/ap/internal/processor/DsComponentProcessor.java
ktor: done

ktor:
Thank you for code review @filiphr. I've made the changes you've proposed and added simple test to confirm that generated mappers contain required annotations.

So far I had no luck with integration test where junit would invoke simple osgi container that would confirm that components are initialized and bind correctly but the DS annotations are pretty simple.

I believe the simple test could suffice is maintained: would only change once OSGi DS specification changes (very rare).

ktor:
Changed to draft because after changes in `OsgiDsComponentProcessor` my DS supplied factory that is references in `uses` attribute of `@Mapper` annotation does not get needed `@Reference` annotation. I will update the test for the case.

filiphr:
>I believe the simple test could suffice if maintained: would only change once OSGi DS specification changes (very rare).

Can you please show me an example test case that I can look into, to see why it isn't working in our test setup. We do have some complex setup of course 😄

ktor:
> Can you please show me an example test case that I can look into

Hi @filiphr, I've finished the "simple" tests, meaning I've copied all the tests from spring decorator and injection strategy to osgi_ds packages and adjusted the "simple" tests to checkup on expected, generated OSGi DS result.

All places that I've failed to test are now commented with _TODO osgi_. Those are methods like `osgiUp`, `osgiDown`. Those methods should be responsible for spinning up a local osgi container in junit test runtime or to run some kind of osgi standardized test suite. I've failed to use one that works in the mapstruct setup, tried following:
1. https://bnd.bndtools.org/chapters/315-launchpad-testing.html and failed to provide "remote workspace" for it to use
2. https://sling.apache.org/documentation/development/osgi-mock.html failed to apply completely, it was some time ago and don't remeber exactly what was not working, maybe I'll give it another chance

ktor:
Hello @cschneider,

Thank you for your presentation: https://www.youtube.com/watch?v=9WGbnfa2BaM. I write to you in hope that you could help me to implement OSGi DS test for services generated with mapstruct mapper generator.

I would be very grateful if you could take a look at one particular test: https://github.com/mapstruct/mapstruct/pull/2400/files#diff-54dfa14bd76fc710acc684c990077482d0fc184f6b440d4e5b05b1d8d22666ccR56 and suggest a way to test if OSGi DS annotations generated with the mapstruct tooling would work in osgi container supporting OSGi R7.

Classes generated with `mvn clean test` that should somehow be included in the test and activated with the osgi runtime / test suite are prepared in the directory `processor/target/compilation-tests/org.mapstruct.ap.test.decorator.osgi_ds.constructor.OsgiDsDecoratorTest/{{name_of_a_test_method}}_eclipse/classes`

Thank you!

PR #2097, opened 2020-05-14
adding custom "nested presence checkers" support #674 and #2084

ivangsa:
adding custom "nested presence checkers" support #674 and #2084

Issue #1216, opened 2017-05-30
Ambiguous mapping method but there is a most specific mapping


filiphr:
I have a util class that has the following methods:

```java
public static java.util.Date toUtilDate(String strDate);
public static java.sql.TimeStamp toTimeStamp(String strDate);
public static java.sql.TimeStamp toTimeStamp2(String strDate);
public static java.sql.Date toSqlDate(String strDate);
```

I want to map from `String` to `java.util.Date`. I am getting an Ambiguous mapping error. I think that if an explicit mapping between `String` to `java.util.Date` exists then we should select that method. If the `toUtilDate` method didn't exists then we should throw the error.

WDYT?

andresmontoyat:
H!, 

I have the following problem, this is my example: 

this is a model object:
```java
 public class CurriculumViewModel extends CurriculumModel
 ...
```

this is my mapper methods: 
```java
public CurriculumModel toCurriculumModel(Curriculum domain);

@Mappings({
            @Mapping(source = "id", target = "id"),
            @Mapping(source = "headLine", target = "headLine"),
            @Mapping(source = "summary", target = "summary"),
            @Mapping(source = "location", target = "location")
 })
public CurriculumViewModel toCurriculumViewModel(Curriculum domain);
```
this is my error message:
```
Ambiguous mapping methods found for mapping collection element to
```

filiphr:
As discussed in the gitter [chat](https://gitter.im/mapstruct/mapstruct-users). We need some more information for the error in order to help you out. You might need to specify on which method exactly does the error occur and in which mapper.

csxiaoshang:
> H!,
> 
> I have the following problem, this is my example:
> 
> this is a model object:
> 
> ```java
>  public class CurriculumViewModel extends CurriculumModel
>  ...
> ```
> 
> this is my mapper methods:
> 
> ```java
> public CurriculumModel toCurriculumModel(Curriculum domain);
> 
> @Mappings({
>             @Mapping(source = "id", target = "id"),
>             @Mapping(source = "headLine", target = "headLine"),
>             @Mapping(source = "summary", target = "summary"),
>             @Mapping(source = "location", target = "location")
>  })
> public CurriculumViewModel toCurriculumViewModel(Curriculum domain);
> ```
> 
> this is my error message:
> 
> ```
> Ambiguous mapping methods found for mapping collection element to
> ```
hi, do you slove this error? i found the same error.

PR #2403, opened 2021-04-04
Conditional mapping

### LINKED ISSUE #2051: [Feature] Conditional mapping

### LINKED ISSUE #2084: Support custom source presence checker expressions for mappings and custom method mappings
Currently mapstruct only allows presence checker methods on the owning source object.

Some types (like java.util.Options, guava Optional, JsonNullable, ...) have presence checker method on the property itself.

With a custom presence checker expression configuration we could write a mapper like:

```
@Mapper

    @SourcePresenceCheckerConfig(
            expression = "java( <SOURCE>.<READ_ACCESSOR>() != null && <SOURCE>.<READ_ACCESSOR>().isPresent() )")
    default <T> T fromOptional(final Optional<T> optional) {
        return optional.orElse(null);
    }

    // nulls are handled inside method
    @SourcePresenceCheckerConfig(nullValueCheckStrategy = NullValueCheckStrategy.NEVER)
    default <T> Optional<T> asOptional(final T optional) {
        return optional != null ? Optional.of(optional) : Optional.empty();
    }

}
```

or a mapping like:
```
@Mapping(target = "prop1", source ="prop1", sourcePresenceCheckerExpression = "java( source != null && source.getProp1().isPresent() )")
TargetWithPlainValues asPlain(SourceWithOptional source);
```

This may address this question:
https://stackoverflow.com/questions/61371019/custom-source-presence-checking-method-name-in-mapstruct

and this issue in a general manner:
https://github.com/mapstruct/mapstruct/issues/674

filiphr:
This is my iteration of the conditional mapping.

There are 2 commits, one which refactors the presence checks into an interface and using Freemarker templates, and the second one that adds the functionality.

There is a new annotation `@Condition` which has to be on a method in order for such a method to be considered as a custom presence check method.

There are new attributes in `@Mapping`:

* `conditionBy` - similar to `qualifiedBy` but only for `@Condition` methods
* `conditionByName` - similar to `qualifiedByName` but only for `@Condition` methods
* `conditionExpression` - which allows to define the expression in Java format.

This PR uses the same concept for discovering custom presence check methods through the `MethodSelectors`. 

Fixes #2051
Fixes #2084

This PR should make PR #2148 and PR #2097 obsolete.

@ivangsa can you please check if this satisfies your requirement? The approach here is slightly different than the one proposed by you in PR #2097. Check `ConditionalMappingTest#optionalLikeConditional` and `OptionalLikeConditionalMapper`.

@Desislav-Petrov can you also have a look at this? This makes it possible to use the source parameters, `@Context` parameters and the source right hand side.

@sjaakd can you also please have a look at this. Check the 2 commits separate, since they are entirely independent, and the first one is there to make it simpler for the conditional mapping.

Todos before merge:

- [x] Add documentation section


processor/src/main/java/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.java
```diff
@@ -52,17 +54,27 @@ public Builder mappingContext(MappingBuilderContext mappingContext) {
...
+                    sourceParameter = parameter;
+                }
             }
             final List<Type> thrownTypes = new ArrayList<>();
             List<SafePropertyEntry> safePropertyEntries = new ArrayList<>();
+            if ( sourceParameter == null ) {
+                throw new IllegalStateException( "Method " + method + " has no source parameter." );
+            }
+
+            String previousPropertyName = sourceParameter.getName();
```
sjaakd, line 70: is it `previousPropertyName` or `originaSourceParameterName`

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -604,30 +608,43 @@ else if ( !sourceReference.isNested() ) {
...
+                    ) );
 
                     String variableName = sourceParam.getName() + "."
                         + propertyEntry.getReadAccessor().getSimpleName() + "()";
                     for (int i = 1; i < sourceReference.getPropertyEntries().size(); i++) {
                         PropertyEntry entry = sourceReference.getPropertyEntries().get( i );
                         if (entry.getPresenceChecker() != null && entry.getReadAccessor() != null) {
-                            sourcePresenceChecker += " && " + variableName + " != null && "
-                                + variableName + "." + entry.getPresenceChecker().getSimpleName() + "()";
+                            presenceChecks.add( new NullPresenceCheck( variableName ) );
```
sjaakd, line 665: I see that we add a `NullPresenceCheck` and a `SourceReferenceMethodPresenceCheck` .. Why not assume delegation of the null check to a default `SourceReferenceMethodPresenceCheck`? Wouldn't that give
1. More room to play for the implementor of `SourceReferenceMethodPresenceCheck`? (null is a value, right)
2. Give more readable code?

Or perhaps I don't understand it sufficient yet at this point in time (first review pass).

processor/src/main/java/org/mapstruct/ap/internal/model/NestedPropertyMappingMethod.java
filiphr: It is actually both. The idea is to replace

```
<#macro localVarName index><#if index == 0>${sourceParameter.name}<#else>${propertyEntries[index-1].name}</#if></#macro>

File: processor/src/main/java/org/mapstruct/ap/internal/model/presence/AllPresenceChecksPresenceCheck.java [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Set;
+
+import org.mapstruct.ap.internal.model.common.ModelElement;
+import org.mapstruct.ap.internal.model.common.PresenceCheck;
+import org.mapstruct.ap.internal.model.common.Type;
+
+/**
+ * @author Filip Hrisafov
+ */
+public class AllPresenceChecksPresenceCheck extends ModelElement implements PresenceCheck {
```
sjaakd, line 20: From this I assume you want to concatenate multiple presence checks.. But as mentioned earlier, there could be a dependency: a presence checker not being executed while a previous one already fired.

I'm not sure if we need that.. E.g.: can you skip the null presence check?  Or couldn't that one be the default to override (instead of add a check after that). ?

sjaakd: How much is this all steered by the strategy enums on method-class-mapper level?

sjaakd: Should we deprecate the strategy enums.. now we have more fine grained control.. ? E.g. should this be 2.0 functionality?

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
filiphr: This is a special one which is only did for #1456. My idea was to make the presence checks independent of each other. The `SourceReferenceMethodPresenceCheck` is also used in the case when we need to call `hasProperty()` on a source bean. In this case we shouldn't do a `null` check for the `source` since it has already been done at the beginning of the method. We want to avoid the following scenario


```

    if ( source == null ) {
        return null;
    }
    
    Target target = new Target();
    
    if ( source != null && source.hasName() ) {
        target.setName( source.getName() );
    }

}
```

processor/src/main/java/org/mapstruct/ap/internal/model/presence/AllPresenceChecksPresenceCheck.java
filiphr: I don't quite follow with what you mean by "a presence checked not being executed while a previous one already fired". This `AllPresenceChecksPresenceCheck` is meant mostly to cover the use case in #1456.

>How much is this all steered by the strategy enums on method-class-mapper level?

This is entirely steared by the strategy enums. All the behaviour stays the same, basically `NullValueCheckStrategy` can be though as `PresenceValueCheckStrategy`. 

>Should we deprecate the strategy enums.. now we have more fine grained control.. ? E.g. should this be 2.0 functionality?

I don't think that we should deprecate them. I think that it should behave the same way as it did before. With this we are giving more options for the users to write presence checks without changing their beans.

sjaakd: > I don't quite follow with what you mean by "a presence checked not being executed while a previous one already fired". This `AllPresenceChecksPresenceCheck` is meant mostly to cover the use case in #1456.

I suspected that the following code is generated:

```java
if ( source != null && pc1( source ) && pc2( source) {
        target.setName( source.getName() );
 }

```

The value `null` will never be offered to the `pc1`. So `pc1` cannot make choices anymore on `null`.. but reading your answer that does not happen right? The same is valid for `pc1` and `pc2` by the way, but at least that is something that the user has under control.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: > The `SourceReferenceMethodPresenceCheck` is also used in the case when we need to call `hasProperty()` on a source bean

So, the `hasXYZ` is one of the 'default' presence checkers? Is this overridden as well as you define a `conditionBy`? (so if you've defined `has` methods.. what will happen in relation to the new stuff)?

processor/src/main/java/org/mapstruct/ap/internal/model/presence/AllPresenceChecksPresenceCheck.java
sjaakd: > This is entirely steared by the strategy enums. All the behaviour stays the same, basically `NullValueCheckStrategy` can be though as `PresenceValueCheckStrategy`.

What about `NullValuePropertyMappingStrategy`..? 

The point is that this interface is getting too complex. You need to define a strategy and then you need to do something in your `@Mapping` to customise this.

I would plea for a solution that we 
1. make behaviour consistent:.  `NullValuePropertyMappingStrategy' and 'NullValueCheckStrategy' 
2. drop the 2 `NullValuePropertyMappingStrategy' and 'NullValueCheckStrategy' 
3. extend `@Condition` to be used on `@MapperConfig` and `@Mapper`
4. have some default `@Conditions` for reuse. Like the `null check` and the `has` check.
5. have a script to apply to source code migrate the current two strategies. I'm thinking about the way that assertj was done from its predecessor. 
6. Rethink the remaining strategy `NullValueMappingStrategy` and at least call it something different: `SourceBeanNullCheck` or apply a special case of the `@Condition` to this scenario as well.

Don't get me wrong.. I really like the approach.  But I think the API is becoming too complex and too dependent. 

So perhaps we should call the next release 2.0 alpha or something like that.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
filiphr: Yes `hasXYZ` is one of the 'default' presence checkers. If you define a custom `@Condition` method that is applicable for that type then, this will be used, i.e. your presence checker on the bean won't be used. Perhaps, those should have highest precedence instead?

processor/src/main/java/org/mapstruct/ap/internal/model/presence/AllPresenceChecksPresenceCheck.java
filiphr: I see what you are saying. In this situation we are not looking for custom presence checkers here, this will work exactly like it did before. The code you think that is generated with `pc1` and `pc2` it won't be generated.

I am not following your plea, perhaps we need to do a sync meeting about it 😉

sjaakd (approved):
You are right about collections I spotted a time ago that since we support automatic mapping we don't need in-situ collection mapping anymore. Perhaps also something we could drop for 2.0 to make it more consistent.

About de strategies.. I still think this way of working could at least offer a viable approach to deprecate the `nullvaluecheck` strategy. After all, you could provide a more generic option in your code:

```java
@Condition
boolean ispresent( Object o ) {
   return o != null;
}
```

Thinking whether we should do this already for 1.5 and mention how to do this (in order to prepare users for 2.0)?

filiphr:
>About de strategies.. I still think this way of working could at least offer a viable approach to deprecate the nullvaluecheck strategy. After all, you could provide a more generic option in your code:

You are right the `NullValueCheckStrategy` indeed has no impact once you add `@Condition`. 

We need to investigate this, but I think such a condition will lead to an "ambiguous error" like I have created in https://github.com/mapstruct/mapstruct/issues/1216

filiphr:
Thanks for the review btw. I have rebased this on top of the main branch and got rid of the used `NullValueCheckStrategy`

PR #2406, opened 2021-04-05
#2303 Generated code should use iteration order preserving LinkedHash(Map|Set) instead of Hash(Map|Set)

### LINKED ISSUE #2303: Retain iteration order for Sorted(Map|Set) values
We experienced a strange bug in our software recently, that took us a while to find. The insight we finally ended up with was:
The iteration order of a specific `Map` value caused the issue. Please consider following example:

```java
class SortedMapDto {
  public SortedMap<String, String> theMap = new TreeMap<>();
}

class TargetDto {
  public Map<String, String> theMap;
}

@Mapper
interface SortedMapMapper {
  TargetDto map(SortedMapDto sortedMapDto);
}
```
The code MapStruct (1.4.1.Final) generates from this is:
```java
// ...
SortedMap<String, String> sortedMap = sortedMapDto.theMap;
if ( sortedMap != null ) {
  targetDto.theMap = new HashMap<String, String>( sortedMap );
}
```

Now as any reasonable Java developer knows, the iteration order of `Map` is not guaranteed. So this bug was of course *our fault* and what MapStruct does is technically correct.

[Immutables](https://immutables.github.io/) though, has a bit more *resilient* take on that by always using `LinkedHashMap` for creating defensive copies of any `Map` types to retain their iteration order.

What I want to suggest is considering to use `LinkedHashMap` (and its counterpart `LinkedHashSet`) just like Immutables for defensive copies instead of `HashMap`.

What's your take on that?

filiphr:
Fixes #2303

### DIFF
--- a/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
@@ -180,7 +180,7 @@ By default the target property will be set to null.
 However:
 
 1. By specifying `nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT` on `@Mapping`, `@BeanMapping`, `@Mapper` or `@MapperConfig`, the mapping result can be altered to return *default* values.
-For `List` MapStruct generates an `ArrayList`, for `Map` a `HashMap`, for arrays an empty array, for `String` `""` and for primitive / boxed types a representation of `false` or `0`.
+For `List` MapStruct generates an `ArrayList`, for `Map` a `LinkedHashMap`, for arrays an empty array, for `String` `""` and for primitive / boxed types a representation of `false` or `0`.
 For all other objects an new instance is created. Please note that a default constructor is required. If not available, use the `@Mapping#defaultValue`.
 
 2. By specifying `nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.IGNORE` on `@Mapping`, `@BeanMapping`, `@Mapper` or `@MapperConfig`, the mapping result will be equal to the original value of the `@MappingTarget` annotated target.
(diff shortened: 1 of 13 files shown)


sjaakd (commented):
Is it safe to do this? I read on SO that 'LinkedMap' is (after a number of entries) actually faster. But is it always available on lightweight implementations? E.g. Android? I did not check thread safeness yet (I guess it will be the same). 

Anyway.. I'm a bit hesitant if we don't have a (configurable) fallback option. E.g. a `MapperConfig#mapImplementation` or something similar. I would not mind if the default is `LinkedHashSet`.. but just in case.

filiphr:
>Is it safe to do this? I read on SO that 'LinkedMap' is (after a number of entries) actually faster. But is it always available on lightweight implementations? E.g. Android? I did not check thread safeness yet (I guess it will be the same).

I didn't check whether the linked map / set are their on lightweight implementations. However, I would imagine that such a basic map / set should be 😄. Nevertheless, I checked and there is [`LinkedHashMap`](https://developer.android.com/reference/java/util/LinkedHashMap) and [`LinkedHashSet`](https://developer.android.com/reference/java/util/LinkedHashSet) on Android as well.

As for thread safeness, both the linked and non linked variants are the same.

```
Anyway.. I'm a bit hesitant if we don't have a (configurable) fallback option. E.g. a MapperConfig#mapImplementation or something similar. I would not mind if the default is LinkedHashSet.. but just in case.

**Comment by filiphr** [comment]
>Anyway.. I'm a bit hesitant if we don't have a (configurable) fallback option. E.g. a `MapperConfig#mapImplementation` or something similar. I would not mind if the default is `LinkedHashSet`.. but just in case.

I don't like the `MapperConfig` approach. However, people can easily do this with an `@ObjectFactory`.

e.g.

```
public interface MapCreator {

    @ObjectFactory
    static <K, V> Map<K, V> createMap() {
        return new HashMap<>();
    }

    @ObjectFactory
    static <T> Set<T> createSet() {
        return new HashSet<>();
    }
}
```

and then use it in `Mapper#uses` or `MapperConfig#uses`

sjaakd:
> I don't like the `MapperConfig` approach. However, people can easily do this with an `@ObjectFactory`.

Ok. perhaps good to mention that in the release message.

sjaakd (approved):

filiphr:
Thanks for the review @sjaakd. Will add that to the release notes.

PR #2419, opened 2021-04-16
unmappedTargetPolicy for @BeanMapping

### LINKED ISSUE #2132: unmappedTargetPolicy for @BeanMapping
In case you want to have a different unmapped-target-policy just for one mapping right now you have to go and move it to a dedicated class setting the `@Mapper` config accordingly.

I think in the past there was a similar issue, but I can't find a solution to it... I expected that `@BeanMapping` has a `unmappedTargetPolicy` property to change it just for this mapping.

Reason to have this:
Sometimes you map to objects you don't control and they've got a lot properties. To prevent defining a lot ignore `@Mapping`s and having troubles upgrading the external dependency that might introduce new targets it could make sense to decide that the unmapped-target-policy should be set to `IGNORE` - but just for this mapping and not for the others.

This doesn't sound like a big deal to add this.. but I would like to get some feedback from you guys @filiphr and @sjaakd if you also think this is helpful before I will start on it.
Maybe there is already a good way to do this so that we don't need this new option at all.

LukeLaz:
Fixes #2132

### DIFF
--- a/core/src/main/java/org/mapstruct/BeanMapping.java
+++ b/core/src/main/java/org/mapstruct/BeanMapping.java
@@ -141,6 +141,15 @@ public @interface BeanMapping {
     String[] ignoreUnmappedSourceProperties() default {};
 
     /**
+     * How unmapped properties of the target type of a mapping should be
+     * reported. The method overrides an unmappedTargetPolicy set by {@link Mapper#unmappedSourcePolicy()} or
+     * {@link MapperConfig#unmappedTargetPolicy() }
+     *
+     * @return The reporting policy for unmapped target properties.
+     */
+    ReportingPolicy unmappedTargetPolicy() default ReportingPolicy.WARN;
+
+    /**
      * The information that should be used for the builder mappings. This can be used to define custom build methods
      * for the builder strategy that one uses.
      *
--- a/core/src/main/java/org/mapstruct/BeanMapping.java
+++ b/core/src/main/java/org/mapstruct/BeanMapping.java
@@ -141,9 +141,9 @@ public @interface BeanMapping {
     String[] ignoreUnmappedSourceProperties() default {};
 
     /**
-     * How unmapped properties of the target type of a mapping should be
-     * reported. The method overrides an unmappedTargetPolicy set by {@link Mapper#unmappedSourcePolicy()} or
-     * {@link MapperConfig#unmappedTargetPolicy() }
+     * How unmapped properties of the target type of a mapping should be reported.
+     * If no policy is configured, the policy given via {@link MapperConfig#unmappedTargetPolicy()} or
+     * {@link Mapper#unmappedTargetPolicy()} will be applied, using {@link ReportingPolicy#WARN} by default.
      *
      * @return The reporting policy for unmapped target properties.
      */
--- a/documentation/src/main/asciidoc/chapter-2-set-up.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-2-set-up.asciidoc
@@ -245,6 +245,7 @@ Supported values are:
 * `IGNORE`: unmapped target properties are ignored
 
 If a policy is given for a specific mapper via `@Mapper#unmappedTargetPolicy()`, the value from the annotation takes precedence.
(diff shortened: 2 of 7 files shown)


chris922:
@LukeLaz : Looks great to me!

Maybe add at least one more test case where you use an `unmappedTargetPolicy` of `ERROR` on class level, but reduce it on method level to `WARN` or something like this? This would verify that overriding the policy will works.

LukeLaz:
Good point @chris922, added one

core/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -140,6 +140,15 @@ NullValuePropertyMappingStrategy nullValuePropertyMappingStrategy()
      */
     String[] ignoreUnmappedSourceProperties() default {};
 
+    /**
+     * How unmapped properties of the target type of a mapping should be
+     * reported. The method overrides an unmappedTargetPolicy set by {@link Mapper#unmappedSourcePolicy()} or
+     * {@link MapperConfig#unmappedTargetPolicy() }
```
mjustin, line 146: More general question: this "override" language matches the language in `Mapper.java`, but not the other overrides in this class.  For example:

```java
     * The strategy to be applied when {@code null} is passed as source bean argument value to this bean mapping. If no
     * strategy is configured, the strategy given via {@link MapperConfig#nullValueMappingStrategy()} or
     * {@link Mapper#nullValueMappingStrategy()} will be applied, using {@link NullValueMappingStrategy#RETURN_NULL}
     * by default.
```

and

```java
     * The strategy to be applied when a source bean property is {@code null} or not present. If no strategy is
     * configured, the strategy given via {@link MapperConfig#nullValuePropertyMappingStrategy()} or
     * {@link Mapper#nullValuePropertyMappingStrategy()} will be applied,
     * {@link NullValuePropertyMappingStrategy#SET_TO_NULL} will be used by default.
```

I do think think I prefer the approach you're using here, but it is inconsistent.  Should the other Javadocs in this class be updated to use this same linguistic approach?

LukeLaz: What is your opinion @chris922?

filiphr: I also think that the new wording is clear. It gives the clear intent of this. I think it makes it also clear that if you set it (even if to the default value), it will override what was in `@Mapper` or `@MapperConfig`.

WDYT @sjaakd?

sjaakd: If I interpret @LukeLaz his comment correctly it comes down to 2 things:
1. don't use the `override` term because its meaning in OO languages (like in `Overrides`). 
2. keep phrasing aligned over all places were we have this situation.
I tend to agree with that. 

However, if something is unclear - like explicitly setting the default value has a different effect than leaving it out altogether - we need to add that in other places.

mjustin: @sjaakd My comment was specifically regarding #2.  Namely, there currently exists different phrasing between `BeanMapping.java` and `Mapper.java`, but this PR is using the `Mapper.java` phrasing within `BeanMapping.java`.  I suspect the phrasings between the two can be made similar.  If not, it's probably better to keep the same phrasing within the same class.

That said, I'm a user of the library, not a contributor, so that probably limits the value of my opinion somewhat.

LukeLaz: Hmmm, on further inspection it seems like phrasing is not aligned already:

```
     * Determines when to include a null check on the source property value of a bean mapping.
     *
     * Can be overridden by the one on {@link MapperConfig}, {@link Mapper}  or {@link Mapping}.
     *
     * @return strategy how to do null checking
     * NullValueCheckStrategy nullValueCheckStrategy() default ON_IMPLICIT_CONVERSION;
```

filiphr:
Thanks for the great PR @LukeLaz. I manually integrated it in https://github.com/mapstruct/mapstruct/commit/cc1562c5adacd09afcfd547b3bc340be95c953bd. It was manual because we changed the way we test so needed to change `@Test` with our new `@ProcessorTest`.

PR #2424, opened 2021-04-24
#2423 Use SetterWrapperForCollectionsAndMapsWithNullCheck if the source has a presence check method

### LINKED ISSUE #2423: Source presence checks on collections with non direct mappings is ignored
When we have the following model:

```
public class SoccerTeam {

    private Collection<Player> players;

    public boolean hasPlayers() {
        return players != null && !players.isEmpty();
    }

    // getters and setters omitted

}

public class Player {

    private String name;

    // getters and setters omitted
}

public class SoccerTeamDto {

    private Collection<PlayerDto> players;

    // getters and setters omitted
}

public class PlayerDto {

    private String name;

    // getters and setters omitted
}
```

Then the following mapper will not use the `hasPlayers` presence check.

```
@Mapper
public interface SoccerTeamMapper {

    SoccerTeamDto map(SoccerTeam team);

}
```

Presence check methods should always be used irregardless of the `NullValueCheckStrategy`.

filiphr:
Fixes #2423


sjaakd (approved):

sjaakd:
LGTM!

filiphr:
Thanks for the review @sjaakd

PR #2433, opened 2021-05-06
CaseEnumTransformationStrategy

jpbassinello:
A new enum transformation strategy that applies case conversion to the source enum.

LOWER:

`SourceEnum.VALUE1 -> TargetEnum.value1`

UPPER:

`SourceEnum.value1 -> TargetEnum.VALUE1`

CAPITAL:

`SourceEnum.ENUM_VALUE -> TargetEnum.Enum_Value`

Fixes: #2445

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/spi/CaseEnumTransformationStrategy.java
+++ b/processor/src/main/java/org/mapstruct/ap/spi/CaseEnumTransformationStrategy.java
@@ -1,0 +1,53 @@
+package org.mapstruct.ap.spi;
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+/**
+ * Applies case transformation to the source enum
+ *
+ * @author jpbassinello
+ *
+ * @since 1.5
+ */
+public class CaseEnumTransformationStrategy implements EnumTransformationStrategy {
+
+    private static final String UPPER = "upper";
+    private static final String LOWER = "lower";
+    private static final String CAPITAL = "capital";
+
+    @Override
+    public String getStrategyName() {
+        return "case";
+    }
+
+    @Override
+    public String transform(String value, String configuration) {
+        switch (configuration.toLowerCase()) {
+            case UPPER:
+                return value.toUpperCase();
+            case LOWER:
+                return value.toLowerCase();
+            case CAPITAL:
+                return capitalize(value);
+            default:
+                return value;
+        }
+    }
+
+    private static String capitalize(String value) {
+        return Arrays.stream(value.split("_"))
+                .map(CaseEnumTransformationStrategy::upperCaseFirst)
+                .collect(Collectors.joining("_"));
+    }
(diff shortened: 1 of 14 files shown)


processor/src/main/java/org/mapstruct/ap/spi/CaseEnumTransformationStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+        return "case";
+    }
+
+    @Override
+    public String transform(String value, String configuration) {
+        switch (configuration.toLowerCase()) {
+            case UPPER:
+                return value.toUpperCase();
+            case LOWER:
+                return value.toLowerCase();
```
filiphr, line 30: Using `Locale.ROOT` when doing `toLowerCase` and `toUpperCase`

processor/src/main/java/org/mapstruct/ap/spi/CaseEnumTransformationStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+            case CAPITAL:
+                return capitalize(value);
+            default:
+                return value;
```
filiphr, line 34: I think that we need to throw an exception if the user has used something that we do not support.

filiphr (changes requested):
Thanks for the PR @jpbassinello.

On first pass it looks great, I've left two comments that we should address before merging this.

On top of that I think that there are some copyright headers missings. You'll see it in the CI failing.

Final note, can we add thew new strategy to the [Custom Name Transformation](https://mapstruct.org/documentation/stable/reference/html/#_custom_name_transformation) section of the docs?

processor/src/main/java/org/mapstruct/ap/spi/CaseEnumTransformationStrategy.java
jpbassinello: make sense to me. Done.

processor/src/main/java/org/mapstruct/ap/spi/CaseEnumTransformationStrategy.java
jpbassinello: Fixed.

jpbassinello:
Thanks @filiphr 
I think I fixed the issues. I also updated the documentation, but I´ve added here https://mapstruct.org/documentation/stable/reference/html/#_custom_name_transformation
instead of the section you suggested. Let me know if it is correct for you.

filiphr:
Thanks for the PR @jpbassinello. I integrated it manually in https://github.com/mapstruct/mapstruct/commit/fdf3dcc8efaf0cdc1a2145dd53c1302346e2838a and I also improved the error message when the configuration is incorrect in https://github.com/mapstruct/mapstruct/commit/4576103752df896d4548c1d618ab892bf540b628

PR #2447, opened 2021-05-15
#2446 Use DefaultLocale and DefaultTimeZone from JUnit Pioneer

### LINKED ISSUE #2446: Use JUnit Pioneer for setting Default Locale and Default Timezone
JUnit Pioneer has [Default Locale and Default TimeZone](https://junit-pioneer.org/docs/default-locale-timezone/) annotations that can be used to set the default Locale / TimeZone for a particular test.

Those annotations work correctly with JUnit Jupiter `@ResourceLock` and running tests in Parallel.

Doing this will achieve better performance when running tests locally. When running the `processor` tests in the default way it takes 8 minutes for the `mvn test` in the processor module. If we use parallel tests it takes 4 minutes.

### LINKED ISSUE #2466: Integration Tests with Java 16

filiphr:
Configure the JUnit Platform to run the processor tests in parallel by running different test classes in concurrent threads

Fixes #2446

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -29,6 +29,7 @@
         <org.eclipse.tycho.compiler-jdt.version>1.6.0</org.eclipse.tycho.compiler-jdt.version>
         <com.puppycrawl.tools.checkstyle.version>8.36.1</com.puppycrawl.tools.checkstyle.version>
         <org.junit.jupiter.version>5.8.0-M1</org.junit.jupiter.version>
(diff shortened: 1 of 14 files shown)


filiphr:
>Now.. that's pretty cool. @work i'm still at junit4. Just today fixed an issue with timezone problems..

It is a pretty cool testing framework. And time to upgrade to JUnit Jupiter, we've done that at my work. Also starting slower, few tests at a time 😉

PR #2465, opened 2021-06-05
#2375 Add records cross module integration test

### LINKED ISSUE #2375: Track Progress of JDK-8258535
Java currently has a bug in regard to how the compiler provides record information from other jars.

See [JDK-8258535](https://bugs.openjdk.java.net/browse/JDK-8258535).

Currently there is a problem on JDK 15 and JDK16.ea. 

Related to #2294 and #2374.

filiphr:
Fixes #2375

### DIFF
--- a/.github/workflows/java-ea.yml
+++ b/.github/workflows/java-ea.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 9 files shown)


sjaakd (approved):

filiphr:
Thanks for the review @sjaakd

PR #2471, opened 2021-06-06
Mapping from map to bean

### LINKED ISSUE #1075: Please add a feature to map from java.util.map to an object

filiphr:
This supersedes #2317. @ckosmowski did an amazing job, I only polished it a bit and fixed the merge conflicts.

@ckosmowski let me know if you don't agree with some of my changes.

Fixes #1075

Open points:

- [x] Add documentation
- [x] Add tests for using non String map as a source parameter mapped into a bean

### DIFF
--- a/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
@@ -650,3 +650,68 @@ public class PersonMapperImpl implements PersonMapper {
 }
 ----
 ====
+
+[[mapping-map-to-bean]]
+=== Mapping Map to Bean
+
+There are situations when a mapping from a `Map<String, ???>` into a specific bean is needed.
+MapStruct offers a transparent way of doing such a mapping by using the target bean properties (or defined through `Mapping#source`) to extract the values from the map.
+Such a mapping looks like:
+
+.Example classes for mapping map to bean
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+public class Customer {
+
+    private Long id;
+    private String name;
+
+    //getters and setter omitted for brevity
+}
+
+@Mapper
+public interface CustomerMapper {
+
+    @Mapping(target = "name", source = "customerName")
+    Customer toCustomer(Map<String, String> map);
+
+}
+----
+====
+
+.Generated mapper for mapping map to bean
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+// GENERATED CODE
+public class CustomerMapperImpl implements CustomerMapper {
+
+    @Override
+    public Customer toCustomer(Map<String, String> map) {
+        // ...
+        if ( map.containsKey( "id" ) ) {
+            customer.setId( Integer.parseInt( map.get( "id" ) ) );
+        }
+        if ( map.containsKey( "customerName" ) ) {
+            customer.setName( source.get( "customerName" ) );
+        }
+        // ...
+    }
+}
+----
+====
+
+[NOTE]
+====
+All existing rules about mapping between different types and using other mappers defined with `Mapper#uses` or custom methods in the mappers are applied.
+i.e. You can map from `Map<String, Integer>` where for each property a conversion from `Integer` into the respective property will be needed.
+====
+
+[WARNING]
+====
+When a raw map or a map that does not have a String as a key is used, then a warning will be generated.
+The warning is not generated if the map itself is mapped into some other target property directly as is.
+====
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
@@ -181,7 +181,11 @@ public enum Message {
     VALUEMAPPING_ANY_REMAINING_OR_UNMAPPED_MISSING( "Source = \"<ANY_REMAINING>\" or \"<ANY_UNMAPPED>\" is advisable for mapping of type String to an enum type.", Diagnostic.Kind.WARNING  ),
     VALUEMAPPING_NON_EXISTING_CONSTANT_FROM_SPI( "Constant %s doesn't exist in enum type %s. Constant was returned from EnumMappingStrategy: %s"),
     VALUEMAPPING_NON_EXISTING_CONSTANT( "Constant %s doesn't exist in enum type %s." ),
-    VALUEMAPPING_THROW_EXCEPTION_SOURCE( "Source = \"<THROW_EXCEPTION>\" is not allowed. Target = \"<THROW_EXCEPTION>\" can only be used." );
+    VALUEMAPPING_THROW_EXCEPTION_SOURCE( "Source = \"<THROW_EXCEPTION>\" is not allowed. Target = \"<THROW_EXCEPTION>\" can only be used." ),
+
+    MAPTOBEANMAPPING_WRONG_KEY_TYPE( "The Map parameter \"%s\" cannot be used for property mapping. It must be typed with Map<String, ???> but it was typed with %s.", Diagnostic.Kind.WARNING ),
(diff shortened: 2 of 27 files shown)


filiphr:
One think that I forgot to add is the warning about raw map or a map that does not have `String` as a key.

Is there something in backwards compatibility way that we need to think about?

ckosmowski:
Thanks for bringing this feature further!

From what i understand about the polishing, it looks great! So i agree 👍 .
One thing that is very important for us about this feature is that we really need to input untyped maps (trusting the user that it will have only Strings as keys) because of the outputs we get from a third party library.

But if i understand it correct, in this case the message will only be a warning and does not stop the mapper from working, am i correct?

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -179,7 +179,11 @@
     VALUEMAPPING_ANY_REMAINING_OR_UNMAPPED_MISSING( "Source = \"<ANY_REMAINING>\" or \"<ANY_UNMAPPED>\" is advisable for mapping of type String to an enum type.", Diagnostic.Kind.WARNING  ),
     VALUEMAPPING_NON_EXISTING_CONSTANT_FROM_SPI( "Constant %s doesn't exist in enum type %s. Constant was returned from EnumMappingStrategy: %s"),
     VALUEMAPPING_NON_EXISTING_CONSTANT( "Constant %s doesn't exist in enum type %s." ),
-    VALUEMAPPING_THROW_EXCEPTION_SOURCE( "Source = \"<THROW_EXCEPTION>\" is not allowed. Target = \"<THROW_EXCEPTION>\" can only be used." );
+    VALUEMAPPING_THROW_EXCEPTION_SOURCE( "Source = \"<THROW_EXCEPTION>\" is not allowed. Target = \"<THROW_EXCEPTION>\" can only be used." ),
+
+    MAPTOBEANMAPPING_WRONG_GENERIC_TYPES( "The Map parameter \"%s\" cannot be used for property mapping. It must be typed with Map<String, ???> but it was typed with %s.", Diagnostic.Kind.WARNING ),
```
sjaakd, line 184: '_WRONG_KEY_TYPE'?

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -179,7 +179,11 @@
...
+    MAPTOBEANMAPPING_UNTYPED( "The Map parameter \"%s\" cannot be used for property mapping. It must be typed with Map<String, ???> but it was untyped.", Diagnostic.Kind.WARNING ),
```
sjaakd, line 185: I guess this one is used only when you try to do a raw map. Shouldn't the message reflect that?. So something along the lines `Raw Map is not allowed`

sjaakd (approved):
Only a few minor's..

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
filiphr: Yes this is for raw maps. I'll adapt the warning message.

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
filiphr: Yes this is better. I'll adapt it

filiphr:
>I guess we do need to update the documentation as well.. right?

You are right @sjaakd. I will add some documentation to this as well.

>One thing that is very important for us about this feature is that we really need to input untyped maps (trusting the user that it will have only Strings as keys) because of the outputs we get from a third party library.
>
>But if i understand it correct, in this case the message will only be a warning and does not stop the mapper from working, am i correct?

@ckosmowski yes there will be a warning. However, the mapping will not work in this case because raw maps are skipped. Your implementation was doing that already. Check:

https://github.com/mapstruct/mapstruct/pull/2471/files#diff-0e91bfe179a1bac245627d8be8971044045f147c81b308a3c3d6f822a76b4804R329-R339

I added the assertions, but the was null anyways. Only a `Map<String, ???>` will be treated as mapping from Map to Bean. You will have to define your mappers that they receive `Map<String, ???>` and pass those untyped maps to them. Your mappers will be safe and you will trust that passing a raw map into a typed map is OK.

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
filiphr: Just saw that that message ends with "but it was untyped". I think that it is OK to do it like that. I'll change it tp "but it was raw"

filiphr:
@sjaakd, I've added some docs, added some more tests and also addressed your comments about the warning messages.

Please have a look and let me know what you think.

documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
```diff
@@ -650,3 +650,68 @@ public class PersonMapperImpl implements PersonMapper {
...
+[subs="verbatim,attributes"]
+----
+// GENERATED CODE
+public class CustomerMapperImpl implements CustomerMapper {
+
+    @Override
+    public Customer toCustomer(Map<String, String> map) {
+        // ...
+        if ( map.containsKey( "id" ) ) {
+            customer.setName( Integer.parseInt( map.get( "name" ) ) );
```
sjaakd, line 696: this doesn't look right I guess that you need to set the id and get the id.. 😄

documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
```diff
@@ -650,3 +650,68 @@ public class PersonMapperImpl implements PersonMapper {
...
+
+[NOTE]
+====
+All existing rules about mapping between different types and using other mappers defined with `Mapper#uses` or custom methods in the mappers are applied.
+i.e. You can map from `Map<String, Integer>` where for each property a conversion from `Integer` into the respective property will be needed.
+====
+
+[WARNING]
+====
+When a raw map or a map that does not have a String as a key is used, then a warning will be generated.
```
sjaakd, line 715: How is the key interpreted in such case.. guess we need to add that in the doc.

sjaakd (changes requested):
@filiphr . LGTM in general. Though the generated code doesn't look as it should be.

documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
filiphr: It isn't, I thought that it is clear at the beginning of the section

>There are situations when a mapping from a `Map<String, ???>` into a specific bean is needed.

documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
filiphr: Good catch. I've fixed this

ckosmowski:
Awesome that this change made it to the project! Thanks guys. I would love to be added to the copyright notice of the project like the other contributors if possible as a reference. Thanks in advance :-)

filiphr:
Thanks for letting us know @ckosmowski. You've been added to the copyright notice https://github.com/mapstruct/mapstruct/commit/0b2c7e58b264adb5674de51634509d70a4578da2. I did a pass of all the commits, but it seems like I missed you. Sorry about that

ckosmowski:
Thanks, nevermind!

Issue #2481, opened 2021-06-12
[suggestion] Add optional reporting if ignored source properties do not exist


incaseoftrouble:
For `unmappedSourcePolicy = ReportingPolicy.WARN` or above, one can use `@BeanMapping(ignoreUnmappedSourceProperties = {...})` to declare source properties which are omitted intentionally. However, an equally simple dual option is missing. I'd suggest having a `ignoredSourcePolicy` configuration which warns / fails if an unknown source property is missing. This is useful for keeping this list "in sync" with reality, especially helpful after refractoring.

I am aware that one could achieve through `ignoreByDefault = true`, but then you'd have to write all mappings explicitly.

Sure, compared to a required missing target property, an ignored missing source property is not really dramatic, but it helps with debugging and keeping code clean. Its definitely a low-priority suggestion.

filiphr:
I am not sure that I understand you completely @incaseoftrouble.

You want to have a warning / error if a source property used in `BeanMapping#ignoreUnmappedSourceProperties` does not exist? i.e. you made a typo?

If that is the case then I would say that we don't need any other option, we should do this out of the box automatically. I think that this is a relatively easy enhancement and anyone from the community can jump in to help with this. Perhaps you would be interested in providing this? We would need to change [this](https://github.com/mapstruct/mapstruct/blob/08016d9ef24a4fb59347953f1f0cba9bcecca810/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java#L253) line and add a check that the property is indeed in the `unprocessedSourceProperties` and if it isn't to have an error or a warning.

incaseoftrouble:
Done :-) I hope I got it right, since its a new feature I am not entirely sure how to test it. I basically copied what `unmappedSourcesPolicy` is doing, but I might have missed something.

filiphr:
Done in PR #2485, thanks @incaseoftrouble

PR #2485, opened 2021-06-13
#2481 Report ignored source properties which are missing

incaseoftrouble:
PR for #2481 - WIP

Optionally get warned / error if ignored source properties are missing (this way you can ensure an exact 1:1 correspondence between the declarations and generated code)

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -98,6 +98,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         private Map<String, Accessor> unprocessedConstructorProperties;
         private Map<String, Accessor> unprocessedTargetProperties;
         private Map<String, Accessor> unprocessedSourceProperties;
+        private Set<String> missingIgnoredSourceProperties;
         private Set<String> targetProperties;
         private final List<PropertyMapping> propertyMappings = new ArrayList<>();
         private final Set<Parameter> unprocessedSourceParameters = new HashSet<>();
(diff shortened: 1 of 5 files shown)


filiphr (changes requested):
Thanks a lot for the PR @incaseoftrouble. 

It is good. However, I think that we do not need to have a new option in the `@Mapper` or `@MapperConfig` for this. I think that we can immediately have a compile error if the user used an unknown source property in `BeanMapping#ignoreUnmappedSourceProperties`. The same error happens if you use `@Mapping(target = "unknown", ignore = true)`.

I know that this might lead to compile errors in existing code. However, in my opinion that is OK.

What do you think @sjaakd?

incaseoftrouble:
> I know that this might lead to compile errors in existing code. However, in my opinion that is OK.

Ok, that's for you guys to decide 👍 I'll wait for your comments and adapt accordingly.

filiphr:
Thanks for your understanding @incaseoftrouble.

An additional comment for why the tests are failing. We have integration tests that are taking all tests from our test suite and compile the (the so called fullFeatureTest). If a certain test has a compile error then the fullFeatureTest will fail.

We do have mappers that lead to compilation errors, as we want to test this. In order to satisfy the integration test such mappers have to contain `Erroneous` in their name. Classes with that in their name will be ignored.

Long story short, once we agree whether we need the new option or not (I am for not) we would need to change the mappers that test the compile errors to have `Erroneous` in their name

sjaakd:
> It is good. However, I think that we do not need to have a new option in the `@Mapper` or `@MapperConfig` for this. I think that we can immediately have a compile error if the user used an unknown source property in `BeanMapping#ignoreUnmappedSourceProperties`.

I don't use this feature a lot, so I'm kind of surprised (or rather shocked 😄 ) that you can specify properties that don't exist in `BeanMapping#ignoreUnmappedSourceProperties`.. So I went ahead and tried by altering a testcase:

```java
package org.mapstruct.ap.test.source.ignore;

import org.mapstruct.Mapper;
import org.mapstruct.ReportingPolicy;

@Mapper(
    unmappedSourcePolicy = ReportingPolicy.ERROR
)
public interface IgnoreUnmappedSourceMapper {

    @BeanMapping(
        ignoreUnmappedSourceProperties = {
            "blah", // <<<<<<<<<<<<<<<<< not existing
            "name",
            "surname"
        }
    )
    PersonDto map(Person person);
}
```

And indeed.. you don't get  an error. 

To answer your question on my opinion @filiphr :
* So, no, we don't need another option. 
* And yes, please generate an error in this case.

Fully agreed upon!

filiphr:
Thanks for looking into this as well @sjaakd. So @incaseoftrouble, please go ahead with what I said earlier, remove the option and automatically add a compile error if a defined source property in `BeanMapping#ignoreUnmappedSourceProperties` is unknown.

incaseoftrouble:
Updated accordingly. Locally, integration tests (`./mvnw install`) succeed.

filiphr:
Thanks for your work on this @incaseoftrouble

PR #2486, opened 2021-06-13
#2108 Make sure Javadoc can be generated with Java 11

### LINKED ISSUE #2108: Find an alternative for Apiviz and to run it on Java 11
Currently it is not possible to do the release with Java 11 because the apiviz doclet does not work with it.

>[ERROR] Exit code: 1 - javadoc: warning - The old Doclet and Taglet APIs in the packages
[ERROR] com.sun.javadoc, com.sun.tools.doclets and their implementations
[ERROR] are planned to be removed in a future JDK release. These
[ERROR] components have been superseded by the new APIs in jdk.javadoc.doclet.
[ERROR] Users are strongly recommended to migrate to the new APIs.

We should find an alternative to it. I had a look at [UMLDoclet](https://github.com/talsma-ict/umldoclet). However, it is not even close to the ones generated by API Viz. 

An alternative is to drop the diagrams creation at all. Not sure if it is worth it or not.

filiphr:
Remove the org.jboss.apiviz.APIviz doclet since it is no longer compatible with Java 11.
Add new group in the Javadoc for the MapStruct Processor SPI
Fix Javadoc warnings

Fixes #2108

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -34,6 +34,11 @@
             <groupId>org.freemarker</groupId>
             <artifactId>freemarker</artifactId>
         </dependency>
+        <!-- Needed here so references to MapStruct Gem classes can be resolved during JavaDoc generation -->
+        <dependency>
+            <groupId>org.mapstruct.tools.gem</groupId>
+            <artifactId>gem-api</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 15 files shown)


distribution/pom.xml
```diff
@@ -34,6 +34,11 @@
             <groupId>org.freemarker</groupId>
             <artifactId>freemarker</artifactId>
         </dependency>
+        <!-- Needed here so references to MapStruct Gem classes can be resolved during JavaDoc generation -->
+        <dependency>
```
sjaakd, line 38: This feels a bit weird.. Why would the javadoc processor require `gem-api`? Is there no way of telling the javadoc processor to exclude this explicitly from analysis.

sjaakd (commented):
LGTM (besides one question)

distribution/pom.xml
filiphr: That's how it works. It looks at our sources and tries to analyze it. The same problem as above with `freemarker`.

Issue #2463, opened 2021-06-04
ObjectFactory with generic source not in generated code in 1.5.0


markbanierink:
We have several classes extending abstract classes:
```java
abstract class Source {
    abstract public boolean isNew();
}

class SourceA extends Source {
    // some properties, etc.
}

class SourceB extends Source {
    // some properties, etc.
}

abstract class Target {
    // some properties, etc.
}

class TargetA extends Target {
    // some properties, etc.
}

class TargetB extends Target {
    // some properties, etc.
}
```
These are mapped with:
```java
@Mapping
TargetA toTarget(SourceA source);

@Mapping
TargetB toTarget(SourceB source);
```

Now, if we have an ObjectFactory using a generic target type:
```java
@ObjectFactory
public <T extends TargetClass> T createTarget(@TargetType Class<T> targetType) {
    return newTarget(targetType);
}
```
this generates:
```java
TargetA toEntity(SourceA source) {
    // some code...

    TargetA targetA = targetFactory.createTarget( TargetA.class );

    // some code...
}
```

However, if we add a second generic for the source argument:
```java
@ObjectFactory
public <T extends Target, S extends Source> T createTarget(S source, @TargetType Class<T> targetType) {
    if (source.isNew())) {
        return newTarget(source);
    }
    return getSomeOtherTarget(targetType, source);
}
```
the generated code looks like:
```java
TargetA toEntity(SourceA source) {
    // some code...

    TargetA targetA = new TargetA()

    // some code...
}
```
Here it should have generated something like:
```java
TargetA toEntity(SourceA source) {
    // some code...

    TargetA targetA = targetFactory.createTarget( source, TargetA.class );

    // some code...
}
```

This used to work in 1.4.2, however, in the current master/1.5.0 it doesn't seem to work anymore.

sjaakd:
Was introduced due to major refactoring of selection mechanism. Thanks for reporting. Please check (if possible) if this solves the issue.

markbanierink:
Yes, it's solved. Thanks!

PR #2487, opened 2021-06-14
#2463 Selection stops when method type-args are flipped

sjaakd:


filiphr (approved):
Looks great @sjaakd

sjaakd:
> Looks great @sjaakd

thanks :)

PR #2492, opened 2021-06-19
#2491 Do not use types not part of java.base in MapStruct processor

### LINKED ISSUE #2491: JaxbElem not used when jaxb-api not on annotation processor path on Java 11
This was reported on our [Gitter Channel](https://gitter.im/mapstruct/mapstruct-users)

>Mapstruct can not generate mapping implementation for `JaxbElement`-s when run with Java 11. I found this issue: mapstruct/mapstruct#1942 which says "upgrade to 1.3.1 solves the problem". But I use 1.4.1 ...
error looks like `error: Can't map property "JAXBElement<String> culture" to "String culture". Consider to declare/implement a mapping method: "String map(JAXBElement<String> value)"`.

Solution is

>seems I had to add JAXB dependency to `annotationProcessor` configuration

I've analyzed this and indeed it is a problem. The problem is that we are using classes from jaxb-api. However, we can work without those classes and everything should work transparently.

filiphr:
MapStruct should not use types that are outside of java.base.
This makes sure that no additional dependencies (such as jaxb-api) are needed on the annotation processor path

Fixes #2491


integrationtest/src/test/resources/jaxbAnnotationProcessorPathsTest/src/main/java/org/mapstruct/itest/jaxb/OrderDetailsDto.java
```diff
@@ -0,0 +1,42 @@
...
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.itest.jaxb;
+
+import java.util.List;
+
+/**
+ * @author Sjaak Derksen
```
sjaakd, line 11: are you sure I'm the author 😄

sjaakd (approved):

integrationtest/src/test/resources/jaxbAnnotationProcessorPathsTest/src/main/java/org/mapstruct/itest/jaxb/OrderDetailsDto.java
filiphr: yes you are, I copied everything from the `jaxbText`. In any case, I got a different idea for this. I'll add some more compilation options such that the current tests are run with `annotationProcessorPaths` and without them.

PR #2500, opened 2021-06-24
Update Named.java

anovitskyi:


filiphr:
Good catch @anovitskyi. Thanks

PR #2510, opened 2021-07-10
#2351 NullValueMappingStrategy for Maps and Iterables

### LINKED ISSUE #2351: NullValueMappingStrategy option to map collections/maps, but not beans
I have a mapper type that's mapping optional list elements from user input to required list elements internally.  MapStruct provides the option to set a `NullValueMappingStrategy`, but that also maps beans to null.  It would be helpful for my use case if `NullValueMappingStrategy` had the option to return default for collection-like objects (map/iterable), but not for beans.  It seems something generically useful as well; often code that null collection elements as equal to empty wouldn't want to do the same for beans.

My current workaround is to add a custom mapping type for lists to my mappers:

```java
    default <T> List<T> mapList(List<T> value) {
        return value == null ? Collections.emptyList() : value;
    }
```

## Proposed solution:
```java
public enum NullValueMappingStrategy {
    RETURN_NULL,
    RETURN_DEFAULT,
    RETURN_DEFAULT_COLLECTIONS,
    RETURN_DEFAULT_BEANS; // For completion's sake
}
```

hpoettker:
With two new parameters for Mapper and MapperConfig, it is now possible to override the nullValueMappingStrategy specifically for MapMapping and IterableMapping.

Fixes #2351

### DIFF
--- a/core/src/main/java/org/mapstruct/Mapper.java
+++ b/core/src/main/java/org/mapstruct/Mapper.java
@@ -204,6 +204,32 @@ public @interface Mapper {
     NullValueMappingStrategy nullValueMappingStrategy() default NullValueMappingStrategy.RETURN_NULL;
 
     /**
+     * The strategy to be applied when {@code null} is passed as source argument value to an {@link IterableMapping} of
+     * this mapper. If unset, the strategy set with {@link #nullValueMappingStrategy()} will be applied. If neither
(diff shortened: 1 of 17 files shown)


filiphr:
Thanks a lot for your PR @hpoettker it has been integrated.

Thanks for changing the ternary operators 😄

PR #2511, opened 2021-07-10
Remove remaining references to Hickory

hpoettker:


filiphr:
Good catch @hpoettker, thanks for the PR

Issue #2438, opened 2021-05-07
Support for optional subtype-only source properties


mjustin:
My source type is a superclass with a certain set of properties, with a subclass that adds additional properties.  My target type contains properties to be mapped from the superclass, as well as some optional properties to be mapped if the source is an instance of the subclass.  It would be convenient to be able to be able to express this mapping relationship in a single annotation-based mapping method, rather than using multiple mapping methods with a custom mapping method to tie them together.

This is certainly related to #131, and possibly involves the same solution.  The difference is that that issue concerns mapping to different target types based on the source type, whereas this involves mapping different target types to a single source type, with optional fields for subclasses.

### Types

```java
public class Source {
    private String property1;
    private String property2;
    private String property3;

    // Getters, setters, & constructors omitted for brevity
}

public class SourceSubclass extends Source {
    private String subclassProperty;

    // Getters, setters, & constructors omitted for brevity
}

public class Target {
    private String target1;
    private String target2;
    private String target3;
    private String target4;

    // Getters, setters, & constructors omitted for brevity
}
```

### Mapper
This is an solution to the problem using the features currently available in MapStruct.

```java
@Mapper
public interface SubclassMapper {
    default Target map(Source source) {
        if (source instanceof SourceSubclass) {
            return mapSubclass((SourceSubclass) source);
        } else {
            return mapSuperclass(source);
        }
    }

    @Mapping(target = "target1", source = "property1")
    @Mapping(target = "target2", source = "property2")
    @Mapping(target = "target3", source = "property3")
    @Mapping(target = "target4", expression = "java(null)")
    Target mapSuperclass(Source source);

    @InheritConfiguration(name = "mapSuperclass")
    @Mapping(target = "target4", source = "subclassProperty")
    Target mapSubclass(SourceSubclass source);
}
```

filiphr:
Thanks for the detailed explanation @mjustin.

I guess you are looking for MapStruct to generate your default `Target map(Source source)` source method for you? If that is the case then the solution is identical. The fact that #131 talks about different target types is only a small detail. If you agree, I would close this one in favour of #131. You can perhaps add your example to it as well.

Just so I understand you correctly. The `mapSuperclass` and `mapSubclass` will still be defined by you. You don't expect MapStruct to generate something like that for you?

mjustin:
@filiphr Sorry I'm just now getting back to your question.

[...]

Issue #366, opened 2014-12-02
Support for abstract class mapping or classes with base class


playbee:
For example (one abstract class and two concrete classes):

```
public abstract class Animal {
  private String name;

  ...
}

public class Cat extends Animal {
  private Boolean mouseCatcher;

  ...
}

public class Dog extends Animal {
  private Integer awardsCount;

  ...
}
```

DTO:

```
public abstract class AnimalDto {
  private String name;

  ...
}

public class CatDto extends AnimalDto {
  private Boolean mouseCatcher;

  ...
}

public class DogDto extends AnimalDto {
  private Integer awardsCount;

  ...
}
```

And mapper:

```
@Mapper
public interface AnimalMapper {
  public CatDto catToCatDto(Cat cat);
  public DogDto dogToDogDto(Dog dog);

  @AbstractMapping // special mark to generate abstract mapping method
  public AnimalDto animalToAnimalDto(Animal animal);

  public List<AnimalDto> animalsToAnimalDtos(List<Animal> animals);
}
```

Generated result:

```
public class AnimalMapperImpl implements AnimalMapper {
  public CatDto catToCatDto(Cat cat) {
    ...
  }

  public DogDto dogToDogDto(Dog dog) {
    ...
  }

  // example of generated method for abstract mapping
  public AnimalDto animalToAnimalDto(Animal animal) {
    if (animal instanceof Cat)
      return catToCatDto((Cat) animal);
    else if (animal instanceof Dog)
      return dogToDogDto((Dog) animal);
    return null;
  }

  public List<AnimalDto> animalsToAnimalDtos(List<Animal> animals) {
    if ( animals == null ) {
        return null;
    }

    List<AnimalDto> list = new ArrayList<AnimalDto>();

    for ( Animal animal : animals ) {
        list.add( animalToAnimalDto( animal ) );
    }

    return list;
  }
}
```

milanov:
Hi guys, is there any development on this issue (or the referenced #131 )? In case not, what would you recommend for an ad-hoc implementation of this looks like? The use case is exactly the one described here: converting between parallel class hierarchies of Entity/Dto objects:
```
  Person                 PersonDto
  /    \                  /    \
Man   Woman           ManDto   WomanDto
```
And the method I would like to implement is `List<? extends PersonDto> peopleToDtos(List<? extends Person> people)`?

amitjindal:
Hi, as @milanov requested could you please provide how to do these>?
[...]

Issue #131, opened 2014-02-16
Support for Type-Refinement mapping (or Downcast Mapping)


agudian:
Suppose we have the following types: `MotorizedVehicle`, `Car extends MotorizedVehicle`, `MotorCycle extends MotorziedVehicle`, and their complementary types `MotorizedVehicleDto`, `CarDto extends MotorizedVehicleDto`, `MotorCycleDto extends MotorziedVehicleDto`.

We also have defined a MotorCycleMapper, a CarMapper and a MotorizedVehicleMapper, each containing methods to map from the entity type to the DTO type.

What I now have in my hand is just a list of MotorizedVehicle instances and I want to map them all to DTOs. But not just to MotorizedVehicleDtos, but to what resembles the most concrete type. 

With the current state, I would have to "instanceof" my way through and call the right mapper with the downcasted object myself. Other mapping frameworks allow to define that the actually returned type may be something more special based on the runtime-type of the object to map (in dozer you have to specify such type pairs in the global config).

I wonder if we should support something like that as well. For exampe:

``` java
@Mapper( uses = { CarMapper.class, MotorCycleMapper.class } )
public MotorizedVehicleMapper {
    @Mapping( withDowncastMapping=true )
    public MotorizedVehicleDto vehicleToDto(MotorizedVehicle vehicle);
}
```

would generate to:

``` java
public MotorizedVehicleMapperImpl {
    private CarMapper carMapper = ...;

    @Override
    public MotorizedVehicleDto vehicleToDto(MotorizedVehicle vehicle) {
        if ( vehicle instanceof Car ) {
            return carMapper.carToCarDto( (Car) vehicle );
        }

        if ( vehicle instanceof MotorCycle ) {
            return ...;
        }

        MotorizedVehicleDto motorizedVehicleDto = new ...
        // the usual mapping stuff
    }
}
```

I know, usually one could nicely handle such stuff in an object-oriented way by upsmarting the entity types, but, well... you know how it is... ;-)

WDYT?

agudian:
The idea here is that we generate a method that selects the refining mapping method at runtime using instanceof checks and downcasts of the source parameter (as in the example above). The possible options are collected from the current mapper's methods and the used mappers.

That way, if a user has an object containing a property of type ```MotorizedVehicle```, then that property mapping can be done with the best downcasted fit automatically. We might need to enforce that the user specifies the behaviour in case none of the "known" options fits.

gunnarmorling:
[...]

PR #2512, opened 2021-07-11
PR for issues #131, #2438, #366

Zegveld:
#131 #2438 #366 : Added support for SubClassMapping annotations. This will allow for hierarchy mappings.

if a parent mapping method is annotated with @SubClassMapping it will now generate an instanceof check inside the parent mapping and generate the subclass mappings if they are not manually defined. See issues #131 #2438 and #366 for more information.

### DIFF
--- a/core/src/main/java/org/mapstruct/SubClassMapping.java
+++ b/core/src/main/java/org/mapstruct/SubClassMapping.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Configures the mapping to handle hierarchy of the source type.
+ * <p>
+ * The subclass to be mapped is to be specified via {@link #sourceClass()}.
+ * The subclass to map to is to be specified via {@link #targetClass()}.
+ * It currently is also required that a mapping method exist that has the {@link #targetClass()}
+ * as the return type and the {@link #sourceClass()} as the only argument.
+ * </p>
+ * <p>
+ * This annotation can be combined with &#64;Mapping annotations.
+ * </p>
+ *
+ * <pre><code class='java'>
+ * &#64;Mapper
+ * public interface MyMapper {
+ *    &#64;SubClassMapping (sourceClass = SourceSubClass.class,
+ *                      targetClass = TargetSubClass.class)
+ *    TargetParent mapParent(SourceParent parent);
+ *
+ *    TargetSubClass mapSubClass(SourceSubClass subInstant);
+ * }
+ * </code></pre>
+ * Below follow examples of the implementation for the mapParent method.
+ * <strong>Example 1:</strong> For parents that cannot be created. (e.g. abstract classes or interfaces)
+ * <pre><code class='java'>
+ * // generates
+ * &#64;Override
+ * public TargetParent mapParent(SourceParent parent) {
+ *     if (parent instanceof SourceSubClass) {
+ *         return mapSubClass( (SourceSubClass) parent );
+ *     }
+ *     else {
+ *         throw new IllegalArgumentException("Not all subclasses are supported for this mapping. Missing for "
+ *                    + parent.getClass());
+ *     }
+ * }
+ * </code></pre>
+ * <strong>Example 2:</strong> For parents that can be created. (e.g. normal classes or interfaces with
+ * &#64;Mappper( uses = ObjectFactory.class ) )
+ * <pre><code class='java'>
+ * // generates
+ * &#64;Override
+ * public TargetParent mapParent(SourceParent parent) {
+ *     TargetParent targetParent1;
+ *     if (parent instanceof SourceSubClass) {
+ *         targetParent1 = mapSubClass( (SourceSubClass) parent );
+ *     }
+ *     else {
+ *         targetParent1 = new TargetParent();
+ *         // ...
+ *     }
+ * }
+ * </code></pre>
+ *
+ * @author Ben Zegveld
+ * @since 1.5
+ */
+@Repeatable( value = SubClassMappings.class )
+@Retention( RetentionPolicy.CLASS )
+@Target( { ElementType.METHOD, ElementType.ANNOTATION_TYPE } )
+public @interface SubClassMapping {
+    /**
+     * @return the source subclass to check for before using the default mapping as fallback.
+     */
+    Class<?> sourceClass();
+
+    /**
+     * @return the target subclass to map the sourceSubClass to.
+     */
+    Class<?> targetClass();
+}
--- a/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
@@ -110,6 +110,34 @@ public interface SourceTargetMapper {
 
 The example demonstrates how to use defaultExpression to set an `ID` field if the source field is null, this could be used to take the existing `sourceId` from the source object if it is set, or create a new `Id` if it isn't. Please note that the fully qualified package name is specified because MapStruct does not take care of the import of the `UUID` class (unless it’s used otherwise explicitly in the `SourceTargetMapper`). This can be resolved by defining imports on the @Mapper annotation (see <<expressions>>).
 
+[[sub-class-mappings]]
+=== Subclass Mapping
+
+When both input and result types have an inheritance relation, you would want the correct specialization be mapped to the matching specialization. Suppose an Apple and a Banana, which are both specializations of Fruit.
+
+.Specifying the sub class mappings of a fruit mapping
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+@Mapper
+public interface FruitMapper {
+
+    @SubClassMapping( sourceClass = AppleDto.class, targetClass = Apple.class )
+    @SubClassMapping( sourceClass = BananaDto.class, targetClass = Banana.class )
+    Fruit map( FruitDto source );
+
+}
+----
+====
+
+If you would just use a normal mapping both the AppleDto and the BananaDto would be made into a Fruit object, instead of an Apple and a Banana object. By using the SubClassMapping an AppleDtoToApple mapping will be done for Apples, and an BananaDtoToBanana mapping will be done for Banana's. If you would try to map a GrapeDto it would still turn into a Fruit. In the case that the Fruit type is an abstract class or an interface, you would get an IllegalArgumentException because it is unknown how to map a GrapeDto. Adding the missing (`@SubClassMapping`) for it will fix that.
+
+[TIP]
+====
+If the mapping method for the subClasses does not exist it will be created and any other annotations on the fruit mapping method are inherited by the newly generated mappings.
+====
+
 [[determining-result-type]]
 === Determining the result type
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -21,6 +21,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
+
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
@@ -40,10 +41,14 @@ import org.mapstruct.ap.internal.model.beanmapping.MappingReference;
 import org.mapstruct.ap.internal.model.beanmapping.MappingReferences;
 import org.mapstruct.ap.internal.model.beanmapping.SourceReference;
 import org.mapstruct.ap.internal.model.beanmapping.TargetReference;
+import org.mapstruct.ap.internal.model.common.Assignment;
 import org.mapstruct.ap.internal.model.common.BuilderType;
+import org.mapstruct.ap.internal.model.common.FormattingParameters;
 import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.common.ParameterBinding;
+import org.mapstruct.ap.internal.model.common.SourceRHS;
 import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.model.common.TypeFactory;
 import org.mapstruct.ap.internal.model.dependency.GraphAnalyzer;
 import org.mapstruct.ap.internal.model.dependency.GraphAnalyzer.GraphAnalyzerBuilder;
 import org.mapstruct.ap.internal.model.source.BeanMappingOptions;
@@ -51,7 +56,9 @@ import org.mapstruct.ap.internal.model.source.MappingOptions;
 import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.model.source.SelectionParameters;
 import org.mapstruct.ap.internal.model.source.SourceMethod;
+import org.mapstruct.ap.internal.model.source.SubClassMappingOptions;
 import org.mapstruct.ap.internal.model.source.selector.SelectedMethod;
+import org.mapstruct.ap.internal.model.source.selector.SelectionCriteria;
 import org.mapstruct.ap.internal.util.Message;
 import org.mapstruct.ap.internal.util.Strings;
 import org.mapstruct.ap.internal.util.accessor.Accessor;
@@ -86,11 +93,10 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
     private final MethodReference finalizerMethod;
 
     private final MappingReferences mappingReferences;
+    private List<SubClassMapping> subClasses;
 
-    public static class Builder {
+    public static class Builder extends AbstractMappingMethodBuilder<Builder, BeanMappingMethod> {
 
-        private MappingBuilderContext ctx;
-        private Method method;
         private Type userDefinedReturnType;
 
         /* returnType to construct can have a builder */
@@ -109,9 +115,13 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         private MethodReference factoryMethod;
         private boolean hasFactoryMethod;
 
-        public Builder mappingContext(MappingBuilderContext mappingContext) {
-            this.ctx = mappingContext;
-            return this;
+        public Builder() {
+            super( Builder.class );
+        }
+
+        @Override
+        protected boolean shouldUsePropertyNamesInHistory() {
+            return true;
         }
 
         public Builder userDefinedReturnType(Type userDefinedReturnType) {
@@ -125,12 +135,12 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         }
 
         public Builder sourceMethod(SourceMethod sourceMethod) {
-            this.method = sourceMethod;
+            super.method( sourceMethod );
             return this;
         }
 
         public Builder forgedMethod(ForgedMethod forgedMethod) {
-            this.method = forgedMethod;
+            super.method( forgedMethod );
             mappingReferences = forgedMethod.getMappingReferences();
             Parameter sourceParameter = first( Parameter.getSourceParameters( forgedMethod.getParameters() ) );
             for ( MappingReference mappingReference: mappingReferences.getMappingReferences() ) {
@@ -164,7 +174,9 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                     // the userDefinedReturn type can also require a builder. That buildertype is already set
                     returnTypeImpl = returnTypeBuilder.getBuilder();
                     initializeFactoryMethod( returnTypeImpl, selectionParameters );
-                    if ( factoryMethod != null || canReturnTypeBeConstructed( returnTypeImpl ) ) {
+                    if ( factoryMethod != null
+                        || hasSubClassMappingsAndIsEitherAbstractOrCanBeConstructed( returnTypeImpl )
+                        || doesNotHaveSubClassMappingsAndCanBeConstructed( returnTypeImpl ) ) {
                         returnTypeToConstruct = returnTypeImpl;
                     }
                     else {
@@ -184,7 +196,9 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 else if ( !method.isUpdateMethod() ) {
                     returnTypeImpl = method.getReturnType();
                     initializeFactoryMethod( returnTypeImpl, selectionParameters );
-                    if ( factoryMethod != null || canReturnTypeBeConstructed( returnTypeImpl ) ) {
+                    if ( factoryMethod != null
+                        || hasSubClassMappingsAndIsEitherAbstractOrCanBeConstructed( returnTypeImpl )
+                        || doesNotHaveSubClassMappingsAndCanBeConstructed( returnTypeImpl ) ) {
                         returnTypeToConstruct = returnTypeImpl;
                     }
                     else {
@@ -332,6 +346,11 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
             }
 
+            List<SubClassMapping> subClasses = new ArrayList<>();
+            for ( SubClassMappingOptions subClassMappingOptions : method.getOptions().getSubClassMappings() ) {
+                subClasses.add( createSubClassMapping( subClassMappingOptions ) );
+            }
+
             MethodReference finalizeMethod = null;
 
             if ( shouldCallFinalizerMethod( returnTypeToConstruct ) ) {
@@ -349,8 +368,58 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 beforeMappingMethods,
                 afterMappingMethods,
                 finalizeMethod,
-                mappingReferences
+                mappingReferences,
+                subClasses
             );
+        }
+
+        private boolean doesNotHaveSubClassMappingsAndCanBeConstructed(Type returnTypeImpl) {
+            return !hasSubclassMappings()
+                && canReturnTypeBeConstructed( returnTypeImpl );
+        }
+
+        private boolean hasSubClassMappingsAndIsEitherAbstractOrCanBeConstructed(Type returnTypeImpl) {
+            return hasSubclassMappings()
+                && isReturnTypeAbstractOrcanBeConstructed( returnTypeImpl );
+        }
+
+        private SubClassMapping createSubClassMapping(SubClassMappingOptions subClassMappingOptions) {
+            TypeFactory typeFactory = ctx.getTypeFactory();
+            Type sourceType = typeFactory.getType( subClassMappingOptions.getSourceClass() );
+            Type targetType = typeFactory.getType( subClassMappingOptions.getTargetClass() );
+
+            SourceRHS rightHandSide = new SourceRHS(
+                "subClassMapping",
+                sourceType,
+                Collections.emptySet(),
+                "SubClassMapping for " + sourceType.getFullyQualifiedName() );
+            SelectionCriteria criteria =
+                SelectionCriteria
+                                 .forMappingMethods(
+                                     new SelectionParameters(
+                                         Collections.emptyList(),
+                                         Collections.emptyList(),
+                                         subClassMappingOptions.getTargetClass(),
+                                         ctx.getTypeUtils() ).withSourceRHS( rightHandSide ),
+                                     null,
+                                     null,
+                                     false );
+            Assignment assignment = ctx
+                                   .getMappingResolver()
+                                   .getTargetAssignment(
+                                       method,
+                                       null,
+                                       targetType,
+                                       FormattingParameters.EMPTY,
+                                       criteria,
+                                       rightHandSide,
+                                       null,
+                                           () -> forgeMapping( rightHandSide, sourceType, targetType ) );
+            return new SubClassMapping( sourceType, targetType, assignment, ctx.getTypeUtils() );
+        }
+
+        private boolean hasSubclassMappings() {
+            return !method.getOptions().getSubClassMappings().isEmpty();
         }
 
         private void initializeMappingReferencesIfNeeded(Type resultTypeToMap) {
@@ -555,6 +624,20 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                     Message.GENERAL_NO_SUITABLE_CONSTRUCTOR,
                     returnType.describe()
                 );
+                error = false;
+            }
+            return error;
+        }
+
+        private boolean isReturnTypeAbstractOrcanBeConstructed(Type returnType) {
+            boolean error = true;
+            if ( !returnType.isAbstract() && !returnType.hasAccessibleConstructor() ) {
+                ctx
+                   .getMessager()
+                   .printMessage(
+                       method.getExecutable(),
+                       Message.GENERAL_NO_SUITABLE_CONSTRUCTOR,
+                       returnType.describe() );
                 error = false;
             }
             return error;
@@ -1600,7 +1683,8 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                               List<LifecycleCallbackMethodReference> beforeMappingReferences,
                               List<LifecycleCallbackMethodReference> afterMappingReferences,
                               MethodReference finalizerMethod,
-                              MappingReferences mappingReferences) {
+                              MappingReferences mappingReferences,
+                              List<SubClassMapping> subClasses) {
         super(
             method,
             existingVariableNames,
@@ -1646,6 +1730,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             }
         }
         this.returnTypeToConstruct = returnTypeToConstruct;
+        this.subClasses = subClasses;
     }
 
     public List<PropertyMapping> getConstantMappings() {
@@ -1668,6 +1753,20 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
     public Type getReturnTypeToConstruct() {
         return returnTypeToConstruct;
+    }
+
+    public boolean hasSubClassMappings() {
+        return !subClasses.isEmpty();
+    }
+
+    public List<SubClassMapping> constructSubClassMappings(List<Parameter> parameters) {
+        subClasses.forEach( subClass -> subClass.updateWithParameters( parameters ) );
+        return subClasses;
+    }
+
+    public boolean isAbstractReturnType() {
+        return getFactoryMethod() == null && !hasConstructorMappings() && returnTypeToConstruct != null
+            && returnTypeToConstruct.isAbstract();
     }
 
     public boolean hasConstructorMappings() {
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
@@ -11,6 +11,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
+
 import javax.lang.model.element.ExecutableElement;
 
 import org.mapstruct.ap.internal.model.beanmapping.MappingReferences;
@@ -347,7 +348,7 @@ public class ForgedMethod implements Method {
 
     @Override
     public MappingMethodOptions getOptions() {
-        return basedOn.getOptions();
+        return MappingMethodOptions.getForgedMethodInheritedOptions( basedOn.getOptions() );
     }
 
     @Override
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
@@ -34,6 +34,7 @@ public class MappingMethodOptions {
         null,
         null,
         null,
+        Collections.emptyList(),
         Collections.emptyList()
     );
 
@@ -45,12 +46,14 @@ public class MappingMethodOptions {
     private EnumMappingOptions enumMappingOptions;
     private List<ValueMappingOptions> valueMappings;
     private boolean fullyInitialized;
+    private List<SubClassMappingOptions> subClassMapping;
 
     public MappingMethodOptions(MapperOptions mapper, Set<MappingOptions> mappings,
                                 IterableMappingOptions iterableMapping,
                                 MapMappingOptions mapMapping, BeanMappingOptions beanMapping,
                                 EnumMappingOptions enumMappingOptions,
-                                List<ValueMappingOptions> valueMappings) {
+                                List<ValueMappingOptions> valueMappings,
+                                List<SubClassMappingOptions> subClassMapping) {
         this.mapper = mapper;
         this.mappings = mappings;
         this.iterableMapping = iterableMapping;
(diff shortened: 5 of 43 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
```diff
@@ -0,0 +1,135 @@
...
+                    .printMessage(
+                        method,
+                        gem.mirror(),
+                        SUBCLASSMAPPING_METHOD_SIGNATURE_NOT_SUPPORTED );
+            return false;
+        }
+
+        TypeMirror sourceSubClass = gem.sourceClass().getValue();
+        TypeMirror targetSubClass = gem.targetClass().getValue();
+        TypeMirror sourceParentType = method.getParameters().get( 0 ).asType();
```
Zegveld, line 97: What is the correct way to determine the sourceParentType? Currently assuming that the first argument of the method is the source that needs to be mapped to the target.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
```diff
@@ -0,0 +1,135 @@
...
+                                          targetSubClass,
+                                          mapperOptions ) );
+        }
+        return subClassMappingOptions;
+    }
+
+    private static boolean isConsistent(SubClassMappingGem gem, ExecutableElement method, FormattingMessager messager,
+                                        TypeUtils typeUtils) {
+
+        if ( method.getParameters().isEmpty() || method.getReturnType().getKind() == TypeKind.VOID ) {
```
Zegveld, line 86: What is the correct way to determine the targetParentType? Currently assuming that the return value of the method is the target that needs to be mapped to.

filiphr:
Wow, thanks a lot for this PR @Zegveld. I haven't gotten around in reviewing this (picking the low hanging fruits first). However, we would look into this to make the 1.5.0.Beta2 release

processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
```diff
@@ -24,6 +24,19 @@
...
     </#if>
 
+    <#if hasSubClassMappings()>
+        <#assign first = true />
+        <#list constructSubClassMappings(parameters) as subClass>
+            <#if !first>else</#if> if (${subClass.sourceArgument} instanceof <@includeModel object=subClass.sourceType/>) {
+                <@includeModel object=subClass.assignment existingInstanceMapping=existingInstanceMapping/>
+            }
+            <#assign first = false />
+        </#list>
```
filiphr, line 34: no need to use the assignment for first. Check the [Accessing iteration state](https://freemarker.apache.org/docs/ref_directive_list.html#ref_list_accessing_state) for Freemarker. You can use `<#if subClass_index == 0>` instead. We are using that for the `afterMappingReferences`. I see that Freemarker has also added some more [Loop variable built-ins](https://freemarker.apache.org/docs/ref_builtins_loop_var.html), but we are not on that version yet 😄

core/src/main/java/org/mapstruct/SubClassMapping.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * @return the source subclass to check for before using the default mapping as fallback.
+     */
+    Class<?> sourceClass();
+
+    /**
+     * @return the target subclass to map the sourceSubClass to.
+     */
+    Class<?> targetClass();
+}
```
filiphr, line 84: From an API point of view I think it would make sense to only have `source` and `target` instead of `sourceClass` and `targetClass`. The annotation already talks about `SubClass`, so it should be clear from that that `source` and `target` are classes.

WDYT @sjaakd?

core/src/main/java/org/mapstruct/SubClassMapping.java
```diff
@@ -0,0 +1,84 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Configures the mapping to handle hierarchy of the source type.
+ * <p>
+ * The subclass to be mapped is to be specified via {@link #sourceClass()}.
+ * The subclass to map to is to be specified via {@link #targetClass()}.
+ * It currently is also required that a mapping method exist that has the {@link #targetClass()}
+ * as the return type and the {@link #sourceClass()} as the only argument.
```
filiphr, line 20: Let's try to get rid of this requirement. I've left bunch of comments around that I think will help for this.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -560,6 +629,20 @@ else if ( !returnType.hasAccessibleConstructor() ) {
             return error;
         }
 
+        private boolean isReturnTypeAbstractOrcanBeConstructed(Type returnType) {
```
filiphr, line 632: Small typo, "Orcan" to "OrCan"

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -349,10 +368,60 @@ else if ( !method.isUpdateMethod() ) {
...
+                                   .getMappingResolver()
+                                   .getTargetAssignment(
+                                       method,
+                                       null,
+                                       targetType,
+                                       FormattingParameters.EMPTY,
+                                       criteria,
+                                       rightHandSide,
+                                       null,
+                                           () -> forgeMapping( rightHandSide, sourceType, targetType ) );
```
filiphr, line 417: I actually think that this invalidates the statement that a method is needed for the `SubClassMapping`.

Can we also have some tests where the configuration from the base mapping is passed to the forged mappings as well?

processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
```diff
@@ -347,7 +348,7 @@ public boolean isObjectFactory() {
 
     @Override
     public MappingMethodOptions getOptions() {
-        return basedOn.getOptions();
+        return MappingMethodOptions.getForgedMethodInheritedOptions( basedOn.getOptions() );
```
filiphr, line 351: Why is this needed? Is this actually the fix for #2438? I would prefer if we can have that separate to reduce the surface area of a single PR.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -294,10 +296,20 @@ else if ( constructor.getParameters().size() == 1 ) {
...
 
 
+        for ( SourceMethod method : methods ) {
+            for ( SubClassMappingOptions subClassMapping : method.getOptions().getSubClassMappings() ) {
+                Type type = typeFactory.getType( subClassMapping.getTargetClass() );
+                extraImports.add( type );
+                type = typeFactory.getType( subClassMapping.getSourceClass() );
+                extraImports.add( type );
+            }
+        }
```
filiphr, line 311: I do not think that this is needed. The `SubClassMapping` already implements the `getImportTypes`, so you only need to call that for each `subClassMapping` in the `BeanMappingMethod#getImportTypes`

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1670,6 +1755,20 @@ public Type getReturnTypeToConstruct() {
...
     }
 
+    public boolean hasSubClassMappings() {
+        return !subClasses.isEmpty();
+    }
+
+    public List<SubClassMapping> constructSubClassMappings(List<Parameter> parameters) {
+        subClasses.forEach( subClass -> subClass.updateWithParameters( parameters ) );
+        return subClasses;
+    }
```
filiphr, line 1765: We strive for the models to be immutable if possible. This code is invoked from the `BeanMappingMethod.ftl`. Can we do this in a different place and only get the `subClassMappings` here?

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
```diff
@@ -188,6 +200,8 @@ public void applyInheritedOptions(SourceMethod templateMethod, boolean isInverse
                 }
             }
 
+            // Do NOT inherit subclass mapping options!!!
```
filiphr, line 203: Nice 😄

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -114,6 +114,9 @@
     ENUMMAPPING_NO_ELEMENTS( "'nameTransformationStrategy', 'configuration' and 'unexpectedValueMappingException' are undefined in @EnumMapping, define at least one of them." ),
     ENUMMAPPING_ILLEGAL_TRANSFORMATION( "Illegal transformation for '%s' EnumTransformationStrategy. Error: '%s'." ),
 
+    SUBCLASSMAPPING_ILLEGAL_SUBCLASS("Class '%s' is not a subclass of '%s'."),
+    SUBCLASSMAPPING_METHOD_SIGNATURE_NOT_SUPPORTED("SubClassMapping annotation can not be used on Suppliers and Runnables."),
```
filiphr, line 118: I do not think that this error message is clear. Having `void` return type means that the user is using update mappings, most likely we do not support that for now, so I would change the message to something like that.

By the way, you can do update mappings with a return type e.g. `Target update(@MappingTarget Target target, Source source)`

filiphr (commented):
Just finished with the review on this one. Really nice work @Zegveld. 

I've left some comments inline and I have some more general comments here.

I see that you are using the `_bugs` folder. We use this usually for actual bugs, not new features. I would suggest instead to use `subclassmapping` as top package and put your test there. You can then use different sub packages to show different use cases. In addition to that I am looking at the tests for #131 and #366. Correct me if I am wrong, but the only difference is the fact that in #131 `Vehicle` is not abstract, right?

Let's try to merge these things together and use the same source types and different target types. Perhaps you can also rename the source to `VehicleEntity` or somehow else, in order to avoid using full qualified names in the tests to make it more readable.

Can we also add some fixture tests, to see how the generated code looks like? As you know we strive to generate code which looks human readable 😄 .

Can we also add some erroneous tests to test the non happy paths for the new messages.

I am really excited in getting this in the upcoming 1.5. 

------
Something that could be good, but maybe we need to do it in a separate PR is to validate that the SubClassMapping make sense. For example if the first sub class mapping does not hide the second one.

e.g.

```
interface Vehicle { }

interface Car extends Vehicle { }

class SUV implements Car { }

class Sedan implements Car { }
```

And then if you defined sub class mapping like:

```
@SubClassMapping(source = Vehicle.class, target = ...)
@SubClassMapping(source = Car.class, target = ...)
```

Those subclass mappings will generate code which will never be executed.

e.g.

```
if (source instanceof Vehicle) {
    // ...
} else if (source instanceof Car) {
    // ...
}
```

After having written all of this, I think it makes sense to tackle it in a different PR. Let's do it like that, once we have this merged.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
```diff
@@ -0,0 +1,135 @@
...
+        if ( nonNull( subClassMappings ) ) {
+            subClassMappingAnnotations = subClassMappings.value().get();
+        }
+        else if ( nonNull( SubClassMappingGem.instanceOn( method ) ) ) {
+            subClassMappingAnnotations = Arrays.asList( SubClassMappingGem.instanceOn( method ) );
+        }
+        else {
+            return Collections.emptyList();
+        }
+
```
filiphr, line 61: My comment here got lost. Let's do this in a similar way like we are doing it for `@Mapping` and `@Mappings`. We don't need to support meta annotations for now, but the location where we pick them up makes more sense in `MethodRetrieval#getMethodRequiringImplementation`. There you also have access to the source parameters, return type etc. This also answers your other 2 questions in this class I beleive.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
filiphr: See https://github.com/mapstruct/mapstruct/pull/2512/#discussion_r697980035

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
filiphr: See https://github.com/mapstruct/mapstruct/pull/2512/#discussion_r697980035

filiphr:
By the way @Zegveld in https://github.com/mapstruct/mapstruct/issues/131#issuecomment-876330455 you are saying that you are not happy with the implementation route you took? Can you explain why? I personally thought that it is quite neat 😄 .

Zegveld:
> By the way @Zegveld in [#131 (comment)](https://github.com/mapstruct/mapstruct/issues/131#issuecomment-876330455) you are saying that you are not happy with the implementation route you took? Can you explain why? I personally thought that it is quite neat  .

Well that was before I made a lot more changes where in I made it quite neat ;)
Fixed it in this commit: https://github.com/rigd-loxia/mapstruct/commit/98c3d8e832d1277d1fa984a41c27ef168f8c1b17

processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
Zegveld: I think this change was to prevent infinite loops, so that the `@SubclassMapping` is not inherited. Will double-check this to be certain.

core/src/main/java/org/mapstruct/SubClassMapping.java
Zegveld: The existing mapping is no longer required, but I'm still not certain about the mappings that use the update route. That might need some more work, but let's see how far we'll get.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
Zegveld: This was a good idea, found a bug in the code with it.
Apparently the ForgedMethod that I created didn't get the MappingReferences of the parent method.
Fixing that as well now.

Zegveld:
Only done the bits so far where I've added a thumbs up.
Still got a few review items to go, but that will have to wait till when I have time again later this week.

core/src/main/java/org/mapstruct/SubClassMapping.java
```diff
@@ -0,0 +1,82 @@
...
+ * }
+ * </code></pre>
+ *
+ * @author Ben Zegveld
+ * @since 1.5
+ */
+@Repeatable( value = SubClassMappings.class )
+@Retention( RetentionPolicy.CLASS )
+@Target( { ElementType.METHOD, ElementType.ANNOTATION_TYPE } )
+public @interface SubClassMapping {
```
mjustin, line 72: Since the term is the single word "[subclass](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.1.4)" in Java, wouldn't `SubclassMapping` be a more appropriate camel casing than `SubClassMapping`?

documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
```diff
@@ -110,6 +110,34 @@ public interface SourceTargetMapper {
...
+
+    @SubClassMapping( sourceClass = AppleDto.class, targetClass = Apple.class )
+    @SubClassMapping( sourceClass = BananaDto.class, targetClass = Banana.class )
+    Fruit map( FruitDto source );
+
+}
+----
+====
+
+If you would just use a normal mapping both the AppleDto and the BananaDto would be made into a Fruit object, instead of an Apple and a Banana object. By using the SubClassMapping an AppleDtoToApple mapping will be done for Apples, and an BananaDtoToBanana mapping will be done for Banana's. If you would try to map a GrapeDto it would still turn into a Fruit. In the case that the Fruit type is an abstract class or an interface, you would get an IllegalArgumentException because it is unknown how to map a GrapeDto. Adding the missing (`@SubClassMapping`) for it will fix that.
```
mjustin, line 134: Minor, but the plural should be "bananas", not "banana's".

```suggestion
If you would just use a normal mapping both the AppleDto and the BananaDto would be made into a Fruit object, instead of an Apple and a Banana object. By using the SubClassMapping an AppleDtoToApple mapping will be done for Apples, and an BananaDtoToBanana mapping will be done for Bananas. If you would try to map a GrapeDto it would still turn into a Fruit. In the case that the Fruit type is an abstract class or an interface, you would get an `IllegalArgumentException` because it is unknown how to map a GrapeDto. Adding the missing (`@SubClassMapping`) for it will fix that.
```

Also, since the Apple/Banana/etc. types are (hypothetical) Java types, should they be done using code blocks?  Or is it better that they remain the normal code formatting?

Zegveld: Also spotted a few other things just now that can be worded a bit better. (Like "Apples" should be "`AppleDto` objects", as "Banana's" should have been "`BananaDto` objects") 
Will also use the code blocks for the hypothetical types and methods. Think that would make it clear whether something is java code related or not. (I'll add a thumbs up after updating it)

core/src/main/java/org/mapstruct/SubClassMapping.java
Zegveld: I'm fine with it either way, I did not look into how compound words are camelcased in this project.

Here is the reason why I used `SubClassMapping` instead of `SubclassMapping`:
For compound words I look for example if I can put the text ` of a ` in between and retain the meaning.
So for `subclass` I can do this, because `sub of a class` is correct. (in this case sub is short for substitute.)

What do you think, @filiphr ?

core/src/main/java/org/mapstruct/SubClassMapping.java
filiphr: I was more referring to the fact about "required that a mapping methods exists". That is not needed, since we can create an automatic method for it. We can mention that it does not support update methods though, no need to look into that for this PR

processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
filiphr: Makes sense, you need to remove the inherited `SubClassMapping`(s).  Let's create a field for this instead of creating it on every `getOptions` call. Those getOptions are called in a lot of places, so it might cause some performance problem.

core/src/main/java/org/mapstruct/SubClassMapping.java
filiphr: Although "SubclassMapping" might be "more" correct, I don't like it. The biggest reason why I like `SubClassMapping` more is that in an IDE I would always search by Typing "SB" and "SubclassMapping" is not returned in this case. However, reading more about it and looking at some other projects, it seems "Subclass" is more preferred (e.g. in the ByteBuddy project). 

In any case, lets leave the naming for the end. It would be easy to change it. We'll talk about it with the team.

filiphr:
>Only done the bits so far where I've added a thumbs up.
Still got a few review items to go, but that will have to wait till when I have time again later this week.

No problem @Zegveld, take as much time as you need. Please ping me once the PR is fully ready for the next iteration of review.

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
Zegveld: So something like: `@SubClassMapping do not support update mappings.`?
Open for suggestions to make this message better, because the one above doesn't feel quite right either.

core/src/main/java/org/mapstruct/SubClassMapping.java
Zegveld: I was referring to the same thing with the remark 'The existing mapping is no longer required'. :)
About the no support for update methods bit, I will add it (if not already there) to the documentation contained in the PR.

Note to self, this post requires a thumbs up when I've updated the documentation.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
Zegveld: Made the flow for the `@SubClassMapping` and the `@SubClassMappings` equal to `@Mapping` and `@Mappings` with some refactoring. Added a common base class and two specific classes for the flow. This should also make it easier to add more implementations using the same route in the future. Something for the second review round.

Also added a sorter to solve inheritence ordering issues, the more specific a class the earlier it should be in the list. But not yet implemented it, might be that it will only get a real implementation when someone runs into this as an issue.

processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
```diff
@@ -24,6 +24,19 @@
...
+        else {
+    </#if>
+    <#if isAbstractReturnType()>
+        throw new IllegalArgumentException("Not all subclasses are supported for this mapping. Missing for " + ${parameters[0].name}.getClass());
```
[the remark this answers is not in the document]
Zegveld: changed it so it uses the same value as the subclassmappings themselves do by getting the parameter name from the first subclassmapping. That way it will be the correct parameter.

On another note, no idea how to find the correct method in the following case:
```java
@SubClassMapping(....)
@Mapping(...., source="otherSource.value")
Target mapSuperClass(SourceSuperClass source, OtherSource otherSource);

@Mapping(....)
Target mapSubClass(SourceSubClass source, OtherSource otherSource);
```
didn't write out the whole thing, but my primary tests show that I'll generate a new method.
`Target mapSourceSubClassToTarget(SourceSubClass source)`
Instead of using the method `mapSubClass` with the additional parameter.

Although my suggestion to those who need this will be create a record containing those 2 arguments so that you only have to pass 1 argument. That way mapstruct will be able to handle it correctly.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
Zegveld: This bit is still in it although I could probably retrieve it some other way, it does not really make sense to use SubClassMappings with the `@MappingTarget` annotation. Seeing as then both the source and the target need to be instanceof checked in order to call the correct method. Why not let the caller handle that, seeing as that one most likely will know what it is because it created the target object.
So leaving this as it is.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
Zegveld: Replaced this with checking if any parameter matches. That way it doesn't matter which parameter the source parameter is.

Zegveld:
@filiphr , Think I've done all the adjustments. So that part is ready for a next iteration of review. Hopefully I haven't missed anything.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
```diff
@@ -0,0 +1,116 @@
...
+            return false;
+        }
+
+        TypeMirror sourceSubClass = gem.source().getValue();
+        TypeMirror targetSubClass = gem.target().getValue();
+        TypeMirror targetParentType = method.getReturnType();
+        boolean isConsistent = true;
+
+        boolean isChildOfAParameter = false;
+        for ( VariableElement parameter : method.getParameters() ) {
```
filiphr, line 58: Doing `method.getParameters()` is not quite correct since this will also take `@Context` parameters into consideration.  You can pass the real source parameters from the `MethodRetrievalProcessor`, since we have them there. I think you can use them there. There is also `targetParameter` and the `resultType` there as well.

filiphr (commented):
Great work @Zegveld. We are closing to the end 😄 . I've left some more comments inline. Will reply to yours and resolve what has been already done by you.

Thanks for bearing with us

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
filiphr: I've left a comment about this in my last review. The current solution will also work for

```
@SubClassMapping(source = SubArg1, target = Target)
@SubClassMapping(source = SubArg2, target = Target)
Target map(SuperArg1 arg1, @Context SuperArg2 arg2);
```

However, it shouldn't. `@Context` arguments should not be treated as potential source.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
filiphr: I've left a comment about how you can retrieve this in my last review.

processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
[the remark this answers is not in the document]
filiphr: Multi argument methods are not reused on our side. So you are correct not to focus on that.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubClassMappingOptions.java
filiphr: Yup I saw your refactoring. I like it a lot 😄 .

I've left a comment about the sorter in my last review. I don't think we need it for now.

documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
```diff
@@ -110,6 +110,40 @@ public interface SourceTargetMapper {
...
+
+[TIP]
+====
+If the mapping method for the subClasses does not exist it will be created and any other annotations on the fruit mapping method are inherited by the newly generated mappings.
+====
+
+[NOTE]
+====
+Combining `@SubClassMapping` with update methods is not supported. If you use an update method with a return value then the subclass mappings will return a new instance based on the mapping, but it will not take into account the `@MappingTarget` parameter.
+The same issue exists for the `@Context` and `@TargetType` parameters.
```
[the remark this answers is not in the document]
Zegveld: But then you would also need to check if the `@MappingTarget` is of the same type as the `@SubClassMapping(target=TargetSub.class, source=SourceSub.class)` so you would end up with 2 instance of checks in order to call the subclass method which would then become: `void updateTargetSubWithSourceSub(@MappingTarget TargetSub target, SourceSub source)`. So the implementation of the mapping method calling that update method would look something like this:
```java
void updateMapping( @mappingTarget Target target, Source source ) {
    if ( source instanceof SourceSub && target instanceof TargetSub ) {
        updateTargetSubWithSourceSub( (TargetSub) target, (SourceSub) source );
    } else if ( source instanceof SourceSub) {
        // what now??? source argument matches subclass of sourceClass but target argument doesn't match subclass of targetClass.
    } else {
        // default update behavior
    }
}
```
But how to search for an existing method or method definition that can be called?

I think it would indeed be best to have that as another Issue: support for subclassmappings in combination with update mappings. Should I create an issue for this so that we can discuss this topic in that issue's thread?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -184,7 +196,9 @@ else if ( userDefinedReturnType != null ) {
                 else if ( !method.isUpdateMethod() ) {
                     returnTypeImpl = method.getReturnType();
                     initializeFactoryMethod( returnTypeImpl, selectionParameters );
-                    if ( factoryMethod != null || canReturnTypeBeConstructed( returnTypeImpl ) ) {
+                    if ( factoryMethod != null
+                        || hasSubClassMappingsAndIsEitherAbstractOrCanBeConstructed( returnTypeImpl )
```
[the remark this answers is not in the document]
Zegveld: That method is still used, inside the others.

Basically there are 2 situations that can happen now.
1. No subclass annotation and return type can be constructed.
2. With subclass annotation and an abstract class or interface as return type or a return type that can be constructed.

So got 2 methods for that now:
1. `doesNotHaveSubClassMappingsAndCanBeConstructed( returnTypeImpl )`
method implementation: `!hasSubclassMappings() && canReturnTypeBeConstructed( returnTypeImpl )`
2. hasSubClassMappingsAndIsEitherAbstractOrCanBeConstructed
method implementation: `hasSubclassMappings() && isReturnTypeAbstractOrCanBeConstructed( returnTypeImpl )`

The methods `isReturnTypeAbstractOrCanBeConstructed` and `canReturnTypeBeConstructed` are the ones that also handle the error message stuff.

Couldn't think of a shorter name that would explain it correctly. So if you have better method suggestions I'm all for it.

processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
```diff
@@ -319,6 +342,10 @@ private SourceMethod getMethodRequiringImplementation(ExecutableType methodType,
             .build();
     }
 
+    private List<SubClassMappingOptions> sortSubClassMappings(Set<SubClassMappingOptions> subClassMappingOptions) {
+        return subClassMappingOptions.stream().sorted( new SubClassInheritenceSorter() ).collect( Collectors.toList() );
```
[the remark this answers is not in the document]
Zegveld: True, the current implementation doesn't do any sorting atm, but just converts it into a list. 
I noticed that a `LinkedHashSet` is used, I guess that the reason for this is to maintain the user defined order while keeping it distinct?
With a `Set` my assumption would be that the order does not matter. Wouldn't it then be better to explicitly use the `LinkedHashSet` to identify the collection instead of a `Set`?

core/src/main/java/org/mapstruct/SubClassMapping.java
filiphr: Discussed this in the team and we decided to go with `SubclassMapping`. I can do that change.

Zegveld:
@filiphr made changes to handle 2 of the issues (marked with the thumbs up), but got some questions about the other 3, see the comments per issue.
I will also do the rename from SubClass to Subclass, that is no problem.

processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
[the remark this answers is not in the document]
Zegveld: Made it explicit that the Set is a LinkedHashSet and not a Set with undefined ordering.
Also removed the sorter because it is not needed anymore.

Zegveld:
Just read the first review again and noticed the bit about:
`Can we also add some fixture tests, to see how the generated code looks like? As you know we strive to generate code which looks human readable smile .`
Where can I find an example of fixture tests so that I can add them for the subclass mappings as well.

Also added a warning for `Something that could be good, but maybe we need to do it in a separate PR is to validate that the SubClassMapping make sense. For example if the first sub class mapping does not hide the second one.`

Zegveld:
I've added fixture tests. Made these separate so that the main variants are covered and not impacted by changes within the general tests. 
@filiphr, hope to hear from you soon for the 2 remaining issues, and also what you think of my solution for the annotation handling order one.

sjaakd (changes requested):
First of all: what a tremendous feat of work. Thanks!. When discussing this with @filiphr I was amazed by the effort done on this.

I did not check the complete implementation. I recon that @filiphr did that already. So I focussed on the API extensions and the implications there.

Now: we do like to go more into meta-annotations in 2.0. So, isolating functionality (`sourceClass` and `targetClass`) into a new annotation is a good idea. I've one consideration here though. 

MapStruct recognises a few major classes of mappings: mapping bean-to-bean, mapping iterable-to-iterable, map-to-map and value-to-value. BeanMapping obviously the most important one. Also, each comes with its own annotation for doing configuration: `@BeanMapping`, `@IterableMapping`, `@MapMapping`, `@ValueMapping`.

There is check functionality that guarantees that  a `MapMapping` is not accidentally applied to `@BeanMapping`. In fact, we do have a small problem with the just released map-to-bean-mapping, which under some conditions gets mixed up with bean mapping.

Having said that: would it not be better to make `@SubclassMapping` a property of `@BeanMapping`? Although more verbose, it would keep functionality together and make sure `@SubclassMapping` is not accidentally applied to another form of mapping.

In 2.0 we could give those mappings a more " independent life", perhaps making even use of the fact that they are part of a `@BeanMapping`. But lets do that step-by-step. 

Next, I was thinking how this functionality would interplay with the new `MapToBean` functionality. Do we have an idea 'bout that?

anyway @Zegveld : what a nice piece of work on an issue which has a long history. Really looking forward to work with you on more issues in the future.

filiphr:
Thanks for the review @sjaakd. However, I have to say that I do not quite agree with moving `@SubclassMapping` to the `BeanMapping`. I believe that `@SubclassMapping` is on the same level as `@ValueMapping` and `@Mapping`.  

Adding it to the `@BeanMapping` to only remove it later on doesn't bring much value I think. @Zegveld also did a really nice refactoring which makes supporting meta annotations work out of the box already.

> and also what you think of my solution for the annotation handling order one.

@Zegveld you mean the refactoring you did, right? I found it really great, already said that in https://github.com/mapstruct/mapstruct/pull/2512#discussion_r703782441 😉 .

-----
@Zegveld I've been a bit busy, so I haven't gotten around in finishing the review with your last changes. I haven't forgotten about this one yet.

Zegveld:
> First of all: what a tremendous feat of work. Thanks!. When discussing this with @filiphr I was amazed by the effort done on this.
> 
> I did not check the complete implementation. I recon that @filiphr did that already. So I focussed on the API extensions and the implications there.
> 
> Now: we do like to go more into meta-annotations in 2.0. So, isolating functionality (`sourceClass` and `targetClass`) into a new annotation is a good idea. I've one consideration here though.
> 
> MapStruct recognises a few major classes of mappings: mapping bean-to-bean, mapping iterable-to-iterable, map-to-map and value-to-value. BeanMapping obviously the most important one. Also, each comes with its own annotation for doing configuration: `@BeanMapping`, `@IterableMapping`, `@MapMapping`, `@ValueMapping`.
>
> There is check functionality that guarantees that a `MapMapping` is not accidentally applied to `@BeanMapping`. In fact, we do have a small problem with the just released map-to-bean-mapping, which under some conditions gets mixed up with bean mapping.
> 
> Having said that: would it not be better to make `@SubclassMapping` a property of `@BeanMapping`? Although more verbose, it would keep functionality together and make sure `@SubclassMapping` is not accidentally applied to another form of mapping.
> 

I didn't think about this initially, but it would be possible to do submappings for iterables or maps as well. This would be useful if the mapping is reused multiple times, but the implementation collection varies. And you could even support implementations of the Collection framework that you do not know exist. (e.g. project specific ones) 
For `@ValueMapping`s I'm not that certain, if this is only enums then no. But if this is also static fields than the same thing could apply. Because static fields can refer to subclassed instances of the source. 
It would require some more work to support the `@IterableMapping`, `@MapMapping` and `@ValueMapping` annotations with subclasses but I do believe that would be possible. Not certain as to how often it would be used if at all though, because the above is pure theoretical.

> In 2.0 we could give those mappings a more " independent life", perhaps making even use of the fact that they are part of a `@BeanMapping`. But lets do that step-by-step.
> 
> Next, I was thinking how this functionality would interplay with the new `MapToBean` functionality. Do we have an idea 'bout that?

I think that could turn out to be an adjusted variant of the `@SubclassMapping` as is, but instead of supplying a source class you would supply a sourceCondition. If this sourceCondition would result in a true value a more specific MapToBean method would be called. Something for in another PR but here is the gist of it:
```
@SubclassMapping( sourceCondition="java(input.containsKey('subBeanSpecificField'))", target=SubBean.class )
Bean mapToBean(Map<String, String> input);

@Mapping(/*specific mapping for this condition*/
SubBean mapToSubBean(Map<String, String> input);
```
which could result in something like:
```
Bean mapToBean(Map<String, String> input) {
    if (input.containsKey('subBeanSpecificField')) {
        return mapToSubBean(input);
    }
    // default mapToBean behaviour
}

SubBean mapToSubBean(Map<String, String> input) {
    // mapToSubBean behaviour
}
```
> anyway @Zegveld : what a nice piece of work on an issue which has a long history. Really looking forward to work with you on more issues in the future.

Likewise.



> > and also what you think of my solution for the annotation handling order one.
>
> @Zegveld you mean the refactoring you did, right? I found it really great, already said that in #2512 (comment) wink .

@filiphr, no I did a refactor which changes several Set's into specifically LinkedHashSet's. Was talking about that one.

> @Zegveld I've been a bit busy, so I haven't gotten around in finishing the review with your last changes. I haven't forgotten about this one yet.

No problem I've got plenty of waiting time. Currently implementing just a lot of instanceof checks manually when I'm using it instead of generating them. :)

sjaakd:
> I didn't think about this initially, but it would be possible to do submappings for iterables or maps as well. This would be useful if the mapping is reused multiple times, but the implementation collection varies. And you could even support implementations of the Collection framework that you do not know exist. (e.g. project specific ones)

You have a point here. Did not consider that. I think @filiphr and I were discussing the change of the default implementation (don't remember if it was `Map` or `List`) a while ago.


> For `@ValueMapping`s I'm not that certain, if this is only enums then no. But if this is also static fields than the same thing could apply. Because static fields can refer to subclassed instances of the source.

That would be really a niche feature. 

I was discussing mis-use of annotations with @filiphr just earlier and he pointed out that we don't have (presumably) checks for `@Mapping` either. So we're a bit half-hearted with checks here anyway. 

Finally, applying it to the new map-to-bean would indeed qualify as a follow up PR.

So, as far as I'm concerned: go ahead.

sjaakd (approved):

filiphr:
I checked out the PR locally today and did some changes in it. I've pushed the changes to https://github.com/filiphr/mapstruct/tree/feature/PR-131 (I couldn't push it to your branch for some reason, most likely you have not allowed maintainers to push changes to it). Can you please have a look at them and let me know whether you agree with them @Zegveld? I tried to do them atomically in order to show my thought process for some of them, everything will be merged into one commit in the end.

>@filiphr, no I did a refactor which changes several Set's into specifically LinkedHashSet's. Was talking about that one.

I saw that one, and I didn't really like it. I do not like having collection implementation types as fields of classes or parameters. I reverted some of that change in a separate PR.

Some other more notable changes I did are:

* We do not support all update methods, not just the ones that are void. Otherwise it is too confusing
* Small Refactoring in the RepeatableMappingAnnotations to reduce the complexity a bit
* Remove obsolete changes in MapperCreationProcessor

Some final things to address before merging:

* I am not sure that the throwing of the `IllegalArgumentException` is a good approach. The reason for that is that if you remove the `SubclassMapping`(s) we will have a compile error, otherwise we throw a runtime error. Can we keep this behaviour consistent? In a follow up PR we can add support for throwing the exception if the user wants it by explicitly asking for it via `@BeanMapping` and in another PR to support sealed classes, i.e. we can know whether the `SubclassMapping`(s) have all classes for the source or not.

I think that this would be the last change I request, together with my changes the PR would be ready for merging and doing a beta release shortly after that. It would be a great addition to the 1.5 release 😄 

------

Side note: Since your email address is private, can you please reach out to me somehow? You could use [Twitter](https://twitter.com/filiphr) or my email address (it is in my commits). I'd like to discuss something with you.

Zegveld:
I did take that route under consideration, but the use case that I had which caused me to write this enhancement was one where I never have an implementation for general usage. In this use case the classes are generated from two xsd's and I need to transfer information between the two models.

The best solution I can think of usage side without having the exception would be to manually create an implementation/extension of the base class which then would throw an IllegalStateException in it's constructor because it should not be created. That one would then be the default implementation that mapstruct falls back to. This does not feel that good as a solution for this problem.

What if we add another annotation `@SubclassMappingStrategy` in which you can configure whether you get an exception or compile error. Hmmm maybe with a field/property named .... can't think of a good name for the property atm. But that property would then influence the behavior. Just like how the other MappingStrategy annotations do.

Zegveld:
@filiphr , I've introduced the SubclassExhaustiveStrategy and did some more renaming/refactoring. Also updated the documentation to have it mentioned.
Hope you have time to review the new changes.

Issue #2515, opened 2021-07-14
Ambiguous constructors error message does not contain the ambiguous constructors


basclaessen:
The ambiguous constructors error message does not contain the ambiguous constructors.
Adding the ambiguous constructors to the message can help fixing the error.

PR #2516, opened 2021-07-14
#2515 Added ambiguous constructors to ambiguous constructor error mes…

basclaessen:
Added ambiguous constructors to ambiguous constructor error message

Fixes: #2515

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -26,6 +26,7 @@ import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
+import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.ElementFilter;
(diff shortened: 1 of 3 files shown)


filiphr:
Good catch @basclaessen. Looking at the tests, I am not sure how I feel about eclipse 😄 . I would rather craft the elements on our side appropriately, so our Eclipse users also see better messages 😄.

Can you please do that as well?

basclaessen:
@filiphr Sure!
I changed the constructors to string conversion. The error message when using the Eclipse compiler is now equal to the error message when using the JDK compiler.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -699,7 +700,14 @@ private ConstructorAccessor getConstructorAccessor(Type type) {
                     method.getExecutable(),
                     GENERAL_AMBIGUOUS_CONSTRUCTORS,
                     type,
-                    Strings.join( constructors, ", " )
+                    constructors.stream()
+                        .map( ExecutableElement::getParameters )
+                        .map( ps -> ps.stream()
+                            .map( VariableElement::asType )
+                            .map( String::valueOf )
+                            .collect( Collectors.joining( ",", type.getName() + "(", ")" ) )
```
filiphr, line 708:
```suggestion
                            .collect( Collectors.joining( ", ", type.getName() + "(", ")" ) )
```

processor/src/test/java/org/mapstruct/ap/test/constructor/erroneous/ErroneousConstructorTest.java
```diff
@@ -29,8 +29,10 @@
             kind = javax.tools.Diagnostic.Kind.ERROR,
             line = 17,
             message = "Ambiguous constructors found for creating org.mapstruct.ap.test.constructor.erroneous" +
-                ".ErroneousAmbiguousConstructorsMapper.PersonWithMultipleConstructors. Either declare parameterless " +
-                "constructor or annotate the default constructor with an annotation named @Default."
+                ".ErroneousAmbiguousConstructorsMapper.PersonWithMultipleConstructors: " +
+                "PersonWithMultipleConstructors(java.lang.String), " +
+                "PersonWithMultipleConstructors(java.lang.String,int). Either declare parameterless constructor " +
```
filiphr, line 34:
```suggestion
                "PersonWithMultipleConstructors(java.lang.String, int). Either declare parameterless constructor " +
```

filiphr (commented):
Thanks for the changes @basclaessen. I've added some small suggestion (which I will accept and then merge this PR).

Issue #2041, opened 2020-03-12
Request to add @Profile supporting features for spring integration


JokerSun:
I have some case that needs add @Profile annotation on mapper, but when I annotate it directly, it doesn't work. 
I guess it hasn't been supported yet.
```
e.g.

@Profile(Constants.PROFILE)
@Mapper(componentModel = "spring")
public interface SomeConverter {
    ...
}
```

filiphr:
`@Profile` is a Spring annotation and we don't have any particular support for those. This is really similar with #1427.

However, this is quite easy to be done with a custom `@ComponentScan` that would only scan the `SomeConverter` package when the profile is active.

I am leaning towards closing this as wontfix. Not sure what the team thinks.

JokerSun:
Well, That sounds reasonable, and does the team have a plan to develop the spring integration? I mean the spring-boot-starter-mapstruct will be useful.

Chessray:
On a more abstract level, this reminds me strongly of the discussion in #990 about a possibility to add _any_ annotation to a Mapper. We seem to have a common topic here.

JokerSun:
@Chessray 
That's it. I think I should close the issue, and thanks for your reply.

PR #2517, opened 2021-07-16
Added new componentModel to improve Dagger2 support

leonardortlima:
Added a new componentModel for Dagger2, as suggested in #1617 and #2337

Dagger2 is mostly compatible with the existing `jsr330` componentModel, but it has 2 issues:

1. For constructor injection, a constructor annotated with `@Inject` should always be present, even if the constructor is empty
2. For field injection, it requires the fields to be package private (no access modifiers)

Issue 1 should be fixed by the current state of this PR
Issue 2 would involve changing more stuff,  and I'm not sure if you'd be interested in supporting it, since constructor injection is preferred with Dagger

Also, I'd like some help fixing the integration tests, if possible. I'm having trouble making both dagger and mapstruct annotation processing work during testing

### DIFF
--- a/integrationtest/src/test/resources/dagger2Test/pom.xml
+++ b/integrationtest/src/test/resources/dagger2Test/pom.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright MapStruct Authors.
+
+    Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.mapstruct</groupId>
+        <artifactId>mapstruct-it-parent</artifactId>
+        <version>1.0.0</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>dagger2Test</artifactId>
+    <packaging>jar</packaging>
+
+    <properties>
+        <dagger.version>2.37</dagger.version>
+    </properties>
+    
+    <dependencies>
+        <dependency>
+            <groupId>com.google.dagger</groupId>
+            <artifactId>dagger</artifactId>
+            <version>${dagger.version}</version>
+        </dependency>
+    </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>org.mapstruct</groupId>
+                            <artifactId>mapstruct-processor</artifactId>
+                            <version>${mapstruct.version}</version>
+                        </path>
+                        <path>
+                            <groupId>com.google.dagger</groupId>
+                            <artifactId>dagger-compiler</artifactId>
+                            <version>${dagger.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+</project>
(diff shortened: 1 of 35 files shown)


processor/src/test/java/org/mapstruct/ap/test/injectionstrategy/dagger2/constructor/Dagger2ConstructorMapperTest.java
```diff
@@ -0,0 +1,90 @@
...
+    CustomerDto.class,
+    CustomerEntity.class,
+    Gender.class,
+    GenderDto.class,
+    CustomerDagger2ConstructorMapper.class,
+    GenderDagger2ConstructorMapper.class,
+    ConstructorDagger2Config.class
+})
+@ComponentScan(basePackageClasses = CustomerDagger2ConstructorMapper.class)
+@Configuration
```
filiphr, line 42: Why are we using Spring for this?

processor/src/test/java/org/mapstruct/ap/test/injectionstrategy/dagger2/emptyconstructor/Dagger2EmptyConstructorMapperTest.java
```diff
@@ -0,0 +1,82 @@
...
+ * Test constructor injection for component model dagger2, when a mapper has no extra dependencies.
+ */
+@WithClasses({
+    Gender.class,
+    GenderDto.class,
+    ConstructorDagger2Config.class,
+    GenderDagger2EmptyConstructorMapper.class
+})
+@ComponentScan(basePackageClasses = GenderDagger2EmptyConstructorMapper.class)
+@Configuration
```
filiphr, line 38: Same question here, why Spring?

integrationtest/src/test/resources/dagger2Test/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+                        <path>
+                            <groupId>org.mapstruct</groupId>
+                            <artifactId>mapstruct-processor</artifactId>
+                            <version>${mapstruct.version}</version>
+                        </path>
+                        <path>
+                            <groupId>com.google.dagger</groupId>
+                            <artifactId>dagger-compiler</artifactId>
+                            <version>${dagger.version}</version>
+                        </path>
```
filiphr, line 49: I know that this are best practices for annotation processors. We are also pushing the same. However, our infrastructure for the integration tests doesn't work the best with this. I think that if you add the `dagger-compiler` as a provided dependency, the tests would work.

filiphr (commented):
Thanks for this PR @leonardortlima.

Great that Issue 1 is covered here.

Regarding Issue 2. I think that we should still offer that possibility. I don't think that we need to change a lot for this. We need to adapt something in `AnnotationBasedComponentModelProcessor#replacementMapperReference` and then in `AnnotationMapperReference` and `AnnotationMapperReference.ftl` be able to create package private fields.

I've left some other comments for the process-tests and the integration-tests. 
I'll look at the CI as well once the tests fail

processor/src/test/java/org/mapstruct/ap/test/injectionstrategy/dagger2/constructor/Dagger2ConstructorMapperTest.java
leonardortlima: These tests were basically copy and pasted from [the JRS330 tests](https://github.com/mapstruct/mapstruct/blob/master/processor/src/test/java/org/mapstruct/ap/test/injectionstrategy/jsr330/constructor/Jsr330ConstructorMapperTest.java), with updated checks for Dagger behavior

integrationtest/src/test/resources/dagger2Test/pom.xml
leonardortlima: I was testing different strategies trying to make the integration tests work, but still had no success

leonardortlima:
Thanks for the feedback @filiphr. 

I will try to implement field injection too

processor/src/test/java/org/mapstruct/ap/test/injectionstrategy/dagger2/constructor/Dagger2ConstructorMapperTest.java
filiphr: Interesting. I think that the JSR 330 tests are like that because Spring supports that as well. Makes sense to only check the generated code here. 

The integration tests are testing the full integration from what I can see, so that's great

filiphr:
@facboy what do you mean by "can you support custom scope annotations while doing this?"

facboy:
you can annotate dagger injectables with any scope, it doesn't have to be `Singleton`.  mapstruct should support specifying the scope annotation for the generated mapper.

filiphr:
Thanks for the explanation @facboy. The initial support will not take that into consideration. The other dependency injection frameworks have such scope functionality, but we do not support it currently. 

That is more related to something like #1427, #990 and #2041.

wonsuc:
Any updates on this?

Rubydesic:
This should also change MapStruct mappers to be marked `@Reusable` instead of `@Singleton`

PR #2526, opened 2021-07-22
#2525 add available transformations to CaseEnumTransformationStrategy exception

### LINKED ISSUE #2525: Provide available case transformation when unknown is used in CaseEnumTransformationStrategy
This issue was inspired by https://stackoverflow.com/questions/68468415/error-to-map-enums-with-mappingconstants-case-transformation.

Ideally we should include the supported case transformations in the error message in 

https://github.com/mapstruct/mapstruct/blob/4576103752df896d4548c1d618ab892bf540b628/processor/src/main/java/org/mapstruct/ap/spi/CaseEnumTransformationStrategy.java#L39-L40

This issue is a simple one and is an ideal candidate for someone from the community to implement it. Please let us know if you are interested to work on it.

amogh94:
Fixes #2525


filiphr:
Thanks for the PR @amogh94. It looks as expected, will merge it once the CI is done

PR #2542, opened 2021-08-09
implicit source properties now get removed from unprocessedSourceProperties

### LINKED ISSUE #2537: Issue with unmappedSourcePolicy and implicit source mapping
I have a mapper in which the source properties consists of just a subset of the target properties.  I'm mapping those properties, but leaving everything else unchanged.  I'm using `@BeanMapping(ignoreByDefault = true)` to ensure and make explicit that I'm only mapping a small subset of properties over.  I discovered that I can specify just the `target` property in a `@Mapping` annotation to opt a field in to this mapping, when the source property name is the same as the target's:

```java
@Mapper
interface MyMapper {
    @BeanMapping(ignoreByDefault = true)
    @Mapping(target = "renamedTarget", source = "renamedSource")
    @Mapping(target = "property") // source = "property" isn't needed
    void copyProperties(Source source, @MappingTarget Target target);
}
```

I would like to specify the unmapped source policy to fail if I don't map all source properties.  This way I'll be notified at build time if any properties get added to the source that I forget to map.  Using `@Mapper(unmappedSourcePolicy = ReportingPolicy.ERROR)` works, but only if I explicitly specify the source property (`@Mapping(target = "property", source = "property")`).  If I don't, I get a build error:

```
Unmapped source property: "property".
```

Given that the source is actually mapped by MapStruct in this scenario, I would expect this to compile without error.

## Minimal example:

```java
@Mapper(unmappedSourcePolicy = ReportingPolicy.ERROR) // Works without the unmappedSourcePolicy
public interface UnmappedSourcePolicyWithImplicitSourceMapper {
    @Mapping(target = "property") // Works if I add source = "property"
    Target map(Source source);
}

public static class Source {
    private String property;

    public String getProperty() { return property; }
    public void setProperty(String property) { this.property = property; }
}

public static class Target {
    private String property;

    public String getProperty() { return property; }
    public void setProperty(String property) { this.property = property; }
}
```

Zegveld:
Fixes #2537.
Using the situation as reported in the issue as test.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2537/ImplicitSourceTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2537/ImplicitSourceTest.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2537;
+
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.ProcessorTest;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.compilation.annotation.CompilationResult;
+import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutcome;
+
+/**
+ * implicit source-target mapping should list the source property as being mapped.
+ *
+ * @author Ben Zegveld
+ */
+@WithClasses( { UnmappedSourcePolicyWithImplicitSourceMapper.class } )
+@IssueKey( "2537" )
+public class ImplicitSourceTest {
+
+    @ProcessorTest
+    @ExpectedCompilationOutcome( value = CompilationResult.SUCCEEDED, diagnostics = {} )
+    public void reportsProperWarningsAndError() {
+    }
+}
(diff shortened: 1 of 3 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_2537/ImplicitSourceTest.java
```diff
@@ -0,0 +1,27 @@
...
+ * implicit source-target mapping should list the source property as being mapped.
+ *
+ * @author Ben Zegveld
+ */
+@WithClasses( { UnmappedSourcePolicyWithImplicitSourceMapper.class } )
+@IssueKey( "2537" )
+public class ImplicitSourceTest {
+
+    @ProcessorTest
+    @ExpectedCompilationOutcome( value = CompilationResult.SUCCEEDED, diagnostics = {} )
```
filiphr, line 24: This is not needed, since if the compilation fails the test will fail as well

filiphr (commented):
Thanks a lot for the speedy fix @Zegveld.

The PR looks great.

I have one small request though. I know that you took the test from #2537. However, when I initially saw it, I was confused why we have `Mapping#target` only.  Such a mapping doesn't make a lot of sense on its own. I then saw the issue and saw that the user has `@BeanMapping(ignoreByDefault = true)`.

Can we please adapt the test to use that as well?

Zegveld:
Updated the test with the review feedback.

filiphr:
Thanks @Zegveld

PR #2548, opened 2021-08-13
Add URL to String built-in conversions

### LINKED ISSUE #2552: Add built in conversion between URL and String

szatyinadam:
In our projects at my job is a frequent use case to convert URL to String. Now we copy the same methods from project to other project to do this conversion.

This built-in conversion would be useful for us and I thought it could be useful for others as well.

Fixes #2552

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
@@ -7,6 +7,7 @@ package org.mapstruct.ap.internal.conversion;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.net.URL;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.time.Duration;
@@ -193,6 +194,8 @@ public class Conversions {
         register( Currency.class, String.class, new CurrencyToStringConversion() );
 
         register( UUID.class, String.class, new UUIDToStringConversion() );
+
+        registerURLConversion();
     }
 
     private void registerJodaConversions() {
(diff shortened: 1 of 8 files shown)


processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
```diff
@@ -193,6 +194,7 @@ public Conversions(TypeFactory typeFactory) {
         register( Currency.class, String.class, new CurrencyToStringConversion() );
 
         register( UUID.class, String.class, new UUIDToStringConversion() );
+        register( URL.class, String.class, new URLToStringConversion() );
```
filiphr, line 197: The `URL` is part of the `java.net` module which means that it might not be available. Lets do this in a similar way as it is done for the `java.sql` module. Look at `registerJavaTimeSqlConversions`

processor/src/test/java/org/mapstruct/ap/test/conversion/url/URLConversionTest.java
```diff
@@ -0,0 +1,56 @@
...
+        target.setURL( "https://mapstruct.org/" );
+
+        Source source = URLMapper.INSTANCE.targetToSource( target );
+
+        assertThat( source ).isNotNull();
+        assertThat( source.getURL() ).isEqualTo( new URL( target.getURL() ) );
+    }
+
+    @ProcessorTest
+    public void shouldHandleInvalidURLString() {
```
filiphr, line 48: This is a great test. Let's add one more test where the mapping method throws `MalformedURLException` as well.

filiphr (commented):
Great work on this @szatyinadam. 

I've left some comments in the PR, it would be good if you can have a look at them.

In addition to the comments, can you please add this to the documentation as well. The section for this is [here](https://github.com/mapstruct/mapstruct/blob/master/documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc#implicit-type-conversions).

szatyinadam:
Thank you @filiphr for the review.
I've fixed what you mentioned and I've added the test you asked and an other one for null checking.
Can you take a look again on this PR?

filiphr:
Thanks a lot for the PR @szatyinadam, I've integrated this and it will be available in the next release.

PR #2550, opened 2021-08-14
#2541 fix incorrect name for TypeVar with ElementType.TYPE_USE for javac-with-errorprone

### LINKED ISSUE #2541: TYPE_USE annotations on method parameters cause them not to match when using errorprone
When using the errorprone javac compiler (which afaik is based on a newer compiler version than jdk8) Mapstruct gets confused when method parameters have `TYPE_USE` annotations (eg as used in `checker-qual`).  It seems to see the annotated type variable and the unannotated type variable as different types:

```
static <T> Optional<T> toOptional(@Nullable T value) {
    return Optional.ofNullable(value);
}
```

Example project: https://github.com/facboy/mapstruct-bug

filiphr:
Fixes #2541


sjaakd (approved):
LGTM

filiphr:
Thanks for the review

PR #2568, opened 2021-08-30
Ignore source properties if ignoreByDefault = true (closes #2560)

### LINKED ISSUE #2560: BeanMapping / ignoreByDefault interaction with unmappedSourcePolicy
A rather simple issue:

If I specify `@BeanMapping(ignoreByDefault = true)` and `unmappedSourcePolicy != ReportingPolicy.IGNORE`, non-mapped source parameters still cause an error. Arguably, this might be the correct behaviour, but I think it is not:

`ignoreByDefault` basically means "I will do everything manually". In my understanding, reporting policies are there to warn you about things the automatic matching is doing. In particular, if you have an `unmappedTargetPolicy` + an unmapped target property with `ignoreByDefault = true`, there are no errors (as, essentially, this throws an `ignore = true` on every target property). In the current state, this is asymmetric for the unmapped sources case, as `@BeanMapping.ignoreUnmappedSourceProperties` practically is the equivalent of `@Mapping(ignore = true)`.

Proposal: `ignoreByDefault` also disables complaints about unmapped source properties :-)

This should be quite easy to implement, if you agree I can submit a PR.

incaseoftrouble:
I implemented this at the "check" level - `unprocessedSourceProperties` is still populated, only throwing the error is suppressed.

BTW: On my machine, `org.mapstruct.ap.test.bugs._2544.Issue2544Test` fails when I run all tests, but this hasn't changed.

Fixes #2560

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -1514,13 +1514,17 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             if ( mappingReferences.isForForgedMethods() ) {
                 return ReportingPolicyGem.IGNORE;
             }
+            // If we have ignoreByDefault = true, unprocessed source properties are not an issue.
+            if ( method.getOptions().getBeanMapping().isignoreByDefault() ) {
+                return ReportingPolicyGem.IGNORE;
+            }
             return method.getOptions().getMapper().unmappedSourcePolicy();
         }
 
         private void reportErrorForUnmappedSourcePropertiesIfRequired() {
             ReportingPolicyGem unmappedSourcePolicy = getUnmappedSourcePolicy();
 
-            if ( !unprocessedSourceProperties.isEmpty() && unmappedSourcePolicy.requiresReport() ) {
+            if ( !unprocessedSourceProperties.isEmpty() && unmappedSourcePolicy.requiresReport()) {
 
                 Message msg = unmappedSourcePolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
                     Message.BEANMAPPING_UNMAPPED_SOURCES_ERROR : Message.BEANMAPPING_UNMAPPED_SOURCES_WARNING;
--- a/processor/src/test/java/org/mapstruct/ap/test/ignorebydefaultsource/IgnoreByDefaultSourcesTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/ignorebydefaultsource/IgnoreByDefaultSourcesTest.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.ignorebydefaultsource;
+
+import javax.tools.Diagnostic.Kind;
+
+import org.mapstruct.ap.testutil.ProcessorTest;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.compilation.annotation.CompilationResult;
+import org.mapstruct.ap.testutil.compilation.annotation.Diagnostic;
+import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutcome;
+
+public class IgnoreByDefaultSourcesTest {
+
+    @ProcessorTest
+    @WithClasses({ SourceTargetMapper.class, Source.class, Target.class })
+    public void shouldSucceed() {
+    }
+
+    @ProcessorTest
+    @WithClasses({ ErroneousSourceTargetMapper.class, Source.class, Target.class })
+    @ExpectedCompilationOutcome(
+        value = CompilationResult.FAILED,
+        diagnostics = {
+            @Diagnostic(type = ErroneousSourceTargetMapper.class,
+                kind = Kind.ERROR,
+                line = 20,
+                message = "Unmapped source property: \"other\".")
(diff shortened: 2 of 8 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1520,7 +1520,9 @@ private ReportingPolicyGem getUnmappedSourcePolicy() {
         private void reportErrorForUnmappedSourcePropertiesIfRequired() {
             ReportingPolicyGem unmappedSourcePolicy = getUnmappedSourcePolicy();
 
-            if ( !unprocessedSourceProperties.isEmpty() && unmappedSourcePolicy.requiresReport() ) {
+            // If we have ignoreByDefault = true, unprocessed source properties are not an issue.
+            if ( !method.getOptions().getBeanMapping().isignoreByDefault() && !unprocessedSourceProperties.isEmpty()
+                && unmappedSourcePolicy.requiresReport()) {
```
filiphr, line 1525: Let's move this logic in `getUnmappedSourcePolicy()`.  We already have something similar there for forged methods

processor/src/test/java/org/mapstruct/ap/test/ignorebydefaultsource/IgnoreByDefaultSourcesTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.compilation.annotation.CompilationResult;
+import org.mapstruct.ap.testutil.compilation.annotation.Diagnostic;
+import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutcome;
+
+public class IgnoreByDefaultSourcesTest {
+
+    @ProcessorTest
+    @WithClasses({ SourceTargetMapper.class })
+    @ExpectedCompilationOutcome(value = CompilationResult.SUCCEEDED)
```
filiphr, line 20: This isn't needed, since if the compilation fails the test will fail as well

filiphr (changes requested):
Thanks a lot @Zegveld.

>I implemented this at the "check" level - `unprocessedSourceProperties` is still populated, only throwing the error is suppressed.

That's what I expected as well. Otherwise, it is a bit more difficult, we do the same for the target.

>BTW: On my machine, `org.mapstruct.ap.test.bugs._2544.Issue2544Test` fails when I run all tests, but this hasn't changed.

Interesting, that seems to test some formatting. So could be related to Windows and Java versions. If you have time it would be good to analyze it on your machine. What kind of OS do you have? Which Java / JDK do you use?


By the way the tests are failing because there is no `WIthClasses` for the `Source` and `Target` that you've added.

I would suggest adding them on top of the test class.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
incaseoftrouble: Ok, will do!

incaseoftrouble:
> By the way the tests are failing because there is no WIthClasses for the Source and Target that you've added.

Yeah, I hastily moved those out of the mapper classes and then was busy doing something else :-D

incaseoftrouble:
> Interesting, that seems to test some formatting. So could be related to Windows and Java versions. If you have time it would be good to analyze it on your machine. What kind of OS do you have? Which Java / JDK do you use?

This run is on Windows 10 build 19042.1165, Java `openjdk version "16.0.1" 2021-04-20`. I run the tests inside IntelliJ, the mentioned java is not the default on my machine. I'll try to pin it down.

incaseoftrouble:
I found the issue :-D Its a locale thing. I have english locale but german formatting settings

filiphr:
>I found the issue :-D Its a locale thing. I have english locale but german formatting settings.

Nice, combinations like this always show up interesting bugs. I have a mix of English and Swiss, so know the problems.
Can you try using `@DefaultLocale` from JUnit Pioneer, have a look at `NumberFormatConversionTest` as an example. If that fixes it for you please provide a PR for it so it is stable.

processor/src/test/java/org/mapstruct/ap/test/ignorebydefaultsource/Source.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.ignorebydefaultsource;
+
+class Source {
```
filiphr, line 8: Interesting that having package protected works with our test harness. Good to know this 😄

filiphr:
Back to the original topic of this PR. Great work @incaseoftrouble, thanks for applying my suggestions. I integrated this manually in https://github.com/mapstruct/mapstruct/commit/9ed4e389f8328b5f62655c82119f05cb59c4a439

processor/src/test/java/org/mapstruct/ap/test/ignorebydefaultsource/Source.java
incaseoftrouble: AFAIK Junit 5 tests are even recommended to be package private! Then you don't have 1000 different `Source` and `Target` :-) (But this actually only happened by accident :-D)

filiphr: Yes we do have 1000 different ones currently 😄 . I know about the tests, and that's what we are trying to do now. 

I wanted to make the mappers as well, but that doesn't work nicely with our test harness, we have some special class loading to make all this work correctly.

Issue #2455, opened 2021-05-21
@MappingTarget does not allow update of contained beans


pdavie:
I am working on a project (using Hibernate) where I have a complex object hierarchy, for example, `User` and `Role` (where a given `User` has a  contained `Role`).  I want to be able to perform the following to change a user's role (make an HTTP PUT request in a REST API):
```
{
    "role": {
        "id": 25
    }
}
```
I have a DTO `UserWriteDto` which contains all updatable `User` properties, including a `RoleReference` which just contains the `id` of the `Role` to set for the `User` (so that other `Role` properties cannot be updated, such as the `Role` `name` and `description` if specified in the JSON).

In the attached ZIP, I have created a rough approximation of the problem with User = A and Role = B where the `User` object contains an instance of an (existing) `Role` object.  In the generated `AMapper.createModel` method, I see (in lines 34-36):
```
...
        if ( dto.b != null ) {
            a.b = bReferenceToB( dto.b );
        }
...
```

However, when I am using `@MappingTarget` in the generated `AMapper.updateModel` method (lines 53-58), I see:
```
...
        if ( dto.b != null ) {
            if ( a.b == null ) {
                a.b = bMapper.fromReference( dto.b );
            }
            bReferenceToB1( dto.b, a.b );
        }
...
```

There seems to be no way to prevent the qualification `if (a.b == null)` on the `a.b = bMapper.fromReference(dto.b)` statement in the generated `updateModel` method.  Note the intention of the `fromReference()` method is that it looks up the existing entity in the database and returns it.  I have not found any way of being able to suppress this unwanted qualification, essentially, I want the `updateModel` setting of `B` to exactly match the code generated for `createModel`, that is, the value of `b` is **always** overwritten (and let Hibernate worry about the details).

I have worked around this specific issue by having an `@AfterMapping` explicitly set the role on the user when it has been changed, but doing this for all bean instances across all of the mappers is a considerable effort.

[mapstruct-mapping-target.zip](https://github.com/mapstruct/mapstruct/files/6519640/mapstruct-mapping-target.zip)

PR #2587, opened 2021-09-17
22455

alija-abadzic:
remove null check on target otherwise field is never updated

#2455


filiphr:
I am closing this PR since it breaks the functionality of MapStruct. See my comment in #2455 to understand why that issue is not working as expected.

Issue #598, opened 2015-07-07
Diagnostics show up on non-Mapper types


agudian:
Suppose I have these types:

``` java
public interface BaseEntityMapper<E extends AbstractEntity, D extends AbstractDto> {
    E toEntity(D src);
    D toDto(E src);
}

@Mapper
public interface CarEntityMapper extends BaseEntityMapper<Car, CarDto> {
// nothing
}
```

If Car and CarDto map nicely, then `CarEntityMapperImpl` will be generated as one would expect and it works great. 

However, if they don't map nicely, e.g. due to unmapped target properties, the diagnostics are generated with the locations of the methods in BaseEntityMapper. If that type is located in an already created jar, the marker might not be visible in some IDEs or in javac.

I think that, generally, if we create a diagnostic for some mapper to generate, we should only use locations within that mapper. In this case, an "unmapped target property" could be located at "extends _BaseEntityMapper_".

WDYT?

gunnarmorling:
Yes, I think that makes sense. I didn't foresee that usage pattern, but it seems legit and as you say all warnings should be raised at the `@Mapper`-annotated class. The warning would have to provide some more context though, e.g. the name of the affected method.

Zegveld:
Thought I'd take my attempt at creating a fix for this issue. 

My idea (which I'm already implementing as you might have noticed by the commit above) is to pass both the MapperElement containing the sourceMethod and the sourceMethodElement itself to the FormattingMessager. There the FormattingMessager will check if the MapperElement contains the sourceMethodElement.
If it does then it will just report it as it currently already does.
If it does not then it will report it as coming from the mapper. Still need to add a bit of message text to display which method it actually originated from.
Currently thinking of adding something like `Error occured at '<Method>' located in '<MethodClass>'.` at the end of the original message.

Also it allows the use of null for the mapper element, because currently I've only adjusted the errors that appear at Method level. This change might impact other PR's because it changes the FormattingMessager interface.

----------------------

A larger example to explain the requested feature: (and a check to see if I understood it correctly)
```
@Mapper
interface Mapper extends BaseMapper<SourceClass, TargetClass> {
    public TargetClass myMap(SourceClass src);
}

interface BaseMapper<S, T> {
    public T map(S s);
}

class SourceClass {
    public String property;
}

class TargetClass {
    public Object property;
}
```
[...]

PR #2608, opened 2021-10-10
#598 superclass mapping methods now throw errors/warnings on mapper class.

Zegveld:
Here is the first (and maybe final) draft of the PR for issue #598. I've made adjustments to the FormattingMessager interface, so this PR will most likely have impact on other PR's that introduce error/warning/info messages during compilation.

FormattingMessager.printMessage now expects 2 elements instead of 1 (those without elements are unaffected):
1. The mapper element (can be null) 
2. The element on which the error triggered (may not be null)

### DIFF
--- a/core/src/main/java/org/mapstruct/BeanMapping.java
+++ b/core/src/main/java/org/mapstruct/BeanMapping.java
@@ -14,6 +14,7 @@ import java.lang.annotation.Target;
 import org.mapstruct.control.MappingControl;
 
 import static org.mapstruct.NullValueCheckStrategy.ON_IMPLICIT_CONVERSION;
+import static org.mapstruct.SubclassExhaustiveStrategy.COMPILE_ERROR;
 
 /**
  * Configures the mapping between two bean types.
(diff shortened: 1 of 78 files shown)


filiphr:
Thanks for the PR @Zegveld. 

I like the idea. However, I am not to happy with the fact that we have to expose this on the `FormattingMessager` itself. I would propose an alternative solution. Instead of doing this on the caller side we could do it by creating a new instance of the `FormattingMessage`. This can be done in the `MapperCreationProcessor` by creating a new instance using the `mapperTypeElement` which is actually the `@Mapper` for which an implementation is being created at the moment and I guess is the one in the `sourceMethod` for you.

Regarding

>when trying to add the message I get different results depending on the compiler:
>jdk: `Occured at 'map(SOURCE)' in 'AbstractMapper'.`
>eclipse: `Occured at 'public abstract TARGET map(SOURCE) ' in 'AbstractMapper'.`

Instead of testing it like that I would rather cast to `Executable` element (if the element is a method) and provide a consistent message from our side.

PR #2625, opened 2021-10-24
#2005 Parameter type should only be checked if we are mapping from a single argument source

### LINKED ISSUE #2005: Getting incorrect "iterable type to non-iterable type" error
Given those three objects :

```java
public class MyDto {

    String name;
    byte[] file;
    int age;

// Getters and setters
}
```
```java
public class MyModel {

    String name;
    byte[] file;
    int age;

// Getters and setters
}
```
```java
public class MyModelWithoutByteArray {

    String name;
    int age;

// Getters and setters
}
```

And the following mapper interface :

```java
@Mapper
public interface DtoMapper {

    MyDto mapToDto(MyModel myModel);

    @Mapping(source = "file", target = "file")
    MyDto mapToDto(byte[] file, MyModelWithoutByteArray myModel);
}
```
The first method produces the following code :
```java
 @Override
    public MyDto mapToDto(MyModel myModel) {
        if ( myModel == null ) {
            return null;
        }

        MyDto myDto = new MyDto();

        myDto.setName( myModel.getName() );
        byte[] file = myModel.getFile();
        if ( file != null ) {
            myDto.setFile( Arrays.copyOf( file, file.length ) );
        }
        myDto.setAge( myModel.getAge() );

        return myDto;
    }
```

whereas, when trying to generate the second one, I get the following error :
```
Error:(15,11) java: Can't generate mapping method from iterable type to non-iterable type.
```

Here's the current configuration in my POM.xml :
```xml
	       <dependency>
			<version>1.3.1.Final</version>
		</dependency>

....

             <plugins>
                  <plugin>
				<version>3.8.0</version>
				<configuration>
					<release>11</release>
					<source>11</source>
					<target>11</target>
					<annotationProcessorPaths>
						<path>
							<version>1.3.1.Final</version>
						</path>
					</annotationProcessorPaths>
					<!-- other annotation processors -->
					<compilerArgs>
						<compilerArg>
							-Amapstruct.defaultComponentModel=spring
						</compilerArg>
					</compilerArgs>
				</configuration>
			</plugin>
		</plugins>

```

filiphr:
Fixes #2005

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
@@ -512,25 +512,39 @@ public class MethodRetrievalProcessor implements ModelElementProcessor<Void, Lis
                 return false;
         }
 
-        Type parameterType = sourceParameters.get( 0 ).getType();
+        if ( sourceParameters.size() == 1 ) {
+            Type parameterType = sourceParameters.get( 0 ).getType();
 
-        if ( isStreamTypeOrIterableFromJavaStdLib( parameterType ) && !resultType.isIterableOrStreamType() ) {
-            messager.printMessage( method, Message.RETRIEVAL_ITERABLE_TO_NON_ITERABLE );
-            return false;
+            if ( isStreamTypeOrIterableFromJavaStdLib( parameterType ) && !resultType.isIterableOrStreamType() ) {
+                messager.printMessage( method, Message.RETRIEVAL_ITERABLE_TO_NON_ITERABLE );
+                return false;
+            }
+
+            if ( !parameterType.isIterableOrStreamType() && isStreamTypeOrIterableFromJavaStdLib( resultType ) ) {
+                messager.printMessage( method, Message.RETRIEVAL_NON_ITERABLE_TO_ITERABLE );
+                return false;
+            }
+
+            if ( parameterType.isPrimitive() ) {
+                messager.printMessage( method, Message.RETRIEVAL_PRIMITIVE_PARAMETER );
+                return false;
+            }
+
+            for ( Type typeParameter : parameterType.getTypeParameters() ) {
+                if ( typeParameter.hasSuperBound() ) {
+                    messager.printMessage( method, Message.RETRIEVAL_WILDCARD_SUPER_BOUND_SOURCE );
+                    return false;
+                }
+
+                if ( typeParameter.isTypeVar() ) {
+                    messager.printMessage( method, Message.RETRIEVAL_TYPE_VAR_SOURCE );
+                    return false;
+                }
+            }
         }
 
         if ( containsTargetTypeParameter ) {
             messager.printMessage( method, Message.RETRIEVAL_MAPPING_HAS_TARGET_TYPE_PARAMETER );
-            return false;
-        }
-
-        if ( !parameterType.isIterableOrStreamType() && isStreamTypeOrIterableFromJavaStdLib( resultType ) ) {
-            messager.printMessage( method, Message.RETRIEVAL_NON_ITERABLE_TO_ITERABLE );
-            return false;
-        }
-
-        if ( parameterType.isPrimitive() ) {
-            messager.printMessage( method, Message.RETRIEVAL_PRIMITIVE_PARAMETER );
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
```diff
@@ -512,9 +513,11 @@ private boolean checkParameterAndReturnType(ExecutableElement method, List<Param
                 return false;
         }
 
-        Type parameterType = sourceParameters.get( 0 ).getType();
+        Optional<Type> parameterType =
```
sjaakd, line 516: I rember that we had discussions some time ago  on whether you should use optionals this way 😄 . 

Anyway, woudn't it be better to wrap line 516 to 567 in an 'if' statement like this:

```java
if souceParameters.size()==1){
 : 

}
```
to make it more explicit?

sjaakd: Zooming further out, this seams to be very much geared towards checking certain types of mapping methods. I'm thinking: should that be checked here really? Or at the moment when we really try to do the iterable-to-iterable mapping in the 'MappingCreationProcessor' ?  I guess it's some kind of fail-fast. 

And just curious for that matter.. what happens if we combine a bean with an interable as both source? Is it than automatically recognized as bean-mapping?

processor/src/test/java/org/mapstruct/ap/test/multisource/MultiSourceMapper.java
```diff
@@ -0,0 +1,67 @@
...
+@Mapper(unmappedTargetPolicy = ReportingPolicy.IGNORE)
+public interface MultiSourceMapper {
+
+    MultiSourceMapper INSTANCE = Mappers.getMapper( MultiSourceMapper.class );
+
+    Target mapFromPrimitiveAndCollection(int value, Collection<String> elements);
+
+    Target mapFromCollectionAndPrimitive(Collection<String> elements, int value);
+
+    class Source {
```
sjaakd, line 26: Source is never used, or is it?

processor/src/test/java/org/mapstruct/ap/test/multisource/MultiSourceMapperTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.ProcessorTest;
+import org.mapstruct.ap.testutil.WithClasses;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Filip Hrisafov
+ */
+public class MultiSourceMapperTest {
```
sjaakd, line 19: Is it really reproducing the problem as described on #2005? That's a combination of bean, array right?

sjaakd (changes requested):

processor/src/test/java/org/mapstruct/ap/test/multisource/MultiSourceMapper.java
filiphr: No it isn't I'll remove it.

processor/src/test/java/org/mapstruct/ap/test/multisource/MultiSourceMapperTest.java
filiphr: It goes a step further. I started like in #2005. However, I realised that it was a more general multi source parameter mapping. We also have similar problems when the first parameter is a primitive

processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
filiphr: haha, you are right. We had a discussion about the optionals used in this way. I'll try to extract something better.

And yes this is done here for fail fast. I don't want to start moving things around as it might make some things more complicated 😄.

>And just curious for that matter.. what happens if we combine a bean with an interable as both source? Is it than automatically recognized as bean-mapping?

Yes it will be recognized as bean mapping. The other types of mappings are mappings with only a single source parameter.

filiphr:
Thanks for the review @sjaakd. I have done the changes you requested. Let me know what you think

sjaakd (approved):

Issue #2553, opened 2021-08-17
New mapping feature not support property paths


jonsalvas:
First of all: thanks for the great job you are doing with this library, it helps us to save a lot of boilerplate code. I was really happy that you finally released the feature to map maps to beans: https://mapstruct.org/documentation/dev/reference/html/#mapping-map-to-bean 
As we use a lot of maps in our project it will be very helpful.

However, I was not really able to make it work in our project so I tried to build a minimal example based on one of your test cases (MapToBeanFromMultipleSource) in the PR. There I noticed that, if the map is a field of a bean the property path in the source parameter is not resolved correctly. Minimized example code:

```
import org.mapstruct.Mapper;
import org.mapstruct.Mapping;

import java.util.Map;

@Mapper
public interface MapToBeanFromMultipleSources {
    @Mapping(target = "string", source = "bean.strings.string")
    Target toTarget(Source bean);

    class Source {
        private Map<String, String> strings;

        public Map<String, String> getStrings() {
            return strings;
        }

        public void setStrings(Map<String, String> strings) {
            this.strings = strings;
        }

    }

    class Target {

        private String string;


        public String getString() {
            return string;
        }

        public void setString(String string) {
            this.string = string;
        }

    }
}
```

I am receiving: "The type of parameter "bean" has no property named "strings.string". Am I missing something, or is this currently not supported?

filiphr:
Thanks for reporting this @jonsalvas.

This is slightly related to #2549. Seems like we no longer allow using the single source parameter name in `Mapping#source` when the parameter is `Map<String, ???>`.

Which means that something like 

```
@Mapping(target = "string", source = "strings.string")
```

should work. However, it doesn't. We will look into this so it works transparently for defined nested mappings. If we add a defined mapping to [`MapToBeanFromMapAndNestedMap`](https://github.com/mapstruct/mapstruct/blob/master/processor/src/test/java/org/mapstruct/ap/test/frommap/MapToBeanFromMapAndNestedMap.java) then it won't work.

We will fix this

PR #2626, opened 2021-10-24
#2624 Nested target methods should be inherited for forged Map to Bean methods

### LINKED ISSUE #2624: Cannot map nested target when mapping from Map to Bean
```
public class Employee {
    private String id;
    private String name;
    private Department department
}


public class Department {
    private String id;
    private String name;
}
```

And the mapper

```
@Mapper
public interface MapToBeanMapper {

    @Mapping(target = "department.id", source = "did")
    @Mapping(target = "department.name", source = "dname")
    Employee fromMap(Map<String, String> map);
}
```

Will not generate the correct code.

filiphr:
Fixes #2624

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2624/Issue2624Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2624/Issue2624Mapper.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2624;
+
+import java.util.Map;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper
+public interface Issue2624Mapper {
+
+    Issue2624Mapper INSTANCE = Mappers.getMapper( Issue2624Mapper.class );
+
+    @Mapping( target = "department.id", source = "did")
+    @Mapping( target = "department.name", source = "dname")
+    Employee fromMap(Map<String, String> source);
+
+    class Employee {
+        private String id;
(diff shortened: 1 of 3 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_2624/Issue2624Mapper.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper
+public interface Issue2624Mapper {
+
+    Issue2624Mapper INSTANCE = Mappers.getMapper( Issue2624Mapper.class );
+
+    @Mapping( target = "department.id", source = "did")
```
sjaakd, line 22: nice.. Just a thought.. you can't map nested sources, right? Since we don't know the type of the 'value' in the map if it's a `Map<String, Object>`.. 

I'm not sure how explicit the documentation is in this area (I remember that we discussed having a `sourceType` in `@Mapping`). And especially with reverting mappings (later on), this might to start to play a role. So also, perhaps we need to extend the documentation in this area.

I don't dare to ask 😄 .. but what happens if you specify a `@Mapping(target = ".", source = "did")`?

sjaakd (approved):

processor/src/test/java/org/mapstruct/ap/test/bugs/_2624/Issue2624Mapper.java
filiphr: >nice.. Just a thought.. you can't map nested sources, right? Since we don't know the type of the 'value' in the map if it's a `Map<String, Object>`..

No we can't, that has already been reported in #2553

>I don't dare to ask 😄 .. but what happens if you specify a `@Mapping(target = ".", source = "did")`?

You just did 😄 . I haven't tested this, but it will try to map whatever type the map value is to the current target. I don't really see a use case for this, so I'll rather wait for someone to complain about it 😄

sjaakd: by all means 👍

PR #2632, opened 2021-10-30
#1752 Always return mapping target when using update methods with return

### LINKED ISSUE #1752: MapStruct should return the target object(which marking with @MappingTarget) when the source object is null
The generated code as follow:

    public TbsSpecification yjlGoodsAttribute2Spec(YjlGoodsAttribute yjlGoodsAttribute, TbsSpecification tbsSpecification) {
        if ( yjlGoodsAttribute == null ) {
            return null;
        }

        tbsSpecification.setModifiedTime( java.time.LocalDateTime.now() );

        return tbsSpecification;
    }

when yjlGoodsAttribute is null, the method should be return tbsSpecification.

filiphr:
Fixes #1752

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
@@ -20,7 +20,7 @@
     </#list>
     <#if !mapNullToDefault>
     if ( <#list sourceParametersExcludingPrimitives as sourceParam>${sourceParam.name} == null<#if sourceParam_has_next> && </#if></#list> ) {
-        return<#if returnType.name != "void"> null</#if>;
+        return<#if returnType.name != "void"> <#if existingInstanceMapping>${resultName}<#if finalizerMethod??>.<@includeModel object=finalizerMethod /></#if><#else>null</#if></#if>;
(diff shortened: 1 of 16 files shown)


sjaakd (changes requested):
I guess we need some spent some words in documentation as well.

For the remainder LGTM 👍

filiphr:
>I guess we need some spent some words in documentation as well.

I checked the docs and we have nothing about this type of update methods. We have something in `MappingTarget` that only provides an example for "Update exist bean and return it".

I'll add something to the release notes though.

sjaakd:
> I checked the docs and we have nothing about this type of update methods. We have something in `MappingTarget` that only provides an example for "Update exist bean and return it".
> 
> I'll add something to the release notes though.

That's fine.. But _should_ we mention these kind of methods in the documentation.? (It's more a completeness question)

filiphr:
>That's fine.. But _should_ we mention these kind of methods in the documentation.? (It's more a completeness question)

Seems I was wrong. We do have something in the docs:

>Instead of `void` you may also set the method's return type to the type of the target parameter, which will cause the generated implementation to update the passed mapping target and return it as well

I think that this is fine and that we do not need any change, since what we say in the docs is what we are doing now as well.

PR #2633, opened 2021-10-30
#2225 Add support for suppressing the generation of the timestamp through Mapper and MapperConfig

### LINKED ISSUE #2225: Expose suppressGeneratorTimestamp in @Mapper annotation
Right now to enable reproducible build timestamp generation must be disabled. This can be done only via compiler argument. But this generates below warning if some project do not declare any MapStruct mappers.

```raw
The following options were not recognized by any processor: '[mapstruct.suppressGeneratorTimestamp]'
```

This can easily happen if having MapStruct configuration done once for many projects, like when using parent pom in Maven multi-module project. Typically then all projects use the same parent pom, but not all declare MapStruct mappers. In such case `javac` generates above warning, and it can't be suppressed.

So to avoid this either:

- generation of timestamp should be disabled by default

  Such approach was already suggested in #1682, but this was closed. IMHO timestamp does not bring any value, but only introduce issues: like with reproducible builds, Gradle build caching (#2194) or in any other case when there is no mapper declared (like in #1638), so disabling it by default will make better user experience.

- add ability to set `suppressGeneratorTimestamp` this on `@Mapper` annotation - then it will not be possible to set it globally once, like in the Maven parent pom, but at least this warning, will go away

filiphr:
Fixes #2225

### DIFF
--- a/core/src/main/java/org/mapstruct/Mapper.java
+++ b/core/src/main/java/org/mapstruct/Mapper.java
@@ -359,14 +359,15 @@ public @interface Mapper {
     Class<? extends Exception> unexpectedValueMappingException() default IllegalArgumentException.class;
 
     /**
-     * Flag indicating whether the creation of a time stamp in the {@code @Generated} annotation should be suppressed.
-     * i.e. not be created.
+     * Flag indicating whether the addition of a time stamp in the {@code @Generated} annotation should be suppressed.
+     * i.e. not be added.
      *
      * The method overrides the flag set in a central configuration set by {@link #config()}
+     * or through an annotation processor option.
      *
-     * @return whether the creation of a timestamp should be suppressed
+     * @return whether the addition of a timestamp should be suppressed
      *
      * @since 1.5
      */
-    boolean suppressGeneratorTimestamp() default false;
+    boolean suppressTimestampInGenerated() default false;
 }
--- a/core/src/main/java/org/mapstruct/Mapper.java
+++ b/core/src/main/java/org/mapstruct/Mapper.java
@@ -357,4 +357,16 @@ public @interface Mapper {
      * @since 1.4
      */
     Class<? extends Exception> unexpectedValueMappingException() default IllegalArgumentException.class;
+
+    /**
+     * Flag indicating whether the creation of a time stamp in the {@code @Generated} annotation should be suppressed.
+     * i.e. not be created.
+     *
+     * The method overrides the flag set in a central configuration set by {@link #config()}
+     *
+     * @return whether the creation of a timestamp should be suppressed
+     *
+     * @since 1.5
+     */
+    boolean suppressGeneratorTimestamp() default false;
 }
--- a/core/src/main/java/org/mapstruct/MapperConfig.java
+++ b/core/src/main/java/org/mapstruct/MapperConfig.java
@@ -330,13 +330,15 @@ public @interface MapperConfig {
     Class<? extends Exception> unexpectedValueMappingException() default IllegalArgumentException.class;
 
     /**
-     * Flag indicating whether the creation of a time stamp in the {@code @Generated} annotation should be suppressed.
-     * i.e. not be created.
+     * Flag indicating whether the addition of a time stamp in the {@code @Generated} annotation should be suppressed.
+     * i.e. not be added.
      *
-     * @return whether the creation of a timestamp should be suppressed
+     * The method overrides the flag set through an annotation processor option.
+     *
+     * @return whether the addition of a timestamp should be suppressed
      *
      * @since 1.5
      */
-    boolean suppressGeneratorTimestamp() default false;
+    boolean suppressTimestampInGenerated() default false;
 }
 
--- a/core/src/main/java/org/mapstruct/MapperConfig.java
+++ b/core/src/main/java/org/mapstruct/MapperConfig.java
@@ -329,5 +329,14 @@ public @interface MapperConfig {
      */
     Class<? extends Exception> unexpectedValueMappingException() default IllegalArgumentException.class;
 
+    /**
+     * Flag indicating whether the creation of a time stamp in the {@code @Generated} annotation should be suppressed.
+     * i.e. not be created.
+     *
+     * @return whether the creation of a timestamp should be suppressed
+     *
+     * @since 1.5
+     */
+    boolean suppressGeneratorTimestamp() default false;
 }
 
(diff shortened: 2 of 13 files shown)


processor/src/test/java/org/mapstruct/ap/test/versioninfo/SuppressTimestampViaMapper.java
```diff
@@ -0,0 +1,16 @@
...
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.versioninfo;
+
+import org.mapstruct.Mapper;
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper(suppressGeneratorTimestamp = true)
```
sjaakd, line 13: I guess, we our fixture checker somehow ignores the TS as well? That code can probably be removed, right?

core/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -357,4 +357,16 @@ NullValuePropertyMappingStrategy nullValuePropertyMappingStrategy() default
...
+     * Flag indicating whether the creation of a time stamp in the {@code @Generated} annotation should be suppressed.
+     * i.e. not be created.
+     *
+     * The method overrides the flag set in a central configuration set by {@link #config()}
+     *
+     * @return whether the creation of a timestamp should be suppressed
+     *
+     * @since 1.5
+     */
+    boolean suppressGeneratorTimestamp() default false;
```
sjaakd, line 371: `suppressGeneratorTimestamp()` .. I guess `suppressGenerationTimestamp()` is better. It's the time-stamp of 'the act of' generation, right?

core/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -357,4 +357,16 @@ NullValuePropertyMappingStrategy nullValuePropertyMappingStrategy() default
      * @since 1.4
      */
     Class<? extends Exception> unexpectedValueMappingException() default IllegalArgumentException.class;
+
+    /**
+     * Flag indicating whether the creation of a time stamp in the {@code @Generated} annotation should be suppressed.
```
sjaakd, line 362: creation == addition?

core/src/main/java/org/mapstruct/MapperConfig.java
```diff
@@ -329,5 +329,14 @@ MappingInheritanceStrategy mappingInheritanceStrategy()
      */
     Class<? extends Exception> unexpectedValueMappingException() default IllegalArgumentException.class;
 
+    /**
+     * Flag indicating whether the creation of a time stamp in the {@code @Generated} annotation should be suppressed.
```
sjaakd, line 333: creation = addtion?

core/src/main/java/org/mapstruct/Mapper.java
```diff
@@ -357,4 +357,16 @@ NullValuePropertyMappingStrategy nullValuePropertyMappingStrategy() default
...
+     * i.e. not be created.
+     *
+     * The method overrides the flag set in a central configuration set by {@link #config()}
```
sjaakd, line 365: mention link to `MapperConfig` annotation..?

core/src/main/java/org/mapstruct/MapperConfig.java
sjaakd: Mention somehow this overrides the processor setting (in the pom.xml in case of maven)

sjaakd (changes requested):
Perhaps also some words in the documentation..

processor/src/test/java/org/mapstruct/ap/test/versioninfo/SuppressTimestampViaMapper.java
filiphr: Yes our fixture checker ignores the TS and the comment, because they are different each time and also they are different on different compilers.

Which code are you referring to? The one that does the ignoring? We still need to ignore the comment though.

core/src/main/java/org/mapstruct/Mapper.java
filiphr: I used the same property as the compiler option. If we are going for something else then perhaps `suppressTimestampGeneration` is better, at least it reads better to me than `suppressGenerationTimestamp`.

core/src/main/java/org/mapstruct/MapperConfig.java
filiphr: Good point. I'll change it and mention the override of the processor setting.

core/src/main/java/org/mapstruct/Mapper.java
filiphr: Not following. This is identical to the other places. Check `componentModel`

processor/src/test/java/org/mapstruct/ap/test/versioninfo/SuppressTimestampViaMapper.java
sjaakd: agreed. but not the TS.. I did not search the code.. But logic demands there would be code  that's irrelevant. Anyway.. its not a big deal because removing this would also require updating all fixtures.

core/src/main/java/org/mapstruct/Mapper.java
sjaakd: ok.. leave as is..

core/src/main/java/org/mapstruct/Mapper.java
sjaakd: ok.. leave as is.

processor/src/test/java/org/mapstruct/ap/test/versioninfo/SuppressTimestampViaMapper.java
sjaakd: `JavaFileAssert#ignoreDelta`

filiphr: It means updating all fixtures and all our tests to actually suppress the timestamp generation

sjaakd: That's what I meant in my comment 12 min ago.. 😉

sjaakd (approved):

filiphr:
Thanks for the review @sjaakd

Issue #2606, opened 2021-10-04
From Date to LocalDate loses one day


UnleashSpirit:
Hi we have a lost of one day whenconverting from Date to LocalDate
I search over old issues but don't find really accurate thing

here the code
```java
import java.time.LocalDate;
class A {
  private LocalDate birthDate;
}

import java.util.Date;
class B {
  private Date naissance;
}

@Mapping(target="birthDate", source="birth")
A toA(B b);

// Generate
if ( b.getBirth() != null ) {
    a.birthDate( LocalDateTime.ofInstant( b.getBirth().toInstant(), ZoneOffset.UTC ).toLocalDate() );
}
```

The problem with this, is that if b.birth = 1980-10-01, the a.birthDate become 1980-09-30 (loses one day)
Some traces : 
```java
// 1980-09-30T23:00:00Z
System.out.println(b.getBirth().toInstant());
// 1980-09-30T23:00
System.out.println(LocalDateTime.ofInstant( b.getBirth().toInstant(), ZoneOffset.UTC ));
// 1980-09-30
System.out.println(LocalDateTime.ofInstant( b.getBirth().toInstant(), ZoneOffset.UTC ).toLocalDate());
```

Is there a way to have a correct conversion, without specifying the expressoion=java(...) myself ?
This works
```java
a.birthDate( LocalDateTime.ofInstant( b.getBirth().toInstant(), ZoneOffset.systemDefault() ).toLocalDate() );
```
Or "without" using LocalDateTime (why using Time for date only ?)
```java
a.birthDate( b.getBirth().toInstant().atZone(ZoneId.systemDefault()).toLocalDate());
```

lostiniceland:
Maybe Mapstruct could make use of `java.time.Clock` which is passed to the Mapper.

For CDI one could then provide an Bean which has the prefered Clock
```
@DefaultBean
@ApplicationScoped
public Clock systemClock() {
    return Clock.systemDefaultZone(); // or with some zone, etc
}
```

UnleashSpirit:
I like the idea of using a Clock bean !

We used it in every "now" cause we need to override the "real" now for our jUnit and Cucumber tests.
```java
LocalDate.now(clock);
OffsetDateTime.now(clock)
[...]
```

PR #2634, opened 2021-10-30
#598: Errors/Warnings now end up on the @Mapper annotated class.

Zegveld:
Introduced a decorator for the FormattingMessager which changes the element on which the error/warning is thrown as well as adding additional text to the error to show where the error/warning originated from.
This PR is for issue #598

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperAnnotatedFormattingMessenger.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperAnnotatedFormattingMessenger.java
@@ -5,6 +5,7 @@
  */
 package org.mapstruct.ap.internal.processor;
 
+import java.util.stream.Collectors;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
@@ -100,20 +101,18 @@ public class MapperAnnotatedFormattingMessenger implements FormattingMessager {
             method.append( " " );
             method.append( ee.getSimpleName() );
             method.append( "(" );
-            boolean first = true;
-            for ( VariableElement element : ee.getParameters() ) {
-                if ( !first ) {
-                    method.append( ", " );
-                }
-                method.append( typeUtils.asElement( element.asType() ).getSimpleName() );
-                method.append( " " );
-                method.append( element.getSimpleName() );
-                first = false;
-            }
+            method.append( ee.getParameters()
+                    .stream()
+                    .map( this::parameterToString )
+                    .collect( Collectors.joining( ", " ) ) );
             method.append( ")" );
             return method.toString();
         }
         return e.toString();
+    }
+
+    private String parameterToString(VariableElement element) {
+        return typeUtils.asElement( element.asType() ).getSimpleName() + " " + element.getSimpleName();
     }
 
     private Message determineDelegationMessage(Element e, Message msg) {
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperAnnotatedFormattingMessenger.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperAnnotatedFormattingMessenger.java
@@ -1,22 +1,38 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
 package org.mapstruct.ap.internal.processor;
 
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
+import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
+import javax.tools.Diagnostic.Kind;
 
 import org.mapstruct.ap.internal.util.FormattingMessager;
 import org.mapstruct.ap.internal.util.Message;
+import org.mapstruct.ap.internal.util.TypeUtils;
 
+/**
+ * Handles redirection of errors/warnings so that they're shown on the mapper instead of hidden on a superclass.
+ *
+ * @author Ben Zegveld
+ */
 public class MapperAnnotatedFormattingMessenger implements FormattingMessager {
 
     private FormattingMessager delegateMessager;
     private TypeElement mapperTypeElement;
-    private Element messageElement;
+    private TypeUtils typeUtils;
 
-    public MapperAnnotatedFormattingMessenger(FormattingMessager delegateMessager, TypeElement mapperTypeElement) {
+    public MapperAnnotatedFormattingMessenger(FormattingMessager delegateMessager, TypeElement mapperTypeElement,
+                                              TypeUtils typeUtils) {
         this.delegateMessager = delegateMessager;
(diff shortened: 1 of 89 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/MapperAnnotatedFormattingMessenger.java
```diff
@@ -0,0 +1,137 @@
...
+            boolean first = true;
+            for ( VariableElement element : ee.getParameters() ) {
+                if ( !first ) {
+                    method.append( ", " );
+                }
+                method.append( typeUtils.asElement( element.asType() ).getSimpleName() );
+                method.append( " " );
+                method.append( element.getSimpleName() );
+                first = false;
+            }
```
filiphr, line 112: We can change the loop with a stream and a collectors joining in theory. You could even add a method that would get the `VariableElement` and return its simple name type and then use a method reference.
e.g.

```suggestion
            method.append( ee.getParameters()
                    .stream()
                    .map( element -> typeUtils.asElement( element.asType() ).getSimpleName() )
                    .collect( Collectors.joining( ", " ) ) );
```

I think that this makes the code slightly clearer to read

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperAnnotatedFormattingMessenger.java
```diff
@@ -0,0 +1,137 @@
...
+    public boolean isErroneous() {
+        return delegateMessager.isErroneous();
+    }
+
+    private Object[] determineDelegationArguments(Element e, Message msg, Object[] args) {
+        if ( methodInMapperClass( e ) ) {
+            return args;
+        }
+        String originalMessage = String.format( msg.getDescription(), args );
+        return new Object[] { originalMessage, constructMethod( e ), e.getEnclosingElement().getSimpleName() };
```
filiphr, line 88: Instead of using `getSimpleName` I would suggest to use the `getQualifiedName` from `QualifiedNameable` if the `e.getEnclosingElement()` returns a `QualifiedNameable` (which it most likely does always here). 

I think this would make it clear in which type this happened.

filiphr (commented):
This looks nice and clean @Zegveld. I've left two comments, not sure if you agree with them or not.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperAnnotatedFormattingMessenger.java
Zegveld: Well you did find the one bit of code that I was not that happy with ;)
The method is too long in my opinion, but then again we have a lot of methods in mapstruct that are longer. I don't mind changing it into the streaming API.

The next question here is what do we want as the method text, as your suggestion here strips the parameter names away.
I'm fine either way, but I believe it to be more clear to report with:
`<ReturnType> <MethodName>(<Param1Type> <Param1Name>, <Param2Type> <Param2Name>...).`
But leaving the parameter names behind reduces it to the actual signature (except that I add the return type as well ;) ) :
`<ReturnType> <MethodName>(<Param1Type>, <Param2Type>...)`

What do you think is better here, @filiphr ?

filiphr: I completely missed the parameter name. My suggestion is wrong then, you should include it. My main suggestion was for the style of the method, i.e. use streaming / StringJoiner instead of the loop, it makes it easier to read it.

I personally do not mind the length that much. We do try to extract smaller methods where it makes sense, but it doesn't work always 😀

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperAnnotatedFormattingMessenger.java
Zegveld: Having a qualified name is not really necessary here I think. You are referencing that class from the code there anyway, since you are either extending or using it.
I also believe it will obscure the source because of message length if we do it like that.

For example:
`This is the original message with an error. Occured at TargetType mapMethod(SourceType source) in GeneralMapper.`
vs
`This is the original message with an error. Occured at TargetType mapMethod(SourceType source) in org.mapstruct.business.component.product.storage.domain.general.GeneralMapper.`
(used a location that has a similar structure as in one of the product that I work on)

It should be clear which `GeneralMapper` it is, because the error appears near the line:
`public abstract class MySubDomainMapper extends GeneralMapper`

Although if you have multiple layers of hierarchy you might not see it immediate, but you can just click into the parent class to find out who that one is extending:
`public abstract class MySubDomainMapper extends MyDomainMapper`
`public abstract class MyDomainMapper extends GeneralMapper`

I'm not sure if the uses mappers that are not annotated with `@Mapper` or `@Mapping` will keep their errors with this change. It might even be that those will get their errors redirected here, come to think of it. But since you are referencing them explicitly with uses you should be able to find them with just the class name.

If on the other hand it is standard behavior for compiler messages to refer to the qualified name instead, then I agree with that it should become the qualified name. I just don't have that much knowledge yet about standard behavior for compiler/annotation processing.

filiphr: You are absolutely right. Having the simple name here makes absolutely more sense. I didn't thought about the fact that you already have the name in the location of the main mapper. Let's leave it as is

Zegveld:
@filiphr , I've just pushed the changes. 

=========== Not part of this PR ============

Although my build just failed locally due to an error with date translation, but I believe that needs to be fixed in another PR, related to #2606. (Basically date translations shouldn't use date/time construct but date construct.)

```
[ERROR] Failures: 
[ERROR] org.mapstruct.ap.test.abstractclass.AbstractClassTest.shouldCreateImplementationOfMethodFromSuper
[ERROR]   Run 1: AbstractClassTest.shouldCreateImplementationOfMethodFromSuper:47->assertResult:62 
Expecting:
 <"Birthday: 25.04.1948">
to be equal to:
 <"Birthday: 26.04.1948">
but was not.
```
I think this is related to DLS usage within the timezone.

filiphr (approved):
Thanks for doing the changes @Zegveld. All looks good to me.

You can go ahead and do the merge. You can squash and merge from the GitHub UI directly.

filiphr:
Regarding the date related problem. It could be related to the way we are setting the time zone for some tests, but not for this one. I do not think that it is linked to that issue since we create a `Calendar` in the `Source` and then in the `SourceTargetMapper` we have a custom mapping for `Calendar` to `String`. Most likely something is messed up there. We could use an explicit timezone in the tests though

Zegveld:
> Regarding the date related problem. It could be related to the way we are setting the time zone for some tests, but not for this one. I do not think that it is linked to that issue since we create a `Calendar` in the `Source` and then in the `SourceTargetMapper` we have a custom mapping for `Calendar` to `String`. Most likely something is messed up there. We could use an explicit timezone in those tests though

Something to keep in mind if we see that test failing more often. It was now just once locally for me. The exact cause is unknown and we can not analyze it because it is not reproducible. If we get a reproducible issue for this then we can look further into it, not worth it to dive into this atm.

Guess I'll be doing my first PR-merge then. 😸  wish me luck 😹

filiphr:
>Guess I'll be doing my first PR-merge then. 😸 wish me luck 😹

From what I can see it all went smoothly. Congrats on your first merge 😉

Issue #2636, opened 2021-11-01
Issue with qualifiedByName used together with defaultValue


RHarryH:
- [x] Is this an issue (and hence not a question)? 

Hello,

Today I decided to upgrade `mapstruct` from `1.3.1.Final` to `1.4.2.Final` and my code stopped to work. I also chcecked `1.5.0.Beta1` and from my perspective the behavior is still wrong. For following mapping:

```java
@Mapping(source = "folder.ancestor.id", target= "parent", defaultValue = "#", qualifiedByName = "uuidToString")
DirectoryNode convert(Folder folder);

@Named("uuidToString")
default String uuidToString(UUID id) {
    return id.toString();
}
```
I'm getting following results:

`1.3.1.Final` **Expected behavior**
```java
if ( folderAncestorId( folder ) != null ) {
    directoryNode.setParent( uuidToString( folderAncestorId( folder ) ) );
}
else {
    directoryNode.setParent( "#" );
}
```
`1.4.2.Final`

The code generating fails with following error
`Can't map "#" to "String parent"`

`1.5.0.Beta1`
```java
UUID id = documentFolderId( document );
if ( id != null ) {
    directoryNode.setParent( uuidToString( id ) );
}
else {
    directoryNode.setParent( uuidToString( UUID.fromString( "#" ) ) );
}
```

`#` is not a correct `UUID` but from my perspective it is the correct indicator of directory root. I don't want to call my named method when applying the default value. I just want to set the value when `id` is `null`.

Zegveld:
Thank you for reporting this issue. I agree that it should indeed just take the default-value as specified instead of attempting to convert it if there is no conversion needed.

I've created a PR for it, and hope to have it fixed in the 1.5.0.Beta2 or if it does not make that one then the 1.5.0 release.

RHarryH:
Many thanks for such fast reaction.

filiphr:
This is a tricky one @RHarryH.

Lets take this example


```java
@Mapper
public interface MyMapper {

    @Mapping(target = "name", source = "firstName", defaultValue = "unknown", qualifiedByName = "uppercase")
    Customer map(CustomerDto dto);

    @Named("uppercase")
    default String uppercase(String value) {
        return value == null ? null : value.toUpperCase(Locale.ROOT);
    }
}
```

Here MapStruct will generate

```java
if ( dto.getFirstName() != null ) {
    customer.setFirstName( uppercase( dto.getFirstName() ) );
}
else {
[...]

PR #2637, opened 2021-11-02
#2636: defaultValue combined with qualified should not convert if not needed

Zegveld:
Qualified should have preference on usage for the normal values, but if a default value is set to use in case that the source is `null`. Then this default value should be used if it has the same type. 

It is best to assume that the user knows what he wants as the value. Therefor trying to map this back and forth is useless, seeing #2636  it even breaks expected functionality in certain cases.


processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -917,7 +917,11 @@ public PropertyMapping build() {
             );
 
             Assignment assignment = null;
-            if ( !targetType.isEnumType() ) {
+            if ( targetType.equals( sourceType ) ) { // No need to lookup a mapping.
```
sjaakd, line 920: I'm not sure here whether there are no side effects. I realise this is a constant. However, zooming out what do you do when the user explicitly wants to convert a constant or default value to a String? 

In other words: do we see a use case were a constant String is converted into another String? At first glance this sounds meaningless. But we do have context objects. Although, perhaps far sought..

So:
```java
@Mapper
 public interface Issue2636Mapper {
     @Mapping( source = "folder.ancestor.id", target = "parent", defaultValue = "#", qualifiedByName = "uuidToString" )
     DirectoryNode convert(Folder folder, @Context State someState);

     @Named( "uuidToString" )
     default String uuidToString(UUID id, @Context State someState) {
         // do something with state and
         return someState.doSomethingWithId(id);
     }

     @Named( "uuidToString" )
     default String defaultToString(String id, @Context State someState) {
         // do something with state and
         return someState.doSomethingWithId(id);
     }

 }
```
BTW: I'm curious to know it the above works without a context. Although I admit the example is far-sought and feels counter intuitive. 

What also bothers me a bit is that all the checks are currently in `MethodResolverImpl`. Here we do allow explicitly same-type mapping methods to override the "direct assignable" types. The check feels a bit out of place here. At the other hand: we don't have any knowledge from the calling side in `MethodResolverImpl` (so it does not it's a constant).

An argument for this solution is that it used to work. Apparently we "broke" something. 

@filiphr WDYT?

sjaakd (changes requested):
Let's discuss this a bit further.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
filiphr: This is a tricky one @sjaakd. It broke due to the fixes we did in #1714. If there are qualifiers then we do not use the direct assignment.

Your example is entirely OK @sjaakd, and it is even valid without the `@Context`. Currently, we will pick up the `defaultToString` method. With this PR that is not the case.

A more simpler example is something like:


```
@Mapper
public interface MyMapper {

    @Mapping(target = "name", source = "firstName", defaultValue = "unknown", qualifiedByName = "uppercase")
    Customer map(CustomerDto dto);

    @Named("uppercase")
    default String uppercase(String value) {
        return value == null ? null : value.toUpperCase(Locale.ROOT);
    }
}
```

Before this PR we would use the `uppercase` method for both, which IMO is correct. With this PR that is no longer the case.

I am more keen on not providing fix for this and telling the user what the actual problem is, since if we start going down this path there will be even more problems. We also have `defaultValueExpression` for using things literally.

Zegveld: Concerning that uppercase example, wouldn't it be weird that the `defaultValue` listed there is not in uppercase itself? The user specifies a `defaultValue`, but instead of that one being used it gets adjusted by the mapping method for the non-default values.

Also I find it weird that the behavior of `defaultValue` and `defaultExpression` differs. The only thing I would expect to be different with these is that one uses a `String` or `Primitive Data Type` input as a default value, while the other can handle all the other types using methods. Using your example of uppercase, wouldn't it be more logical to suggest here to use either `defaultValue="UNKNOWN"` or `defaultExpression=java(uppercase("unknown"))`?

On the other hand the example that @sjaakd gave is one I did not think of. That situation would make the `defaultValue` usage the way I thought of it break. Currently I end up with at least 2 other options to solve this situation:

1. Maybe an option is to leave it up to the user to decide how to handle these situations. We could introduce a setting in the `@MappingControl` to handle this. [The changes would look something like this.](https://github.com/Zegveld/mapstruct/compare/feature/2636...Zegveld:feature/2636_alternative) That alternative solution is not finished yet, it needs more situations for testing it. Including those with an `@Context` annotation and perhaps quite a bit more work.
2. Update the Reference Guide to mention that `defaultValue` and `constant` are also mapped through the qualified mappings if possible, and that if this is not wanted they should use the `defaultExpression` variant instead.

Perhaps the 2nd option is the better one, that we document it and if people want to influence it we can always add the 1st option upon request.

filiphr: My example was extremely simplified. The implementation logic can be anything, perhaps you have some kind of a mapping by key and you need to get an id from the DB, similar to what Sjaak showed, but not with a `@Context`.

I think that option 2 is the best here. Update the Reference Guide and also add something to the javadoc of those methods as well. Perhaps we can also add some tests for this, to verify that everything works like we expect it to (like it is now), since as we can see there are no tests failing currently 😄

Zegveld:
Updated this PR for documentation, reverted the change and added tests for three different cases:
* invalid use of `defaultValue` combined with qualifiers.
* Valid use of `defaultValue` combined with qualifiers.
* Use of `defaultExpression` combined with qualifiers.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: > Your example is entirely OK @sjaakd, and it is even valid without the `@Context`. Currently, we will pick up the `defaultToString` method. With this PR that is not the case.

The `@Context` was intended to show that even a defaultValue might be context dependent.

documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc
```diff
@@ -703,3 +703,40 @@ public interface MovieMapper {
 ====
 Although the used mechanism is the same, the user has to be a bit more careful. Refactoring the name of a defined qualifier in an IDE will neatly refactor all other occurrences as well. This is obviously not the case for changing a name.
 ====
+
+=== Combining qualifiers with defaults
+When using `defaultValue` with `qualifiedByName` or `qualifiedBy` be aware that the default value will the passed into the conversion method. In the case that this is not wanted you need to use `defaultExpression` instead of `defaultValue`.
```
sjaakd, line 708: Perhaps: a better approach:

Please note that the `Mapping#defaultValue` is in essence a `String`. This `String` needs to be converted to the `Mapping#target` . Providing a  `qualifiedByName` or `qualifiedBy` can lead to errors if the `Mapping#source` is not a `String`, since it forces MapStruct to use that method. To solve this, please provide an approriate mapping method, that transforms the `String` into the desired type of `Mapping#target` and also annotate this method with the same  `qualifiedByName` or `qualifiedBy`  annotation.

I don't think `defaultExprerssion` is intended as an alternative to `defaultValue`.  (do not want to encourage its use).

sjaakd (changes requested):

documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc
sjaakd: This is a matter of taste: but I really dislike the `defaultExpression`. It was always intended as a kind of a 'gap stop'

sjaakd: Or perhaps we should document both. I don't know. If you guys really feel that `defaultExpression` is the way to go.

Zegveld: I'm thinking of adjusting the text to the following:

----
Please note that the `Mapping#defaultValue` is in essence a `String`, which needs to be converted to the `Mapping#target`. Providing a `Mapping#qualifiedByName` or `Mapping#qualifiedBy` will force MapStruct to use that method. If you want different behavior for the `Mapping#defaultValue`, then please provide an appropriate mapping method. This mapping method needs to transforms a `String` into the desired type of `Mapping#target` and also be annotated so that it can be found by the `Mapping#qualifiedByName` or `Mapping#qualifiedBy`.
----

This is not yet done in code, but I figured we could discuss it here beforehand. That way we can get a correct description out there for the guide.

I guess the other open bit for this issue is the javadoc enhancements to make it more clear that the `qualifiedByName` and `qualifiedBy` will also be used for the `defaultValue`. I think it would suffice to add this to the javadoc of the `qualifiedBy`/`qualifiedByName` bit, but not needed for the `defaultValue` one.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
sjaakd: > Also I find it weird that the behavior of `defaultValue` and `defaultExpression` differs. The only thing I would expect to be different with these is that one uses a `String` or `Primitive Data Type` input as a default value, while the other can handle all the other types using methods.

Some history: They are in essence different because the `expression` we introduced long time ago with the idea the user would have a means to solve all kinds of shortcomings that we did not have an answer for (yet) 😄 . The reason that it has a syntax like `java(...)` is that we anticipated other syntaxes, like `el(..)` for expression language.  I even thought about using antlr and used the jsr of el to generate an AST for generarting code. But I deemed it to cumbersome. 

So in the end we decided to do no interpretation at all. In contrast: a `constant` was supposed to be a `String` from the start. Later we added the relatively new option to interpret java literals: constants for the primitive types. Like `5d`, `3L` etc.

The default variants evolved somewhere in the middle.

documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc
filiphr: I too am not a big fan of the expression semantics as well. However, I do think that in this case it is OK to mention it. Your last proposal is OK @Zegveld. I would only add one more sentence in the end "An alternative to avoid the usage of the qualified mapping method `defaultValueExpression` can be used.".

>I guess the other open bit for this issue is the javadoc enhancements to make it more clear that the `qualifiedByName` and `qualifiedBy` will also be used for the `defaultValue`. I think it would suffice to add this to the javadoc of the `qualifiedBy`/`qualifiedByName` bit, but not needed for the `defaultValue` one.

I agree that adding extra javadoc to the `qualifiedBy` / `qualifiedByName` to clarify that it is also used for the `defaultValue` is enough.

Zegveld:
Updated the PR with the javadoc adjustment and the asciidoc. I did leave the examples in the asciidoc, I think that should be fine.
@sjaakd and @filiphr can you check if this would suffice for documentation?
Best is to do a full compare because then you will see it all.

filiphr:
LGTM @Zegveld

sjaakd (changes requested):
I've got some tiny remarks still (sorry @Zegveld ). When fixed, you can go ahead and merge.

Zegveld:
> I've got some tiny remarks still (sorry @Zegveld ). When fixed, you can go ahead and merge.

@sjaakd , no problem, that is why I requested the review.
Though I see some remarks on outdated code, could not reply on those in thread so those replies are below:

>> and also be annotated
>
>I guess it needs to be the same annotation as you require for the `Mapping#target`. Perhaps make that explicit.

how would the following need to be adjusted to make it more explicit?
> and also be annotated so that it can be found by the `Mapping#qualifiedByName` or `Mapping#qualifiedBy`.

>> I guess the other open bit for this issue is the javadoc enhancements to make it more clear that the `qualifiedByName` and `qualifiedBy` will also be used for the `defaultValue`. I think it would suffice to add this to the javadoc of the `qualifiedBy`/`qualifiedByName` bit, but not needed for the `defaultValue` one.
>
>Agreed! And mention also `#Mapping.constant# `

I think that for `constant` it is obvious that it would use the `qualifiedBy`/`qualifiedByName`, because why else would you use that combination. If you want to fix it for a constant then you remove the `qualifiedBy`/`qualifiedByName` bit from the `@Mapping`.
However I will add it to the constant javadoc: 
> You can use `#qualifiedBy()` or `#qualifiedByName()` to force the use of a conversion method even when one would not apply. (e.g. `String` to `String`)

sjaakd (approved):

sjaakd:
@Zegveld : nice job.. thanks for taking this one..

Issue #1166, opened 2017-03-31
Nesting on null target property


softdays:
When I use the following mapping configuration with a nested property:
`   
@Mapping(target = "nestedId", source = "nested.id")
ObjectA map(ObjectB source);

@InheritInverseConfiguration
ObjectB map(ObjectA source);
`
Mapstruct generates the inverse mapping below:
`
ObjectB objectB = new ObjectB();

Nested nested = new Nested();
objectB.setNested(nested);

nested.setId(source.getNestedId());
`
But when `nested` is `null` I get an instance of ObjectB with a not-null nested property containing a null id.
I need to get an instance of ObjectB with null value on its nested property.

Can I configure MapStruct in this way?

filiphr:
@softdays You are not using the latest 1.2.0.Beta2. I think that your issue is similar as #1011. 

However, I think that currently you are always going to get a non `null` instance of the nested property with a `null` id if the `source.nestedId` is `null`.

Do you have a suggestion how we could do what you need?

In order to do something like that we will have to check all the properties that are being mapped and if all of them are `null` then return `null`, otherwise create a new object and map. However, this can trigger a large chain of mappings if you are mapping multiple target properties in the same level

softdays:
Yes you're right, I was using 1.1.0.Final. And yes I can confirm that this is the same using 1.2.0.Beta2.
Indeed I had not actually perceived all the potential complexity of my issue...
Do you think that it could be addressed with a new `NullValue*Strategy`? Or maybe with a specific property on the property mapping?
I am aware of not bringing enough proposal

filiphr:
@softdays maybe a new `NullValue*Strategy` could help. However, this will need quite a few changes in how we generate the methods.

Maybe you can use an `@AfterMapping` and set the property to `null` explicitly if `source.nestedId` is `null`

sjaakd:
I guess that the proper solution would-be to implement multi-argument factory methods or constructor selection as proposed in pr #874. In that way the user could control which nested properties would lead to creating new objects and which should then be completed by mapstruct automapping.

filiphr:
@sjaakd you are right, theoretically multi-argument factory methods will solve the problem. Thinking a bit more about the linked PR constructors can also be seen as a way of factory methods. That means that if we introduce multi value factories, then we could use it there as well

sjaakd:
[...]

Issue #2610, opened 2021-10-16
Support conditional mapping for source parameters


filiphr:
With #2051 we added support for conditional mappings. Currently this only works for properties. However, it would be good to be able to apply the same logic for the source parameters.

e.g. If we have the following mapper

```java
@Mapper
public interface CarMapper {

    CarDto map(Car car);

}
```

We currently generated something like:

```java
public class CarMapperImpl implements CarMapper {

    public CarDto map(Car car) {
        if ( car == null ) {
            return null;
        }

        // perform mapping
    }
}
```

It would be good to be able to define a custom method that would tell is if the car should be mapped or not:

e.g. we could generate something like

```java
public class CarMapperImpl implements CarMapper {

    public CarDto map(Car car) {
        if ( isCarPresent(car) ) {
            return null;
        }

        // perform mapping
    }
}
```

where `isCarPresent` is a method on the `CarMapper`. It can be a method that is available anywhere in the same discovery mechanism we have for the property conditional methods.

I believe that this issue is not super difficult to implement and would appreciate if someone from the community is interested in stepping up.

filiphr:
This is more complicated that I expected, therefore moving it to 1.6 so we can think it through a bit more

PR #2640, opened 2021-11-04
Return null if all mapped values are null which is controlled by ValueMappingStrategy.RETURN_NULL_ON_ALL_NULL_VALUE

tim12s:
This will return null if all mapped values are null which is controlled by ValueMappingStrategy.RETURN_NULL_ON_ALL_NULL_VALUE.   Issue is #1166 .

The solution generates a null check after processing for bean mapping and map mapping like:

```
  protected Address orderDTOToAddress(OrderDTO orderDTO) {
        if ( orderDTO == null ) {
            return null;
        }

        address.setLine1( orderDTO.getLine1() );
        address.setLine2( orderDTO.getLine2() );
        address.setLine3( orderDTO.getLine3() );

        if ( address.getLine1() == null
            && address.getLine2()== null
            && address.getLine3() == null)
            return null;

        return address;
    }
```

and 

```
    protected NestedMiddle orderDTOToNestedMiddle(OrderDTO orderDTO) {
        if ( orderDTO == null ) {
            return null;
        }

        NestedMiddle nestedMiddle = new NestedMiddle();
        nestedMiddle.setLeaf( orderDTOToNestedLeaf( orderDTO ) );

        if ( nestedMiddle.getLeaf() == null )
            return null;

        return nestedMiddle;
    }
```

This is the least disruptive change.  As seen in my comments #1166 the solution to avoid the creation of the object (check in advance of the object assignment) will require major changes for which I am not yet suitable to make.

### DIFF
--- a/core/src/main/java/org/mapstruct/NullValueMappingStrategy.java
+++ b/core/src/main/java/org/mapstruct/NullValueMappingStrategy.java
@@ -30,5 +30,12 @@ public enum NullValueMappingStrategy {
      * <li>For map mapping methods an empty map will be returned.</li>
      * </ul>
      */
-    RETURN_DEFAULT;
+    RETURN_DEFAULT,
+
+    /**
+     * If {@code null} is passed to a mapping method, or if all the mapped values from the passed object are null, then
+     * {@code null} will be returned.
+     */
+    RETURN_NULL_ON_ALL_NULL_VALUES;
+
 }
--- a/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
@@ -212,6 +212,50 @@ However, by specifying `nullValueMappingStrategy = NullValueMappingStrategy.RETU
 
 The strategy works in a hierarchical fashion. Setting `nullValueMappingStrategy` on mapping method level will override `@Mapper#nullValueMappingStrategy`, and `@Mapper#nullValueMappingStrategy` will override `@MapperConfig#nullValueMappingStrategy`.
 
+[[mapping-result-for-all-null-mapping-values]
+=== Controlling mapping result for all-'null' properties in source bean mappings (excludes update mapping methods)
+
+Mapstruct offers control over the object to create when the value of all source argument mapping methods equal 'null'.  By default the value returned wis controlled only considers whether the source object is null and returns 'null' if the object is 'null'.
+
+By specifying 'nullValueMappingStrategy = NullValueMappingStrategy.RETURN_NULL_ON_ALL_NULL_VALUES` on `@BeanMapping`, `@IterableMapping`, `@MapMapping`, or globally on `@Mapper` or `@MapperConfig`, the mapping result can be altered to a 'null' value.  This means for:
+
+* *Bean mappings*: an 'null' target bean will be returned if all mapping methods are null.
+* *Iterables / Arrays*: an empty iterable will be returned.
+* *Maps*: an empty map will be returned.
+
+The strategy works in a hierarchical fashion. Setting `nullValueMappingStrategy` on mapping method level will override `@Mapper#nullValueMappingStrategy`, and `@Mapper#nullValueMappingStrategy` will override `@MapperConfig#nullValueMappingStrategy`.
+
+.generated all mapped value null check
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+protected Address orderDTOToAddress(OrderDTO orderDTO) {
+    if ( orderDTO == null ) {
+        return null;
+    }
+
+    Address address = new Address();
+
+    address.setLine1( orderDTO.getLine1() );
+    address.setLine2( orderDTO.getLine2() );
+    address.setLine3( orderDTO.getLine3() );
+
+    if (   ( address . getLine1() == null )
+        && ( address . getLine2() == null )
+        && ( address . getLine3() == null )
+    ) return null;
+
+    return address;
+}
+----
+====
+
+[NOTE]
+====
+Any default value assigned, or any primitive value that can not be part of a null check, will result in the all-value null check not being performed.
+====
+
 
 [[mapping-result-for-null-properties]]
 === Controlling mapping result for 'null' properties in bean mappings (update mapping methods only).
--- a/processor/src/main/java/org/mapstruct/ap/internal/gem/NullValueMappingStrategyGem.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/gem/NullValueMappingStrategyGem.java
@@ -12,16 +12,25 @@ package org.mapstruct.ap.internal.gem;
  */
 public enum NullValueMappingStrategyGem {
 
-    RETURN_NULL( false ),
-    RETURN_DEFAULT( true );
+    RETURN_NULL( false , false),
+    RETURN_DEFAULT( true, false ),
+    RETURN_NULL_ON_ALL_NULL_VALUES ( false , true);
 
     private final boolean returnDefault;
 
-    NullValueMappingStrategyGem(boolean returnDefault) {
+    private final boolean returnNullOnAllNullValues;
+
+    NullValueMappingStrategyGem(boolean returnDefault, boolean returnNullOnAllNull) {
         this.returnDefault = returnDefault;
+        this.returnNullOnAllNullValues = returnNullOnAllNull;
     }
 
     public boolean isReturnDefault() {
         return returnDefault;
     }
+
+    public boolean isReturnNullOnAllNullValues() {
+        return returnNullOnAllNullValues;
+    }
+
 }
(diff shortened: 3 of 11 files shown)


tim12s:
@sjaakd @filiphr Not sure if I'm configuring the tests correctly.  Let me know if there is any changes. 

I'm considering adding a mode that will consider only the mapped methods  which do not have defaults/constants assigned. I.e. there must be at least 1 nullable value which is not a constant/default value.

filiphr:
Thanks a lot for your work on this @tim12s. I've added https://github.com/mapstruct/mapstruct/issues/1166#issuecomment-962421107 to the issue and I'd like to reiterate here as well.

Perhaps we need to take a step back from what is proposed in the issues and resolve this through #2610. This way users will have full control of what needs to happen.

Lets take your example

we can have something like this:

```java
public PresenceUtils {

    @Condition
    public static boolean isPresentOrderDto(OrderDTO order) {
        if (order == null) {
            return false;
        }

        if (order.getLine1() == null) {
            return false;
        }
    
        return true;
    }

}
```

Then the generated method will look like

```java
  protected Address orderDTOToAddress(OrderDTO orderDTO) {
        if ( PresenceUtils.isPresentOrderDto( orderDTO) ) {
            return null;
        }

        address.setLine1( orderDTO.getLine1() );
        address.setLine2( orderDTO.getLine2() );
        address.setLine3( orderDTO.getLine3() );

        return address;
    }
```

Doing it like this the performance will be better as well since there will be no unnecessary object creations and method invocations for the `Address`. In addition to that constant values will keep working as they do now and would be driven by the current `NullValueMappingStrategy`.

What do you think about this?

tim12s:
Hi,

Well mapstruct is principally a code automation process - I don't really want to write the code for basic checks.  @Condition should be for semantic checks not basic mapping logic.

In principle I think the @Condition is a great idea.  I question how it will deal with 2+ property mappings and nested checks.

Another 2 points: 
1. Inverse Mappings - Cannot automatically generate inverse mappings with @Condition as there is no guarantee that the condition coded is not a semantic condition.

2. Nested Middle Mapping - Does not solve the middle nested mapping problem.  How will @Condition solve (f.3) below and (f.5) below?

My first attempt tried to bring all of the logic before the memory allocation.  See (b.target.1) and (c.target.2) above in my prior comments.  It would be easy to wrap that logic into an automatically generated condition check. 



Consider the following...

**Snippet (f.1):**
```
@Mapper
public interface OrderMapper {

    @Mapping(source = "customerName", target = "customer.name")
    @Mapping(source = "nestedLeafOid", target = "nestedMiddle.leaf.oid")
    @Mapping(source = "line1", target = "customer.address.line1")
    @Mapping(source = "line2", target = "customer.address.line2")
    @Mapping(source = "line3", target = "customer.address.line3")
    Order convertToOrder(OrderDTO orderDTO);

    @InheritInverseConfiguration(name = "convertToOrder")
    OrderDTO convertToOrderDTO(Order order);

}

```

This generates


**Snippet (f.2):**
```
    @Override
    public Order convertToOrder(OrderDTO orderDTO) {
        if ( orderDTO == null ) {
            return null;
        }

        Order order = new Order();

        order.setCustomer( orderDTOToCustomer( orderDTO ) );
        order.setNestedMiddle( orderDTOToNestedMiddle( orderDTO ) );

        return order;
    }
```


and 


**Snippet (f.3):**
```
    protected NestedMiddle orderDTOToNestedMiddle(OrderDTO orderDTO) {
        if ( orderDTO == null ) {
            return null;
        }

        NestedMiddle nestedMiddle = new NestedMiddle();

        nestedMiddle.setLeaf( orderDTOToNestedLeaf( orderDTO ) );

        return nestedMiddle;
    }
```


and

**Snippet (f.4):**
```
    protected NestedLeaf orderDTOToNestedLeaf(OrderDTO orderDTO) {
        if ( orderDTO == null ) {
            return null;
        }

        NestedLeaf nestedLeaf = new NestedLeaf();

        nestedLeaf.setOid( orderDTO.getNestedLeafOid() );

        return nestedLeaf;
    }
```


and

**Snippet (f.5):**
```
    protected Customer orderDTOToCustomer(OrderDTO orderDTO) {
        if ( orderDTO == null ) {
            return null;
        }

        Customer customer = new Customer();

        customer.setAddress( orderDTOToAddress( orderDTO ) );
        customer.setName( orderDTO.getCustomerName() );

        return customer;
    }

```



This might be the simplest way to automatically generate the checks and mix them through the generated code.  Again, when it comes to @Conditional, I don't want to write this manually unless it is a semantic check that needs higher order logic.

```
public boolean mapCustomer(Order orderDTO) {
    return orderDTO != null && (
       orderDTO.getName() != null
    || mapAddress(orderDTO));
}

public boolean mapAddress(Order orderDTO) {
    return orderDTO != null &&
    ( orderDTO.getLine1() != null
    || orderDTO.getLine2() != null
    || orderDTO.getLine3() != null);
}

public boolean mapNestedMiddle(Order orderDTO) {
    return orderDTO != null &&
    ( orderDTO.getNestedLeafOid() != null);
}
```

filiphr:
>Well mapstruct is principally a code automation process - I don't really want to write the code for basic checks. `@condition` should be for semantic checks not basic mapping logic.

Indeed MapStruct is a code generator. However, it is a code generator for bean mappings and not for condition checks.

You could always write another code generator that will generate methods with `@Condition`. 

Your examples are indeed quite complex and I do not think that MapStruct needs to solve all of that. You can easily write something that would make it possible to use different conditional methods.


e.g.

```java
@Mapper
public interface OrderMapper {

    @Mapping(source = "nestedLeafOid", target = "nestedMiddle.leaf.oid")
    @Mapping(source = "orderDTO", target = "customer", conditionQualifiedByName = "mapCustomerFromOrder")
    Order convertToOrder(OrderDTO orderDTO);

    @Mapping(source = "customerName", target = "name")
    @Mapping(source = "orderDTO", target = "address", conditionQualifiedByName = "mapAddressFromOrder")
    Customer convertToCustomer(OrderDTO orderDTO);

    @Mapping(source = "line1", target = "line1")
    @Mapping(source = "line2", target = "line2")
    @Mapping(source = "line3", target = "line3")
    Address convertToAddress(OrderDTO orderDTO);

    @InheritInverseConfiguration(name = "convertToOrder")
    OrderDTO convertToOrderDTO(Order order);

    @Condition
    @Named("mapCustomerFromOrder")
    default boolean mapCustomerFromOrder(OrderDTO orderDTO) {
        return orderDTO != null && (
                orderDTO.getCustomerName() != null
                        || mapAddressFromOrder(orderDTO));
    }

    @Condition
    @Named("mapAddressFromOrder")
    default boolean mapAddressFromOrder(OrderDTO orderDTO) {
        return orderDTO != null && (orderDTO.getLine1() != null
                || orderDTO.getLine2() != null
                || orderDTO.getLine3() != null);
    }


    @Condition
    @Named("mapNestedMiddleFromOrder")
    default boolean mapNestedMiddleFromOrder(OrderDTO orderDTO) {
        return orderDTO != null &&
                ( orderDTO.getNestedLeafOid() != null);
    }

}
```

This should generate something resembling of what you would expect. Yes I know that it is not the most fun code to write, but it is highly tailored to the example you have.

Perhaps we could enhance MapStruct to add some hook points that would be able to generate the implementations for the `@Condition` methods.

This would of course require an additional `@Mapping` definition for the inverse configuration `convertToOrderDTO`. However, this could be expanded by adding something like


```java
@Mapping(source = "orderDTO", target = "customer", conditionQualifiedByName = "mapCustomerFromOrder",
        nested = {
            @Mapping(source = "customerName", target = "name"),
            @Mapping(source = "orderDTO", target = "address", conditionQualifiedByName = "mapAddressFromOrder")
        }
    )
```

NB: `Mapping#nested` does not exist currently. It is an example for something that we could add. Thinking now about it, perhaps that's how we should have implemented the nested target mappings.

-----

I would mark this so we can discuss this within the team to decide what direction we would like to go towards.

sjaakd:
> Perhaps we could enhance MapStruct to add some hook points that would be able to generate the implementations for the `@Condition` methods.

I like that idea.... What if we make some pre-defined `@Condition`..  Perhaps along the lines of what we do with `@MappingControl`?

So have an annotation like this in a `org.mapstruct.condition` package: 

```java 
@Contition
public @interface ConditionNullOnAllNull{}
```

When used in the above mapping:

```java
@Mapper
public interface OrderMapper {

    @Mapping(source = "nestedLeafOid", target = "nestedMiddle.leaf.oid")
    @Mapping(source = "orderDTO", target = "customer", conditionQualifiedBy = ConditionNullOnAllNull.class )
    Order convertToOrder(OrderDTO orderDTO);
 
   // other methods
```

Heck.. we could even extend that with the other options of `NullValueMappingStrategy` and deprecate this option in the future. You could add a property to `@ConditionNullOnAllNull` to tailor nested mapping.

Implementation wise we could isolate this by adding another processor to the chain, e.g. `ConditionCreationPostProcessor` or something like that.

filiphr:
>I like that idea.... What if we make some pre-defined `@Condition`.. Perhaps along the lines of what we do with `@MappingControl`?

Before adding some of our own conditions, I think it might make sense to see the request from the community about this? I can easily see this as an extra project that could add certain annotations or even single annotations that would generate the conditional code needed for MapStruct.

>Implementation wise we could isolate this by adding another processor to the chain, e.g. `ConditionCreationPostProcessor` or something like that.

That's what I had in mind for this. To have an SPI that would provide the method definition, although it would rely on internal MapStruct classes. Perhaps providing something from our side would make sense. In any case I would like to see more request from the community for this feature.

PR #2653, opened 2021-11-13
#2553 Support source property paths for maps

### LINKED ISSUE #2553: New mapping feature not support property paths

filiphr:
Fixes #2553 

This PR does some small changes in the way we acquire read accessors. The refactorings in the PR, especially the new methods in the `Type` can open the door for better support for things like Jackson `JsonNode`,  `ResultSet`, etc.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -48,6 +48,8 @@ import org.mapstruct.ap.internal.util.Nouns;
 import org.mapstruct.ap.internal.util.TypeUtils;
 import org.mapstruct.ap.internal.util.accessor.Accessor;
 import org.mapstruct.ap.internal.util.accessor.AccessorType;
+import org.mapstruct.ap.internal.util.accessor.MapValueAccessor;
+import org.mapstruct.ap.internal.util.accessor.MapValuePresenceChecker;
 
 import static org.mapstruct.ap.internal.util.Collections.first;
 
@@ -60,6 +62,7 @@ import static org.mapstruct.ap.internal.util.Collections.first;
  * through {@link TypeFactory}.
  *
  * @author Gunnar Morling
+ * @author Filip Hrisafov
  */
 public class Type extends ModelElement implements Comparable<Type> {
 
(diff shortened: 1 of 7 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -597,6 +600,48 @@ public Type asRawType() {
         }
     }
 
+    public Accessor getReadAccessor(String propertyName) {
+        if ( isMapType() ) {
+            List<Type> typeParameters = getTypeParameters();
+            if ( typeParameters.size() == 2 && typeParameters.get( 0 ).isString() ) {
```
sjaakd, line 606: perhaps move this check to a method `hasMapSignature`...

sjaakd (approved):
Just a tiny remark.. If all testcases pass: LGTM

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: Good point, I have extracted a method `hasStringMapSignature` for this.

PR #2667, opened 2021-11-21
#2666 Presence Check should be applied to source parameters when used in `@Mapping`

### LINKED ISSUE #2666: @Condition not working when the source in the Mapping is the source parameter

filiphr:
Fixes #2666

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -559,11 +559,17 @@ public class PropertyMapping extends ModelElement {
 
             // parameter reference
             if ( propertyEntry == null ) {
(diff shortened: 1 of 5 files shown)


Zegveld (approved):
looks good to me.

sjaakd (approved):
I played around a bit with `@Mapper( nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT ) ` to see if we could get rid on the parameter null check. That's not possible.  The check simply moves around the mappings.

So, to complete this kind of functionality,  I think we should extend this with: `NullValueMappingStrategy.NONE` WDYT?. 

That would be a new PR BTW 😄

filiphr:
>So, to complete this kind of functionality, I think we should extend this with: `NullValueMappingStrategy.NONE` WDYT?.

I think that #2610 is the approach we need to take for the parameter presence checks.

PR #2675, opened 2021-11-25
#2673: the getCandidates method now correctly returns false.

### LINKED ISSUE #2673: Optional "wrapping" pattern broken with 1.5.0.Beta

The following pattern worked as a way to map from `Optional` to non `Optional`:

```
    default <T> Optional<T> wrap(T value) {
        return Optional.ofNullable(value);
    }

    default <T> T unwrap(Optional<T> value) {
        return value.orElse(null);
    }
```

However, with 1.5.0.Beta compilation seems to fail on an unrelated field.

For example, my classes are:

```
public record FromClass(int a, Optional<String> b) {}

public record ToClass(int a, String b) {}
```

During compilation I get:

```
error: Internal error in the mapping processor: java.lang.IllegalArgumentException: int     at jdk.compiler/com.sun.tools.javac.model.JavacTypes.getDeclaredType0(JavacTypes.java:272)      at jdk.compiler/com.sun.tools.javac.model.JavacTypes.getDeclaredType(JavacTypes.java:241)       at org.mapstruct.ap.internal.util.AbstractTypeUtilsDecorator.getDeclaredType(AbstractTypeUtilsDecorator.java:119)       at org.mapstruct.ap.internal.model.source.MethodMatcher$GenericAnalyser.resolve(MethodMatcher.java:381)         at org.mapstruct.ap.internal.model.source.MethodMatcher$GenericAnalyser.lineUp(MethodMatcher.java:174)          at org.mapstruct.ap.internal.model.source.MethodMatcher$GenericAnalyser.access$000(MethodMatcher.java:110)      at org.mapstruct.ap.internal.model.source.MethodMatcher.matches(MethodMatcher.java:71)          at org.mapstruct.ap.internal.model.source.SourceMethod.matches(SourceMethod.java:497)   at org.mapstruct.ap.internal.model.source.selector.TypeSelector.lambda$getMatchingParameterBinding$0(TypeSelector.java:164)     at java.base/java.util.ArrayList.removeIf(ArrayList.java:1672)
...
```

If I explicitly ignore the mappings for the unrelated field `a` then the compilation works as expected.

Zegveld:
There is a special case where the getCandidates method incorrectly returned true.

Fixes #2673

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2673/Issue2673Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2673/Issue2673Mapper.java
@@ -29,38 +29,38 @@ public interface Issue2673Mapper {
     }
 
     class Target {
-        private final int a;
-        private final String b;
+        private final int primitive;
+        private final String nonPrimitive;
 
-        public Target(int a, String b) {
-            this.a = a;
-            this.b = b;
+        public Target(int primitive, String nonPrimitive) {
+            this.primitive = primitive;
+            this.nonPrimitive = nonPrimitive;
         }
 
-        public int getA() {
-            return a;
+        public int getPrimitive() {
+            return primitive;
         }
 
-        public String getB() {
-            return b;
+        public String getNonPrimitive() {
+            return nonPrimitive;
         }
     }
 
     class Source {
-        private final int a;
-        private final Optional<String> b;
+        private final int primitive;
+        private final Optional<String> nonPrimitive;
 
-        public Source(int a, Optional<String> b) {
-            this.a = a;
-            this.b = b;
+        public Source(int primitive, Optional<String> nonPrimitive) {
+            this.primitive = primitive;
+            this.nonPrimitive = nonPrimitive;
         }
 
-        public int getA() {
-            return a;
+        public int getPrimitive() {
+            return primitive;
         }
 
-        public Optional<String> getB() {
-            return b;
+        public Optional<String> getNonPrimitive() {
+            return nonPrimitive;
         }
     }
 }
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2673/Issue2673Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2673/Issue2673Mapper.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2673;
+
+import java.util.Optional;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Ben Zegveld
+ */
+@Mapper
+public interface Issue2673Mapper {
+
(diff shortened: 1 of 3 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_2673/Issue2673Test.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+@WithClasses({
+    Issue2673Mapper.class
+})
+class Issue2673Test {
+
+    @ProcessorTest( Compiler.JDK )
+    @IssueKey( "2673" )
+    void shouldcompileWithTheJdkCompiler() {
+        // This issue does not cause a problem for the eclipse compiler.
```
filiphr, line 24: If there is no problem for the eclipse compiler why not test with it as well?

processor/src/test/java/org/mapstruct/ap/test/bugs/_2673/Issue2673Mapper.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Optional;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Ben Zegveld
+ */
+@Mapper
+public interface Issue2673Mapper {
```
filiphr, line 17: Let's go with some more descriptive names than `a`, `b`.

filiphr (changes requested):
Looks good to me. I've left some comments that should be addressed I think.

Do you by any chance know why this is happening for the javac compiler only? Is it a bug that we should perhaps report?

processor/src/test/java/org/mapstruct/ap/test/bugs/_2673/Issue2673Mapper.java
Zegveld: renamed them to `primitive`, `nonPrimitive` respectively.

Zegveld:
> Looks good to me. I've left some comments that should be addressed I think.
> 
> Do you by any chance know why this is happening for the javac compiler only? Is it a bug that we should perhaps report?

fixed the commented bits. The reason it happens for the javac compiler is because it does what the API says. We called a method with a primitive type that should not be called with a primitive type. The eclipse compiler is less restrictive apparently.


excerpt from the API:
```
{@return the type corresponding to a type element and actual type arguments} Given the type element for Set and the type mirror for String, for example, this method may be used to get the parameterized type Set<String>.

....

IllegalArgumentException - if too many or too few type arguments are given, or if an inappropriate type argument or type element is provided
```
Logically a primitive cannot be a type argument, you will need to use the boxed version of it. (`Integer` instead of `int`) 
So in this case we were calling it with an inappropriate type argument.

processor/src/test/java/org/mapstruct/ap/test/bugs/_2673/Issue2673Test.java
```diff
@@ -0,0 +1,24 @@
...
+ * @author Ben Zegveld
+ */
+@WithClasses({
+    Issue2673Mapper.class
+})
+class Issue2673Test {
+
+    @ProcessorTest
+    @IssueKey( "2673" )
+    void shouldcompile() {
```
filiphr, line 22:
```suggestion
    void shouldCompile() {
```

PR #2702, opened 2021-12-25
#2686 Add documentation about when mappers are injected

### LINKED ISSUE #2686: Bean is not injected with "uses" attribute of @Mapper annotation
I have the following `@Mapper` annotation with `LocationDetailsMapper` injection
```
@Mapper(
        componentModel = "spring",
        uses = {
                LocationDetailsMapper.class,
        }
)
```

I expect that I can use the following `@Mapping`:
```
@Mapping(target = "details", expression = "java(locationDetailsMapper.mapDetails(someSource1, someSource2))")
```

However, generated code does not have locationDetailsMapper injection

The only way to have mapper injected to use it as in the snippet above is to provide another `@Mapping(target = "smth", source = "someSource1.smth")`, which would use that LocationDetailsMapper method like `OtherMyClass map(MyClass smth)` somehow and only then I can use expressions like `expression = "java(locationDetailsMapper.mapDetails(someSource1, someSource2))"`, which is a bit unexpected behavior - as I thought if I specify something in `uses` I expect it to be injected unconditionally

To sum up - mapper is injected only in case mapstruct can automatically detect where to use that mapper and if it can not do so it does not inject the specified mapper (which is exactly my case when I want to specify my custom mapping with 2 or more arguments in `expression`)

filiphr:
Fixes #2686

### DIFF
--- a/documentation/src/main/asciidoc/chapter-4-retrieving-a-mapper.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-4-retrieving-a-mapper.asciidoc
@@ -117,7 +117,7 @@ public interface CarMapper {
 ----
 ====
 
-The generated mapper will inject all classes defined in the **uses** attribute.
+The generated mapper will inject all classes defined in the **uses** attribute and MapStruct has detected that they are actually used on some of the mappings.
 When `InjectionStrategy#CONSTRUCTOR` is used, the constructor will have the appropriate annotation and the fields won't.
 When `InjectionStrategy#FIELD` is used, the annotation is on the field itself.
 For now, the default injection strategy is field injection, but it can be configured with <<configuration-options>>.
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/chapter-4-retrieving-a-mapper.asciidoc
```diff
@@ -117,7 +117,7 @@ public interface CarMapper {
 ----
 ====
 
-The generated mapper will inject all classes defined in the **uses** attribute.
+The generated mapper will inject all classes defined in the **uses** attribute and MapStruct has detected that they are actually used on some of the mappings.
```
Zegveld, line 120: does the injection also take place if a static method is used from the defined class?

Perhaps a better wording would be:
`The generated mapper will inject classes defined in the **uses** attribute if MapStruct has detected that it needs to use an instance of it for a mapping.`

Zegveld (changes requested):
I think it is still to ambiguous, see comment.

filiphr:
You are right @Zegveld. Your suggestion is way better. I've added that

Zegveld (approved):

filiphr:
Thanks for the review @Zegveld

PR #2705, opened 2021-12-25
Fix broken reference to constants within nested enums

### LINKED ISSUE #2704: Fail to import nested enum when an enum constant is referenced

While implementing #2616, a use case has been forgotten.

unshare:
Fixes #2704


filiphr:
Good catch @unshare. Thanks for the PR

PR #2708, opened 2021-12-27
2696: Invert `@SubclassMappings` with `@InheritInverseConfiguration`.

### LINKED ISSUE #2696: Allow @InheritInverseConfiguration with @SubclassMappings
using `@InheritInverseConfiguration` results in an error, I had to explicitly write the `@SubclassMapping`s manually. It would be useful to be able to inherit them, either through the standard annotation or with an ad-hoc one.

_Originally posted by @pennal in https://github.com/mapstruct/mapstruct/issues/2695#issuecomment-998924888_

Zegveld:
`@InheritConfiguration` still not possible with `@SubclassMapping`'s, but `@InheritInverseConfiguration` is possible with `@SubclassMapping`'s.

Added a new error in case of identical source classes in the `@SubclassMapping`'s.
For normal `@SubclassMapping`'s this was easy to spot, but with the `@InheritInverseConfiguration` it was suddenly possible, and you would end up with unreachable code.
```java
if ( A instanceof B ) {
  // do mapping
} else if ( A instanceof B ) { // <--- will never be true
  // do different mapping //  <--- unreachable code
}
```

Fixes #2696

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
@@ -189,7 +189,7 @@ public class MappingMethodOptions {
             }
             else {
                 if ( templateOptions.getValueMappings() != null ) {
-                    // iff there are also inherited mappings, we inverse and add them.
+                    // if there are also inherited mappings, we inverse and add them.
                     for ( ValueMappingOptions inheritedValueMapping : templateOptions.getValueMappings() ) {
                         ValueMappingOptions valueMapping =
                             isInverse ? inheritedValueMapping.inverse() : inheritedValueMapping;
@@ -202,6 +202,7 @@ public class MappingMethodOptions {
             }
 
             if ( isInverse ) {
+                // normal inheritence of subclass mappings will result runtime in infinite loops.
                 setSubclassMapping( SubclassMappingOptions.copyForInverseInheritance(
                           templateOptions.getSubclassMappings(),
                           sourceMethod,
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
@@ -130,10 +130,6 @@ public class MappingMethodOptions {
         this.valueMappings = valueMappings;
     }
 
-    public void setSubclassMapping(Set<SubclassMappingOptions> subclassMappings) {
-        this.subclassMappings = subclassMappings;
-    }
-
     public MapperOptions getMapper() {
         return mapper;
     }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
@@ -7,13 +7,11 @@ package org.mapstruct.ap.internal.model.source;
 
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import javax.lang.model.element.AnnotationMirror;
-import javax.lang.model.type.TypeMirror;
 
 import org.mapstruct.ap.internal.gem.CollectionMappingStrategyGem;
 import org.mapstruct.ap.internal.model.common.Type;
@@ -50,7 +48,7 @@ public class MappingMethodOptions {
     private EnumMappingOptions enumMappingOptions;
     private List<ValueMappingOptions> valueMappings;
     private boolean fullyInitialized;
-    private Map<TypeMirror, SubclassMappingOptions> subclassMapping;
+    private Set<SubclassMappingOptions> subclassMappings;
 
     private SubclassValidator subclassValidator;
 
@@ -59,7 +57,7 @@ public class MappingMethodOptions {
                                 MapMappingOptions mapMapping, BeanMappingOptions beanMapping,
                                 EnumMappingOptions enumMappingOptions,
                                 List<ValueMappingOptions> valueMappings,
-                                Set<SubclassMappingOptions> subclassMapping, SubclassValidator subclassValidator) {
+                                Set<SubclassMappingOptions> subclassMappings, SubclassValidator subclassValidator) {
         this.mapper = mapper;
         this.mappings = mappings;
         this.iterableMapping = iterableMapping;
@@ -67,7 +65,7 @@ public class MappingMethodOptions {
         this.beanMapping = beanMapping;
         this.enumMappingOptions = enumMappingOptions;
         this.valueMappings = valueMappings;
-        setSubclassMapping( subclassMapping ); // use setter because of special behavior.
+        this.subclassMappings = subclassMappings;
         this.subclassValidator = subclassValidator;
     }
 
@@ -109,7 +107,7 @@ public class MappingMethodOptions {
     }
 
     public Set<SubclassMappingOptions> getSubclassMappings() {
-        return new LinkedHashSet<>( subclassMapping.values() );
+        return subclassMappings;
     }
 
     public void setIterableMapping(IterableMappingOptions iterableMapping) {
@@ -132,11 +130,8 @@ public class MappingMethodOptions {
         this.valueMappings = valueMappings;
     }
 
-    public void setSubclassMapping(Set<SubclassMappingOptions> subclassMapping) {
-        this.subclassMapping = new LinkedHashMap<>();
-        for ( SubclassMappingOptions subclassMappingOptions : subclassMapping ) {
-            this.subclassMapping.put( subclassMappingOptions.getSource(), subclassMappingOptions );
-        }
+    public void setSubclassMapping(Set<SubclassMappingOptions> subclassMappings) {
+        this.subclassMappings = subclassMappings;
     }
 
     public MapperOptions getMapper() {
@@ -215,7 +210,7 @@ public class MappingMethodOptions {
 
             if ( isInverse ) {
                 // normal inheritence of subclass mappings will result runtime in infinite loops.
-                Set<SubclassMappingOptions> inheritedMappings = SubclassMappingOptions.copyForInverseInheritance(
+                List<SubclassMappingOptions> inheritedMappings = SubclassMappingOptions.copyForInverseInheritance(
                                               templateOptions.getSubclassMappings(),
                                               getBeanMapping() );
                 addAllNonRedefined( sourceMethod, annotationMirror, inheritedMappings );
@@ -242,15 +237,15 @@ public class MappingMethodOptions {
     }
 
     private void addAllNonRedefined(SourceMethod sourceMethod, AnnotationMirror annotationMirror,
-                                    Set<SubclassMappingOptions> inheritedMappings) {
-        Set<TypeMirror> redefinedTargets = new HashSet<>( subclassMapping.keySet() ); // unlink the keyset
-        for ( SubclassMappingOptions subclassMappingOptions : inheritedMappings ) {
-            if ( !redefinedTargets.contains( subclassMappingOptions.getSource() ) ) {
+                                    List<SubclassMappingOptions> inheritedMappings) {
+        Set<SubclassMappingOptions> redefinedSubclassMappings = new HashSet<>( subclassMappings );
+        for ( SubclassMappingOptions subclassMappingOption : inheritedMappings ) {
+            if ( !redefinedSubclassMappings.contains( subclassMappingOption ) ) {
                 if ( subclassValidator.isValidUsage(
                                           sourceMethod.getExecutable(),
                                           annotationMirror,
-                                          subclassMappingOptions.getSource() ) ) {
-                    subclassMapping.put( subclassMappingOptions.getSource(), subclassMappingOptions );
+                    subclassMappingOption.getSource() ) ) {
+                    subclassMappings.add( subclassMappingOption );
                 }
             }
         }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
@@ -7,11 +7,13 @@ package org.mapstruct.ap.internal.model.source;
 
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.type.TypeMirror;
 
 import org.mapstruct.ap.internal.gem.CollectionMappingStrategyGem;
 import org.mapstruct.ap.internal.model.common.Type;
@@ -48,7 +50,7 @@ public class MappingMethodOptions {
     private EnumMappingOptions enumMappingOptions;
     private List<ValueMappingOptions> valueMappings;
     private boolean fullyInitialized;
-    private Set<SubclassMappingOptions> subclassMapping;
+    private Map<TypeMirror, SubclassMappingOptions> subclassMapping;
 
     private SubclassValidator subclassValidator;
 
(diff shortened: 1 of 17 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
```diff
@@ -84,7 +92,7 @@ private static boolean isConsistent(SubclassMappingGem gem, ExecutableElement me
                         targetSubclass.toString() );
             isConsistent = false;
         }
-        subclassValidator.isInCorrectOrder( method, gem.mirror(), targetSubclass );
+        subclassValidator.isValidUsage( method, gem.mirror(), targetSubclass );
```
filiphr, line 95: We never use the return type from `isValidUsage`. Should we do that and assign it to `isConsistent` if it is `false`?

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
```diff
@@ -166,6 +174,41 @@ private static void addAndValidateInstance(SubclassMappingGem subclassMapping, E
...
+                    new SubclassMappingOptions(
+                        messager,
+                        typeUtils,
+                        subclassMapping,
+                        sourceSubclass,
+                        targetSubclass,
+                        beanMappingOptions ) );
+    }
+
+    public static Set<SubclassMappingOptions> copyForInverseInheritance(Set<SubclassMappingOptions> subclassMappings,
```
filiphr, line 188: Why don't we pass the `message` and `typeUtils` to this method instead of dragging them in the `SubclassMappingOptions`? It doesn't feel right to me to add those things to the model.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
```diff
@@ -31,9 +32,16 @@
 
     private final TypeMirror source;
     private final TypeMirror target;
+    private final FormattingMessager messager;
+    private final TypeUtils typeUtils;
+    private final SubclassMappingGem subclassMapping;
```
filiphr, line 37: Where is this used?

filiphr (changes requested):
Thanks for your work on this @Zegveld. I've left some comments that are unclear to me.

Also it would be good to add a test case where you can override something in the `InheritInverseConfiguration` method.

e.g.  It should be possible to fix `ErroneousInverseSubclassMapper` by doing

```
@SubclassMapping( target = Bike.class, source = BikeDto.class )
@SubclassMapping( target = Car.class, source = CarDto.class )
Vehicle mapInverse(VehicleDto dto)
```

Of course we also need to be able to carry over an additional `SubclassMapping` that is not defined. Is it possible to be able to distinquish certain things like this? I think yes, because the `source` has to anyways be unique in the first method and thus all the inversed mappings will have a unique `target`.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
Zegveld: it is used at another location:
https://github.com/mapstruct/mapstruct/pull/2708/files#diff-86e6eb61ca7c458ab2ea7274f6950ab596031e8a450771263d5c7492ed151862R198

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
Zegveld: I thought of doing that, but then I had trouble with determining the correct annotation etc for the message to pop-up on. I'll see if I can think of a better solution for this issue.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
Zegveld: inside isValidUsage

Zegveld:
> Also it would be good to add a test case where you can override something in the `InheritInverseConfiguration` method.
> 
> e.g. It should be possible to fix `ErroneousInverseSubclassMapper` by doing
> 
> ```
> @SubclassMapping( target = Bike.class, source = BikeDto.class )
> @SubclassMapping( target = Car.class, source = CarDto.class )
> Vehicle mapInverse(VehicleDto dto)
> ```
> 
> Of course we also need to be able to carry over an additional `SubclassMapping` that is not defined. Is it possible to be able to distinquish certain things like this? I think yes, because the `source` has to anyways be unique in the first method and thus all the inversed mappings will have a unique `target`.

I did think about this, and came to the following conclusion: Inheriting the inverse mappings only makes sense if both source and target are already unique. The reason for this is that the following behavior does not make sense:
```
class ElectricBike extends Bike {
  String voltage;
}
class ManualBike extends Bike {
  int numberOfGearShifts;
}
abstract class Bike {
  String size;
}
class BikeDto {
  String voltage;
  int numberOfGearShifts;
  String size;
}
@Mapper(subclassExhaustiveStrategy = SubclassExhaustiveStrategy.RUNTIME_EXCEPTION)
class BikeMapper {
    @SubclassMapping( source = ElectricBike.class, target = BikeDto.class )
    @SubclassMapping( source = ManualBike.class, target = BikeDto.class )
    BikeDto map(Bike bike);

    @InheritInverseConfiguration
    @SubclassMapping( source = BikeDto.class, target = ManualBike.class )
    Bike mapInverse(BikeDto dto);
}
```
using the following bit of code you get different results:
```
BikeMapper bikeMapper = /*retrieveMapper*/;
ElectricBike eBike = /*createElectricBike*/;
Bike thereAndBackAgain = bikeMapper.mapInverse(bikeMapper.map(eBike));
/* you would expect thereAndBackAgain to be the same eBike as it was before, 
but now it suddenly became a ManualBike. */
```
In this case you would want to have an `inverseCondition` or something, that way you can specify what to do instead of the instanceof check. Might also be useful for a normal `Condition` to replace the instanceof check and do the mapping still. This would even allow multiple `source`'s at that point, thus removing the uniqueness in case a condition is present. I believe this is better solved in a different PR.

Perhaps for now would be a better option to add a flag to the `InheritInverseConfiguration` with which you can disable the `SubclassMapping`s inheritence. That way the user can still use all other mappings reversed, but define all the `SubclassMapping`s manually.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
Zegveld: Solved this by moving the SubclassValidator further outside. Now the same validator will be used for both the methods `SubclassMapping`s as well as the inverse inherited `SubclassMapping`s.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
Zegveld: fixed this as specified using the `bitwise and` (`&` instead of `&&`) operator.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
Zegveld: got removed during fixing of previous remark.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
```diff
@@ -84,7 +86,7 @@ private static boolean isConsistent(SubclassMappingGem gem, ExecutableElement me
                         targetSubclass.toString() );
             isConsistent = false;
         }
-        subclassValidator.isInCorrectOrder( method, gem.mirror(), targetSubclass );
+        isConsistent = isConsistent & subclassValidator.isValidUsage( method, gem.mirror(), targetSubclass );
```
filiphr, line 89: I personally always have problems with the bitwise operator. Can we use something else here perhaps?

e.g.
```
return subclassValidator.isValidUsage( method, gem.mirror(), targetSubclass ) && isConsistent;
```

or even a simple if

```
if (!subclassValidator.isValidUsage( method, gem.mirror(), targetSubclass )) {
    isConsistent = false;
}
```

filiphr (commented):
Thanks for your changes @Zegveld.

>I did think about this, and came to the following conclusion: Inheriting the inverse mappings only makes sense if both source and target are already unique. The reason for this is that the following behavior does not make sense:

I agree with you that it does not make sense to do that. However, it is up to you to do that, perhaps you have 10 different `SubclassMappings` and 9 out of those are unique, the 10th one is a special one reusing something else due to different reasons, it can be a catch all.

Now if you want to inherit those options your only option is to copy the 9 mappings and add a 10th specific one.

In my opinion we should take the same spirit as the `@Mapping`. You can map from the same source multiple times, when you reverse we allow people to provide different `@Mapping` for some target.

I do understand the change that you pulled the usage of the `SubclassValidator` further up, but I am not really feeling it. There are a lot of places that needed to be changed so we can use this. I will try to check your branch out and see if I can think of something about that

Zegveld:
There is one problem with that, would the inherited mappings come before or after the mappings directly on the method?
Or perhaps they would come in between?
```
@SubclassMapping(target = before.class, ...)
@InheritInverseConfiguration
@SubclassMapping(target = after.class, ...)
```
Because the position is more important for subclassmappings than for normal mappings.

filiphr:
The order of the `@SubclassMapping` is indeed an important one. If you ask me I would say that the order should be in the same place as it was before, unless it is a new not defined target, then it goes to the top (highest priority).

We can wait for feedback from users about this and perhaps even add support for custom ordering through something like `SubclassMapping#dependsOn`, but I would decouple that from this PR. This should behave the same way as it does for `@Mapping`

Zegveld:
> The order of the `@SubclassMapping` is indeed an important one. If you ask me I would say that the order should be in the same place as it was before, unless it is a new not defined target, then it goes to the top (highest priority).

Actually it was easier to make the new order primary, and have any inverse inherited mappings that are redefined skipped.
This will allow you to redefine the entire order if needed, by writing it out. This way you still do not need to rewrite your normal `@Mapping`s just to change the `@SubclassMapping`s order.

for example:
```
@SubclassMapping( source = SourceA.class, target = TargetA.class )
@SubclassMapping( source = SourceB.class, target = TargetB.class )
@SubclassMapping( source = SourceC.class, target = TargetC.class )
@SubclassMapping( source = SourceD.class, target = TargetD.class )
@SubclassMapping( source = SourceA2.class, target = TargetA.class )
Target from(Source source);

@InheritInverseMapping
@SubclassMapping( source = TargetD.class, target = SourceD.class )
@SubclassMapping( source = TargetB.class, target = SourceB.class )
Source from(Target source);
```
will end up as
```
@SubclassMapping( source = TargetD.class, target = SourceD.class )
@SubclassMapping( source = TargetB.class, target = SourceB.class )
@SubclassMapping( source = TargetA.class, target = SourceA.class )
@SubclassMapping( source = TargetC.class, target = SourceC.class )
Source from(Target source);
```

Also added support for overriding a double target issue, by redefining it as seen in above example.

filiphr:
>This way you still do not need to rewrite your normal `@Mappings` just to change the `@SubclassMappings` order.

I thought that this was implied already 😄 , as to me those are 2 independent things. Anyways, I am going through look at your last changes now

processor/src/test/java/org/mapstruct/ap/test/subclassmapping/SubclassMappingTest.java
```diff
@@ -79,6 +95,24 @@ void subclassMappingInheritsMapping() {
...
+        HatchBack.class,
+        InverseOrderSubclassMapper.class
+    } )
+    void subclassMappingOverridesInverseInheritsMapping() {
+        VehicleCollection vehicles = new VehicleCollection();
+        Car car = new Car();
+        car.setVehicleManufacturingCompany( "BenZ" );
+        vehicles.getVehicles().add( car );
+
+        VehicleCollectionDto result = InverseOrderSubclassMapper.INSTANCE.map( vehicles );
```
filiphr, line 109: Don't we want to test the inverse mapping here and check the type of the generated objects?

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
```diff
@@ -46,22 +50,25 @@
     private EnumMappingOptions enumMappingOptions;
     private List<ValueMappingOptions> valueMappings;
     private boolean fullyInitialized;
-    private Set<SubclassMappingOptions> subclassMapping;
+    private Map<TypeMirror, SubclassMappingOptions> subclassMapping;
```
filiphr, line 53: We can't use `TypeMirror` as a key in a Map. The `equals` implementation of `TypeMirror` might return `false` for types that are actually the same. That's why we use `Types#isSameType` to check these things. Can we not use the FQN of the type instead?

filiphr (changes requested):
Went through it again and there are still some things that we need to sort out before we merge this.

Why do we need to the map for tracking the order? Why can't we do something like we do for the `mappingOptions`? I know that the `MappingOptions` have `equals` and `hashCode` implemented, but the `SubclassMappingOptions` do not have that.

Zegveld:
> Went through it again and there are still some things that we need to sort out before we merge this.
> 
> Why do we need to the map for tracking the order? Why can't we do something like we do for the `mappingOptions`? I know that the `MappingOptions` have `equals` and `hashCode` implemented, but the `SubclassMappingOptions` do not have that.

Indeed we do not need the map. Originally I had the map because that one does replace it, while a Set does not replace the same item. But when rewriting I ended up checking if it is already there, and if so skipping it. Which basically becomes the behavior of the Set, so I should refactor this back into a Set and add the equals/hashcode to the SubclassMappingOptions.
This should also fix 2 of the issues with the code you mentioned.

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
Zegveld: Map not needed anymore, implemented an equals on SubclassMappingOptions that uses TypeUtils.isSameType with source MirrorType.

processor/src/test/java/org/mapstruct/ap/test/subclassmapping/SubclassMappingTest.java
Zegveld: Indeed, fixed the tests.

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
```diff
@@ -102,7 +109,7 @@ public EnumMappingOptions getEnumMappingOptions() {
     }
 
     public Set<SubclassMappingOptions> getSubclassMappings() {
-        return subclassMapping;
+        return new LinkedHashSet<>( subclassMapping.values() );
```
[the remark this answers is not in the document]
Zegveld: It is no longer a Map, but a Set so it is now in sync again without creating a new HashSet.

processor/src/test/java/org/mapstruct/ap/test/subclassmapping/SubclassMappingTest.java
```diff
@@ -66,19 +81,38 @@ void existingMappingsAreUsedWhenFound() {
...
+        CarDto carDto = new CarDto();
+        carDto.setMaker( "BenZ" );
+        vehiclesDto.getVehicles().add( carDto );
 
-        VehicleCollectionDto result = SimpleSubclassMapper.INSTANCE.map( vehicles );
+        VehicleCollection result = SimpleSubclassMapper.INSTANCE.mapInverse( vehiclesDto );
 
         assertThat( result.getVehicles() )
-            .extracting( VehicleDto::getMaker )
+            .extracting( Vehicle::getVehicleManufacturingCompany )
```
filiphr, line 94: Don't we care only about the return type? I thought that this is what we are testing? At least that is what we test in the original method

filiphr (commented):
Thanks a lot for your changes @Zegveld. I've added 2 more questions, one is about the testing, since I am confused a bit about it, and the second one is a small style question for a setter

processor/src/test/java/org/mapstruct/ap/test/subclassmapping/SubclassMappingTest.java
Zegveld: That's is already covered in the previous test (line 57). This one is that the `@Mapping` is also inherited by the inverse subclass mapping.

filiphr: You are right, my mistake

filiphr (approved):
Thanks a lot for all the changes @Zegveld. It looks good to me now. Please go ahead with squashing and merging it.

PR #2710, opened 2021-12-27
2709 Corrected description for example demonstrating default expression

### LINKED ISSUE #2709: Correct documentation 10.3 example 77 demonstrating default expression
The documentation describes example 77 as demonstration for two source properties to one target property but in example 77 the usage of a default expression is demonstrated. The same explanation is shown for previous example 75.

Goni-Dev:
Fixes #2709

### DIFF
--- a/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
@@ -88,7 +88,7 @@ Default expressions are a combination of default values and expressions. They wi
 
 The same warnings and restrictions apply to default expressions that apply to expressions. Only Java is supported, and MapStruct will not validate the expression at generation-time.
 
-The example below demonstrates how two source properties can be mapped to one target:
+The example below demonstrates how a default expression can be used to set a value when the source attribute is `null`:
 
 .Mapping method using a default expression
 ====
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
```diff
@@ -88,7 +88,7 @@ Default expressions are a combination of default values and expressions. They wi
 
 The same warnings and restrictions apply to default expressions that apply to expressions. Only Java is supported, and MapStruct will not validate the expression at generation-time.
 
-The example below demonstrates how two source properties can be mapped to one target:
+The example below demonstrates how a default expression can be used to set a value when the source attribute is `null`:
```
filiphr, line 91:
```suggestion
The example below demonstrates how a default expression can be used to set a value when the source attribute is not present (e.g. is `null`):
```

filiphr (changes requested):
Thanks a lot @Goni-Dev.

I've added a suggestion change since the `defaultExpression` is also used when presence checks are used, either through presence checks on the bean itself or using the `@Condition`.

What do you think about my proposal?

Goni-Dev:
Thanks a lot for reviewing. 
I'm totally fine with the proposal.

filiphr:
Thanks @Goni-Dev

PR #2723, opened 2022-01-20
#2719: additional note at builders documentation about before-/aftermapping impact.

### LINKED ISSUE #2719: Clarify documentation about builders and lifecycle `@AfterMapping` / `@BeforeMapping` methods
This is a heads-up for people using Lombok `@Builder` annotation, IT IS NOT a MapStruct issue so please feel free to close immediately, it will at least then be visible to others via a search.

There was an undocumented change (or at the very least a change that is not clearly explained) In lombok version 1.18.22 which caused us issues in an older project when the Springboot version was bumped to 2.6.2 as it included Lombok at 1.18.22. In effect this change has made visible `@Builder` (and possibly/probably other annotations?) to other annotation processors when these annotations were not previously visible.

What does this (possibly) mean for MapStruct users given that MapStruct now has visibility of the `@Builder` annotation?

If you are using `@AfterMapping` in any mapper to map a Lombok class annotated with `@Builder` AND you are NOT passing the builder (as opposed to the mapping target) into the after mapping method, you're after mapping code will no longer be processed.

For reference, the Lombok PR that included this change is [#2718](https://github.com/projectlombok/lombok/pull/2718) and confirmation of the changed behaviour is given in this issue [#3083](https://github.com/projectlombok/lombok/issues/3083).

Discussion around MapStruct and using builders as arguments to after mapping methods can be found at MapStruct issue #1862 

Hopefully this can help others avoid a day of frustration like we just endured in trying to track this down.

Zegveld:
Fixes #2719

### DIFF
--- a/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
@@ -445,6 +445,11 @@ The <<object-factories>> are also considered for the builder type.
 E.g. If an object factory exists for our `PersonBuilder` then this factory would be used instead of the builder creation method.
 ======
 
+[NOTE]
+======
+Detected builders influence `@BeforeMapping` and `@AfterMapping` behavior. See chapter `Mapping customization with before-mapping and after-mapping methods` for more information.
+======
+
 .Person with Builder example
 ====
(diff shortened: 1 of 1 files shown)


filiphr:
Thanks @Zegveld

PR #2727, opened 2022-01-22
#2629 Use ModuleElement when getting type element

### LINKED ISSUE #2629: Detecting 'javax.annotation.processing.Generated' in modular Java does not seem to work

In modular Java, the code generated by mapstruct uses `javax.annotation.processing.Generated`, but the source module does not require module `java.compiler`. Because of this problem, the generated code cannot be compiled.

I found the following description in the documentation, I am not sure how `mapstruct` decides whether to use `javax.annotation.processing.Generated` :

> In Java 9 java.annotation.processing.Generated was added (part of the java.compiler module), if this annotation is available then it will be used.

Is there an alternative of `@Generated` in modular Java?

 JDK version: 11
mapstruct version: 1.4.2.Final
building tool version: gradle 7.2

filiphr:
Prior to this MapStruct would only use `Elements#getTypeElement`.
With this PR if the mapper being generated is within a module MapStruct will use that module
for the methods that are needed (getTypeElement and getPackageElement)

Fixes #2629.

I have to say that I am not the proudest of this piece of code. However, it does it's job. Everything works fine now when running with the module system.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
@@ -6,8 +6,6 @@
 package org.mapstruct.ap.internal.util;
 
 import java.io.Writer;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -15,11 +13,13 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import javax.annotation.processing.ProcessingEnvironment;
+import javax.lang.model.SourceVersion;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
+import javax.lang.model.element.ModuleElement;
 import javax.lang.model.element.Name;
 import javax.lang.model.element.PackageElement;
 import javax.lang.model.element.TypeElement;
@@ -36,91 +36,45 @@ import static javax.lang.model.util.ElementFilter.methodsIn;
 
 public abstract class AbstractElementUtilsDecorator implements ElementUtils {
 
-    private static final Method MODULE_OF_METHOD;
-    private static final Method GET_TYPE_ELEMENT_WITH_MODULE_METHOD;
-    private static final Method GET_PACKAGE_ELEMENT_WITH_MODULE_METHOD;
-
-    static {
-        Method moduleOfMethod;
-        Method getTypeElementWithModuleMethod;
-        Method getPackageElementWithModuleMethod;
-        try {
-            moduleOfMethod = ElementUtils.class.getMethod( "getModuleOf", Element.class );
-            Class<?> moduleElementClass = Class.forName( "javax.lang.model.element.ModuleElement" );
-            getTypeElementWithModuleMethod = ElementUtils.class.getMethod(
-                "getTypeElement",
-                moduleElementClass,
-                CharSequence.class
-            );
-            getPackageElementWithModuleMethod = ElementUtils.class.getMethod(
-                "getPackageElement",
-                moduleElementClass,
-                CharSequence.class
-            );
-
-        }
-        catch ( NoSuchMethodException | ClassNotFoundException e ) {
-            moduleOfMethod = null;
-            getTypeElementWithModuleMethod = null;
-            getPackageElementWithModuleMethod = null;
-        }
-
-        MODULE_OF_METHOD = moduleOfMethod;
-        GET_TYPE_ELEMENT_WITH_MODULE_METHOD = getTypeElementWithModuleMethod;
-        GET_PACKAGE_ELEMENT_WITH_MODULE_METHOD = getPackageElementWithModuleMethod;
-    }
-
     private final Elements delegate;
     /**
-     * The module element when running with the module system.
+     * The module element when running with the module system,
      * {@code null} otherwise.
      */
     private final Element moduleElement;
 
+    @IgnoreJRERequirement
     AbstractElementUtilsDecorator(ProcessingEnvironment processingEnv, TypeElement mapperElement) {
         this.delegate = processingEnv.getElementUtils();
-        Element moduleElement;
-        if ( MODULE_OF_METHOD == null ) {
-            moduleElement = null;
+        if ( SourceVersion.RELEASE_8.compareTo( processingEnv.getSourceVersion() ) <= 0 ) {
+            // We are running with Java 8 or lower
+            this.moduleElement = null;
         }
         else {
-            try {
-                moduleElement = (Element) MODULE_OF_METHOD.invoke( this.delegate, mapperElement );
-            }
-            catch ( IllegalAccessException | InvocationTargetException e ) {
-                moduleElement = null;
-            }
+            this.moduleElement = this.delegate.getModuleOf( mapperElement );
         }
-
-        this.moduleElement = moduleElement;
     }
 
     @Override
+    @IgnoreJRERequirement
     public PackageElement getPackageElement(CharSequence name) {
-        if ( moduleElement == null ) {
-            return delegate.getPackageElement( name );
+        if ( this.moduleElement == null ) {
+            return this.delegate.getPackageElement( name );
         }
 
-        try {
-            return (PackageElement) GET_PACKAGE_ELEMENT_WITH_MODULE_METHOD.invoke( delegate, moduleElement, name );
-        }
-        catch ( IllegalAccessException | InvocationTargetException e ) {
-            return delegate.getPackageElement( name );
-        }
+        // If the module element is not null then we must be running on Java 8+
+        return this.delegate.getPackageElement( (ModuleElement) moduleElement, name );
     }
 
     @Override
+    @IgnoreJRERequirement
     public TypeElement getTypeElement(CharSequence name) {
-        if ( moduleElement == null ) {
-            return delegate.getTypeElement( name );
+        if ( this.moduleElement == null ) {
+            return this.delegate.getTypeElement( name );
         }
 
-        try {
-            return (TypeElement) GET_TYPE_ELEMENT_WITH_MODULE_METHOD.invoke( delegate, moduleElement, name );
-        }
-        catch ( IllegalAccessException | InvocationTargetException e ) {
-            return delegate.getTypeElement( name );
-        }
+        // If the module element is not null then we must be running on Java 8+
+        return this.delegate.getTypeElement( (ModuleElement) moduleElement, name );
     }
 
     @Override
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
@@ -6,6 +6,8 @@
 package org.mapstruct.ap.internal.util;
 
 import java.io.Writer;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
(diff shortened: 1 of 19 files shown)


processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
```diff
@@ -34,20 +36,91 @@
 
 public abstract class AbstractElementUtilsDecorator implements ElementUtils {
 
+    private static final Method MODULE_OF_METHOD;
+    private static final Method GET_TYPE_ELEMENT_WITH_MODULE_METHOD;
+    private static final Method GET_PACKAGE_ELEMENT_WITH_MODULE_METHOD;
+
+    static {
```
sjaakd, line 43: why not use the java version? >8 this method is available. == 8 not. I'm not in favour of using reflection if that's not strictly needed. So.. you could deduct the java version here (I guess when we've some logic already when we write the @Generated annotation).

You'll need to compile MapStruct to java9. There are some questions arising from this: what would happen if you use the mapstruct processor compiled to 9 in an javac 8 compiler (runtime)?

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
```diff
@@ -34,20 +36,91 @@
...
+     * The module element when running with the module system.
+     * {@code null} otherwise.
+     */
+    private final Element moduleElement;
 
-    AbstractElementUtilsDecorator(ProcessingEnvironment processingEnv) {
+    AbstractElementUtilsDecorator(ProcessingEnvironment processingEnv, TypeElement mapperElement) {
         this.delegate = processingEnv.getElementUtils();
+        Element moduleElement;
+        if ( MODULE_OF_METHOD == null ) {
```
sjaakd, line 83: this would then be a java version check. can be a static initialised boolean.

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
```diff
@@ -34,20 +36,91 @@
...
     }
 
     @Override
     public PackageElement getPackageElement(CharSequence name) {
-        return delegate.getPackageElement( name );
+        if ( moduleElement == null ) {
+            return delegate.getPackageElement( name );
+        }
+
+        try {
```
sjaakd, line 104: same here

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
```diff
@@ -34,20 +36,91 @@
...
     }
 
     @Override
     public TypeElement getTypeElement(CharSequence name) {
-        return delegate.getTypeElement( name );
+        if ( moduleElement == null ) {
+            return delegate.getTypeElement( name );
+        }
+
+        try {
```
sjaakd, line 118: same here..

sjaakd (changes requested):
I'm not in favour of using reflection for java versions > 8 for each call to resolve a type. I guess this will downgrade performance considerably. But perhaps I'm wrong.

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
filiphr: This is not only about the `@Generated` notation. It is also about types that are available or not, they might be available during the annotation processing, but not in the actual module of the mapper (if someone uses module-info.java).

In order for us to be able to use the methods during compile time then we will need to either use a Multi Release Jar where we will offload this or change the baseline of MapStruct.

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
filiphr: We still need to invoke the method somehow, we can't compile with target 8 and invoke a Java 9 method.

filiphr:
>I'm not in favour of using reflection for java versions > 8 for each call to resolve a type. I guess this will downgrade performance considerably. But perhaps I'm wrong.

There have been significant improvements done in reflection in newer Java versions, especially in Java 17. 
I have not measured the performance hit here. However, the reflection methods for each type will be used only if the user is using modules, otherwise the retrieval of the `moduleElement` returns `null` which means that will will do the first check only. 

I can look into using multi release jars if you think it might be better.

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
filiphr: The support for Jakarta in #2738 might also be affected by this. e.g. you might have both javax and jakarta available, but in your module descriptor you are only using jakarta, in this case this new mechanism will properly see only the jakarta packages.

sjaakd: >> This is not only about the @Generated notation

I understand that. I was just referring to the code that establishes the runtime java version. 

So in short:
* what would happen if we go to java9 for compiling the processor, especially when use the processor in a javac v8? Would that work if we don't call the java9 methods?
* So in java9 the required methods on `Elements` are available. We only should not call those in a javac 8 environment. 
* we could use the static part to establish if we run the processor in java>8 or java==8. The checks in the methods would than boil down to selecting the proper method.

sjaakd:
> I can look into using multi release jars if you think it might be better.

I would not prefer that. But see my points above. Can we compile the processor to java9? What would happen if we run the processor on a javac 8? Is that at all possible?

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
filiphr: >what would happen if we go to java9 for compiling the processor, especially when use the processor in a javac v8? Would that work if we don't call the java9 methods?

In order to achieve what you are saying we need to change the baseline, which means that the bytecode will have a newer byte code version and you won't be able to run the processor on javac8.

There are 3 options for we supporting Java Modules properly:

* Use reflection like it is done in this PR and like we do for Records
* Use Multi Release Jar and have a different implementation for Java 9+ for this particular class
* Change our baseline to Java 11 or 17. This option is out of the question for now, perhaps in 2.0 we can change our baseline as well. We could still target the generated code to be correct on Java 8, but you would have to use our baseline to run the processor.

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
sjaakd: Are you sure about that? I thought that `target 8` means that it compiles target 8 compatible byte-code. I don't think the compiler actively checks the if a method (with a since java9) is present or not.  What would otherwise be the idea of having a `target 8` decoupled from a `source 8`? . Also, the classes in `javax.lang.model` are an extension to java and not the core right?

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
sjaakd: I agree that we should not change the entire baseline. Java 8 is too much in use right now.

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
filiphr: OK I did some analysis. If we only use `source` and `target` we can make this work. However, if we switch to using the new `release` flag then there will be a compile error (same error as we would run the compilation with Java 8). This does mean that we have to build the processor module with at least java 9.

> Also, the classes in `javax.lang.model` are an extension to java and not the core right?

What do you mean an extension? They are part of the OpenJDK.

In any case if you are really against the reflection then I would prefer to use [Multi-Release JAR](https://openjdk.java.net/jeps/238). Keep in mind that this is not new artifact, we are still going to provide a single dependency. Since Java 9 you can provide different implementations of certain classes for different Java versions in the same JAR under `META-INF/versions/<javaVersion>`

sjaakd: I mean.. records an lamda's are java core features (probably implemented in native  byte-code for that matter)

The javax stuff isn't, right?

sjaakd: so my proposal would be: go to a higher java version but set the target bytecode to 8. I think you still would be able to call the methods if you run inside 9 java (you compiled to 9 level source). But it would also generate javac 8 compatible byte code since the target level is 8. But it could be that I'm all wrong about the sof course 😄 . Never hurts to try.

filiphr:
I wrote in the comment, but also want to write here.

Not sure what you think a Multi-Release JAR. This is not a new artifact, we are still going to provide a single dependency. Since Java 9 you can provide different implementations of certain classes for different Java versions in the same JAR under META-INF/versions/<javaVersion>. Have a look at [JEP 238: Multi-Release JAR FIles](https://openjdk.java.net/jeps/238) for more information.

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
filiphr: I tried it already and we only use `source` and `target` we can achieve what we want. We'll need to make sure our tests setup runs properly. Let me push that here.

filiphr:
@sjaakd I've done the changes not to use reflection. Can you have a look and let me know what you think

processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
```diff
@@ -36,91 +36,45 @@
...
-        }
-        catch ( IllegalAccessException | InvocationTargetException e ) {
-            return delegate.getPackageElement( name );
-        }
+        // If the module element is not null then we must be running on Java 8+
+        return this.delegate.getPackageElement( (ModuleElement) moduleElement, name );
     }
 
     @Override
+    @IgnoreJRERequirement
```
sjaakd, line 70: cool.. 👍 didn't know that one existed.

processor/src/main/java/org/mapstruct/ap/internal/util/IgnoreJRERequirement.java
```diff
@@ -0,0 +1,16 @@
...
+package org.mapstruct.ap.internal.util;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.TYPE })
+public @interface IgnoreJRERequirement {
```
sjaakd, line 15: Ok.. we defined it ourselves.. 😄

sjaakd (approved):

sjaakd:
>> Since Java 9 you can provide different implementations of certain classes for different Java versions in the same JAR under META-INF/versions/<javaVersion>. Have a look at JEP 238: Multi-Release JAR FIles for more information.

Ah.. I read that a while ago but forgot about that. Thanks

sjaakd:
Just out of curiosity: where do you define the java versions now? I would have expected somewhere in the maven compile plugin?

filiphr:
>Just out of curiosity: where do you define the java versions now? I would have expected somewhere in the maven compile plugin?

The java version (`target` and `source`) is defined in the parent `pom.xml` with `1.8`. The current changes are changing the minimum java version to build the `processor` module to be Java 11 and adapt the `animal-sniffer-maven-plugin` to ignore the fact that we are using Java 9 methods.

>Did you go for the multi-release-jar and where is that configured? The yaml only tells CI how to build / how to run tests. But I miss where we actually define stuff.

No I haven't gone with the Multi-Release JAR yet. The last approach is the one you proposed, we build the processor with Java 11 with target and source of Java 8. The GitHub YAML has been updated to be able to run the integration tests with Java 8 in order for the `fullFeature` test to run and we to be able to validate that everything runs properly when our users are compiling with Java 8.

sjaakd:
> No I haven't gone with the Multi-Release JAR yet. The last approach is the one you proposed, we build the processor with Java 11 with target and source of Java 8. The GitHub YAML has been updated to be able to run the integration tests with Java 8 in order for the `fullFeature` test to run and we to be able to validate that everything runs properly when our users are compiling with Java 8.

Cool.. Perhaps we should do the `record` implementation in the future like this as well..

But by all means.. If you agree yourself 😄 , merge..

filiphr:
Thanks for the review @sjaakd, it helped iron it out and find a better solution to this.

PR #2729, opened 2022-01-22
#2728 Add new WithTestDependency annotation for our processor testing

### LINKED ISSUE #2728: Support defining which test dependencies should be available during testing
In order to be able to test things such as #2567 we need to be able to dynamically add test dependencies to the testing compile path.

filiphr:
Adding this dependency allows us to dynamically pick the dependencies that we want to have on the test compilation classpath.
It would allow us to more granularly test things with different dependencies, such as javax inject and jakarta inject

Fixes #2728


sjaakd (approved):
cool 👍  please go ahead. Nice idea.

filiphr:
Thanks @sjaakd. Apart from us needing this, this could also be another step into completely releasing a dedicated compile testing library 😉

PR #2731, opened 2022-01-23
#1661 Add support for globally disabling builders

### LINKED ISSUE #1661: Opt-out  of using builders via a processor option

filiphr:
Fixes #1661

### DIFF
--- a/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
@@ -532,7 +532,7 @@ Otherwise, you would need to write a custom `BuilderProvider`
 
 [TIP]
 ====
-In case you want to disable using builders then you can use the `NoOpBuilderProvider` by creating a `org.mapstruct.ap.spi.BuilderProvider` file in the `META-INF/services` directory with `org.mapstruct.ap.spi.NoOpBuilderProvider` as it's content.
+In case you want to disable using builders then you can set pass the MapStruct processor option `mapstruct.disableBuilders` to the compiler. e.g. `-Amapstruct.disableBuilders=true`.
 ====
 
 [[mapping-with-constructors]]
--- a/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
@@ -532,7 +532,7 @@ Otherwise, you would need to write a custom `BuilderProvider`
 
 [TIP]
 ====
-In case you want to disable using builders then you can set pass the MapStruct processor option `mapstruct.disableBuilders` to the compiler. e.g. `-Amapstruct.disableBuilders=true`.
+In case you want to disable using builders then you can pass the MapStruct processor option `mapstruct.disableBuilders` to the compiler. e.g. `-Amapstruct.disableBuilders=true`.
 ====
 
 [[mapping-with-constructors]]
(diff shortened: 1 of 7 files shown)


documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
```diff
@@ -532,7 +532,7 @@ Otherwise, you would need to write a custom `BuilderProvider`
 
 [TIP]
 ====
-In case you want to disable using builders then you can use the `NoOpBuilderProvider` by creating a `org.mapstruct.ap.spi.BuilderProvider` file in the `META-INF/services` directory with `org.mapstruct.ap.spi.NoOpBuilderProvider` as it's content.
+In case you want to disable using builders then you can set pass the MapStruct processor option `mapstruct.disableBuilders` to the compiler. e.g. `-Amapstruct.disableBuilders=true`.
```
sjaakd, line 535: I guess 'set pass' should be 'pass' or 'set'

sjaakd (approved):
I thought we were going to deprecate options in the future in favour of global config.

documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
filiphr: Good spotting this, yes it should be one of those. I'll fix it

filiphr:
>I thought we were going to deprecate options in the future in favour of global config.

I think that this is something that we should investigate when we start working towards 2.0

PR #2733, opened 2022-01-23
#1997 Use builders to construct empty objects in update wrapper

### LINKED ISSUE #1997: MapStruct does not always use builders in target-object
Hi,

we try to use MapStruct with Lombok builders but MapStruct does **not always** seem to recognize and use the builders. Our particular problematic case is when the **target** object has a field which only has a builder.

Used MapStruct version: 1.3.1.FINAL
Used Lombok version: 1.18.10

It would be very helpful for us if you could let us know if this is considered a bug or not.
Looking forward to your feedback!

**Here is an example:**

Source POJO:
```
package example.model.source;

import lombok.Builder;
import lombok.Data;

@Data
@Builder
public class Car {
	private String model;
}
```

Target POJOs:
```
package example.model.target;

import lombok.Builder;
import lombok.Data;

@Data
@Builder
public class CarInsurance {
	private CarDetail carDetail;   // this is the problematic case
}

@Data
@Builder
public class CarDetail {
	private String model;
}
```

Mapper:
```
package example.mapper;

import example.model.source.Car;
import example.model.target.CarInsurance;
import org.mapstruct.Mapper;
import org.mapstruct.Mapping;

@Mapper
public interface CarInsuranceMapper {
	@Mapping(source = "model", target = "carDetail.model")
	void map(Car source, @MappingTarget CarInsurance target);
}
```

MapStruct generates this mapper:
```
package example.mapper;

import example.model.source.Car;
import example.model.target.CarDetail;
import example.model.target.CarInsurance;
import javax.annotation.Generated;
import org.springframework.stereotype.Component;

@Generated(
    value = "org.mapstruct.ap.MappingProcessor"
)
@Component
public class CarInsuranceMapperImpl implements CarInsuranceMapper {

    @Override
    public void map(Car source, CarInsurance target) {
        if ( source == null ) {
            return;
        }

        if ( target.getCarDetail() == null ) {
            target.setCarDetail( new CarDetail() );  // THIS CREATES A COMPILE ERROR
        }
        carToCarDetail( source, target.getCarDetail() );
    }

    protected void carToCarDetail(Car car, CarDetail mappingTarget) {
        if ( car == null ) {
            return;
        }

        mappingTarget.setModel( car.getModel() );
    }
}
```

And the compiler fails with:
```
CarInsuranceMapperImpl.java:22: error: constructor CarDetail in class CarDetail cannot be applied to given types;
            target.setCarDetail( new CarDetail() );

```

It would work if MapStruct would generate:
```
if ( target.getCarDetail() == null ) {
    target.setCarDetail( CarDetail.builder().build() );
}
```

The annotation processor ordering in the Gradle build:
```
	implementation 'org.mapstruct:mapstruct'
	implementation 'org.projectlombok:lombok'
	annotationProcessor 'org.mapstruct:mapstruct-processor'
	annotationProcessor 'org.projectlombok:lombok'
```

filiphr:
Fixes #1997

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/MethodReference.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/MethodReference.java
@@ -6,6 +6,7 @@
 package org.mapstruct.ap.internal.model;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
@@ -58,8 +59,10 @@ public class MethodReference extends ModelElement implements Assignment {
     private final Type definingType;
     private final List<ParameterBinding> parameterBindings;
     private final Parameter providingParameter;
+    private final List<MethodReference> methodsToChain;
     private final boolean isStatic;
     private final boolean isConstructor;
+    private final boolean isMethodChaining;
 
     /**
      * Creates a new reference to the given method.
@@ -95,6 +98,8 @@ public class MethodReference extends ModelElement implements Assignment {
         this.isStatic = method.isStatic();
         this.name = method.getName();
         this.isConstructor = false;
+        this.methodsToChain = Collections.emptyList();
+        this.isMethodChaining = false;
    }
 
     private MethodReference(BuiltInMethod method, ConversionContext contextParam) {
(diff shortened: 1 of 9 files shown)


sjaakd (approved):
Factory methods accept source parameters (I believe). However, I don't think we use them in property mappings, right? (Only for constructing a return type).  The case that you test is creating a default empty object with a builder. 

Would it be interesting to se what happens if we do accept a source parameter (for the non default case)? I guess that would be a different PR, right? The use case would be an extension of this one though. Jus as using object factories for property mappings.

filiphr:
>Factory methods accept source parameters (I believe). However, I don't think we use them in property mappings, right? (Only for constructing a return type). The case that you test is creating a default empty object with a builder.

Yes they do access source parameters and they still do. This is done in line 420 in `PropertyMapping`. However, if there is no factory method to construct an object then we do something similar like we do in the `BeanMappingMethod`. 

The change that this PRs is doing is to use the builder if there is one for constructing the target object for update mappings.

PR #2736, opened 2022-01-29
#2682 Change RetentionPolicy of `@DecoratedWith` to `CLASS`

### LINKED ISSUE #2682: `@DecoratedWith` does not work with incremental compilation in Gradle 7.3
Demo project for this issue: https://github.com/laech/mapstruct-spring-gradle-incremental-issue

If you checkout the demo project, it contains a mapper and a decorator:

```java
@DecoratedWith(PersonMapperDecorator.class)
interface PersonMapper {
}
```

```java
abstract class PersonMapperDecorator implements PersonMapper {

  @Autowired
  @Qualifier("delegate")

  @Override
  public PersonDto personToPersonDto(Person person) {
    PersonDto dto = delegate.personToPersonDto(person);
    dto.name += "!!";
    return dto;
  }
}
```

After initial compilation via `./gradlew compileJava` and inspecting the generated sources under `build/generated/sources/annotationProcessor/java/main/com/example/demo` we should have a correctly generated mapper and its decorator:

```java
@Generated(
    date = "2021-12-08T14:16:55+1300",
    comments = "version: 1.5.0.Beta1, compiler: IncrementalProcessingEnvironment from gradle-language-java-7.3.1.jar, environment: Java 17 (Private Build)"
)
@Component
@Primary
class PersonMapperImpl extends PersonMapperDecorator {
}
```

```java
@Generated(
    date = "2021-12-08T14:02:40+1300",
    comments = "version: 1.5.0.Beta1, compiler: IncrementalProcessingEnvironment from gradle-language-java-7.3.1.jar, environment: Java 17 (Private Build)"
)
@Component
@Qualifier("delegate")
class PersonMapperImpl_ implements PersonMapper {

    @Override
    public PersonDto personToPersonDto(Person person) {
        if ( person == null ) {
            return null;
        }
        PersonDto personDto = new PersonDto();
        personDto.name = person.name;
        return personDto;
    }
}
```

Now make a change in `PersonMapperDecorator.java`, such as adding a random comment, then compile again with `./gradlew compileJava` and the generated classes become:

```java
@Generated(
    date = "2021-12-08T14:14:11+1300",
    comments = "version: 1.5.0.Beta1, compiler: IncrementalProcessingEnvironment from gradle-language-java-7.3.1.jar, environment: Java 17 (Private Build)"
)
@Component
class PersonMapperImpl implements PersonMapper {

    @Override
    public PersonDto personToPersonDto(Person person) {
        if ( person == null ) {
            return null;
        }
        PersonDto personDto = new PersonDto();
        personDto.name = person.name;
        return personDto;
    }
}
```

```java
@Generated(
    date = "2021-12-08T14:10:35+1300",
    comments = "version: 1.5.0.Beta1, compiler: IncrementalProcessingEnvironment from gradle-language-java-7.3.1.jar, environment: Java 17 (Private Build)"
)
@Component
@Qualifier("delegate")
class PersonMapperImpl_ implements PersonMapper {

    @Override
    public PersonDto personToPersonDto(Person person) {
        if ( person == null ) {
            return null;
        }
        PersonDto personDto = new PersonDto();
        personDto.name = person.name;
        return personDto;
    }
}
```

The problem is the new `PersonMapperImpl` now contains the wrong code. A workaround is to rebuild the whole project, which is problematic for large projects.

**Update:**

This issue is reproducible on Gradle 7.3, but not on Gradle 7.2.

filiphr:
In some circumstances (used with other types of aggregating processors, e.g. Spring)
the Gradle incremental compilation works correctly only for classes annotated with the `CLASS` or `RUNTIME`
retention policy.

The `@DecoratedWith` is the only annotation from MapStruct that was `SOURCE` retention.
With this commit we are changing its retention policy in order for better compatibility with the Gradle Incremental compilation

Fixes #2682

### DIFF
--- a/core/src/main/java/org/mapstruct/DecoratedWith.java
+++ b/core/src/main/java/org/mapstruct/DecoratedWith.java
@@ -145,7 +145,7 @@ import java.lang.annotation.Target;
  * @author Gunnar Morling
  */
 @Target(ElementType.TYPE)
-@Retention(RetentionPolicy.SOURCE)
+@Retention(RetentionPolicy.CLASS)
 public @interface DecoratedWith {
 
     /**


core/src/main/java/org/mapstruct/DecoratedWith.java
```diff
@@ -145,7 +145,7 @@
  * @author Gunnar Morling
  */
 @Target(ElementType.TYPE)
-@Retention(RetentionPolicy.SOURCE)
+@Retention(RetentionPolicy.CLASS)
```
sjaakd, line 148: I see this is one of the last ones that does not have the retention policy 'CLASS'.  Only 'Experimental'  remaining.

I guess a more important reason is that some annotations need to be referenced in libraries. Although, that's probably not the case for `@DecoratedWith` but also not for `@Mapper`. I forgot why runtime was not sufficient for the latter one, but I guess`@DecoratedWith` needs to be in line with `@Mapper`

sjaakd (approved):

core/src/main/java/org/mapstruct/DecoratedWith.java
filiphr: `Experimental` is indeed only `SOURCE`, but that one is only for documentation, so that is fine.

`@Mapper` might actually be used in other libraries, e.g. if you use other mappers via `Mapper#uses`. `RUNTIME` is not needed since we don't want our annotations to be present during the runtime execution.

filiphr:
Thanks for the review @sjaakd

PR #2737, opened 2022-01-29
#2677 Use type without bounds when looking for read /  presence accessor in a SourceReference

### LINKED ISSUE #2677: Cannot map from a covariant (extends) generic type

I am trying to use a covariant generic type (`<? extends X>`) in the mapping parameter and the result is

```
Internal error in the mapping processor: java.lang.NullPointerException  
at org.mapstruct.ap.internal.util.workarounds.SpecificCompilerWorkarounds.replaceTypeElementIfNecessary(SpecificCompilerWorkarounds.java:99)  
at org.mapstruct.ap.internal.util.Executables.getAllEnclosedExecutableElements(Executables.java:98)  
at org.mapstruct.ap.internal.model.common.Type.getAllMethods(Type.java:665)  
at org.mapstruct.ap.internal.model.common.Type.getPropertyReadAccessors(Type.java:517)  
at org.mapstruct.ap.internal.model.beanmapping.SourceReference.push(SourceReference.java:423)  
at org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.applyTargetThisMapping(BeanMappingMethod.java:1207)  
at org.mapstruct.ap.internal.model.BeanMappingMethod$Builder.build(BeanMappingMethod.java:268)  
at org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMappingMethods(MapperCreationProcessor.java:378)  
at org.mapstruct.ap.internal.processor.MapperCreationProcessor.getMapper(MapperCreationProcessor.java:152)  
at org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:123)  
at org.mapstruct.ap.internal.processor.MapperCreationProcessor.process(MapperCreationProcessor.java:74)  
at org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:338)  
at org.mapstruct.ap.MappingProcessor.processMapperTypeElement(MappingProcessor.java:318)  
at org.mapstruct.ap.MappingProcessor.processMapperElements(MappingProcessor.java:267)  
at org.mapstruct.ap.MappingProcessor.process(MappingProcessor.java:166)  
at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:980)  
at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:896)  
at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1222)  
at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1334)  
at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.processAnnotations(JavaCompiler.java:1258)  
at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:936)  
at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.lambda$doCall$0(JavacTaskImpl.java:104)  
at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.handleExceptions(JavacTaskImpl.java:147)  
at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.doCall(JavacTaskImpl.java:100)  
at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.call(JavacTaskImpl.java:94)  
at org.codehaus.plexus.compiler.javac.JavaxToolsCompiler.compileInProcess(JavaxToolsCompiler.java:126)  
at org.codehaus.plexus.compiler.javac.JavacCompiler.performCompile(JavacCompiler.java:174)  
at org.apache.maven.plugin.compiler.AbstractCompilerMojo.execute(AbstractCompilerMojo.java:1134)  
at org.apache.maven.plugin.compiler.CompilerMojo.execute(CompilerMojo.java:187)  
at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:137)  
at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:210)  
at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:156)  
at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:148)  
at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:117)  
at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call(MultiThreadedBuilder.java:190)  
at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call(MultiThreadedBuilder.java:186)  
at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)  
at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)  
at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)  
at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)  
at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)  
at java.base/java.lang.Thread.run(Thread.java:834)
```

The simplified code that produces this is:
```java
@Mapper
public interface GenericMapper {
    @Mapping(target = ".", source = "value")
    Output map(Wrapper<? extends Parent> in); // Internal error in the mapping processor

    @Mapping(target = ".", source = "value")
    Output mapFromParent(Wrapper<Parent> in); // works

    @Mapping(target = ".", source = "value")
    Output mapFromChild(Wrapper<Child> in); // works
}
```
with classes
```java
class Wrapper<T> {
    private boolean status;
    private T value;

    public void setStatus(boolean status) { this.status = status; }
    public void setValue(T value) { this.value = value; }
    public boolean isStatus() { return status;}
    public T getValue() { return value; }
}

class Parent {
    private int id;

    public int getId() { return id; }
    public void setId(int id) { this.id = id; }
}

class Child extends Parent {
    private String whatever;

    public String getWhatever() { return whatever; }
    public void setWhatever(String whatever) { this.whatever = whatever; }
}

class Output {
    int id;
    boolean status;

    public int getId() { return id; }
    public void setId(int id) { this.id = id; }
    public boolean isStatus() { return status; }
    public void setStatus(boolean status) { this.status = status; }
}
```

Mapstruct: 1.4.2.Final
JDK: Amazon Coretto 11.0.4
Maven: 3.6.3
IntelliJ Idea: 2021.2.3 Ultimate

filiphr:
Fixes #2677

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
@@ -311,11 +311,12 @@ public class SourceReference extends AbstractReference {
             Type newType = type;
             for ( int i = 0; i < entryNames.length; i++ ) {
                 boolean matchFound = false;
-                ReadAccessor readAccessor = newType.getReadAccessor( entryNames[i] );
+                Type noBoundsType = newType.withoutBounds();
+                ReadAccessor readAccessor = noBoundsType.getReadAccessor( entryNames[i] );
                 if ( readAccessor != null ) {
-                    PresenceCheckAccessor presenceChecker = newType.getPresenceChecker( entryNames[i] );
+                    PresenceCheckAccessor presenceChecker = noBoundsType.getPresenceChecker( entryNames[i] );
                     newType = typeFactory.getReturnType(
-                        (DeclaredType) newType.getTypeMirror(),
+                        (DeclaredType) noBoundsType.getTypeMirror(),
                         readAccessor
                     );
                     sourceEntries.add( forSourceReference(
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2677/Issue2677Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2677/Issue2677Mapper.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2677;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper
+public interface Issue2677Mapper {
+
+    Issue2677Mapper INSTANCE = Mappers.getMapper( Issue2677Mapper.class );
+
+    @Mapping(target = "id", source = "value.id")
+    Output map(Wrapper<? extends Parent> in);
+
+    @Mapping(target = ".", source = "value")
+    Output mapImplicitly(Wrapper<? extends Parent> in);
+
+    @Mapping(target = "id", source = "value.id")
+    Output mapFromParent(Wrapper<Parent> in);
+
+    @Mapping(target = "id", source = "value.id")
+    Output mapFromChild(Wrapper<Child> in);
+
+    class Wrapper<T> {
+        private final T value;
+        private final String status;
+
+        public Wrapper(T value, String status) {
+            this.value = value;
+            this.status = status;
+        }
+
+        public String getStatus() {
+            return status;
+        }
+
+        public T getValue() {
+            return value;
+        }
+    }
+
+    class Parent {
+        private final int id;
+
+        public Parent(int id) {
+            this.id = id;
+        }
+
+        public int getId() {
+            return id;
+        }
+    }
+
+    class Child extends Parent {
+        private final String whatever;
+
+        public Child(int id, String whatever) {
+            super( id );
+            this.whatever = whatever;
+        }
+
+        public String getWhatever() {
+            return whatever;
+        }
+    }
+
+    class Output {
+        private int id;
+        private String status;
+
+        public int getId() {
+            return id;
+        }
+
+        public void setId(int id) {
+            this.id = id;
+        }
+
+        public String getStatus() {
+            return status;
+        }
+
+        public void setStatus(String status) {
+            this.status = status;
+        }
+    }
+}
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2677/Issue2677Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2677/Issue2677Mapper.java
@@ -29,6 +29,12 @@ public interface Issue2677Mapper {
     @Mapping(target = "id", source = "value.id")
     Output mapFromChild(Wrapper<Child> in);
 
+    @Mapping( target = "value", source = "wrapperValue")
+    Wrapper<String> mapToWrapper(String wrapperValue, Wrapper<? super Parent> wrapper);
+
+    @Mapping(target = "id", source = "value.id")
+    Output mapWithPresenceCheck(Wrapper<? extends ParentWithPresenceCheck> in);
+
     class Wrapper<T> {
         private final T value;
         private final String status;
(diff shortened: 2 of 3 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
```diff
@@ -311,11 +311,12 @@ private void reportErrorOnNoMatch( Parameter parameter, String[] propertyNames,
             Type newType = type;
             for ( int i = 0; i < entryNames.length; i++ ) {
                 boolean matchFound = false;
-                ReadAccessor readAccessor = newType.getReadAccessor( entryNames[i] );
+                Type noBoundsType = newType.withoutBounds();
```
sjaakd, line 314: I'm trying to grasp what is happening here. So I did setup a breakpoint on when the presenceChecker of the noBoundsType differs from the newType presenceChecker. It's  not hit.  I'm doing something wrong here. Can you explain why there's a diff?

filiphr: I didn't add a test case for the presence checker, but the test case fails if we don't use `withoutBounds` for getting the `ReadAccessor`

sjaakd: I don't think that we do this at the right place. You are now testing with a upper bound. Did you try using a lower bound?

So in your test case:

```java
    @Mapping(target = "id", source = "value.id")
    Output map(Wrapper<? super Parent> in);

    @Mapping(target = ".", source = "value")
    Output mapImplicitly(Wrapper<? super Parent> in);
```

You probably need to change your inheritance structure for that.

The point is: you change it at the spot where we define the `SourceReference`. We should probably check it at the spot where we try to make a match to the getter respectively the presence checker.

filiphr: >I don't think that we do this at the right place. You are now testing with a upper bound. Did you try using a lower bound?

We do not allow using lower bounds for source parameters. At least not when there is a single parameter. Which means that for the example above there would be a compile error. 

I'll try to come up with some more tests where we do implicit without a `target = "."`

filiphr: I've added some more tests, not sure if it is what you were looking for. They still work without my changes. There can be other corner cases like auto generating from things like `Wrapper<? extends SomeType>`, but I don't think that this belong to this issue

sjaakd: Ok.. then go ahead and merge. I was indeed looking for cases that failed. 

And  I was thinking how this code ties in to the method selection mechanism (I did not investigate that). But it looked like you were making a kind of pre-selection / pre-filter (so limiting the cases that you "feed"  to the method selection prior to calling method selection later on). If that's the case: then this should be fixed there (in method selection) and not in source reference. Hope this clarifies my comment.

sjaakd (approved):

processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
filiphr: Yes it does make sense. However, the problem was not about method selection, the problem was that the we were not getting any properties for the wild card type.

PR #2738, opened 2022-01-29
#2567 Add support for Jakarta Injection

### LINKED ISSUE #2567: Add support for Jakarta Dependency Injection API

Hi,

I think it would be great if Mapstruct would support the [Jakarta dependency inject annotations](https://jakarta.ee/specifications/dependency-injection/2.0/):

* jakarta.inject.Inject
* jakarta.inject.Singleton
* jakarta.inject.Named

This would would make generated mappers compatible with Jakarta EE 9, Micronaut 3, etc.  

Could this be as simple as copying the Jsr330 component model with replaced annotation names?

filiphr:
Support for Jakarta is done in 2 ways:

* current jsr330 component model - In this case Jakarta Inject will be used if javax.inject is not available
* new jakarta component model - In this case Jakarta Inject will always be used

Fixes #2567

### DIFF
--- a/core/src/main/java/org/mapstruct/DecoratedWith.java
+++ b/core/src/main/java/org/mapstruct/DecoratedWith.java
@@ -103,12 +103,12 @@ import java.lang.annotation.Target;
  * private PersonMapper personMapper; // injects the decorator, with the injected original mapper
  * </pre>
  *
- * <h3>3. Component model 'jsr330'</h3>
+ * <h3>3. Component model 'jsr330' or 'jakarta'</h3>
  * <h4>Referencing the original mapper</h4>
  * <p>
- * JSR 330 doesn't specify qualifiers and only allows to specifically name the beans. Hence, the generated
- * implementation of the original mapper is annotated with
- * {@code @javax.inject.Named("fully-qualified-name-of-generated-impl")} and {@code @Singleton} (please note that when
+ * JSR 330 / Jakarta Inject doesn't specify qualifiers and only allows to specifically name the beans. Hence,
+ * the generated implementation of the original mapper is annotated with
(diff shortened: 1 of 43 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/Jsr330ComponentProcessor.java
```diff
@@ -58,21 +60,35 @@ protected boolean requiresGenerationOfDecoratorClass() {
...
             Collections.singletonList( '"' + mapper.getPackageName() + "." + mapper.getName() + '"' )
         );
     }
 
     private Annotation inject() {
-        return new Annotation( getTypeFactory().getType( "javax.inject.Inject" ) );
+        return new Annotation( getType( "Inject" ) );
+    }
+
+    private Type getType(String simpleName) {
```
sjaakd, line 81: I noticed that this implementation differs from the `JakartaComponentProcessor`. Perhaps factor this method out as static (util) to be used in both the `JSR330ComponentProcessor` and this one.

sjaakd (changes requested):

processor/src/main/java/org/mapstruct/ap/internal/processor/Jsr330ComponentProcessor.java
filiphr: It differs because my goal was for the `JSR330ComponentProcessor` to support both javax and jakarta, with a precedence over javax. This means that if you do not have `javax` available and you use `jsr330` MapStruct will create the correct mappers with Jakarta. This would mean that you don't need to do any changes in your mappers if you just switch dependencies.

However, if you explicitly use the `jakarta` component model then we are only going to use jakarta annotations, won't do any special fallbacks.

sjaakd (approved):

PR #2740, opened 2022-02-01
#2715: Updated documentation to reflect impact of nullValuePropertyMappingStrategy on conditions.

### LINKED ISSUE #2715: Condition on @MappingTarget
When using condition (regardless of the way it has been qualified) with @MappingTarget generated code assign `null` (ie. default value) in else block although [news](https://mapstruct.org/news/2021-07-18-mapstruct-1_5_0_Beta1-is-out/), [documentation  ](https://mapstruct.org/documentation/dev/reference/html/#conditional-mapping) and their examples state that mapping will be skipped if condition isn't met. Moreover it doesn't generate any else statement when operating on primitives making it inconsistent.

Tested on both MapStruct 1.5.0.Beta1 and Beta2.
Here is code I've used to produce this behaviour:

    public abstract class ExemplaryMapper {
    
      static class SourceClass {
    
        private boolean primitiveBoolean;
        private Boolean objectBoolean;
    
        public boolean isPrimitiveBoolean() {
          return this.primitiveBoolean;
        }
    
        public Boolean getObjectBoolean() {
          return this.objectBoolean;
        }
    
        public void setPrimitiveBoolean(boolean primitiveBoolean) {
          this.primitiveBoolean = primitiveBoolean;
        }
    
        public void setObjectBoolean(Boolean objectBoolean) {
          this.objectBoolean = objectBoolean;
        }
      }
    
      static class TargetClass {
        private boolean primitiveBoolean;
        private Boolean objectBoolean;
    
        public boolean isPrimitiveBoolean() {
          return this.primitiveBoolean;
        }
    
        public Boolean getObjectBoolean() {
          return this.objectBoolean;
        }
    
        public void setPrimitiveBoolean(boolean primitiveBoolean) {
          this.primitiveBoolean = primitiveBoolean;
        }
    
        public void setObjectBoolean(Boolean objectBoolean) {
          this.objectBoolean = objectBoolean;
        }
      }
    
      @Condition
      @ConditionAnnotation
      public boolean condition(boolean opt) {
        return opt;
      }
    
      @Condition
      @ConditionAnnotationObj
      public boolean conditionObj(Boolean opt) {
        return opt;
      }
    
    
      @Qualifier
      @Retention(RetentionPolicy.RUNTIME)
      @Target(ElementType.METHOD)
      public static @interface ConditionAnnotation {
      }
    
      @Qualifier
      @Retention(RetentionPolicy.RUNTIME)
      @Target(ElementType.METHOD)
      public static @interface ConditionAnnotationObj {
      }
    
      @Mapping(target = "primitiveBoolean", source = "primitiveBoolean",
          conditionQualifiedBy = ConditionAnnotation.class)
      @Mapping(target = "objectBoolean", source = "objectBoolean",
          conditionQualifiedBy = ConditionAnnotationObj.class)
      public abstract TargetClass map(@MappingTarget TargetClass targetObj, SourceClass sourceObj);
    
    }

And here is the result of generation:

    @Override
    public TargetClass map(TargetClass targetObj, SourceClass sourceObj) {
        if ( sourceObj == null ) {
            return null;
        }

        if ( condition( sourceObj.isPrimitiveBoolean() ) ) {
            targetObj.setPrimitiveBoolean( sourceObj.isPrimitiveBoolean() );
        }
        if ( conditionObj( sourceObj.getObjectBoolean() ) ) {
            targetObj.setObjectBoolean( sourceObj.getObjectBoolean() );
        }
        else {
            targetObj.setObjectBoolean( null );
        }

        return targetObj;
    }

Zegveld:
closes #2715

### DIFF
--- a/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
@@ -344,6 +344,8 @@ public class CarMapperImpl implements CarMapper {
 
         if ( isNotEmpty( car.getOwner() ) ) {
             carDto.setOwner( car.getOwner() );
+        } else {
+            carDto.setOwner( null );
         }
 
         // Mapping of other properties
@@ -362,6 +364,11 @@ If there is a custom `@Condition` method applicable for the property it will hav
 [NOTE]
 ====
 Methods annotated with `@Condition` in addition to the value of the source property can also have the source parameter as an input.
+====
+
+[NOTE]
+====
+`nullValuePropertyMappingStrategy` will influence what to do when the condition is not met.
 ====
 
 <<selection-based-on-qualifiers>> is also valid for `@Condition` methods.
--- a/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
@@ -344,6 +344,54 @@ public class CarMapperImpl implements CarMapper {
 
         if ( isNotEmpty( car.getOwner() ) ) {
             carDto.setOwner( car.getOwner() );
+        }
+
+        // Mapping of other properties
+
+        return carDto;
+    }
+}
+----
+====
+
+When using this in combination with an update mapping method it will replace the `null-check` there, for example:
+
+.Update mapper using custom condition check method
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+@Mapper
+public interface CarMapper {
+
+    CarDto carToCarDto(Car car, @MappingTarget CarDto carDto);
+
+    @Condition
+    default boolean isNotEmpty(String value) {
+        return value != null && !value.isEmpty();
+    }
+}
+----
+====
+
+The generated update mapper will look like:
+
+.Custom condition check in generated implementation
+====
+[source, java, linenums]
+[subs="verbatim,attributes"]
+----
+// GENERATED CODE
+public class CarMapperImpl implements CarMapper {
+
+    @Override
+    public CarDto carToCarDto(Car car, CarDto carDto) {
+        if ( car == null ) {
+            return carDto;
+        }
+
+        if ( isNotEmpty( car.getOwner() ) ) {
+            carDto.setOwner( car.getOwner() );
         } else {
             carDto.setOwner( null );
         }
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
```diff
@@ -344,6 +344,8 @@ public class CarMapperImpl implements CarMapper {
 
         if ( isNotEmpty( car.getOwner() ) ) {
             carDto.setOwner( car.getOwner() );
+        } else {
+            carDto.setOwner( null );
```
filiphr, line 348: Are you sure that we create this branch? I don't think that we do this for normal mappings?

documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
```diff
@@ -364,6 +366,11 @@ If there is a custom `@Condition` method applicable for the property it will hav
 Methods annotated with `@Condition` in addition to the value of the source property can also have the source parameter as an input.
 ====
 
+[NOTE]
+====
+`nullValuePropertyMappingStrategy` will influence what to do when the condition is not met.
```
filiphr, line 371: This is only valid for update mappings and not for other normal mappings. We have an entire section about [Controlling mapping result for null properties](https://mapstruct.org/documentation/stable/reference/html/#mapping-result-for-null-properties) that talks about these things. So perhaps we need to link to that.

filiphr (commented):
I am not sure that this will clarify the documentation. Have a look at my comments

documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
Zegveld: Indeed you are right. so that bit shouldn't be there, but perhaps there should be an example with an update method instead as well.

documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
Zegveld: Correct, I missed that bit. Will rewrite this as well as the above.

Zegveld:
> I am not sure that this will clarify the documentation. Have a look at my comments

I agree, I totally missed the difference between an update mapper and a normal mapper here. Rewritten it to add an example with an update mapper instead, so that it becomes clear that it replaces the null-check instead of being something separate.

filiphr (approved):
LGTM. Thanks @Zegveld

PR #2749, opened 2022-02-08
Mapping map keys with invalid chars for methods

### LINKED ISSUE #2748: Mapping embedded map keys that contain characters that aren't valid for Java method names
I encountered an error when trying to generate a mapper class that maps an object with an embedded map property into a target with a simple property.  The source map contained the key "specific/value", and this key would get converted into a method name with that was invalid for Java, and would fail to compile.

As an example, there source looked somewhat like this:
```
...
public class Source {
...
private Map<String, String> annotations;
...
public Map<String, String getAnnotations() {...}
public void setAnnotations(Map<String,String> annotations) {...}
}
```

And the target was something like this:
```
...
public class Target {
...
private String specificValue;
...
public String getSpecificValue() {...}
...
public void setSpecificValue(String specificValue) {...}
...
}
```

And the @Mapping annotation I was trying to use:
```
@Mapping(source = "annotations.specific/value", target = "specificValue")
Target sourceToTarget(Source trainingPortal);
```

cdelashmutt-pivotal:
This is an attempt to try and allow mapping of maps that have keys that contain invalid characters for Java method names.  Fixes #2748

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/Strings.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/Strings.java
@@ -183,7 +183,15 @@ public class Strings {
 
             if ( firstAlphabeticIndex < identifier.length()) {
                 // If it is not consisted of only underscores
-                return identifier.substring( firstAlphabeticIndex ).replace( "[]", "Array" );
+                String firstAlphaString = identifier.substring( firstAlphabeticIndex ).replace( "[]", "Array" );
+
+                return firstAlphaString.codePoints()
+                    .map(codePoint -> Character.isJavaIdentifierPart(codePoint) || codePoint == (int) '.' ? codePoint : (int) '_')
+                    .collect(
+                        StringBuilder::new,
+                        StringBuilder::appendCodePoint,
+                        StringBuilder::append
+                    ).toString();
             }
 
             return identifier.replace( "[]", "Array" );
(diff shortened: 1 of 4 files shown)


processor/src/main/java/org/mapstruct/ap/internal/util/Strings.java
```diff
@@ -183,7 +183,15 @@ public static String sanitizeIdentifierName(String identifier) {
...
-                return identifier.substring( firstAlphabeticIndex ).replace( "[]", "Array" );
+                String firstAlphaString = identifier.substring( firstAlphabeticIndex ).replace( "[]", "Array" );
+
+                return firstAlphaString.codePoints()
+                    .map(codePoint -> Character.isJavaIdentifierPart(codePoint) || codePoint == (int) '.' ? codePoint : (int) '_')
+                    .collect(
+                        StringBuilder::new,
+                        StringBuilder::appendCodePoint,
+                        StringBuilder::append
+                    ).toString();
```
filiphr, line 194: `codePoints()` is available from Java 9, we can't use this in this path.

In addition to that `codePoint` can never be `.`, since `.` are treated in a special way in MapStruct for nesting mappings.

Let's use good old iteration here.

filiphr (changes requested):
Thanks for the PR @cdelashmutt-pivotal. Unfortunately, this will not work on Java 8. The CI should fail now.

I have left a comment explaining why it won't work.

filiphr:
@cdelashmutt-pivotal I did some small changes to adapt your contribution to work on Java 8 as well. Are you OK with the changes?

@Zegveld can you please do a review as well?

filiphr (approved):

Zegveld (approved):
Looks good to me.

PR #2759, opened 2022-02-18
#2758: fix crash with MappingTarget in Condition

### LINKED ISSUE #2758: Using MappingTarget inside a Condition causes a FreeMarker template error.
When using a condition that also uses the `@MappingTarget` we get an exception.

Reproduction scenario:
```
package updateOnlyNull;

import org.mapstruct.Condition;
import org.mapstruct.Mapper;
import org.mapstruct.Mapping;
import org.mapstruct.Named;

@Mapper
public interface UpdateMapper {

    @Mapping( target = "name", conditionQualifiedByName = "nameCheck" )
    void map(Person person, @MappingTarget Person target);

    @Condition // enabling this will cause faulty codegen.
    @Named( "nameCheck" )
    default boolean check(Person person, @MappingTarget Person target) {
        return person.getName() != null && target.getName() == null;
    }
}
```
Will cause the following generated code with 1.5.0.Beta2

```
package updateOnlyNull;

@Generated(
    date = "2022-02-18T22:22:43+0100",
    comments = "version: 1.5.0.Beta2, compiler: Eclipse JDT (IDE) 1.4.50.v20210914-1429, environment: Java 17.0.1 (Azul Systems, Inc.)"
)
public class UpdateMapperImpl implements UpdateMapper {

    @Override
    public void map(Person person, Person target) {
        if ( person == null ) {
            return;
        }

        if ( check( person, FreeMarker template error: The following has evaluated to null or missing: ==> ext.targetBeanName [in template "org/mapstruct/ap/internal/model/MethodReference.ftl" at line 42, column 15] ---- Tip: It's the step after the last dot that caused this error, not those before it. ---- Tip: If the failing expression is known to be legally refer to something that's null or missing, either specify a default value like myOptionalVar!myDefault, or use <#if myOptionalVar??>when-present<#else>when-missing</#if>. (These only cover the last step of the expression; to cover the whole expression, use parenthesis: (myOptionalVar.foo)!myDefault, (myOptionalVar.foo)?? ---- ---- FTL stack trace ("~" means nesting-related): - Failed at: ${ext.targetBeanName} [in template "org/mapstruct/ap/internal/model/MethodReference.ftl" in macro "arguments" at line 42, column 13] - Reached through: #list parameterBindings as param [in template "org/mapstruct/ap/internal/model/MethodReference.ftl" in macro "arguments" at line 37, column 5] - Reached through: @arguments [in template "org/mapstruct/ap/internal/model/MethodReference.ftl" in macro "methodCall" at line 30, column 53] - Reached through: @methodCall [in template "org/mapstruct/ap/internal/model/MethodReference.ftl" at line 22, column 9] ---- Java stack trace (for programmers): ---- org.mapstruct.ap.shaded.freemarker.core.InvalidReferenceException: [... Exception message was already printed; see it above ...] at ......```

### PR DESCRIPTION
**Pull request by Zegveld**
fallback to param.variableName if ext.targetBeanName is not present in MethodReference handling.

fixes #2758

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/MethodReference.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/MethodReference.ftl
@@ -41,7 +41,7 @@
             <#-- a class is passed on for casting, see @TargetType -->
             <@includeModel object=inferTypeWhenEnum( ext.targetType ) raw=true/>.class<#t>
         <#elseif param.mappingTarget>
-            ${ext.targetBeanName}<#if ext.targetReadAccessorName??>.${ext.targetReadAccessorName}</#if><#t>
+            <#if ext.targetBeanName??>${ext.targetBeanName}<#else>${param.variableName}</#if><#if ext.targetReadAccessorName??>.${ext.targetReadAccessorName}</#if><#t>
         <#elseif param.mappingContext>
             ${param.variableName}<#t>
         <#elseif param.sourceRHS??>
--- a/processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMappingTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMappingTest.java
@@ -7,8 +7,8 @@ package org.mapstruct.ap.test.conditional.basic;
 
 import java.util.Arrays;
 import java.util.Collections;
-
 import org.junit.jupiter.api.extension.RegisterExtension;
+
 import org.mapstruct.ap.testutil.IssueKey;
 import org.mapstruct.ap.testutil.ProcessorTest;
 import org.mapstruct.ap.testutil.WithClasses;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: processor/src/main/resources/org/mapstruct/ap/internal/model/MethodReference.ftl [hunk 1]
```diff
@@ -41,7 +41,7 @@
             <#-- a class is passed on for casting, see @TargetType -->
             <@includeModel object=inferTypeWhenEnum( ext.targetType ) raw=true/>.class<#t>
         <#elseif param.mappingTarget>
-            ${ext.targetBeanName}<#if ext.targetReadAccessorName??>.${ext.targetReadAccessorName}</#if><#t>
+            <#if ext.targetBeanName??>${ext.targetBeanName}<#else>${param.variableName}</#if><#if ext.targetReadAccessorName??>.${ext.targetReadAccessorName}</#if><#t>
```
Zegveld, line 44: Might be better to try and fix this at the location where the `targetBenName` value is not set.

filiphr (approved):
LGTM

Issue #1574, opened 2018-08-05
Add annotations to Generated code


vadim-shb:
I just want to share an idea... I think it would be great to be able to add any annotations you want to generated classes/methods.
Something like this:
```
@Annotated({
  @Component,
  @Qualifier("fooMapper")
})
@Mapper
public interface FooMapper {
  @Annotated({
    @Cacheable(value = "cache", key = "mappedFoo #foo.id" )
  })
  Foo map(Foo foo);
}
```

So, the content of `@Annotated` would be added to generated code like this:
```
@Component,
@Qualifier("fooMapper")
public class FooMapperImpl implements FooMapper {
  @Cacheable(value = "cache", key = "mappedFoo #foo.id" )
  Foo map(Foo foo) {
  ...
  }
}
```
With this functionality we wouldn't need @Mapper(componentModel = "spring") for example, and many annotation-based issues would gone.

sjaakd:
Interesting. I'm not an expert on Spring but to my understanding some of the annotations end up on a referenced mapper instance. Like '@Inject' for CDI. How would that work with your idea?

vadim-shb:
I'm not sure if I've understood your question correctly, so please rephrase it then, or add an example...

In context of Spring `@Inject` would work just fine with the example above (if package with FooMapper.java is scanned for beans by Spring):

```
@Component
public class SomeBean {
  @Inject
  @Qualifier("fooMapper")
  FooMapper mapper

  ...
}
```

sjaakd:
When you use a mapper, ('@Mapper#use') MapStruct will generate that code when the component model is Spring. Don't know if it also evaluates the annotations on the used mapper

vadim-shb:
Ok, I've never used `@Mapper#use`, and after quick pick I still not see where I would like to use it, but it was probably a mistake to talk that 

> we wouldn't need @Mapper(componentModel = "spring")

But it still might be useful in different situations... Leave it here to discuss...

filiphr:
@vadim-shb the request is interesting. 

Regarding:

>With this functionality we wouldn't need @Mapper(componentModel = "spring") for example, and many annotation-based issues would gone.

This is wrong. If you use the default `componentModel` than when you use other mappers with `Mapper#uses` MapStruct will use the `Mappers` factory to create the used mapper or invoke it's default constructor. This is wrong.

[...]

PR #2790, opened 2022-03-17
First test for Mapper with additional annotations.

### LINKED ISSUE #990: Add option to generate Apache Camel's @Converter annotation into mappers and methods
I'm working on a project where we're using both Apache Camel and Mapstruct with CDI.

Camel offers a @Converter annotation as described [here](https://camel.apache.org/type-converter.html). In order to use a generated Mapstruct-Mapper as @Converter, I have to write a "gluing class" looking somewhat like this:

    package sample.conversion;
    import org.apache.camel.Converter;
    import javax.inject.Inject;
    @Converter
    public class TypeConverter {
        @Inject
        private Mapper mapper;
        @Converter
        public TargetType toTarget(SourceType input) {
            return mapper.toTarget(input);
        }
    }

If instead the generated mapper class and and the `toTarget` method already had the @Converter annotation, I could drop this class altogether.

Chessray:
Fixes #990

### DIFF
--- a/core/src/main/java/org/mapstruct/Mapper.java
+++ b/core/src/main/java/org/mapstruct/Mapper.java
@@ -376,5 +376,12 @@ public @interface Mapper {
      */
     boolean suppressTimestampInGenerated() default false;
 
+    /**
+     * Additional annotations to be added to the generated Mapper implementation
+     *
+     * @return A list of annotation types to be added to the generated Mapper implementation
+     *
+     * @since 1.5
+     */
     Class<? extends Annotation>[] implementationAnnotations() default {};
 }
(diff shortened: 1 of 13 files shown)


filiphr:
I would refer this to @Zegveld since I know that he is also working on something like this as part of #1574. The other issue is slightly more complicated that it allows to put values in the annotations as well.

Chessray:
It's kind of awkward that these issues have been open for years, and then within the course of less than 3 weeks, @Zegveld and myself both independently find time and willingness to do something about it. 🤦🏼‍♂️  I agree that my attempt is the less sophisticated one. Let me know if we need anything to get the other one across the finish line.

filiphr:
It is really interesting to see how 2 people independently in the same time do this 😄. Thanks for the understanding @Chessray

PR #2798, opened 2022-03-23
#2797: When adding imports for types also check if the types own imports are needed.

### LINKED ISSUE #2797: Compilation error due to missing import with nested classes.
Example project: https://github.com/laech/mapstruct-nested-class-issue

Mapper generated with v1.5.0.Beta1 works fine containing:

```java
import com.example.model.BasePerson.Names;
import com.example.model.Example.Person;

...
public class ExampleMapperImpl implements ExampleMapper {
    ...
    private String personNamesLast(Person person) {
        ...
        Names names = person.names;
        ...
```

Mapper generated with v1.5.0.Beta2 or v1.5.0.RC1 fails to compile:

```java
import com.example.model.Example;
...
public class ExampleMapperImpl implements ExampleMapper {
    ...
    private String personNamesLast(Example.Person person) {
        ...
        BasePerson.Names names = person.names;  // <-- compiler error
        ...
```

Due to `BasePerson` is missing in the imports.

Zegveld:
Type's own imports consist of itself and top levels. Sometimes top level imports are used instead of the Type itself. In this case the top level type needs to be added to the import listing.

fixes #2797

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
@@ -10,14 +10,15 @@ import java.util.Collection;
 import java.util.List;
 import java.util.SortedSet;
 import java.util.TreeSet;
+
 import javax.lang.model.type.TypeKind;
+import org.mapstruct.ap.internal.util.ElementUtils;
 
 import org.mapstruct.ap.internal.model.common.Accessibility;
 import org.mapstruct.ap.internal.model.common.ModelElement;
 import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.common.TypeFactory;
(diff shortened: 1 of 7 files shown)


filiphr (commented):
I don't like the change because it would lead to repeating a lot of things. In almost all of our cases when we get import type we do `Type#getImportTypes`. I pushed a commit with the place that was causing this problem. Please let me know if you agree with it and if you do feel free to merge it

Zegveld:
Looks good to me, although I am wondering if there are also other locations where there are nested imports missing. Is it only the `NestedPropertyMappingMethod` that uses it or are there also other locations?

I guess we'll get other issue reports if that is the case, so merging this one.

filiphr:
There could indeed be some other places where this is not working. However, I'd like to fix those places in their respective places when we get reports that it is't working

PR #2805, opened 2022-04-02
#2794 Compile error when condition expression used with constant or expression

### LINKED ISSUE #2794: No compile error when `conditionExpression` and `expression` are used together
Using MapStruct 1.5.0.RC1, imagine the following (or a matching class hierarchy, doesn't matter)

```java
// domain
interface Result {}
record Success(Instant timestamp) implements Result {}
record Failure(String reason) implements Result {}

// public API
record ResultApi(boolean success, Instant timestamp, String reason) {}
```

and the following mapper:
```java
@Mapping(target = "success", expression = "java(result instanceof Success)")
@Mapping(target = "timestamp",
    expression = "java(result instanceof Success ? ((Success) result).getTimestamp() : null)")
@Mapping(target = "reason",
    expression = "java(result instanceof Failure ? ((Failure) result).getReason() : null)")
ResultApi resultToResultApi(Result result);
```

This works. I tried changing the mapper to
```java
@Mapping(target = "success", expression = "java(result instanceof Success)")
@Mapping(target = "timestamp",
    conditionExpression = "java(result instanceof Success)",
    expression = "java(((Success) result).getTimestamp())")
@Mapping(target = "reason",
    conditionExpression = "java(result instanceof Failure)",
    expression = "((Failure) result).getReason())")
ResultApi resultToResultApi(Result result);
```

however the conditionals never appeared in the generated code.

filiphr:
Fixes #2794


Zegveld (approved):
Looks good to me.

PR #2808, opened 2022-04-04
#2807: Include LifeCycleMethod importTypes in the list of importTypes.

### LINKED ISSUE #2807: No import is added to generated class when using '@BeforeMapping' on a 'used' mapper
If we use a "baseMapper" class that the only method called is a "@BeforeMapping", than a implementation class is generated without the proper import. This issue was tested at versions: "1.4.2.final" and "1.5.0.RC1"

If we use any of regular(no "BeforeMapping") methods from baseMapper the important is added to the generated class.

```
public class BaseMapper {
    private BaseMapper() {   }

    public static String mapUuidToString(UUID uuid) {
        return uuid == null ? null : uuid.toString();
    }

    @BeforeMapping
    public static <T, D> List<D> handleNullLists(Iterable<T> source) {
        return source == null ? new ArrayList() : null;
    }
}
```

The code below generates a no working implementation class, since a import to `BaseMapper` is missing:
```
@Mapper(componentModel = "spring", uses = {BaseMapper.class})
public interface NoWorkingSample {

    List<String> map(List<Integer> list);
}
```

The code below generates a working implementation class, since we are also using the `mapUuidToString` method from the `BaseMapper`:
```
@Mapper(componentModel = "spring", uses = {BaseMapper.class})
public interface WorkingSample {

    List<String> map(List<UUID> list);
}
```

- [X] Is this an issue (and hence not a question)?

Zegveld:
fixes #2807

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/MappingMethod.java
@@ -5,11 +5,9 @@
  */
 package org.mapstruct.ap.internal.model;
 
-import static org.mapstruct.ap.internal.util.Strings.getSafeVariableName;
-import static org.mapstruct.ap.internal.util.Strings.join;
-
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


filiphr (approved):
LGTM

PR #2809, opened 2022-04-04
Fixes NullPointerException on void mapping methods

intellitour:
When `ee.getReturnType()` returns `void`, `typeUtils.asElement( ee.getReturnType() )` returns `null`. So, `typeUtils.asElement( ee.getReturnType() ).getSimpleName()` crashes with `NullPointerException`. 

This PR fixes this situation.


filiphr:
Thanks for the PR @intellitour, can you please also add a test case that causes this to fail?

sjaakd (changes requested):
can you add a simple reproducer? I'm struggling to see what issue this fixes

filiphr:
>but here is one thing I forgot to mention: the issue occurs with Kotlin, not Java.

If you can't find Java code that will cause this error, then please do share a Kotlin example. Perhaps it is some kind of a Kotlin bug, and it would be good to report it and analyze it.

filiphr:
Closing this in favour of PR #2868.

Thanks for the PR in any case @intellitour. The fix is the same as yours.

We didn't merge it earlier, because it was lacking tests and we couldn't understand why it was happening.

Issue #2688, opened 2021-12-13
Access to the target property name 


HemalathaGowdar:
Expose target property to method arguments

filiphr:
Can you please provide an example?

HemalathaGowdar:
@Mapper(config = MapperBConfig.class , uses = EValueMapper.class)
@Component 
public abstract  class AttribMapper {
@Mapping(Source = "indicator" , target= "ïndicator")
public abstract TargetAttributes TargetAttributesMapper( Source Attributes @Mapping TargetAttributes );

@condition
public boolean conditioncheck(String value)
{
if(contains in map attribute check) // basically am trying to map to the target attribute if that attribute is required by consumer 
return true
else return false
}

This method should get invoke for all attributes mapping.
I also tried Conditionexpression to check whether this attribute in teh consumer list to map or not . This also not coming in the generated impl class

Could you please suggest here.

Zegveld:
If I understand correctly what you want is something like I've got a `Map<String, Object>` named `source` and another `Map<String, Object>` named `target` and I only want to map the entries from `source` for which there is already an entry present in `target`?
Trying to do this using conditions with Iterables or Maps is not supported yet, see #1610 for more information.

However writing the mapping yourself might not be that much additional code, and mapstruct can still use it in the mappingchain like this:
```
@Mapper
public abstract class ConditionalMapMapper {
    static class Item {
        Map<String, String> value;
       // getters and setters
    }

    static class ItemDto {
        Map<String, String> value;
       // getters and setters
    }

    abstract Item map(ItemDto dto, @MappingTarget Item item);

    void map(Map<String, String> source, @MappingTarget Map<String, String> target) {
        for ( Entry<String, String> entry : source.entrySet() ) {
            if ( target.containsKey( entry.getKey() ) ) {
                target.put( entry.getKey(), entry.getValue() );
            }
        }
    }
}
```

The map method with the source and target map will be used in the implementation to map from/to a map.

```
public class ConditionalMapMapperImpl extends ConditionalMapMapper {

    @Override
    Item map(ItemDto dto, Item item) {
        if ( dto == null ) {
            return null;
        }

        if ( dto.getValue() != null ) {
            if ( item.getValue() == null ) {
                item.setValue( new LinkedHashMap<String, String>() );
            }
            map( dto.getValue(), item.getValue() );
        }
        else {
            item.setValue( null );
        }

        return item;
    }
}
```

Hope this helps.

[...]

PR #2834, opened 2022-04-28
#2688: Access to the target property name

ivlcic:
### Implemented

Access to the target property name https://github.com/mapstruct/mapstruct/issues/2688
As agreed in discussion: https://github.com/mapstruct/mapstruct/discussions/2831

### Rationale
Currently there is no option to do run-time filtering based on target property name.

This comes very handy when mapping larger object graphs from **ORM** frameworks, to block
unnecessary loading from database and reuse mapping code.

With this feature you can for instance block further mapping in runtime based on some `@Context` variable state and reuse mapper code without having to foresee every possible property mapping combination.

### Implementation
Implementation introduces support for new **presence check method parameter** annotated with `@TargetPropertyName`.

It is implemented with minimum impact to current code base, so it might be in architecture/elegance sense problematic.

It introduces no changes any current features; it's only an addition.

Unit tests are provided in `org/mapstruct/ap/test/conditional`.

### Example 1
```java
@Mapper
public interface MyMapper {

    MyMapper INSTANCE = Mappers.getMapper( MyMapper.class );

    Employee map(EmployeeDto employee);

    @Condition
    default <T> boolean isNotEmpty(Collection<T> collection, @TargetPropertyName String propName) {
        if ( "addresses".equalsIgnoreCase( propName ) ) {
            return false;
        }
        return collection != null && !collection.isEmpty();
    }

    @Condition
    default boolean isNotBlank(String value, @TargetPropertyName String propName) {
        if ( propName.equalsIgnoreCase( "lastName" ) ) {
            return false;
        }
        return value != null && !value.trim().isEmpty();
    }
}
```

### Example 2
```java
@Mapper
public interface MyMapper {

    MyMapper INSTANCE = Mappers.getMapper( MyMapper.class );

    Employee map(EmployeeDto employee, @Context PresenceUtils utils);

    Address map(AddressDto addressDto, @Context PresenceUtils utils);

    class PresenceUtils {
        Set<String> visited = new LinkedHashSet<>();

        @Condition
        public boolean decide(String value, @TargetPropertyName String propName) {
            return true; //decide if continue mapping
        }
    }
}
```

### Complex Example
```java
interface Entity {}
interface Dto {}

class Employee implements Entity { // setters and getters ommited
    String name;
    Employee boss;
    Address address;
    List<Employee> subordinates;
    String title;
}

class Address implements Entity { // setters and getters ommited
    String street;
}

class EmployeeDto implements Dto { // default constructor, setters and getters ommited
    String name;
    EmployeeDto boss;
    AddressDto address;
    List<EmployeeDto> subordinates;
    String title;

    public EmployeeDto(String name, EmployeeDto boss, AddressDto address) {
        this.name = name;
        this.boss = boss;
        this.address = address;
    }

    public EmployeeDto(String name, AddressDto address, List<EmployeeDto> subordinates) {
        this.name = name;
        this.address = address;
        this.subordinates = subordinates;
    }
}

class AddressDto implements Dto { // default constructor, setters and getters ommited
    String street;
    
    public AddressDto(String street) {
        this.street = street;
    }
}

class PathSegment {
    String name;
    boolean cycleDetected;

    public PathSegment(String name, boolean cycleDetected) {
        this.name = name;
        this.cycleDetected = cycleDetected;
    }
}

class CollectionPathSegment extends PathSegment {
    int index;

    public CollectionPathSegment(String name, boolean cycleDetected) {
        super( name, cycleDetected );
    }
}

class CollectionElementPathSegment extends PathSegment {
    int index;
    public CollectionElementPathSegment(String name, int index, boolean cycleDetected) {
        super( name, cycleDetected );
        this.index = index;
    }
}

class MapContext {
    Set<String> ignoreProps = new HashSet<>();
    Set<Integer> mapped = new HashSet<>();
    Deque<PathSegment> path = new LinkedList<>();
    Deque<String> visited = new LinkedList<>();
    String lastProp = null;

    private String pathToString() {
        StringBuilder strPath = new StringBuilder();
        for ( PathSegment segment : path ) {
            if ( segment instanceof CollectionElementPathSegment ) {
                strPath.append( segment.name );
            } else {
                if ( strPath.length() > 0 ) {
                    strPath.append( "." );
                }
                strPath.append( segment.name );
            }
        }
        if (path.isEmpty()) {
            return lastProp;
        }
        strPath.append( "." );
        strPath.append( lastProp );
        return strPath.toString();
    }

    @Condition
    public boolean collect(@TargetPropertyName String propName) {
        PathSegment segment = path.peekLast();
        if ( ignoreProps.contains( propName ) || (segment != null && segment.cycleDetected)) {
            return false;
        }
        this.lastProp = propName;
        this.visited.offerLast( pathToString() );
        return true;
    }
}

interface FirmMapper<E extends Entity, D extends Dto> {

    E map(D dto, @Context MapContext ctx);

    @BeforeMapping
    default void before(Object dto, @Context MapContext ctx) {
        if ( dto == null ) {
            return;
        }
        boolean cycleDetected = !ctx.mapped.add( dto.hashCode() );

        if ( ctx.lastProp == null ) {
            return;
        }
        if ( dto instanceof Collection ) {
            ctx.path.offerLast( new CollectionPathSegment( ctx.lastProp, cycleDetected ) );
        } else {
            PathSegment segment = ctx.path.peekLast();
            if ( segment instanceof CollectionPathSegment ) {
                ctx.path.offerLast(
                        new CollectionElementPathSegment(
                                "[" + ( (CollectionPathSegment) segment ).index + "]",
                                ( (CollectionPathSegment) segment ).index,
                                cycleDetected
                        )
                );
                ( (CollectionPathSegment) segment ).index++;
            } else {
                ctx.path.offerLast( new PathSegment( ctx.lastProp, cycleDetected ) );
            }
        }
    }

    @AfterMapping
    default void after(Object dto, @Context MapContext ctx) {
        ctx.path.pollLast();
        PathSegment segment = ctx.path.peekLast();
        if ( segment != null ) {
            ctx.lastProp = segment.name;
        } else {
            ctx.lastProp = null;
        }
    }
}

@Mapper
interface EmployeeMapper extends FirmMapper<Employee, EmployeeDto> {
    EmployeeMapper INSTANCE = Mappers.getMapper( EmployeeMapper.class );
}

@Mapper
interface AddressMapper extends FirmMapper<Address, AddressDto> {
    AddressMapper INSTANCE = Mappers.getMapper( AddressMapper.class );
}

// ... more mappers

public void test) {
    EmployeeMapper mapper = EmployeeMapper.INSTANCE;
    MapContext context = new MapContext();
    //context.ignoreProps.add( "boss" ); // glob - like exclude / include could be implemented

    EmployeeDto bossDto = new EmployeeDto( "Boss", new AddressDto("Testing st. 18"), new ArrayList<>() );
    bossDto.subordinates.add( new EmployeeDto( "Employee1", bossDto, new AddressDto("Street 1") ) );
    bossDto.subordinates.add( new EmployeeDto( "Employee2", bossDto, new AddressDto("Street 2") ) );
    bossDto.subordinates.add( new EmployeeDto( "Employee3", bossDto, new AddressDto("Street 3") ) );

    Employee boss = mapper.map( bossDto, context );
    assertThat( List.of(
            "name",
            "boss",
            "address",
            "address.street",
            "subordinates",
            "subordinates[0].name",
            "subordinates[0].boss",
            "subordinates[0].address",
            "subordinates[0].address.street",
            "subordinates[0].subordinates",
            "subordinates[0].title",
            "subordinates[1].name",
            "subordinates[1].boss",
            "subordinates[1].address",
            "subordinates[1].address.street",
            "subordinates[1].subordinates",
            "subordinates[1].title",
            "subordinates[2].name",
            "subordinates[2].boss",
            "subordinates[2].address",
            "subordinates[2].address.street",
            "subordinates[2].subordinates",
            "subordinates[2].title",
            "title"
    ) ).isEqualTo( context.visited );
}

```


core/src/main/java/org/mapstruct/TargetProperty.java
```diff
@@ -0,0 +1,24 @@
...
+ * This annotation marks a <em>presence check method</em> parameter as a property name parameter.
+ * <p>
+ *   This parameter enables conditional filtering based on source property name in run-time.
+ * </p>
+ * @author Nikola Ivačič.
+ * @since 1.5
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.CLASS)
+public @interface TargetProperty {
```
sjaakd, line 23: I would perhaps opt for a slightly longer name: `TargetPropertyName`. This leaves an option open to have a `TargetPropertyValue` (e.g. a Supplier function)

core/src/main/java/org/mapstruct/TargetProperty.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation marks a <em>presence check method</em> parameter as a property name parameter.
+ * <p>
+ *   This parameter enables conditional filtering based on source property name in run-time.
```
sjaakd, line 16: target property iso source property

core/src/main/java/org/mapstruct/TargetProperty.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation marks a <em>presence check method</em> parameter as a property name parameter.
```
sjaakd, line 14: Make a link to `@Condition` and make the use of this annotation exclusive to methods annotated with `@Condition`

processor/src/main/java/org/mapstruct/ap/internal/model/PresenceCheckMethodResolver.java
```diff
@@ -5,11 +5,6 @@
  */
 package org.mapstruct.ap.internal.model;
 
-import java.util.ArrayList;
```
sjaakd, line 8: Try to avoid reformatting..

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
```diff
@@ -5,18 +5,19 @@
  */
 package org.mapstruct.ap.internal.model.common;
 
-import java.util.List;
```
sjaakd, line 8: Try to avoid formatting.

processor/src/main/java/org/mapstruct/ap/internal/model/PresenceCheckMethodResolver.java
```diff
@@ -47,7 +47,6 @@ public static PresenceCheck getPresenceCheck(
         }
 
         MethodReference methodReference = getPresenceCheckMethodReference( method, matchingMethod, ctx );
-
```
sjaakd, line 50: Try to avoid formatting

core/src/main/java/org/mapstruct/TargetProperty.java
sjaakd: Also explicitly limit the type to `String` in this text. Perhaps have a testcase in which you use another java type and see if you get a decent error. Add error handling?

sjaakd (changes requested):
@ivlcic I need to review the implementation a bit more thoroughly. But we have a `dependsOn` in `@Mapping`.. without digging (yet) to deep into this problem.. Isn't that usable.?

ivlcic:
> @ivlcic I need to review the implementation a bit more thoroughly. But we have a `dependsOn` in `@Mapping`.. without digging (yet) to deep into this problem.. Isn't that usable.?

I don't really see how this is connected. Basically what I needed was to block contained (sub)collection(s) mapping at runtime based on target property name while reusing same mapper. 

I have a REST client which can filter out returned properties from deep nested structures (GraphML like), and this is done via runtime parameters. That's why I can't foresee every possible combination and use different mapper (and hence different compile time `@Mapping`) for specific desired client response.

core/src/main/java/org/mapstruct/TargetProperty.java
ivlcic: That was my initial implementation but it was suggested to me to rename it to `@TargetProperty` :D.

core/src/main/java/org/mapstruct/TargetProperty.java
ivlcic: Done

core/src/main/java/org/mapstruct/TargetProperty.java
ivlcic: 1. Done
2. Not sure if I'm capable to make correct error handling.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Parameter.java
ivlcic: Done

processor/src/main/java/org/mapstruct/ap/internal/model/PresenceCheckMethodResolver.java
ivlcic: Done

filiphr:
I'll need to have a better look into this. However, I would like to thank you for the contribution, the examples look really good (although one might be a bit too complex 😄 ).

>I don't really see how this is connected. Basically what I needed was to block contained (sub)collection(s) mapping at runtime based on target property name while reusing same mapper.

I think that @sjaakd wrote that because of the "Complex Example" that you have. Which is doing something really complex, and I do understand why you would need it in your code, we would need to see if we want to keep it in our tests (we would need to scratch our head always when we see it)

>btw: I squashed commits on my branch again after review. I guessed that this would be easier to compare to master. I hope I didn't spoil your workflow.

It is all fine, we are squashing everything when we merge anyways. The diff is also complete towards the merge branch in a PR, so it is all good.

_____

Since we are already out with 1.5.0.RC1, I would like to target this towards 1.6 (with hopefully faster release this time)

core/src/main/java/org/mapstruct/TargetProperty.java
filiphr: Indeed, me and @Zegveld suggested `@TargetProperty` initially. We have `@TargetType` for the `Class` of the type. In any case, we can always easily change this before the merge if we thing `@TagetProperty` is a better name (I think that @ivlcic changed it back to `@TargetPropertyName`)

core/src/main/java/org/mapstruct/TargetProperty.java
filiphr: We always try to be as specific to users with our error messages. However, I do agree, we can do this enhancement in an additional PR.

ivlcic:
> I think that @sjaakd wrote that because of the "Complex Example" that you have. Which is doing something really complex, and I do understand why you would need it in your code, we would need to see if we want to keep it in our tests (we would need to scratch our head always when we see it)

I'm sorry for that "Complex example" and similar - derived tests. I'm a complete noob to Mapstruct and I had only two days to implement "_runtime object graph path filtering_". So I started `@TargetPropertyName` implementation through tests just to prove to my self, that I can pull it off with a **single, minimal and useful** change to Mapstruct's code base. 

I was trying to convince (you) the contributors why this would be useful and is unachievable with `@Mapping` annotation and combination of other Mapstruct's features. So I left it in PR's tests. Therefore I have no problem with removing the complex tests (and/or any other code that you don't like for that matter). 

You can just leave a comment and I'll fix/remove it.

processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMethodForCollectionMapperWithTargetPropertyName.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.mapstruct.Condition;
+import org.mapstruct.Mapper;
+import org.mapstruct.TargetPropertyName;
+import org.mapstruct.factory.Mappers;
+
+import java.util.Collection;
+
+/**
+ * @author Nikola Ivačič
```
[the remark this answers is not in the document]
ivlcic: :rofl:, hmn I guess you can rename me :rofl: Ivacic is as good as Ivačič. Replace it with your name if its more friendly ...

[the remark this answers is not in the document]
ivlcic: I had this issue once with maven-javadoc-plugin but can't remember which project and why it didn't pick the project.build.sourceEncoding

processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMappingTest.java
```diff
@@ -119,6 +122,218 @@ public void conditionalMethodInUsesContextMapper() {
...
+
+        TargetPropertyNameModel.EmployeeDto employeeDto = new TargetPropertyNameModel.EmployeeDto();
+        employeeDto.setFirstName( "  " );
+        employeeDto.setLastName( "Testirovich" );
+        employeeDto.setCountry( "US" );
+        employeeDto.setAddresses( List.of( new TargetPropertyNameModel.AddressDto( "Testing St. 6" ) ) );
+
+        TargetPropertyNameModel.Employee employee = mapper.map( employeeDto, utils );
+        assertThat( employee.getLastName() ).isEqualTo( "Testirovich" );
+        assertThat( employee.getFirstName() ).isEqualTo( "  " );
```
sjaakd, line 233: shouldn't you assert country?

processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMappingTest.java
```diff
@@ -119,6 +122,218 @@ public void conditionalMethodInUsesContextMapper() {
...
+      ConditionalMethodWithTargetPropertyNameInContextMapper.PresenceUtils utils =
+        new ConditionalMethodWithTargetPropertyNameInContextMapper.PresenceUtils();
+
+      TargetPropertyNameModel.EmployeeDto employeeDto = new TargetPropertyNameModel.EmployeeDto();
+      employeeDto.setLastName( "  " );
+      employeeDto.setCountry( "US" );
+      employeeDto.setAddresses( List.of( new TargetPropertyNameModel.AddressDto( "Testing St. 6" ) ) );
+
+      TargetPropertyNameModel.Employee employee = mapper.map( employeeDto, utils );
+      assertThat( employee.getLastName() ).isNull();
```
sjaakd, line 262: shouldn't you assert lastname and firstname as well?

processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMappingTest.java
```diff
@@ -119,6 +122,218 @@ public void conditionalMethodInUsesContextMapper() {
...
+      ConditionalMethodWithTargetPropertyNameInContextMapper.PresenceUtilsAllProps allPropsUtils =
+        new ConditionalMethodWithTargetPropertyNameInContextMapper.PresenceUtilsAllProps();
+
+      employeeDto = new TargetPropertyNameModel.EmployeeDto();
+      employeeDto.setLastName( "Tester" );
+      employeeDto.setCountry( "US" );
+      employeeDto.setAddresses( List.of( new TargetPropertyNameModel.AddressDto( "Testing St. 6" ) ) );
+
+      employee = mapper.map( employeeDto, allPropsUtils );
+      assertThat( employee.getLastName() ).isEqualTo( "Tester" );
```
sjaakd, line 281: other assertions?

processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMappingTest.java
```diff
@@ -119,6 +122,218 @@ public void conditionalMethodInUsesContextMapper() {
...
+      bossEmployeeDto.setAddresses( List.of( new TargetPropertyNameModel.AddressDto( "Testing St. 6" ) ) );
+
+      employeeDto = new TargetPropertyNameModel.EmployeeDto();
+      employeeDto.setLastName( "Tester" );
+      employeeDto.setCountry( "US" );
+      employeeDto.setBoss( bossEmployeeDto );
+      employeeDto.setAddresses( List.of( new TargetPropertyNameModel.AddressDto( "Testing St. 6" ) ) );
+
+      employee = mapper.map( employeeDto, allPropsUtilsWithSource );
+      assertThat( employee.getLastName() ).isEqualTo( "Tester" );
```
sjaakd, line 312: other assertions / null, notNull?

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
```diff
@@ -263,7 +265,7 @@ private SourceMethod(Builder builder, MappingMethodOptions mappingMethodOptions)
     private boolean determineIfIsObjectFactory() {
         boolean hasNoSourceParameters = getSourceParameters().isEmpty();
         boolean hasNoMappingTargetParam = getMappingTargetParameter() == null;
-        return !isLifecycleCallbackMethod() && !returnType.isVoid()
+        return !isLifecycleCallbackMethod() && !returnType.isVoid() && !returnType.isPrimitive()
```
sjaakd, line 268: I think I can understand why you added `!returnType.isPrimitive()`.. But is it strictly part of this PR? Should there be a separate testcase validating this?

sjaakd (changes requested):
some small comments after more thorough review.. In general it looks good to me.. Nice job.

processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMethodForCollectionMapperWithTargetPropertyName.java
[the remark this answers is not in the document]
sjaakd: Sorry for that. I've seen this before as well.. perhaps that we should check why it gives error notices.. Could be that I'm the only one seeing this..

core/src/main/java/org/mapstruct/TargetProperty.java
sjaakd: Ok.. I can live with that as well.. See #2863

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
ivlcic: This is the one I wasn't happy about. It is too hackish to me since its not explicit. I was wondering if it would be better to invoke `determineIfIsPresenceCheck` instead of `!returnType.isPrimitive()`, but then the method leveling is kinda wrong.
I would need some consulting from your behalf on this...

ivlcic: I came across execution when object factory method was detected instead of presence check. What I assumed was: object factory can't return a primitive ... but that is implicit ... and that's why I'm not happy with that IF

processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMethodForCollectionMapperWithTargetPropertyName.java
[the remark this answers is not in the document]
ivlcic: I'm using Linux with default locale set to UTF-8 so its hard for me to replicate. It's much easier to rename myself ... It's not like I'm gonna be famous :rofl:

processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMappingTest.java
ivlcic: Will fix tomorrow

processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMappingTest.java
ivlcic: Will fix tomorrow

processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMappingTest.java
ivlcic: Will fix tomorrow

core/src/main/java/org/mapstruct/TargetProperty.java
sjaakd: Ok.. given your example: the `Type` in `TargetType` refers to `class` or `interface`. So `Name` in `@TargetPropertyName` refers to a text or - in other words - `String`.  It leaves the option open for extension towards the future. I'm not bothered by the non-verbosity, since a proper IDE completes it anyway..

@filiphr : Enough on my side of the argument. I leave it up for you tot decide 😄 .

sjaakd:
>You can just leave a comment and I'll fix/remove it.

I'm fine with it. I first focused on what I observed (did not have too much time either yesterday) and at first glance I thought it also concerned the order of property-mapping (which is a kind of strange assumption at hindsight admittedly). I can see - in your example - you got your inspiration from the life-cycle avoidance context testcases. I'm also happy to see that this did not make it to the unit testcases.

core/src/main/java/org/mapstruct/TargetProperty.java
Zegveld: Thinking out of the box here, but if someone wants to manage all their properties using an `enum` or some other container. Why wouldn't we allow this as long as we can convert to that type? It does open up some nice management options for people who want to use that route.

The above suggestion should be a separate PR though, but at that point having it already listed as `@TargetProperty` might be more useful. Although I can understand changing it to `@TargetPropertyName` because this is more verbose and still to the point even with the above suggestion.

I'm fine with either annotation.

processor/src/main/java/org/mapstruct/ap/internal/model/source/SourceMethod.java
Zegveld: For this to be the same as in the rest of the flow wouldn't it make more sense to say here:
```java
boolean hasNoTargetPropertyNameParam = getTargetPropertyNameParameter() == null;
return !isLifecycleCallbackMethod() && !returnType.isVoid()
            && hasNoMappingTargetParam && hasNoTargetPropertyNameParam
            && ( hasObjectFactoryAnnotation || hasNoSourceParameters );
```
If I correctly understand that you want to exclude methods that have a `@TargetPropertyName` from being seen as an object factory?

ivlcic: Yep, That's what I was looking for! Thanks!

Zegveld (changes requested):
It feels to me that there is a lot more pass-through code present then necessary.
The set of changes that I request all focus on the same topic, using the same style as is done for `TargetType`.
I haven't placed remarks at each location where adjustments might be required.

The tests look good, which should make it easier to refactor this.

processor/src/main/resources/org/mapstruct/ap/internal/model/MethodReference.ftl
```diff
@@ -44,6 +44,8 @@
             <#if ext.targetBeanName??>${ext.targetBeanName}<#else>${param.variableName}</#if><#if ext.targetReadAccessorName??>.${ext.targetReadAccessorName}</#if><#t>
         <#elseif param.mappingContext>
             ${param.variableName}<#t>
+        <#elseif param.targetPropertyName>
+            "${param.targetPropertyNameValue}"<#t>
```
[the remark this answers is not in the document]
ivlcic: > can't we retrieve this value inside the freemarker tree itself?
I can't find a way properly reference `PropertyMapping` model and its `name` property in `arguments` ftl macro.
#noob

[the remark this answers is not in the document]
Zegveld: The `PropertyMapping` knows the `name`, so in the `PropertyMapping.ftl` you can pass it on at the `<@includeModel object=assignment` call.
You see there for example `targetType=targetType`, the first `targetType` is the name it will be known in the assignment area. There it is refered to as `ext.targetType`. The second `targetType` is the java bean `targetType` from the `PropertyMapping` class, the `getTargetType()` method.
You will need to pass it on like this to get it to where you use it at `MethodReference.ftl`.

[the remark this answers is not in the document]
ivlcic: I'm some what guessing, but is `PropertyMapping` assignment accessible in ftl for presence check methods?
If I use:
```
@Condition
default boolean isNotBlank(String value, @TargetType Class<?> cls) {...
```
I get null for cls and NPE in `org.mapstruct.ap.internal.model.MethodReference#inferTypeWhenEnum`.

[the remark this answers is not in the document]
Zegveld: It seems you have found a bug.

I can reproduce it with 1.5.1.Final. The current value of `@TargetType` here should have been `Class<String>` I think, which is the `TargetType` of the parameter value for which the `isNotBlank` check takes place. So don't think that `@TargetType` is that useful for `@Condition` since you are already using the correct type in the previous parameter.

Let's fix this in a separate issue (#2882) with it's own PR.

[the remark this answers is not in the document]
Zegveld: The PR #2883 is a [nice example of how to pass on information through freemarker](https://github.com/mapstruct/mapstruct/pull/2883/files).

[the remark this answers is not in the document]
ivlcic: What name do you suggest for template var? I think that `name` might be too generic, but on the other hand that is a field name in `PropertyMapping`?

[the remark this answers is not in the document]
ivlcic: Ok, I think I'm done. I have one more observation regarding #2883 (I don't know if I'm right since I didn't write a test):
I believe `<#macro handleLocalVarNullCheck needs_explicit_local_var>` in CommonMacros.ftl should be fixed in the same way as `<#macro handleSourceReferenceNullCheck>` or the `Class<?> targetType` parameter passing might fail if presence check method is in `@Context` parameter.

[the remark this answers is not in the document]
ivlcic: And another thing: I didn't pass `ext.targetPropertyName` in every possible assigment (as is `ext.targetType`) since it's only relevant in presence check methods. I hope that's ok.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -84,10 +85,22 @@ public <T extends Method> List<SelectedMethod<T>> getMatchingMethods(Method mapp
...
         List<ParameterBinding> availableParams = new ArrayList<>( method.getParameters().size() + 3 );
 
+        for (SelectedMethod<T> selectedMethod : methods) {
+          Parameter propertyNameParameter = selectedMethod.getMethod().getTargetPropertyNameParameter();
+          if (propertyNameParameter != null) {
+            availableParams.add( ParameterBinding
+              .fromParameter( propertyNameParameter ) );
+          }
+        }
+
```
Zegveld, line 103: is this still needed?

Can't we solve this in the same manner as is done for `@TargetType` in the `addMappingTargetAndTargetTypeBindings` method?

Zegveld (changes requested):
Starts to look great, just one issue remaining concerning the way things are handled at the ParameterBindings.

I think it is also good to add a compiler error in case the `@TargetPropertyName` annotation is used in a method other than one annotated with `@Condition`. Because that currently is not supported. What do you think, @filiphr?

Keep up the good work, @ivlcic.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
ivlcic: I'll look into it and try to solve it as you suggested. (currently its needed ...)

Zegveld (changes requested):

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -149,6 +151,33 @@ else if ( pb.isTargetType() ) {
...
+                break;
+            }
+        }
+        if ( !targetPropertyNameAvailable ) {
+            for ( SelectedMethod<T> selectedMethod : methods ) {
+                Parameter propertyNameParameter = selectedMethod.getMethod().getTargetPropertyNameParameter();
+                if ( propertyNameParameter != null ) {
+                    availableParams.add( ParameterBinding.fromParameter( propertyNameParameter ) );
+                }
+            }
```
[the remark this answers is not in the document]
ivlcic: Sorry for misunderstanding and thanks for your patience.
I would need some further assistance regarding error checking:
- Do we put it in another ticket?
- Where would you put error checking? My take is (from the top of my head): all `MethodResolver`s except `PresenceCheckMethodResolver`
- Is error message constant something like: `GENERAL_TARGET_PROPERTY_NAME_IN_WRONG_METHOD_TYPE` ok?

sjaakd (approved):
Nice piece of work @ivlcic .. thanks!

filiphr:
>I think it is also good to add a compiler error in case the `@TargetPropertyName` annotation is used in a method other than one annotated with `@Condition`. Because that currently is not supported.

I do not think that we are doing that for `@TargetType`, not sure if we need to do it for this at the beginning to be honest.

filiphr:
In any case if we need to add custom checks for some annotations we can always add a custom `Processor` that would validate those. e.g. A non string method parameter annotated with `@TargetPropertyName`

filiphr (changes requested):
I finally to the time to have a proper look into this.

Thanks @sjaakd and @Zegveld for the reviews.

@ivlcic, I've pushed some small cleanup changes. Hope you are OK with them.

This is almost ready to be merged. I have 2 things to ask:

* Can we please add some documentation for this?
* Can we please move the tests to their own package and dedicated test. e.g. `org.mapstruct.ap.test.conditional.targetpropertyname`.  Would also be good to extract the different implementation of the `TargetPropertyNameModel` in their own classes. I think that it will make the tests a bit more readable.

ivlcic:
- I've added some (minimal) documentation (chapter 10).
- I've moved tests to its own package and moved classes as suggested.

ivlcic:
@Zegveld, @filiphr, @sjaakd ... Thanks for all your help with this!

One small side note:
In `processor/src/main/resources/org/mapstruct/ap/internal/model/macro/CommonMacros.ftl` we pass down `targetPropertyName=ext.targetPropertyName` and `targetType=ext.targetType` in `<#macro handleSourceReferenceNullCheck>`

Correction for `targetType=ext.targetType` is missing in `<#macro handleLocalVarNullCheck needs_explicit_local_var>` or it will fail in obscure usage of `@TargetType` when `@Conditional` method is defined in `@Context` variable.

Regards!

ivlcic:
I there anything more to be done from my side?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -149,6 +150,25 @@ else if ( pb.isTargetType() ) {
         }
     }
 
+    /**
+     * Adds default parameter bindings for the target-property-name if not already available
+     *
+     * @param availableParams Already available params, new entries will be added to this list
+     */
+    private <T extends Method> void addTargetPropertyNameBindings(List<ParameterBinding> availableParams) {
```
filiphr, line 158: Why not do what is done here in `addMappingTargetAndTargetTypeBindings`?

filiphr (commented):
I did a final pass of this and started to integrate it into main. I have only one question @ivlcic.

Can you please let me know what you think about that?

I'll merge and fix the conflcits afterwards (almost done with that locally)

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
ivlcic: Huh, I guess I was too narrow minded or too afraid of existing code changes.

I totally agree: additional method can be removed.

The question is will you do it, or shall I? 

If you prefer for me to do it, should I rename the private method to note all three actions i.e.: `addMappingTargetTargetTypeAndPropertyNameBindings` (:-S I don't like the name but the reality is such).

In any case If you want to change the code, I don't consider my consent necessary - since you're the expert.

ivlcic: Feel free to change the code in any way it makes more sense to you!

Zegveld: I agree with the naming `addMappingTargetAndTargetTypeAndPropertyNameBindings` being a bit too much, perhaps something like `addMissingAnnotationParamterBindings` would be shorter and still correct.

filiphr: I was only asking if there was any specific reason. Since there isn't, I'll adapt it. I've changed the name locally to `addTargetRelevantBindings`.

filiphr:
I've manually merged and polished this a bit in https://github.com/mapstruct/mapstruct/commit/8fa286fe4ce790cc2a65adff55c813bc9717437a. Thanks a lot for the contribution @ivlcic

PR #2844, opened 2022-05-06
#2839 keep thrown types when creating a new ForgedMethod with the same arguments

### LINKED ISSUE #2839: Compilation error when mapping fields with the same type due to not wrapping in a `try-catch` block
The mapper for the following class cannot be generated:

```java
...
public final class FinancingProductDto {
  ...
  private final List<CommitmentFeeDto> conditionedCommitmentFees;
  private final List<CommitmentFeeDto> unconditionedCommitmentFees;
...
```

The second mapped field is not wrapped in a `try-catch` block:

```java
...
        try {
            conditionedCommitmentFees = commitmentFeeDtoListToCommitmentFeeList( dto.getConditionedCommitmentFees() );
        }
        catch ( IllegalAccessException e ) {
        }
        catch ( InvocationTargetException e ) {
        }
        catch ( InstantiationException e ) {
        }
        catch ( NoSuchMethodException e ) {
        }
        unconditionedCommitmentFees = commitmentFeeDtoListToCommitmentFeeList( dto.getUnconditionedCommitmentFees() );
...
```

The following error is occurring: `unreported exception java.lang.IllegalAccessException; must be caught or declared to be thrown`

Here is an example project to reproduce the issue: https://github.com/hakan0xFF/mapstruct-mapping-issue

The error occurs also with the latest stable version `1.4.2.Final`.

hakan0xFF:
fixes #2839

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2839/CarMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2839/CarMapper.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2839;
+
+import java.util.UUID;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Hakan Özkan
+ */
+@Mapper
+public abstract class CarMapper {
+
+    public static final CarMapper MAPPER = Mappers.getMapper( CarMapper.class );
+
+    public abstract Car toEntity(CarDto dto);
+
+    public abstract CarDto toDto(Car entity);
+
+    protected Id mapId(UUID id) throws IllegalAccessException {
+        throw new IllegalAccessException();
+    }
+
+    protected UUID mapId(Id id) throws IllegalAccessException {
+        throw new IllegalAccessException();
+    }
+}
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2839/CarMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2839/CarMapper.java
@@ -5,8 +5,6 @@
  */
 package org.mapstruct.ap.test.bugs._2839;
 
-import java.util.UUID;
-
 import org.mapstruct.Mapper;
 import org.mapstruct.factory.Mappers;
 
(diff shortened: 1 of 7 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_2839/CarMapper.java
```diff
@@ -0,0 +1,32 @@
...
+public abstract class CarMapper {
+
+    public static final CarMapper MAPPER = Mappers.getMapper( CarMapper.class );
+
+    public abstract Car toEntity(CarDto dto);
+
+    public abstract CarDto toDto(Car entity);
+
+    protected Id mapId(UUID id) throws InstantiationException {
+        return id != null ? new Id( id ) : null;
```
sjaakd, line 26: It would be nice to actually trigger the exception based on the data and verify that scenario in your test.

processor/src/test/java/org/mapstruct/ap/test/bugs/_2839/CarMapper.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    protected UUID mapId(Id id) throws NoSuchMethodException {
+        return id != null ? id.getId() : null;
```
sjaakd, line 30: It would be be nice to trigger the exception based on the data and verify that.

sjaakd (changes requested):

processor/src/test/java/org/mapstruct/ap/test/bugs/_2839/CarMapper.java
hakan0xFF: Done. Thx for the feedback.

processor/src/test/java/org/mapstruct/ap/test/bugs/_2839/CarMapper.java
hakan0xFF: I've removed `throws` from this method as it's not needed here.

filiphr: It is not only for removing the `throws`. I think what @sjaakd wanted to see was to throw an exception in one of these methods and verify that in the tests.

hakan0xFF: Oh, I see. Done, the exceptions are thrown now. One additional thing that I noticed is that checked exceptions are failing however unchecked exceptions are just fine.

filiphr:
@hakan0xFF I've applied the requested changes by @sjaakd to your branch and pushed it. Once the tests are green I am going to merge this PR. Thanks for your work on it.

PR #2868, opened 2022-06-04
#2867 Fix NPE when reporting message on parent mappers

### LINKED ISSUE #2867: NullPointerException when reporting errors for non Mappers
With #598 we improved our diagnostics to show the errors from inherited classes on the mapper that is being created.

However, this caused an NPE.

See https://github.com/jhipster/generator-jhipster/pull/18820 from JHiipster as an example.

The problem that is trying to be solved in PR #2809, is the fix for this. We missed this, because the PR was lacking the information about why this is happening.

filiphr:
Fixes #2867


filiphr:
@sjaakd or @Zegveld if some of you can have a look at this one it would be great.

I'd like to do a 1.5.1.Final release with this fix.

Zegveld (approved):

Issue #2806, opened 2022-04-02
Date tests sometimes (less than 0.1% of the time) fail


Zegveld:
I've seen tests fail randomly, so this will also make some mappings inconsistent at rare moments. 
Chance of happening seems to be below 0.1% of the time, but would be nice if this was 0%.

1) DateConversionTest.shouldApplyDateFormatForConversionsJdk11
```
[INFO] Running org.mapstruct.ap.test.injectionstrategy.jsr330._default.Jsr330DefaultCompileOptionFieldMapperTest
Error:  Tests run: 22, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 0.906 s <<< FAILURE! - in org.mapstruct.ap.test.conversion.date.DateConversionTest
Error:  org.mapstruct.ap.test.conversion.date.DateConversionTest.shouldApplyDateFormatForConversionsJdk11  Time elapsed: 0.337 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 

Expecting:
 <"05.07.2013">
to be equal to:
 <"06.07.2013">
but was not.
	at org.mapstruct.ap.test.conversion.date.DateConversionTest.shouldApplyDateFormatForConversionsJdk11(DateConversionTest.java:66)
```

2) NestedMappingMethodInvocationTest
[Another failed test](https://github.com/mapstruct/mapstruct/runs/5611036594?check_suite_focus=true):
```
 [INFO] Running org.mapstruct.ap.test.nestedmethodcall.NestedMappingMethodInvocationTest
Error:  Tests run: 9, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.525 s <<< FAILURE! - in org.mapstruct.ap.test.nestedmethodcall.NestedMappingMethodInvocationTest
Error:  org.mapstruct.ap.test.nestedmethodcall.NestedMappingMethodInvocationTest.shouldMapViaMethodAndConversion  Time elapsed: 0.252 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 

Expecting:
 <2013-07-05T22:00:00.000 (java.util.Date)>
to be equal to:
 <2013-07-06T00:00:00.000 (java.util.Date)>
but was not.
	at org.mapstruct.ap.test.nestedmethodcall.NestedMappingMethodInvocationTest.shouldMapViaMethodAndConversion(NestedMappingMethodInvocationTest.java:74)
```

3) VariableNamingTest
[...]

PR #2869, opened 2022-06-05
Try to stabilise some date conversion tests 

filiphr:
Related to #2806. 

We try to stabilise some date conversion tests by read locking the default time zone.


sjaakd:
What is the purpose of changing one testcase? There are more of them that change locale or default timezone.. Shouldn't we do them all?

PR #2872, opened 2022-06-05
#2837 Add support for text blocks in expressions

### LINKED ISSUE #2837: Add support for Java Text Blocks
- [Y] Is this an issue (and hence not a question)? 

The following mapping does not work:
```
    @Mapping(target = "fieldName", expression =
            """
                java(
                    EnumType1.ENUM_VALUE_1 == dto.getValueOfEnum1()
                    ? EnumType2.ENUM_VALUE_2
                    : dto.getValueOfEnum2()
                )    
            """)
```
The compile-time error is `java: Value for expression must be given in the form "java(<EXPRESSION>)".`.

It's possible that the problem is at `org.mapstruct.ap.internal.model.source.MappingOptions` line 40:
`private static final Pattern JAVA_EXPRESSION = Pattern.compile( "^java\\((.*)\\)$" );`

In theory, this block of code should work:
```
entity.setFieldName(
                    EnumType1.ENUM_VALUE_1 == dto.getValueOfEnum1()
                    ? EnumType2.ENUM_VALUE_2
                    : dto.getValueOfEnum2());
```

Perhaps adding `Pattern.MULTILINE` would work?

filiphr:
This PR adds support for Java Text Blocks by changing the pattern we use to match an expression to use the `DOTALL` flag (dot matches new lines) and also adding extra whitespaces before `java` and after the closing of the expression. The extra whitespaces are there because a user might make a mistake when typing the expression and have leading whitespace in the string. We don't want to be too strict there.

Fixes #2837

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
@@ -37,7 +37,7 @@ import org.mapstruct.tools.gem.GemValue;
  */
 public class MappingOptions extends DelegatingOptions {
 
-    private static final Pattern JAVA_EXPRESSION = Pattern.compile( "^java\\((.*)\\)$" );
+    private static final Pattern JAVA_EXPRESSION = Pattern.compile( "^\\s*java\\((.*)\\)\\s*$", Pattern.DOTALL );
 
     private final String sourceName;
     private final String constant;
--- a/integrationtest/src/test/java/org/mapstruct/itest/tests/MavenIntegrationTest.java
+++ b/integrationtest/src/test/java/org/mapstruct/itest/tests/MavenIntegrationTest.java
@@ -122,6 +122,13 @@ public class MavenIntegrationTest {
     void recordsCrossModuleTest() {
     }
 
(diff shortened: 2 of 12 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
```diff
@@ -37,7 +37,7 @@
  */
 public class MappingOptions extends DelegatingOptions {
 
-    private static final Pattern JAVA_EXPRESSION = Pattern.compile( "^java\\((.*)\\)$" );
+    private static final Pattern JAVA_EXPRESSION = Pattern.compile( "^\\s*java\\((.*)\\)\\s*$", Pattern.DOTALL );
```
sjaakd, line 40: I guess that @Zegveld made a remark here during our meeting..

sjaakd (approved):

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
filiphr: Yea he did, he had a better idea for how to do this. Without using a `Pattern`, but I'll leave that improvement to him 😄

Issue #674, opened 2015-10-26
Support Java 8 java.util.Optional type


agudian:
It's a little bit more than adding a built-in conversion:
* For `Optional<T> prop` source property:
  * add check for `prop.isPresent()` in addition to the null-check (in every spot where we currently use our null-check assignment wrapper)
  * using `prop.get()` to fetch the value
* For `Optional<T> prop` target property:
  * use `Optional.empty()` instead of `null` for null-assignments, or `Optional.ofNullable(..)` for other assignments.

Optionally (pun intended) also support the equivalent in Guavas `com.google.common.base.Optional`, which uses different names for the factory methods.

gunnarmorling:
> add check for prop.isPresent() in addition to the null-check 

Do you really think *in addition*, not *instead*? Anyone setting an `Optional` seems to strongly dislike their co-workers ;)

In that light, related to the discussion around the "source value presence checking" strategy, I think there "IS_NULL" should be renamed to something like "DEFAULT" to abstract from the fact whether "is null" or "isPresent()" is invoked by the default means of value presence checking.

remmeier:
it would be great if the request here is implemented in a more generic fashion. Meaning that the developer has to possiblity to provide an own isPresent(sourceObj, sourcePropertyName, targetObj, targetPropertyName), ignore(...) or skip(...) method to skip the mapping process for some of the properties. Some additional use cases:

- JPA/Hibernate proxies: ignore proxies if not loaded (can kill the performance)
- security: filter properties the user is not privileged to see

leimer:
Is there a way to enable mapstruct to handle java.util.Optionals? I'd love to provide a basic mapper config which is able to unwrap an optional and reuse it in my concrete mappers.

I have an object which contains an optional address, which contains an optional contact, which contains an emailaddress. I would like to map this object into an object which only holds the emailaddress.
In a naive way this would look like:
`@Mapping(source = "address.contact.email", target = "email")`
Of course this won't work :-(

Can you help me?

leimer:
I found a way by adding special getters for fields with optionals:
```
    private Optional<Contact> contact;

    public Contact getNullableContact() {
        return contact.orElse(null);
    }
```
the mapping configuration is now:
`@Mapping(source = "nullableInvoiceAddress.nullableContact.email", target = "email")` 

any better ideas?

filiphr:
[...]

PR #2874, opened 2022-06-06
#674 Support Java 8 java.util.Optional type

ibogdanchikov:
Adds support for mapping between `java.util.Optional<T>` and `T`. See #674 for details.

Asking the maintainers for a code review. Thank you for your time.

This is my first time ever trying to contribute to an open source project, so I apologize in advance for anything I might have done wrong.

### DIFF
--- a/documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc
@@ -125,6 +125,8 @@ public interface CarMapper {
 * Between `java.net.URL` and `String`.
 ** When converting from a `String`, the value needs to be a valid https://en.wikipedia.org/wiki/URL[URL] otherwise a `MalformedURLException` is thrown.
 
+* Between `java.util.Optional<T>` and `T`.
+
 [[mapping-object-references]]
 === Mapping object references
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
@@ -23,6 +23,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.UUID;
 
 import org.mapstruct.ap.internal.model.common.Type;
@@ -41,6 +42,8 @@ public class Conversions {
     private final Map<Key, ConversionProvider> conversions = new HashMap<>();
     private final Type enumType;
     private final Type stringType;
+    private final Type objectType;
+    private final Type optionalType;
     private final TypeFactory typeFactory;
 
     public Conversions(TypeFactory typeFactory) {
@@ -48,6 +51,8 @@ public class Conversions {
 
         this.enumType = typeFactory.getType( Enum.class );
         this.stringType = typeFactory.getType( String.class );
+        this.objectType = typeFactory.getType( Object.class );
+        this.optionalType = typeFactory.getType( Optional.class );
 
         //native types <> native types, including wrappers
         registerNativeTypeConversion( byte.class, Byte.class );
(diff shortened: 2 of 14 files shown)


filiphr:
Thanks a lot for the PR @ibogdanchikov. On first sight it looks OK. 

I saw that you want for the build in method approach. However, I don't think that this is needed. I think that this would be more suited to work with inline conversions. And do things like `Optional.ofNullable( ... )` and `optional.get()` for the mapping for from and to.

In addition to this, for `Optional#get` to be able to be used I think that we need to add some out of the box conditional method that would do the check if the optional is present.

Basically the code that is generated to look something like:

```java
targetOptionalProperty.setToOptionalProp( Optional.ofNullable( source.getToOptionalProp() ) );
if ( source.getFromOptionalProp().isPresent() ) {
    targetOptionalProperty.setFromOptionalProp( source.getFromOptionalProp().get() );
}
```

filiphr:
>This is my first time ever trying to contribute to an open source project, so I apologize in advance for anything I might have done wrong.

By the way. Congrats on your first open source contribution. We are happy to see that you chose MapStruct 😄

ibogdanchikov:
> I saw that you want for the build in method approach. However, I don't think that this is needed. I think that this would be more suited to work with inline conversions.

I totally agree. I think this would be a much better solution. May I try and implement inline conversion instead? Would you be OK with that?

filiphr:
Yes of course @ibogdanchikov. Feel free to try the other approach. Let us know if you have any questions

ibogdanchikov:
@filiphr I've reimplemented Optional mapping using inline conversion. Could you please take a look?

jlamb1:
thank you for taking this on! @filiphr do you mind looking over this?

filiphr:
I've started looking into this and we want and will have this in 1.6. One thing I want to fix is the null checks that are being done on `Optional` properties. We shouldn't be doing that, and I am not sure how easy it would be to achieve this.

ibogdanchikov:
@filiphr Could you please elaborate on how would you like to fix the null checks? I would be very grateful if you could provide trivial examples showing in what ways the generated code would look wrong and how would you like it to look. Thank you.

filiphr:
Sorry for taking so long time @ibogdanchikov.

What I basically would like to see is instead of:

```java
if ( source.getFromOptionalProp() != null ) {
    target.setFromOptionalProp( source.getFromOptionalProp().orElse( null ) );
}
```

I would like to see


```java
if ( source.getFromOptionalProp().isPresent() ) {
    target.setFromOptionalProp( source.getFromOptionalProp().get() );
}
```

or if we had some out-of-the-box presence check method for optionals:

```java
if ( isPresent( source.getFromOptionalProp() ) ) {
    target.setFromOptionalProp( source.getFromOptionalProp().get() );
}
```

ibogdanchikov:
@filiphr Thank you very much for the examples, they were very helpful! I adjusted the implementation according to these examples.

Specifically, I added the following condition to `CommonMacros.ftl`:

```
<#if getSourceType().isOptionalType()>.isPresent()<#else> != null</#if>
```

and changed `OptionalToObjectConversion.getToExpression` to `"<SOURCE>.get()"`, so now it generates the following code:

```java
if ( source.getFromOptionalProp().isPresent() ) {
    target.setFromOptionalProp( source.getFromOptionalProp().get() );
}
```

Please let me now if this solution is acceptable. Thank you.

filiphr:
This looks better @ibogdanchikov. Thanks a lot. I'll get your branch and play around with it before merging it

Tristan107:
Hi @filiphr, did you find something blocking the merge of this feature ?

filiphr:
>Hi @filiphr, did you find something blocking the merge of this feature ?

Unfortunately, I haven't had the time I want to look better into this one. Recently there was another PR (#3183) for this functionality. So I'll need to analyze both and see what works best for MapStruct

filiphr:
I had some time to look into this and #3183. The approach which is taken in #3183 is more flexible and allows mapping between different types e.g. `Optional<Integer>` -> `Optional<String>`. However, in this PR that doesn't work. Therefore, I am going to close this in favour of #3183. I am really sorry for this @ibogdanchikov, I hope you understand.

MakraGitHub:
Optional<User> selectByEmail(String email) does not support constructor in spring boot myBatis query 
How to fixed it>

PR #2884, opened 2022-06-10
Log erroneousElement

micheljung:


micheljung:
Nevermind, the element is `null` anyway (at least in my case).

Issue #2891, opened 2022-06-15
`SubclassExhaustiveStrategy.RUNTIME_EXCEPTION` option does not work if the superclass has a non-empty constructor


prtnv:
- [x] Is this an issue (and hence not a question)? 

If the superclass has a non-empty constructor, then MapStruct ignores the `subclassExhaustiveStrategy = SubclassExhaustiveStrategy.RUNTIME_EXCEPTION` option when generating code

MapStruct 1.5.1.Final, JDK 17

Reproduction scenario:

```
  @BeanMapping( subclassExhaustiveStrategy = SubclassExhaustiveStrategy.RUNTIME_EXCEPTION )
  @SubclassMapping( source = Source1.class, target = Target1.class )
  @SubclassMapping( source = Source2.class, target = Target2.class )
  AbstractTarget map( AbstractSource source );

  abstract class AbstractTarget {
    private final String name;

    protected AbstractTarget(String name) {
      this.name = name;
    }
    // getter
  }

  class Target1 extends AbstractTarget {
    protected Target1(String name) {
      super(name);
    }
  }

  class Target2 extends AbstractTarget {
    protected Target2(String name) {
      super(name);
    }
  }

  abstract class AbstractSource {
    private final String name;

    protected AbstractSource(String name) {
      this.name = name;
    }
    // getter
  }

  class Source1 extends AbstractSource {
    protected Source1(String name) {
      super(name);
    }
  }

  class Source2 extends AbstractSource {
    protected Source2(String name) {
      super(name);
    }
  }
```
For this example MapStruct generates the following code
```
@Override
public AbstractTarget map(AbstractSource source) {
    if ( source == null ) {
        return null;
    }
    if (source instanceof Source1) {
        return source1ToTarget1( (Source1) source );
    }
    else if (source instanceof Source2) {
        return source2ToTarget2( (Source2) source );
    }
    else {
        String name = null;
        name = source.getName();
        AbstractTarget abstractTarget = new AbstractTarget( name );
        return abstractTarget;
    }
}
```
I was expecting a new IllegalArgumentException to be thrown in the last else statement. And also this generated code doesn't compile due to `new AbstractTarget( name )`

After some debugging, I think that if change the `isAbstractReturnType` method in the `BeanMappingMethod#isAbstractReturnType` class from
```
public boolean isAbstractReturnType() {
    return getFactoryMethod() == null && !hasConstructorMappings() && returnTypeToConstruct != null
[...]

PR #2893, opened 2022-06-18
#2891 Fix subclass mapping while superclass has non-empty constructor

prtnv:
Fixes https://github.com/mapstruct/mapstruct/issues/2891


filiphr:
thanks for the PR @prtnv. I had a look, and I realized that the bug is in another place. We should never have had the `hasConstructorMappings` check in that place. There shouldn't have been a constructor mapping method created for the abstract type.

I've pushed a change with that fix. Are you OK with it?

prtnv:
@filiphr thanks for feedback. All looks good for me

PR #2904, opened 2022-06-21
Feature/2892

### LINKED ISSUE #2892: Make test-utils a separate module
The mapstruct test-utils are handy for testing code generators.
I would like to also use these test-utils for testing our generators.

Can this be supplied as a separate module from the rest of the code.

Zegveld:
- [ ] Merge main branch and fix moved imports. (Periodically needed)
- [ ] Write in the README.md how to use the testutils in other modules.
- [ ] Stop passing the configuration interface along and have the static retrieval method cache the action.
- [ ] Make what the methods on the configuration interface are used for more obvious with javadoc and renaming.
- [ ] Move to another repository. (Finishing action)

closes #2892

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/internal/model/source/SelectionParametersTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/internal/model/source/SelectionParametersTest.java
@@ -5,14 +5,11 @@
  */
 package org.mapstruct.ap.internal.model.source;
 
-import static org.assertj.core.api.Assertions.assertThat;
-
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-
 import javax.lang.model.element.AnnotationMirror;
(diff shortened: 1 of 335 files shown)


filiphr:
This would ideally be in a new project under a new organization. It doesn't fit well with what MapStruct is doing.

Anyways, the move to a dedicated non publishable module is a good step to separate it more from MapStruct.

What I would suggest in order to avoid a lot of changes is to limit certain things for now.

* `IssueKey` should not be moved
* `WithJakartaInject`, `WithJavaxInject`, `WithJavaxJaxb`, `WithJoda`, `WithJoda` should not be moved. They are all meta annotations for `WithTestDependency`. 
* For easier review, I would also suggest if possible to do everything in the current packages and add new things that are need. e.g. the `ProjectTestConfiguration`
* Seems like checkstyle should be an optional dependency as well

I haven't done a thorough analysis of this yet though

Zegveld:
The diff should be a lot better now. The reverted changes were all import updates, because of moved items.

PR #2923, opened 2022-07-06
#2922 Fix protobuf tests for M1 Macs

### LINKED ISSUE #2922: Build fails when running mvn test on an M1 Mac
The problem is caused by the `MavenIntegrationTest.protobufBuilderTest`, because it cannot find the required `com.google.protobuf:protoc` library version for the `osx-aarch_64` architecture. Same applies to the `io.grpc:protoc-gen-grpc-java` library.

[mvn-test-protobuf-failure-log.txt](https://github.com/mapstruct/mapstruct/files/9054535/mvn-test-protobuf-failure-log.txt)

Versions of mentioned libraries used in the parent `pom.xml` and in the `protobuf-maven-plugin` are quite old and don't have any support for the `osx-aarch_64` architecture.

Minimal versions that allow build to succeed and tests to pass are:

* `com.google.protobuf:protoc:3.17.3`
* `io.grpc:protoc-gen-grpc-java:1.42.1`

These versions do not have native support for the `osx-aarch_64` architecture, but use this architecture's name as an alias for the `osx-x86_64` architecture that is backward compatible via the Rosetta 2 emulator.

I also tried the latest versions of mentioned libraries and they also seem to work fine (i.e. build succeeds and tests pass). `io.grpc:protoc-gen-grpc-java` still doesn't have native support for `osx-aarch_64`, but `com.google.protobuf:protoc` does since version `3.21.0`, so I personally would vote for using these latest versions, if that would be possible.

I'm ready to create a pull request if the core team will see no problems with updating the versions of these libraries.

ibogdanchikov:
Fixes #2922.

### DIFF
--- a/integrationtest/src/test/resources/protobufBuilderTest/pom.xml
+++ b/integrationtest/src/test/resources/protobufBuilderTest/pom.xml
@@ -23,7 +23,7 @@
 
     <properties>
         <os.mavenplugin.version>1.6.0</os.mavenplugin.version>
-        <protobuf.mavenplugin.version>0.5.1</protobuf.mavenplugin.version>
+        <protobuf.mavenplugin.version>0.6.1</protobuf.mavenplugin.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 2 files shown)


mamadaliev (approved):

filiphr:
Thanks for the fixes @ibogdanchikov

Issue #2378, opened 2021-03-11
Kotlin Code Generation and Empty Constructor Preference


Fyzxs:
- [x] Is this an issue (and hence not a question)? 

The key components of the issue I'm encountering
* OpenApi DTO specification has non-required fields
* Generation as immutable kotlin data classes
* Being generated, cannot use `@Default`
* MapStruct selecting empty constructor when present (w/o @Default)

OpenApi will generate Kotlin code for a DTO with unrequired fields as nullable. The Kotlin code will look something like
```kotlin
data class Example{
    val requiredField: String,
    val notRequiredField: String? = null
}
```
The compiled code has a primary constructor with all fields and an empty constructor. The empty constructor is preferred by MapStruct. Since the fields are immutable, no data gets transferred between the source and target object.
The constructor that should be used exists, there's just preference for the empty constructor.

My thought on this is to have an option to tell MapStruct to use the constructor with the most arguments.

filiphr:
This one is a tricky one @Fyzxs. What should happen if there are constructors with the same number of arguments?

Since we are talking about Kotlin Data classes, I think that this issue is the same as #2281. I haven't used Kotlin that much. However, I think that a possible solution would be if we can somehow pick the constructor that matches the data class in its full from the AST generated by Kotlin

Fyzxs:
At least part of #2281 is the same. The issue I'm encountering is that the empty constructor is being picked up and nothing mapped. I'm not encountering the issue with the `#copy` methods.

I think using the primary/default constructor out of the Kotlin class (independent of it being a data class) when available would be the best choice, or at least an option to configure it.

I was unable to see a way to get that constructor in the MapStruct code. I'm not familiar with the functionality used, so was fumbling around blind.

For a simple class
```kotlin
class Example(
    val imProp1: String,
    var mProp2: String? = null
)
```
there will be a constructor generated with this signature
```
public synthetic <init>(Ljava/lang/String;Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
```
(Note - Some of my actual code had an inserted `int` before the `DefaultConstructorMarker`, don't know why it's different here... It might not actually be different... when I decompile the bytecode to java, this is the signature that results
`public Example(String var1, String var2, int var3, DefaultConstructorMarker var4)`)

[...]

Issue #2281, opened 2020-11-24
Kotlin: problems with data classes


jakoss:
- [x] Is this an issue (and hence not a question)? 

I'm trying to use MapStruct with kotlin codebase. For the most part i'm pretty successfull, but i have 2 issues that are weird and require workarounds that shouldn't be necessary.

First issue that i have is that if i try to map data class with single field i have to explicitly ignore "copy" target in mapper annotation because i'm getting warning "Unmapped target property: "copy"". When i add second field - it's no longer needed. Example:
```kotlin
data class SinglePropertyTest(val property: String)
data class SinglePropertyTestDto(val property: String)

data class DoublePropertyTest(val property1: String, val property2: Int)
data class DoublePropertyTestDto(val property1: String, val property2: Int)

@Mapper
interface SinglePropertyMapper {
    // this gives me warning - Unmapped target property: "copy"
    fun toDto(singlePropertyTest: SinglePropertyTest): SinglePropertyTestDto

    // this works just fine
    fun toDto(doublePropertyTest: DoublePropertyTest): DoublePropertyTestDto
}
```

Seconds issue is more dangerous. If class have all fields nullable with default value set to null - mapper is silently failing to map any properties (all properties in target class are null). Example:
```kotlin
fun main(args: Array<String>) {
    val allNullClass = AllNullClass("test", null)
    val mapper = Mappers.getMapper(AllNullMapper::class.java)
    val mappedDto = mapper.toDto(allNullClass)

    // property1 here is null !
    assert(mappedDto.property1 != null)
    assert(mappedDto.property2 == null)
}

data class AllNullClass(
    val property1: String? = null,
    val property2: Int? = null
)

data class AllNullClassDto(
    val property1: String? = null,
    val property2: Int? = null
)

@Mapper
interface AllNullMapper {
    fun toDto(allNullClass: AllNullClass): AllNullClassDto
}
```

All in all - great library, almost perfect for my current usage!

filiphr:
Both problems seem related to the way the Kotlin byte code (AST) looks like when we try to create the mappers.

I guess that the first issue is due to the fact that Kotlin creates some method that looks like:

```
SinglePropertyTest#copy(String)
```

and since this is a fluent property we treat it as property. Seems like this property is not there when having multiple properties.

The second problem is the fact that Kotlin creates multiple constructors (one of them being the empty constructor) and thus MapStruct picking that one.

[...]

PR #2933, opened 2022-07-11
Instantiate Kotlin data classes using the _primary_ constructor

jawo-a:
A lot of Kotlin projects define DTOs using data classes with optional parameters like
```kotlin
data class Person(
    val name: String? = null,
    val email: String? = null
)
```

Unfortunately these data classes do not work with MapStruct because the Kotlin compiler generates multiple constructors under the hood.
Issues #2281, #2378 describe this problem.

This PR is an attempt to make MapStruct use the primary constructor of the data class to instantiate it, and ignore all other generated constructors.

### How it works?
Recognizes data classes and the primary constructor by
- the [annotation `kotlin.Metadata`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-metadata/), that is present on any class file produced by the Kotlin compiler
- and by the presence of  `componentN()` methods, as described in the [documentation about data classes](https://kotlinlang.org/docs/data-classes.html).

Note: _There might be false positives._

### How to review?
- Start with the changes to [`BeanMappingMethod`](https://github.com/mapstruct/mapstruct/pull/2933/files#diff-a631248fddd6c9d017e345f4e71bbac0fa3eb1ff5e26801c45dba616867af23a).
- Then take a look at [`KotlinType`](https://github.com/mapstruct/mapstruct/pull/2933/files#diff-9681dbb19064dcbb27a6bbb4f8f45893a1f3e1ff81e9726fd21285492e521fcd).
- The rest are tests.

### TODO 
- [ ] Update the reference documentation

### Improvement ideas
- Instead of using the number of `componentN()` methods, use their exact return types to more accurately determine the primary constructor.

### DIFF
--- a/integrationtest/src/test/resources/kotlinDataTest/src/main/java/org/mapstruct/itest/kotlin/data/CustomerEntity.java
+++ b/integrationtest/src/test/resources/kotlinDataTest/src/main/java/org/mapstruct/itest/kotlin/data/CustomerEntity.java
@@ -11,7 +11,9 @@ package org.mapstruct.itest.kotlin.data;
 public class CustomerEntity {
 
     private String name;
+    private int age;
     private String mail;
+    private String job;
 
     public String getName() {
         return name;
@@ -21,11 +23,27 @@ public class CustomerEntity {
         this.name = name;
     }
 
+    public int getAge() {
+        return age;
+    }
+
+    public void setAge(int age) {
+        this.age = age;
+    }
+
     public String getMail() {
         return mail;
     }
 
     public void setMail(String mail) {
         this.mail = mail;
+    }
+
+    public String getJob() {
+        return job;
+    }
+
+    public void setJob(String job) {
+        this.job = job;
     }
 }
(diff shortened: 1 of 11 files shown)


sskrla:
You could also maybe use the generated copy method as a form of detection?

jawo-a:
@sskrla  Yes, I've taken a look at decompiled data classes and the generated `copy()` method has exactly the same parameters as the primary constructor, although I'm not sure if that is actually guaranteed by Kotlin.

filiphr:
This is an interesting approach @jawo-a. However, I am not sure that I like the taken approach with using the component method count. I would prefer in using the [kotlinx-metadata-jvm](https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/ReadMe.md) library as an optional dependency instead. Perhaps with a `mapstruct-kotlin-extensions` or some other extension that will pull the needed Kotlin dependencies and that will have the needed code to make Kotlin things work properly.

DavidRomao:
Hi @filiphr , how would you suggest that `mapstruct-kotlin-extensions` extension be done ? Would it be okay through the creation of a Constructor Provider SPI, where that extenstion package would then implement  ? Otherwise mapstruct would load the current implementation

jawo-a:
@DavidRomao I've already prepared a change with a Constructor Provider SPI. Going to open a PR in the next few weeks.

MiguelAngelLV:
Any change with this?

filiphr:
Closing this in favor of #3978

PR #2959, opened 2022-08-15
Feature/2954 allowing generic mappings

### LINKED ISSUE #2954: Generic type use , An error was obtained  --- Can't generate mapping method for a generic type variable source.
**background**

I want to mapping entity to VO  for PageInfo，The code :

```
@Mapper
public interface PageInfoConvert {
    PageInfoConvert INSTANCE = Mappers.getMapper(PageInfoConvert.class);

    @Mapping(target = "list",expression  = "java(voList)")
    <T,S> PageInfo<T> convert(PageInfo<S> pageInfo, @Context List<T> voList);
}
```

When compile, an error was obtained  ------ Can't generate mapping method for a generic type variable source.

**environment**
Jdk : 11
Mapstruct：1.5.0.RC1

I didn't get answer on the Internet
Please help me. Thanks

Zegveld:
First step in allowing generic mappings as long as the generic itself doesn't change.

closes #2954

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -96,6 +96,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
     private final MethodReference finalizerMethod;
 
     private final MappingReferences mappingReferences;
+    private List<Type> typeParameters;
 
     public static class Builder extends AbstractMappingMethodBuilder<Builder, BeanMappingMethod> {
 
@@ -1758,6 +1759,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         }
         this.returnTypeToConstruct = returnTypeToConstruct;
         this.subclassMappings = subclassMappings;
+        this.typeParameters = method.getTypeParameters();
     }
 
     public List<PropertyMapping> getConstantMappings() {
@@ -1801,6 +1803,10 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
     public MethodReference getFinalizerMethod() {
         return finalizerMethod;
+    }
+
+    public List<Type> getTypeParameters() {
+        return typeParameters;
     }
 
     @Override
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
@@ -539,11 +539,6 @@ public class MethodRetrievalProcessor implements ModelElementProcessor<Void, Lis
                     messager.printMessage( method, Message.RETRIEVAL_WILDCARD_SUPER_BOUND_SOURCE );
                     return false;
                 }
-
-                if ( typeParameter.isTypeVar() ) {
-                    messager.printMessage( method, Message.RETRIEVAL_TYPE_VAR_SOURCE );
-                    return false;
-                }
             }
         }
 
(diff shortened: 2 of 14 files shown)


filiphr:
I am not sure that we should relax the requirement we have for this. However, I am going to look into the PR into detail a bit later.

filiphr:
I need to play around with this to see what happens when the sources are not generic themselves.

processor/src/test/java/org/mapstruct/ap/test/generics/GenericsTest.java
```diff
@@ -8,27 +8,140 @@
...
+        GenericTargetContainer.class,
+        GenericSourceContainer.class,
+        GenericContainerMapper.class,
+    })
+    public void mapsSameTypeCorrectly() {
+        String containedObject = "Test";
+        GenericSourceContainer<String> container = new GenericSourceContainer<>( containedObject, "otherValue" );
+
+        GenericTargetContainer<String> resultString = GenericContainerMapper.INSTANCE.map( container );
+        assertThat( resultString.getContained() ).isSameAs( containedObject );
```
filiphr, line 68: Doing `isSameAs` with a String isn't really recommended. If we want to test this properly we should use some other object.

processor/src/test/java/org/mapstruct/ap/test/generics/GenericsTest.java
```diff
@@ -8,27 +8,140 @@
...
+        assertThat( resultString.getOtherValue() ).isEqualTo( container.getOtherValue() );
+
+        Integer replacement = 1234;
+        GenericTargetContainer<Integer> resultContext =
+            GenericContainerMapper.INSTANCE.mapWithContext( container, replacement );
+        assertThat( resultContext.getContained() ).isSameAs( replacement );
```
filiphr, line 74: Same story here, we can't be using `isSameAs` for integer. We should try with actual objects here.

processor/src/test/java/org/mapstruct/ap/test/generics/GenericsTest.java
```diff
@@ -8,27 +8,140 @@
...
+        assertThat( resultContext.getOtherValue() ).isEqualTo( container.getOtherValue() );
+
+        GenericTargetContainer<Integer> resultSecondParameter =
+            GenericContainerMapper.INSTANCE.mapWithSecondParameter( container, replacement );
+        assertThat( resultSecondParameter.getContained() ).isSameAs( replacement );
```
filiphr, line 79: Same story here for the `isSameAs`

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -96,6 +96,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
     private final MethodReference finalizerMethod;
 
     private final MappingReferences mappingReferences;
+    private List<Type> typeParameters;
```
filiphr, line 99: Why can't we make this final? I also see that this is not in the `getImportTypes()`. How will it with with things such as `extends` and `super`?

processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
```diff
@@ -552,7 +549,7 @@ private boolean checkParameterAndReturnType(ExecutableElement method, List<Param
         }
 
         for ( Type typeParameter : resultType.getTypeParameters() ) {
-            if ( typeParameter.isTypeVar() ) {
+            if ( resultType.isIterableOrStreamType() && typeParameter.isTypeVar() ) {
```
filiphr, line 552: Why do we need to check the `resultType` here?

processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
```diff
@@ -528,13 +519,19 @@ private boolean checkParameterAndReturnType(ExecutableElement method, List<Param
...
+            }
+
             for ( Type typeParameter : parameterType.getTypeParameters() ) {
                 if ( typeParameter.hasSuperBound() ) {
                     messager.printMessage( method, Message.RETRIEVAL_WILDCARD_SUPER_BOUND_SOURCE );
                     return false;
                 }
 
-                if ( typeParameter.isTypeVar() ) {
+                if ( parameterType.isIterableOrStreamType() && resultType.isIterableOrStreamType()
```
filiphr, line 533: Why do we need to check for `isIterableOrStreamType` here?

filiphr (commented):
I had a look at this @Zegveld and I have some questions for some things that I couldn't understand.

I am also not sure that support a type parameter for the `@Context` parameter makes sense here. The `@Context` parameter is meant to be passed to other places, e.g. `@AfterMapping`, `@BeforeMapping`, etc. How will this affect that functionality?

I think that we should limit the generic support only to valid source parameters and nothing more.

How does this work with `@MappingTarget` parameters?

Issue #2825, opened 2022-04-20
SubclassMapping stackoverflow exception


Zegveld:
When using subclass mappings the method itself is accepted as a possible conversion method.
This is not wanted, work around is to manually define the subclass mapping methods.

Reproduction scenario:
```
class Animal {
    String name;
}
class Dog extends Animal {
    String race;
}
class Cat extends Animal {
    String race;
}
class TargetAnimal {
    String name;
    String race;
}

@SubclassMapping( target = TargetAnimal.class, source = Dog.class )
@SubclassMapping( target = TargetAnimal.class, source = Cat.class )
TargetAnimal map(Animal source);
```

What happends is that the `TargetAnimal map(Animal)` method is called for each subclass, instead of new methods being generated.
The following 2 methods are wanted here:
```
TargetAnimal map(Dog dog);
TargetAnimal map(Cat cat);
```

Work-around is to create these 2 mapping methods manually.

PR #2962, opened 2022-08-18
fix: #2825 SubclassMapping stackoverflow exception

chenzijia12300:
fix [SubclassMapping stackoverflow exception](https://github.com/mapstruct/mapstruct/issues/2825)


filiphr (commented):
Thanks for working on this @chenzijia12300. I'll need to check it out locally, as I am not sure that the fix in the `MappingResolverImpl` is the right one.

filiphr (changes requested):
@chenzijia12300 I've adapted the change that is needed to fix this.

Can you perhaps explain it my why we need so many mappers in the tests? Isn't it enough to only have the `AnimalMapper` for this issue?

chenzijia12300:
Sorry for submitting more than one useless test class, I added multiple mappers at that time to better debug the logic of ResolvingAttempt class

filiphr:
>Sorry for submitting more than one useless test class, I added multiple mappers at that time to better debug the logic of ResolvingAttempt class

No need to apologise, I was only being curious. I've pushed a commit that removes the extra test classes. Will merge once the CI is done

PR #2968, opened 2022-08-20
#2949 Do not inverse inherit BeanMapping#ignoreUnmappedSourceProperties

### LINKED ISSUE #2949: Ignoring unmapped source properties in inverse inheritance
I ran into a problem when upgrading from `1.4.2.Final` to `1.5.2.Final` regarding ignoring unmapped source properties and inverse inheritance.

Consider the mapping:
```java
    @InheritInverseConfiguration(name = "toModel")
    public abstract Product toEntity(ProductDto model);

    @BeanMapping(ignoreUnmappedSourceProperties = { "property1", "property2", "property3" })
    public abstract ProductDto toModel(Product entity);
```

In version `1.4.2.Final` the result was as expected with the properties being ignored in the generated mapper.
In compile-time I got the warning for `toEntity` with the message `Unmapped target properties; "property1, property2, property3` (with the quotes exactly as I wrote them), which is weird in itself as I expected that these properties would be ignored as `source` and `target` switch places in inverse configuration.
Nevertheless, this was only a warning and I was not bothered by it.

Then I upgraded `mapstruct` to version `1.5.2.Final` and I started getting the error `Ignored unknown source properties: "property1, property2, property3`.
It seems that the `ignoreUnmappedSourceProperties` was inherited as-is and the role was not switched from `source` to `target`.
The `toEntity` method behaves like I want to ignore `source` properties when I actually want to ignore `target` properties as read from the inverse configuration.

There does not seem to be a way to lower the level from `ERROR` to `WARN` to at least suppress the compile error.
I can work around this concrete issue explicitly defining `@Mapping(target = "property1", ignore = true)` on the `toEntity` method for each property.
But this means that I have to duplicate all property names on both sides in all the mappers currently using this inverse configuration approach.

This is probably related to #2481 as this error message was introduced with that change.

It seems that this is a bug.. If I want to ignore some source properties there is a high chance that these properties do not exist on the target type.
In that case I will always get a compile error unless I duplicate the property names using the `@Mapping` annotation.
So I have to approach in two different ways to ignore properties that exist in only 1 of these classes.

filiphr:
Fixes #2949

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
@@ -59,7 +59,7 @@ public class BeanMappingOptions extends DelegatingOptions {
                                                    TypeUtils typeUtils, TypeFactory typeFactory
     ) {
         if ( beanMapping == null || !isConsistent( beanMapping, method, messager ) ) {
-            BeanMappingOptions options = new BeanMappingOptions( null, null, null, mapperOptions );
+            BeanMappingOptions options = new BeanMappingOptions( null, Collections.emptyList(), null, mapperOptions );
             return options;
         }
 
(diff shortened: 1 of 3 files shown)


sjaakd (approved):
LGTM

Issue #2773, opened 2022-02-28
Copy `@Deprecated` annotation from method or mapper to implementation


filiphr:
As requested in https://github.com/mapstruct/mapstruct/issues/1574#issuecomment-881692018 if a method or a mapper that MapStruct implements is annotated with`@Deprecated` we need to copy this annotation in the generated code.

PR #2969, opened 2022-08-20
feat：#2773

chenzijia12300:
Try to implement feat: #2773

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -85,6 +85,7 @@ import static org.mapstruct.ap.internal.util.Message.PROPERTYMAPPING_CANNOT_DETE
  */
 public class BeanMappingMethod extends NormalTypeMappingMethod {
 
+    private final List<Annotation> annotations;
     private final List<PropertyMapping> propertyMappings;
     private final Map<String, List<PropertyMapping>> mappingsByParameter;
     private final Map<String, List<PropertyMapping>> constructorMappingsByParameter;
@@ -112,6 +113,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         private final Set<Parameter> unprocessedSourceParameters = new HashSet<>();
         private final Set<String> existingVariableNames = new HashSet<>();
         private final Map<String, Set<MappingReference>> unprocessedDefinedTargets = new LinkedHashMap<>();
+        private final List<Annotation> annotations = new ArrayList<>();
 
         private MappingReferences mappingReferences;
         private MethodReference factoryMethod;
@@ -214,6 +216,12 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 // If the return type cannot be constructed then no need to try to create mappings
                 return null;
             }
+            AdditionalAnnotationsBuilder additionalAnnotationsBuilder =
+                new AdditionalAnnotationsBuilder(
+                    ctx.getElementUtils(),
+                    ctx.getTypeFactory(),
+                    ctx.getMessager() );
+            annotations.addAll( additionalAnnotationsBuilder.getProcessedAnnotations( method.getExecutable() ) );
 
             /* the type that needs to be used in the mapping process as target */
             Type resultTypeToMap = returnTypeToConstruct == null ? method.getResultType() : returnTypeToConstruct;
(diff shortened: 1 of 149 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/Mapper.java
```diff
@@ -155,7 +155,11 @@ private Mapper(TypeFactory typeFactory, String packageName, String name,
         this.customPackage = customPackage;
         this.customImplName = customImplName;
         customAnnotations.forEach( this::addAnnotation );
-
+        TypeElement typeElement = this.getMapperDefinitionType().getTypeElement();
+        if ( typeElement.getAnnotation( Deprecated.class ) != null
+                && typeFactory.isTypeAvailable( Deprecated.class.getCanonicalName() )) {
```
filiphr, line 160: I don't think that we need to check if `java.lang.Deprecated` is available. This comes from `java.base`, which means it will always be available on the classpath.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -95,13 +95,10 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
     private final Type returnTypeToConstruct;
     private final BuilderType returnTypeBuilder;
     private final MethodReference finalizerMethod;
-
```
filiphr, line 98: Can you please revert all your changes in this file since they are only removing and adding lines

processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
```diff
@@ -10,6 +10,9 @@
     <#nt><@includeModel object=annotation/>
 </#list>
 <#if overridden>@Override</#if>
+<#if deprecated>
+    @Deprecated
+</#if>
```
filiphr, line 15: Can we perhaps change this and all the other places to:
```suggestion
<#if deprecated>@Deprecated</#if>
```

filiphr (changes requested):
Thanks for the changes @chenzijia12300.

I've left some comments. Something more general, I do not think that adding a default `deprecatedMethod`  to the top level `Method` is the right approach for this. While reviewing this one I realised that we are missing some things for annotations in other methods and I opened #2983.

Sorry, but we'll need to wait a bit before we can integrate this one. Would you be interested in working on some other issue in the meantime?

processor/src/main/java/org/mapstruct/ap/internal/model/Mapper.java
chenzijia12300: Thank you for pointing it out, it has been changed

processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
chenzijia12300: If the if statements are all on the same line, it will output an extra blank line when @deprecated is not present, so I want to avoid it

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
chenzijia12300: Thank you for pointing it out, it has been changed

chenzijia12300:
@filiphr Thank you for reviewing my code and pointing out the relevant bad code, I'm happy to deal with other issues

filiphr:
I think that we can finally come back to this one @chenzijia12300 and I think that doing it from scratch would be the best.

I think that the best solution for this is to do this in `AdditionalAnnotationsBuilder` by overriding the `getProcessedAnnotations` method and adding the `@Deprecated` if the `Element` is annotated with `@Deprecate`. You won't need to do any other changes then I believe, since everything will be handled properly by the new logic you added in #2986. 

Do you agree with this @chenzijia12300?

chenzijia12300:
@filiphr Sounds like a great idea, I'll try to implement it

processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
filiphr: OK I didn't know that. Good point. In any case, we already agreed for the new approach, so we won't need any changes in the ftl files.

filiphr:
Closing this PR in favour of #2996

PR #2970, opened 2022-08-21
Add support for Jakarta CDI

### LINKED ISSUE #2950: Support for Jakarta @ApplicationScoped is missing
https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#bean_defining_annotations defines a bunch of bean defining annotations. 

Prior to Jakarta, 'cdi' component model on a mapper resulted in an @ApplicationScoped implementation class, which could get properly resolved.

However, with Jakarta CDI spec 3.0, @ApplicationScoped no longer 'sits' in javax. but in jakarta. The result of this is that mapstruct cannot be used in such an arrangement.

filiphr:
Fixes #2950

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/CdiComponentProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/CdiComponentProcessor.java
@@ -12,6 +12,8 @@ import java.util.List;
 import org.mapstruct.ap.internal.gem.MappingConstantsGem;
 import org.mapstruct.ap.internal.model.Annotation;
 import org.mapstruct.ap.internal.model.Mapper;
+import org.mapstruct.ap.internal.model.common.Type;
+import org.mapstruct.ap.internal.util.AnnotationProcessingException;
 
 /**
  * A {@link ModelElementProcessor} which converts the given {@link Mapper}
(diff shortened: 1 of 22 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/CdiComponentProcessor.java
```diff
@@ -48,4 +50,24 @@ protected boolean requiresGenerationOfDecoratorClass() {
     protected boolean additionalPublicEmptyConstructor() {
         return true;
     }
+
+    private Type getType(String simpleName) {
```
sjaakd, line 54: `getCDIType`

sjaakd (approved):

processor/src/main/java/org/mapstruct/ap/internal/processor/CdiComponentProcessor.java
filiphr: I didn't want to complicate things. This also fetches the `@Inject` from the inject package as well

Felk:
Hello, I am curious if this is landing in a release any time soon? I am playing around with [Quarkus 3.0.0 Alpha4](https://github.com/quarkusio/quarkus/releases/tag/3.0.0.Alpha4), which includes the transition from the `javax` to `jakarta` namespace. Being able to just switch the component model from `CDI` to `JAKARTA_CDI` would mean mapstruct is well-prepared with a clear migration path once Quarkus 3.0 hits stable 😁

edeandrea:
I too am looking for this feature because of Quarkus 3 & the move to the jakarta namespace. RIght now I really have no workaround as `@Singleton` does not work (because I mock/spy using Mockito) and you can't mock/spy on `@Singleton` beans. It needs to be `@ApplicationScoped`.

edeandrea:
@filiphr / @sjaakd Please see previous comments.

filiphr:
This has now been backported to 1.5.4

PR #2972, opened 2022-08-21
#2897 Always import types defined in `Mapper#imports`

### LINKED ISSUE #2897: Missing import of nested class
I have classes with imports of nested class. _UtilClass.FactoryClass_ in example below. 
This worked in 1.4.2.. After upgrade to 1,5 generated code lacks import declared in annotation property 

```
@Mapper(imports = UtilClass.FactoryClass.class)
public interface ExampleMapper {

  @Mapping(target = "firstName", expression = "java(FactoryClass.doSomething( source ))")
  Person map(Source source);
}
```

filiphr:
Fixes #2897

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
@@ -307,7 +307,7 @@ public class MapperCreationProcessor implements ModelElementProcessor<List<Sourc
 
 
         for ( TypeMirror extraImport : mapperOptions.imports() ) {
-            Type type = typeFactory.getType( extraImport );
+            Type type = typeFactory.getAlwaysImportedType( extraImport );
             extraImports.add( type );
         }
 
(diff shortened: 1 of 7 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -307,7 +307,7 @@ private SortedSet<Type> getExtraImports(TypeElement element,  MapperOptions mapp
 
 
         for ( TypeMirror extraImport : mapperOptions.imports() ) {
-            Type type = typeFactory.getType( extraImport );
+            Type type = typeFactory.getAlwaysImportedType( extraImport );
```
sjaakd, line 310: Would `explicit` be a better term? Or even `explicitImportByMappingAnnotation`..

sjaakd: some javadoc otherwise..

sjaakd (approved):
suggestion for naming added

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
filiphr: I named it `getAlwaysImportedType` because we might need to use it for something else in the future. You never know what can happen.

I'll add Javadoc though

PR #2973, opened 2022-08-21
#2945 Stabilise top level imports

### LINKED ISSUE #2907: Unused import warning of nested classes
There are some unused import warnings on generated classes on version 1.5.2 that was not happening on 1.4.x

When I have a Mapper that uses a nested class for example:

Source:
```
package demo;

import java.util.Set;
import lombok.Data;

@Data
public class Source {
	private Set<SourceNested> nested;
}
```

```
package demo;

import lombok.Data;

@Data
public class SourceNested {
	private String prop;
}
```

Target:

```
package demo;

import lombok.Data;

@Data
public class Target {
	
	private TargetNested[] nested;
	
	@Data
	public static class TargetNested {
		private String prop;
	}
}
```

Mapper:
```
package demo.mapper;

import org.mapstruct.Mapper;

import demo.Source;
import demo.Target;

public interface DemoMapper {

    Target map(Source source);
    
}
```

Then the generated class have an unnecessary import

```
package demo.mapper;

import demo.Source;
import demo.SourceNested;
import demo.Target;
import demo.Target.TargetNested; // <------------- HERE IS WARNING. NOT NEED THIS IMPORT
import java.util.Set;
import org.springframework.stereotype.Component;

@Generated(
    date = "2022-06-24T19:26:33+0900",
    comments = "version: 1.5.2.Final, compiler: Eclipse JDT (IDE) 1.4.100.v20220318-0906, environment: Java 17.0.3 (Eclipse Adoptium)"
)
@Component
public class DemoMapperImpl implements DemoMapper {

    @Override
        if ( source == null ) {
            return null;
        }

        Target target = new Target();

        target.setNested( sourceNestedSetToTargetNestedArray( source.getNested() ) );

        return target;
    }

    protected Target.TargetNested sourceNestedToTargetNested(SourceNested sourceNested) {
        if ( sourceNested == null ) {
            return null;
        }

        Target.TargetNested targetNested = new Target.TargetNested();

        targetNested.setProp( sourceNested.getProp() );

        return targetNested;
    }

    protected Target.TargetNested[] sourceNestedSetToTargetNestedArray(Set<SourceNested> set) {
        if ( set == null ) {
            return null;
        }

        Target.TargetNested[] targetNestedTmp = new Target.TargetNested[set.size()];
        int i = 0;
        for ( SourceNested sourceNested : set ) {
            targetNestedTmp[i] = sourceNestedToTargetNested( sourceNested );
            i++;
        }

        return targetNestedTmp;
    }
}
```

Source code example:
[demo.zip](https://github.com/mapstruct/mapstruct/files/8975386/demo.zip)

### LINKED ISSUE #2945: Missing import in generated mapper when referencing a nested enum from an unrelated class
Given a nested enumeration,
```java
public class TestEnumHolder {
    public enum TestEnum {
        VALUE_1,
        VALUE_2;
    }
}
```

A source class (record for brievty),
```java
public record TestEntity(String testProperty) {}
```

A target class,
```java
public class TestDTO {
    private TestEnumHolder.TestEnum testProperty;

    public TestEnumHolder.TestEnum getTestProperty() {
        return testProperty;
    }

    public void setTestProperty(TestEnumHolder.TestEnum testProperty) {
        this.testProperty = testProperty;
    }
}
```

A mapper,
```java
@Mapper
public interface TestMapper {
    TestDTO map(TestEntity entity);
}
```

When compiling the mapper with Mapstruct 1.5.2, the following error occurs:
`TestMapperImpl.java:[23,66] package TestEnumHolder does not exist`

When inspecting the generated mapper source, the import is missing for the enum.
This used to work with Mapstruct 1.4.1.

This is the reproduction case packaged as a Maven project:
[mapstruct-nested-enum-repro.zip](https://github.com/mapstruct/mapstruct/files/9218836/mapstruct-nested-enum-repro.zip)

filiphr:
Make sure that GeneratedType always gets the imported types
from a Type before adding them

Fixes #2945 
Fixes #2907

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
@@ -258,8 +258,10 @@ public abstract class GeneratedType extends ModelElement {
             return;
         }
 
-        if ( needsImportDeclaration( typeToAdd ) ) {
-            collection.add( typeToAdd );
+        for ( Type type : typeToAdd.getImportTypes() ) {
+            if ( needsImportDeclaration( type ) ) {
+                collection.add( type );
+            }
         }
     }
 
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2907/Issue2907Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2907/Issue2907Test.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2907;
+
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.mapstruct.ap.test.bugs._2907.mapper.Issue2907Mapper;
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.ProcessorTest;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.GeneratedSource;
+
+/**
+ * @author Filip Hrisafov
+ */
+@IssueKey("2907")
+@WithClasses({
+    Issue2907Mapper.class,
+    Source.class,
+    SourceNested.class,
+    Target.class,
+})
+class Issue2907Test {
+
+    @RegisterExtension
+    final GeneratedSource generatedSource = new GeneratedSource();
+
+    @ProcessorTest
+    void shouldNotGeneratedImportForNestedClass() {
+        generatedSource.forMapper( Issue2907Mapper.class )
+            .containsNoImportFor( Target.TargetNested.class );
+    }
+}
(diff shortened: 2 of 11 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
```diff
@@ -258,8 +258,10 @@ protected void addIfImportRequired(Collection<Type> collection, Type typeToAdd)
             return;
         }
 
-        if ( needsImportDeclaration( typeToAdd ) ) {
-            collection.add( typeToAdd );
```
sjaakd, line 262: ok.. how could this work in the first place? Now we don't add typeToAdd itself, but the required imported types..

sjaakd (approved):
Just a question

processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
filiphr: Have a look at the `Type` implementation. We are a bit smart there when we need to do an import and only use the `TopLevel.Nested` for nested. There is a complex logic for computing whether we need to import a type or not.

When I did the initial work for this I changed a lot of places where we were adding types to actually return `Type#getImportedTypes`, but apparently this is not everywhere (considering the bug reports we've been receiving)

PR #2984, opened 2022-08-24
Stabilise top level imports

filiphr:
Some more stabilisation because it seems like the combination of
* https://github.com/mapstruct/mapstruct/pull/2973
* https://github.com/mapstruct/mapstruct/pull/2972

Are not compatible with each other, and it lead to some test failures on main

Fixes #2990

PR #2985, opened 2022-08-25
#2950 Disable CDI in the full features tests on Java 8

filiphr:
Fixes problems on Java 8 when running the full feature tests with CDI (the new tests were added for #2950). It depends on #2984 for the processor tests to pass


Zegveld:
#2990 fixes the failed tests of the non-JDK 8 builds.

Zegveld (approved):

PR #2992, opened 2022-08-28
#2952 Do not treat a getter as an alternative write accessor when using CollectionMappingStrategy#TARGET_IMMUTABLE

### LINKED ISSUE #2952: Private collection field with getter gives incorrect compilation error
Erroneous `Unmapped target property` compilation error when private collection field has a `get` method but no `set` method. 

Versions: Java 11, Mapstruct 1.5.2-Final.

Given the following classes:

```java
public class Source {
    private String sourceField;

    public Source() {
        // no-op
    }

    public String getSourceField() {
        return sourceField;
    }

    public void setSourceField(String sourceField) {
        this.sourceField = sourceField;
    }
}

public class Target {
    private final Map<String, String> attributes = new HashMap<>();
    private String targetField;

    public Target() {
        // no-op
    }

    public Map<String, String> getAttributes() {
        return attributes;
    }

    public String getTargetField() {
        return targetField;
    }

    public void setTargetField(String targetField) {
        this.targetField = targetField;
    }
}

@Mapper
public interface S2TMapper {
    @Mapping(target = "targetField", source = "sourceField")
    Target map(Source source);
}
```
When the compiler is run we get the following compilation error.

image

If the `getAttributes` method is ommitted, everything compiles fine, so it seems Mapstruct mistakenly believes the field should be writable on the basis that it has an accessor.

NOTE: I haven't tested this with other collection type classes.

Reporoducible project [here](https://github.com/andye2004/mapstruct-issue.git)

### LINKED ISSUE #2967: CollectionMappingStrategy.TARGET_IMMUTABLE does not ignore target collections with getters
The problem arises from #2952.

Having the example from there:

```java
public class Source {
    private String sourceField;

    public Source() {
        // no-op
    }

    public String getSourceField() {
        return sourceField;
    }

    public void setSourceField(String sourceField) {
        this.sourceField = sourceField;
    }
}

public class Target {
    private final Map<String, String> attributes = new HashMap<>();
    private String targetField;

    public Target() {
        // no-op
    }

    public Map<String, String> getAttributes() {
        return attributes;
    }

    public String getTargetField() {
        return targetField;
    }

    public void setTargetField(String targetField) {
        this.targetField = targetField;
    }
}

@Mapper
public interface S2TMapper {
    @Mapping(target = "targetField", source = "sourceField")
    Target map(Source source);
}
```

We can expect that using `@Mapper(collectionStrategy = CollectionStrategy.TARGET_IMMUTABLE)` is going to ignore the map or collection property that does not have a setter. However, that is not the case.

To top it of, if the `Source` has a field `attributes` of the same type then we are getting the error:

> No write accessor found for property "attributes" in target type.

filiphr:
Fixes #2967

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -782,6 +782,13 @@ public class Type extends ModelElement implements Comparable<Type> {
                     // an adder has been found (according strategy) so overrule current choice.
                     candidate = adderMethod;
                 }
+
+                if ( cmStrategy == CollectionMappingStrategyGem.TARGET_IMMUTABLE
+                    && candidate.getAccessorType() == AccessorType.GETTER ) {
+                    // If the collection mapping strategy is target immutable
+                    // then the getter method cannot be used as a setter
+                    continue;
+                }
             }
             else if ( candidate.getAccessorType() == AccessorType.FIELD  && ( Executables.isFinal( candidate ) ||
                 result.containsKey( targetPropertyName ) ) ) {
--- a/core/src/main/java/org/mapstruct/CollectionMappingStrategy.java
+++ b/core/src/main/java/org/mapstruct/CollectionMappingStrategy.java
@@ -12,8 +12,8 @@ package org.mapstruct;
  * Next, the trailing {@code s} indicates the plural form.
  * The table explains the options and how they are applied to the presence / absence of a
  * {@code set-s}, {@code add-} and / or {@code get-s} method on the target object.
- *
- * <table>
+ * <p>
+ * <table summary="Collection mapping strategy options">
  *     <tr>
  *         <th>Option</th>
  *         <th>Only target set-s Available</th>
--- a/core/src/main/java/org/mapstruct/CollectionMappingStrategy.java
+++ b/core/src/main/java/org/mapstruct/CollectionMappingStrategy.java
@@ -12,8 +12,8 @@ package org.mapstruct;
  * Next, the trailing {@code s} indicates the plural form.
  * The table explains the options and how they are applied to the presence / absence of a
  * {@code set-s}, {@code add-} and / or {@code get-s} method on the target object.
- * <p>
- * <table summary="Collection mapping strategy options">
+ * <table>
+ *     <caption>Collection mapping strategy options</caption>
  *     <tr>
  *         <th>Option</th>
  *         <th>Only target set-s Available</th>
--- a/core/src/main/java/org/mapstruct/CollectionMappingStrategy.java
+++ b/core/src/main/java/org/mapstruct/CollectionMappingStrategy.java
@@ -7,6 +7,54 @@ package org.mapstruct;
 
 /**
  * Strategy for propagating the value of collection-typed properties from source to target.
+ * <p>
+ * In the table below, the dash {@code -} indicates a property name.
+ * Next, the trailing {@code s} indicates the plural form.
+ * The table explains the options and how they are applied to the presence / absence of a
+ * {@code set-s}, {@code add-} and / or {@code get-s} method on the target object.
+ *
+ * <table>
+ *     <tr>
+ *         <th>Option</th>
+ *         <th>Only target set-s Available</th>
+ *         <th>Only target add- Available</th>
+ *         <th>Both set-s/add- Available</th>
+ *         <th>No set-s/add- Available</th>
+ *         <th>Existing Target ({@code @TargetType})</th>
+ *     </tr>
+ *     <tr>
+ *         <td>{@link #ACCESSOR_ONLY}</td>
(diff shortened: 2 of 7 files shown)


filiphr:
@sjaakd if you have some time it would be good to have a look at this one. Especially, because we are changing a test case, which in my opinion should have never been done like that.

Basically, if the `CollectionMappingStrategy#TARGET_IMMUTABLE` strategy is used collection / map getter methods cannot be used as alternative write accessors, because we'll ignore them anyways. Therefore, I believe that we should not even include them when figuring out the unmapped target properties

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -782,6 +782,13 @@ else if ( candidate.getAccessorType() == AccessorType.GETTER ) {
                     // an adder has been found (according strategy) so overrule current choice.
                     candidate = adderMethod;
                 }
+
+                if ( cmStrategy == CollectionMappingStrategyGem.TARGET_IMMUTABLE
```
sjaakd, line 786: Hmmm. The [CollectionMappingStrategy](https://github.com/mapstruct/mapstruct/blob/main/core/src/main/java/org/mapstruct/CollectionMappingStrategy.java) documentation is pretty lame reading it at hindsight.  I've been reading it a couple of times, but its intended usage (e.g. JPA?) does not really become clear from the text.. does it? And what is the default behavior (if you omit the strategy all together)? 

Also, I guess the description for each value should describe the "regular" and the "update" behavior separately.

sjaakd: Why limit this only to  the `TARGET_IMMUTABLE` option? I would say that in general the presence of only a `getter` does not constitute a reason to give a "unmapped target property" message.  In my opinion, only when an accessor with the clear intention of writing (`setter`, `adder`) is there with no corresponding source (mapping) the "unmapped target property" message should be given.

I guess that the behavior is even backward compatible. For existing usage the users have "fixed" such errors. We simply drop the error in new usage (clear release notes though).

sjaakd (changes requested):

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
filiphr: We have an extensive documentation about the `CollectionMappingStrategy` (https://mapstruct.org/documentation/stable/reference/html/#collection-mapping-strategies). By default it is accessor only (check `Mapper#collectionMappingStrategy`. I think that the table in the docs is quite clear, perhaps we need to add that table to the javadoc.

> I would say that in general the presence of only a `getter` does not constitute a reason to give a "unmapped target property" message. 

I would tend to agree with you on this one. However, if we remove this here it would mean that there won't be any target property that you can map. And for collections and map we do support things like `target.getNames().add(name)` and also for update mappings doing things like
```
target.getNames().clear()
target.getNames().addAll()
```

This would be a rather larger breaking change, since getters will no longer be considered for populating collections.

sjaakd: > perhaps we need to add that table to the javadoc.

That would be a good idea 👍

sjaakd: > This would be a rather larger breaking change, since getters will no longer be considered for populating collections.

But I guess there are two separate things:
1.  using a getter when present: no problem with that.. keep the existing functionality
2. generating  a  "unmapped target property" fault when a getter is there and there's no corresponding source property. That's the thing I like to address in a more generic way.

filiphr: >That would be a good idea 👍

I'll add the table.

>generating a "unmapped target property" fault when a getter is there and there's no corresponding source property. That's the thing I like to address in a more generic way.

If we do this, isn't this a bit of a strange behaviour for the user? They had no warning / error before. They add a getter on the source side and not all of a sudden it gets mapped. It would be confusing in my opinion.

sjaakd: or for this specific case we change from an error to a warning..

filiphr: @sjaakd I've added the table to the javadoc. 

Regarding the other stuff, I think that it makes more sense to do it in a different issue.  Can you create it or should we wait for someone to actually need this?

sjaakd: I'm afraid we'll forget that.. To be honest. My comments dated from a month ago and I forgot already.. Anyway.. I leave it to you if you want to fix that.

sjaakd (approved):

filiphr:
Thanks for the review @sjaakd.

PR #2993, opened 2022-08-28
#1216 Pick candidate method with most specific return type

### LINKED ISSUE #1216: Ambiguous mapping method but there is a most specific mapping

### LINKED ISSUE #2840: Ambiguous mapping methods when upgrading to Mapstruct 1.5.0.RC1
When upgrading from Mapstruct 1.4.2 to 1.5.0.RC1 I get _Ambiguous mapping methods_ error.

Given the following MappingConfiguration I want to make use of the support of the new Jakarta annotations.

```java
@MapperConfig(
    componentModel = "jakarta",
    nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS,
    nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.IGNORE)
public class MicronautMappingConfig {}
```

The following mapper maps `Short` and `Integer` to their primitive value which used to work in Mapstruct 1.4.2

```java
@Mapper(config = MicronautMappingConfig.class)
public interface NumberousMapper {

    NumberDto toNumberDto(Short shortValue, Integer intValue);

    default int toInt(Number number) {
        return number.intValue();
    }

    default short toShort(Number number) {
        return number.shortValue();
    }
}

class NumberDto {

    private int intValue;
    private short shortValue;

    public int getIntValue() {
        return intValue;
    }

    public void setIntValue(int intValue) {
        this.intValue = intValue;
    }

    public short getShortValue() {
        return shortValue;
    }

    public void setShortValue(short shortValue) {
        this.shortValue = shortValue;
    }
}
```

With Mapstruct 1.5.0.RC1 I get the following error

```
NumberousMapper.java:8: error: Ambiguous mapping methods found for mapping property "Integer intValue" to int: int toInt(Number number), short toShort(Number number). See https://mapstruct.org/faq/#ambiguous for more info.
    NumberDto toNumberDto(Short shortValue, Integer intValue);
              ^
```

### LINKED ISSUE #2921: Upgrade from 1.4.1 to 1.5.2 broke primitive to wrapper classes mapping
We had a mapper that mapped `Integer` to `int` properties from an internal model to a dto (method1). In the same mapper we had a method that mapped from Integer to Short(method2). This method was used in an expression on a different mapping method(method3). After upgrading to 1.5.2 that method Integer->Short (method2) is now used instead of generating a proper Integer to int mapping like it did in 1.4.2(for method1). 
This is bad for values that uses more than 16 bits and very dangerous as it can pass a lot of tests.
Is there an annotation that can be put on a method to restrict it to be used only on expressions and not be picked up by mapstruct just by types?
Thanks

filiphr:
When there are multiple candidate methods returning different types.
We should be able to use the method with the most specific return type (if such a method exists)

Fixes #1216


processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/MethodSelectors.java
```diff
@@ -36,7 +36,9 @@ public MethodSelectors(TypeUtils typeUtils, ElementUtils elementUtils, TypeFacto
             new InheritanceSelector(),
             new CreateOrUpdateSelector(),
             new SourceRhsSelector(),
-            new FactoryParameterSelector() );
+            new FactoryParameterSelector(),
+            new MostSpecificResultTypeSelector()
```
sjaakd, line 40: I thought such Selector was there already.. I remember @agudian writing something with distances..

filiphr: I think that you are referring to the `InheritanceSelector`. That is is using distances. However, it is only for the source types. This one is for the return type. I was also contemplating whether to do this in the `TargetTypeSelector`. However, that was a bit too early, since you might have a least specific type with a qualifier

processor/src/test/java/org/mapstruct/ap/test/bugs/_2840/Issue2840Test.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+@IssueKey("2840")
+@WithClasses({
+    Issue2840Mapper.class,
+})
+class Issue2840Test {
+
+    @ProcessorTest
+    void shouldUseMethodWithMostSpecificReturnType() {
+        Issue2840Mapper.Target target = Issue2840Mapper.INSTANCE.map( (short) 10, 50 );
```
sjaakd, line 25: Should we create an update method as well? Does that make sense at all (or is that cover by the `InheritanceSelector` ? Man.. I really need to get back into this 😄

filiphr: Adding an update method isn't really needed. We are testing the picking of the custom conversion functions. Which means that the return type is always in those, it has nothing to do with the mapping method. The `InheritanceSelector` is for other things, like picking the right object factory method.

processor/src/test/java/org/mapstruct/ap/test/selection/resulttype/ErroneousAmbiguousMostSpecificResultTypeSelectingMapper.java
```diff
@@ -0,0 +1,31 @@
...
+public interface ErroneousAmbiguousMostSpecificResultTypeSelectingMapper {
+
+    @Mapping( target = "apple", source = "fruit")
+    AppleFamily map(FruitFamily fruitFamily);
+
+    default GoldenDelicious toGolden(IsFruit fruit) {
+        return fruit != null ? new GoldenDelicious( fruit.getType() ) : null;
+    }
+
+    default Apple toApple1(IsFruit fruit) {
```
sjaakd, line 24: I saw you used boxed type conversion. Should we have a testcase with the `Number` inheritance somewhere?

filiphr: Did you check `Issue2840Mapper` as part of this PR?

sjaakd:
Apparently i missed that one.. sorry. Hi ahead and merge

Op zo 4 sep. 2022 12:35 schreef Filip Hrisafov ***@***.***>:

filiphr:
I've removed the test cases we had for #2840 from this PR since the actual fix for that is in #3032. This PR is only adding a new feature

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/MostSpecificResultTypeSelector.java
```diff
@@ -0,0 +1,45 @@
...
+                                                                         Type returnType, SelectionCriteria criteria) {
+        if ( candidates.size() < 2 || !criteria.isForMapping() || criteria.getQualifyingResultType() != null) {
+            return candidates;
+        }
+
+        List<SelectedMethod<T>> result = new ArrayList<>();
+
+        for ( SelectedMethod<T> candidate : candidates ) {
+            if ( candidate.getMethod()
+                .getResultType()
```
sjaakd, line 32: since you specifically use the resultType.. should we have an update method test?

sjaakd (approved):

sjaakd:
General mechanism LGTM. One question though..

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/MostSpecificResultTypeSelector.java
filiphr: Good point. I'll try to add such test case

Issue #2947, opened 2022-08-01
Mapstruct ignores the generic bounds when using explicit mappings


bgK:
Given a generic source class with a member implementing an interface,
```java
public interface TestInterface {}

public record TestEntity<T extends TestInterface>(T data) {}
```

A target class,
```
public record TestDto(String data) {}
```

And a mapper,
```
@Mapper
public abstract class TestMapper {
    @Mapping(target = "data", source = "data") // Removing this works around the issue
    public abstract TestDto map(TestEntity<?> action);

    String mapToString(TestInterface rawData) {
        return rawData.toString();
    }
}
```

When compiling with Mapstruct 1.5.2, the following error occurs: `TestMapper.java:[12,5] Can't map property "Object data" to "String data". Consider to declare/implement a mapping method: "String map(Object value)".`.

Note that removing the explicit mapping `@Mapping(target = "data", source = "data")` works around the issue (Obviously not possible in more complex situations).

This used to work with Mapstruct 1.4.1.

This is the reproduction case packaged as a Maven project:
[mapstruct-generic-bounds.zip](https://github.com/mapstruct/mapstruct/files/9232708/mapstruct-generic-bounds.zip)

filiphr:
Thanks for the detailed issue examples and the repro case. This is indeed a bug and we need to look into fixing it.

filiphr:
This bug was caused by the fix of #2677. It will be tricky to fix it

Issue #2070, opened 2020-04-13
SourceReference refactorings


filiphr:
Similar as #2069, but for `SourceReference`. This is not needed for #73. However, with PR #2068 we realised that there can be some simplifications for the `SourceReference` as well.

sjaakd:
Also #2193 should be taken into consideration. When target is null, the logic inside source reference needs to resolve the source with the target name (and not in bean mapping at line 1071..(else branch) when no other mapping annotation field  (expression, constant, etc) is set.

filiphr:
How can `target` be `null`? That is not possible, `target` is the only mandatory parameter in the `@Mapping` annotation

PR #2995, opened 2022-08-28
#2947: fix for wildcard bounded values. 

Zegveld:
Got the given example situation working and all other tests still pass.
- [ ] Add testcase with super wildcard.
- [ ] rename test package to be more generic for extends/super wildcard usage. maybe `typebound` instead of `wildcardextends`.

Already created the PR so that the fix is already reviewable. (Only planning on adding more tests not changing more code)

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/beanmapping/SourceReference.java
@@ -311,7 +311,7 @@ public class SourceReference extends AbstractReference {
             Type newType = type;
             for ( int i = 0; i < entryNames.length; i++ ) {
                 boolean matchFound = false;
-                Type noBoundsType = newType.withoutBounds();
+                Type noBoundsType = newType.hasTypeBound() ? newType.getTypeBound() : newType;
                 ReadAccessor readAccessor = noBoundsType.getReadAccessor( entryNames[i] );
                 if ( readAccessor != null ) {
                     PresenceCheckAccessor presenceChecker = noBoundsType.getPresenceChecker( entryNames[i] );
@@ -432,6 +432,7 @@ public class SourceReference extends AbstractReference {
         PropertyEntry deepestProperty = getDeepestProperty();
         if ( deepestProperty != null ) {
             Type type = deepestProperty.getType();
+            type = type.hasTypeBound() ? type.getTypeBound() : type;
             Map<String, ReadAccessor> newDeepestReadAccessors = type.getPropertyReadAccessors();
             String parameterName = getParameter().getName();
             String deepestPropertyFullName = deepestProperty.getFullName();
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -1264,6 +1264,10 @@ public class Type extends ModelElement implements Comparable<Type> {
         return boundingBase;
     }
 
+    public boolean hasTypeBound() {
+        return getTypeBound() != null;
+    }
+
     public boolean hasAccessibleConstructor() {
         if ( hasAccessibleConstructor == null ) {
             hasAccessibleConstructor = false;
--- a/processor/src/test/java/org/mapstruct/ap/test/generics/wildcardextends/Source.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/generics/wildcardextends/Source.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.generics.wildcardextends;
+
+/**
+ * @author Ben Zegveld
+ */
+public class Source<T extends WildcardedInterface> {
+    private T object;
+
+    public T getObject() {
+        return object;
+    }
+
+    public void setObject(T object) {
+        this.object = object;
+    }
+}
--- a/processor/src/test/java/org/mapstruct/ap/test/generics/wildcardextends/Target.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/generics/wildcardextends/Target.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.generics.wildcardextends;
+
+/**
+ * @author Ben Zegveld
+ */
+public class Target {
+    String object;
+
+    public String getObject() {
+        return object;
+    }
+
+    public void setObject(String object) {
+        this.object = object;
+    }
+}
--- a/processor/src/test/java/org/mapstruct/ap/test/generics/wildcardextends/WildcardExtendsMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/generics/wildcardextends/WildcardExtendsMapper.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.generics.wildcardextends;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+/**
+ * @author Ben Zegveld
+ */
+@Mapper
+public abstract class WildcardExtendsMapper {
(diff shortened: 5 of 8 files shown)


filiphr:
I also had a look at this one when I was analysing it and it was tricky, because on the first step like you've done now it is being fixed. However, I think that if you do `@Mapping( target = "object", source = "object.contents" )` it will no longer work, or something similar to that. I think once you do without bounds for `<?>` you get `Object` type for `?`. However, it is worth exploring ways to fix this properly.

filiphr:
@Zegveld I really dislike the fix in this PR. It is way to intrusive and changes a lot of things that shouldn't really be affected. I'll give it a go and try to see if I can come up with another fix for this one.

Zegveld:
The one bit I don't like myself about the current change is the "?" comparison bit at the `Filters` class.
Though the changes at `Type` feel logical to me.

As a result of the changes an parameter at `TypeFactory.getReturnType(...)` became obsolete which resulted in the changes there.

filiphr:
>Though the changes at `Type` feel logical to me.

All that looks way too error prone. We should create a correct accessor with a correct type and not pass this things down to the `Type` to resolve it. When `@Mapping` is not used everything works fine, so we need to only fix it in the creation of the `SourceReference`. Perhaps we need to tackle #2070 once and for all :)

Zegveld:
Seeing as tackling #2070 would indeed mean that the code there would change again, I've done some more rework and moved all the required actions onto the `Filters` class. At the same time I've refactored the "?" comparison away. Still not 100% happy with the new result but it is improving.

sjaakd:
I guess I need to dig in to this material as well. I'll try to have a look sometime this week..

Zegveld:
I believe the current state is the best I can do atm. if you have any idea's for improvements I'm open to them.

processor/src/main/java/org/mapstruct/ap/internal/util/Filters.java
```diff
@@ -111,7 +114,17 @@ public Map<String, ReadAccessor> recordAccessorsIn(Collection<Element> recordCom
     }
 
     private TypeMirror getReturnType(ExecutableElement executableElement) {
-        return getWithinContext( executableElement ).getReturnType();
+        TypeMirror returnType = getWithinContext( executableElement ).getReturnType();
+        if ( returnType.getKind() == TypeKind.WILDCARD ) {
```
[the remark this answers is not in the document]
sjaakd: hmm. can return type be wildcards (that have an extends bound)?

sjaakd (commented):
Its a while ago.. but for me this: http://www.angelikalanger.com/GenericsFAQ/JavaGenericsFAQ.html was really useful. I'll have a better look at the case soon.

filiphr:
I really do not feel comfortable removing one of the methods from the `TypeFactory`. Imagine if we have something like:

```java
public interface OtherWildCardInterface extends WildcardInterface {
}
```

```java
@Mapper
public interface WildcardNestedInheritedExtendsMapper {
    WildcardNestedInheritedExtendsMapper INSTANCE = Mappers.getMapper( WildcardNestedInheritedExtendsMapper.class );

    @Mapping( target = "object", source = "contained.value")
    Target map(SourceContainerInherited<? extends OtherWildCardInterface> source);

    default String mapToString(WildcardInterface rawData) {
        return rawData.getContents();
    }

    default String mapToString(OtherWildCardInterface rawData) {
        return rawData.getContents() + "other";
    }
}
```

Or perhaps another raw types that need to be returned in the scope of the currently declared type. There might be some other edge cases where we resolve these things.

---

In addition to that in the `Filters` we have

```java
        if ( returnType.getKind() == TypeKind.WILDCARD ) {
            if ( ( (WildcardType) returnType ).getExtendsBound() != null ) {
                return ( (WildcardType) returnType ).getExtendsBound();
            }
            return ( (TypeVariable) executableElement.getReturnType() ).getUpperBound();
        }
```

I am confused by that. We check the `returnType` and if that does not have an extends bound we assume that we can cast the executable element to `TypeVariable`. 

---

I'll need some more time to process this changes and try to wrap my head around them

Zegveld:
with `SourceContainerInherited<? extends OtherWildCardInterface>` then the type returned is `OtherWildCardInterface`, because the returnType of the container internal method is a `WILDCARD` with an `extendsBound`, so that `extendsBound` is chosen.

I'll move the `Issue2677Mapper` and corresponding test to the location where the new tests are as well. Then this case is also visible in the code-review.

----

concerning the Filters bit `returnType.getKind() == TypeKind.WILDCARD` can only be true when `executable.getReturnType()` is a `TypeVariable`. 

It however is not true that because `executable.getReturnType()` is a `TypeVariable` then the `returnType.getKind()` must be a wildcard. There are other types that also fit the `TypeVariable` grouping, like `TypeKind.TYPEVAR`.

----

Think I'll also add some inline comments as to how and why the code works like this, and which situation is covered by what line.

processor/src/main/java/org/mapstruct/ap/internal/util/Filters.java
```diff
@@ -111,7 +114,24 @@ public Map<String, ReadAccessor> recordAccessorsIn(Collection<Element> recordCom
     }
 
     private TypeMirror getReturnType(ExecutableElement executableElement) {
-        return getWithinContext( executableElement ).getReturnType();
+        TypeMirror returnType = getWithinContext( executableElement ).getReturnType();
```
sjaakd, line 117: rename `getWithinContext` to `getMethodAsMemberOfDefiningClass`. guess thats a more appropriate name

processor/src/main/java/org/mapstruct/ap/internal/util/Filters.java
```diff
@@ -111,7 +114,24 @@ public Map<String, ReadAccessor> recordAccessorsIn(Collection<Element> recordCom
...
+        if ( returnType.getKind() == TypeKind.WILDCARD ) {
+            // we got 2 versions of wildcards: '?' or '? extends ...'
+            if ( ( (WildcardType) returnType ).getExtendsBound() != null ) {
+                // in case of '? extends ...' we want to return the '...' part.
+                return ( (WildcardType) returnType ).getExtendsBound();
+            }
+            // in case of just '?', we want to look at the original methods return type and see if that one has an
```
sjaakd, line 124: I guess this can only be if the method is generic. Because the `getWithinContext` would already have tackled the class defined type-var's right?

processor/src/test/java/org/mapstruct/ap/test/generics/typebounds/WildcardedInterface.java
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.generics.typebounds;
+
+/**
+ * @author Ben Zegveld
+ */
+public interface WildcardedInterface {
```
sjaakd, line 11: I understand that this is the base for a wildcard.. But the interface itself does not contain wildcards. Hence the name is somewhat confusing. Would `UpperBound` be a [better name](http://www.angelikalanger.com/GenericsFAQ/FAQSections/TypeArguments.html#What%20is%20a%20bounded%20wildcard?)?

processor/src/test/java/org/mapstruct/ap/test/generics/typebounds/WildcardedInterfaceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.generics.typebounds;
+
+/**
+ * @author Ben Zegveld
+ */
+class WildcardedInterfaceImpl implements WildcardedInterface {
```
sjaakd, line 11: same here: `UpperBoundImpl`?

processor/src/test/java/org/mapstruct/ap/test/generics/typebounds/WildcardedInterface.java
sjaakd: Or `WildCardBase` or just `WildCard`

sjaakd (commented):
LGTM.. some naming remarks..

processor/src/main/java/org/mapstruct/ap/internal/util/Filters.java
Zegveld: That is correct.

processor/src/test/java/org/mapstruct/ap/test/generics/typebounds/WildcardedInterface.java
Zegveld: went with `WildCard` and `WildCardImpl`

processor/src/test/java/org/mapstruct/ap/test/generics/typebounds/WildcardedInterfaceImpl.java
Zegveld: went with `WildCard` and `WildCardImpl`

PR #2996, opened 2022-08-29
feat: support copy @Deprecated

### LINKED ISSUE #2773: Copy `@Deprecated` annotation from method or mapper to implementation

chenzijia12300:
Fixes #2773


processor/src/main/java/org/mapstruct/ap/internal/model/AdditionalAnnotationsBuilder.java
```diff
@@ -90,6 +92,33 @@ protected void addInstances(AnnotateWithsGem gem, Element source, Set<Annotation
...
+    }
+
+    private Set<Annotation> addDeprecatedAnnotation(Element source, Set<Annotation> annotations) {
+        Type deprecatedType = typeFactory.getType( Deprecated.class );
+        Optional<? extends AnnotationMirror> optionalAnnotationMirror = source.getAnnotationMirrors().stream()
+                .filter( annotationMirror -> DEPRECATED_ANNOTATION_FQN.equals(
+                        annotationMirror.getAnnotationType().asElement().toString() ) ).findFirst();
+        if (optionalAnnotationMirror.isPresent() &&
+                annotations.stream().anyMatch( annotation -> annotation.getType().equals( deprecatedType ) ) ) {
+            messager.printMessage(
```
filiphr, line 108: I think that for this particular case we don't have to add an error. We can have a warning that is saying that the method is already using `@AnnotateWith(Deprecated.class)` and that doing it twice is obsolete.

filiphr (changes requested):
Thanks a  lot for your patience @chenzijia12300. This PR now looks really clean 😄. I've left one comment about the error, which I think that we should not be that strict.

In addition to the comment, can you please revert the changes you've done with the imports in the `AdditionalAnnotationsBuilder` and try to keep the same order as it was before. We do have some IDE specific settings in https://github.com/mapstruct/mapstruct/blob/main/etc that you can use to have the right formatting and import order. If you are using IntelliJ then you can import https://github.com/mapstruct/mapstruct/blob/main/etc/mapstruct.xml as a code style.

Zegveld:
I'm missing the 2 annotation elements of `@Deprecated`: `since` and `forRemoval`.

Using a `Gem` for the annotation will allow you to find the values of these.
You will need to add `Deprecated` to the `GemGenerator` class so that it becomes available for usage.
Another advantage of using `Gem` is that you do not need to search for the deprecation annotation yourself but the generated `Gem` has methods that do this for you.

processor/src/main/java/org/mapstruct/ap/internal/model/AdditionalAnnotationsBuilder.java
Zegveld: @filiphr , should the warning take into account the situation where the `@Deprecation` and the `@AnnotateWith` don't match?

for example:
```java
@Deprecated( since = "1.5" )
@AnnotateWith(
    value = Deprecated.class,
    elements = {
        @Element( name = "forRemoval", boolean = true ),
        @Element( name = "since", string = "1.5.0" )
    }
)
```

chenzijia12300:
> 

Thank you for your guidance😊

processor/src/main/java/org/mapstruct/ap/internal/model/AdditionalAnnotationsBuilder.java
chenzijia12300: Okay, so when there is a duplicate definition, a warning is raised and the type defined in @AnnotateWith(Deprecated.class) is returned

chenzijia12300: Okay, so when there is a duplicate definition, a warning is raised and the type defined in @AnnotateWith(Deprecated.class) is returned

Zegveld:
Just looked at the reason the JDK 8 tests failed.
I totally missed this, but the `since` and `forRemoval` annotation elements are only available since java 9.
So some of the tests must be annotated that they require at least java 11, while the others also work with java 8.
`@EnabledForJreRange(min = JRE.JAVA_11)`
This might also require some tests to be near duplicates, one without the annotation elements for java 8+ and one with for java 11+.

chenzijia12300:
> ,
Thanks for the suggestion, I'm having a problem with how to exclude certain classes from compiling in a particular JDK version. I think @EnabledForJreRange only prevents the execution of the test class or test method, it does not prevent the compilation of the class.
I see a FullFeatureCompilationExclusionCliEnhancer class, please can be here for the specific class compilation exclusion?

> Just looked at the reason the JDK 8 tests failed. I totally missed this, but the `since` and `forRemoval` annotation elements are only available since java 9. So some of the tests must be annotated that they require at least java 11, while the others also work with java 8. `@EnabledForJreRange(min = JRE.JAVA_11)` This might also require some tests to be near duplicates, one without the annotation elements for java 8+ and one with for java 11+.

Thanks for the suggestion, I'm having a problem with how to exclude certain classes from compiling in a particular `JDK` version. I think `@EnabledForJreRange` only prevents the execution of the test class or test method, it does not prevent the compilation of the class.
I see a `FullFeatureCompilationExclusionCliEnhancer` class, please can be here for the specific class compilation exclusion?

filiphr:
>I think `@EnabledForJreRange` only prevents the execution of the test class or test method, it does not prevent the compilation of the class.

Indeed that prevents the execution of a specific test class or test method. The processors tests are not run with Java 8, they are only run with Java 11 and later.

>I see a `FullFeatureCompilationExclusionCliEnhancer` class, please can be here for the specific class compilation exclusion?

Spot on @chenzijia12300. That class is the one that can add additional excludes for the full feature integration test. Keep in mind that you most likely will also need to add an `additionalExclude` in 
https://github.com/mapstruct/mapstruct/blob/68571de01bd9b06cef647a3d9ad073f0f0cd2456/integrationtest/src/test/resources/fullFeatureTest/pom.xml#L22-L28

and 

https://github.com/mapstruct/mapstruct/blob/68571de01bd9b06cef647a3d9ad073f0f0cd2456/integrationtest/src/test/resources/fullFeatureTest/pom.xml#L33-L50

chenzijia12300:
> Spot on @chenzijia12300. That class is the one that can add additional excludes for the full feature integration test. Keep in mind that you most likely will also need to add an `additionalExclude` in

Thank you for the tip 😊

processor/src/test/java/org/mapstruct/ap/test/annotatewith/deprecated/newer/DeprecatedTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.mapstruct.ap.testutil.compilation.annotation.Diagnostic;
+import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutcome;
+import org.mapstruct.factory.Mappers;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author orange add
+ */
+public class DeprecatedTest {
```
filiphr, line 28: I see that there is some small misunderstanding of the use of the integration `fullFeatureTest`.

The full feature integration test does nothing more than compiling all the mappers that are in our processor test module. However, it doesn't run any tests on its own. 

In addition to that, our processor tests are currently only running with Java 11 and later. Which means that we can easily move this test to the package above `org.mapstruct.ap.test.annotatewith.deprecated` and remove the one from `org.mapstruct.ap.test.annotatewith.deprecated.older`. I would also remove the `@EnabledForJreRange` from this tests, since it is not needed.

Can we also please change the names of the mappers a bit. e.g. everything from the `older` package can just be moved to the package above `org.mapstruct.ap.test.annotatewith.deprecated` and the prefix `Older` can be removed. Everything in the `newer` package can be moved to a different package called `java11`, remove the prefix `Newer` and do things like `DeprecatedWithAttributesMapper`, etc. 

I am asking for this in order to reduce some of the additional test complexity and to be immediately clear when someone reads the names of the classes.

processor/src/main/java/org/mapstruct/ap/internal/model/AdditionalAnnotationsBuilder.java
```diff
@@ -90,6 +92,44 @@ protected void addInstances(AnnotateWithsGem gem, Element source, Set<Annotation
...
+        if ( annotations.stream().anyMatch( annotation -> annotation.getType().equals( deprecatedType ) ) ) {
+            messager.printMessage(
+                    source,
+                    deprecatedGem.mirror(),
+                    Message.ANNOTATE_WITH_DUPLICATE,
+                    deprecatedType.describe() );
+            return annotations;
+        }
+        List<AnnotationElement> annotationElements = new ArrayList<>();
+        for ( Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry : deprecatedGem.mirror()
```
filiphr, line 116: We don't really need to use the mirror and strings for this. The `DeprecatedGem` should have methods that allow to have access to the values. You can rewrite the loop to:

```java
if ( deprecatedGem.since().hasValue() ) {
    annotationElements.add( new AnnotationElement(
        AnnotationElementType.STRING,
        "since",
        Collections.singletonList( deprecatedGem.since().getValue() )
    ) );
}
        
if ( deprecatedGem.forRemoval().hasValue() ) {
    annotationElements.add( new AnnotationElement(
        AnnotationElementType.BOOLEAN,
        "forRemoval",
        Collections.singletonList( deprecatedGem.forRemoval().getValue() )
    ) );
}
```

filiphr (changes requested):
Thanks for the changes @chenzijia12300. Few more tiny changes and we are ready to merge this.

processor/src/test/java/org/mapstruct/ap/test/annotatewith/deprecated/newer/DeprecatedTest.java
chenzijia12300: Thank you for your patient guidance, I will correct it

filiphr (approved):
LGTM thanks for all the changes @chenzijia12300. @Zegveld do you have anything extra to add or can I merge this?

Issue #2963, opened 2022-08-19
Can provide converter between enum and int 


Ikki-Dai:
provide enum copy to int and int copy to enum 
such as Mybatis EnumOrdinalTypeHandler

```
int  = enums.ordinal()

enums.values()[int]
```

filiphr:
This is a good enhancement @Ikki-Dai. Thanks for raising it.

Would you be interested in providing this functionality through a PR? This issue is a good one for anyone that is interested in contributing to MapStruct. Therefore, if @Ikki-Dai is not interested in providing a PR, please reach out to us in this issue if you would like to work on this.

jccampanero:
If  @Ikki-Dai is not interested in providing this functionality through a PR himself, I would be glad to help you with this issue.

filiphr:
Thanks for the offer @jccampanero. I am going to wait until Friday for @Ikki-Dai  in case they are interested. Otherwise, I'll assign the issue to you

Ikki-Dai:
Thanks @jccampanero and @filiphr, I have no time to provide PR these days. and @jccampanero,  pls do it.

I'm glad to provide some advice and review, if i have spare time.

filiphr:
Thanks for letting use know @Ikki-Dai. @jccampanero I've assigned this issue to you let us know if you need some pointers

jccampanero:
Sorry for the late reply.

Thank you very much @Ikki-Dai, and thank you very much you too @filiphr.

I will try to do my best to implement the required functionality. I will contact you if I need any help.

Thank you both very much again.

liuxiansheng990710:
@filiphr On the basis of mapstruct, I realized the conversion of enum and int. May I have a try?

filiphr:
@liuxiansheng990710 this issue is already assigned to @jccampanero. If you would like to contribute, feel free to have a look at the issue tracker for some other issue.

jccampanero:
Thank you very much @filiphr and sorry for the late reply, I was unable to work on the task until yesterday.

I pushed an initial version of the code.

I defined the unit tests in a new package named `enumtointeger` under `org.mapstruct.ap.test.conversion`. Is it the right place? I noticed some tests related to `enum` to `String` in the `org.mapstruct.ap.test.value.enum2string` package.

Please let me know if I need to make any changes, I'll be happy to do so.

Thank you very much.

filiphr:
[...]

PR #3000, opened 2022-08-30
Add support for enum to integer conversion

jccampanero:
Adds support for enum to integer conversion, according to `enum.ordinal()`. See #2963 for details.

@filiphr, for now I have left the work split into the commits of the original branch, so that code review can be done more easily and to make it easier to add additional changes if needed. I will squash the commits into one after the code review if you consider it appropriate.

Thank you very much.


processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
```diff
@@ -324,10 +330,14 @@ private void register(String sourceTypeName, Class<?> targetClass, ConversionPro
     }
 
     public ConversionProvider getConversion(Type sourceType, Type targetType) {
-        if ( sourceType.isEnumType() && targetType.equals( stringType ) ) {
+        if ( sourceType.isEnumType() &&
+                ( targetType.equals( stringType ) || targetType.equals( integerType ) || targetType.equals( intType ) )
```
filiphr, line 334: You don't need the `intType` here you can do `targetType.getBoxedEquivalent().equals( integerType )` instead

processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
```diff
@@ -324,10 +330,14 @@ private void register(String sourceTypeName, Class<?> targetClass, ConversionPro
...
+        ) {
             sourceType = enumType;
         }
-        else if ( targetType.isEnumType() && sourceType.equals( stringType ) ) {
+        else if ( targetType.isEnumType() &&
+                ( sourceType.equals( stringType ) || sourceType.equals( integerType ) || sourceType.equals( intType ) )
```
filiphr, line 339: You don't need the `intType` here you can do `sourceType.getBoxedEquivalent().equals( integerType )` instead

filiphr (changes requested):
Looks great @jccampanero. There are only 2 tiny things that we should do before we merge this.

processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
jccampanero: Thank you very much for the feedback @filiphr. I implemented the requested change.

Please, let me know if any additional work needs to be done.

processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
jccampanero: Thank you very much again for the feedback @filiphr, I implemented the requested change.

Please do not hesitate to contact me to make any changes you deem appropriate.

processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
```diff
@@ -324,10 +328,18 @@ private void register(String sourceTypeName, Class<?> targetClass, ConversionPro
     }
 
     public ConversionProvider getConversion(Type sourceType, Type targetType) {
-        if ( sourceType.isEnumType() && targetType.equals( stringType ) ) {
+        if ( sourceType.isEnumType() &&
+                ( targetType.equals( stringType ) ||
+                  targetType.equals( integerType ) ||
+                  targetType.getBoxedEquivalent().equals( integerType ) )
```
filiphr, line 334: When doing `getBoxedEquivalent` if the target type is not primitive it returns itself. This means that we can just do:

```suggestion
                  targetType.getBoxedEquivalent().equals( integerType ) )
```

processor/src/main/java/org/mapstruct/ap/internal/conversion/Conversions.java
```diff
@@ -324,10 +328,18 @@ private void register(String sourceTypeName, Class<?> targetClass, ConversionPro
...
+        ) {
             sourceType = enumType;
         }
-        else if ( targetType.isEnumType() && sourceType.equals( stringType ) ) {
+        else if ( targetType.isEnumType() &&
+                ( sourceType.equals( stringType ) ||
+                  sourceType.equals( integerType ) ||
+                  sourceType.getBoxedEquivalent().equals( integerType ) )
```
filiphr, line 341: When doing `getBoxedEquivalent` if the source type is not primitive it returns itself. This means that we can just do:

```suggestion
                  sourceType.getBoxedEquivalent().equals( integerType ) )
```

filiphr (changes requested):
Thanks for the changes @jccampanero. One last request from my side and we are good to go to merge this.

jccampanero:
Thank you very much @filiphr and sorry, I missed that point in my previous commit.

Please, let me know if you need anything else, I will gladly do it.

filiphr (approved):

filiphr:
Thanks a lot for your contribution @jccampanero. I've merged this to main

jccampanero:
You are welcome @filiphr; on the contrary, thank you for this great library.

PR #3003, opened 2022-09-01
Avoid unnecessary unboxing of Boolean

prasanth08:
Changed Boolean.valueOf() to Boolean.parseBoolean() - Boolean.valueOf() returns Boolean while the attributes in Options Class is boolean, Boolean.parseBoolean() returns boolean.
Changed Collections.sort() to List.sort() - Collections.sort() internally calls the List.sort().
Fix Java - doc


processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
```diff
@@ -63,7 +62,7 @@
  * <ul>
  * <li>retrieve mapping methods</li>
  * <li>create the {@code Mapper} model</li>
- * <li>perform enrichments and modifications (e.g. add annotations for dependency injection)</li>
+ * <li>perform enrichment's and modifications (e.g. add annotations for dependency injection)</li>
```
filiphr, line 65: Why do you think that we need to change this? Right now it is a plural noun of enrichment. With the change it becomes something else.

filiphr (changes requested):
Your changes are OK @prasanth08, only one small comment about one of the Javadocs. 

Apart from that, can you please rebase this on top of main. I see that there are some conflicts

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
prasanth08: I agree @filiphr  .Looks like the apostrophe came in from intellij and sonar plugin just before pushing the change.

filiphr:
Thanks @prasanth08

PR #3005, opened 2022-09-02
#2743 BeanMappingOptions should not be inherited for forged methods

### LINKED ISSUE #2743: Ignored unknown source property error, but property exist.
From my perspective, I found a bug, because of the inconsistent behaving of ```ignoreUnmappedSourceProperties```.

I defined these classes:
```
data class Inside(val value : String)
data class InsideDto(val value : String)
data class Problem(val problem : String)
data class ProblemDto(val problem : String)
data class Outside(val inside : Inside, val problem : Problem)
data class OutsideDto(val insideDto : InsideDto, val problem : ProblemDto)
```

Inside of a Mapper which has this Config:
```
@MapperConfig(
    componentModel = "spring",
    unmappedTargetPolicy = ReportingPolicy.ERROR
)
```

I declared:
```
@Mapper(config = MapstructConfig::class)
interface Converter {
    @BeanMapping(ignoreUnmappedSourceProperties = ["inside"])
    fun test(
        outside: Outside,
        insideDto: InsideDto
    ): OutsideDto
}
```

this throws the following error:
```error: Ignored unknown source property: "inside". public abstract [placeholder].OutsideDto test(@org.jetbrains.annotations.NotNull```

I found a workaround. I just have to add an extra mapper declaration inside of my Converter for the Problem mapping:
```
fun solution(problem: Problem) : ProblemDto
```

In this case, everything works. For me, this looks like a bug, because it is not consistent behaviour.

### PR DESCRIPTION
**Pull request by filiphr**
Fixes #2743

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_2743/Issue2743Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_2743/Issue2743Mapper.java
@@ -12,9 +12,7 @@ import org.mapstruct.ReportingPolicy;
 /**
  * @author Filip Hrisafov
  */
-@Mapper(
-    unmappedSourcePolicy = ReportingPolicy.ERROR
-)
+@Mapper(unmappedSourcePolicy = ReportingPolicy.ERROR)
 public interface Issue2743Mapper {
 
     @BeanMapping(ignoreUnmappedSourceProperties = { "number" })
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: processor/src/test/java/org/mapstruct/ap/test/bugs/_2743/Issue2743Mapper.java [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+import org.mapstruct.BeanMapping;
+import org.mapstruct.Mapper;
+import org.mapstruct.ReportingPolicy;
+
+/**
+ * @author Filip Hrisafov
+ */
+@Mapper(
+    unmappedSourcePolicy = ReportingPolicy.ERROR,
+    unmappedTargetPolicy = ReportingPolicy.ERROR
```
Zegveld, line 17: only setting `unmappedSourcePolicy` to `Error` should suffice for the issue, shouldn't it?

filiphr: You are right. I've removed it

PR #3007, opened 2022-09-03
Feature/2663

### LINKED ISSUE #2663: 2-step mapping with generics does not work correctly.
Consider the following situation: (removed getter/setters to keep it short)
```java
class Request {
    Nullable<ChildRequest> child;
}

class ChildRequest {
    String name;
}

class RequestDto {
    JsonNullable<ChildRequestDto> child;
}

class ChildRequestDto {
    String name;
}
```

with the following mapper:
```java
@Mapper
public interface RequestMapper {

    RequestMapper INSTANCE = Mappers.getMapper( RequestMapper.class );

    Request map(RequestDto dto);

    default <T> Nullable<T> jsonNullableToNullable(JsonNullable<T> jsonNullable) {
        if ( jsonNullable.isPresent() ) {
            return Nullable.of( jsonNullable.get() );
        }
        return Nullable.undefined();
    }

    default JsonNullable<ChildRequest> mapJsonNullableChildren(JsonNullable<ChildRequestDto> dtos) {
        if (dtos.isPresent()) {
            return JsonNullable.of( mapChild( dtos.get() ) );
        } else {
            return JsonNullable.undefined();
        }
    }

    ChildRequest mapChild(ChildRequestDto dto);
}
```

This will not compile, but throw the following errors:
```
ERROR Issue2377Mapper.java:19 Nullable<ChildRequest> does not have an accessible constructor.
ERROR Issue2377Mapper.java:19 Can't map property "JsonNullable<ChildRequestDto> child" to "Nullable<ChildRequest> child". Consider to declare/implement a mapping method: "Nullable<ChildRequest> map(JsonNullable<ChildRequestDto> value)".
```

instead of generating the following code:
```
        request.setChild( jsonNullableToNullable( mapJsonNullableChildren( dto.getChild() ) ) );
```

Discovered during analysis of #2377 .

Zegveld:
Was doing a lot of generics stuff lately, so figured I'd also give this one a try.
Debugging I noticed that it did the second match check with the `sourceType` that had the unconverted generic `T`.
Figured I'd give it a shot at replacing this generic to get successful matching going.

closes #2663


processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -1359,6 +1376,44 @@ public ResolvedPair resolveParameterToType(Type declared, Type parameterized) {
         return new ResolvedPair( this, this );
     }
 
+    /**
+     * Resolves generic types using the declared and parameterized types as input.
```
sjaakd, line 1380: would be nice to mention as example an even farther nested structure.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -584,6 +585,22 @@ public Type withoutBounds() {
         );
     }
 
+    private Type replaceGeneric(Type oldGenericType, Type newType) {
+        if ( !typeParameters.contains( oldGenericType ) || newType == null ) {
+            return this;
+        }
+        newType = newType.getBoxedEquivalent();
```
sjaakd, line 592: is this needed? I guess you can't use primitive types as generic?

sjaakd (commented):
only one question remaining.. LGTM solution is far more simple than I thought. Nice work @Zegveld 👍

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
Zegveld: added the following example:
```
     * this: List<Optional<T>>
     * declaredType: JAXBElement<T>
     * parameterizedType: JAXBElement<BigDecimal>
     * result: List<Optional<BigDecimal>>
```
hope this is what you meant with nested.

Also figured I'd add some multi-value examples:
```
     * this: Map<K, V>
     * declaredType: JAXBElement<K>
     * parameterizedType: JAXBElement<BigDecimal>
     * result: Map<BigDecimal, V>
     *
     * this: Map<KEY, VALUE>
     * declaredType: HashMap<VALUE, KEY>
     * parameterizedType: HashMap<BigDecimal, String>
     * result: Map<String, BigDecimal>
```

The javadoc preview didn't look that nice so fixed that as well.

sjaakd (approved):

filiphr:
I haven't reviewed the PR fully yet. However, I tried it for the example in #3163 and it isn't working as expected. We should investigate whether this should also fix that.

Zegveld:
> I haven't reviewed the PR fully yet. However, I tried it for the example in #3163 and it isn't working as expected. We should investigate whether this should also fix that.

I'll take a look at it.

Zegveld:
It seems to be a different problem. I've added the fix here, but perhaps it should be a separate PR.

filiphr:
>It seems to be a different problem. I've added the fix here, but perhaps it should be a separate PR.

I see. Indeed it does make sense to put the fix in a separate PR. I initially thought that it is the same root cause (I didn't really try it out)

PR #3010, opened 2022-09-04
Feature/2955

### LINKED ISSUE #2955: Method annotate with AfterMapping is not called
Method annotate with `@AfterMapping` is no more called in 1.5.2.Final version when using a method with 2 arguments.
It works in 1.4.2.Final and it works when using a method with one argument and an object returned.
Example:
```
void updateDomain(MyDto dto, @MappingTarget MyDomain domain);

@AfterMapping
public default MyDomain doAfterMapping(@MappingTarget MyDomain domain, MyDto dto) {
    return domain;
}
```

Zegveld:
When collecting lifecycle callback methods also collect the methods with returntype equal to the lifecycle callback result type.
Currently this is only done when no other callback methods are found, perhaps this should always be done.

Fixes #2955


processor/src/main/java/org/mapstruct/ap/internal/model/LifecycleMethodResolver.java
```diff
@@ -144,6 +144,16 @@ private static List<LifecycleCallbackMethodReference> collectLifecycleCallbackMe
             method.getReturnType(),
             SelectionCriteria.forLifecycleMethods( selectionParameters ) );
 
+        if ( matchingMethods.isEmpty() ) {
```
sjaakd, line 147: Instead of redoing this with the `getResultType`. Can't we do it with the `getResultType`  from the start. I guess that's always populated, right?

```java
    @Override
    public Type getResultType() {
        return mappingTargetParameter != null ? mappingTargetParameter.getType() : returnType;
    }

```

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
```diff
@@ -197,11 +201,13 @@ else if ( matchingParameterAssignmentVariants.size() == 1 ) {
             // we had some matching assignments before, but when checking the parameter names we can't find an
             // appropriate one, in this case the user must chose identical parameter names for the mapping and lifecycle
             // method
-            messager.printMessage(
-                selectedMethod.getExecutable(),
-                Message.LIFECYCLEMETHOD_AMBIGUOUS_PARAMETERS,
-                mappingMethod
-            );
+            if ( !warningAlreadyGiven( mappingMethod, selectedMethod ) ) {
```
sjaakd, line 204: We really should look at redesigning the (error)messages and the way they are handled in mapping-method-trees with nested mappings.  This becomes too tricky to keep track off. Feels to much of localized fix.

sjaakd, line 204: under which condition does this occur?  When I read the comments my guess is it has some conflicting method signatures with the same parameters and it cannot select the right one based on names.

So for instance:

`A methodX(B one, C two);`
`A methodX(C three, B four)`

And the calling method does not have similar parameter names.


Could it be that the error is than given for both methods? I guess that's correct. Right?

processor/src/main/java/org/mapstruct/ap/internal/model/LifecycleMethodResolver.java
Zegveld: Will attempt this, this should also fix that what I did at the other remark is not necessary.

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/TypeSelector.java
Zegveld: it runs twice through the code, which is why the warning was suddenly given twice. fixing the double run through the code would also reduce this warning to only once.

Zegveld: As to when the warning occurs:

When you have as mapping method: 
`A methodX(String something, String somethingElse)`
and as matching method: 
`A methodY(String totallyDifferent, String alsoTotallyDifferent)`
As mapstruct cannot decide which 2 variables should match.

With mapping method:
`A methodX(B one, C two)`
and matching method:
`A methodX(B three, C four)`
this does not occur since the types are already unique enough to identify how to match.

Zegveld:
So basically this entire PR is changing 1 method call and a lot of tests to make certain it works as expected.

filiphr (changes requested):
The PR needs a bit of clean up before we can merge it.

I see that we have some tests which are testing that stack overflow errors are thrown. I do not believe that we should have such tests. It is enough to have a test which fails without the fixes and it is green after we fix it.

I also see that we are modifying some existing tests such as the `NumberMapperContext` and removing certain things from the `NumberMapperWithContext`.  I think that we should not remove nor update existing things and only add new fresh tests. If something is failing with the new changes then the changes are most likely not right.

Zegveld:
> The PR needs a bit of clean up before we can merge it.
> 
> I see that we have some tests which are testing that stack overflow errors are thrown. I do not believe that we should have such tests. It is enough to have a test which fails without the fixes and it is green after we fix it.

Looking at the tests again, the tests in the cycles package can be removed.

> I also see that we are modifying some existing tests such as the `NumberMapperContext` and removing certain things from the `NumberMapperWithContext`. I think that we should not remove nor update existing things and only add new fresh tests. If something is failing with the new changes then the changes are most likely not right.

The test added here is actually the correct one. Did some clean-up so that this becomes better visible.

filiphr:
It looks much cleaner now @Zegveld, thanks. 

I played around with this and I can actually recommend simplifying it a bit more.  Looking at the generated code we actually have 2 methods that will be called:

* `NumberMapperContext#getInstance`
* `NumberMapperContext#visitNumber`

If we change the test case to look something like:

```
    @ProcessorTest
    void numberUpdateMappingWithContextCallsCache() {
        Number target = new Number();
        NumberMapperWithContext.INSTANCE.integerToNumber( 2342, target );

        try {
            assertThat( NumberMapperContext.getVisited() ).containsExactly( target );
            NumberMapperContext.clearVisited();
            NumberMapperContext.putCache( target );
            NumberMapperWithContext.INSTANCE.integerToNumber( 25342, target );
            assertThat( NumberMapperContext.getVisited() ).isEmpty();
        }
        finally {
            NumberMapperContext.clearCache();
            NumberMapperContext.clearVisited();
        }
    }
```

WDYT @Zegveld?

Zegveld:
@filiphr that test fails, because the putCache object ends up in a different hash-bucket then the object used for retrieval during getCache.

I did change the test somewhat and removed the cache checking. Because I feel that is what you were aiming for.

filiphr:
I had a typo in it @Zegveld. The second call needs to be done with `2342`. That way the correct thing is done in the map

PR #3016, opened 2022-09-07
Make possible to have multiple accessors for overloaded methods

velo:
Proposed fix for #1135

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -63,6 +63,7 @@ import org.mapstruct.ap.internal.util.Message;
 import org.mapstruct.ap.internal.util.Strings;
 import org.mapstruct.ap.internal.util.accessor.Accessor;
 import org.mapstruct.ap.internal.util.accessor.AccessorType;
+import org.mapstruct.ap.internal.util.accessor.AccessorCandidates;
 import org.mapstruct.ap.internal.util.accessor.ParameterElementAccessor;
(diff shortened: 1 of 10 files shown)


Zegveld:
I haven't taken a good look at the production code yet, but the following was what I noticed about the test-code.

I feel that the test setup has too many components intertwined. What is it that is actually being tested?
- MapperConfig
- Mapping method with different source/target
- Overloaded setters
Or is this something that only fails when all three these components are combined?

If I'm correct this is mainly focussed on the `overloaded setters`. So it would be better to have that as the single component in the test setup.

Then concerning the test itself. Currently it just tests if it compiles. You do nothing to check the value inside the resulting object. This feels like it would be the same as having no code inside the test method. I believe it would be better to also test if the resulting value is as wanted. Perhaps also test the reverse mapping as it is present to see if you get something that is equal to the original object back.

Next it would be nice to have an unhappy-flow test, an overloaded situation where there is still no match for the types. The expectation here would be the compilation error that is normally thrown when you need to supply a mapping method yourself.

velo:
Hi @Zegveld thank you for your feedback.

Based on that, I simplified the original tests provided by alex.

Also added the unhappy path.

Please take a look and lemme know what you think

Cheers

velo:
build failed with
```
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/mapstruct/mapstruct/integrationtest/target/tmp/org.mapstruct.itest.tests.freeBuilderBuilderTest/javac/src/main/java/org/mapstruct/itest/freebuilder/PersonMapper.java:[17,12] Several possible source properties for target property "address".
```

I didn't touched that, in any case, I will merge latest master into my fork and see if helps

PR #3017, opened 2022-09-08
#3015 No need to pass annotations to the forge method

### LINKED ISSUE #3015: Annotations are automatically passed in forged methods
Currently the `@AnnotateWith` annotation modification on the original method is automatically passed to its `forgeMethod` method, should we change it to an optional one, regarding whether it is automatically passed to its derived method

e.g. currently for

```
@Mapper
public interface AnnotationForgeMapper {

    @AnnotateWith( CustomMethodOnlyAnnotation.class )
    List<Target> map(List<Source> sources);

}

```
will generate

```
public class AnnotationForgeMapperImpl implements AnnotationForgeMapper {

    @CustomMethodOnlyAnnotation
    @Override
    public List<Target> map(List<Source> sources) {
        if ( sources == null ) {
            return null;
        }

        List<Target> list = new ArrayList<Target>( sources.size() );
        for ( Source source : sources ) {
            list.add( sourceToTarget( source ) );
        }

        return list;
    }

    @CustomMethodOnlyAnnotation

    protected Target sourceToTarget(Source source) {
        if ( source == null ) {
            return null;
        }

        Target target = new Target();

        target.setValue( source.getValue() );

        return target;
    }
}
```
We can see that the `sourceToTarget` method also carries the `CustomMethodOnlyAnnotation` annotation

chenzijia12300:
fix #3015

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
@@ -118,8 +118,11 @@ public class ValueMappingMethod extends MappingMethod {
                 LifecycleMethodResolver.beforeMappingMethods( method, selectionParameters, ctx, existingVariables );
             List<LifecycleCallbackMethodReference> afterMappingMethods =
                 LifecycleMethodResolver.afterMappingMethods( method, selectionParameters, ctx, existingVariables );
-            List<Annotation> annotations = Collections.emptyList();
-            if ( !(method instanceof ForgedMethod) ) {
+            List<Annotation> annotations;
+            if ( method instanceof ForgedMethod ) {
+                annotations = Collections.emptyList();
+            }
+            else {
                 annotations = new ArrayList<>();
                 AdditionalAnnotationsBuilder additionalAnnotationsBuilder =
                     new AdditionalAnnotationsBuilder(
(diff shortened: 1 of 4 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
```diff
@@ -117,13 +118,17 @@ else if ( sourceType.isString() && targetType.isEnumType() ) {
                 LifecycleMethodResolver.beforeMappingMethods( method, selectionParameters, ctx, existingVariables );
             List<LifecycleCallbackMethodReference> afterMappingMethods =
                 LifecycleMethodResolver.afterMappingMethods( method, selectionParameters, ctx, existingVariables );
-            AdditionalAnnotationsBuilder additionalAnnotationsBuilder =
+            List<Annotation> annotations = Collections.emptyList();
```
sjaakd, line 121: you initialize the list of annotations twice.. Why not write:
```java
List<Annotations> annotations;
if ( method instanceof ForgedMethod ) {
   annotations = Collections.emptyList();
}
else {
  annotations = new ArrayList();
  // ...
}
```

sjaakd (changes requested):

processor/src/main/java/org/mapstruct/ap/internal/model/ValueMappingMethod.java
chenzijia12300: @sjaakd Thank you for the review, I just don't like writing else, I will try to modify it :)

filiphr:
Thanks @chenzijia12300

PR #3027, opened 2022-09-21
#3018: subclassmapping now correctly applies mappingcontrol configuration.

### LINKED ISSUE #3018: SubclassMapping doesn't generate mapping methods
- [ x] Is this an issue (and hence not a question)? 

Sorry for [crossposting](https://stackoverflow.com/questions/73648381/deepclone-of-same-class-with-subclassmapping) but I just want to reference this issue somewhere in code.

So is it really necessary to include mapping methods for all mentioned subclasses even when we have them in `SubclassMapping` annotation? I initially thought that this is the issue of `DeepClone` but it seems like it's problem of `SubclassMapping` and probably the same type in input/output (source/target) of the method.

Zegveld:
fixes #3018

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/subclassmapping/DeepCloneMethodMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/subclassmapping/DeepCloneMethodMapper.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.subclassmapping;
+
+import org.mapstruct.BeanMapping;
+import org.mapstruct.Mapper;
+import org.mapstruct.NullValueMappingStrategy;
+import org.mapstruct.SubclassMapping;
+import org.mapstruct.ap.test.subclassmapping.mappables.Bike;
+import org.mapstruct.ap.test.subclassmapping.mappables.Car;
+import org.mapstruct.ap.test.subclassmapping.mappables.Vehicle;
+import org.mapstruct.control.DeepClone;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface DeepCloneMethodMapper {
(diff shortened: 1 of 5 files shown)


processor/src/test/java/org/mapstruct/ap/test/subclassmapping/DeepCloneMethodMapper.java
```diff
@@ -0,0 +1,26 @@
...
+import org.mapstruct.control.DeepClone;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface DeepCloneMethodMapper {
+    DeepCloneMethodMapper INSTANCE = Mappers.getMapper( DeepCloneMethodMapper.class );
+
+    @SubclassMapping( source = Car.class, target = Car.class )
+    @SubclassMapping( source = Bike.class, target = Bike.class )
+    @BeanMapping( mappingControl = DeepClone.class, nullValueMappingStrategy = NullValueMappingStrategy.RETURN_NULL )
```
filiphr, line 24: Why is the `nullValueMappingStrategy` needed here?

filiphr (commented):
Looks good to me. I only have one question regarding the tests.

processor/src/test/java/org/mapstruct/ap/test/subclassmapping/DeepCloneMethodMapper.java
Zegveld: Because it is not allowed to only define `mappingControl` at the moment:
`Compilation failed. Diagnostics: [DiagnosticDescriptor: ERROR DeepCloneMethodMapper.java:24 'nullValueMappingStrategy', 'nullValuePropertyMappingStrategy', 'resultType' and 'qualifiedBy' are undefined in @BeanMapping, define at least one of them.]]`

Do we want to fix that as well with this PR?

filiphr: Got it, no need to fix it in this. I've created #3040 to fix that.

PR #3032, opened 2022-09-26
#2840, #2913, #2921: MethodMatcher should not match widening methods

### LINKED ISSUE #2840: Ambiguous mapping methods when upgrading to Mapstruct 1.5.0.RC1

### LINKED ISSUE #2913: Conversion of BigDecimal to primitive double wrong with 1.5.2
After migrating from 1.4.2 to 1.5.2 a test failed. The result from BigDecimal to double is not the same anymore. Please have a look. Thanks.

https://github.com/urswiss/mapstruct-bigdecimal-to-double

### LINKED ISSUE #2921: Upgrade from 1.4.1 to 1.5.2 broke primitive to wrapper classes mapping

filiphr:
In the MethodMatcher we need to do a special check when the target type is primitive. The reason for that is that a Long is assignable to a primitive double. However, doing that means that information can be lost and thus we should not pick such methods. When the target type is primitive, then a method will be matched if and only if boxed equivalent of the target type is assignable to the boxed equivalent of the candidate return type.

Fixes #2840 
Fixes #2913 
Fixes #2921

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MethodMatcher.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MethodMatcher.java
@@ -84,7 +84,7 @@ public class MethodMatcher {
         if ( !analyser.candidateReturnType.isVoid() ) {
             if ( targetType.isPrimitive() ) {
                 // If the target type is primitive
-                // then we are going to check is its boxed equivalent
+                // then we are going to check if its boxed equivalent
                 // is assignable to the candidate return type
                 // This is done because primitives can be assigned from their own narrower counterparts
                 // directly without any casting.


sjaakd (approved):
LGTM Filip. Go ahead

processor/src/main/java/org/mapstruct/ap/internal/model/source/MethodMatcher.java
```diff
@@ -82,6 +82,17 @@ boolean matches(List<Type> sourceTypes, Type targetType) {
         // (the relation target / target type, target type being a class)
 
         if ( !analyser.candidateReturnType.isVoid() ) {
+            if ( targetType.isPrimitive() ) {
+                // If the target type is primitive
+                // then we are going to check is its boxed equivalent
```
Zegveld, line 87: typo: `check is its...` should be `check if its...`

Zegveld (approved):
Looks good to me, did find a single typo. ;)

filiphr:
Thanks for the reviews

PR #3033, opened 2022-09-27
fix(sec): upgrade freemarker to 2.3.30

ren-jq101:
Upgrade freemarker from 2.3.21 to 2.3.30 for vulnerability fix:
- [MPS-2022-12438](https://www.oscs1024.com/hd/MPS-2022-12438)


filiphr:
Thanks for the PR @ren-jq101. There were some more things that we needed to do for the upgrade. That's why I've created PR #3046 and will close this PR in favour of that one.

Issue #3037, opened 2022-09-28
Would it be possible to add  ElementType.ANNOTATION_TYPE as a target for @ValueMapping?


heipl-paul:
### Use case

I write a bunch of mappers for gRPC which suggests `UNSPECIFIED` as a default value for integration with languages that offer the possibility of enum members that were not predefined, and auto-generates `UNRECOGNIZED` (at least in java, it does).

I always want to map both of these with

```java
@ValueMapping(source = "UNSPECIFIED", target = MappingConstants.THROW_EXCEPTION)
@ValueMapping(source = "UNRECOGNIZED", target = MappingConstants.THROW_EXCEPTION)
```

Which, okay, isn't that big of a deal. But since I've been doing a lot of gRPC mappings, I've become accustomed to Mapping Composition. And instead of having a dozen mappings that are all ignoring various gRPC artifact fields, I extract them in their own specific annotation. 

```java
@Retention(RetentionPolicy.CLASS)
@Mapping(target = "mergeFrom", ignore = true)
@Mapping(target = "clearField", ignore = true)
@Mapping(target = "clearOneof", ignore = true)
@Mapping(target = "unknownFields", ignore = true)
@Mapping(target = "mergeUnknownFields", ignore = true)
@Mapping(target = "allFields", ignore = true)
public @interface IgnoreGrpcOnlyFields {
}
```

And then the following for a specific gRpc (proto) class:

```java
@IgnoreGrpcOnlyFields
@Mapping(target = "cityBytes", ignore = true)
@Mapping(target = "provinceBytes", ignore = true)
@Mapping(target = "zipCodeBytes", ignore = true)
// etc ...
public @interface IgnoreAddressGrpcOnlyFields {
}
```

I assume and hope that adding Mapping Composition to \@ValueMapping wouldn't represent a big technical challenge. And right now, the lack of this feature feels inconsistent. 

### Sample code

In the off case that I'm not clear enough, sample code of what I want to do:

```java
@ValueMapping(source = "UNSPECIFIED", target = MappingConstants.THROW_EXCEPTION)
@ValueMapping(source = "UNRECOGNIZED", target = MappingConstants.THROW_EXCEPTION)
public @interface IgnoreGrpcEnumFields {
}
```

```java
@IgnoreGrpcEnumFields
TargetEnum toEntity(SourceGrpcEnum source);
```
### Generated Code

Self evident, it would not change generated code, this only relates to reading the annotation metadata.

### Possible workarounds

[...]

PR #3041, opened 2022-10-01
#3037 Enhanced composition annotation

chenzijia12300:
Try to implement #3037
Let `IterableMapping`,`MapMapping`,`ValueMapping`,`ValueMappings` and other annotations support the combination pattern


filiphr:
@chenzijia12300 I would suggest that we simplify this PR and only focus on the support for meta annotations for `@ValueMapping` and `@ValueMappings` in a same way as it is done for `@Mapping` and `@Mappings`. I would leave the rest of the annotations as they are. The `@IterableMapping`, `@MapMapping` etc. are different type of annotations.

chenzijia12300:
@filiphr Okay, I will make changes. (●'◡'●)
I would like to ask for @IterableMapping, @MapMapping and other non-repeatable annotations, do you think it is necessary to make it support combination?

filiphr:
>Okay, I will make changes. (●'◡'●)

Thanks.

>I would like to ask for `@IterableMapping`, `@MapMapping` and other non-repeatable annotations, do you think it is necessary to make it support combination?

For now I don't think that we need this. I would rather wait for someone to request this feature than doing it immediately. I don't see a big benefit for it at the moment as well. There are some tricky edge cases with this as well. like what happens if you have defined it on 2 different annotations and things like that

chenzijia12300:
@filiphr Thank you for your patience in explaining
Incidentally, I seem to have screwed up this branch, so I'm going to create a new branch to implement this feature

PR #3043, opened 2022-10-02
#3037 Make @ValueMapping and @ValueMappings support combinations

### LINKED ISSUE #3037: Would it be possible to add  ElementType.ANNOTATION_TYPE as a target for @ValueMapping?

chenzijia12300:
Fixes #3037

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/mappingcomposition/value/CustomValueAnnotation.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/mappingcomposition/value/CustomValueAnnotation.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.mappingcomposition.value;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+import org.mapstruct.MappingConstants;
+import org.mapstruct.ValueMapping;
+import org.mapstruct.ValueMappings;
+
+/**
+ * @author orange add
+ */
+@Retention( RetentionPolicy.CLASS )
+@ValueMappings({
+    @ValueMapping(source = "EXTRA", target = "SPECIAL"),
+    @ValueMapping(source = MappingConstants.ANY_REMAINING, target = "DEFAULT")
+})
+public @interface CustomValueAnnotation {
+}
--- a/core/src/main/java/org/mapstruct/ValueMapping.java
+++ b/core/src/main/java/org/mapstruct/ValueMapping.java
@@ -84,7 +84,7 @@ import java.lang.annotation.Target;
  */
 @Repeatable(ValueMappings.class)
 @Retention(RetentionPolicy.CLASS)
-@Target(ElementType.METHOD)
+@Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})
 public @interface ValueMapping {
     /**
      * The source value constant to use for this mapping.
--- a/core/src/main/java/org/mapstruct/ValueMappings.java
+++ b/core/src/main/java/org/mapstruct/ValueMappings.java
@@ -40,7 +40,7 @@ import java.lang.annotation.Target;
  *
  * @author Sjaak Derksen
  */
-@Target(ElementType.METHOD)
+@Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})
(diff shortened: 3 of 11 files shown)


processor/src/test/java/org/mapstruct/ap/test/mappingcomposition/value/CustomValueAnnotation.java
```diff
@@ -0,0 +1,24 @@
...
+import org.mapstruct.ValueMappings;
+
+/**
+ * @author orange add
+ */
+@Retention( RetentionPolicy.CLASS )
+@ValueMappings({
+    @ValueMapping(source = "EXTRA", target = "SPECIAL"),
+    @ValueMapping(source = MappingConstants.ANY_REMAINING, target = "DEFAULT")
+})
```
filiphr, line 22: Let's use `ValueMapping` only here.

e.g.

```suggestion
@ValueMapping(source = "EXTRA", target = "SPECIAL"),
@ValueMapping(source = MappingConstants.ANY_REMAINING, target = "DEFAULT")
```

filiphr (changes requested):
Thanks @chenzijia12300. The implementation looks good. However, I think that we need to change the location of the test a bit.

We should not touch the `CompositionTest`. Instead I would suggest creating a new `ValueCompositionTest` in a new package `org.mapstruct.ap.test.value.composition` and also put the `CustomValueAnnotation` and `ValueMappingCompositionMapper` in that package.

processor/src/test/java/org/mapstruct/ap/test/mappingcomposition/value/CustomValueAnnotation.java
chenzijia12300: Have modified

chenzijia12300:
@filiphr Thank you for pointing it out, I have made the changes  ;D

sjaakd (approved):

filiphr:
Everything looks good @chenzijia12300, I was about to merge this. However, I remembered that perhaps it will be good to add an example to the documentation as well. Can you please add a new section to the [Mapping Values](https://mapstruct.org/documentation/stable/reference/html/#mapping-enum-types) chapter?

chenzijia12300:
@filiphr   No problem, I'll try to add an example to the documentation

sjaakd (approved):

filiphr:
Thanks @chenzijia12300

Issue #3040, opened 2022-09-29
Cannot use only BeanMapping#mappingControl 


filiphr:
### Expected behavior

It should be possible to only use `BeanMapping#mappingControl`

### Actual behavior

There is a compile error

### Steps to reproduce the problem

Use `@BeanMapping#mappingControl` on a method

### MapStruct Version

1.5.2

filiphr:
This is an easier good issue that can be implemented by someone from the community. If someone is interested to work on this issue please comment here so that we can assign it to you.

chenzijia12300:
@filiphr Hello, I would like to try

filiphr:
Yes sure @chenzijia12300, go for it

PR #3045, opened 2022-10-02
#3040 fix: Cannot use only BeanMapping#mappingControl

chenzijia12300:
fix: #3040


filiphr:
This looks good @chenzijia12300. However, regarding the tests, I would rather change `DeepCloneMethodMapper` and remove the `nullValueMappingStrategy` from that `@BeanMapping`. It is not needed in there anyways.

sjaakd (approved):

filiphr:
Thanks @chenzijia12300

PR #3049, opened 2022-10-05
Abstract constructor

velo:
This is most useful on CDI environments.

I have a few abstract class mappers that have a few spring inject beans. But, I can't use constructor injection as is. This change would allow me to use constructor injection


nfalco79:
I have the same needs. In 1.5.2.Final the generated class duplicate the definition of private field and the constructor, so it works. Moving to 1.5.3.Final the generator has been changed and the generated class does not duplicate anymore the definition of constructor and fields and this cause compile issue

abstract class
```
@MapperConfig(//
        componentModel = "spring",
        collectionMappingStrategy = CollectionMappingStrategy.SETTER_PREFERRED, //
        builder = @Builder(disableBuilder = true), //
        unmappedTargetPolicy = ReportingPolicy.WARN)
public abstract class IRecentSearchMapper {

    public ObjectMapper objectMapper;

    @Autowired
    public IRecentSearchMapper(ObjectMapper objectMapper) {
        this.objectMapper = objectMapper;
    }
    ...
}
```


generated class in 1.5.2
```
@Component
public class IRecentSearchMapperImpl extends IRecentSearchMapper {
    public ObjectMapper objectMapper;

    @Autowired
    public IRecentSearchMapper(ObjectMapper objectMapper) {
        this.objectMapper = objectMapper;
    }
    ...
}
```

generated class in 1.5.3
```
@Component
public class IRecentSearchMapperImpl extends IRecentSearchMapper {
    ...
}
```

filiphr:
This PR is identical to #3016. Therefore, I am going to close it.

@nfalco79 , seems like your problem is with something else. I am not sure what changed, but I would suggest that you open a separate issue.

PR #3056, opened 2022-10-15
#3054: Allow abstract return type when all directly sealed subtypes are covered by subclass mappings.

### LINKED ISSUE #3054: Do not require subclassExhaustiveStrategy when source is a sealed class and all subtypes are specified
### Use case

The `@SubclassMapping` and `@BeanMapping(subclassExhaustiveStrategy = SubclassExhaustiveStrategy.RUNTIME_EXCEPTION)` annotations can be combined to allow mapping to an abstract class or interface.  If a given source input is not one of the specified types, a `RuntimeException` will be thrown.

```java
@SubclassMapping(target = SubclassTarget.class, source = SubclassSource.class)
@BeanMapping(subclassExhaustiveStrategy = SubclassExhaustiveStrategy.RUNTIME_EXCEPTION)
AbstractTarget map(AbstractSource source);
```

=>

```java
@Override
public AbstractTarget map(AbstractSource source) {
    if ( source == null ) {
        return null;
    }

    if (source instanceof SubclassSource) {
        return subclassSourceToSubclassTarget( (SubclassSource) source );
    }
    else {
        throw new IllegalArgumentException("Not all subclasses are supported for this mapping. Missing for " + source.getClass());
    }
}
```

However, in certain scenarios it is possible to guarantee at compile time that all possible source types are accounted for.  In this case, `subclassExhaustiveStrategy` seems superfluous, and could arguably be dropped.  One scenario where we know all source types are accounted for is if the source is a [sealed class](https://openjdk.org/jeps/409) (or sealed interface), and all classes it permits are handled.

#### Example

Take the following type mapper & type hierarchy:

```java
@Mapper
public interface SealedSourceMapper {
    @SubclassMapping(target = FinalSubclassTarget.class, source = FinalSubclassSource.class)
    @SubclassMapping(target = FinalGrandchildTarget.class, source = FinalGrandchildSource.class)
    @SubclassMapping(target = NonSealedSubclassTarget.class, source = NonSealedSubclassSource.class)
    Target mapToAbstractClass(SealedAbstractClassSource source);
```

```java
public abstract sealed class SealedAbstractClassSource 
        permits FinalSubclassSource, SealedSubclassSource, NonSealedSubclassSource {
    private String property;
    public String getProperty() {return property;}
    public void setProperty(String property) {this.property = property;}
}

public final class FinalSubclassSource extends SealedAbstractClassSource {
}

public abstract sealed class SealedSubclassSource extends SealedAbstractClassSource 
        permits FinalGrandchildSource {
}

public final class FinalGrandchildSource extends SealedSubclassSource {
}

public non-sealed class NonSealedSubclassSource extends SealedAbstractClassSource {
}

public abstract class Target {
    private String property;
    public String getProperty() {return property;}
    public void setProperty(String property) {this.property = property;}
}

public class FinalSubclassTarget extends Target {
}

public class FinalGrandchildTarget extends Target {
}

public class NonSealedSubclassTarget extends Target {
}
```

In this case, at compile time, it can be determined that every possible type of the `SealedAbstractClassSource` source argument is accounted for.  Therefore, specifying a `subclassExhaustiveStrategy` should not be required since it is known at compile time to be exhaustive.

Note: this is just as applicable if the source type is a sealed interface as if it were a sealed class.

### Generated Code

```java
public class SealedSourceMapperImpl implements SealedSourceMapper {
    @Override
    public Target mapToAbstractClass(SealedAbstractClassSource source) {
        if ( source == null ) {
            return null;
        }

        if (source instanceof FinalSubclassSource) {
            return finalSubclassSourceToFinalSubclassTarget( (FinalSubclassSource) source );
        }
        else if (source instanceof FinalGrandchildSource) {
            return finalGrandchildSourceToFinalGrandchildTarget( (FinalGrandchildSource) source );
        }
        else if (source instanceof NonSealedSubclassSource) {
            return nonSealedSubclassSourceToNonSealedSubclassTarget( (NonSealedSubclassSource) source );
        }
        else {
            // Same exception type as "Pattern Matching for switch" JEP if incompatible separate compilation
            throw new IncompatibleClassChangeError("Not all subclasses are supported for this mapping. Missing for " + source.getClass());
        }
    }
}
```

### Possible workarounds

This is a minor quality of life enhancement that can easily be worked around by adding `@BeanMapping(subclassExhaustiveStrategy = SubclassExhaustiveStrategy.RUNTIME_EXCEPTION)` to the mapping method.

```java
    @SubclassMapping(target = FinalSubclassTarget.class, source = FinalSubclassSource.class)
    @SubclassMapping(target = FinalGrandchildTarget.class, source = FinalGrandchildSource.class)
    @SubclassMapping(target = NonSealedSubclassTarget.class, source = NonSealedSubclassSource.class)
    @BeanMapping(subclassExhaustiveStrategy = SubclassExhaustiveStrategy.RUNTIME_EXCEPTION)
    Target mapToAbstractClass(SealedAbstractClassSource source);
```

### MapStruct Version

org.mapstruct:mapstruct:1.5.3.Final

Zegveld:
@filiphr do you know how to add a failure test for java 17+ situations?

Currently I've only added the happy flow situation to the integration test.

Fixes #3054

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -441,29 +441,42 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
         private boolean isCorrectlySealed() {
             try {
-                if ( method.getMappingSourceType().isSealed() ) {
-                    List<? extends TypeMirror> unusedPermittedSubclasses =
-                        new ArrayList<>( method.getMappingSourceType().getPermittedSubclasses() );
-                    method.getOptions().getSubclassMappings().forEach( subClassOption -> {
-                        unusedPermittedSubclasses
-                                                 .stream()
-                                                 .filter(
-                                                     permitted -> ctx
-                                                                     .getTypeUtils()
-                                                                     .isSameType(
-                                                                         permitted,
-                                                                         subClassOption.getSource() ) )
-                                                 .findFirst()
-                                                 .ifPresent( unusedPermittedSubclasses::remove );
-                    } );
-                    return unusedPermittedSubclasses.isEmpty();
-                }
-                return false;
+                Type mappingSourceType = method.getMappingSourceType();
+                return isCorrectlySealed( mappingSourceType );
             }
             catch ( Exception e ) {
                 // SEALED not supported.
                 return false;
             }
+        }
+
+        private boolean isCorrectlySealed(Type mappingSourceType) {
+            if ( mappingSourceType.isSealed() ) {
+                List<? extends TypeMirror> unusedPermittedSubclasses =
+                    new ArrayList<>( mappingSourceType.getPermittedSubclasses() );
+                method.getOptions().getSubclassMappings().forEach( subClassOption -> {
+                    unusedPermittedSubclasses
+                                             .stream()
+                                             .filter(
+                                                 permitted -> ctx
+                                                                 .getTypeUtils()
+                                                                 .isSameType(
+                                                                     permitted,
+                                                                     subClassOption.getSource() ) )
+                                             .findFirst()
+                                             .ifPresent( unusedPermittedSubclasses::remove );
+                } );
+                for ( Iterator<? extends TypeMirror> iterator = unusedPermittedSubclasses.iterator();
+                                iterator.hasNext(); ) {
+                    TypeMirror typeMirror = iterator.next();
+                    Type type = ctx.getTypeFactory().getType( typeMirror );
+                    if ( type.isAbstract() && isCorrectlySealed( type ) ) {
+                        iterator.remove();
+                    }
+                }
+                return unusedPermittedSubclasses.isEmpty();
+            }
+            return false;
         }
 
         private void initializeMappingReferencesIfNeeded(Type resultTypeToMap) {
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -436,8 +436,36 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         }
 
         private boolean isAbstractReturnTypeAllowed() {
-            return method.getOptions().getBeanMapping().getSubclassExhaustiveStrategy().isAbstractReturnTypeAllowed()
-                && !method.getOptions().getSubclassMappings().isEmpty();
+            return !method.getOptions().getSubclassMappings().isEmpty()
+                && ( method.getOptions().getBeanMapping().getSubclassExhaustiveStrategy().isAbstractReturnTypeAllowed()
+                    || isCorrectlySealed() );
+        }
+
+        private boolean isCorrectlySealed() {
+            try {
+                if ( method.getMappingSourceType().isSealed() ) {
+                    List<? extends TypeMirror> unusedPermittedSubclasses =
+                        new ArrayList<>( method.getMappingSourceType().getPermittedSubclasses() );
+                    method.getOptions().getSubclassMappings().forEach( subClassOption -> {
+                        unusedPermittedSubclasses
+                                                 .stream()
+                                                 .filter(
+                                                     permitted -> ctx
+                                                                     .getTypeUtils()
+                                                                     .isSameType(
+                                                                         permitted,
+                                                                         subClassOption.getSource() ) )
+                                                 .findFirst()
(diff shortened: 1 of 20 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -436,8 +436,36 @@ private SubclassMapping createSubclassMapping(SubclassMappingOptions subclassMap
...
+                        new ArrayList<>( method.getMappingSourceType().getPermittedSubclasses() );
+                    method.getOptions().getSubclassMappings().forEach( subClassOption -> {
+                        unusedPermittedSubclasses
+                                                 .stream()
+                                                 .filter(
+                                                     permitted -> ctx
+                                                                     .getTypeUtils()
+                                                                     .isSameType(
+                                                                         permitted,
+                                                                         subClassOption.getSource() ) )
```
mjustin, line 457: I haven't played around with this, but would changing this check to be either the "same type" check or a recursive "isCorrectlySealed" check be enough to fully cover a sealed subtype hierarchy, rather than just the case where it's only one-level deep and only the direct subtypes are covered?

filiphr:
>@filiphr do you know how to add a failure test for java 17+ situations?

In order to achieve this we are going to need to use java 17 for our processor test suite. We used to have something like this back when we were supporting java 7. The pattern would be to use `java17` in test suite packages and then make sure that the `MavenIntegrationTest#fullFeatureTest` ignores all packages with `java17` when run with Java 8.

btw @Zegveld, I've been really busy with other things and haven't had the time to properly review things. I hope I can catch up soon.

Zegveld:
@filiphr, no worries. I've been busy myself (and still am atm) so it will take a while before I have more time then a few minutes to look at things here myself.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -434,8 +434,49 @@ private SubclassMapping createSubclassMapping(SubclassMappingOptions subclassMap
...
+                && ( method.getOptions().getBeanMapping().getSubclassExhaustiveStrategy().isAbstractReturnTypeAllowed()
+                    || isCorrectlySealed() );
+        }
+
+        private boolean isCorrectlySealed() {
+            try {
+                Type mappingSourceType = method.getMappingSourceType();
+                return isCorrectlySealed( mappingSourceType );
+            }
+            catch ( Exception e ) {
```
filiphr, line 447: What kind of an exception can happen here?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -434,8 +434,49 @@ private SubclassMapping createSubclassMapping(SubclassMappingOptions subclassMap
...
+                                             .stream()
+                                             .filter(
+                                                 permitted -> ctx
+                                                                 .getTypeUtils()
+                                                                 .isSameType(
+                                                                     permitted,
+                                                                     subClassOption.getSource() ) )
+                                             .findFirst()
+                                             .ifPresent( unusedPermittedSubclasses::remove );
+                } );
```
filiphr, line 468: I think that this would be more readable if we write in without streams. Perhaps use an iterator` for removing  from the `unusedPermittedSubclasses`

filiphr (changes requested):
I have some questions regarding the implementation. When you have some time have a look at my comments @Zegveld

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
Zegveld: true, reading the code there should be no chance of an exception happening here when not java 17.

filiphr: In that case let's remove the try catch. We should handle it properly if something throws an exception

Zegveld:
I've made the wanted changes, and they were definitely an improvement.

>> @filiphr do you know how to add a failure test for java 17+ situations?
> In order to achieve this we are going to need to use java 17 for our processor test suite.
So I guess this would become a new issue to move the processor test suite to java 17?
(And perhaps repeat this every time a java LTS version is released?)

filiphr:
>So I guess this would become a new issue to move the processor test suite to java 17?
(And perhaps repeat this every time a java LTS version is released?)

We don't need to do it now. Perhaps we can create an issue for when we change our LTS version. I am not sure how easy / difficult is to have the test suite with a different java version in Maven

PR #3064, opened 2022-10-30
#2863 Add validation of String type to @TargetPropertyName

### LINKED ISSUE #2863: Add validation of `String` type for `@TargetPropertyName`
Give a clear error message if another type is used.

paparadva:
Check that a parameter annotated with @TargetPropertyName is String and report an error.

With the following mapper:

```java
@Mapper
public interface ErroneousNonStringTargetPropertyNameParameter {

    Employee map(EmployeeDto employee);

    @Condition
    default boolean isNotBlank(String value, @TargetPropertyName int propName) { // propName is int
        return value != null && !value.trim().isEmpty();
    }
}
```

Compilation will fail with this error message:

> error: @TargetPropertyName can only by applied to a String parameter.

Fixes #2863

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
@@ -28,6 +28,7 @@ import org.mapstruct.ap.internal.gem.MappingsGem;
 import org.mapstruct.ap.internal.gem.ObjectFactoryGem;
 import org.mapstruct.ap.internal.gem.SubclassMappingGem;
 import org.mapstruct.ap.internal.gem.SubclassMappingsGem;
+import org.mapstruct.ap.internal.gem.TargetPropertyNameGem;
 import org.mapstruct.ap.internal.gem.ValueMappingGem;
 import org.mapstruct.ap.internal.gem.ValueMappingsGem;
 import org.mapstruct.ap.internal.model.common.Parameter;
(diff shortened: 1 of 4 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
```diff
@@ -406,7 +406,13 @@ private boolean hasFactoryAnnotation(ExecutableElement method) {
         return ObjectFactoryGem.instanceOn( method ) != null;
     }
 
-    private boolean isValidPresenceCheckMethod(ExecutableElement method, Type returnType) {
+    private boolean isValidPresenceCheckMethod(ExecutableElement method, List<Parameter> parameters, Type returnType) {
+        for ( Parameter param : parameters ) {
+            if ( param.isTargetPropertyName() && !param.getType().isString() ) {
+                messager.printMessage( method, Message.RETRIEVAL_TARGET_PROPERTY_NAME_WRONG_TYPE );
```
filiphr, line 412: We can actually pass the annotation mirror from the `@TargetPropertyName` here as well. We also should use the parameter name element for the error element.

e.g.

```java
message.printMessage(
    param.getElement(),
    TargetPropertyNameGem.instanceOn( param.getElement() ).mirror() ),
    Message. RETRIEVAL_TARGET_PROPERTY_NAME_WRONG_TYPE
);
```

filiphr (changes requested):
The PR is great @paparadva.

I've only left one small comment around the error. You can actually see how it will look better in the error line if the erroneous mapper has the method like:

```java
default boolean isNotBlank(
    String value,
    @TargetPropertyName int propName
) {
        return value != null && !value.trim().isEmpty();
}
```

paparadva:
I updated isValidPresenceCheckMethod, the error message now points to propName:
```
    default boolean isNotBlank(String value, @TargetPropertyName int propName) {
                                             ^
```

filiphr:
Thanks for your work on this @paparadva

PR #3068, opened 2022-11-03
Add support for globally defining nullValueMapMappingStrategy and nullValueIterableMappingStrategy

### LINKED ISSUE #2953: Feature request, additional global defaults
I was wondering if there was a reason that the null value mapping strategies were not added to the Options class in order to override the built-in defaults? If there is no reason preventing this from being done I was hoping it could be added as feature moving forward?

I'm specifically interested in the`nullValueIterableMappingStrategy` and `nullValueMapMappingStrategy` as accepted best practice in this instance is to return empty collections and maps as opposed to null values. Basically I want to set this policy globally instead of via the currently available options.

Cheers, Andy.

filiphr:
Fixes #2953

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/nullvaluemapping/NullValueIterableMappingStrategyTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/nullvaluemapping/NullValueIterableMappingStrategyTest.java
@@ -7,6 +7,7 @@ package org.mapstruct.ap.test.nullvaluemapping;
 
 import org.mapstruct.ap.test.nullvaluemapping._target.CarDto;
 import org.mapstruct.ap.test.nullvaluemapping.source.Car;
+import org.mapstruct.ap.testutil.IssueKey;
 import org.mapstruct.ap.testutil.ProcessorTest;
 import org.mapstruct.ap.testutil.WithClasses;
 import org.mapstruct.ap.testutil.compilation.annotation.ProcessorOption;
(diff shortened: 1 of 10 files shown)


processor/src/test/java/org/mapstruct/ap/test/nullvaluemapping/NullValueIterableMappingStrategyTest.java
```diff
@@ -0,0 +1,43 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Filip Hrisafov
+ */
+@WithClasses({
+    CarDto.class,
+    Car.class
+})
+public class NullValueIterableMappingStrategyTest {
```
Zegveld, line 23: missing the issuekey annotation here I believe.

processor/src/test/java/org/mapstruct/ap/test/nullvaluemapping/NullValueMapMappingStrategyTest.java
```diff
@@ -0,0 +1,43 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Filip Hrisafov
+ */
+@WithClasses({
+    CarDto.class,
+    Car.class
+})
+public class NullValueMapMappingStrategyTest {
```
Zegveld, line 23: missing the issuekey annotation here I believe.

Zegveld (changes requested):
Looks good to me, just a minor thing about missing `IssueKey` annotations at the tests.

Zegveld (approved):

mikybars:
Somebody could explain why this feature is not present in the latest release?

filiphr:
The feature is not present in 1.5 because the fix version of #2953 is 1.6

PR #3070, opened 2022-11-04
Pass spi namespaced processor options to custom spi

### LINKED ISSUE #3071: Support configurable Spi Implementations
### Use case

Spi Implementations have a better opportunity for reusability if they are configurable. Currently MapStruct's `MappingProcessor` does not pass the compilerArgs down to the `MapStructProcessingEnvironment` which is received on the `init` methods of the Spi classes (such as `AccessorNamingStrategy`).

Example of such use case: https://github.com/entur/mapstruct-spi-protobuf/pull/133 (this approach ultimately failed because a workaround was needed to retrieve the `options` that MapStruct does not pass through).

### Generated Code

To resolve this we can just pass the options down to `MapStructProcessingEnvironment`. To be safe, we can limit this to only options under the namespace of `mapstruct.ap.spi`.

### Possible workarounds

There are no viable work arounds. An attempt was made to have the options loaded from a parallel processor but there is no guarantee of processor initialization order.

### MapStruct Version

MapStruct 1.5.3

ro0sterjam:
Currently we are unable to configure custom mapstruct spi implementations because the compilerArg options are not passed down to the spi classes.

This PR exposes these options to `MapStructProcessingEnvironment`. I've limited it to only options under the namespace of `mapstruct.ap.spi.` to limit the exposure of unrelated options.

Fixes #3071


filiphr:
@ro0sterjam have a look at my comments in #3071

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
```diff
@@ -393,6 +406,26 @@ public TypeElement visitTypeAsClass(TypeElement e, Void p) {
...
+     * @return the custom supported options
+     */
+    private static Set<String> getCustomSupportedOptions() {
+        final CustomSupportedOptionsProvider customSupportedOptionsProvider =
+            Services.get( CustomSupportedOptionsProvider.class, new DefaultCustomSupportedOptionsProvider() );
+        final Set<String> customSupportedOptions = customSupportedOptionsProvider.getCustomSupportedOptions();
+
+        // Ensure custom options are not in the mapstruct namespace
+        customSupportedOptions.forEach( option -> {
+            if (option.startsWith( "mapstruct" )) {
```
ro0sterjam, line 421: is this enough to check that the custom option does not start with `mapstruct` or do we want to be more thorough?

ro0sterjam:
@filiphr thank you for your guidance. Can you take a look at my latest commit and tell me what you think? If it looks good I'll proceed with updating the documentation.

ro0sterjam:
Added docs and test

ro0sterjam:
Bump on this.

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
filiphr: this is enough yes

filiphr:
Thanks for your work on this @ro0sterjam. There is no need to do a bump. We are working on this in our free time and we don't always have time to speedily review it. The PR will be part of the 1.6 release and we'll review it soon.

ro0sterjam:
Thanks @filiphr, completely understood and apologies for the bump.

filiphr:
@ro0sterjam I did some small polishing. Can you please check it out and let me know if you agree with it? If yes then I will resolve the conflict and merge it manually (unless you want to resolve the conflict and force push to your branch)

ro0sterjam:
Looked over the changes and no objections from me! Good idea about supporting multiple `AdditionalSupportedOptionsProvider`.

filiphr:
This has been integrated into main. Thanks @ro0sterjam

Issue #3072, opened 2022-11-05
MapStruct recognizes 'remove' method as property


trks1970:
### Expected behavior

I have  JPA Entity with two methods for bidirectional collection manipulation:
```
@ManyToOne(optional = false, fetch = FetchType.LAZY)
    private AssetHierarchyNodeEntity parent;

@OneToMany(
        mappedBy = "parent",
        cascade = CascadeType.ALL,
        orphanRemoval = true
    )
    private Set<AssetHierarchyNodeEntity> children = new HashSet<>();

    public AssetHierarchyNodeEntity addChild(final AssetHierarchyNodeEntity assetHierarchyNode) {
        assetHierarchyNode.setParent(this);
        children.add(assetHierarchyNode);
        return assetHierarchyNode;
    }

    public AssetHierarchyNodeEntity removeChild(final AssetHierarchyNodeEntity assetHierarchyNode) {
        children.remove(assetHierarchyNode);
        assetHierarchyNode.setParent(null);
        return assetHierarchyNode;
    }
```

when creating a mapper, mapstruct complains, that `removeChild` is an unmapped property. (strangely it does not complain about addChild not being mapped)

### Actual behavior

Mapstruct should only map real properties

### Steps to reproduce the problem

```
import lombok.EqualsAndHashCode;
import lombok.Getter;
import lombok.Setter;
import lombok.ToString;

import javax.persistence.CascadeType;
import javax.persistence.Column;
import javax.persistence.Entity;
import javax.persistence.FetchType;
import javax.persistence.GeneratedValue;
import javax.persistence.GenerationType;
import javax.persistence.Id;
import javax.persistence.ManyToOne;
import javax.persistence.OneToMany;
import javax.persistence.SequenceGenerator;
import javax.persistence.Table;
import javax.persistence.Version;
import java.util.HashSet;
import java.util.Set;

@Entity
@Table(name = "asset_hierarchy_node")
@Getter
@Setter
@EqualsAndHashCode(onlyExplicitlyIncluded = true)
@ToString
public class AssetHierarchyNodeEntity {
    @Id
    @SequenceGenerator(
        name = "seq_asset_hierarchy_node",
        sequenceName = "seq_asset_hierarchy_node",
        initialValue = 1000,
        allocationSize = 1
    )
    @GeneratedValue(
        strategy = GenerationType.SEQUENCE,
        generator = "seq_asset_hierarchy"
    )
    @Column(name = "id", nullable = false, updatable = false)
[...]

PR #3082, opened 2022-11-16
#3072 Fixes proper ignoring of "removers" that often is accompanying …

jarlesat:
Ignore unmapped "removers" in target classes. Fluent removers are generated by tools like [OpenAPI generator](https://github.com/OpenAPITools/openapi-generator), preventing the use of `unmappedTargetPolicy=ERROR` unless adding `@Mapping(target='removeItem', ignore=true)` for every single remover.


jarlesat:
Should this branch be rebased on main, as main now is version 1.7-SNAPSHOT? Anything else that should be done with it?

filiphr:
Sorry I have not reviewed this. @jarlesat it would indeed be good to rebase it on top of main.

Additionally, if you can move the test in `org.mapstruct.ap.test.collection.remover` it would be great. Now it us under `bugs`, but this is not a bug, it's a new feature.

jarlesat:
Is it ok to move the test to org.mapstruct.ap.spi and extend it to a complete test of DefaultAccessorNamingStrategy? There does not seem to be any testing of that class from.

jarlesat:
Please let me know if there are anything more that should be done, test cases or anything. Maybe branch-name?

PR #3127, opened 2022-12-30
#3125: Allow subclassmapping inheritance for methods with identical signature.

### LINKED ISSUE #3125: Support `@InheritConfiguration` for `@SubclassMapping` in methods with identical signature
### Expected behavior

Working on #3120 i realized that @SubclassMapping wasn't been inherited.

So in a mapper like this i was expecting that TestMapper#mapAnother will inherit all the annotations and basically behave the same way as TestMapper#map:

```
@Mapper
public interface TestMapper {
    TestMapper INSTANCE = Mappers.getMapper( TestMapper.class );

    @SubclassMapping(source = CarDto.class, target = Car.class)
    @SubclassMapping(source = BikeDto.class, target = Bike.class)
    @Mapping(source = "maker", target = "vehicleManufacturingCompany")
    Vehicle map(VehicleDto vehicle);

    @InheritConfiguration
    Vehicle mapAnother(VehicleDto vehicleDto);
}
```

### Actual behavior

But mapstruct generate this, correctly inheriting the mapping annotation but leaving out the subclass related stuff:

```
public class TestMapperImpl implements TestMapper {

    @Override
    public Vehicle map(VehicleDto vehicle) {
        if ( vehicle == null ) {
            return null;
        }

        if (vehicle instanceof CarDto ) {
            return  carDtoToCar( (CarDto) vehicle );
        }
        else if (vehicle instanceof BikeDto ) {
            return  bikeDtoToBike( (BikeDto) vehicle );
        }
        else {
            Vehicle vehicle1 = new Vehicle();

            vehicle1.setVehicleManufacturingCompany( vehicle.getMaker() );
            vehicle1.setName( vehicle.getName() );

            return vehicle1;
        }
    }

    @Override
    public Vehicle mapAnother(VehicleDto vehicleDto) {
        if ( vehicleDto == null ) {
            return null;
        }

        Vehicle vehicle = new Vehicle();

        vehicle.setVehicleManufacturingCompany( vehicleDto.getMaker() );
        vehicle.setName( vehicleDto.getName() );

        return vehicle;
    }

    protected Car carDtoToCar(CarDto carDto) {
        if ( carDto == null ) {
            return null;
        }

        Car car = new Car();

        car.setVehicleManufacturingCompany( carDto.getMaker() );
        car.setName( carDto.getName() );
        car.setManual( carDto.isManual() );

        return car;
    }

    protected Bike bikeDtoToBike(BikeDto bikeDto) {
        if ( bikeDto == null ) {
            return null;
        }

        Bike bike = new Bike();

        bike.setVehicleManufacturingCompany( bikeDto.getMaker() );
        bike.setName( bikeDto.getName() );
        bike.setNumberOfGears( bikeDto.getNumberOfGears() );

        return bike;
    }
}
```

What the expected behavior says.

### MapStruct Version

MapStruct 1.5.3

Zegveld:
closes #3125

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
@@ -21,7 +21,6 @@ import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.common.TypeFactory;
 import org.mapstruct.ap.internal.util.FormattingMessager;
 import org.mapstruct.ap.internal.util.Message;
-import org.mapstruct.ap.internal.util.TypeUtils;
 import org.mapstruct.ap.internal.util.accessor.Accessor;
 
 /**
@@ -31,7 +30,6 @@ import org.mapstruct.ap.internal.util.accessor.Accessor;
  */
 public class MappingMethodOptions {
     private static final MappingMethodOptions EMPTY = new MappingMethodOptions(
-        null,
         null,
         Collections.emptySet(),
         null,
@@ -55,15 +53,12 @@ public class MappingMethodOptions {
 
     private SubclassValidator subclassValidator;
 
-    private TypeUtils typeUtils;
-
-    public MappingMethodOptions(TypeUtils typeUtils, MapperOptions mapper, Set<MappingOptions> mappings,
+    public MappingMethodOptions(MapperOptions mapper, Set<MappingOptions> mappings,
                                 IterableMappingOptions iterableMapping,
                                 MapMappingOptions mapMapping, BeanMappingOptions beanMapping,
                                 EnumMappingOptions enumMappingOptions,
                                 List<ValueMappingOptions> valueMappings,
                                 Set<SubclassMappingOptions> subclassMappings, SubclassValidator subclassValidator) {
-        this.typeUtils = typeUtils;
         this.mapper = mapper;
         this.mappings = mappings;
         this.iterableMapping = iterableMapping;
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
```diff
@@ -232,6 +245,29 @@ public void applyInheritedOptions(SourceMethod sourceMethod, SourceMethod templa
...
+        return parametersAreOfIdenticalTypeAndOrder( templateMethod, sourceMethod )
+            && resultTypeIsTheSame( templateMethod, sourceMethod );
+    }
+
+    private boolean parametersAreOfIdenticalTypeAndOrder(SourceMethod templateMethod, SourceMethod sourceMethod) {
+        if (templateMethod.getParameters().size() != sourceMethod.getParameters().size()) {
+            return false;
+        }
+        for ( int i = 0; i < templateMethod.getParameters().size(); i++ ) {
+            if (!typeUtils.isSameType( templateMethod.getParameters().get( i ).getType().getTypeMirror(),
```
filiphr, line 258: I think that here we can just use `tempalteMethod.getParameter.get( i ).getType().equals( sourceMethod.getParameters().get( i ).getType() )`

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
```diff
@@ -232,6 +245,29 @@ public void applyInheritedOptions(SourceMethod sourceMethod, SourceMethod templa
...
+                                       sourceMethod.getParameters().get( i ).getType().getTypeMirror() ) ) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean resultTypeIsTheSame(SourceMethod templateMethod, SourceMethod sourceMethod) {
+        return typeUtils.isSameType( templateMethod.getResultType().getTypeMirror(),
```
filiphr, line 267: Same here we can use the equals of the type

filiphr (changes requested):
Looks good to me @Zegveld. I've added one comment in order to avoid the need of the `TypeUtils` in the `MappingMethodOptions`

filiphr:
Thanks @Zegveld

PR #3128, opened 2022-12-30
#3126: do not change source parameter name for subclassmappings

### LINKED ISSUE #3126: `@SubclassMapping` not working with `@Mapping` nested properties and target all
### Expected behavior

It should be able to map the nested properties from source to target subclass.

Usage

```java
@SubclassMapping(source = HomeAddressDO.class, target = HomeAddressResponseDto.class)
@SubclassMapping(source = OfficeAddressDO.class, target = OfficeAddressResponseDto.class)
@Mapping(source = "auditable", target = ".")
AddressResponseDto toDto(AddressDO addressDo);
```

### Actual behavior

`MapStruct` complains the following:

```
No property named "addressDo.auditable.createdBy" exists in source parameter(s). Did you mean "auditable"?
```

image

I have created a reproduce at [here](https://github.com/bwgjoseph/mapstruct-with-subclass)

In summary, this is the setup

```java
public abstract class AddressDO {
    public abstract AddressDOBuilder<?, ?> toBuilder();

    private String id;
    private Auditable auditable;
}

public class Auditable {
    private String createdBy;
    private LocalDateTime createdAt;
}

public class HomeAddressDO extends AddressDO {
  // some fields
}

public class OfficeAddressDO extends AddressDO {
  // some fields
}

public interface AddressResponseDto {}

public record HomeAddressResponseDto(String id, String street, String postalCode, String unit, String createdBy, String createdAt) implements AddressResponseDto {}

public record OfficeAddressResponseDto(String id, String building, String street, String postalCode, String unit, String createdBy, String createdAt) implements AddressResponseDto {}

@Mapper(
    componentModel = MappingConstants.ComponentModel.SPRING,
)
public interface AddressMapper {
    @SubclassMapping(source = HomeAddressDO.class, target = HomeAddressResponseDto.class)
    @SubclassMapping(source = OfficeAddressDO.class, target = OfficeAddressResponseDto.class)
    @Mapping(source = "auditable", target = ".") // < offending line
    AddressResponseDto toDto(AddressDO addressDo);
}
```

### MapStruct Version

1.5.3.Final

Zegveld:
keep parameter name intact for subclassmappings to ensure inherited mappings keep working.

closes #3126

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
@@ -141,6 +141,7 @@ public class ForgedMethod implements Method {
         return new ForgedMethod(
             name,
             sourceType,
+            new Parameter(basedOn.getSourceParameters().get( 0 ).getName(), sourceType),
             returnType,
             basedOn.getContextParameters(),
             basedOn,
@@ -153,7 +154,12 @@ public class ForgedMethod implements Method {
     private ForgedMethod(String name, Type sourceType, Type returnType, List<Parameter> additionalParameters,
                          Method basedOn, ForgedMethodHistory history, MappingReferences mappingReferences,
                          boolean forgedNameBased) {
+        this( name, sourceType, determineSourceParameterName( sourceType, additionalParameters ), returnType,
+            additionalParameters, basedOn, history, mappingReferences, forgedNameBased );
 
+    }
+
+    private static Parameter determineSourceParameterName(Type sourceType, List<Parameter> additionalParameters) {
         // establish name
         String sourceParamSafeName;
         if ( additionalParameters.isEmpty() ) {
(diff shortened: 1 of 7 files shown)


filiphr (commented):
I am not sure that I like the fix for this one to be honest. It feels to me like we are not fixing it in the right place. 

Can't we fix this in the right place, i.e. make sure that the generated mappings do not contain the source parameter, or only include the source parameter when there are multiple source parameters?

Zegveld:
> I am not sure that I like the fix for this one to be honest. It feels to me like we are not fixing it in the right place.
> 
> Can't we fix this in the right place, i.e. make sure that the generated mappings do not contain the source parameter, or only include the source parameter when there are multiple source parameters?

The correct fix (imho) would be to redo the analysis of the mappings after transferring the mapping annotations for each of the newly generated subclass mapping methods. But the problem here is I do not know how/where to fix this behavior.

Introducing different behavior for single or multiple source parameters will not be pretty. So I'm against that option. You do not want to constantly check whether or not a mapping is on a method with a _single_ or _multiple_ parameters.

What shall we do with this issue:
* use this fix and create a new issue to correctly fix this
* try to fix the way analysis is done for the forged subclass mapping methods by
  * updating the mappings to replace the old parameter with the new parameter value
  * copying the original (not analyzed) mapping information to the forged subclass mapping methods
  * not caching the mapping information but re-evaluating it constantly
  * some other method I did not think of that would also fix this issue

It would be nice to have the issue at least fixed in the next release.

filiphr:
I think that the best approach would be to fix it by passing the mapping annotations as they are to each bean mapping method generation and resolve the target this mapping on that level. That's "not caching the mapping information but re-evaluating it constantly" I believe. Relying on things like the name of the parameters in the code we generate is a bit fragile and hacky in my opinion

>It would be nice to have the issue at least fixed in the next release.

I agree with this. The issue is currently targeting 1.6, so we can make sure that we fix it in that one.

filiphr:
@Zegveld can you please check PR #3281. It is my suggestion for fixing #3126 instead of this PR here

Zegveld:
@filiphr's PR is better then this one. Closing this one.

PR #3130, opened 2023-01-02
#3129: detect collection adders using Iterable setter/getter

### LINKED ISSUE #3126: `@SubclassMapping` not working with `@Mapping` nested properties and target all

### LINKED ISSUE #3129: CollectionMappingStrategy.ADDER_PREFERRED is not supported for Iterables
### Expected behavior

Having a source of type:
```java
public class Source {
    private java.util.List<String> edges;
    private java.util.List<String> points;

    public Source(List<String> edges, List<String> points) {
        this.edges = edges;
        this.points = points;
    }

    public List<String> getEdges() {
        return edges;
    }

    public List<String> getPoints() {
        return points;
    }
}
```
and a target of type:
```java
@lombok.Builder
@lombok.Value
public class Target{
    final @lombok.Singular("addEdge") com.google.common.collect.ImmutableList<String> edges;
    final @lombok.Singular("addPoint") java.util.List<String> points;
}
```
and a mapper interface:
```java
import org.mapstruct.CollectionMappingStrategy;
import org.mapstruct.Mapper;

@Mapper(collectionMappingStrategy = CollectionMappingStrategy.ADDER_PREFERRED)
public interface PathMapper {
    Target map(Source source);
}
```

, the mapper implementation uses `target.addEdge` to copy edges because of `CollectionMappingStrategy.ADDER_PREFERRED` is used.

### Actual behavior

`target.addEdge` is not used by the generated mapper implementation. Note that but `target.addPoint` is used as expected.

Mapper implementation code:
```java
public class MapperImpl implements Mapper {
    public MapperImpl() {
    }

        if (source == null) {
            return null;
        } else {
            Target.TargetBuilder target = Target.builder();
            target.edges(source.getEdges());
            if (source.getPoints() != null) {
                Iterator var3 = source.getPoints().iterator();

                while(var3.hasNext()) {
                    String point = (String)var3.next();
                    target.addPoint(point);
                }
            }

            return target.build();
        }
    }
}
```

Use the `Source`, `Target` and `Mapper` classes to reproduce the issue. Lombok 1.18.24 was used with a default `lombok.singular.useGuava=false` setting. Note that  `lombok.singular.useGuava=true` would cause that both target properties (`edges`, `points`) will use [guava immutables](https://projectlombok.org/features/Builder) and the mapper implementation will skip usage of `target.addPoint` contrary to the expectation.

The issue is also reproducible without lombok, the actual Target class was transpiled by lombok to: 
```java
// unrelated methods to this issue (equals,hashcode ...)  were removed
import com.google.common.collect.ImmutableList;
import java.util.ArrayList;
import java.util.Collection;
import java.util.Collections;
import java.util.List;

public final class Target {
    private final ImmutableList<String> edges;
    private final List<String> points;

    Target(ImmutableList<String> edges, List<String> points) {
        this.edges = edges;
        this.points = points;
    }
    public static TargetBuilder builder() {
        return new TargetBuilder();
    }
    public ImmutableList<String> getEdges() {
        return this.edges;
    }
    public List<String> getPoints() {
        return this.points;
    }
    public static class TargetBuilder {
        private ImmutableList.Builder<String> edges;
        private ArrayList<String> points;

        TargetBuilder() {
        }

        public TargetBuilder addEdge(String addEdge) {
            if (this.edges == null) {
                this.edges = ImmutableList.builder();
            }

            this.edges.add(addEdge);
            return this;
        }

        public TargetBuilder edges(Iterable<? extends String> edges) {
            if (edges == null) {
                throw new NullPointerException("edges cannot be null");
            } else {
                if (this.edges == null) {
                    this.edges = ImmutableList.builder();
                }

                this.edges.addAll(edges);
                return this;
            }
        }

        public TargetBuilder addPoint(String addPoint) {
            if (this.points == null) {
                this.points = new ArrayList();
            }

            this.points.add(addPoint);
            return this;
        }

        public TargetBuilder points(Collection<? extends String> points) {
            if (points == null) {
                throw new NullPointerException("points cannot be null");
            } else {
                if (this.points == null) {
                    this.points = new ArrayList();
                }

                this.points.addAll(points);
                return this;
            }
        }

        public Target build() {
            ImmutableList<String> edges = this.edges == null ? ImmutableList.of() : this.edges.build();
            List points;
            switch (this.points == null ? 0 : this.points.size()) {
                case 0:
                    points = Collections.emptyList();
                    break;
                case 1:
                    points = Collections.singletonList((String)this.points.get(0));
                    break;
                default:
                    points = Collections.unmodifiableList(new ArrayList(this.points));
            }

            return new Target(edges, points);
        }
    }
}
```

### MapStruct Version

1.5.3.Final

sranka:
Closes #3129

The code herein fixes the detection of collection adders to use a more generic `Iterable` as a setter argument (or a getter return value). As consequence, a lombok generated builder setter (fluent) method:
```
public LombokImmutablePathBuilder edges(Iterable<? extends String> edges) {
  // ...
  return this;
}
```
is now also considered as a candidate for searching of an associated adder:
```
public LombokImmutablePathBuilder addEdge(String edge) {
  // ...
  return this;
}
```

Using the adders directly has a performance benefit when using lombok builders. The mapper implementation then does not need to create intermediary ArrayList values to map collection properties. For example, in place of this code (before this PR):

```java
...
ActionRule.ActionRuleBuilder actionRule = ActionRule.builder();
if (actionRules.getCostList() != null) {
    actionRule.costs(this.costListToCostIterable(actionRules.getCostList()));
}
...
protected Iterable<Cost> costListToCostIterable(List<b.Cost> list) {
    if (list == null) {
        return null;
    } else {
        ArrayList<Cost> iterable = new ArrayList(list.size());
        Iterator var3 = list.iterator();

        while(var3.hasNext()) {
            com.goeuro.search2.model.proto.Cost cost = (com.goeuro.search2.model.proto.Cost)var3.next();
            iterable.add(this.costMapper.map(cost));
        }

        return iterable;
    }
}
```

, the new mapper implementation is then optimal (this PR):

```java
ActionRule.ActionRuleBuilder actionRule = ActionRule.builder();
if (actionRules.getCostList() != null) {
    Iterator var3 = actionRules.getCostList().iterator();

    while(var3.hasNext()) {
        Cost costList = (Cost)var3.next();
        actionRule.addCost(this.costMapper.map(costList));
    }
}
```

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/builder/lombok/LombokBuilderAccessorNamingStrategy.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/builder/lombok/LombokBuilderAccessorNamingStrategy.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.builder.lombok;
+
+import org.mapstruct.ap.spi.DefaultAccessorNamingStrategy;
+
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.ExecutableElement;
+import java.util.List;
+
+/**
+ * LombokBuilderAccessorNamingStrategy is an extension of a default naming strategy
+ * that recognizes lombok mapstruct adders defined by a default @lombok.Singular annotation.
+ * Lombok adders don't have the 'add' prefix OOTB, an adder is herein recognized by
+ * the presence of another builder method with an 's' suffix.
+ *
+ * @see <a href="https://projectlombok.org/features/BuilderSingular">Lombok @Singular</a>
+ */
+public class LombokBuilderAccessorNamingStrategy extends DefaultAccessorNamingStrategy {
+
+    private boolean isLombokBuilderAdder(ExecutableElement method) {
+        // check that the method is defined in a lombok Builder class
+        Element enclosingElement = method.getEnclosingElement();
+        if (enclosingElement == null || !enclosingElement.getSimpleName().toString().endsWith( "Builder" )) {
+            return false;
+        }
+        Element parentElement = enclosingElement.getEnclosingElement();
+        if (parentElement == null || (enclosingElement.getKind() != ElementKind.CLASS
+                && !enclosingElement.getKind().name().equals( "RECORD" ))) {
+            return false;
+        }
+        // check if there is a plural setter for this adder in the builder
+        List<? extends Element> enclosedElements = enclosingElement.getEnclosedElements();
+        String plural1 = method.getSimpleName().toString() + "s";
+        for (Element enclosedElement : enclosedElements) {
+            if (!(enclosedElement instanceof ExecutableElement)) {
+                continue;
+            }
+            ExecutableElement setterMethod = (ExecutableElement) enclosedElement;
+            String name = enclosedElement.getSimpleName().toString();
+            if (plural1.equals( name )) {
+                if (setterMethod.getParameters().size() != 1) {
+                    continue;
+                }
+                return true;
+            }
+        }
+        return false;
+    }
+
+    protected boolean isFluentSetter(ExecutableElement method) {
+        return super.isFluentSetter( method ) &&
+                !isLombokBuilderAdder( method );
+    }
+
+    public boolean isAdderMethod(ExecutableElement method) {
+        return super.isAdderMethod( method ) || isLombokBuilderAdder( method );
+    }
+}
--- a/processor/src/test/java/org/mapstruct/ap/test/builder/lombok/PathMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/builder/lombok/PathMapper.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.builder.lombok;
+
+import org.mapstruct.CollectionMappingStrategy;
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.Mappings;
+
+@Mapper(collectionMappingStrategy = CollectionMappingStrategy.ADDER_PREFERRED)
+public interface PathMapper {
+
+    @Mappings({
+        @Mapping(target = "edges", source = "edges"),
+    })
+    LombokImmutablePath map(MutablePath source);
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -898,7 +898,7 @@ public class Type extends ModelElement implements Comparable<Type> {
 
         List<Accessor> candidates;
 
-        if ( collectionProperty.isCollectionType() ) {
+        if ( collectionProperty.isIterableType() ) {
             candidates = getAccessorCandidates( collectionProperty, Iterable.class );
         }
         else if ( collectionProperty.isStreamType() ) {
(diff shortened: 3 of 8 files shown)


processor/src/test/java/org/mapstruct/ap/test/builder/lombok/LombokBuilderAccessorNamingStrategy.java
filiphr, line 1: I don't understand the use of this class. There is an easy workaround for this if you use `@Singular("add<singular")`. If we are going in the territory of supporting `@Singular` properly then we need to also support something like `@Singular("custom")`. 

I would prefer that we do not go in that direction and only perhaps support adders for iterables as well.

processor/src/test/java/org/mapstruct/ap/test/builder/lombok/PathMapper.java
```diff
@@ -0,0 +1,20 @@
...
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.Mappings;
+
+@Mapper(collectionMappingStrategy = CollectionMappingStrategy.ADDER_PREFERRED)
+public interface PathMapper {
+
+    @Mappings({
+        @Mapping(target = "edges", source = "edges"),
+    })
```
filiphr, line 18: This is obsolete. MapStruct is going to map this implicitly

filiphr (changes requested):
I've left a comment on the `LombokBuilderAccessorNamingStrategy`. I think that the solution for this is fine (i.e. support adder methods for iterables). However, I would like to decouple this from Lombok entirely.

Can you also simplify the test classes a bit. This can be easily reproduced by having a class that looks like:

```java
public class ImnutablePath {
    private final List<String> edges = new ArrayList<>();

    public Iterable<String> getEdges() {
        return edges;
    }

    public void setEdges(Iterable<String> edges) {
        this.edges.clear();
        for(String edge: edges) {
            this.edges.add(edge.toUpperCase());
        }
    }

    public void addEdge(String edge) {
        this.edges.add(edge);
    }
}
```

The tests should also be moved in the `org.mapstruct.ap.test.collection.adder.AdderTest`

filiphr:
@sranka you don't need to do anything on this PR. I believe that PR #3166 is going to solve the problem

sranka:
@filiphr yeah, it looks similar, closing this PR

PR #3166, opened 2023-02-17
#3165: support for array to collection with adder preferred

### LINKED ISSUE #3129: CollectionMappingStrategy.ADDER_PREFERRED is not supported for Iterables

### LINKED ISSUE #3165: Array to List mapping with adder fails
### Expected behavior

I would expect that given 
```java
public class User {
    private String id;
    private String email;
    private Permission[] permissions;
    // more code omitted for brevity
}
```
I'd be able to get a mapper to map the Java POJO to
```
message UserDTO {
    string id = 1;
    string email = 2;
    repeated PermissionDTO permissions = 3;
// more code omitted for brevity
}
```
Which  would map the `permission` array to the proto `permissionsList` through the adder pattern.

### Actual behavior

The auto-code generation fails because the generated code tries to directly add the whole array instead of iterating over it, map each value on its own and add it to the list.

I modified the public Protobuff example so it fails
https://github.com/eroznik/mapstruct-examples/commit/f4c3a4eae7a678b8ba83820bdee4532f6d02b622

The pushed code is using the latest snapshot, mostly because I was using the latest snapshot version while testing it out locally. Keep in mind that the linked commit won't build.

### MapStruct Version

All versions that I tested.

eroznik:
This PR offers a potential fix for the problem that I described in #3165 .

Fixes #3165 
Fixes #3129

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
@@ -42,8 +42,11 @@ public class AdderWrapper extends AssignmentWrapper {
         if ( getSourceType().isCollectionType() ) {
             adderType = first( getSourceType().determineTypeArguments( Collection.class ) );
         }
-        else { // array
-            adderType =  getSourceType().getComponentType();
+        else if ( getSourceType().isArrayType() ) {
+            adderType = getSourceType().getComponentType();
+        }
+        else { // iterable
+            adderType = first( getSourceType().determineTypeArguments( Iterable.class ) );
         }
     }
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
@@ -39,7 +39,12 @@ public class AdderWrapper extends AssignmentWrapper {
         // localVar is iteratorVariable
         String desiredName = Nouns.singularize( adderIteratorName );
         rhs.setSourceLoopVarName( rhs.createUniqueVarName( desiredName ) );
-        adderType = first( getSourceType().determineTypeArguments( Collection.class ) );
+        if ( getSourceType().isCollectionType() ) {
+            adderType = first( getSourceType().determineTypeArguments( Collection.class ) );
+        }
+        else { // array
+            adderType =  getSourceType().getComponentType();
+        }
     }
 
     @Override
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/SourceRHS.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/SourceRHS.java
@@ -149,6 +149,9 @@ public class SourceRHS extends ModelElement implements Assignment {
             else if ( sourceType.isStreamType() ) {
                 return first( sourceType.determineTypeArguments( Stream.class ) );
             }
+            else if ( sourceType.isArrayType() ) {
+                return sourceType.getComponentType();
+            }
         }
         return sourceType;
     }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/SourceRHS.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/SourceRHS.java
@@ -152,6 +152,9 @@ public class SourceRHS extends ModelElement implements Assignment {
             else if ( sourceType.isArrayType() ) {
                 return sourceType.getComponentType();
             }
+            else if ( sourceType.isIterableType() ) {
+                return first( sourceType.determineTypeArguments( Iterable.class ) );
+            }
         }
         return sourceType;
     }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -523,7 +523,7 @@ public class PropertyMapping extends ModelElement {
             Assignment result = rightHandSide;
 
             String adderIteratorName = sourcePropertyName == null ? targetPropertyName : sourcePropertyName;
-            if ( result.getSourceType().isCollectionType() ) {
+            if ( result.getSourceType().isIterableType() ) {
                 result = new AdderWrapper( result, method.getThrownTypes(), isFieldAssignment(), adderIteratorName );
             }
             else if ( result.getSourceType().isStreamType() ) {
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_3165/Issue3165Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_3165/Issue3165Mapper.java
@@ -21,21 +21,29 @@ public interface Issue3165Mapper {
 
     class Source {
         private String[] pets;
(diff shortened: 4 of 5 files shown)


filiphr:
This fix looks OK. However, there are tests missing. If you would like for us to accept this PR please provide tests. 

By the way the support is not really linked to protobuf. It is rather linked to not being able to map an array to a collection / iterable using an adder. It would be good to add tests like that (not with protobuf).

eroznik:
@filiphr i've just added a test case for the issue, lmk if thats what you expected.. (:

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
```diff
@@ -39,7 +39,12 @@ public AdderWrapper( Assignment rhs,
         // localVar is iteratorVariable
         String desiredName = Nouns.singularize( adderIteratorName );
         rhs.setSourceLoopVarName( rhs.createUniqueVarName( desiredName ) );
-        adderType = first( getSourceType().determineTypeArguments( Collection.class ) );
+        if ( getSourceType().isCollectionType() ) {
+            adderType = first( getSourceType().determineTypeArguments( Collection.class ) );
+        }
+        else { // array
+            adderType =  getSourceType().getComponentType();
+        }
```
filiphr, line 47: This won't work if the source type is iterable i.e. `Iterable<String>`. Can you please add a test case for this as well?
```suggestion
        if ( getSourceType().isArrayType() ) {
            adderType  =  getSourceType().getComponentType();
        }
        else { // iterable
            adderType = first( getSourceType().determineTypeArguments( Collection.class ) );
            adderType =  getSourceType().getComponentType();
        }
```

processor/src/main/java/org/mapstruct/ap/internal/model/common/SourceRHS.java
```diff
@@ -149,6 +149,9 @@ public Type getSourceTypeForMatching() {
             else if ( sourceType.isStreamType() ) {
                 return first( sourceType.determineTypeArguments( Stream.class ) );
             }
+            else if ( sourceType.isArrayType() ) {
+                return sourceType.getComponentType();
+            }
```
filiphr, line 154: I think that we might need to have another `else if ( sourceType.isIterableType() )` here as well.

filiphr (commented):
The test you added is good. Would be nice if you can also add a test where the source is `Iterable<String>`. I think that it won't work without the changes I am mentioning in the comments

filiphr:
I think that with the addition of the other tests we are going to show that this is also going to fix #3129 and actually make PR #3130 obsolete

processor/src/main/java/org/mapstruct/ap/internal/model/common/SourceRHS.java
eroznik: you are correct :+1:

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/AdderWrapper.java
eroznik: added

eroznik:
Added test case for iterable mapping + the required code changes so the test passes

filiphr:
Thanks @eroznik

PR #3204, opened 2023-03-18
Improve line location report for invalid qualifier for SubclassMapping

### LINKED ISSUE #3202: Improve line location report for invalid qualifier for SubclassMapping
In PR #3120 we added support for `qualifiedBy` and `qualifiedByName`. Our generic logic seems like it is putting the error line location on the method element itself. We should instead use the `@SubclassMapping` element for reporting the error.

This issue is a good issue to get started with MapStruct. If the original creator of the PR @EvaristeGalois11 is not interested in adding this, then I would invite someone from the community to give it a go.

EvaristeGalois11:
Fix #3202 

I had to remove one of the `@SubclassMapping` annotations from the erroneous mappers because otherwise the tests would fail.

It seems that if more than one annotations of the same type as `AnnotationMirror` are present on the mapping method the compiler will fallback to emit the error on the method itself. Eclipse compiler seems to work fine regardless of the number of annotations. Is this an already known issue?

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
@@ -7,7 +7,10 @@ package org.mapstruct.ap.internal.model.source;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
+import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.type.TypeMirror;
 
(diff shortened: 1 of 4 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/SubclassMappingOptions.java
```diff
@@ -34,14 +36,16 @@ public class SubclassMappingOptions extends DelegatingOptions {
     private final TypeMirror target;
     private final TypeUtils typeUtils;
     private final SelectionParameters selectionParameters;
+    private final SubclassMappingGem subclassMappingGem;
```
filiphr, line 39: Let's name this `subclassMapping`. We are not adding the `Gem` prefix to fields and variables. e.g. have a look at `MappingOptions`

filiphr (changes requested):
Thanks @EvaristeGalois11. The PR looks great, only one small cosmetic comment regarding the prefix `Gem` in fields and variables. We are usually not using it in other places, e.g. `MappingOptions`, `BeanMappingOptions`.

>It seems that if more than one annotations of the same type as `AnnotationMirror` are present on the mapping method the compiler will fallback to emit the error on the method itself. Eclipse compiler seems to work fine regardless of the number of annotations. Is this an already known issue?

This is indeed a known issue. I think that newer Java compiler versions have fixed this. The eclipse compiler always worked fine. You don't need to remove anything from the code you can use the `alternativeLine` in the `@Diagnostic`. You can actually remove in one place and leave in the other, to make sure that we really do report in the right place in both cases. As an example with this similar problem is `AnnotateWithTest#mapperWithIdenticalAnnotationRepeated`

EvaristeGalois11:
Thank you @filiphr for your feedback! I have updated the PR with what you said.

I had to change the `copyForInverseInheritance` method a bit because i couldn't find a combination of variables name that wasn't awkard, I hope it isn't a problem.

filiphr:
Thanks @EvaristeGalois11

PR #3207, opened 2023-03-20
add kora framework support

slutmaker:
Recently, a lightweight and fast compile–time framework has appeared - [kora](https://github.com/Tinkoff/kora). Unfortunately, it can't recognize @Singleton and other annotations from Jakarta. So this pull request adds a new type of component model - **kora**.

### DIFF
--- a/core/src/main/java/org/mapstruct/MappingConstants.java
+++ b/core/src/main/java/org/mapstruct/MappingConstants.java
@@ -149,6 +149,12 @@ public final class MappingConstants {
          */
         public static final String JAKARTA_CDI = "jakarta-cdi";
 
+        /**
+         * The generated mapper is a Kora component
+         *
+         */
+        public static final String KORA = "kora";
+
     }
 
 }
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -102,6 +102,14 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- Kora -->
+        <dependency>
+            <groupId>ru.tinkoff.kora</groupId>
+            <artifactId>common</artifactId>
+            <version>0.11.6</version>
+            <scope>test</scope>
+        </dependency>
+
         <!-- Spring -->
         <dependency>
             <groupId>org.springframework</groupId>
@@ -286,6 +294,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
+                    <testSource>17</testSource>
+                    <testTarget>17</testTarget>
                     <annotationProcessorPaths>
                         <path>
                             <groupId>org.mapstruct.tools.gem</groupId>
--- a/processor/src/main/java/org/mapstruct/ap/internal/gem/MappingConstantsGem.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/gem/MappingConstantsGem.java
@@ -53,6 +53,8 @@ public final class MappingConstantsGem {
         public static final String JAKARTA = "jakarta";
 
         public static final String JAKARTA_CDI = "jakarta-cdi";
+
+        public static final String KORA = "kora";
      }
 
 }
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/KoraComponentProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/KoraComponentProcessor.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.internal.processor;
+
+import org.mapstruct.ap.internal.gem.MappingConstantsGem;
+import org.mapstruct.ap.internal.model.Annotation;
+import org.mapstruct.ap.internal.model.Mapper;
+
+import java.util.Collections;
+import java.util.List;
+
+public class KoraComponentProcessor extends AnnotationBasedComponentModelProcessor {
+
+    @Override
+    protected String getComponentModelIdentifier() {
+        return MappingConstantsGem.ComponentModelGem.KORA;
+    }
+
+    @Override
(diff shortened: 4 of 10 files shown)


filiphr:
Thanks for this PR @AbstractCoderX. From what I can see there are no tests for this. In addition to that I see that this is relatively new framework (first release is less than a 6 months ago). Their documentation is also not in English, so I can't really understand how it actually works.

>Unfortunately, it can't recognize @singleton and other annotations from Jakarta

I think that a better approach would be to try and get the kora team to support the jakarta-inject API. That is the most resilient approach and easiest to integrate with other frameworks

AbstractCoderX:
> I think that a better approach would be to try and get the kora team to support the jakarta-inject API. That is the most resilient approach and easiest to integrate with other frameworks

Kora reacts only to its @Component annotation. They do not want to support jakarta-inject, since the Kora Framework has a completely different api: there is no alternatives for @Inejct, @Typed, string qualifiers, etc.

filiphr:
@AbstractCoderX if there is no alternative to `@Inject` what's the point of using a custom component model? The idea of the component model in MapStruct is to be able to inject mappers and services defined via `Mapper#uses`. If kore does not have a concept of `@Inject` then it makes no sense to support it from our side.

Looking at your contribution it seems like you can easily just use the upcoming `@AnnotateWith(Component.class)` to annotate the class. Also the decorator won't work. 

Please provide some test cases so that we can see how this would work

AbstractCoderX:
> @AbstractCoderX if there is no alternative to `@Inject` what's the point of using a custom component model? The idea of the component model in MapStruct is to be able to inject mappers and services defined via `Mapper#uses`. If kore does not have a concept of `@Inject` then it makes no sense to support it from our side.

Kora has a dependency injection mechanism, only the api is different, there is no alternative of the @Inject annotation, it is not needed in kora, it uses only a constructor for dependency injection.



> Looking at your contribution it seems like you can easily just use the upcoming `@AnnotateWith(Component.class)` to annotate the class. Also the decorator won't work.

`@AnnotateWith(Component.class)` will help me, but I did not find a release where this annotation is present, 1.6 has not been released yet.

filiphr:
>Kora has a dependency injection mechanism, only the api is different, there is no alternative of the @Inject annotation, it is not needed in kora, it uses only a constructor for dependency injection.

MapStruct supports both constructor and setter injection. This means that kora only works with constructor injection. 

I am a bit reluctant to incorporate such a new framework into the core of MapStruct. This can also be part of another module and maintained by the kora team. All the other dependency injection frameworks support JSR 330, which only has 6 annotations:

* `@Inject`
* `@Named`
* `@Provider`
* `@Qualifier`
* `@Scope`
* `@Singleton`

I really do not understand why the kora team doesn't just map those to their own. Micronaut and Spring also have their own annotations, but they also support JSR 330. 

Can you please point me to an issue in the kora issue tracker where they are rejecting the JSR 330 support?

AbstractCoderX:
> MapStruct supports both constructor and setter injection. This means that kora only works with constructor injection.
> 
> I am a bit reluctant to incorporate such a new framework into the core of MapStruct. This can also be part of another module and maintained by the kora team. All the other dependency injection frameworks support JSR 330, which only has 6 annotations:
> 
>     * `@Inject`
> 
>     * `@Named`
> 
>     * `@Provider`
> 
>     * `@Qualifier`
> 
>     * `@Scope`
> 
>     * `@Singleton`
> 
> 
> I really do not understand why the kora team doesn't just map those to their own. Micronaut and Spring also have their own annotations, but they also support JSR 330.

The bottom line is that the core has a different philosophy. Spring, Quarkus, Micronaft are aimed at backward compatibility with old solutions and ease of transition from them. In kora, dependency graph construction happens at compile time, unlike micronaut/quarkus. There are no string qualifiers in kora, instead tags are used with a class literal. There are no scopes for components in the kora, all components are singleton. Most of the annotations from JSR 330 just don't fit into the philosophy of the kora, so they don't want to support them.



> Can you please point me to an issue in the kora issue tracker where they are rejecting the JSR 330 support?

I communicated with them about this in their telegram chat, not in the github/jira issue.


I'm in the process of implementing the tests now but ran into some problems, I'll write to gitter for help.

filiphr:
>The bottom line is that the core has a different philosophy. Spring, Quarkus, Micronaut are aimed at backward compatibility with old solutions and ease of transition from them. In kora, dependency graph construction happens at compile time, unlike micronaut/quarkus.

Micronaut also creates the dependency graph during compile time. There is also [Dagger](https://dagger.dev/), which is a fully static compile-time dependency injection framework for Java, Kotlin and Android. It also supports the JSR 330 annotations. I don't know how Kora works, but I really do not see why it can be a problem to support some of those annotations. They can use the annotations during compile time, it doesn't mean that if you use JSR 330 you have to do something on runtime.

AbstractCoderX:
I've added tests, but they only test the generatedSources. Due to the fact that mapper implementations are not available at compile time to the kora annotationProcessor in the mapstruct-processor, kora can't build the graph (because it builds it at compile time). However, checking the resources is quite enough, the rest is already checking the work of the koraframework itself, which we don’t really need.

zabelovq:
any updates on this?

filiphr:
The added tests are not enough. If we are going to have support for this, I would need to see an entire example with having Kora actually running. This can be an integration test, we have integration test that work with other annotation processors (e.g. Immutables, FreeBuilder, etc.). We also need to have support for decorators.

If Kora does not have such support then MapStruct cannot provide such a component model.

slutmaker:
Kora added the mapstruct-extension module, this PR can be closed.

filiphr:
Thanks for letting us know @slutmaker. From what I can also see the kora framework has been marked as a public archive. So not sure if this means that it isn't developed anymore or it has been moved somewhere else

slutmaker:
> Thanks for letting us know @slutmaker. From what I can also see the kora framework has been marked as a public archive. So not sure if this means that it isn't developed anymore or it has been moved somewhere else

GitHub has decided to ban all opensource repositories of the Russian company Tinkoff. Kora continues to be developed, but not on github

PR #3219, opened 2023-03-28
#2987 Support for defining Javadoc in the generated mapper implementation

### LINKED ISSUE #2987: Support for defining Javadoc in the generated mapper implementation
Javadoc validation checks generated code by default and generates warnings for missing tags on generated code.
It could be resolved by copying javadoc into the generated code because simply ignoring generated code causes other issues such as missing types.

jccampanero:
The proposed change provides a new annotation allowing the definition of Javadoc comments in the generated `Mapper` implementation.

For instance, the following definition:

```java
@Javadoc(
    value = "This is the description",
    authors = { "author1", "author2" },
    deprecated = "Use {@link OtherMapper} instead",
    since = "0.1"
)
```

will generate something like:

```java
/**
* This is the description
* 
* @author author1
* @author author2
* 
* @deprecated Use {@link OtherMapper} instead
* @since 0.1
*/
```

The annotation allows users to write Text blocks:

```java
@Javadoc(
    """
    This is the description
                   
    @author author1
    @author author2
                   
    @deprecated Use {@link OtherMapper} instead
    @since 0.1
    """
)
```

or plain old Java `String`s:

```java
@Javadoc("This is the description\n"
                   + "\n"
                   + "@author author1\n"
                   + "@author author2\n"
                   + "\n"
                   + "@deprecated Use {@link OtherMapper} instead\n"
                   + "@since 0.1\n")
```

Fixes #2987.


processor/src/main/java/org/mapstruct/ap/internal/model/Javadoc.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    private final String value;
+    private final List<String> authors;
+    private final String deprecated;
+    private final String since;
+
+    private Javadoc(String value, List<String> authors, String deprecated, String since) {
+        this.value = value;
+        this.authors = authors;
```
filiphr, line 63: We can do the defensive part of the authors here. i.e. `this.authors = authors != null ? Collections.unmodifiableList( authors ) : Collections.emptyList()`.

Doing this would simplify both the `hasAuthors` and `getAuthors`

processor/src/main/resources/org/mapstruct/ap/internal/model/GeneratedType.ftl
```diff
@@ -14,6 +14,9 @@ package ${packageName};
 import ${importedType};
 </#list>
 
+<#if hasJavadoc()>
+<#nt><@includeModel object=javadoc/>
+</#if>
```
filiphr, line 19: I don't think that we need the `hasJavadoc` method. We can do it in a similar way as it is done for the constructor.

e.g.

```
<#if javadoc??><#nt> <@includeModel object=javadoc /></#if>
```

processor/src/main/resources/org/mapstruct/ap/internal/model/Javadoc.ftl
```diff
@@ -0,0 +1,25 @@
...
+    Copyright MapStruct Authors.
+
+    Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+
+-->
+<#-- @ftlvariable name="" type="org.mapstruct.ap.internal.model.Javadoc" -->
+/**
+<#list value?split("\n") as line><#nt>*<#if line?has_content> </#if>${line?trim}
+</#list>
+<#if hasAuthors()>
```
filiphr, line 12: We don't need authors we can just do `<#if !authors.isEmpty()>`

processor/src/main/resources/org/mapstruct/ap/internal/model/Javadoc.ftl
```diff
@@ -0,0 +1,25 @@
...
+*
+<#list authors as author> <#nt>* @author ${author?trim}
+</#list>
+</#if>
+<#if hasDeprecated()>
```
filiphr, line 17: We don't need the presence check we can do `<#if deprecated?has_content>`

processor/src/main/resources/org/mapstruct/ap/internal/model/Javadoc.ftl
```diff
@@ -0,0 +1,25 @@
...
+*
+<#nt>* @deprecated ${deprecated?trim}
+</#if>
+<#if hasSince()>
```
filiphr, line 21: We don't need the presence check we can do `<#if since?has_content>`

processor/src/test/java/org/mapstruct/ap/test/javadoc/JavadocTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.ProcessorTest;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.GeneratedSource;
+
+/**
+ * @author Jose Carlos Campanero Ortiz
+ */
+@IssueKey("2987")
+@WithClasses(AnnotateWithEnum.class)
```
filiphr, line 20: Why is this needed?

processor/src/test/java/org/mapstruct/ap/test/javadoc/JavadocTest.java
```diff
@@ -0,0 +1,49 @@
...
+public class JavadocTest {
```
filiphr, line 21: I think that for this test it would be better if we can use fixtures to test it. This will make sure that we test the entire javadoc properly.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -441,6 +444,23 @@ else if ( method.isStreamMapping() ) {
...
     }
 
+    private Javadoc getJavadoc(TypeElement element) {
+        JavadocGem javadocGem = JavadocGem.instanceOn( element );
+
+        if ( javadocGem == null ) {
+            return null;
+        }
+
+        Javadoc javadoc = new Javadoc.Builder()
```
filiphr, line 454: Should we perhaps have an error or warning if nothing is defined? It doesn't bring much value to use `@Javadoc` without anything

processor/src/main/java/org/mapstruct/ap/internal/model/Mapper.java
```diff
@@ -128,6 +135,7 @@ public Mapper build() {
     private final boolean customPackage;
     private final boolean customImplName;
     private Decorator decorator;
+    private Javadoc javadoc;
```
filiphr, line 138: Why not `final`?

core/src/main/java/org/mapstruct/Javadoc.java
```diff
@@ -0,0 +1,20 @@
...
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
+public @interface Javadoc {
```
filiphr, line 15: Can we please add some javadoc with examples for this? Please also add javadoc for all the other methods.

core/src/main/java/org/mapstruct/Javadoc.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
```
filiphr, line 14: Why `RetentionPolicy.CLASS`?

filiphr (changes requested):
This looks really good @jccampanero. Thanks a lot for your work on this and sorry that it took me so long to review it.

I've left some comments that I think we should address before we merge this to main.

In addition to the comments. It would perhaps be good to add something in the documentation for this and to also add something in the javadoc of `@Mapper`, perhaps a `@see`.

What do you think @jccampanero

processor/src/main/resources/org/mapstruct/ap/internal/model/Javadoc.ftl
jccampanero: Thank you for all the comments related to the templates: sorry, I am afraid that my FreeMaker skills are a bit rusty.

processor/src/main/java/org/mapstruct/ap/internal/model/Javadoc.java
jccampanero: I agree with with.

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
jccampanero: That was a very good point indeed.

filiphr: Yes using the `FormattingMessenger` is the right approach for providing feedback. Have a look at something like `IterableMappingOptions#isConsistent` to see how we are doing it in other places

processor/src/main/resources/org/mapstruct/ap/internal/model/Javadoc.ftl
filiphr: No need to apologise. That's why we have PR, we all learn from each other 😉

processor/src/test/java/org/mapstruct/ap/test/javadoc/JavadocTest.java
jccampanero: Sorry, I probably copied and pasted the test skeleton from another class and I forgot to remove that code.

processor/src/test/java/org/mapstruct/ap/test/javadoc/JavadocTest.java
jccampanero: Sorry, I probably copied and pasted the test skeleton from another class and I forgot to remove that code.

core/src/main/java/org/mapstruct/Javadoc.java
jccampanero: Sorry @filiphr , I don't fully understand you here. Please, can you elaborate?

filiphr: The java compiler uses this `RetentionPolicy` to decide what to do with the annotation in the byte code. There are 3 retention policies:

* `SOURCE` - Annotations will be discarded by the compiler
* `CLASS` - Annotations will be recorded in the class file
* `RUNTIME` - Annotations will be recorded in the class file and retained by the VM at runtime

I think that for this particular annotation we can use the `SOURCE` retention policy, since we only need it when generating the code for a particular mapper.

jccampanero: Obvious, sorry @filiphr.

core/src/main/java/org/mapstruct/Javadoc.java
jccampanero: Of course I have provided an initial implementation. I'm sorry, I should have included the necessary comments from the beginning.

jccampanero:
> This looks really good @jccampanero. Thanks a lot for your work on this and sorry that it took me so long to review it.
> 
> I've left some comments that I think we should address before we merge this to main.
> 
> In addition to the comments. It would perhaps be good to add something in the documentation for this and to also add something in the javadoc of `@Mapper`, perhaps a `@see`.
> 
> What do you think @jccampanero

Thank you very much @filiphr.

There is no need to apologize; on the contrary, I am sorry for the time in providing a pull request.

In addition, thank you very much for your time and for reviewing the code.

I have been working in the changes you proposed: only the tests are still pending, and probably improving the Javadoc comments of the `Javadoc` annotation.

Of course, I agree with you: I will include a reference to the new functionality in the Javadoc comments of `Mapper` and in the library documentation. What section of the documentation do you think would be best suited to define the new functionality?

sjaakd:
Nice feature 👍

processor/src/main/java/org/mapstruct/ap/internal/processor/MapperCreationProcessor.java
```diff
@@ -810,4 +830,15 @@ private void reportErrorWhenNonMatchingName(SourceMethod onlyCandidate, SourceMe
             onlyCandidate.getName()
         );
     }
+
+    private boolean isConsistent( JavadocGem gem, TypeElement element, FormattingMessager messager ) {
+        if ( !gem.value().hasValue()
+            && !gem.authors().hasValue()
+            && !gem.deprecated().hasValue()
+            && !gem.since().hasValue() ) {
+            messager.printMessage( element, Message.JAVADOC_NO_ELEMENTS );
```
filiphr, line 839: It would be good to pass the `AnnotationMirror` for the javadoc here as well. e.g.
```java
messager.printMessage( element, gem.mirror(), Message.JAVADOC_NO_ELEMENTS );
```

filiphr (commented):
The comments in the Javadoc look good to me.

I've left one small comment for the printing of the error message if something is missing. You could also add an erroneous test for that as well.

For the documentation, I think that we can add an additional section to `chapter-3-defining-a-mapper.asciidoc` below the "Adding annotations" section, i.e. at the end.

filiphr:
Thanks for the changes @jccampanero. I've pushed one small polishing commit for the tests. Can you please add some docs and then we can merge this?

jccampanero:
Thank you very much for the changes performed in the tests @filiphr.

Of course, I will update the documentation to include the `Javadoc` functionality.

Can I update `JavadocTest` to use fixtures as you asked me to do too?

filiphr:
>Can I update `JavadocTest` to use fixtures as you asked me to do too?

I forgot about this. Yes please go ahead and update that test to use fixtures.
I did the polishing thinking that I'll merge it immediately, but then I saw that the documentation was not there yet.

jccampanero:
Thank you very much @filiphr.

Yes, sorry, I have been busy the last week.

filiphr:
>Yes, sorry, I have been busy the last week.

No need to apologise @jccampanero. We are all doing this in our free time, as you can see I am also not the most free person  as well 😄

documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
```diff
@@ -761,3 +761,103 @@ public class MyConverterImpl implements MyConverter {
 }
 ----
 ====
+
+
+[[javadoc]]
+=== Adding Javadoc comments
+
+MapStruct provides support for defining Javadoc comments in the generated mapper implementation using the
+`org.mapstruct.Javadoc annotation.
```
filiphr, line 770:
```suggestion
`org.mapstruct.Javadoc` annotation.
```

filiphr (commented):
Only one small typo in the docs

jccampanero:
I implemented the required changes, the one related to the documentation and the other one related to the use of fixtures in the tests.

@filiphr, please, consider review carefully the fixtures provided: I think they are right but I was unable to run the tests in IntelliJ, for any reason the `fixtures` directory are not being copied to the test output folder in my local setup.

Should we create additional test cases?

jccampanero:
> > Yes, sorry, I have been busy the last week.
> 
> No need to apologise @jccampanero. We are all doing this in our free time, as you can see I am also not the most free person as well 😄

Thank you very much always for your kindness @filiphr, I really appreciate it.

processor/src/test/java/org/mapstruct/ap/test/javadoc/JavadocTest.java
jccampanero: I implemented the tests using fixtures.

As I mentioned, please, consider review carefully the fixtures provided: I think they are right but I was unable to run the tests in IntelliJ, for any reason the fixtures directory are not being copied to the test output folder in my local setup.

Thank you @filiphr.

jccampanero:
@filiphr I would like to perform an additional change before closing the PR if you consider it appropriate.

The generated Javadoc comments always start right straight the new line character, no matter how I configure the FreeMarker template. I mean, it generates:

```
/**
*
```

but I think this should be:

```
/**
 *
```

Note the space before the asterisk.

Please, can you advice me about how I should modify the FreeMarker template to deal with that situation? Thank you in advance.

jccampanero:
> @filiphr I would like to perform an additional change before closing the PR if you consider it appropriate.
> 
> The generated Javadoc comments always start right straight the new line character, no matter how I configure the FreeMarker template. I mean, it generates:
> 
> ```
> /**
> *
> ```
> 
> but I think this should be:
> 
> ```
> /**
>  *
> ```
> 
> Note the space before the asterisk.
> 
> Please, can you advice me about how I should modify the FreeMarker template to deal with that situation? Thank you in advance.

After reviewing the code I think the issue could be motivated by `org.mapstruct.ap.internal.writer.IndentationCorrectingWriter` and the way it seems to handle whitespace at the beginning of new lines.

If I'm right, I think it's preferable to use the current implementation of the code, as it generates correct Javadoc comments, which is ultimately the only important thing in this case, to avoid changing the current functionality implemented for the rest of the templates.

filiphr:
>After reviewing the code I think the issue could be motivated by `org.mapstruct.ap.internal.writer.IndentationCorrectingWriter` and the way it seems to handle whitespace at the beginning of new lines.

You are right @jccampanero. The `IndendationCorrectingWriter` is the one which is removing the whitespaces. We'll need to do something special there to properly handle this. I would say that for now it is OK how it looks like. We can create a separate issue for this if you want.

filiphr:
This has been integrated. Thanks @jccampanero

jccampanero:
You are welcome, and thank you for your support @filiphr.

Let's tackle the next issue!!

Issue #3229, opened 2023-04-04
Implement `InjectionStrategy.SETTER`


MLNW:
### Use case

It is possible that there is a dependency cycle between multiple mappers, e.g.:

```java
@Mapper(componentModel = "spring",  uses = { FooMapper.class })
public interface BarMapper {
    Bar toBo(BarDto dto);
    BarDto toDto(Bar obj) ;
}

@Mapper(componentModel = "spring",  uses = { BarMapper.class })
public interface ForMapper {
    Foo toBo(FooDto dto);
    FooDto toDto(Foo obj) ;
}
```

In that case newer versions of SpringBoot will fail at build time. This has been reported in https://github.com/mapstruct/mapstruct/issues/2741. https://github.com/mapstruct/mapstruct/issues/2741#issuecomment-1065410535 noted the preferred workaround, i.e., use `SETTER` injection instead of `FIELD` or `CONSTRUCTOR`.

Therefore, we should implement an additional `InjectionStrategy.SETTER` to work around issues like this.

### Generated Code

```java
@Mapper(componentModel = "spring",  uses = { FooMapper.class }, injectionStrategy = InjectionStrategy.SETTER)
public interface BarMapper {
    Bar toBo(BarDto dto);
    BarDto toDto(Bar obj) ;
}
```

```java
@Generated(...)
@Component
public class FooMapperImpl implements FooMapper {

    private BarMapper barMapper;

    ...

    public void setBarMapper(BarMapper barMapper) {
        this.barMapper = barMapper;
    }

}
```

### Possible workarounds

A work around is to use an external dependency described in the referenced issue: https://github.com/mapstruct/mapstruct/issues/2741#issuecomment-1067304830

### MapStruct Version

MapStruct 1.5.3.Final

filiphr:
This makes sense as an enhancement @MLNW. The MapStruct team is busy with other things. However, if someone from the community is interested in working on this issue we are more than happy to help. @MLNW in case you are interested, you have the priority for working on this

MLNW:
@filiphr I had a look through the code base and a short try of implementing this but in the end I wasn't able to find a way to add a non-mapping function to the generated mapper, as that is pretty much what needs to be done: add a `setFoo` function to the mapper.

Do you have an idea how to best accomplish this task?

MLNW:
@filiphr I have added my progress as a draft PR. Maybe that will help in guiding me to the finish line.

filiphr:
[...]

PR #3265, opened 2023-05-07
#3229: Implement InjectionStrategy.SETTER

MLNW:
This is my work in progress for #3229. So far I have been unable to add a new method on the generated mapper. I think everything else is atleast well prepared. For the last bits I could use some support.

### DIFF
--- a/documentation/src/main/asciidoc/chapter-4-retrieving-a-mapper.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-4-retrieving-a-mapper.asciidoc
@@ -102,7 +102,7 @@ A mapper which uses other mapper classes (see <<invoking-other-mappers>>) will o
 [[injection-strategy]]
 === Injection strategy
 
-When using <<using-dependency-injection,dependency injection>>, you can choose between field and constructor injection.
+When using <<using-dependency-injection,dependency injection>>, you can choose between constructor, field, or setter injection.
 This can be done by either providing the injection strategy via `@Mapper` or `@MapperConfig` annotation.
 
 .Using constructor injection
@@ -120,8 +120,12 @@ public interface CarMapper {
 The generated mapper will inject classes defined in the **uses** attribute if MapStruct has detected that it needs to use an instance of it for a mapping.
 When `InjectionStrategy#CONSTRUCTOR` is used, the constructor will have the appropriate annotation and the fields won't.
 When `InjectionStrategy#FIELD` is used, the annotation is on the field itself.
+When `InjectionStrategy#SETTER` is used the annotation is on a generated setter method.
 For now, the default injection strategy is field injection, but it can be configured with <<configuration-options>>.
 It is recommended to use constructor injection to simplify testing.
+
+When you define mappers in Spring with circular dependencies compilation may fail.
+In that case utilize the `InjectionStrategy#SETTER` strategy.
 
 [TIP]
 ====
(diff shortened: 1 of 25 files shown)


filiphr:
Thanks for your work on this @MLNW. What I would suggest is that you have a look at the `HelperMethod`. Perhaps the `Setter` can be such a method as well. Then in the `buildSetters` you can use the `SupportingMappingMethod` to create a `MappingMethod` and add that to the `Mapper` methods.  Have a look at the `CreateDecimal` to see how that is done. You'll need to override the `name` as well for this.

processor/src/main/resources/org/mapstruct/ap/internal/model/Setter.ftl
```diff
@@ -0,0 +1,10 @@
+ <#--
+
+    Copyright MapStruct Authors.
+
+    Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+
+-->
+public void ${name}( <@includeModel object=parameter/> ) {
+    this.<@includeModel object=field/> = <@includeModel object=parameter/>;
+}
```
MLNW, line 10: I'm unable to make this mapping work. The result looks like this:

```java
    public void setGenderSpringSetterMapper(  ) {
        this. = ;
    }
```

MLNW: @filiphr this should be the last big thing that needs to be fixed but for the life of me I'm not able to figure out how this is supposed to work. Some help would be greatly appreciated

filiphr:
@MLNW I had a quick look and did some polishing. Can you please have a look and let me know what you think.

The reason why it was not working was due to the fact that the parameters available for the template were coming from the `SupportingMethod` and not from the `Setter`. I've changed this a bit and I also added support for decorators

MLNW:
Thanks for the help @filiphr!

Looks good to me. If you have any other things that need changing let me know.

BTW: Should I squash everything together?

documentation/src/main/asciidoc/chapter-4-retrieving-a-mapper.asciidoc
```diff
@@ -102,7 +102,7 @@ A mapper which uses other mapper classes (see <<invoking-other-mappers>>) will o
 [[injection-strategy]]
 === Injection strategy
 
-When using <<using-dependency-injection,dependency injection>>, you can choose between field and constructor injection.
+When using <<using-dependency-injection,dependency injection>>, you can choose between constructor, field, or method injection (the latter is only supported when using Spring).
```
filiphr, line 105: Why are you saying that method inject is only supported with Spring?

filiphr (commented):
Thanks @MLNW. I've left one comment and it seems like there are some Checkstyle errors. Can you please have a look at it?

>BTW: Should I squash everything together?

No need, I squash everything when I do the merge.

documentation/src/main/asciidoc/chapter-4-retrieving-a-mapper.asciidoc
MLNW: My first implementation was only targeting Spring and there is only a test for Spring. I guess with your changes this might work for other frameworks as well. In that case a test for those should be added as well.

BTW: I called it method injection though I guess that is wrong and should be called setter injection?

filiphr: Indeed setter injection is better, thanks for changing it.

Yes my changes made it work in general. Will push some jsr330 and jakarta tests as well

PR #3267, opened 2023-05-08
#1454 Support for lifecycle methods on type being built with builders

### LINKED ISSUE #1454: Support for lifecycle methods on type being build with builders

thunderhook:
Hi,

this is my first PR for mapstruct. Disclaimer: Because of the IntelliJ Community Edition I do not have any highlighting or formatting for `.ftl` files, so bear with me. 😅 

I tried to be as consistent as possible with the naming of the lifecycle method reference collections (`beforeMappingReferencesWithFinalizedReturnType` & `afterMappingReferencesWithFinalizedReturnType`) but if you have better suggestions, please tell me.

There is one edge-case i stumbled accross. With supporting `@AfterMapping` on the targets builder and finalized object, the following combination can be done by using methods **both returning** the result:

```java
    // snippet of processor/src/test/java/org/mapstruct/ap/test/builder/lifecycle/MappingContext.java
    @AfterMapping
    public Order afterWithBuilderTargetReturningTarget(@MappingTarget Order.Builder orderBuilder) {
        invokedMethods.add( "afterWithBuilderTargetReturningTarget" );

        return orderBuilder.create();
    }

    @AfterMapping
    public Order afterWithTargetReturningTarget(@MappingTarget Order order) {
        invokedMethods.add( "afterWithTargetReturningTarget" );

        return order;
    }

```

Which results in the following mapper:

```java

public class OrderMapperImpl implements OrderMapper {

    @Override
    public Order map(OrderDto source, MappingContext context) {

        // ... removed for brevity ...

        context.afterWithBuilderTarget( source, order );
        Order target = context.afterWithBuilderTargetReturningTarget( order );
        if ( target != null ) {
            return target; // in this case it will return here
        }

        // this code will not be reached
        Order orderResult = order.create();

        context.afterWithTargetType( source, Order.class );
        context.afterWithTarget( source, orderResult );
        Order target1 = context.afterWithTargetReturningTarget( orderResult );
        if ( target1 != null ) {
            return target1;
        }

        return orderResult;
    }
```

This makes sense but might be confusing. I was wondering if MapStruct should throw an error in that case?
For me, I had to adapt the test, that `context#afterWithBuilderTargetReturningTarget` returns null so that he other `@AfterMappings` get called.

Fixes #1454
---

- [x] documentation still has to be adapted [here](https://github.com/thunderhook/mapstruct/blob/693e1ab12410ae8fbbbdb371a505ed547e74a9fc/documentation/src/main/asciidoc/chapter-12-customizing-mapping.asciidoc#L265-L265)

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -94,6 +94,9 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
     private final Type returnTypeToConstruct;
     private final BuilderType returnTypeBuilder;
     private final MethodReference finalizerMethod;
+    private final String finalizedResultName;
+    private final List<LifecycleCallbackMethodReference> beforeMappingReferencesWithFinalizedReturnType;
+    private final List<LifecycleCallbackMethodReference> afterMappingReferencesWithFinalizedReturnType;
 
     private final MappingReferences mappingReferences;
 
@@ -368,8 +371,35 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
             MethodReference finalizeMethod = null;
 
+            List<LifecycleCallbackMethodReference> beforeMappingReferencesWithFinalizedReturnType = new ArrayList<>();
+            List<LifecycleCallbackMethodReference> afterMappingReferencesWithFinalizedReturnType = new ArrayList<>();
             if ( shouldCallFinalizerMethod( returnTypeToConstruct ) ) {
                 finalizeMethod = getFinalizerMethod();
+
+                Type actualReturnType = method.getReturnType();
+
+                beforeMappingReferencesWithFinalizedReturnType.addAll( filterMappingTarget(
+                    LifecycleMethodResolver.beforeMappingMethods(
+                        method,
+                        actualReturnType,
+                        selectionParameters,
+                        ctx,
+                        existingVariableNames
+                    ),
+                    false
+                ) );
+
+                afterMappingReferencesWithFinalizedReturnType.addAll( LifecycleMethodResolver.afterMappingMethods(
+                    method,
+                    actualReturnType,
+                    selectionParameters,
+                    ctx,
+                    existingVariableNames
+                ) );
+
+                // remove methods without parameters as they are already being invoked
+                removeMappingReferencesWithoutSourceParameters( beforeMappingReferencesWithFinalizedReturnType );
+                removeMappingReferencesWithoutSourceParameters( afterMappingReferencesWithFinalizedReturnType );
             }
 
             return new BeanMappingMethod(
@@ -383,10 +413,16 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 returnTypeBuilder,
                 beforeMappingMethods,
                 afterMappingMethods,
+                beforeMappingReferencesWithFinalizedReturnType,
+                afterMappingReferencesWithFinalizedReturnType,
                 finalizeMethod,
                 mappingReferences,
                 subclasses
             );
+        }
+
+        private void removeMappingReferencesWithoutSourceParameters(List<LifecycleCallbackMethodReference> references) {
+            references.removeIf( r -> r.getSourceParameters().isEmpty() && r.getReturnType().isVoid() );
         }
 
         private boolean doesNotAllowAbstractReturnTypeAndCanBeConstructed(Type returnTypeImpl) {
@@ -706,7 +742,6 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
          * Find a factory method for a return type or for a builder.
          * @param returnTypeImpl the return type implementation to construct
          * @param @selectionParameters
-         * @return
          */
         private void initializeFactoryMethod(Type returnTypeImpl, SelectionParameters selectionParameters) {
             List<SelectedMethod<SourceMethod>> matchingFactoryMethods =
@@ -1380,7 +1415,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
          * <p>
          * When a target property matches its name with the (nested) source property, it is added to the list if and
          * only if it is an unprocessed target property.
-         *
+         * <p>
          * duplicates will be handled by {@link #applyPropertyNameBasedMapping(List)}
          */
         private void applyTargetThisMapping() {
@@ -1766,6 +1801,8 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                               BuilderType returnTypeBuilder,
                               List<LifecycleCallbackMethodReference> beforeMappingReferences,
                               List<LifecycleCallbackMethodReference> afterMappingReferences,
+                              List<LifecycleCallbackMethodReference> beforeMappingReferencesWithFinalizedReturnType,
+                              List<LifecycleCallbackMethodReference> afterMappingReferencesWithFinalizedReturnType,
                               MethodReference finalizerMethod,
                               MappingReferences mappingReferences,
                               List<SubclassMapping> subclassMappings) {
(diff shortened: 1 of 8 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1055,7 +1089,7 @@ private boolean handleDefinedNestedTargetMapping(Set<String> handledTargets, Typ
                 .existingVariableNames( existingVariableNames )
                 .build();
 
-            unprocessedSourceParameters.removeAll( holder.getProcessedSourceParameters() );
+            holder.getProcessedSourceParameters().forEach( unprocessedSourceParameters::remove );
```
filiphr, line 1092: Why is this changed?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1830,6 +1869,18 @@ public List<SubclassMapping> getSubclassMappings() {
         return subclassMappings;
     }
 
+    public String getFinalizedResultName() {
+        return getResultName() + "Result";
+    }
```
filiphr, line 1874: I would initialize the `finalizedResultName` in the constructor. Otherwise, this can potentially lead to compiler errors due to the name already being used.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -339,6 +341,33 @@ else if ( !method.isUpdateMethod() ) {
...
+                afterMappingReferencesWithFinalizedReturnType.addAll( LifecycleMethodResolver.afterMappingMethods(
+                    method,
+                    actualReturnType,
+                    selectionParameters,
+                    ctx,
+                    existingVariableNames
+                ) );
+                removeMappingReferencesWithoutSourceParameters(afterMappingReferencesWithFinalizedReturnType);
+            }
+
```
filiphr, line 370: Doesn't it make more sense to initialize this in the same place where we create the `finalizeMethod` (i.e. line 401)? I think that in that case we won't need to do the if checks that are here, since we know we only need those lifecycles when using builders

filiphr (commented):
This looks really good @thunderhook. I've left some comments in line. Regarding your remarks:

>There is one edge-case i stumbled accross. With supporting `@AfterMapping` on the targets builder and finalized object, the following combination can be done by using methods both returning the result:

I think that it is fine to have it as is now. We have clearly documented how things work when something is returned from a lifecycle method, so it is completely valid to use it like that if someone wants to do it. Keep in mind that `afterWithBuilderTargetReturningTarget` in our example always returns an instance, but in an actual implementation you might return an instance based on certain logic in your own method. Adapting the test for us is fine.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
thunderhook: IntelliJ shows a warning: `Call to 'set.removeAll(list)' may work slowly` therefore I applied the quick fix to prevent a possible performance problem. I wasn't aware of this until that moment. More information here: https://stackoverflow.com/q/28671903/1213544

I honestly don't know why I only fixed this warning, especially because that part of the code is unrelated to this ticket. Therefore I reverted it.

thunderhook:
Checkstyle failed because of the 2000 line limit in `BeanMappingMethod` . Therefore I joined some lines. If you have another idea what could be extracted, please let me know (I thought of the inner `Builder` class but it needs too many internals (private fields/classes) of this class)

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
thunderhook: Yes you're right. I had a bad feeling putting this in the getter.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
filiphr: Interesting. I didn't know about this. The problem there is the fact that the `holder` returns a `List<Parameter>` which doesn't make much sense to be honest. It can be a `Collection` (and then the problem goes away). The thing is anyways that this parameters are the source parameters of a mapping method, so I doubt that we will ever go in the range where this can become a problem. Seems like a new warning from IntelliJ though, I haven't seen it before.

We can fix it in a different PR. I prefer like that, because it makes it easier to review the code and to focus on the main functionality in a single PR

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1783,9 +1798,13 @@ private BeanMappingMethod(Method method,
         this.propertyMappings = propertyMappings;
         this.returnTypeBuilder = returnTypeBuilder;
         this.finalizerMethod = finalizerMethod;
+        this.finalizedResultName = getResultName() + "Result";
```
filiphr, line 1801: We are still having the problem where there might be a clash. I think what you need to do is 
```
if ( this.finalizerMethod != null ) {
    this.finalizedResultName = Strings.getSafeVariableName( getResultName() + "Result", existingVariableNames )
    existingVariableNames.add( this.finalizedResultName );
}
else {
    this.finalizedResultName = null;
}
```

filiphr (changes requested):
>Checkstyle failed because of the 2000 line limit in `BeanMappingMethod`

merging some the lines looks OK. Perhaps we need to refactor some things there (or increase this limit of 2000).

I don't like the wrappings in `forEach` and `Optional.ofNullable` though. It creates extra classes. Let's increase the line limit (I am also adding some new things to that class now for something else, so it would fail anyways)

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
thunderhook: Didn't know about this method. Thank you, I will have a look.

thunderhook:
> > Checkstyle failed because of the 2000 line limit in `BeanMappingMethod`
> 
> merging some the lines looks OK. Perhaps we need to refactor some things there (or increase this limit of 2000).
> 
> I don't like the wrappings in `forEach` and `Optional.ofNullable` though. It creates extra classes. Let's increase the line limit (I am also adding some new things to that class now for something else, so it would fail anyways)

Yeah, you're right. I will revert the changes to make life easier when you rebase

thunderhook:
I increased the `FileLength` limit to 2500 to get the pipeline green. Feel free to adapt it as you wish afterwards.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -383,12 +411,18 @@ else if ( !method.isUpdateMethod() ) {
...
+                beforeMappingReferencesWithFinalizedReturnType,
+                afterMappingReferencesWithFinalizedReturnType,
                 finalizeMethod,
                 mappingReferences,
                 subclasses
             );
         }
 
+        private void removeMappingReferencesWithoutSourceParameters(List<LifecycleCallbackMethodReference> references) {
+            references.removeIf( r -> r.getSourceParameters().isEmpty() && r.getReturnType().isVoid() );
```
filiphr, line 423: Let's add a comment why we are doing this. The reason why this is done is because lifecycle methods without any parameters are going to be already invoked. Mostly for future us, when we scratch our head why this is being done 😄

filiphr (commented):
Looks really good @thunderhook.

Two last things that I just remembered about. One I commented inline and the other on is the documetation. In the [Mapping customization with before-mapping and after-mapping methods](https://mapstruct.org/documentation/stable/reference/html/#customizing-mappings-with-before-and-after) we have a not mentioning the fact that `@AfterMapping`  / `@BeforeMapping` methods are not being invoked.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
thunderhook: Thank you for explaining why I did this. I already already forgot the reason 2 weeks after implementing it... 😆  I added a comment.

thunderhook:
> ... the other on is the documetation. In the [Mapping customization with before-mapping and after-mapping methods](https://mapstruct.org/documentation/stable/reference/html/#customizing-mappings-with-before-and-after) we have a not mentioning the fact that `@AfterMapping` / `@BeforeMapping` methods are not being invoked.

I'm not sure if I understood you correctly. I tried to tackle the documentation part and replaced the **Important:** with a NOTE-block going into detail with builders. Have a look at the MR.

Documenting and formulating is definitely not my forte, I tried my best. 😅

filiphr (approved):
The docs are perfect. Thanks a lot. I'll merge it once the builds are done

PR #3276, opened 2023-05-13
Configure maven retry behavior for Windows CI pipeline

jccampanero:
I have included the following configuration property in the MAVEN_ARGS variable defined in main.yml to enable retry in Windows:

-Dmaven.wagon.http.retryHandler.count=3

The CI defined pipeline for that operating system constantly  timeouts when fetching maven dependencies: perhaps the change could solve the issue.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 env:
   SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
   SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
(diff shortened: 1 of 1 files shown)


jccampanero:
Sorry, no, I think it doesn't solve the issue.

I will close the PR. Sorry for the incoveniences.

filiphr:
Thanks for trying this out. Perhaps we need to remove Windows from the PR builds. It is really annoying that it constantly fails.

jccampanero:
You are welcome @filiphr. Yes, I agree with you, sometimes it is even frustrating. 

If you want, we may try configuring verbose mode (`-e -X`) while running the windows pipeline for one time and see if it gives us some feedback before completely close the PR. What do you thing?

Issue #3231, opened 2023-04-06
`@Default` on Java Record's constructor isn't respected


AlexElin:
### Expected behavior

Record's constructor having `@Default` is called in the generated code.

### Actual behavior

Record's constructor having `@Default` isn't called in the generated code.  
The generated code is below.

```java
class TaskMapperImpl implements TaskMapper {

    @Override
    public Task from(TaskDto testDbRecord) {
        if ( testDbRecord == null ) {
            return null;
        }

        String id = null;
        Long number = null;

        id = testDbRecord.id();
        number = testDbRecord.number();

        Task task = new Task( id, number );

        return task;
    }
}
```

### Steps to reproduce the problem

1. Configure the maven compiler plugin with the settings below (it's taken from [docs](https://mapstruct.org/documentation/installation/))
2. `mvn clean compile` for the below code snippets

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-compiler-plugin</artifactId>
    <version>3.11.0</version>
    <configuration>
        <release>19</release>
        <annotationProcessorPaths>
            <path>
                <groupId>org.mapstruct</groupId>
                <artifactId>mapstruct-processor</artifactId>
                <version>1.5.3.Final</version>
            </path>
        </annotationProcessorPaths>
    </configuration>
</plugin>
```

```java
record Task(String id, Long number) {
    @Default
    Task(String id) {
        this(id, 1L);
    }
}
```
```java
record TaskDto(String id, Long number) {
}
```
```java
@Mapper
interface TaskMapper {
    Task from(TaskDto testDbRecord);
}
```

### MapStruct Version

1.5.3.Final

filiphr:
The way we work with records is by directly using the record components, we don't even look into the available constructors. This works like that on purpose. I'll change this to an enhancement. 

Can you please explain the use case for using `@Default` in a record @AlexElin?

AlexElin:
The use case is that I have a record with multiple constructors.  
And I want a specific one with `@Default` be used by Mapstruct during the mapping.

My vision of expected behavior is inspired the way as Mapstruct works for plain Java class having `@Default`-annotated constructor ([docs](https://mapstruct.org/documentation/1.5/reference/html/))

filiphr:
Done in PR #3284. Thanks @thunderhook

PR #3284, opened 2023-05-23
#3231 prefer record constructor annotated with @Default

thunderhook:
Hi, this is a small change. Maybe this could make it into 1.6.0. 🙂


filiphr:
Thanks @thunderhook

PR #3297, opened 2023-06-01
#3296: skip default methods when determining prototype methods.

### LINKED ISSUE #3296: Ignore default and static methods in `@MapperConfig`
### Expected behavior

I have added an `@AfterMapping` method to a `@MapperConfig` interface:

```java
@MapperConfig(
        componentModel = "spring",
        uses = CommonMapper.class
)
public interface CommonMapperConfig<P extends CommonPayload, E extends CommonEntity> extends CommonMapper<P, E> {

    @AfterMapping
    default void afterMapping(@MappingTarget E entity, P payload) {
        if (entity.getNumber() > 0) {
            entity.setPositive(true);
        }
    }
}
```

### Actual behavior

However, the code throws this when compiling:

```java
C:\dev\env\experiment\cloned\mapstruct-aftermapping\src\main\java\com\example\mapstruct\mapper\CommonMapperConfig.java:21: error: @AfterMapping can only be applied to an implemented method.
    default void afterMapping(@MappingTarget E entity, P payload) {
                 ^
```

My example code is stored [here on GitHub](https://github.com/dirkniblick/mapstruct-aftermapping). If I remove the @AfterMapping method, the code compiles and generates a mapper.

At first I thought I was crazy for remembering being able to do this before, with the effect being that the method would be available to all of the extended mappers. But, if I change the version of Mapstruct from 1.5.5.Final to 1.4.2.Final, this works:

```java
@Generated(
    date = "2023-06-01T10:04:02-0500",
    comments = "version: 1.4.2.Final, compiler: IncrementalProcessingEnvironment from gradle-language-java-8.0.jar, environment: Java 17.0.2 (Oracle Corporation)"
)
@Component
public class SpecificMapperImpl implements SpecificMapper {

    @Override
    public SpecificPayload toPayload(SpecificEntity entity) {
        if ( entity == null ) {
            return null;
        }

        SpecificPayload specificPayload = new SpecificPayload();

        specificPayload.setNumber( entity.getNumber() );
        specificPayload.setName( entity.getName() );

        return specificPayload;
    }

    @Override
    public SpecificEntity toNewEntity(SpecificPayload payload) {
        if ( payload == null ) {
            return null;
        }

        SpecificEntity specificEntity = new SpecificEntity();

        specificEntity.setNumber( payload.getNumber() );
        specificEntity.setName( payload.getName() );

        afterMapping( specificEntity, payload );

        return specificEntity;
    }

    @Override
    public SpecificEntity toUpdatedEntity(SpecificPayload payload, SpecificEntity entity, EntityContext context) {
        if ( payload == null ) {
            return null;
        }

        entity.setNumber( payload.getNumber() );
        entity.setName( payload.getName() );

        afterMapping( entity, payload );

        return entity;
    }
}
```

I changed versions to 1.5.0.Final and the compiler complains again. Is this illegal now?

### MapStruct Version

1.5.5.Final

Zegveld:
closes #3296

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
@@ -124,6 +124,11 @@ public class MethodRetrievalProcessor implements ModelElementProcessor<Void, Lis
         List<SourceMethod> methods = new ArrayList<>();
         for ( ExecutableElement executable : elementUtils.getAllEnclosedExecutableElements( typeElement ) ) {
 
+            if ( executable.isDefault() ) {
+                // skip the default methods because these are not prototypes.
+                continue;
+            }
+
             ExecutableType methodType = typeFactory.getMethodType( mapperAnnotation.mapperConfigType(), executable );
             List<Parameter> parameters = typeFactory.getParameters( methodType, executable );
             boolean containsTargetTypeParameter = SourceMethod.containsTargetTypeParameter( parameters );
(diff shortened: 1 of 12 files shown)


filiphr (approved):
LGTM. Perhaps we should also ignore static methods?

Zegveld:
> LGTM. Perhaps we should also ignore static methods?

good point, adding that situation as well to this PR. Since that also causes the same error.

filiphr:
Thanks @Zegveld

PR #3298, opened 2023-06-01
Allow inheriting the mapper attribute.

thedrow:
Draft for now.


thedrow:
@filiphr I can't run the integration tests myself locally. See [log](https://gist.github.com/thedrow/64480336fe4892c0df662823b8dfe1b1).
Can you please approve the workflow?

filiphr:
What is the purpose of inheriting the mapper? Also I don't like the fact that the retention is now runtime. What is the purpose of that?

PR #3326, opened 2023-07-12
#3309 - Add `BeanMapping#unmappedSourcePolicy`

### LINKED ISSUE #3309: Add `BeanMapping#unmappedSourcePolicy`
### Discussed in https://github.com/mapstruct/mapstruct/discussions/3308

<div type='discussions-op-text'>

<sup>Originally posted by **meilleureVie** June 15, 2023</sup>
There is some case both BeanMapping#unmappedSourcePolicy and BeanMapping#unmappedTargetPolicy will be helpful.
let take an example of 2 class (class Student with more than 20 properties and StudentDto with 2 attributes just name and firstname)

So in our mapper we are looking for 2 methods :
1. a first method **Student toStudent(StudentDto studentDto)** that can create a new Student and just fill the name and the firstname.
2. a second method **StudenDto toStudentDto(Student student)** that can map the name and firstname of a student to studentDto

For me, in order to achieve that goal and also take advantage of typesafe warning, it will be easier for me to do something like this :
1. in the first case
```
@BeanMapping(unmappedTargetPolicy = ReportingPolicy.IGNORE, unmappedSourcePolicy = ReportingPolicy.WARN)
Student toStudent(StudentDto studentDto);
```

2. in the second case
```
@BeanMapping(unmappedTargetPolicy = ReportingPolicy.WARN, unmappedSourcePolicy = ReportingPolicy.IGNORE)
StudentDto toStudentDto(Student student);
```

But i don't know why BeanMapping#unmappedSourcePolicy does not exists ?
So i'm wondering why and also wondering if you plan another best way to achieve that goal.
thanks

</div>

Note: This is an easy issue, so if someone from the community is interested in working on this please let us know

venkatesh2090:
Add support for specifying `unmappedSourcePolicy` similar to `unmappedTargetPolicy` on `BeanMapping`.

Fixes #3309


core/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -152,6 +152,17 @@ NullValuePropertyMappingStrategy nullValuePropertyMappingStrategy()
...
     String[] ignoreUnmappedSourceProperties() default {};
 
+    /**
+     * How unmapped properties of the source type of a mapping should be reported.
+     * If no policy is configured, the policy given via {@link MapperConfig#unmappedSourcePolicy()} or
+     * {@link Mapper#unmappedSourcePolicy()} will be applied, using {@link ReportingPolicy#WARN} by default.
+     *
+     * @return The reporting policy for unmapped source properties.
+     *
+     * @since
```
filiphr, line 162: Feel free to add 1.6 here

core/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -152,6 +152,17 @@ NullValuePropertyMappingStrategy nullValuePropertyMappingStrategy()
...
+     */
+    ReportingPolicy unmappedSourcePolicy() default ReportingPolicy.WARN;
```
filiphr, line 164: Please use `ReportingPolicy.IGNORE` which is the same as the one used `Mapper`

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1636,11 +1636,11 @@ private ReportingPolicyGem getUnmappedTargetPolicy() {
             if ( mappingReferences.isForForgedMethods() ) {
                 return ReportingPolicyGem.IGNORE;
             }
-            // If we have ignoreByDefault = true, unprocessed target properties are not an issue.
-            if ( method.getOptions().getBeanMapping().isignoreByDefault() ) {
```
filiphr, line 1640: Why is this changed needed? This method is for the unmapped target policy and you are adding functionality for the unmapped source policy. Perhaps the null check on line 1639 is obsolete, but let's try to keep the PRs focused on the main functionality and leave things like this for other PRs

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1736,9 +1736,12 @@ private ReportingPolicyGem getUnmappedSourcePolicy() {
             if ( mappingReferences.isForForgedMethods() ) {
                 return ReportingPolicyGem.IGNORE;
             }
-            // If we have ignoreByDefault = true, unprocessed source properties are not an issue.
-            if ( method.getOptions().getBeanMapping().isignoreByDefault() ) {
```
filiphr, line 1740: I think that we can leave this as is and then just return 
```java
method.getOptions().getBeanMapping().unmappedSourcePolicy();
```

That makes sure that if it is not set on the bean mapping then the next one will be used.

filiphr (changes requested):
Thanks for your work on this @venkatesh2090.

Have a look at my comments. In addition to that, can you please add tests with different policies and also a combination of policy on `@Mapper` and `@BeanMapping` to show that it is being picked up properly

core/src/main/java/org/mapstruct/BeanMapping.java
```diff
@@ -152,6 +152,17 @@ NullValuePropertyMappingStrategy nullValuePropertyMappingStrategy()
      */
     String[] ignoreUnmappedSourceProperties() default {};
 
+    /**
+     * How unmapped properties of the source type of a mapping should be reported.
+     * If no policy is configured, the policy given via {@link MapperConfig#unmappedSourcePolicy()} or
+     * {@link Mapper#unmappedSourcePolicy()} will be applied, using {@link ReportingPolicy#WARN} by default.
```
filiphr, line 158:
```suggestion
     * {@link Mapper#unmappedSourcePolicy()} will be applied, using {@link ReportingPolicy#IGNORE} by default.
```

filiphr (changes requested):
Thanks for the changes. I've added one last comment and we'll merge once that is done and the build is green

venkatesh2090:
I was refactoring the tests and writing more which broke more stuff. I'll fix it and push it once it's done. FYI here's the error I get during the tests.

```shell
ERROR BeanMappingSourcePolicyMapper.java:20 'nullValueMappingStrategy', 'nullValuePropertyMappingStrategy', 'resultType' and 'qualifiedBy' are undefined in @BeanMapping, define at least one of them.
```

It might take a bit more time. My bad.

venkatesh2090:
I did manage to fix it, but I have one concern.

This error message in [Message](https://github.com/mapstruct/mapstruct/blob/main/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java#L27) looks outdated. It is being used in [BeanMappingOptions.isConsistent](https://github.com/mapstruct/mapstruct/blob/main/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java#L104) where it checks for more values than the error message points out. The message is inconsistent I think.

filiphr (changes requested):
Thanks @venkatesh2090, one last comment for the integration tests to pass and we are done. You are right, the error message in the `@BeanMapping` is inconsistent. We can change this in a different PR. The message doesn't need to enumarate all options, we can just say `@BeanMapping` must have at least one option defined

filiphr:
Thanks @venkatesh2090, this has been integrated into main

Issue #2946, opened 2022-07-29
Allow @MappingTarget to be used with @SubclassMapping


heanssgen-troy:
Currently, it is an error to use @MappingTarget in conjunction with @SubclassMapping.

This leads to the scenario where we can convert from a DTO into a subclass structure correctly, but cannot _update_ that subclass structure using another DTO. This is particularly annoying in a nested structure (such as queries and subqueries).

The current workaround is to write each subclass as an Update method and then use an @AfterMapping method to do the instanceof checks and call the appropriate update methods in the subclass. This is not ideal, as 'inherited' properties from the superclass will be mapped twice (causing excess database reads)/

It would be immensely helpful if the update method could inherit the @SubclassMapping annotations and perform the instanceof checks directly in the mapper. In the even the mapper could not find an appropriate subclass type, either throwing an exception or downcasting and using the current functionality would be good.

filiphr:
This does sound like an interesting enhancements @heanssgen-troy. Would you be willing to work on adding this to the project?

filiphr:
As I've said in https://github.com/mapstruct/mapstruct/issues/2946#issuecomment-1201504305, this is a good enhancement. We are busy with some other issues. Are you @jonsalvas perhaps interested in adding this functionality?

Errore418:
I agree that this is a big limitation of the SubclassMapping annotation. I'm willing to try to implement it. Do you @filiphr have some tips to begin with? This is my first time contributing to this project.

csabasulyok:
Would also be interested in this feature.

D3PSI:
> Would also be interested in this feature.

same here, this would be really useful

D3PSI:
i'd be willing to give implementing this myself a shot, if someone could provide guidance on what would need to be done in order to achieve this

filiphr:
Thanks @D3PSI. 

The best place to start would be in the [BeanMappingMethod#createSubclassMapping](https://github.com/mapstruct/mapstruct/blob/03563d8ffea0ca5ff4eadb297795f2407001c5dd/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java#L398). Have a look at how we create the subclass mappings for normal methods. There is also `SubclassMappingOptions#isConsistent` where we actually add the compiler error that we do not support them on update methods.

Have a look at `SubclassMappingTest` for the different test scenarios that we have.

Please let us know if you are still interested in working on this and I'd assign this issue to you

[...]

PR #3330, opened 2023-07-16
\#2946 allow update methods with `@SubclassMapping`

D3PSI:
this change implements the functionality to generate update methods
(`@MappingTarget` methods) in conjunction with `@SubclassMapping`


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1942,6 +1955,10 @@ public boolean hasSubclassMappings() {
         return !subclassMappings.isEmpty();
     }
 
+    public boolean isUpdateMethod() {
+        return Parameter.getMappingTargetParameter( getParameters() ) != null;
+    }
+
```
filiphr, line 1961: Remove this and use `isExistingInstanceMapping` instead

processor/src/main/java/org/mapstruct/ap/internal/model/ForgedMethod.java
```diff
@@ -171,6 +171,9 @@ private ForgedMethod(String name, Type sourceType, Type returnType, List<Paramet
         Parameter sourceParameter = new Parameter( sourceParamSafeName, sourceType );
         this.parameters.add( sourceParameter );
         this.parameters.addAll( additionalParameters );
+        if ( basedOn.isUpdateMethod() ) {
+            this.parameters.add( Parameter.forForgedMappingTarget( returnType ) );
+        }
```
filiphr, line 176: Why is this needed? Why don't we pass everything in the additional parameters?

processor/src/test/java/org/mapstruct/ap/test/subclassmapping/SubclassMappingTest.java
```diff
@@ -227,21 +227,11 @@ void subclassMappingInheritsInverseMappingWithCompositeMapping() {
...
-        @Diagnostic(type = ErroneousSubclassUpdateMapper.class,
-            kind = javax.tools.Diagnostic.Kind.ERROR,
-            line = 25,
-            message = "SubclassMapping annotation can not be used for update methods."
-        )
-    })
-    void unsupportedUpdateMethod() {
+    @WithClasses({ SubclassUpdateMapper.class })
+    @ExpectedCompilationOutcome(value = CompilationResult.SUCCEEDED)
+    void updateMethod() {
```
filiphr, line 234: Remove this test and write more thorough tests

filiphr (changes requested):
Thanks for your work on this @D3PSI. I've left some comments.

In addition to the comments I have some other remarks:

* Please add some more tests, perhaps by copying the existing tests for normal mapping and applying them for update mappings
* How does the check for subclass mapping exhaustion look like? From what I saw we are doing `if (source instanceof SourceType && target instanceof TargetType)` then the mapping is done. However, this means that people can make mistakes. Therefore, we need to add some errors and other things like that for this. There are also other scenarios where the target type you are updating might not matter.

Having written the second remark we might need to think for a bit better solution for subclass update mappings. Especially in combinations of what you would like to upgrade and from what.

e.g. Consider the following

```java
class Vehicle {}
class Car extends Vehicle {}
class Bike extends Vehicle {}
```

```java
class VehicleDto {}
class CarDto extends VehicleDto {}
class BikeDto extends VehicleDto {}
```

When doing update mappings there are various scenarios

```java
@Mapper
public interface CustomMapper {

    void update(@MappingTarget Vehicle target, VehicleDto source);

}
```

So what you can now do is the following:

* Use `CarDto`, `BikeDto` or `VehicleDto` to update the target
* Use `CarDto` to update the `Car` target
* Use `CarDto` to update the `Bike` target
...

As you can see with updates there are way more possible combinations that can be done.

s-jepsen:
Any progress on this PR?

heanssgen-troy:
Just putting my 2c into this issue (I made the original suggestion a while back). I believe it would be prudent to generate a single mapping method per _permutation_. You would then use the @SubclassMapping as the source of truth, only falling back to the most specific common declared ancestor if an _exact_ match was not found.

In the case of second-generation ancestors, we would add a new annotation to specify the common ancestor of each defined subclass mapping. So if I created an inheritance tree as seen below:

```
@Mapper
public interface CustomMapper {

    @SubclassMapping(source =CarDTO.class, target=Car.class)
    @SubclassMapping(source =BikeDTO.class, target=Bike.class)
    @SubclassMapping(source =F150.class, target=F150DTO.class)
    @SubclassMapping(source =Corvette.class, target=CorvetteDTO.class)
    @SubclassParent(source=Supercar.class, target=SupercarDTO.class)
    void update(@MappingTarget Vehicle target, VehicleDto source);

}
````

This would build the tree of:
image

In this case, the `F150` and `Bike` would be downcast to a `Vehicle` if no exact type-matched mapper was found (because `Car` was not declared as a `@SubclassParent`). If the `Corvette` was similarly downcast, it would match `Supercar` before it matched `Vehicle` because of the provided `@SubclassParent`.

This would mean we would **not** generate a mapping method that allows the `Corvette` to be upgraded using a `SupercarDTO` because the user did not ask us to explicitly create that relation.

In this case, I think its okay to omit non-explicit downcasts of 2nd-generation-onward ancestors, because otherwise you get the whole problem of generating a mapper for the cross product of every class and every DTO that is an ancestor of the @SubclassMapping. If we allow the craziness to unfold, we are going to get people asking why they cannot provide a `Corvette` to the `F150` mapping because they both can be downcast to `Car`.

Without the @SubparentMapping you would get this:

Imagine a heirarchy like A -> B -> C

You would have to generate:

Adto -> A
Bdto -> B
Bdto -> A
Cdto -> C
Cdto -> B
Cdto -> A

etc. etc.
Now add C -> D and C -> E

Instead of it being linear, you now have to add:
Ddto -> D
Ddto -> C
Ddto -> B
Ddto -> A
Edto -> E
Edto -> C
Edto -> B
Edto -> A

With the @SubparentMapping(source=Cdto.class, target=C.class)

Ddto -> D
Ddto -> C
Edto -> E
eDTO -> C

Issue #3334, opened 2023-07-18
Add option to map Collections to Immutable Lists


holistic-developer:
### Use case

By default, when generating a mapping method using an interface, the following would lead to generated code returning an ArrayList.

    public List<Abc> convertToAbcList(List<AbcDto> abcs);

generates:

```
public List<Abc> convertToAbcList(List<AbcDto> abcs) {
        if (abcs == null) {
            return null;
        } else {
            List<Abc> list = new ArrayList(abcs.size());
            Iterator var3 = abcs.iterator();

            while(var3.hasNext()) {
                AbcDto abcDto = (AbcDto)var3.next();
                list.add(this.convertToAbc(abcDto));
            }

            return list;
        }
    }
```

(_convertToAbc_ is also defined in the same interface)

There could be an option in the @Mapping annotation to control whether or not an immutable collection is returned from the mapping function.

### Generated Code

```
public List<Abc> convertToAbcList(List<AbcDto> abcs) {
        if (abcs == null) {
            return null;
        } else {
            List<Abc> list = new ArrayList(abcs.size());
            Iterator var3 = abcs.iterator();

            while(var3.hasNext()) {
                AbcDto abcDto = (AbcDto)var3.next();
                list.add(this.convertToAbc(abcDto));
            }

            return List.copyOf(list);
        }
    }
```
`List.copyOf(list)` wraps the existing collection into an unmodifiable version.

Alternatively the following would also be possible:
```
public List<Abc> convertToAbcList(List<AbcDto> abcs) {
        if (abcs == null) {
            return null;
        } else {
            return abcs.stream().map(convertToAbc).toList();
        }
    }
```

### Possible workarounds

My current workaround is to override these mapping functions by providing a default implementation that does the mapping using a stream and the toList collector.
```
default public List<Abc> convertToAbcList(List<AbcDto> abcs) {
    return abcs.stream().map(this::convertToAbc).toList();
}
```

### MapStruct Version

1.5.5.Final

filiphr:
Thanks for the detailed request @holistic-developer. We'll have this into consideration. Currently, we are focused on the 1.6 release, so we can start looking into this maybe after that. If someone is interested in contributing this, feel free to let us know

j-be:
I got bored on a rainy Sunday afternoon, so I gave it a go. See #3356.

I tried to be as little invasive as possible.

[...]

PR #3356, opened 2023-08-06
#3334: Support returning unmodifiable collections and maps

j-be:


filiphr:
Thanks for your work on this @j-be. I managed to review it and it is a really good direction. There are some things that are still missing:

* Need support for mapping `Stream` to `List`, `Collection` etc. (this is part of the `StreamMappingMethod`
* We need error handling for mapping to types that do not support `unmodifiable` e.g. if you do `@IterableMapping(unmodifiable = true)` when the return type is `ArrayList` or some other type of collection that cannot be called on `Collections.unmodifiableXXX` then we need to have a compile error from MapStruct instead of leaving the compile error on the generated code.

Issue #3303, opened 2023-06-07
Mapping from Map to Map and specify source and target values


JM92103:
### Use case

I have a use case where I need to convert a Map<String, String> to a Map<String, String> but manipulate the keys. Currently I am trying to do 
```
@Mapping(target = "value",  source = "Cash_Amount")
@Mapping(target = "valueDate",  source = "Date_Of_Money_Movement")
Map<String, String> fromCsvToStpTransaction(Map<String, String> rec);
```
But the generated code is ignoring these values and just doing a basic Map conversion like below
```
@Override
public Map<String, String> fromCsvToStpTransaction(Map<String, String> rec) {
    if ( rec == null ) {
        return null;
    }

    Map<String, String> map = new LinkedHashMap<String, String>( Math.max( (int) ( rec.size() / .75f ) + 1, 16 ) );

    for ( java.util.Map.Entry<String, String> entry : rec.entrySet() ) {
        String key = entry.getKey();
        String value = entry.getValue();
        map.put( key, value );
    }

    return map;
}
```

### MapStruct Version

1.5.5

filiphr:
This is a valid request @JM92103, I don't see why something like this shouldn't be supported. However, I would expect the implementation to look like:

```java
public class MyMapperImpl implements MyMapper {

    private final Map<String, String> fromCsvToStpTransactionKeyMappings;

    public MyMapperImpl() {
        this.fromCsvToStpTransactionKeyMappings = new HashMap<>();
        this.fromCsvToStpTransactionKeyMappings.put( "Cash_Amount", "value" );
        this.fromCsvToStpTransactionKeyMappings.put( "Date_Of_Money_Movement", "valueDate" );
    }

    @Override
    public Map<String, String> fromCsvToStpTransaction(Map<String, String> rec) {
        if ( rec == null ) {
            return null;
        }

        Map<String, String> map = new LinkedHashMap<>( Math.max( (int) ( rec.size() / .75f ) + 1, 16 ) );

        for ( java.util.Map.Entry<String, String> entry : rec.entrySet() ) {
            String key = entry.getKey();
            String value = entry.getValue();
            map.put( this.fromCsvToStpTransactionKeyMappings.getOrDefault( key, key ), value );
        }

        return map;
    }
}
```

The `fromCsvToStpTransactionKeyMappings` should only be created if the mapping method has `@Mapping`.

I'll put this for "up for grabs" in case someone from the community is interested in working on this.

Issue #3293, opened 2023-05-30
Provide possibility to map maps with enum keys to beans


ruedigerk:
### Use case

Today, MapStruct allows mapping maps with String keys to beans (see https://mapstruct.org/documentation/stable/reference/html/#mapping-map-to-bean).

Please extend this capability to allow mapping maps with enum keys to beans. This could be done by first mapping the enum value to String an then applying the map to bean mapping as of today. This was also suggested in this comment of ticket #717 (https://github.com/mapstruct/mapstruct/issues/717#issuecomment-197248073).

This could be further extended to allow mapping maps with any key type that is mappable to String.

### MapStruct Version

1.5.5

filiphr:
Are you sure that you referenced the right comment @ruedigerk? From what I can see that comment talks about `Map<String, ?>` and not `Map<Enum<?>, ?>`

In any case you are requesting something like:

```java
public enum FieldEnum {

    name,
    lastName,
    age
}

public class Target {

    private String name;
    private String lastName;
    private String age;

    // getters and setters omitted for brevity
}

@Mapper
public interface MapWithEnumMapper {

    Target map(Map<FieldEnum, String> map);

}
```

and this would generate something like:

```java
public class MapWithEnumMapperImpl implements MapWithEnumMapper{
    @Override
    public Target map(Map<FieldEnum, String> map) {
        if ( map == null ) {
            return null;
        }

        Target target = new Target();

        if ( map.containsKey( FieldEnum.name ) ) {
            target.setName( map.get( FieldEnum.name ) );
        }

        if ( map.containsKey( FieldEnum.lastName ) ) {
            target.setLastName( map.get( FieldEnum.lastName ) );
        }

        if ( map.containsKey( FieldEnum.age ) ) {
            target.setAge( map.get( FieldEnum.age ) );
        }

        return target;
    }
}
```

please let me know if I got this right.

ruedigerk:
Hello @filiphr,

yes, you are right, that is what I meant. It seems, I might have been a little tired when I wrote the ticket, because as you have said, the referenced comment talks about something else. 

I was just now thinking about what I meant with the sentence "This could be done by first mapping the enum value to String an then applying the map to bean mapping as of today". My idea was to first map the enum keys to Strings, and then using these Strings a the field names of the object to map to. But that would require setting the fields by reflection and therefore doesn't make a whole lot of sense with MapStruct. So let's disregard that sentence of mine.

[...]

PR #3367, opened 2023-08-28
#3303 Mapping from Map to Map and specify source and target keys of type String

thunderhook:
Tackling map mapping with source and target keys.

Things to note:

- The variable containing key mappings is not built safe like in `org.mapstruct.ap.internal.model.SupportingField#getSafeField`. This would need a bigger change in the `MapperCreationProcessor`. However, a collision with a field named like `method.getName() + "KeyMappings"` is very unlikely.
- Only `@Mapping` annotations with set `source` and `target` are considered, the rest is ignored, like `expression` or `ignore` (I felt that it is not worth to throw an error in that case)
- This PR tackles only mapping of  maps with keys of type string (`Map<String, ?>`). Any other type does not feel like a valid use-case (well, `CharSequence` eventually?)
    - Due to this restriction it was easier to implement. Because the type of the keyMappings field are now hard-coded as `Map<String, String>` in the FreeMarker template.
    - I found no easy solution to get a type of `Map<String, String>` with the `TypeFactory`. If you have a solution for this, please let me know. I could then remove the `MapKeyMappingField.ftl` and use a typed `Field.ftl`

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/MapKeyMappingConstructorFragment.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/MapKeyMappingConstructorFragment.ftl
@@ -1,0 +1,12 @@
+<#--
+
+    Copyright MapStruct Authors.
+
+    Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+
+-->
+<#-- @ftlvariable name="" type="org.mapstruct.ap.internal.model.MapKeyMappingConstructorFragment" -->
+this.${field.variableName} = new java.util.HashMap<>();
+<#list keyMappings as entry>
+  this.${field.variableName}.put( "${entry.source}", "${entry.target}" );
+</#list>
--- a/processor/src/test/java/org/mapstruct/ap/test/collection/map/AnnotatedMapMappingTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/collection/map/AnnotatedMapMappingTest.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.collection.map;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.mapstruct.ap.test.collection.map.other.ImportedType;
+import org.mapstruct.ap.testutil.IssueKey;
+import org.mapstruct.ap.testutil.ProcessorTest;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.runner.GeneratedSource;
+
+import static java.lang.System.lineSeparator;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.entry;
+
+/**
+ * Test for implementation of {@code Map} mapping methods with @{@link org.mapstruct.Mapping Mappings}
+ *
+ * @author Oliver Erhart
+ */
+@WithClasses({
+    AnnotatedSourceTargetMapper.class,
+    CustomNumberMapper.class,
+    Source.class,
+    Target.class,
+    ImportedType.class
+})
+@IssueKey("3303")
+public class AnnotatedMapMappingTest {
+
+    @RegisterExtension
+    final GeneratedSource generatedSource = new GeneratedSource();
+
+    @ProcessorTest
+    public void shouldContainTwoKeyMappingMapsInConstructor() {
+        generatedSource.forMapper( AnnotatedSourceTargetMapper.class )
+            .content()
+            .contains(
+                "private final Map<String, String> stringMapToStringMapUsingTargetParameterKeyMappings;" )
+            .contains(
+                "private final Map<String, String> stringMapToStringMapUsingTargetParameterAndReturnKeyMappings;" )
+            .contains( "private final Map<String, String> stringMapWithMappingAnnotationsKeyMappings;" )
+            .contains(
+                "private final Map<String, String> stringMapWithMappingAnnotationsAndDifferentNameKeyMappings;" )
+            .doesNotContain( "private final Map<String, String> stringMapWithInvalidMappingAnnotationsKeyMappings;" )
+            .doesNotContain( "private final Map<String, String> longMapWithMappingAnnotationsKeyMappings;" )
+            .contains( expectedConstructor() );
+    }
+
+    private static String expectedConstructor() {
+        return "public AnnotatedSourceTargetMapperImpl() {" +
+            lineSeparator() +
+            constructorInitOf( "stringMapToStringMapUsingTargetParameterKeyMappings" ) +
+            lineSeparator() +
+            constructorInitOf( "stringMapToStringMapUsingTargetParameterAndReturnKeyMappings" ) +
+            lineSeparator() +
+            constructorInitOf( "stringMapWithMappingAnnotationsKeyMappings" ) +
+            lineSeparator() +
+            constructorInitOf( "stringMapWithMappingAnnotationsAndDifferentNameKeyMappings" ) +
+            "    }";
+    }
+
+    private static String constructorInitOf(String parameterName) {
+        return
+            "        this." + parameterName + " = new java.util.HashMap<>();" +
+                lineSeparator() +
+                "        this." + parameterName + ".put( \"sourceKey\", \"targetKey\" );" +
+                lineSeparator() +
+                "        this." + parameterName + ".put( \"nonExistentSourceKey\", \"otherTargetKey\" );" +
+                lineSeparator();
+    }
+
+    @ProcessorTest
+    public void shouldMapBasedOnKeyMappings() {
+        Map<String, String> values = new HashMap<>();
+        values.put( "sourceKey", "value" );
+        values.put( "unmappedKey", "otherValue" );
+
+        Map<String, String> target = AnnotatedSourceTargetMapper.INSTANCE.stringMapWithMappingAnnotations( values );
+
+        assertThat( target ).isNotNull()
+            .hasSize( 2 )
+            .contains(
+                entry( "targetKey", "value" ),
+                entry( "unmappedKey", "otherValue" )
+            );
+    }
+
+    @ProcessorTest
+    public void shouldNotMapBasedOnKeyMappingsBecauseOfTypeLong() {
+        Map<Long, Long> values = new HashMap<>();
+        values.put( 1L, 2L );
+        values.put( 3L, 4L );
+
+        Map<Long, Long> target = AnnotatedSourceTargetMapper.INSTANCE.longMapWithMappingAnnotations( values );
+
+        assertThat( target ).isNotNull()
+            .hasSize( 2 )
+            .isEqualTo( values );
+    }
+
+}
--- a/processor/src/test/java/org/mapstruct/ap/test/collection/map/AnnotatedMapMappingTest.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/collection/map/AnnotatedMapMappingTest.java
@@ -48,6 +48,8 @@ public class AnnotatedMapMappingTest {
             .contains( "private final Map<String, String> stringMapWithMappingAnnotationsKeyMappings;" )
             .contains(
                 "private final Map<String, String> stringMapWithMappingAnnotationsAndDifferentNameKeyMappings;" )
+            .contains(
+                "private final Map<String, String> stringLongMapToStringIntegerMapWithMappingAnnotationsKeyMappings;" )
             .doesNotContain( "private final Map<String, String> stringMapWithInvalidMappingAnnotationsKeyMappings;" )
             .doesNotContain( "private final Map<String, String> longMapWithMappingAnnotationsKeyMappings;" )
             .contains( expectedConstructor() );
@@ -63,6 +65,8 @@ public class AnnotatedMapMappingTest {
             constructorInitOf( "stringMapWithMappingAnnotationsKeyMappings" ) +
             lineSeparator() +
             constructorInitOf( "stringMapWithMappingAnnotationsAndDifferentNameKeyMappings" ) +
+            lineSeparator() +
+            constructorInitOf( "stringLongMapToStringIntegerMapWithMappingAnnotationsKeyMappings" ) +
             "    }";
     }
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AnnotatedConstructor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AnnotatedConstructor.java
@@ -10,6 +10,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import org.mapstruct.ap.internal.model.common.ConstructorFragment;
 import org.mapstruct.ap.internal.model.common.ModelElement;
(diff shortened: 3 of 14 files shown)


processor/src/main/resources/org/mapstruct/ap/internal/model/MapKeyMappingConstructorFragment.ftl
```diff
@@ -0,0 +1,12 @@
+<#--
+
+    Copyright MapStruct Authors.
+
+    Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+
+-->
+<#-- @ftlvariable name="" type="org.mapstruct.ap.internal.model.MapKeyMappingConstructorFragment" -->
+this.${field.variableName} = new java.util.HashMap<>();
```
filiphr, line 9: For this you can then use an `IterableCreation` in the `MapKeyMappingConstructorFragment`. Most likely we'll need a new method in `IterableCreation`

e.g.

```java
    public static IterableCreation create(Type resultType) {
        return new IterableCreation( resultType, null, null );
    }
```

I think that the new line will then be 
```freemarker
this.${field.variableName} = <@includeModel object=iterableCreation useSizeIfPossible=false/>;
```

processor/src/test/java/org/mapstruct/ap/test/collection/map/AnnotatedMapMappingTest.java
```diff
@@ -0,0 +1,112 @@
...
+    @ProcessorTest
+    public void shouldNotMapBasedOnKeyMappingsBecauseOfTypeLong() {
+        Map<Long, Long> values = new HashMap<>();
+        values.put( 1L, 2L );
+        values.put( 3L, 4L );
+
+        Map<Long, Long> target = AnnotatedSourceTargetMapper.INSTANCE.longMapWithMappingAnnotations( values );
+
+        assertThat( target ).isNotNull()
+            .hasSize( 2 )
```
filiphr, line 108: `isNotNull` and `hasSize` are obsolete since we are doing `isEqualTo` already

processor/src/test/java/org/mapstruct/ap/test/collection/map/AnnotatedMapMappingTest.java
```diff
@@ -0,0 +1,112 @@
...
+        values.put( "unmappedKey", "otherValue" );
+
+        Map<String, String> target = AnnotatedSourceTargetMapper.INSTANCE.stringMapWithMappingAnnotations( values );
+
+        assertThat( target ).isNotNull()
+            .hasSize( 2 )
+            .contains(
+                entry( "targetKey", "value" ),
+                entry( "unmappedKey", "otherValue" )
+            );
```
filiphr, line 96: we can remove `isNotNull` and `hasSize` if we use `containsOnly` instead of `contains`

processor/src/test/java/org/mapstruct/ap/test/collection/map/AnnotatedMapMappingTest.java
```diff
@@ -0,0 +1,112 @@
...
+    ImportedType.class
+})
+@IssueKey("3303")
+public class AnnotatedMapMappingTest {
+
+    @RegisterExtension
+    final GeneratedSource generatedSource = new GeneratedSource();
+
+    @ProcessorTest
+    public void shouldContainTwoKeyMappingMapsInConstructor() {
```
filiphr, line 41: I would prefer if we have a fixture for this and we test the contents. Makes it a bit easier to see what gets generated as well

processor/src/test/java/org/mapstruct/ap/test/collection/map/AnnotatedMapMappingTest.java
```diff
@@ -0,0 +1,112 @@
...
+ * Test for implementation of {@code Map} mapping methods with @{@link org.mapstruct.Mapping Mappings}
+ *
+ * @author Oliver Erhart
+ */
+@WithClasses({
+    AnnotatedSourceTargetMapper.class,
+    CustomNumberMapper.class,
+    Source.class,
+    Target.class,
+    ImportedType.class
```
filiphr, line 32: Why are these needed? I guess it is a copy paste thing. Let's remove it if that is the case

filiphr (changes requested):
Really nice PR @thunderhook. Thanks for your work on this.

I've left some comments in the PR review, still need to play with it to see the actual generated code.

Here are some comments for your notes:

>The variable containing key mappings is not built safe like in ...

I've left a review comment for this. If we do it like I suggest there, I think that we can use the `SupportingField#getSafeField`

>Only `@Mapping` annotations with set `source` and `target` are considered, the rest is ignored, like `expression` or `ignore` (I felt that it is not worth to throw an error in that case)

If we are going to start something that we didn't support before then in my opinion we should be as strict as possible. This means that it is fine to only support `source` and `target`. However, if someone sets something else then we should have a compile error. 

Alternatively we can also go with a new annotation for this since for this `source` is mandatory, but not `target`. I do see people asking for things like we have in `BeanMapping#ignoreByDefault` and maybe expanding more on how value for a particular key should be mapped using the `qualifedByName` and stuff like this. Having said all of this I think it would be better for us if we go with a new annotation e.g. `@MapKeyMapping` (open to other name suggestions as well)

I am not saying that we should add all this customizations in this PR, but I would like to anticipate things and make sure that we are implementing this right. We do have #3293 where something else around map mapping is being requested.

>This PR tackles only mapping of maps with keys of type string (`Map<String, ?>`). Any other type does not feel like a valid use-case (well, `CharSequence` eventually?)

That is completely OK. However, see my comment for the validations. We also need to add tests for mappings from `Map<X, Y>` to `Map<String, ?>`. We do support converting `X` to `String` through some kind of a mapping or if it is some of the out of the box mappings from `X` to `String`. Check `org.mapstruct.ap.test.collection.map.SourceTargetMapper` as an example.

```java
Map<String, Long> map(Map<Long, Long> source);
```

>I found no easy solution to get a type of `Map<String, String>` with the `TypeFactory`. If you have a solution for this, please let me know. I could then remove the `MapKeyMappingField.ftl` and use a typed `Field.ftl`

You can get the right map by doing something like:

In the `TypeFactory` add:

```java
    public Type getMapType(String keyCanonicalName, String valueCanonicalName) {
        TypeElement mapTypeElement = getTypeElement( Map.class.getCanonicalName() );
        TypeElement keyTypeElement = getTypeElement( keyCanonicalName );
        TypeElement valueTypeElement = getTypeElement( valueCanonicalName );
        DeclaredType mapType = typeUtils.getDeclaredType(
            mapTypeElement,
            keyTypeElement.asType(),
            valueTypeElement.asType()
        );
        return getType( mapType );
    }
```

The `getTypeElement` looks like:

```java
    private TypeElement getTypeElement(String canonicalName) {
        TypeElement typeElement = elementUtils.getTypeElement( canonicalName );

        if ( typeElement == null ) {
            throw new AnnotationProcessingException(
                "Couldn't find type " + canonicalName + ". Are you missing a dependency on your classpath?"
            );
        }

        return typeElement;
    }
```

This is extracted from `private Type getType(String canonicalName, boolean isLiteral)`

thunderhook:
Thank you for your detailed and thorough review.

> Alternatively we can also go with a new annotation for this since for this `source` is mandatory, but not `target`.

Agreed with a new annotation (didn't think of the `target` not being mandatory btw).

I guess this should also be pre JDK-8 compatible? So there will be `@MapKeyMappings` and `@MapKeyMapping`. I personally like the name.
A new annotation might also lead to a better separation of code.

> I am not saying that we should add all this customizations in this PR, but I would like to anticipate things and make sure that we are implementing this right.

Yeah. I will start with the "basic" key mapping feature via `source` and `target`. We can then see how far I got, and then you can decide to give it another round or release it with the simple mapping.
You (or the team) can then prioritize the next features, maybe #3293 because it was a feature request.

----

Thank you for the detailled comments (especially about the type(s)), I will incorporate them as I refactor the code.

filiphr:
>I guess this should also be pre JDK-8 compatible? So there will be `@MapKeyMappings` and `@MapKeyMapping`. I personally like the name.

The need of `@MapKeyMappings` is not for pre JDK-8 compatibility. It is needed for all JDKs, due to the need of `@Repeatable`.

-----

Side note, can you please reach out to me on my email? You can find it in any of my commits.

thunderhook:
@filiphr
Sorry for the wall of text 🙈 

I made good progress but then I confused myself with the conversion of  `Map<X, Y>` to `Map<String, ?>`
Let's say, we have a map of Company-IDs (Long) with their amounf of employees (Long):

| Company ID  | Employees |
| ------------- | ------------- |
| 1 | 100  |
| 2 | 200  |
| 3 | 300  |

### Scenario 1 Mapping Long to String without custom mapping method

Let's say that company 1 and 2 need to be switched. With the original implementation that would've looked like...
```java
    @MapKeyMapping( target = "1L", source = "2L" )
    @MapKeyMapping( target = "2L", source = "1L" )
    Map<String, Long> mapLongKeyMapToStringKeyMap(Map<Long, Long> source);


    // GENERATED CODE
    private final Map<String, String> mapLongKeyMapToStringKeyMapKeyMappings;

    public MapKeyMappingCommonConversionMapperImpl() {
        this.mapLongKeyMapToStringKeyMapKeyMappings = new LinkedHashMap<String, String>();
        this.mapLongKeyMapToStringKeyMapKeyMappings.put( "1L", "2L" );
        this.mapLongKeyMapToStringKeyMapKeyMappings.put( "2L", "1L" );
    }

    public Map<String, Long> mapLongKeyMapToStringKeyMap(Map<Long, Long> source) {

        Map<String, Long> map = new LinkedHashMap<String, Long>();

        for ( java.util.Map.Entry<Long, Long> entry : source.entrySet() ) {
            String key = String.valueOf( entry.getKey() ); // generates "1" instead of "1L" 
            Long value = entry.getValue();
            map.put( this.mapLongKeyMapToStringKeyMapKeyMappings.getOrDefault( key, key ), value ); // "1" != "1L"
        }

        return map;
    }
```

If **there is no custom mapping** method from Long to String, `String.valueOf()` is being used to convert the key (see generated code).
When you define `"1L"` like in `Mapping#constant` this will not hit the Key `"1"` in the mapping key map.
Workaround for this: just use "1" instead. This is bad because the constant expression then has a different behaviour in `@Mapping` and `@MapKeyMapping`.

### Scenario 2: Mapping Long to String with custom mapping

Let's say we have a custom mapping method (which is not qualified, since `qualifiedByName` is not implemented for this yet...) from Long to String, which maps like:

```java
String map(Long value) {
    return "Company" + value;
}
```

How should map key mappings be defined?

1. define the source typed as the source key type and define the target typed as the target key type
2. define both typed as the the target key type
3. define both typed as the the source key type

```java
    // Option 1
    @MapKeyMapping( target = "Company1", source = "2L" )
    @MapKeyMapping( target = "Company2", source = "1L" )
    Map<String, Long> option1(Map<Long, Long> source);

    // Option 2
    @MapKeyMapping( target = "Company1", source = "Company2" )
    @MapKeyMapping( target = "Company2", source = "Company1" )
    Map<String, Long> option2(Map<Long, Long> source);

    // Option 3
    @MapKeyMapping( target = "1L", source = "2L" )
    @MapKeyMapping( target = "2L", source = "1L" )
    Map<String, Long> option3(Map<Long, Long> source);


    // GENERATED CODE
    private final Map<Long, String> option1KeyMappings;
    private final Map<String, String> option2KeyMappings;
    private final Map<Long, Long> option3KeyMappings;

    public MapKeyMappingCommonConversionMapperImpl() {

        this.option1KeyMappings = new LinkedHashMap<Long, String>();
        this.option1KeyMappings.put( (long) 1L, "Company2" );
        this.option1KeyMappings.put( (long) 2L, "Company1" );

        this.option2KeyMappings = new LinkedHashMap<Long, String>();
        this.option2KeyMappings.put( "Company1", "Company2" );
        this.option2KeyMappings.put( "Company2", "Company1" );

        this.option3KeyMappings = new LinkedHashMap<Long, Long>();
        this.option3KeyMappings.put( (long) 1L, (long) 2L );
        this.option3KeyMappings.put( (long) 2L, (long) 1L );
    }

```

#### Option 1

Using a diffrent type between source and target needs a different access inside the mapping method:

```java
// GENERATED CODE
    @Override
    public Map<String, Long> option1(Map<Long, Long> source) {
        if ( source == null ) {
            return null;
        }

        Map<String, Long> map = new LinkedHashMap<String, Long>( Math.max( (int) ( source.size() / .75f ) + 1, 16 ) );

        for ( java.util.Map.Entry<Long, Long> entry : source.entrySet() ) {
            String key = String.valueOf( entry.getKey() );
            Long value = entry.getValue();
            Long sourceMappingKey = entry.getKey();
            map.put( this.option1KeyMappings.getOrDefault( sourceMappingKey, key ), value );
        }

        return map;
    }
```

Implementation-wise I am certain that we should go with option 1.
It supports a good type-safety during compile time. When an invalid mapping is defined, the mapper cannot be generated, e.g.:
```java
    @MapKeyMapping( target = "Company1", source = "NOT_A_LONG") // cannot generate mapper
    Map<String, Long> mapLongKeyMapToStringKeyMap(Map<Long, Long> source);
```

Cons: seems unintuitive to define different types

Remember the problem from Scenario 1 with `"1" != "1L"`? This would be solved with option 1.

#### Option 2

I think the second option is most understandable from a users perspective. Or is it? I am not sure. However this can lead to wrong (and missed) mappings.
What if I define a mappig that will never work, because it will never be mapped like that, e.g.:
```java
    @MapKeyMapping( target = "Company1", source = "CompanyThatNeverHits")
    Map<String, Long> mapLongKeyMapToStringKeyMap(Map<Long, Long> source);
```

This is just not mapped and the user doesn't get informed.

#### Option 3
I don't think that option 3 is really an option. Just using the source key type is hard to understand, right?

### Scenario 3 Map to (or even from) an enum

I reuse the code to get the expression of source/target like it is handled with the `constant` field. So besides primitive types, also enums are allowed.

Let's say we have a map with month and employees count, and all employees from `FEBRUARY` must be switched with `MARCH` (only reasonable example I can find atm)

```java
@MapKeyMapping( target = "FEBRUARY", source = "MARCH" )
@MapKeyMapping( target = "MARCH", source = "FEBRUARY" )
Map<String, Long> map(Map<java.util.Month, Long> source);
```

This works. It does not matter if the key mapping map is backed with a String or Month map.

Let's add a custom enum conversion method:
```java
String mapMonth(Month value) {
    return value.name().substring(0, 3);
}
```

How about the mapping? Must it be changed to the following? This example should help to chose for an option from above
```java
// option 1
@MapKeyMapping( target = "FEB", source = "MARCH" )
@MapKeyMapping( target = "MAR", source = "FEBRUARY" )
// option 2
@MapKeyMapping( target = "FEB", source = "MAR" )
@MapKeyMapping( target = "MAR", source = "FEB" )
// option 3
@MapKeyMapping( target = "FEBRUARY", source = "MARCH" )
@MapKeyMapping( target = "MARCH", source = "FEBRUARY" )
Map<String, Long> map(Map<java.util.Month, Long> source);
```


### Scenario 4 (bonus) Mapping from Long to Long (or: same source key type as target key type)

I can see no reason why this should not work. And this will work when the key mapping map supports different types than String. This is currently implemented and working as expected.

```java
@MapKeyMapping( taraget = "1L", source = "2L")
@MapKeyMapping( taraget = "2L", source = "1L")
Map<Long, Long> mapLongsToLongs(Map<Long, Long> source);
```

### Conclusion

During writing this wall of text (took me like an hour...) I could think about it more and analyze it. I want to know your opinion about the possible options, and what would you choose?
Is there another option, like disabling the support of one of these things (like really just support `@MapKeyMapping` when source and target have the same type)?

Thanks in advance. Take your time to answer it. I would love to push my current code, but it is really mixed up with those three options.

Issue #3089, opened 2022-11-23
Improve support for Map attributes for immutables 


GVladi:
### Expected behavior

The mapstruct should not produce a "Unmapped target properties" for "put" and "putAll" methods.

### Actual behavior

The mapstruct supports org.immutables objects already. In case an object has got a java.util.Map attributes like:
`public abstract Map<String, String> getAttributes();`

the immutables generates the following methods in the Builder class for this attribute:
`public final Builder putAttributes(String key, String value)`
`public final Builder putAttributes(Map.Entry<String, ? extends String> entry)`
`public final Builder attributes(Map<String, ? extends String> entries)`
`public final Builder putAllAttributes(Map<String, ? extends String> entries)`

The mapstruct produces a Warning for "put" and "putAll" methods:
`WARNING ItemMapper.java:24 Unmapped target properties: "putAttributes
putAllAttributes".`
as it believes they are fluentSetter.

As a workaround, the warnings can be suppressed by:
`@Mapping(target = "putAttributes", ignore = true)`

### Steps to reproduce the problem

To reproduce the issue, I have add `Map<String, String> getAttributes()` to the `org.mapstruct.ap.test.bugs._1801.domain.Item`:

```
Index: processor/src/test/java/org/mapstruct/ap/test/bugs/_1801/dto/ImmutableItemDTO.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/processor/src/test/java/org/mapstruct/ap/test/bugs/_1801/dto/ImmutableItemDTO.java b/processor/src/test/java/org/mapstruct/ap/test/bugs/_1801/dto/ImmutableItemDTO.java
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_1801/dto/ImmutableItemDTO.java	(revision fd27380185fe83feb6d26308c7d358025783ff8f)
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_1801/dto/ImmutableItemDTO.java	(date 1669213502716)
@@ -7,7 +7,10 @@
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
+import java.util.Collections;
+import java.util.LinkedHashMap;
 
 /**
  * Immutable implementation of {@link ItemDTO}.
@@ -19,9 +22,11 @@
  */
[...]

PR #3368, opened 2023-08-30
#3089 Improve support for Map attributes for immutables

thunderhook:
Work was already done by @GVladi, thank you very much. I just cleaned up the generated class to be checkstyle-conform.

@filiphr is it possible to change the commit author? Would be nice, because I don't want to take credit for someone else's work.


filiphr (changes requested):
>@filiphr is it possible to change the commit author? Would be nice, because I don't want to take credit for someone else's work.

Yes I can do that. Is there a commit by @GVladi that I can look into in order to be able to get the email (event if it is the noreply github one)? I would also need the consent from @GVladi that is is OK to put him as an author.

@thunderhook I see that @GVladi shared a patch in the issue, which I missed. I would prefer if we add a dedicated test for this in the `_3089` package. Can you perhaps do that @thunderhook? I will add both you and @GVladi as authors in that case, @GVladi as the main one.

GVladi:
@filiphr I don't mind, if you put me on the authors list ;-)

thunderhook:
Of course. I have provided a separate test. Just tell me if there is something else you want me to change.

filiphr:
Thanks @GVladi and @thunderhook. I've merged this manually in f61a3acec301de2a781bd46a56cdcf5c83ca159e

PR #3421, opened 2023-10-19
Fixed the misleading default of defaultValue in Mapping

verhasi:
In the org.mapstruct.Mapping.java the defaultValue() is provided with a default ""; Actually during the source generation the default empty value is not applied if defaultValue is not provided in the annotation. It is working perfectly when you provide the defaultValue="" in the annotation. The problem is that the definition of the default is misleading and even the IDE (e.g. the IDEA) propose to remove the redundant parameter. But removing the parameter result is null value in target.

### DIFF
--- a/core/src/main/java/org/mapstruct/Mapping.java
+++ b/core/src/main/java/org/mapstruct/Mapping.java
@@ -442,7 +442,7 @@ public @interface Mapping {
      *
      * @return Default value to set in case the source property is {@code null}.
      */
-    String defaultValue() default "";
+    String defaultValue();
 
     /**
      * Determines when to include a null check on the source property value of a bean mapping.


filiphr:
If this change is applied then `defaultValue` will become a mandatory parameter and this is something that we do not want to do.

For an annotation processor we know whether the user has explicitly defined a value for the property or they have not defined it. So if you want to use an empty string as a default value you need to do `@Mapping(target = "target", defaultValue = "")`. 

I would suggest that you raise an issue with IntelliJ to treat annotations like this differently. We could improve the Javadoc perhaps, but I'll leave that as a different PR.

verhasi:
Hi Filip!

That is a good point that I was not aware of. It means we must add the default if we do not want the defaultValue to be mandatory. On the other hand, the declared value of the default is not used at all. If it does not matter what value it has (as null could not be declared) then it could be e.g. "null value will be used if not declared". In this way the code became self-documented and the mentioned warning would emerge only if someone needs the defaultValue to be "null value will be used if not declared". I suppose it is much less frequent than the empty string.
If you agree that it is a good idea I am going to update the PR.

filiphr:
Until now no one has complained about the IntelliJ warning. Therefore, I would prefer if we just leave it as is. An alternative could be to look into the MapStruct IntelliJ plugin and see if it can be solved through there, but that is independent of this repo

Issue #3400, opened 2023-10-09
Remove unnecessary casts to long


shollander:
### Expected behavior

When use a defaultValue with a long value, ex:
```
class MyObject {
    private Long myField;
    ....
}

class MyMapper {
    @Mapping(target="myField", source="myField", defaultValue="0L")
    abstract MyObject map(MyOtherObject obj);
}
```
The generated code should not produce any compiler warnings.

### Actual behavior
```
class MyMapperImpl {
    abstract MyObject map(MyOtherObject obj) {
        ...
        if(obj.getMyField() != null) {
            myObject.setMyField( obj.getMyField() );
        } else {
            obj.setMyField( (long) 0L );         //   <--- This cast is unnecessary and produces a warning when compiling with -Xlint:all
        }
    }
}
```
### Steps to reproduce the problem

See expected behavior above.

### MapStruct Version

1.5.5

thunderhook:
Thanks @dehasi for the fix!

PR #3431, opened 2023-10-31
#3400 Remove unnecessary casts to long

dehasi:
Address issue [#3400](https://github.com/mapstruct/mapstruct/issues/3400)

An unnecessary cast appears during code generation, because `LongAnalyzer#getLiteral` returns `int` instead of `long`.

The problem happens because we `NativeTypes.getLiteral` returns an incorrect value.
It happens in [PropertyMapping.java:914](https://github.com/dehasi/mapstruct/blob/main/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java#L915)

```java
try {
  // getLiteral returns int instead of long
  baseForLiteral = NativeTypes.getLiteral( targetType.getFullyQualifiedName(), constantExpression );
}
```

This happens in the implementation of [NativeTypes$LiteralAnalyzer:51](https://github.com/dehasi/mapstruct/blob/main/processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java#L51)
for `Long`.
```java
private interface LiteralAnalyzer {

    void validate(String s);

    Class<?> getLiteral();
}
```

[LongAnalyzer#getLiteral()](https://github.com/dehasi/mapstruct/blob/main/processor/src/main/java/org/mapstruct/ap/internal/util/NativeTypes.java#L366)
returns `int`, then it ends up in decorated assignments as a cast.

```java
public Class<?> getLiteral() { // in LongAnalyzer
    return int.class;
}
```
 
**Similar problem** is in `ByteAnalyzer`, `ShortAnalyzer`, and `DoubleAnalyzer`. I can fix them as well in this PR.

**However**. Fix affects [ConstantsMapperImpl](https://github.com/mapstruct/mapstruct/blob/main/processor/src/test/resources/fixtures/org/mapstruct/ap/test/source/constants/ConstantsMapperImpl.java#L36) in fixtures because it expects explicit casting. Which makes me think that maybe [#3400](https://github.com/mapstruct/mapstruct/issues/3400) is not a bug but a feature :)

**Build:** Before creating the PR, I run `mvn clean verify`. After `08:23 min` maven produced `BUILD SUCCESS`.
So it works on my machine :)


dehasi:
@filiphr @thunderhook Could you please take a look?
That's my first PR to MapStruct.

stepanenkomaks (approved):

thunderhook:
Hi @dehasi and thank you for the pull request.

I also tried to tackle this issue a month ago and stopped it quite confused.

> Similar problem is in `ByteAnalyzer`, `ShortAnalyzer`, and `DoubleAnalyzer`. I can fix them as well in this PR.
>
> However. Fix affects [ConstantsMapperImpl](https://github.com/mapstruct/mapstruct/blob/main/processor/src/test/resources/fixtures/org/mapstruct/ap/test/source/constants/ConstantsMapperImpl.java#L36) in fixtures because it expects explicit casting. Which makes me think that maybe #3400 is not a bug but a feature :)

Changing the `ByteAnalyzer` to `byte.class` and `ShortAnalyzer` to `short.class` results in compile errors in the `ConstantsTest`.

```
DiagnosticDescriptor: ERROR ConstantsMapperImpl.java:24 incompatible types: possible lossy conversion from int to byte
DiagnosticDescriptor: ERROR ConstantsMapperImpl.java:25 incompatible types: int cannot be converted to java.lang.Byte
DiagnosticDescriptor: ERROR ConstantsMapperImpl.java:26 incompatible types: possible lossy conversion from int to short
DiagnosticDescriptor: ERROR ConstantsMapperImpl.java:27 incompatible types: int cannot be converted to java.lang.Short
```

Therefore this must have some implicit meaning or impact (like you said, maybe a possible feature).

Since there are no test failures when the literal class is changed in `LongAnalyzer` and `DoubleAnalyzer`, it could also mean that the tests don't cover the uses case for double and long.

Maybe this needs a detailed test that does not only test the constant mapping, but also all kind of mappings between those possible primitive types and reference types, e.g.
```
int -> long
int -> Long
Integer -> long
Integer -> Long
long -> Long
Long -> Long
...
```

The [`ConstantsTarget`](https://github.com/mapstruct/mapstruct/blob/main/processor/src/test/java/org/mapstruct/ap/test/source/constants/ConstantsTarget.java) seems like a good example of what types could be mapped.

This might ensure that the removal of this (unnecessary) long cast does not break anything else.

What do you think?

filiphr:
To be honest I am. not sure why we are doing the cast for the double and long. Maybe @sjaakd would know.

I think that doing the changes for the `LongAnalyzer` and `DoubleAnalyzer` is fine, the others do need the casts, otherwise there are some other issues.

Also, I would remove the new tests. It is enough to have the changes in the `ConstantsMapperImpl` fixture.

dehasi:
@filiphr @thunderhook 
I changed `float` -> `double` in `DoubleAnalyzer`, deleted tests, and updated `ConstantsMapperImpl`.

filiphr:
Thanks for the changes @dehasi and thanks @thunderhook for merging it

PR #3438, opened 2023-11-10
Fix Typo Mappper in SubclassMapping Doc

wandi34:
Just a small fix for a typo in the JavaDoc I recognized during use of the lib.


thunderhook:
Thanks @wandi34!

PR #3445, opened 2023-11-20
Expression or condition 

### LINKED ISSUE #3413: Using `Mapping#expression` and `Mapping#conditionaQualifiedBy(Name)` should lead to compile error
### Expected behavior

Using a combination for `Mapping#expression` and `Mapping#conditionQualifiedBy(Name)` should lead to compile error

### Actual behavior

There is no compile error and the `Mapping#conditionQualifiedBy(Name)` is simple ignored

```
@Mapper
public interface MyMapper {
    @Mapping(target = "value", expression = "java(source.getValue())", conditionQualifiedByName = "condition")
    Target map(Source source);
}
```

This is related to #3405

### MapStruct Version

1.5.5

the-mgi:
fixes #3413

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
@@ -207,13 +207,13 @@ public class MappingOptions extends DelegatingOptions {
         if ( gem.source().hasValue() && gem.constant().hasValue() ) {
             message = Message.PROPERTYMAPPING_SOURCE_AND_CONSTANT_BOTH_DEFINED;
         }
-        else if (gem.expression().hasValue() && gem.conditionQualifiedByName().hasValue()) {
+        else if ( gem.expression().hasValue() && gem.conditionQualifiedByName().hasValue() ) {
             message = Message.PROPERTYMAPPING_EXPRESSION_AND_CONDITION_QUALIFIED_BY_NAME_BOTH_DEFINED;
         }
         else if ( gem.source().hasValue() && gem.expression().hasValue() ) {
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
```diff
@@ -207,6 +207,9 @@ private static boolean isConsistent(MappingGem gem, ExecutableElement method,
         if ( gem.source().hasValue() && gem.constant().hasValue() ) {
             message = Message.PROPERTYMAPPING_SOURCE_AND_CONSTANT_BOTH_DEFINED;
         }
+        else if (gem.expression().hasValue() && gem.conditionQualifiedByName().hasValue()) {
```
thunderhook, line 210: formatting nitpick: please add a whitespace (see the formatter in [CONTRIBUTING.md](https://github.com/mapstruct/mapstruct/blob/main/CONTRIBUTING.md)) - 6 lines beyond there is also a whitespace missing, would be nice if you could also add it there 

(wondering why checkstyle did not fail)

thunderhook (changes requested):
Hi @the-mgi!

Sorry it took so long to review. (now that I see it, 2 days isn't that long 😅 )

Would you be so kind and add a unit test? This would look similar to the[`Issue2668Test`](https://github.com/mapstruct/mapstruct/blob/main/processor/src/test/java/org/mapstruct/ap/test/bugs/_2668/Issue2668Test.java#L23-L42).

Thanks in advance!

the-mgi:
test case added, added proper whitespace

the-mgi:
FYI there is a failing test `mapstruct/processor/src/test/java/org/mapstruct/ap/test/collection/iterabletoset/FruitsMapper.java` on local on main branch, and on expression-or-condition.

thunderhook (approved):

PR #3460, opened 2023-12-02
Bump some versions for quick win

mosesonline:
I looked into the code to learn how it works and saw some old versions of libs which should be save to bump version in a minor version jump 1.5 ->1.6. I decided to provide this quick win.

Everything works fine.


thunderhook:
Thanks!

Issue #3306, opened 2023-06-14
Can not find symbol


TodyPlay:
### Expected behavior

![image](https://github.com/mapstruct/mapstruct/assets/54938830/29b17cf0-7f6f-4dc1-ac7e-62683ef8b8b7)
multilevel and use expression

### Actual behavior

![image](https://github.com/mapstruct/mapstruct/assets/54938830/1be24ac0-de67-46b7-b938-b885c22d7230)
just one symbol , so can not find symbol : options

### Steps to reproduce the problem

multilevel and use expression
` @Mapping(target = "questionAnalyse.questionDistributions", expression = "java(this.toDistributionList(questionPo, options))")`

### MapStruct Version

1.5.3.Final

Zegveld:
MapStruct is not going to analyse the contents of the expression. It will also not pass according to analysis unneeded parameters. So this behavior is expected for MapStruct. 

However if you annotate the argument `options` with `@Context` mapstruct will pass this on to submethods and then it does become available at the location where it currently is missing.

Hope this helps with getting this working for you.

TodyPlay:
![image](https://github.com/mapstruct/mapstruct/assets/54938830/1e8a0356-5c67-41df-8b46-33ae2cd81f8a)
@Zegveld thanks for you help. but new problem....

filiphr:
If you use `@Context` then the parameter will be passed in the methods below so it will fix one problem, but then it causes another problem because you are using it as a source in other places. You'll need to either wrap everything in one class and use that or pass the list twice (once as `@Context` and once as a normal mapping method). I would advise wrapping the 3 source parameters into one class.

Going to close this since it works as designed and it isn't an issue

Issue #3144, opened 2023-01-25
Breaking change: Mapping from Map to Bean


kirbylink:
### Expected behavior

When target and source object contains maps (Map<String, ?>)  with same name, then the new function "Mapping from Map to Bean" should not be used

Following Object is used with MapStruct 1.4.2.Final:
```java
@Getter
@Builder(setterPrefix = "with")
public class DispatchServiceRequest {
	private final String topic;
	private final String text;
	private final Map<String, Object> attachments;
	private final Date timestamp;
	private final UUID messageId;
	private final DispatcherService dispatcherService;
	private final Delay delay;
}
```
The corresponding Mapper that worked out of the box with MapStruct 1.4.2.Final:
```java
@Mapper(componentModel = "spring")
public interface DispatchServiceRequestMapper {
	@Mapping(target = "messageId", ignore = true)
	DispatchServiceRequest mapToDispatchServiceRequest(String topic, String text, Map<String, Object> attachments, Date timestamp, DispatcherService dispatcherService, Delay delay);
}
```

### Actual behavior

With version 1.5.x.Final i get the following errors:

> Multiple markers at this line
	- Can't map property "Object topic" to "String topic". Consider to declare/implement a mapping method: "String map(Object value)".
	- Can't map property "Object attachments" to "Map<String,Object> attachments". Consider to declare/implement a mapping method: "Map<String,Object> map(Object value)".
	- Can't map property "Object text" to "String text". Consider to declare/implement a mapping method: "String map(Object value)".
	- Can't map property "Object timestamp" to "Date timestamp". Consider to declare/implement a mapping method: "Date map(Object value)".
	- Can't map property "Object delay" to "Delay delay". Consider to declare/implement a mapping method: "Delay map(Object value)".
	- Can't map property "Object dispatcherService" to "Dispatcher.DispatcherService dispatcherService". Consider to declare/implement a mapping method: "Dispatcher.DispatcherService map(Object value)".

It seems, that the new Feature [3.10. Mapping Map to Bean](https://mapstruct.org/documentation/stable/reference/html/#mapping-map-to-bean) breaks the old behaviour.
When i remove the parameter `Map<String, Object> attachments` from the mapper method, it will compile.
When i add for each parameter a mapping like
```java
@Mapper(componentModel = "spring")
public interface DispatchServiceRequestMapper {
	@Mapping(target = "messageId", source = "")
	@Mapping(target = "text", source = "text")
[...]

Issue #3066, opened 2022-11-02
Can't use `@Mapping#source` for map fields containg the dot character


sajkol:
### Expected behavior

Using the new map->bean mapping feature it is not possible to use `@Mapping#source` for keys containing a `.` character. Instead, the `Map` type itself seems to be considered as the source type and its methods are proposed as alternatives.

It's still possible to achieve that with `expression` of course (as in the commented out line in the example), but it's far less elegant as just using the `source`.

### Actual behavior

```
error: No property named "a.b" exists in source parameter(s). Did you mean "a.bytes"?
        @Mapping(target = "value", source = "a.b")
```

### Steps to reproduce the problem

```
package com.example;

import org.junit.jupiter.api.Test;
import org.mapstruct.Mapper;
import org.mapstruct.Mapping;
import org.mapstruct.factory.Mappers;

import java.util.Map;

import static org.assertj.core.api.Assertions.assertThat;

class MapstructExampleTest {

    public static record ExampleBean(String value) { }

    @Mapper
    public interface ExampleMapper {

        @Mapping(target = "value", source = "a.b")
        //@Mapping(target = "value", expression = "java(source.get(\"a.b\"))")
        ExampleBean map(Map<String, String> source);
    }

    @Test
    void shouldMapToBean() {
        Map<String, String> source = Map.of("a.b", "test");
        var output = Mappers.getMapper(ExampleMapper.class).map(source);
        assertThat(output.value()).isEqualTo("test");
    }

}
```

### MapStruct Version

1.5.3

chenzijia12300:
Hi, I guess you can take a look at this [document](https://mapstruct.org/documentation/stable/reference/html/#mapping-map-to-bean)

sajkol:
Thank you, I have of course read through the documentation and described behavior seems to contradict what's written there:

> using the target bean properties (or defined through Mapping#source) to extract the values from the map

The example even shows usage of `Mapping#source` to directly name the field in the source map from which a value should be taken. There's no mention here of the `.` character being treated in a special way as it is for bean mappings - but in that case, it's used to support nested beans and the characters can't be used in a field name anyway. For String map keys it's no longer a special character and unless there is a plan to have support for multi-level maps as mapping sources, there's no point in having any special handling for it. Furthermore, the error message suggests the `Map` type itself is treated as source type (as opposed to map contents), which also supports the idea that this is not intended behavior.

chenzijia12300:
[...]

PR #3469, opened 2023-12-12
#3306 Support for map keys that contain dots when maps are mapped

thunderhook:
Tackling #3066 for map keys containing dots.
It was tricky to not reintroduce #3144 again.


filiphr:
@thunderhook how about something like `Map<String, Source>` if we use dotted then this would work properly right now, right? However, with this PR this will no longer work, right?

thunderhook:
I was wondering why there was no test handling this. I just tried the following:

```java
@Mapper
public interface MapToBeanFromMapWithSource {

    MapToBeanFromMapWithSource INSTANCE = Mappers.getMapper( MapToBeanFromMapWithSource.class );

    @Mapping(target = "targetName", source = "name")
    // @Mapping(target = "targetName", source = "source.name")
    // @Mapping(target = "targetName", source = "sourceA.name")
    // @Mapping(target = "targetName", source = "source.sourceA.name")
    Target toTarget(Map<String, Source> source);

    class Source {

        private String name;

        public String getName() {
            return name;
        }

        public void setName(String name) {
            this.name = name;
        }
    }

    class Target {

        String targetName;

        public String getTargetName() {
            return targetName;
        }

        public void setTargetName(String targetName) {
            this.targetName = targetName;
        }
    }

}
```

```java
    @ProcessorTest
    @WithClasses(MapToBeanFromMapWithSource.class)
    void shouldMap() {
        Map<String, MapToBeanFromMapWithSource.Source> sourceMap = new HashMap<>();
        MapToBeanFromMapWithSource.Source sourceA = new MapToBeanFromMapWithSource.Source();
        sourceA.setName( "value" );
        sourceMap.put( "sourceA", sourceA );
        sourceMap.put( "sourceB", new MapToBeanFromMapWithSource.Source() );

        MapToBeanFromMapWithSource.Target target = MapToBeanFromMapWithSource.INSTANCE.toTarget( sourceMap );

        assertThat( target.getTargetName() ).isEqualTo( "value" );
    }
```

And no matter what, it always shows:
```
DiagnosticDescriptor: ERROR MapToBeanFromMapWithSource.java:22
Can't map property "MapToBeanFromMapWithSource.Source name" to "String targetName".
Consider to declare/implement a mapping method: "String map(MapToBeanFromMapWithSource.Source value)
```

So it seems like this has never worked out of the box. But maybe I was doing something wrong. Do you have a working example in the main branch?

filiphr:
@thunderhook when I try your test on main it works when one of the following mappings are used

```java
@Mapping(target = "targetName", source = "sourceA.name")
@Mapping(target = "targetName", source = "source.sourceA.name")
```

The other two

```java
@Mapping(target = "targetName", source = "name")
@Mapping(target = "targetName", source = "source.name")
```

do not work because when using `name` the type that you get is `Source` and thus the compile error to provide a mapping from `Source` to `String`. When using `source.name` it is the same as using `name` since the source parameter is called `source`.

thunderhook:
Thanks for testing. Looks like I made a typo or something.
Yes, you're right. It works on `main`, but not on my branch.

To be backwards compatible, we would need some sort of control, whether to use a dot as a parameter segment or as a key containing a dot. But that leads to other complexities.
Think of nested maps, where the top level needs to be treated as a parameter segment, and the nested map as a string. An intermediate method would be needed, and nesting like this would be error-prone: `sourceA.innerMap.some.key.with.dots.person.firstName`, which should be separated like this:
`sourceA - innerMap - some.key.with.dots - person - firstName`.

But, there may be another way: Using the dot is a common and regular pattern  for parameter segments. It is more of an exception to use it as a string like `"a.b"`. So the user may need to escape the dot, if it is meant to be a string, such as: `"a\.b"`.
This is the same with regular expressions where the dot is also meant as a wildcard, and therefore needs to be escaped.

WDYT? Is there already some kind of escaping? I only know of the MappingConstants with their `<...>` naming syntax. But I don't think a qualifier like this such as `<DOT>` would be helpful.

thunderhook:
Here's my take with the escaped character. It is now possible to uses nested mappings with dotted keys as I stated in my previous comment. Please have a look.

filiphr:
I think having a special escape character like there is in regular expressions is completely fine. I'd try to find some time to review this. I would like to first focus on some of the remaining things for 1.6 before start to add some new functionality as well.

PR #3483, opened 2023-12-25
#1064 Provide a switch to turn off CheckStyle on generated sources

### LINKED ISSUE #1064: Provide a switch to turn off CheckStyle on generated sources
The CheckStyle run for the generated sources consumes a considerable amount of time. Disabling it in the code (`CompilingStatement`) cuts down build time from 2 minutes to 1 minute for the processor module for me. Hence having an option to turn it off would be highly desirable, ideally it'd be just controlled via the existing option `-Dcheckstyle.skip=true`.

thunderhook:
Skip checkstyle ifthe system property `checkstyle.skip[=true]` is set (except in the rare case that someone sets it to `checkstyle.skip=false`).

This reducces my build time from 7min 34sec to 6min 49sec. An improvement about 11%. 👍 

Fixes #1064

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingExtension.java
+++ b/processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingExtension.java
@@ -188,10 +188,7 @@ abstract class CompilingExtension implements BeforeEachCallback {
     }
 
     private static boolean skipCheckstyleBySystemProperty() {
-
-        String skipCheckstyleProperty = System.getProperty( "checkstyle.skip" );
-
-        return skipCheckstyleProperty != null && !skipCheckstyleProperty.equals( "false" );
+        return Boolean.parseBoolean( System.getProperty( "checkstyle.skip" ) );
     }
 
     private void assertCheckstyleRules() throws Exception {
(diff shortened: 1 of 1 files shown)


processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilingExtension.java
```diff
@@ -182,11 +182,18 @@ private void assertResult(CompilationOutcomeDescriptor actualResult, ExtensionCo
...
+        if ( !findAnnotation( testClass, DisableCheckstyle.class ).isPresent() && !skipCheckstyleBySystemProperty() ) {
             assertCheckstyleRules();
         }
     }
 
+    private static boolean skipCheckstyleBySystemProperty() {
+
+        String skipCheckstyleProperty = System.getProperty( "checkstyle.skip" );
+
+        return skipCheckstyleProperty != null && !skipCheckstyleProperty.equals( "false" );
```
filiphr, line 194:
```suggestion
    private static boolean skipCheckstyleBySystemProperty() {
        return Boolean.parseBoolean( System.getProperty( "checkstyle.skip" ) );
```

filiphr (commented):
Looks good to me. I have one small simplification suggestion though

Issue #3360, opened 2023-08-14
Subclass mapping warns about unmapped property that is mapped in referenced mapper


peterjeschke:
### Expected behavior

The provided example should compile successfully without warnings. Also, no unused code should be generated

### Actual behavior

In the code example, the `AbstractBoDtoMapper` is configured to cause a compile error for unmapped properties. The generated code uses the mapper for the concrete subclasses (`AImplMapper` and `BImplMapper`), which both properly map all properties, including the property "legacyMyString" which needs manual mapping. Even though everything should be correct (and indeed, the generated code works as expected, if the error is disabled), Mapstruct thinks that the property "legacyMyString" was not mapped and thus shows an error.

Additionally, if the unmapped*Policies are set to default again (e.g. by removing the lines in the `@Mapper` annotation), the generated `AbstractBoDtoMapperImpl` contains code that isn't actually used anywhere:

```java


    protected ExampleObjectBo exampleObjectDtoToExampleObjectBo(ExampleObjectDto exampleObjectDto) {
        if (exampleObjectDto == null) {
            return null;
        } else {
            ExampleObjectBo exampleObjectBo = new ExampleObjectBo();
            exampleObjectBo.setExampleString(exampleObjectDto.getExampleString());
            return exampleObjectBo;
        }
    }

    protected ExampleObjectDto exampleObjectBoToExampleObjectDto(ExampleObjectBo exampleObjectBo) {
        if (exampleObjectBo == null) {
            return null;
        } else {
            ExampleObjectDto exampleObjectDto = new ExampleObjectDto();
            exampleObjectDto.setExampleString(exampleObjectBo.getExampleString());
            return exampleObjectDto;
        }
    }
```

Since the code isn't used, it shouldn't be generated by Mapstruct

### Steps to reproduce the problem

See https://github.com/peterjeschke/mapstruct-reproducer for a small example. The error is shown when compiling the project

### MapStruct Version

1.5.5.Final

filiphr:
Thanks for the detailed bug report @peterjeschke.

This is indeed a bug that we need to look into and fix. There are indeed 2 problems, but they might be linked to the same root cause.

When the target classes in `AbstractBoDtoMapper` are not `abstract` then the warning for the unmapped is valid and the previously unused methods are invalid.

We'll look into it and fix it.

filiphr:
We should look into #3331 when working on this fix

PR #3487, opened 2023-12-27
#3331 Do not handle defined mappings if the result type is abstract due to runtime exception subclass exhaustive strategy

### LINKED ISSUE #3331: Abstract class fields mapping issues
### Expected behavior

Mapstruct can map fields of abstract classes

### Actual behavior

Mapstruct is not mapping fields of abstract classes - various problems occur

Source classes:
```java
public abstract class Vehicle {
	private final String name;
	Vehicle(String name) {
		this.name = name;
	}
	public String getName() {
		return name;
	}
}

public class Car extends Vehicle {
	private final int numOfDoors;
	Car(String name, int numOfDoors) {
		super(name);
		this.numOfDoors = numOfDoors;
	}
	public int getNumOfDoors() {
		return numOfDoors;
	}
}

public class Motorbike extends Vehicle {
	private final boolean allowedForMinor;
	Motorbike(String name, boolean allowedForMinor) {
		super(name);
		this.allowedForMinor = allowedForMinor;
	}
	public boolean isAllowedForMinor() {
		return allowedForMinor;
	}
}
```

Target classes:
```java
public abstract class VehicleDto {
	private final String name;
	public VehicleDto(String name) {
		this.name = name;
	}
}

public class CarDto extends VehicleDto {
	private final int numOfDoors;
	public CarDto(String name, int numOfDoors) {
		super(name);
		this.numOfDoors = numOfDoors;
	}
}

public class MotorbikeDto extends VehicleDto {
	private final boolean allowedForMinor;
	public MotorbikeDto(String name, boolean allowedForMinor) {
		super(name);
		this.allowedForMinor = allowedForMinor;
	}
}
```

Mapper config:
```java
@Mapper(subclassExhaustiveStrategy = SubclassExhaustiveStrategy.RUNTIME_EXCEPTION)
public interface VehicleMapper {
	@SubclassMapping(target = CarDto.class, source = Car.class)
	@SubclassMapping(target = MotorbikeDto.class, source = Motorbike.class)
	VehicleDto mapToDto(Vehicle vehicle);
}
```

Everything works fine so far, the mapper class is generated and mapping correctly all the fields. 

1.
The problem occurs when I'm trying to add anything related to name field in VehicleDto, for example:
```java
@Mapper(subclassExhaustiveStrategy = SubclassExhaustiveStrategy.RUNTIME_EXCEPTION)
public interface VehicleMapper {
	@SubclassMapping(target = CarDto.class, source = Car.class)
	@SubclassMapping(target = MotorbikeDto.class, source = Motorbike.class)
	@Mapping(target = "name", constant = "noname") // <-------- added this one --------
	VehicleDto mapToDto(Vehicle vehicle);
}
```

In this case I'm getting a compilation error:
>Unknown property "name" in result type VehicleDto. Did you mean "null"?

2.
If I also add getter to the abstract target class:
```java
public abstract class VehicleDto {
	private final String name;
	public VehicleDto(String name) {
		this.name = name;
	}
	public String getName() {
		return name;
	}
}
```
It's still not working but the build error is different:
>Property "name" has no write accessor in VehicleDto.

Isn't Mapstruct supposed to handle both these cases correctly and map the name field?

I'm using Java 17, Spring Boot 3.1.1, no Lombok.

### MapStruct Version

1.5.5.Final

filiphr:
Fixes #3331

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -284,11 +284,10 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
             initializeMappingReferencesIfNeeded( resultTypeToMap );
 
-            boolean hasSubclassMappingsWithAllSupportedClasses = hasSubclassMappingsWithAllSupportedClasses(
-                resultTypeToMap );
+            boolean shouldHandledDefinedMappings = shouldHandledDefinedMappings( resultTypeToMap );
 
 
-            if ( hasSubclassMappingsWithAllSupportedClasses ) {
+            if ( shouldHandledDefinedMappings ) {
                 // map properties with mapping
                 boolean mappingErrorOccurred = handleDefinedMappings( resultTypeToMap );
                 if ( mappingErrorOccurred ) {
@@ -1052,8 +1051,8 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         }
 
         /**
-         * Determine whether the result type has subclass mappings with all supported classes.
-         * It has if any of the following is true:
+         * Determine whether defined mappings should be handled on the result type.
+         * They should be, if any of the following is true:
          * <ul>
          *     <li>The {@code resultTypeToMap} is not abstract</li>
          *     <li>There is a factory method</li>
@@ -1063,9 +1062,9 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
          * and the chosen strategy is runtime exception.
          *
          * @param resultTypeToMap the type in which the defined target properties are defined
-         * @return {@code true} if there are subclass mappings with all supported classes, {@code false} otherwise
+         * @return {@code true} if defined mappings should be handled for the result type, {@code false} otherwise
          */
-        private boolean hasSubclassMappingsWithAllSupportedClasses(Type resultTypeToMap) {
+        private boolean shouldHandledDefinedMappings(Type resultTypeToMap) {
             if ( !resultTypeToMap.isAbstract() ) {
                 return true;
             }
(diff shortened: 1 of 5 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -284,10 +284,16 @@ else if ( !method.isUpdateMethod() ) {
...
 
-            // map properties with mapping
-            boolean mappingErrorOccurred = handleDefinedMappings( resultTypeToMap );
-            if ( mappingErrorOccurred ) {
-                return null;
+            boolean hasSubclassMappingsWithAllSupportedClasses = hasSubclassMappingsWithAllSupportedClasses(
+                resultTypeToMap );
+
+
+            if ( hasSubclassMappingsWithAllSupportedClasses ) {
```
thunderhook, line 291: Maybe there is another way to check if the `SubclassExhaustiveStrategy.RUNTIME_EXCEPTION` option is set? Maybe via `isAbstractReturnTypeAllowed()`?

I mean, perhaps this method should only be called, when the `SubclassExhaustiveStrategy.RUNTIME_EXCEPTION` is set, and not in every bean mapping method? This could have side effects in future code changes.

e.g.:
```java
if ( isRuntimeExceptionSubclassExhaustiveStrategy
     && hasSubclassMappingsWithAllSupportedClasses(...) ) {
  // ...
}
```

filiphr: You are right. It didn't quite feel right to use this naming. I'll think of a better solution for this

filiphr: I've renamed this. Is it clearer now?

filiphr: Going back a bit through some of the issues and my commits, I realized why I named it like I did. I am using that in https://github.com/filiphr/mapstruct/commit/d5cc43a19552967c7c682b9a1daec994a771445d which is a fix for #3360. @thunderhook do you think that the new local variable name is clear for that other commit as well?

thunderhook:
Looks good to me. Thanks for renaming the variable.

PR #3492, opened 2023-12-28
#3463 DefaultBuilderProvider should be able to handle methods in parent interfaces

### LINKED ISSUE #3463: Interface inherited build method is not found
### Expected behavior

I have a weird setup, I'm mapping from my own dto to a libraries equivalent dto. The library has builders for its entities, however their builders are exposed via an interface that extends another interface that contains a generic build method. I would expect this build method to be found and used. 

### Actual behavior

I get a compilation error "Person does not have an accessible constructor."

```java
// provided by library
public interface EntityBuilder<T> {
    T build();
}

// provided by library
public class Person {
    private final String name;

    private Person(String name) {
        this.name = name;
    }

    public static Builder builder() {
        return new BuilderImpl();
    }

    public interface Builder extends EntityBuilder<Person> {
        Builder name(String name);
    }

    private static final class BuilderImpl implements Builder {
        private String name;

        private BuilderImpl() {
        }

        public Builder name(String name) {
            this.name = name;
            return this;
        }

        @Override
        public Person build() {
            return new Person(name);
        }
    }
}

//my own dto
public record PersonDto(String name) {
}

@Mapper
public interface PersonMapper {
    Person map(PersonDto source);
}
```

From my quick debugging it looks like the problem is that https://github.com/mapstruct/mapstruct/blob/main/processor/src/main/java/org/mapstruct/ap/spi/DefaultBuilderProvider.java#L240 does not yield the `build` method

### MapStruct Version

1.5.5.Final

filiphr:
Fixes #3463

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/spi/DefaultBuilderProvider.java
+++ b/processor/src/main/java/org/mapstruct/ap/spi/DefaultBuilderProvider.java
@@ -239,15 +239,12 @@ public class DefaultBuilderProvider implements BuilderProvider {
      * Searches for a build method for {@code typeElement} within the {@code builderElement}.
      * <p>
      * The default implementation iterates over each method in {@code builderElement} and uses
-     * {@link DefaultBuilderProvider#isBuildMethod(ExecutableElement, DeclaredType, TypeElement)} to check if the method is a
-     * build method for {@code typeElement}.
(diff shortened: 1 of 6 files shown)


thunderhook (approved):
Looks good to me. I was wondering if something should be added to the docs. But there is nothing stated about parent (builder) classes or interfaces in general. So I think it's fine to leave it as it is.

PR #3498, opened 2024-01-01
add missing generic to MappingOptions

hduelme:


filiphr:
Good catch @hduelme. Thanks

PR #3499, opened 2024-01-01
add missing generic type to Javadoc Builder

hduelme:
Currently there the type of `authors` is missing. I added it as `List<String>`


filiphr:
Thanks @hduelme

PR #3500, opened 2024-01-02
Fix typo chapter-10-advanced-mapping-options.asciidoc

ChanyutJean:
Typo at `@MaperConfig`.


filiphr:
Good catch. Thanks @ChanyutJean

PR #3501, opened 2024-01-02
use primitive types in NativeTypes

hduelme:
`Double.parseDouble` returns a primitive `double` and `Float.parseFloat` returns a primitive`float`. By using the primitive types directly instead of the wrapper types unnecessary boxing can be avoided.


filiphr:
Thanks @hduelme

PR #3506, opened 2024-01-03
#3495 when updating targets with collections without a setter, clear them only when null-check or condition is satisfied

### LINKED ISSUE #3495: Clear immutable collection only when condition is met
### Expected behavior

```java
if ( h5s.getPurchasedProducts() != null ) {
  if ( savePurchasedProducts( proto.getPurchasedProductsList() ) ) {
      h5s.getPurchasedProducts().clear();
      h5s.getPurchasedProducts().addAll( purchasedProductInputProtoListToPurchasedProductList( proto.getPurchasedProductsList() ) );
  }
}
```

### Actual behavior

```java
if ( h5s.getPurchasedProducts() != null ) {
  h5s.getPurchasedProducts().clear();
  if ( savePurchasedProducts( proto.getPurchasedProductsList() ) ) {
      h5s.getPurchasedProducts().addAll( purchasedProductInputProtoListToPurchasedProductList( proto.getPurchasedProductsList() ) );
  }
}
```

in the generated code the destination collection is always emptied even if the mapping condition is false, I would expect it not to be touched as happens for the other fields.

### MapStruct Version

1.5.5.Final

thunderhook:
With this change, updating targets without a setter now works exactly like updating a target with a collection setter. So hopefully no more confusion with this.

This breaks a very old test, see `Issue289Test`. However, I feel that this was always a bug. @filiphr Let me know what you think.

Closes #3495

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/GetterWrapperForCollectionsAndMaps.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/assignment/GetterWrapperForCollectionsAndMaps.ftl
@@ -10,10 +10,10 @@
 <@lib.sourceLocalVarAssignment/>
 if ( ${ext.targetBeanName}.${ext.targetWriteAccessorName}<@lib.handleWriteAccesing /> != null ) {
     <@lib.handleExceptions>
-      <#if ext.existingInstanceMapping>
-        ${ext.targetBeanName}.${ext.targetWriteAccessorName}<@lib.handleWriteAccesing />.clear();
-      </#if>
       <@lib.handleLocalVarNullCheck needs_explicit_local_var=false>
+        <#if ext.existingInstanceMapping>
+          ${ext.targetBeanName}.${ext.targetWriteAccessorName}<@lib.handleWriteAccesing />.clear();
+        </#if>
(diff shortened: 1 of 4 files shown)


filiphr:
I don't agree that this was a bug. Let's take scenario from the `Issue289Test`. The generated code that is on main looks like:

```java
    target.getCollection().clear();
    Collection<TargetElement> collection = sourceElementCollectionToTargetElementCollection( source.getCollection() );
    if ( collection != null ) {
        target.getCollection().addAll( collection );
    }
}
```

This means that if target is set we need to update its values with whatever is in the source collection. This means that we clear the existing collection and then we add the existing elements from the source. With this change we can no longer remove values from the target collection.


If we look at what is happening when there is a setter

```java
    Collection<TargetElement> collection = sourceElementCollectionToTargetElementCollection( source.getCollection() );
    if ( collection != null ) {
        target.getCollection().clear();
        target.getCollection().addAll( collection );
    }
    else {
        target.setCollection( null );
    }
}
else {
    Collection<TargetElement> collection = sourceElementCollectionToTargetElementCollection( source.getCollection() );
    if ( collection != null ) {
        target.setCollection( collection );
    }
}
```

With this if the target collection is not null then we either clear the collection (when source is there) or we just set it to `null`. When the target is null we only set the source collection if it isn't null.

I'll post in the issue as well.

thunderhook:
Closed, see https://github.com/mapstruct/mapstruct/issues/3495#issuecomment-1937548261.

PR #3508, opened 2024-01-07
#3323 Support access to the source property name

### LINKED ISSUE #3323: Add `@SourcePropertyName` to handle a property name of the source object
A suggestion from https://github.com/mapstruct/mapstruct/issues/3313#issuecomment-1627449047

thunderhook:
Finally took the time to figure it out. I used the patch of @aboqasem mentioned in https://github.com/mapstruct/mapstruct/issues/3323#issuecomment-1763729690 and finished the implementation. I think he should also be mentioned as co-author.

Closes #3323

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -70,6 +70,7 @@ import static org.mapstruct.ap.internal.model.common.Assignment.AssignmentType.D
 public class PropertyMapping extends ModelElement {
 
     private final String name;
+    private final String sourcePropertyName;
     private final String sourceBeanName;
     private final String targetWriteAccessorName;
     private final ReadAccessor targetReadAccessorProvider;
@@ -286,6 +287,7 @@ public class PropertyMapping extends ModelElement {
             }
 
             return new PropertyMapping(
+                sourcePropertyName,
                 targetPropertyName,
                 rightHandSide.getSourceParameterName(),
                 targetWriteAccessor.getSimpleName(),
@@ -1099,16 +1101,17 @@ public class PropertyMapping extends ModelElement {
         ReadAccessor targetReadAccessorProvider,
         Type targetType, Assignment propertyAssignment,
         Set<String> dependsOn, Assignment defaultValueAssignment, boolean constructorMapping) {
-        this( name, null, targetWriteAccessorName, targetReadAccessorProvider,
+        this( name, name, null, targetWriteAccessorName, targetReadAccessorProvider,
             targetType, propertyAssignment, dependsOn, defaultValueAssignment,
             constructorMapping
         );
     }
 
-    private PropertyMapping(String name, String sourceBeanName, String targetWriteAccessorName,
-        ReadAccessor targetReadAccessorProvider, Type targetType,
-        Assignment assignment,
-        Set<String> dependsOn, Assignment defaultValueAssignment, boolean constructorMapping) {
+    private PropertyMapping(String sourcePropertyName, String name, String sourceBeanName,
+                            String targetWriteAccessorName, ReadAccessor targetReadAccessorProvider, Type targetType,
+                            Assignment assignment,
+                            Set<String> dependsOn, Assignment defaultValueAssignment, boolean constructorMapping) {
+        this.sourcePropertyName = sourcePropertyName;
         this.name = name;
         this.sourceBeanName = sourceBeanName;
         this.targetWriteAccessorName = targetWriteAccessorName;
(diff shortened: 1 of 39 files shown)


aboqasem:
Thanks a lot for the acknowledgement, @thunderhook!

I will have to take a look at it sometime to learn how it was implemented.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -1099,16 +1101,17 @@ private PropertyMapping(String name, String targetWriteAccessorName,
         ReadAccessor targetReadAccessorProvider,
         Type targetType, Assignment propertyAssignment,
         Set<String> dependsOn, Assignment defaultValueAssignment, boolean constructorMapping) {
-        this( name, null, targetWriteAccessorName, targetReadAccessorProvider,
+        this( name, name, null, targetWriteAccessorName, targetReadAccessorProvider,
```
filiphr, line 1104: I don't think that we need to pass the name as a source here. Constant expressions do not have a source property.

processor/src/test/java/org/mapstruct/ap/test/conditional/propertyname/sourcepropertyname/ConditionalMethodForCollectionMapperWithSourcePropertyName.java
```diff
@@ -0,0 +1,47 @@
...
+
+    ConditionalMethodForCollectionMapperWithSourcePropertyName INSTANCE
+        = Mappers.getMapper( ConditionalMethodForCollectionMapperWithSourcePropertyName.class );
+
+    @Mapping(target = "country", source = "originCountry")
+    Employee map(EmployeeDto employee);
+
+    @Condition
+    default <T> boolean isNotEmpty(Collection<T> collection, @SourcePropertyName String propName) {
+        if ( "addresses".equalsIgnoreCase( propName ) ) {
```
filiphr, line 34: Perhaps we need a test where the source and target property names are different?

processor/src/test/java/org/mapstruct/ap/test/conditional/propertyname/sourcepropertyname/ConditionalMethodInMapperWithSourcePropertyName.java
```diff
@@ -0,0 +1,38 @@
...
+
+    ConditionalMethodInMapperWithSourcePropertyName INSTANCE
+        = Mappers.getMapper( ConditionalMethodInMapperWithSourcePropertyName.class );
+
+    @Mapping(target = "country", source = "originCountry")
+    Employee map(EmployeeDto employee);
+
+    @Condition
+    default boolean isNotBlank(String value, @SourcePropertyName String propName) {
+        if ( propName.equalsIgnoreCase( "lastName" ) ) {
```
filiphr, line 33: Perhaps we use a source that has a different source name?

processor/src/test/java/org/mapstruct/ap/test/conditional/propertyname/sourcepropertyname/ConditionalMethodInUsesMapperWithSourcePropertyName.java
```diff
@@ -0,0 +1,42 @@
...
+        = Mappers.getMapper( ConditionalMethodInUsesMapperWithSourcePropertyName.class );
+
+    @Mapping(target = "country", source = "originCountry")
+    Employee map(EmployeeDto employee);
+
+    class PresenceUtils {
+
+        @Condition
+        public boolean isNotBlank(String value, @SourcePropertyName String propName) {
+            if ( propName.equalsIgnoreCase( "lastName" ) ) {
```
filiphr, line 35: Perhaps we use a source that has a different property name?

filiphr (commented):
Looks good to me. I think some small changes could be done though to make the tests clearer. Let me know what you think @thunderhook

thunderhook:
Thank you! Good points. I have rewritten the tests so that the collection also has a different name.
But now the mapping seems to be generated in a different order? At least `PresenceUtilsAllPropsWithSource` now behaves strangely and sometimes removes the `boss.` prefix from the path.

I'm not sure how to solve it. I don't quite understand the logic that is implemented with those `Deque`s.
A possible approach is to use a source name that behaves the same (makes the test less readable).
WDYT?

thunderhook:
Got it. As suspected, this problem was caused by a reordered mapping method. The `PresenceUtilsAllPropsWithSource` worked by accident because the nested address mapping happened at the end. The `@AfterMapping after` method was called by the intermediate method that handled the collection, and therefore went up the path twice.

@filiphr Could you please take another look? Thanks in advance.

filiphr:
Thanks @thunderhook. The changes looked good, I've merged this

PR #3522, opened 2024-02-08
#2788 unmapped source properties message improved for forged methods

### LINKED ISSUE #2788: Improving error messages for auto generated mappings

The documentation of `Mapper#disableSubMappingMethodsGeneration` states that if one had to use it, the case should be reported, and here we go.

In our case we had a deeply nested hierarchy of types. For each of those types, a corresponding DTOs exists in another package, but these DTOs do not necessarily contain all of the fields of the source types. When working with the default setting of `disableSubMappingMethodsGeneration == false`, the error messages contain only the raw names of the unmapped source fields, but do not yield the information to which type each of the unmapped fields belongs. Hence turning off auto generation made it a lot easier to find the corresponding culprits. Maybe the error messages could be improved to something like `Unmapped source property: typeA.typeB.foo`.

Zegveld:
updated the message for unmapped source properties to have the same structure as the unmapped target properties messages.
fixes #2788

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -1682,14 +1682,16 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             }
             else if ( !unprocessedTargetProperties.isEmpty() && unmappedTargetPolicy.requiresReport() ) {
 
-                Message unmappedPropertiesMsg =
-                                unmappedTargetPolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
-                                Message.BEANMAPPING_UNMAPPED_TARGETS_ERROR :
-                                Message.BEANMAPPING_UNMAPPED_TARGETS_WARNING;
-                Message unmappedForgedPropertiesMsg =
-                                unmappedTargetPolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
-                                Message.BEANMAPPING_UNMAPPED_FORGED_TARGETS_ERROR :
-                                Message.BEANMAPPING_UNMAPPED_FORGED_TARGETS_WARNING;
+                Message unmappedPropertiesMsg;
+                Message unmappedForgedPropertiesMsg;
+                if ( unmappedTargetPolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ) {
+                    unmappedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_TARGETS_ERROR;
+                    unmappedForgedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_FORGED_TARGETS_ERROR;
+                }
+                else {
+                    unmappedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_TARGETS_WARNING;
+                    unmappedForgedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_FORGED_TARGETS_WARNING;
+                }
 
                 reportErrorForUnmappedProperties(
                     unprocessedTargetProperties,
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -1713,14 +1713,16 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         private void reportErrorForUnmappedSourcePropertiesIfRequired() {
             ReportingPolicyGem unmappedSourcePolicy = getUnmappedSourcePolicy();
             if ( !unprocessedSourceProperties.isEmpty() && unmappedSourcePolicy.requiresReport() ) {
-                Message unmappedPropertiesMsg =
-                                unmappedSourcePolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
-                                Message.BEANMAPPING_UNMAPPED_SOURCES_ERROR :
-                                Message.BEANMAPPING_UNMAPPED_SOURCES_WARNING;
-                Message unmappedForgedPropertiesMsg =
-                                unmappedSourcePolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
-                                Message.BEANMAPPING_UNMAPPED_FORGED_SOURCES_ERROR :
-                                Message.BEANMAPPING_UNMAPPED_FORGED_SOURCES_WARNING;
+                Message unmappedPropertiesMsg;
+                Message unmappedForgedPropertiesMsg;
+                if ( unmappedSourcePolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ) {
+                    unmappedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_SOURCES_ERROR;
+                    unmappedForgedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_FORGED_SOURCES_ERROR;
+                }
+                else {
+                    unmappedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_SOURCES_WARNING;
+                    unmappedForgedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_FORGED_SOURCES_WARNING;
+                }
 
                 reportErrorForUnmappedProperties(
                     unprocessedSourceProperties,
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -1682,53 +1682,19 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             }
             else if ( !unprocessedTargetProperties.isEmpty() && unmappedTargetPolicy.requiresReport() ) {
 
-                if ( !( method instanceof ForgedMethod ) ) {
-                    Message msg = unmappedTargetPolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
-                        Message.BEANMAPPING_UNMAPPED_TARGETS_ERROR : Message.BEANMAPPING_UNMAPPED_TARGETS_WARNING;
-                    Object[] args = new Object[] {
-                        MessageFormat.format(
-                            "{0,choice,1#property|1<properties}: \"{1}\"",
-                            unprocessedTargetProperties.size(),
-                            Strings.join( unprocessedTargetProperties.keySet(), ", " )
-                        )
-                    };
+                Message unmappedPropertiesMsg =
+                                unmappedTargetPolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
+                                Message.BEANMAPPING_UNMAPPED_TARGETS_ERROR :
+                                Message.BEANMAPPING_UNMAPPED_TARGETS_WARNING;
+                Message unmappedForgedPropertiesMsg =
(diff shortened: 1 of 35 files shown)


processor/src/test/java/org/mapstruct/ap/test/accessibility/referenced/ReferencedAccessibilityTest.java
```diff
@@ -21,6 +21,7 @@
  * @author Sjaak Derksen
  */
 @WithClasses( { Source.class, Target.class, ReferencedSource.class, ReferencedTarget.class } )
+@IssueKey("2788")
```
filiphr, line 24: Why don't we create a dedicated test for this?

processor/src/test/java/org/mapstruct/ap/test/accessibility/referenced/SourceTargetMapperDefaultOther.java
```diff
@@ -7,14 +7,15 @@
...
+import org.mapstruct.ReportingPolicy;
 import org.mapstruct.ap.test.accessibility.referenced.a.ReferencedMapperDefaultOther;
 import org.mapstruct.factory.Mappers;
 
 /**
  *
  * @author Sjaak Derksen
  */
-@Mapper(uses = ReferencedMapperDefaultOther.class)
+@Mapper(uses = ReferencedMapperDefaultOther.class, unmappedSourcePolicy = ReportingPolicy.WARN)
```
filiphr, line 18: Why do we need to change an existing test mapper?

processor/src/test/java/org/mapstruct/ap/test/accessibility/referenced/SourceTargetMapperDefaultSame.java
```diff
@@ -7,13 +7,14 @@
...
 import org.mapstruct.Mapping;
+import org.mapstruct.ReportingPolicy;
 import org.mapstruct.factory.Mappers;
 
 /**
  *
  * @author Sjaak Derksen
  */
-@Mapper(uses = ReferencedMapperDefaultSame.class)
+@Mapper(uses = ReferencedMapperDefaultSame.class, unmappedSourcePolicy = ReportingPolicy.WARN)
```
filiphr, line 17: Why do we need to change an existing test mapper?

processor/src/test/java/org/mapstruct/ap/test/accessibility/referenced/SourceTargetMapperPrivate.java
```diff
@@ -7,13 +7,14 @@
...
 import org.mapstruct.Mapping;
+import org.mapstruct.ReportingPolicy;
 import org.mapstruct.factory.Mappers;
 
 /**
  *
  * @author Sjaak Derksen
  */
-@Mapper(uses = ReferencedMapperPrivate.class)
+@Mapper(uses = ReferencedMapperPrivate.class, unmappedSourcePolicy = ReportingPolicy.WARN)
```
filiphr, line 17: Why do we need to change an existing test mapper?

processor/src/test/java/org/mapstruct/ap/test/accessibility/referenced/SourceTargetMapperProtected.java
```diff
@@ -7,13 +7,14 @@
...
 import org.mapstruct.Mapping;
+import org.mapstruct.ReportingPolicy;
 import org.mapstruct.factory.Mappers;
 
 /**
  *
  * @author Sjaak Derksen
  */
-@Mapper(uses = ReferencedMapperProtected.class)
+@Mapper(uses = ReferencedMapperProtected.class, unmappedSourcePolicy = ReportingPolicy.WARN)
```
filiphr, line 17: Why do we need to change an existing test mapper?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1682,53 +1682,19 @@ private void reportErrorForUnmappedTargetPropertiesIfRequired() {
...
-                        )
-                    };
+                Message unmappedPropertiesMsg =
+                                unmappedTargetPolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
+                                Message.BEANMAPPING_UNMAPPED_TARGETS_ERROR :
+                                Message.BEANMAPPING_UNMAPPED_TARGETS_WARNING;
+                Message unmappedForgedPropertiesMsg =
+                                unmappedTargetPolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
+                                Message.BEANMAPPING_UNMAPPED_FORGED_TARGETS_ERROR :
+                                Message.BEANMAPPING_UNMAPPED_FORGED_TARGETS_WARNING;
```
filiphr, line 1692:
```suggestion
                Message unmappedPropertiesMsg;
                Message unmappedForgedPropertiesMsg;
                if ( unmappedTargetPolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ) {
                    unmappedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_TARGETS_ERROR;
                    unmappedForgedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_FORGED_TARGETS_ERROR;
                }
                else {
                    unmappedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_TARGETS_WARNING;
                    unmappedForgedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_FORGED_TARGETS_WARNING;
                }
```

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1746,22 +1712,65 @@ private ReportingPolicyGem getUnmappedSourcePolicy() {
...
-                Message msg = unmappedSourcePolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
-                    Message.BEANMAPPING_UNMAPPED_SOURCES_ERROR : Message.BEANMAPPING_UNMAPPED_SOURCES_WARNING;
+                Message unmappedPropertiesMsg =
+                                unmappedSourcePolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
+                                Message.BEANMAPPING_UNMAPPED_SOURCES_ERROR :
+                                Message.BEANMAPPING_UNMAPPED_SOURCES_WARNING;
+                Message unmappedForgedPropertiesMsg =
+                                unmappedSourcePolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ?
+                                Message.BEANMAPPING_UNMAPPED_FORGED_SOURCES_ERROR :
+                                Message.BEANMAPPING_UNMAPPED_FORGED_SOURCES_WARNING;
```
filiphr, line 1723:
```suggestion
                Message unmappedPropertiesMsg;
                Message unmappedForgedPropertiesMsg;
                if ( unmappedSourcePolicy.getDiagnosticKind() == Diagnostic.Kind.ERROR ) {
                    unmappedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_SOURCES_ERROR;
                    unmappedForgedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_FORGED_SOURCES_ERROR;
                }
                else {
                    unmappedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_SOURCES_WARNING;
                    unmappedForgedPropertiesMsg = Message.BEANMAPPING_UNMAPPED_FORGED_SOURCES_WARNING;
                }
```

filiphr (changes requested):
main changes look OK, I've added some suggestions to reduce the duplication a bit.

I've left some comments about the test changes. I am not a fan of changing existing tests, why are we doing that? Why don't we write dedicated tests for this?

Zegveld:
> main changes look OK, I've added some suggestions to reduce the duplication a bit.
> 
> I've left some comments about the test changes. I am not a fan of changing existing tests, why are we doing that? Why don't we write dedicated tests for this?

Not changing the existing test will make it less visible what is being tested by the test.
The current test class `ReferencedAccessibilityTest.java` should test both target and source messages, otherwise the test class is incorrectly named in my opinion.

I can split it to be more precise, so that we get separate tests for source and target. This will be obvious in the classes then.
These three categories will be visible then:
* No messages
* Target messages
* Source messages

Zegveld:
> > main changes look OK, I've added some suggestions to reduce the duplication a bit.
> > I've left some comments about the test changes. I am not a fan of changing existing tests, why are we doing that? Why don't we write dedicated tests for this?
> 
> Not changing the existing test will make it less visible what is being tested by the test. The current test class `ReferencedAccessibilityTest.java` should test both target and source messages, otherwise the test class is incorrectly named in my opinion.
> 
> I can split it to be more precise, so that we get separate tests for source and target. This will be obvious in the classes then. These three categories will be visible then:
> 
> * No messages
> * Target messages
> * Source messages

Took another look at the code. and I think I'll alter the tests still, but do it in another way entirely.
I'm going to introduce tests for the target/source messages, and remove the current messages out of the current tests, because that does not seem to be what the test is actually about, but a side-effect which is added to those tests.

filiphr:
Ideally we should not touch the existing tests and add new ones for this. Or update the ones that are really testing the error messages. The `ReferencedAccessibilityTest` doesn't sound like the right one.

The `DottedErrorMessageTest` might be more correct for this. Perhaps even the two `UnmappedSourceTest`

Zegveld:
> Ideally we should not touch the existing tests and add new ones for this. Or update the ones that are really testing the error messages. The `ReferencedAccessibilityTest` doesn't sound like the right one.
> 
> The `DottedErrorMessageTest` might be more correct for this. Perhaps even the two `UnmappedSourceTest`

Indeed the `DottedErrorMessageTest` does sound like the correct location. I didn't find that one earlier, because the text `Mapping from property` is not present in that test, instead it has it in a static variable `Mapping from "+ PROPERTY`.

I do feel like changing `ReferencedAccessibilityTest` so that the message is not part of it anymore is the right call, because it is using a side-effect to test behavior while that should be explicitly tested.
(I could do that in a separate PR though, easy enough to separate.)

filiphr:
>I do feel like changing `ReferencedAccessibilityTest` so that the message is not part of it anymore is the right call, because it is using a side-effect to test behavior while that should be explicitly tested.

I am not sure that I am following. What is wrong with that test? The test is properly showing that private methods are not going to be picked up by MapStruct and thus lead to the default warning.

Zegveld:
> > I do feel like changing `ReferencedAccessibilityTest` so that the message is not part of it anymore is the right call, because it is using a side-effect to test behavior while that should be explicitly tested.
> 
> I am not sure that I am following. What is wrong with that test? The test is properly showing that private methods are not going to be picked up by MapStruct and thus lead to the default warning.

default settings is in my opinion not something that the test should know of. If at some point in time the default setting changes, a lot of tests need to be fixed because they assumed that default settings would never change.

But it is something that can be discussed outside of this PR.

filiphr:
>default settings is in my opinion not something that the test should know of. If at some point in time the default setting changes, a lot of tests need to be fixed because they assumed that default settings would never change.

We can agree to disagree about this. In my opinion it is totally OK for a lot of tests to fail if we change a default, this also means that users will be affected by this and it is a good signal that something is not OK. 
In any case I do agree we can discuss this outside of this PR. Let me know when you are done with this so that I can give it another pass.

Zegveld:
I've accepted your suggestions for the production code, and rewritten the tests to be a part of `DottedErrorMessageTest`.
This did result in changing the target tests there, because otherwise they would be hard to distinguish from eachother.

Should be good for another pass.

filiphr (approved):

filiphr:
>This did result in changing the target tests there, because otherwise they 

Yea those are fine since those are around testing the messages we are adapting now. Will merge once CI is green

Issue #702, opened 2015-11-26
retain annotations at runtime


rmannibucau:
@Mappings (and @Mapping) don't have a runtime retention. This is fine for mapstruct but I wanted to introspect the annotations in my project to generate a custom adoc report from the mapping class.

I see 2 solutions:

- mapstruct adds a report option/solution
- the retention is updated to let users rely on this model for their own purposes

gunnarmorling:
Hi, couldn't you generate your report at build time, too, using an annotation processor? To me it seems like that'd be better suited for such once-off task (the mapper config cannot change at runtime anyways).

> mapstruct adds a report option/solution

What would such report entail?

gunnarmorling:
Btw. if you feel adventurous, you could try and plug into the MapStruct processing chain. This will allow you to access the mapper model which is the basis for the code generation, so you could examine this to extract whatever information you need.

To do so, you'd have to implement [ModelElementProcessor](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/processor/ModelElementProcessor.java), register it via a `META-INF/services/...` file and add it to the processor path. It's not a public SPI at this point and I haven't tried it myself with external modell element processors, but it may be interesting to try.

rmannibucau:
Report "constraint" is to not add another module and to not package the code. Thought to MEProc but didnt find a nice way to do it with mvn without multiple build phase but sounds like the best solution ATM.

remmeier:
+1 it would be great if there is a standarized way of having all the evaluated mappings between source and target (not the raw annotations) available at runtime if desired.

we plan to use mapstruct to map JPA entities to DTO. As a consequence, we would like to do filtering, paging and sorting on DTOs. For this purpose, the paths/naming in DTO "queries" need to be translated back into entity "queries".

remmeier:
I looked more closely at the processors. Would there be an interest for a PR that adds a  "MapperRuntimeModelRenderingProcessor" processor similar to the MapperRenderingProcessor that has its own set of templates to generate mapping informations (src <-> target properties) into the mapping classes? The processor could be enabled/triggered by an annotation like @RuntimeModel on the mapper class. Some further Model classes would be necessary at runtime to have an API to access the model. The thing could also be moved into a dedicated project.

gunnarmorling:
Hi, generating some sort of metamodel for inspecting mappings is a very
interesting idea.

I like the idea of making this some kind of side project, so this may be
[...]

PR #3530, opened 2024-02-16
Mapper, Mapping, Mappings - Change RetentionPolicy to RUNTIME

WeierE1:
Changing for Mapper, Mapping, Mappings the RetentionPolicy from CLASS to RUNTIME. This is to get at runtime the defintion of the mapping. Can be used for example to convert a sorting definition of a database object that is used in a query in the corresponding sorting defintion of the dto that can be used in the frontend.


filiphr:
Closing this for now since we have #702 where this has been discussed and I've also asked a question in https://github.com/mapstruct/mapstruct/discussions/3528#discussioncomment-8584384. We are not that keen in changing the retention policy to runtime.

I don't understand how changing this would help with something on the runtime.

PR #3542, opened 2024-03-01
#3524 provide tests with delomboked classes annotated with lomboks @SuperBuilder annotation

### LINKED ISSUE #3524: Improve tests to show that Lombok `@SuperBuilder` is supported
It is currently not clear if Lombok `@SuperBuilder` is supported by MapStruct (see https://github.com/mapstruct/mapstruct-idea/issues/159#issuecomment-1826883528). It may work by accident.

We're using it extensively in our company code. So it is critical for us to know if we should disable our builders or maybe even need to redesign our domain objects.

Therefore I want to provide a set of integration tests to show that it works.

If it has been proven, we can look at the following tickets:
* https://github.com/mapstruct/mapstruct-idea/issues/159#issue-1990285833
* respectively this PR: https://github.com/mapstruct/mapstruct-idea/pull/161

@filiphr Does that sound okay to you?

thunderhook:
Haven't found that many different scenarios. I hope it's enough to show that it works.

If you have any further wishes or suggestions, please let me know.

Closes #3524


filiphr (changes requested):
Thanks @thunderhook, this looks OK to me. I have only one small request. Let's remove Lombok from the package the javadoc. This is about a super builder, so let's only keep that in the naming. Let's also remove the things that are not needed for the tests like the `equals` and `hashCode` methods.

thunderhook:
@filiphr Thanks for the suggestions. I tried to keep it as "original" as possible, hence the equals/hashCode, but no problem. I will remove it.

The `package-info.java` was was an attempt to show that other (de)lombok(ed) tests can also be added there.

So just name the package `org.mapstruct.ap.test.superbuilder' and remove any trace of Lombok?

filiphr:
>So just name the package `org.mapstruct.ap.test.superbuilder' and remove any trace of Lombok?

Indeed, rename the package and remove the Lombok traces. Try to limit it to the minimum that we need to test super builders. Thanks

filiphr:
Thanks @thunderhook

Issue #1610, opened 2018-09-20
Allow conditions/filters mapping iterables and maps


chris922:
I think I already saw some questions to this topic here or on stackoverflow and in our projects I sometimes missed a feature like this as well:

Mapping iterables will right now always map all iterables from A to B. It is often helpful to remove some entries in the result B. Right now this can be achieved with e. g. an `@AfterMapping` that removes these entries or create an own mapping method.

Allowing the user specifying some conditions would be helpful to let mapstruct still do most of the magic.

Some examples:

```java
// E1
@IterableMapping(condition = "java( sourceValue != null && sourceValue.startsWith(\"1\") )")
List<Integer> mapIntegersFilteredBySourceValue(List<String> source);

// E2
@IterableMapping(condition = "java( targetValue != null && targetValue < 5 )")
List<Integer> mapIntegersFilteredByTargetValue(List<String> source);

// E3
@IterableMapping(conditionQualifiedByName = "filterString")
List<Integer> mapIntegersFilteredBySourceValueInExtraMethod(List<String> source);

@Named("filterString")
boolean filterString(String sourceValue) {
  return sourceValue != null && sourceValue.startsWith("1");
}

// E4
@IterableMapping(conditionQualifiedByName = "filterInteger")
List<Integer> mapIntegersFilteredByTargetValueInExtraMethod(List<String> source);

@Named("filterInteger")
boolean filterInteger(Integer targetValue) {
  return targetValue != null && targetValue < 5;
}
```

`conditionQualifiedBy` annotation should be supported as well. And same things should be available for `@MapMapping` - here it should be possible to filter by `key` or `value`.

Things that should be clarified: 
* When providing a condition as an inline expression, what variable names should be available?
* There should be a way to filter _before_ and _after_ mapping the value, how to determine at which point the user would like to have his condition? Rely on variable names? Dividing the attributes in something like `sourceCondition` and `targetCondition`? Some other idea?
* Name it `condition` (thus add entry if result is `true`) or `filter` (thus remove entry if result is `true`)?

WDYT? 

_(I wasn't able to find an issue with the same topic.. even if I think I read something like this somewhere...)_

lazydays79:
+1

sjaakd:
@chris922 .. I guess this is very similar as #487.. right? its on how you copy merge collections.

chris922:
Hmm, haven't yet thought about update methods. In my description I just talked about regular mapping methods.

[...]

PR #3543, opened 2024-03-03
Add support for conditions on source parameters + fix incorrect use of source parameter in presence check method

### LINKED ISSUE #2610: Support conditional mapping for source parameters

### LINKED ISSUE #3270: @Condition badly implemented
### Expected behavior

@Condition should be a wrapper of the existing generated code (generated without this annotation). Not an overwritter of existing generated code.

### Actual behavior

When using @Condition, the generated code is rewritten which is bad (in comparison when not present). It needs to wrap it, not rewrite it.

1) Create a classical mapper class : 

```
@Mapper(componentModel = "spring", uses = {TaxeMapper.class})
public interface FormuleMapper {

       Formule formuleEntityToDTO(FormuleOption formule);
}
```

2) Generated code : 

```
@Generated(
    date = "2023-05-10T15:52:58+0200",
    comments = "version: 1.5.5.Final, compiler: javac, environment: Java 18.0.2.1 (Oracle Corporation)"
)
@Component
public class FormuleMapperImpl implements FormuleMapper {
@Override
    public Formule formuleEntityToDTO(FormuleOption formule) {
        if ( formule == null ) {
            return null;
        }

        Formule formule1 = new Formule();

        formule1.setId( formule.getId() );
        formule1.setNom( formule.getNom() );
        formule1.setCode( formule.getCode() );
        formule1.setOrdre( formule.getOrdre() );
        formule1.setCommentaire( formule.getCommentaire() );
        formule1.setFormuleAssemblage( formule.getFormuleAssemblage() );
        formule1.setProduit( produitToIdNom( formule.getProduit() ) );
        formule1.setTaxes( ligneTaxeSetToIdNomRepartitionList( formule.getTaxes() ) );
        formule1.setBrouillon( formule.isBrouillon() );
        formule1.setType( getTypeFormule( formule.getType() ) );
        if ( formule.getMinimumPrimeTechniquePure() != null ) {
            formule1.setMinimumPrimeTechniquePure( formule.getMinimumPrimeTechniquePure().doubleValue() );
        }
        return formule1;
    }
}
```

Steps above are OK. Now let's introduce @Condition with generated code to compare : 

1) 

```
@Mapper(componentModel = "spring", uses = {TaxeMapper.class})
public interface FormuleMapper {
  
    Formule formuleEntityToDTO(FormuleOption formule);

@Condition
    default boolean targetFormuleOptionNotNull(fr.cfdp.services.produit.model.entity.produit.FormuleOption targetFormuleOption) {
        return targetFormuleOption != null ;
    }
}
```

2)

```
@Override
    public Formule formuleEntityToDTO(FormuleOption formule) {
        if ( formule == null ) {
            return null;
        }

        Formule formule1 = new Formule();

        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setId( formule.getId() );
        }
        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setNom( formule.getNom() );
        }
        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setCode( formule.getCode() );
        }
        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setOrdre( formule.getOrdre() );
        }
        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setCommentaire( formule.getCommentaire() );
        }
        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setFormuleAssemblage( formule.getFormuleAssemblage() );
        }
        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setProduit( produitToIdNom( formule.getProduit() ) );
        }
        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setTaxes( ligneTaxeSetToIdNomRepartitionList( formule.getTaxes() ) );
        }
        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setBrouillon( formule.isBrouillon() );
        }
        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setType( getTypeFormule( formule.getType() ) );
        }
        if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setMinimumPrimeTechniquePure( formule.getMinimumPrimeTechniquePure().doubleValue() );
        }
        return formule1;
    }
```

So can you explain this difference in generated code : 

```
if ( formule.getMinimumPrimeTechniquePure() != null ) {
            formule1.setMinimumPrimeTechniquePure( formule.getMinimumPrimeTechniquePure().doubleValue() );
        }
```

become : 

```
 if ( targetFormuleOptionNotNull( formule ) ) {
            formule1.setMinimumPrimeTechniquePure( formule.getMinimumPrimeTechniquePure().doubleValue() );
        }
```

Which ends with a NullPointerException in my unit tests.

Thanks for lookup source problem.

### MapStruct Version

1.5.5.Final

### LINKED ISSUE #3459: Method with `@Condition` acts strange when applying to source
### Expected behavior

when `@Condition` annotated method's parameter type is same to mapping source
```java
    DTO toDto(E entity);

    @Condition
    default boolean isEntityAvailable(E entity) {
        if (entity == null) {
            return false;
        }
        return Persistence.getPersistenceUtil().isLoaded(entity);
    }
```
hope to generate something like
```java
    @Override
    public SubCategoryDto toDto(SubCategory entity) {
        if ( entity == null ) {
            return null;
        }
        if ( !isEntityAvailable( entity ) ) {
            return null;
        }
        SubCategoryDto.SubCategoryDtoBuilder<?, ?> subCategoryDto = SubCategoryDto.builder();
        subCategoryDto.id( entity.getId() );
        subCategoryDto.name( entity.getName() );
        subCategoryDto.nick( entity.getNick() );
        return SubCategoryDto.build();
    }
```
or just ignore it
```java
    @Override
    public SubCategoryDto toDto(SubCategory entity) {
        if ( entity == null ) {
            return null;
        }
        SubCategoryDto.SubCategoryDtoBuilder<?, ?> subCategoryDto = SubCategoryDto.builder();
        subCategoryDto.id( entity.getId() );
        subCategoryDto.name( entity.getName() );
        subCategoryDto.nick( entity.getNick() );
        return SubCategoryDto.build();
    }
```

### Actual behavior

but now I got this
```java
    @Override
    public SubCategoryDto toDto(SubCategory entity) {
        if ( entity == null ) {
            return null;
        }
        SubCategoryDto.SubCategoryDtoBuilder<?, ?> subCategoryDto = SubCategoryDto.builder();

        if ( isEntityAvailable( entity ) ) {
            subCategoryDto.id( entity.getId() );
        }
        if ( isEntityAvailable( entity ) ) {
            subCategoryDto.name( entity.getName() );
        }
        if ( isEntityAvailable( entity ) ) {
            subCategoryDto.nick( entity.getNick() );
        }

        return SubCategoryDto.build();
    }
```

using JPA with Mapstruct
I have a base interface:
```java
public interface EntityMapper<DTO extends EntityDto, E extends BasicEntity> {
    E toEntity(DTO dto);

    DTO toDto(E entity);

    // try to avoid JPA lazy loading issue
    // and this one worked well
    @Condition
    default boolean isCollectionAvailable(Collection<E> collection) {
        if (collection == null) {
            return false;
        }
        return Persistence.getPersistenceUtil().isLoaded(collection);
    }

    // try to avoid JPA lazy loading issue
    // got problem with this
    @Condition
    default boolean isEntityAvailable(E entity) {
        if (entity == null) {
            return false;
        }
        return Persistence.getPersistenceUtil().isLoaded(entity);
    }

    @MapperConfig(mappingControl = DeepClone.class,
                  mappingInheritanceStrategy = MappingInheritanceStrategy.AUTO_INHERIT_ALL_FROM_CONFIG)
    interface Config {
    }
}
```
and entity mappers:
```java
@Mapper(config = EntityMapper.Config.class, uses = {SubCategoryMapper.class})
public interface ParentCategoryMapper extends EntityMapper<ParentCategoryDto, ParentCategory> {
}

@Mapper(config = EntityMapper.Config.class, uses = {ParentCategoryMapper.class})
public interface SubCategoryMapper extends EntityMapper<SubCategoryDto, SubCategory> {
}
```
and then I got this:
```java
    @Override
    public SubCategoryDto toDto(SubCategory entity) {
        if ( entity == null ) {
            return null;
        }
        SubCategoryDto.SubCategoryDtoBuilder<?, ?> subCategoryDto = SubCategoryDto.builder();
        // Calling isEntityAvailable() so much times does not make any sense.
        if ( isEntityAvailable( entity ) ) {
            subCategoryDto.id( entity.getId() );
        }
        if ( isEntityAvailable( entity ) ) {
            subCategoryDto.name( entity.getName() );
        }
        if ( isEntityAvailable( entity ) ) {
            subCategoryDto.nick( entity.getNick() );
        }
        if ( isEntityAvailable( entity ) ) {
            subCategoryDto.theOrder( entity.getTheOrder() );
        }
        if ( isEntityAvailable( entity ) ) {
            subCategoryDto.icon( entity.getIcon() );
        }
        if ( isEntityAvailable( entity ) ) {
            subCategoryDto.link( entity.getLink() );
        }
        // this is right
        if ( ParentCategoryMapper.isEntityAvailable( entity.getParentCategory() ) ) {
            subCategoryDto.parentCategory( ParentCategoryMapper.toDto( entity.getParentCategory() ) );
        }
        return SubCategoryDto.build();
    }
```
and what I want is:
```java
    @Override
    public SubCategoryDto toDto(SubCategory entity) {
        if ( entity == null ) {
            return null;
        }
        if ( !isEntityAvailable( entity ) ) {
            return null;
        }
        SubCategoryDto.SubCategoryDtoBuilder<?, ?> subCategoryDto = SubCategoryDto.builder();
        subCategoryDto.id( entity.getId() );
        subCategoryDto.name( entity.getName() );
        subCategoryDto.nick( entity.getNick() );
        subCategoryDto.theOrder( entity.getTheOrder() );
        subCategoryDto.icon( entity.getIcon() );
        subCategoryDto.link( entity.getLink() );
        if ( ParentCategoryMapper.isEntityAvailable( entity.getParentCategory() ) ) {
            subCategoryDto.parentCategory( ParentCategoryMapper.toDto( entity.getParentCategory() ) );
        }
        return SubCategoryDto.build();
    }
```

### MapStruct Version

MapStruct 1.6.0.Beta1

filiphr:
This commit does a backwards incompatible change in regard to how presence check methods are used when only the source parameter is needed for the presence check. e.g.

```java
@Condition
public static boolean shouldMap(Employee employee) {
    // ...
}
```

is now only applied to source parameters or a property of type `Employee`. The method is no longer applied to all the parameters of a mapping method with a source parameter `Employee`. If you still want to do that then the method should be adapted

```java
@Condition
public static boolean shouldMap(Employee employee, @TargetPropertyName String targetPropertyName) {
    // ...
}
```

Instead of the `@TargetPropertyName` it can be `@SourcePropertyName`.

Fixes #2610 
Fixes #3459
Fixes #3270

NB: Perhaps we need to add some more tests for #3459 and #3270, I wanted to first see if this idea is worth pursuing.
-----

An alternative that we could do about the use of the right parameter would be if we introduce something like `@SourceParameter` and / or `@TargetProperty`. This people can explicitly decide if they want to apply a conditional method only on a source parameter or only on a target property. The 2 new annotations might also be used in other places

### DIFF
--- a/core/src/main/java/org/mapstruct/Condition.java
+++ b/core/src/main/java/org/mapstruct/Condition.java
@@ -80,7 +80,7 @@ import java.lang.annotation.Target;
  * This annotation can also be used as a meta-annotation to define the condition strategy.
  *
  * @author Filip Hrisafov
- * @see SourceCondition
+ * @see SourceParameterCondition
  * @since 1.5
  */
 @Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
--- a/core/src/main/java/org/mapstruct/Condition.java
+++ b/core/src/main/java/org/mapstruct/Condition.java
@@ -39,7 +39,7 @@ import java.lang.annotation.Target;
  * <strong>Note:</strong> The usage of this annotation is <em>mandatory</em>
  * for a method to be considered as a presence check method.
  *
- * <pre><code>
+ * <pre><code class='java'>
  * public class PresenceCheckUtils {
  *
  *   &#64;Condition
@@ -57,7 +57,7 @@ import java.lang.annotation.Target;
  * <p>
  * The following implementation of {@code MovieMapper} will be generated:
  *
- * <pre><code>
+ * <pre><code class='java'>
  * public class MovieMapperImpl implements MovieMapper {
  *
  *     &#64;Override
--- a/core/src/main/java/org/mapstruct/Condition.java
+++ b/core/src/main/java/org/mapstruct/Condition.java
@@ -11,7 +11,8 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * This annotation marks a method as a <em>presence check method</em> to check for presence in beans.
+ * This annotation marks a method as a <em>presence check method</em> to check for presence in beans
+ * or it can be used to define additional check methods for something like source parameters.
  * <p>
  * By default, bean properties are checked against {@code null} or using a presence check method in the source bean.
  * If a presence check method is available then it will be used instead.
@@ -20,11 +21,19 @@ import java.lang.annotation.Target;
  * The following parameters are accepted for the presence check methods:
  *     <ul>
  *         <li>The parameter with the value of the source property.
- *         e.g. the value given by calling {@code getName()} for the name property of the source bean</li>
+ *         e.g. the value given by calling {@code getName()} for the name property of the source bean
+ *         - only possible when using the {@link ConditionStrategy#PROPERTIES}
+ *         </li>
  *         <li>The mapping source parameter</li>
  *         <li>{@code @}{@link Context} parameter</li>
- *         <li>{@code @}{@link TargetPropertyName} parameter</li>
- *         <li>{@code @}{@link SourcePropertyName} parameter</li>
+ *         <li>
+ *             {@code @}{@link TargetPropertyName} parameter -
+ *             only possible when using the {@link ConditionStrategy#PROPERTIES}
+ *         </li>
+ *         <li>
+ *             {@code @}{@link SourcePropertyName} parameter -
+ *             only possible when using the {@link ConditionStrategy#PROPERTIES}
+ *         </li>
  *     </ul>
  *
  * <strong>Note:</strong> The usage of this annotation is <em>mandatory</em>
@@ -45,7 +54,7 @@ import java.lang.annotation.Target;
  *     MovieDto map(Movie movie);
  * }
  * </code></pre>
- *
+ * <p>
  * The following implementation of {@code MovieMapper} will be generated:
  *
  * <pre><code>
@@ -67,13 +76,21 @@ import java.lang.annotation.Target;
  *     }
  * }
  * </code></pre>
+ * <p>
+ * This annotation can also be used as a meta-annotation to define the condition strategy.
  *
  * @author Filip Hrisafov
- * @since 1.5
  * @see SourceCondition
+ * @since 1.5
  */
-@Target({ ElementType.METHOD })
+@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
 @Retention(RetentionPolicy.CLASS)
 public @interface Condition {
+
+    /**
+     * @return the places where the condition should apply to
+     * @since 1.6
+     */
+    ConditionStrategy[] appliesTo() default ConditionStrategy.PROPERTIES;
 
 }
--- a/core/src/main/java/org/mapstruct/Condition.java
+++ b/core/src/main/java/org/mapstruct/Condition.java
@@ -11,9 +11,9 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * This annotation marks a method as a <em>presence check method</em> to check check for presence in beans.
+ * This annotation marks a method as a <em>presence check method</em> to check for presence in beans.
  * <p>
- * By default bean properties are checked against {@code null} or using a presence check method in the source bean.
+ * By default, bean properties are checked against {@code null} or using a presence check method in the source bean.
  * If a presence check method is available then it will be used instead.
  * <p>
  * Presence check methods have to return {@code boolean}.
@@ -48,8 +48,7 @@ import java.lang.annotation.Target;
  *
  * The following implementation of {@code MovieMapper} will be generated:
  *
- * <pre>
- * <code>
+ * <pre><code>
  * public class MovieMapperImpl implements MovieMapper {
  *
  *     &#64;Override
@@ -67,11 +66,11 @@ import java.lang.annotation.Target;
  *         return movieDto;
  *     }
  * }
- * </code>
- * </pre>
+ * </code></pre>
  *
  * @author Filip Hrisafov
  * @since 1.5
+ * @see SourceCondition
  */
 @Target({ ElementType.METHOD })
 @Retention(RetentionPolicy.CLASS)
--- a/core/src/main/java/org/mapstruct/SourceParameterCondition.java
+++ b/core/src/main/java/org/mapstruct/SourceParameterCondition.java
@@ -28,7 +28,7 @@ import java.lang.annotation.Target;
  * <pre><code>
  * public class PresenceCheckUtils {
  *
- *   &#64;SourceCondition
+ *   &#64;SourceParameterCondition
  *   public static boolean isDefined(Car car) {
  *      return car != null &#38;&#38; car.getId() != null;
  *   }
@@ -69,6 +69,6 @@ import java.lang.annotation.Target;
 @Target({ ElementType.METHOD })
 @Retention(RetentionPolicy.CLASS)
 @Condition(appliesTo = ConditionStrategy.SOURCE_PARAMETERS)
-public @interface SourceCondition {
+public @interface SourceParameterCondition {
 
 }
--- a/core/src/main/java/org/mapstruct/SourceParameterCondition.java
+++ b/core/src/main/java/org/mapstruct/SourceParameterCondition.java
@@ -25,7 +25,7 @@ import java.lang.annotation.Target;
  * <strong>Note:</strong> The usage of this annotation is <em>mandatory</em>
  * for a method to be considered as a source check method.
  *
- * <pre><code>
+ * <pre><code class='java'>
  * public class PresenceCheckUtils {
  *
  *   &#64;SourceParameterCondition
@@ -43,7 +43,7 @@ import java.lang.annotation.Target;
  *
  * The following implementation of {@code CarMapper} will be generated:
  *
- * <pre><code>
+ * <pre><code class='java'>
  * public class CarMapperImpl implements CarMapper {
  *
  *     &#64;Override
--- a/core/src/main/java/org/mapstruct/ConditionStrategy.java
+++ b/core/src/main/java/org/mapstruct/ConditionStrategy.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct;
+
+/**
+ * Strategy for defining what to what a condition (check) method is applied to
+ *
+ * @author Filip Hrisafov
+ * @since 1.6
+ */
+public enum ConditionStrategy {
+    /**
+     * The condition method should be applied whether a property should be mapped.
+     */
+    PROPERTIES,
+    /**
+     * The condition method should be applied to check if a source parameters should be mapped.
+     */
+    SOURCE_PARAMETERS,
+}
--- a/core/src/main/java/org/mapstruct/SourceCondition.java
+++ b/core/src/main/java/org/mapstruct/SourceCondition.java
@@ -68,6 +68,7 @@ import java.lang.annotation.Target;
  */
 @Target({ ElementType.METHOD })
 @Retention(RetentionPolicy.CLASS)
+@Condition(appliesTo = ConditionStrategy.SOURCE_PARAMETERS)
 public @interface SourceCondition {
 
 }
--- a/core/src/main/java/org/mapstruct/SourceCondition.java
+++ b/core/src/main/java/org/mapstruct/SourceCondition.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation marks a method as a <em>check method</em> to check if a source parameter needs to be mapped.
+ * <p>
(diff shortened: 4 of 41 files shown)


thunderhook:
Sorry, I haven't had the time and mental complexity to think this through.
This is tricky, thanks for your effort in trying to tackle this issues.

Users would now have to provide the `@SourcePropertyName` or `@TargetPropertyName` just to get the condition to work (maybe like it used to).
So if the property name is unused (which it most likely will be), the IDE and other static code analysis tools will warn the user about the unused parameter. For example, we are currently improving our code quality with SonarQube, and false positives like this are a nuisance.
It also feels like a workaround.

So I would therefore prefer the second approach with the specific annotation, as you said "something like `@SourceParameter` and / or `@TargetProperty`".

Whichever way we go, users will still wonder why this is needed (e.g. "why do I have to add something extra?"). So this should be well documented in the docs and release notes.
But I also think that this backwards incompatible change needs to be done, and that this is the right time to do it.

I just played with your branch and saw that a condition written in the old way is simply not used. Is it possible to throw a compile error (or a warning)? (You have better insight into the underlying code)

filiphr:
>Sorry, I haven't had the time and mental complexity to think this through.
This is tricky, thanks for your effort in trying to tackle this issues.

No rush, I've been working on this one for a while, it took me a bit of time to wrap my head properly around it as well. Thanks a lot for going through it.

>So I would therefore prefer the second approach with the specific annotation, as you said "something like `@SourceParameter` and / or `@TargetProperty`".

That was my gut feeling as well, or perhaps have a different annotation that would be applicable for source parameters condition check.

>I just played with your branch and saw that a condition written in the old way is simply not used.

I didn't like that at all as well. I'll look into an alternative for this

filiphr:
@thunderhook, I gave another go at this. Basically, there is a new `@SourceCondition` (open to better naming suggestions) that should be used for source parameter checks. Let me know what you think about this approach

thunderhook:
I really like this approach.

One thing I noticed while playing with the branch: Is there anything against a condition method being a `@SourceCondition` and a `@Condition` at the same time? I do not see why this should not be allowed. Maybe there are some scenarios where this is useful.

With the current implementation, the method won't be used if you add a `@Condition` there:
https://github.com/filiphr/mapstruct/blob/efc374163242119df2b27d83f412a0ce699b7bf9/processor/src/test/java/org/mapstruct/ap/test/conditional/basic/ConditionalMethodForSourceBeanMapper.java#L22

I haven't tried it explicitly, but I don't think it would be picked up as a presence check method either.

Otherwise, I really like it!

filiphr:
You are indeed right. It won't be caught due to the `MethodFamilySelector`

```
method.getMethod().isObjectFactory() == criteria.isObjectFactoryRequired()
                && method.getMethod().isLifecycleCallbackMethod() == criteria.isLifecycleCallbackRequired()
                && method.getMethod().isPresenceCheck() == criteria.isPresenceCheckRequired()
                && method.getMethod().isSourceParameterCheck() == criteria.isSourceParameterCheckRequired()
```

I'll make sure that it works in that case as well.

I should also add some error handlings:

* Using `@SourceCondition` with `@TargetPropertyName` or `@SourcePropertyName` should lead to a compile error


An alternative to the `@SourceCondition`  could be something like:

```java
public enum ConditionStrategy {
    PROPERTIES,
    SOURCE_PARAMETERS,
}
```

then the condition would be

```java
@Target({ ElementType.METHOD })
public @interface Condition {

    ConditionStrategy[] appliesTo() default ConditionStrategy.PROPERTIES;

}
```

it can then be used

* `@Condition(appliesTo = ConditionStrategy.SOURCE_PARAMETERS)` - when we want to apply a condition only on source parameters
* `@Condition(appliesTo = { ConditionStrategy.PROPERTIES, ConditionStrategy.SOURCE_PARAMETERS})` - when it applied both to properties and source parameters

I am not sure which approach I like more. What do you think @thunderhook

thunderhook:
I haven't fully thought this through, but I think the second approach might be more flexible when introducing conditions for other scenarios. Like collection items #1610 or `TARGET_PARAMETERS`. Don't know if there are other scenarios.

If there are other strategies added, then it would make sense to provide a `ConditionStrategy.ALL`.

Another idea might be to support both approaches. Where `@SourceCondition` is a meta-annotation like:

```java
@Condition( appliesTo = ConditionStrategy.SOURCE_PARAMETERS )
public @interface SourceCondition {

}
```

This would allow users to write their own Condition meta annotations, like:
```java
@Condition( appliesTo = { ConditionStrategy.SOURCE_PARAMETERS, ... } )
public @interface MyOwnEntityCondition {

}
```

However, this may cause problems if multiple annotations are set. Should it cause a compile error? If not, should they be combined? What has higher priority? 
```java
@SourceCondition
@Condition(appliesTo = { ConditionStrategy.PROPERTIES, ConditionStrategy.SOURCE_PARAMETERS})
Target map(Source source);
```

So maybe it's better to stick to one of these approaches rather than a combination of the two.

filiphr:
Really good points @thunderhook. It would indeed work nicely for #1610 as well. That would basically be an additional `ConditionStrategy#ITERABLE_ELEMENTS`

>Another idea might be to support both approaches. Where `@SourceCondition` is a meta-annotation like:

I liked this even more. We could indeed provide `@SourceParameterCondition` or `@SourceCondition` that would be meta annotated with `@Condition(appliesTo = ConditionStrategy.SOURCE_PARAMETERS)`. Users can define their own annotations if they want to as well.

>However, this may cause problems if multiple annotations are set. Should it cause a compile error? If not, should they be combined? What has higher priority?

I don't think that this would cause problems. We would basically just merge all of the applicable `appliesTo` into a single set.

Thanks a lot for the feedback @thunderhook, I'll try to rework this in that direction :).

filiphr:
@thunderhook I finally found the time to go over what we discussed last time. Can you please have a look at this and let me know what you think

core/src/main/java/org/mapstruct/SourceParameterCondition.java
```diff
@@ -0,0 +1,74 @@
...
+ * </code></pre>
+ *
+ * @author Filip Hrisafov
+ * @since 1.6
+ * @see Condition @Condition
+ */
+@Target({ ElementType.METHOD })
+@Retention(RetentionPolicy.CLASS)
+@Condition(appliesTo = ConditionStrategy.SOURCE_PARAMETERS)
+public @interface SourceParameterCondition {
```
thunderhook, line 72: Have only seen usages with erroneus usages in tests. Can we alos have a happy case test?

thunderhook (commented):
@filiphr I really like it. Just some small nitpicks.

The documentation can also be improved in another branch, if you want to merge it sooner than later.

core/src/main/java/org/mapstruct/SourceParameterCondition.java
```diff
@@ -0,0 +1,74 @@
...
+ * The following parameters are accepted for the presence check methods:
+ *     <ul>
+ *         <li>The mapping source parameter</li>
+ *         <li>{@code @}{@link Context} parameter</li>
+ *     </ul>
+ *
+ * <strong>Note:</strong> The usage of this annotation is <em>mandatory</em>
+ * for a method to be considered as a source check method.
+ *
+ * <pre><code>
```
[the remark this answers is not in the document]
filiphr: Good point, I'll change it. I've also created #3588 so that we do it everywhere consistently.

core/src/main/java/org/mapstruct/Condition.java
```diff
@@ -67,14 +75,22 @@
  *         return movieDto;
  *     }
  * }
- * </code>
- * </pre>
+ * </code></pre>
+ * <p>
+ * This annotation can also be used as a meta-annotation to define the condition strategy.
```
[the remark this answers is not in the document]
filiphr: Good point, I'll do it as part of this PR

core/src/main/java/org/mapstruct/SourceParameterCondition.java
filiphr: Good point. I'll add a test with this as well. And also with the multiple applies to.

filiphr: Just checked, we are actually using this in `ConditionalMethodForSourceBeanMapper` actually.

filiphr:
Thanks a lot for the review @thunderhook. All your comments make a lot of sense. I've added some docs, added a new test and also added the class to the `<code>` tag. Do you think that the doc is enough?

thunderhook (approved):
@filiphr Yes, I think this is enough. Thanks for your hard work on this. 👍 Merge at your own will!

filiphr:
Thanks for the review @thunderhook

Issue #3539, opened 2024-02-25
Support `BeanMapping#ignoredTargets` instead of repeating `@Mapping(target = "foo", ignore = true)`


filiphr:
### Discussed in https://github.com/mapstruct/mapstruct/discussions/3537

<div type='discussions-op-text'>

<sup>Originally posted by **efenderbosch-atg** February 22, 2024</sup>
Instead of a bunch of repeated annotations like this:
```java
    @Mapping(target = "foo", ignore = true)
    @Mapping(target = "bar", ignore = true)
    @Mapping(target = "baz", ignore = true)
    @Mapping(target = "qux", ignore = true)
```
Something similar to `@BeanMapping::ignoreUnmappedSourceProperties`?

```java
    @BeanMapping(ignoreTargets = ["foo", "bar", "baz", "qux"])
````
</div>

Issue #2599, opened 2021-09-30
qualifiedBy for a list of nested beans has no effect


Korashen:
- [X] Is this an issue (and hence not a question)? 

Using qualifiedBy with annotations for ambiguous mapping methods does not work for a list of nested beans.

The code snippets in this post are shortened. The full code with reproducable results can be found here: https://github.com/Korashen/MapStructIssue

An ID class with the ambiguous methods
```java
public class MyId{
    private int appiid;
    private int id;
    private int version;

    // Constructor, Getter and Setter

    // The ambiguous methods
    public String asFormatA()
    public String asFormatB()
}
```

A domain entity to use the ID class and nested bean `Relation`
```java
public class MyEntity {
    private MyId myId;
    private String randomField;
    private List<MyId> anIdList;
    private List<Relation> relationList;

    // Constructor, Getter and Setter
}
```
The nested bean
```java
public class Relation {
    private MyId from;
    private MyId to;
    private String relationType;

    // Constructor, Getter and Setter
}
```

The API model looks like the same, but instead of the `MyId` objects, it uses simple strings. Example of MyEntity as API.
```java
public class MyApiModel {
    private String id;
    private String randomField;
    private List<String> anIdList;
    private List<MyApiRelation> relationList; // Also uses strings instead of MyIds

    // Constructor, Getter and Setter
```

Now, to disambiguate the mapping of MyId to String, I use a simple annotation and qualifiedBy:
```java
@Qualifier
@Target(ElementType.METHOD)
@Retention(RetentionPolicy.CLASS)
@interface FormatB {
}
```

For the mapper, please note, the original logic of the `default`s is a bit more complex than this example, so expressions will not work in my case, while they surely will for this example
```java
@Mapper
public interface MyIdMapper {

    default String mapA(MyId id){
        return id.asFormatA();
    }

    @FormatB
    default String mapB(MyId id){
        return id.asFormatB();
    }
    // Other mappings
}
```

```java
@Mapper(uses = {MyIdMapper.class})
public interface MyEntityMapper {

    @Mapping(source = "myId", target = "id", qualifiedBy = {FormatB.class}) // 1
    @Mapping(target = "anIdList", qualifiedBy = {FormatB.class}) // 2
    @Mapping(target = "relationList.from", qualifiedBy = {FormatB.class}) // 3
    @Mapping(target = "relationList.to", qualifiedBy = {FormatB.class}) // 3
    MyApiModel map(MyEntity entity);

}
```
[...]

Issue #1958, opened 2019-10-22
Allow ignoring multiple target properties


varunvora:
It would be great if mapstruct provided a way of ignoring multiple target properties.

Instead of writing this
```
@Mapping(target = "property1", ignore = true)
@Mapping(target = "property2", ignore = true)
@Mapping(target = "property3", ignore = true)
```

I would prefer
```
@Mapping(target = {"property1", "property2", "property3"}, ignore = true)
```

filiphr:
@varunvora `@Mapping` is for a single mapping. 

If we do what you suggest then what would do:

```
@Mapping(target = {"property1", "property2", "property3"}, source = "sourceProperty")
```

Not sure if this is better or not from an API point of view.

xumk:
@filiphr may be add `@PropertyIgnore({"property1", "property2", "property3"})` or `@Ingnore({"property1", "property2", "property3"})`? if  this decision is suitable, I can try to do it

sam0410:
Hi, Any update on this?

filiphr:
@sam0410 if there is no comment it means that there isn't any update.

The idea from @xumk is not that bad. We could internally transform this into `@Mapping(target = "...", ignore = true)` and do our mapping. Are you interested in working on this @sam0410?

Currently you can always define your own custom `@Mapping` annotation that will ignore bunch of properties.
Apart from that you can also use `BeanMapping#ignoreByDefault` to ignore all properties and explicitly define the properties you have.

xumk:
@filiphr i can to try do it on next week

filiphr:
No problem. Whenever you have the time

dschulten:
I discovered another, related use case. JAXB Beans can be generated with builder methods like withFoo, withBar and Mapstruct warns that it cannot map them. So a variant of this feature request would be to allow ignoring all methods that start with a particular prefix. Maybe even on the Mappings level:
```
@Mappings(ignorePattern="with.+", value={
    @Mapping(target="fooDto", source="foo"),
    @Mapping(target="barDto", source="bar")
})
...
```

PR #3545, opened 2024-03-04
#3539 Support BeanMapping#ignoredTargets instead of repeating @Mapping(target = "foo", ignore = true)

chenzijia12300:
This commit attempts to implement [#2599](https://github.com/mapstruct/mapstruct/issues/3539), mapping `BeanMapping#ignoredTargets` to multiple `Mapping#ignore` annotations. 

If there are any necessary changes, I'll do my best to address them.

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -7,7 +7,6 @@ package org.mapstruct.ap.internal.model;
 
 import java.text.MessageFormat;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
@@ -284,20 +283,6 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             }
 
             initializeMappingReferencesIfNeeded( resultTypeToMap );
-
-            if ( beanMapping != null ) {
-                for (String ignoreUnmapped : beanMapping.getIgnoreUnmappedTargetProperties()) {
-                    MappingOptions forIgnore = MappingOptions.
-                            forBeanMappingIgnoreTargets( ignoreUnmapped, method.getExecutable() );
-
-                    mappingReferences.getMappingReferences().add( new MappingReference(
-                            forIgnore,
-                            new TargetReference(null, Arrays.asList( ignoreUnmapped.split(  "\\."  ) )),
-                            null
-
-                    ) );
-                }
-            }
 
             boolean shouldHandledDefinedMappings = shouldHandledDefinedMappings( resultTypeToMap );
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -7,6 +7,7 @@ package org.mapstruct.ap.internal.model;
 
 import java.text.MessageFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
@@ -283,6 +284,18 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             }
 
             initializeMappingReferencesIfNeeded( resultTypeToMap );
+
+            if ( beanMapping != null ) {
+                for (String ignoreUnmapped : beanMapping.getIgnoreUnmappedTargetProperties()) {
+                    MappingOptions forIgnore = MappingOptions.forIgnore( ignoreUnmapped );
+                    mappingReferences.getMappingReferences().add( new MappingReference(
+                            forIgnore,
+                            new TargetReference(null, Arrays.asList( ignoreUnmapped.split(  "\\."  ) )),
+                            null
+
+                    ) );
+                }
+            }
 
             boolean shouldHandledDefinedMappings = shouldHandledDefinedMappings( resultTypeToMap );
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -287,7 +287,9 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
             if ( beanMapping != null ) {
                 for (String ignoreUnmapped : beanMapping.getIgnoreUnmappedTargetProperties()) {
-                    MappingOptions forIgnore = MappingOptions.forIgnore( ignoreUnmapped );
+                    MappingOptions forIgnore = MappingOptions.
+                            forBeanMappingIgnoreTargets( ignoreUnmapped, method.getExecutable() );
+
                     mappingReferences.getMappingReferences().add( new MappingReference(
                             forIgnore,
                             new TargetReference(null, Arrays.asList( ignoreUnmapped.split(  "\\."  ) )),
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
@@ -35,6 +35,7 @@ public class BeanMappingOptions extends DelegatingOptions {
 
     private final SelectionParameters selectionParameters;
     private final List<String> ignoreUnmappedSourceProperties;
+    private final List<String> ignoreUnmappedTargetProperties;
     private final BeanMappingGem beanMapping;
 
     /**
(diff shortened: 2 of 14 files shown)


filiphr:
Thanks for your PR @chenzijia12300. 

Can we perhaps add a test case when an unknown property is used in `BeanMapping#ignoreTargets`? How would the compile error look like? Where would it be reported?

chenzijia12300:
Hi @filiphr, thanks for your review, I have added a test case

processor/src/test/java/org/mapstruct/ap/test/unmappedtarget/beanmapping/BeanMappingIgnoreTargetTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.mapstruct.ap.testutil.compilation.annotation.Diagnostic;
+import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutcome;
+
+@IssueKey("3539")
+public class BeanMappingIgnoreTargetTest {
+
+    @ProcessorTest
+    @WithClasses({Source.class, Target.class, BeanMappingIgnoreTargetMapper.class})
+    public void shouldIgnoreTargetProperty() {
+        BeanMappingIgnoreTargetMapper.INSTANCE.convert( new Source() );
```
zyberzebra, line 21: Shouldn't there be an assertion? :)
```java
    public void shouldIgnoreTargetProperty() {
        Source source = new Source();
        Source.NestedSource nested = new Source.NestedSource();
        source.setAge( 18 );
        source.setName( "John" );
        source.setNested( nested );
        nested.setScore( 50.0 );
        Target convert = BeanMappingIgnoreTargetMapper.INSTANCE.convert( source );
        assertThat(convert).hasNoNullFieldsOrPropertiesExcept( "address" );
        assertThat(convert.getNested()).hasNoNullFieldsOrPropertiesExcept( "flag" );
    }
``` 

SHouldn't we add a test like that? :) Or is it enough thats it does not generate a warning that we have unmapped targets

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -284,6 +285,20 @@ else if ( !method.isUpdateMethod() ) {
 
             initializeMappingReferencesIfNeeded( resultTypeToMap );
 
+            if ( beanMapping != null ) {
```
zyberzebra, line 288: Maybe a new method?

zyberzebra (commented):
Just checked it out and had a look :)

zyberzebra:
I guess we have to update the intelliJ plugin if this is implemented ;)
image

processor/src/test/java/org/mapstruct/ap/test/unmappedtarget/beanmapping/BeanMappingIgnoreTargetMapper.java
```diff
@@ -0,0 +1,20 @@
...
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface BeanMappingIgnoreTargetMapper {
+
+    BeanMappingIgnoreTargetMapper INSTANCE =
+            Mappers.getMapper( BeanMappingIgnoreTargetMapper.class );
+
+    @BeanMapping( ignoreTargets = {"nested.flag", "address"} )
+    Target convert( Source source );
```
zyberzebra, line 19:
```java
    @Mapping( target = "address", ignore = true)
    @BeanMapping( ignoreTargets = {"nested.flag", "address"} )
    Target convert( Source source );
``` 
Should this lead to warning? What happens if we do that.

processor/src/test/java/org/mapstruct/ap/test/unmappedtarget/beanmapping/BeanMappingIgnoreTargetMapper.java
```diff
@@ -0,0 +1,20 @@
...
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface BeanMappingIgnoreTargetMapper {
+
+    BeanMappingIgnoreTargetMapper INSTANCE =
+            Mappers.getMapper( BeanMappingIgnoreTargetMapper.class );
+
+    @BeanMapping( ignoreTargets = {"nested.flag", "address"} )
```
zyberzebra, line 18: Shouldn't this also lead to: Target property "address" must not be mapped more than once.
```java
    @BeanMapping( ignoreTargets = {"nested.flag", "address", "nested.flag"} )
``` 

```java
    @Mapping( target = "address", ignore = true)
    @Mapping( target = "address", ignore = true)
``` 
does

processor/src/test/java/org/mapstruct/ap/test/unmappedtarget/beanmapping/BeanMappingIgnoreTargetTest.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    @ProcessorTest
+    @WithClasses({Source.class, Target.class, ErroneousBeanMappingIgnoreTargetMapper.class})
+    @ExpectedCompilationOutcome(
+            value = CompilationResult.FAILED,
+            diagnostics = {
+                    @Diagnostic(type = ErroneousBeanMappingIgnoreTargetMapper.class,
+                            kind = javax.tools.Diagnostic.Kind.ERROR,
+                            line = 19,
```
filiphr, line 31: The reported line should be on the `@BeanMapping` not on the method.

processor/src/test/java/org/mapstruct/ap/test/unmappedtarget/beanmapping/BeanMappingIgnoreTargetTest.java
filiphr: I agree about this one. We should add an assertion as well.

filiphr (changes requested):
I think that the best solution for this would be to do it in the `MapperCreationProcessor` similar to how we are doing it for `@BeanMapping( ignoreByDefault = true )`. 

Additionally, as @zyberzebra mentioned in some other comments. We need to make sure that people do not use duplicates and that you can't define things in `@BeanMapping#ignoreTargets` and `@Mapping` in the same time.

chenzijia12300:
@zyberzebra  @filiphr  Thank you very much for your review. I will try to modify these problems.

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
```diff
@@ -355,6 +355,27 @@ public void applyIgnoreAll(SourceMethod method, TypeFactory typeFactory,
         }
     }
 
+    public void applyIgnoreTargets(SourceMethod method, FormattingMessager messager,
+                                    List<String> ignoreTargets ) {
+        for ( MappingOptions mapping : mappings ) {
+            if ( !mapping.isIgnored() && ignoreTargets.contains( mapping.getTargetName() ) ) {
```
filiphr, line 361: I think that we should always report an error if something is defined twice. And the error should be similar to what we have for `PROPERTYMAPPING_DUPLICATE_TARGETS`

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingMethodOptions.java
```diff
@@ -355,6 +355,27 @@ public void applyIgnoreAll(SourceMethod method, TypeFactory typeFactory,
...
+                        getBeanMapping().getMirror(),
+                        Message.BEANMAPPING_IGNORE_TARGETS_WITH_MAPPING_TARGET_ERROR,
+                        ignoreTargets,
+                        mapping.getTargetName()
+                );
+            }
+        }
+
+        for (String ignoreTarget : ignoreTargets) {
+            MappingOptions mapping =
```
filiphr, line 373: Here we should validate whether `ignoreTarget` is `.` and then provide an error similar to `BEANMAPPING_IGNORE_BY_DEFAULT_WITH_MAPPING_TARGET_THIS`

filiphr (changes requested):
Thanks for the changes @chenzijia12300. I've added some more comments that I saw in my second review.

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
```diff
@@ -35,6 +35,7 @@ public class BeanMappingOptions extends DelegatingOptions {
 
     private final SelectionParameters selectionParameters;
     private final List<String> ignoreUnmappedSourceProperties;
+    private final List<String> ignoreUnmappedTargetProperties;
```
ronodhirSoumik, line 38: Isnt it better to use Set instead of list for both `ignoreUnmappedSourceProperties` and `ignoreUnmappedTargetProperties`

filiphr:
Closing this since we did #1958 an now you can use `@Ignored` to ignore multiple target properties

Issue #2257, opened 2020-10-29
[feature suggestion] Support component injection in abstract base class


incaseoftrouble:
Another one :-)

Suppose I have a mapper like this:
```java
@Mapper // with proper configuration of componentModel, injectionStrategy = CONSTRUCTOR, etc.
public abstract class Mapper {
  @Autowired
  private SpringComponent component;

  // mapping methods
}
```
This works, however using the recommended constructor injection

```java
@Mapper // with proper configuration of componentModel etc.
public abstract class Mapper {
  private final SpringComponent component;

  @Autowired
  Mapper(SpringComponent component) {
    this.component = component;
  }

  // mapping methods
}
```
does not work, since the generated mapper always a default constructor to be present. I suggest that if a single `@Autowired` constructor (or equivalent for other component models) is found, it takes the arguments and adds it to the generated constructor (including `@Qualifier` annotations etc.)

This is sort of low priority because using field injection on the base mapper still works, but would be nice to have I think.

filiphr:
I think that we had this request when we were implementing this construction injection strategy. However, it is a bit tricky to get it right for all the different component models.

I think that the complexity of the implementation is not worth it.

And you can always use setter injection to make your mappers more easily testable.

e.g.

```
@Mapper // with proper configuration of componentModel etc.
public abstract class Mapper {
  private final SpringComponent component;
  
  @Autowire
  void setComponent(SpringComponent component) {
    this.component = component;
  }

  // mapping methods
}
```

If you are interested you can give it a go at doing a draft for this so we can see how it will look like.

By the way you don't need to add `@Autowired` on constructors anymore. Spring handles that when you have a single constructor.

incaseoftrouble:
> By the way you don't need to add @Autowired on constructors anymore. Spring handles that when you have a single constructor.

I was aware of that, just wanted to emphasize ;-)

> However, it is a bit tricky to get it right for all the different component models.

Mh I guessed so. It should be rather easy for spring, right? You could literally just copy the constructor signature and append whatever is needed.

Admittedly, I don't know much about code generation / introspection and other component models. You probably would want to support that for all component models, right?

filiphr:
Yes for Spring it would be rather easy as there is only a single constructor. However, `cdi` currently needs to have an empty constructor for `ApplicationScoped` beans. If we are doing something we should make it work for all.

[...]

PR #3546, opened 2024-03-04
#2257 Adding abstract class mappers with defined constructor

TheAndreyy:
For spring mappers you can inject components by constructor when it is defined, mapper implementation will call it's super constructor.

Adding calling super canonical costructor only for Spring mappers as decided in referenced issue by project maintainers

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/CanonicalConstructor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/CanonicalConstructor.java
@@ -1,0 +1,39 @@
+package org.mapstruct.ap.internal.model;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.mapstruct.ap.internal.model.common.ModelElement;
+import org.mapstruct.ap.internal.model.common.Parameter;
+import org.mapstruct.ap.internal.model.common.Type;
+
+public class CanonicalConstructor extends ModelElement implements Constructor {
+
+    private final String name;
+    private final List<Parameter> parameters;
+
+    public CanonicalConstructor(String name, List<Parameter> parameters) {
+        this.name = name;
+        this.parameters = parameters;
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    public List<Parameter> getParameters() {
+        return parameters;
+    }
+
+    @Override
+    public Set<Type> getImportTypes() {
+        Set<Type> types = new HashSet<>();
+        for ( Parameter parameter : parameters ) {
+            types.addAll( parameter.getType().getImportTypes() );
+        }
+        return types;
+    }
+
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/SpringComponentProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/SpringComponentProcessor.java
@@ -15,15 +15,21 @@ import java.util.Set;
 import org.mapstruct.ap.internal.gem.MappingConstantsGem;
 import org.mapstruct.ap.internal.model.Annotation;
 import org.mapstruct.ap.internal.model.Mapper;
+import org.mapstruct.ap.internal.model.CanonicalConstructor;
 import org.mapstruct.ap.internal.model.annotation.AnnotationElement;
 import org.mapstruct.ap.internal.model.annotation.AnnotationElement.AnnotationElementType;
+import org.mapstruct.ap.internal.model.common.Parameter;
 
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.Element;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Modifier;
 import javax.lang.model.element.PackageElement;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
 
 import static javax.lang.model.element.ElementKind.PACKAGE;
+import static javax.lang.model.util.ElementFilter.constructorsIn;
 
 /**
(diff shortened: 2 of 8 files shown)


filiphr:
Thanks for the PR @TheAndreyy. From what I could see this only works for spring. In order for us to be able to accept this we would need to have full support for the rest of the component models as well (for those where this is possible).

Is this something that you can work on to finish the PR?

TheAndreyy:
Hey @filiphr, thank for the review.

Yes, I'd be happy to implement this. Now it only works for Spring as you've designed it in this [comment](https://github.com/mapstruct/mapstruct/issues/2257#issuecomment-723260107).

As I'm monstly fimiliar with Spring, I have to ask you for more requirements regarding other component models constructors. For example if I add this for "default" ComponentModel, getting Mapper instance like this `Mappers.getMapper( MapperWithCanonicalConstructor.class )` won't work.

filiphr:
In that comment I was saying that it would be easy for Spring, but that we should make it work for all 😄.

>As I'm monstly fimiliar with Spring, I have to ask you for more requirements regarding other component models constructors.

For the non default ones, I would say similar to Spring.

For the default component model we could do something similar like we are doing for the fields. i.e. if it is a mapper use `Mappers.getMapper` when calling the support, otherwise instantiate the class.

TheAndreyy:
@filiphr Ok, I'll add it to other component models.

I'm not sure, I understand correctly your solution. You mean to create additional no-args-constructor that supplies super constructor with empty members, similar like in `NewDatatypeFactoryConstructorFragment.ftl`?
I see it something like this:
```java
public MapperWithCanonicalConstructor() {
    super( field1Type.newInstance(), field2Type.newInstance());
}
```

> when calling the support

What do you mean by that?

filiphr:
>I'm not sure, I understand correctly your solution. You mean to create additional no-args-constructor that supplies super constructor with empty members

Not sure if we are talking about the default component model now. If it is about the default I mean to generate something like:

```java
public MyCustomMapperImpl() {
    super( Mappers.getMapper(MyOtherMapper.class), new MyUtils() );
}
```

basically we need to do exactly what we are doing when `MyCustomMapper` has `@Mapper(uses = { MyOtherMapper.class, MyUtils.class })`

>What do you mean by that?

oops, I meant to say "When calling the supper", not idea why I said "support"

TheAndreyy:
@filiphr I've added changes as you requested. I've tried to satisfy requirements of a constructor for all component models, but not all have tests with constructor injection so I couldn't copy them ;) Waiting for your code suggestions!

TheAndreyy:
Hey @filiphr, hope you had good rest during Easter break. If you find some time for code review I'd be grateful :smile:

processor/src/test/java/org/mapstruct/ap/test/canonicalconstructor/defaultcomponentmodel/UserCanonicalConstructorMapper.java
```diff
@@ -0,0 +1,40 @@
...
+package org.mapstruct.ap.test.canonicalconstructor.defaultcomponentmodel;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.ap.test.canonicalconstructor.shared.AddressDto;
+import org.mapstruct.ap.test.canonicalconstructor.shared.AddressEntity;
+import org.mapstruct.ap.test.canonicalconstructor.shared.UserDto;
+import org.mapstruct.ap.test.canonicalconstructor.shared.UserEntity;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
```
zyberzebra, line 15: If you add additional Mappers via `@Mapper(uses=.....)` you get errors.
`ERROR UserCanonicalConstructorMapper.java:20 No implementation was created for UserCanonicalConstructorMapper due to having a problem in the erroneous element null.`

Could you try that?

TheAndreyy: Hey @zyberzebra, thanks for the review. Good catch, I'll look into it to fix this :sweat_smile:

filiphr:
>Hey @filiphr, hope you had good rest during Easter break. If you find some time for code review I'd be grateful 😄

Sorry @TheAndreyy, I've been a bit busy before and after the Easter break and didn't have much time to review this. I'll try to check it out and provide feedback to you asap

processor/src/test/java/org/mapstruct/ap/test/canonicalconstructor/defaultcomponentmodel/UserCanonicalConstructorMapper.java
zyberzebra: let me know if you need some help. I'm trying to fix it also but it gets kinda tricky here : `org.mapstruct.ap.internal.processor.AnnotationBasedComponentModelProcessor#process`

zyberzebra: i tried to implement something. It works but is kinda "dirty"
https://github.com/mapstruct/mapstruct/pull/3582

TheAndreyy:
Hi, long time no see ;) I've added support for `@Mapper(uses=.....)`.

Thanks for pointing this out @zyberzebra, I haven't used it before. In my opinion now it works more elegant - see test `UserCanonicalConstructorWithUsesOtherMapper`. Sorry I couldn't use your solution, a lot of other tests were failing and I wasn't able to fix it.

Waiting for review, I'm counting on you @filiphr that you find some time :)

TheAndreyy:
Hi @filiphr, I see I've missed to check code formatting, now pipeline should pass. Waiting for code review suggestions :smile:

PR #3560, opened 2024-03-22
add ksp via room compiler processing

### LINKED ISSUE #2522: Feature request: Support KSP
For Kotlin projects, Mapstruct currently makes use of KAPT. This is generally quite slow since it requires generating Java stubs before the annotation processing step.

KSP offers an alternative to this by making it easier for annotation processors like Mapstruct to more directly use Kotlin's compiler plugin infrastructure. Based on initial benchmarks, using KSP can speed up build times by as much as 2x.

https://github.com/google/ksp

StefanLobbenmeierObjego:
WIP - just to get a feel for how much effort this would be.

Aims to fix #2522

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -726,7 +726,6 @@
                                 <requireJavaVersion>
                                     <version>[${minimum.java.version},)</version>
                                 </requireJavaVersion>
-                                <DependencyConvergence />
                                 <requirePluginVersions />
                                 <requireMavenVersion>
                                     <version>[3.0.3,)</version>
--- a/processor/pom.xml
+++ b/processor/pom.xml
@@ -25,6 +25,7 @@
             Therefore we set this empty property here-->
         <jacocoArgLine />
         <minimum.java.version>11</minimum.java.version>
+        <kotlin.version>1.9.0</kotlin.version>
     </properties>
 
     <repositories>
@@ -50,6 +51,16 @@
             <groupId>androidx.room</groupId>
             <artifactId>room-compiler-processing</artifactId>
             <version>2.6.1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.auto</groupId>
+            <artifactId>auto-common</artifactId>
+            <version>0.11</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.devtools.ksp</groupId>
+            <artifactId>symbol-processing-api</artifactId>
+            <version>1.9.0-1.0.13</version>
         </dependency>
         <dependency>
             <groupId>org.mapstruct.tools.gem</groupId>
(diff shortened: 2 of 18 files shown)


processor/pom.xml
```diff
@@ -369,6 +395,68 @@
                     </parameter>
                 </configuration>
             </plugin>
+            <plugin>
```
StefanLobbenmeierObjego, line 398: not necessarily needed, just added automatically by intellij when I copied over kotlin example from room. But they are fairly small, so we can also keep them in java

parent/pom.xml
```diff
@@ -726,7 +726,6 @@
                                 <requireJavaVersion>
                                     <version>[${minimum.java.version},)</version>
                                 </requireJavaVersion>
-                                <DependencyConvergence />
```
StefanLobbenmeierObjego, line 729: room has some libraries with kotlin 1.8.0 and some with kotlin 1.9.0, causing conflicts. Maybe someone who knows maven better than me knows how to resolve that conflict, but just to keep going I turned it off so I can first focus on the rest

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
```diff
@@ -198,19 +201,20 @@ public SourceVersion getSupportedSourceVersion() {
         return SourceVersion.latestSupported();
     }
 
+    @NotNull
     @Override
-    public boolean process(final Set<? extends TypeElement> annotations, final RoundEnvironment roundEnvironment) {
+    public Set<XElement> process(@NotNull XProcessingEnv processingEnv, @NotNull Map<String, ? extends Set<? extends XElement>> elementsByAnnotation, boolean isLastRound) {
```
StefanLobbenmeierObjego, line 206: the processingEnv was a field in the superclass before, I was not sure if it stays the same object during the whole lifetime so I added it as a parameter in several locations, would have to check if field works as well

processor/src/main/java/org/mapstruct/ap/MappingProcessor.java
```diff
@@ -427,21 +425,22 @@ private <P, R> R process(ProcessorContext context, ModelElementProcessor<P, R> p
...
-
-                @Override
-                public TypeElement visitTypeAsClass(TypeElement e, Void p) {
-                    return e;
-                }
-
-            }, null
-        );
+    private XElement asTypeElement(XElement element) {
+        return element;
```
StefanLobbenmeierObjego, line 429: there does not seem to be a way to convert an XElement to an XElementType, but also did not see why this was needed

processor/src/main/java/org/mapstruct/ap/internal/processor/DefaultModelElementProcessorContext.java
```diff
@@ -68,7 +71,7 @@ public DefaultModelElementProcessorContext(ProcessingEnvironment processingEnvir
     }
 
     @Override
-    public Filer getFiler() {
+    public XFiler getFiler() {
```
StefanLobbenmeierObjego, line 74: the XFiler API is quite different to the Filer API, but it should be able to do the same. One notable difference is that it seems to require a source (guessing for performance reasons), which might require a lot of parameters to be added where the source is far from the Filer. It might also be possible to start off without a source at the start and add this performance improvement later

Issue #3133, opened 2023-01-04
Redundant null check in generated code when adding parameters


mjiderhamn:
### Expected behavior

Given `Foo1`/`Foo2` as simple beans
```
@Mapper
public interface FooMapper {
    void updateFoo(Foo1 input, @MappingTarget Foo2 toUpdate);
}
```

properly generates to 
```
    @Override
    public void updateFoo(Foo1 input, Foo2 toUpdate) {
        if ( input == null ) {
            return;
        }

        toUpdate.setBar( input.getBar() );
    }
```

The expectation is to be able to add additional parameters (to be used in expressions) without SpotBugs/FindBugs reporting bugs in the generated code.

### Actual behavior

```
@Mapper
public interface FooMapper {
    void updateFoo(Foo1 input, @MappingTarget Foo2 toUpdate, boolean baz);
}
```
generates
```
    @Override
    public void updateFoo(Foo1 input, Foo2 toUpdate, boolean baz) {
        if ( input == null ) {
            return;
        }

        if ( input != null ) {
            toUpdate.setBar( input.getBar() );
        }
    }
```
for which SpotBugs/FindBugs complains about Redundant null check `RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE`.

### Steps to reproduce the problem

See above

### MapStruct Version

1.5.3

filiphr:
This is a really specific edge case which only occurs when using update mappings with multiple source parameters when only one of the source parameters is non primitive.

I would mark this issue as up-for-grabs in case someone from the community would like to look into this.

kuza2010:
Same problem here:
```java
@Mapper(
        componentModel = "spring",
        collectionMappingStrategy = CollectionMappingStrategy.ADDER_PREFERRED,
        uses = { DocumentMapper.class }
)
public interface ChelovekMapper {
    @Mapping(source = "openRequest", target = "openList")
    Chelovek fromOpenRequest(OpenRequest openRequest);
}
```

generate the following code:
``` java
    @Override
    public Chelovek fromOpenRequest(OpenRequest openRequest) {
        if ( openRequest == null ) {
            return null;
        }

        Chelovek chelovek = new Chelovek();

        chelovek.setId( openRequest.geId() );
        if ( openRequest != null ) {
            chelovek.addDocumentList( DocumentMapper.fromOpenRequest( openRequest) );
        }

        // other mappings
        return chelovek;
    }
```

filiphr:
[...]

PR #3583, opened 2024-04-28
#3133 Redundant null check in generated code when adding parameters 

zyberzebra:
- verify redundant null checks in mappings are avoided
- targets mapper methods that involve update mappings and ensures that null checks are not applied redundantly if already covered


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
zyberzebra, line 1: I used the code formatter in IntelliJ: `etc/mapstruct.xml`

filiphr: When you are formatting it is best to only format the changed lines and also when committing it is best to only stage changes that are not formatting changes.

I'll clean up the not needed formatting changes now, that's not a problem

zyberzebra: Aaaah... makes sense! Thanks for clearing that up 😁

zyberzebra:
Somthing must be configured incorrectly in my IDE... I'm not sure why though

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -2096,7 +2096,16 @@ private boolean doesNotNeedPresenceCheckForSourceParameter(PropertyMapping mappi
...
+        return mapping.getAssignment().isSourceReferenceParameter()
+            || isCheckedByGuardClause( mapping );
+    }
+
+    private boolean isCheckedByGuardClause(PropertyMapping mapping) {
+        boolean isUpdateMethod = getReturnTypeToConstruct() == null && isMapNullToDefault();
+        return !isUpdateMethod
+            && presenceChecksByParameter.size() == 1
+            && presenceChecksByParameter.containsKey( mapping.getSourceBeanName() )
+            && mapping.getAssignment().getSourcePresenceCheckerReference() == null;
```
filiphr, line 2108: Why is this needed? All tests pass without this, did I remove some edge case in the existing test cases?

filiphr (commented):
Nice work @zyberzebra. I've done some small polishing (see my commits). Let me know if I've removed too much.

Can you please also rebase this on top of main?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
zyberzebra: Are you refering to only the last condition? 
`mapping.getAssignment().getSourcePresenceCheckerReference() == null`

filiphr: Yes indeed, only the last condition

zyberzebra: That is indeed a really good question.
I'm going to rebase and try a little if I can find out why I thought I need it.
If I can reproduce it, I'm going to remove the condition and update the PR.

filiphr: Please also check if the class and methods I removed from the tests are causing everything to be green

zyberzebra: I checked out the older revision and none of my Testcases covered that line. 
Sadly I can't tell you my initial thought when I wrote that line.

zyberzebra:
@filiphr Is just a rebase needed again?

filiphr:
Not sure what the problem is @zyberzebra, the logs from the builds are gone.

PR #3585, opened 2024-04-28
#3458 generic fix for method override comparison due to eclipse bug.

### LINKED ISSUE #3458: Default method implementation not taking into consideration
### Expected behavior

_No response_

### Actual behavior

_No response_

Architectural constraint:

```
public interface MapperOutput<DOMAIN, RESPONSE> {

	RESPONSE toResponse(DOMAIN domain);
	
}
```

sometimes I don't need to map DOMAIN to RESPONSE, I only have to return the domain. 

So I have provided a default implementation to return the domain, and this is working fine.

```
@Mapper(componentModel = MappingConstants.ComponentModel.SPRING, injectionStrategy = InjectionStrategy.CONSTRUCTOR)
public interface EmployeeDeleteManyMapper extends  MapperOutput<Void, Void> {
	
	@Override
	default Void toResponse(Void domain) {
		return domain;
	}
	
}
```

Since this is a common behavior, I have decided to introduce a new generic interface called IdentityMapperOutput that return the domain directly as a response

```
public interface IdentityMapperOutput<T> extends MapperOutput<T, T> {

	@Override
	default T toResponse(T domain) {
		return domain;
	}
	
}
```

but now, when I'm extending this interface with the related default method, It gives me an error:

"Void does not have an accessible constructor. Occured at 'RESPONSE toResponse(DOMAIN domain)' in 'MapperOutput'."

```
@Mapper(componentModel = MappingConstants.ComponentModel.SPRING, injectionStrategy = InjectionStrategy.CONSTRUCTOR)
public interface EmployeeDeleteManyMapper extends  IdentityMapperOutput<Void> {
	
}
```

It's not only about Void, it's in general, cause I have tried also with Long:

"Ambiguous constructors found for creating java.lang.Long: Long(long), Long(java.lang.String). Either declare parameterless constructor or annotate the default constructor with an annotation named @Default. Occured at 'RESPONSE toResponse(DOMAIN domain)' in 'MapperOutput'"

### MapStruct Version

1.5.5.Final

Zegveld:
instead of passing on the mapper type as parent, use the type of the class where the method is enclosed.

fixes #3458

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/util/AbstractElementUtilsDecorator.java
@@ -178,7 +178,7 @@ public abstract class AbstractElementUtilsDecorator implements ElementUtils {
             // This can happen when diamond inheritance is used with interfaces
             return;
         }
-        addMethodNotYetOverridden( alreadyAdded, methodsIn( element.getEnclosedElements() ), parentType );
+        addMethodNotYetOverridden( alreadyAdded, methodsIn( element.getEnclosedElements() ) );
 
         if ( hasNonObjectSuperclass( element ) ) {
             addEnclosedMethodsInHierarchy(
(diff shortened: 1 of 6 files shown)


Zegveld:
Closing this, as @filiphr showed in the issue that there is another situation in which this would result in a problem.

PR #3587, opened 2024-04-28
#3565 Presence check methods should not be considered as valid mapping candidates

### LINKED ISSUE #3565: @Conditional treated as ambiguous mapping for methods returning Boolean/boolean
### Expected behavior

If I have a method annotated with `@Condition` that serves as a presence check for Java Optionals, and I have a separate mapping method that unboxes an `Optional`, the Condition method should not be treated as a valid mapper method.

### Actual behavior

Conditional method, which needs to return boolean, clashes with mapping methods that return Boolean and whose input argument type is the same, e.g. 
```
@Condition
boolean conditionalMethod(TypeA typeAObject) { }
```
clashes with
```
@Mapper
interface MyMapper {
boolean mappingMethod(TypeA typeAObject);
}
```

Specifically, this is true when attempting to use a Conditional method to check the if the Boolean wrapped in an `Optional` `isPresent`, because it will return a boolean, while the mapping method that unwraps the Boolean also does this.

MapStruct 1.5.3, Spring Boot 3.1.6

```
import org.mapstruct.*;
import java.util.Optional;

@Mapper
interface MyMapper {
    default <T> T mapFromOptional(Optional<T> value) { return value.orElse((Object) null); }
    
    @Condition
    default boolean isOptionalPresent(Optional value) { return value.isPresent(); }

    MyTargetDTO map(MySourceDTO sourceDto);
}

class MySourceDTO {
    private Boolean isCondition;
    public Optional<Boolean> getIsCondition() { return Optional.ofNullable(this.isCondition); }
}

class MyTargetDTO {
    private String isCondition;
    public Optional<String> getIsCondition() { return Optional.ofNullable(this.isCondition); }
    public void setIsCondition(String isCondition) { this.isCondition = isCondition; }
}
```

Result:

> Ambiguous 2step methods found, mapping Optional sourceDto to String. Found conversionY( methodX ( parameter ) ): conversionY: Boolean -->String, method(s)X: T mapFromOptional(Optional<T> value); conversionY: boolean-->String, method(s)X: T mapFromOptional(Optional<T> value);

It seems to me like MapStruct is taking the `@Condition` method, and using it as a valid mapping method, instead of using it only for Conditional Mapping presence checking. 
If I understand correctly, the mapping steps would be:

1. Call the @Mapper's map, to convert the `MySourceDTO` to `MyTargetDTO`.
2. Since the getter returns an `Optional<Boolean>`, it needs to call the `@Mapper`'s `mapFromOptional`, for the `sourceDto.isCondition` property. 

This will resolve the mapping as `default Boolean mapFromOptional(Optional<Boolean> value)`, which has a similar signature as `default boolean isOptionalPresent(Optional value)`, so that must look ambiguous to MapStruct. However, the latter method is NOT a mapping method, merely a condition.

Methods annotated with `@Condition` should NOT be considered for mappings.

### MapStruct Version

MapStruct 1.5.3

filiphr:
Fixes #3565

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/creation/MappingResolverImpl.java
@@ -470,6 +470,9 @@ public class MappingResolverImpl implements MappingResolver {
         }
 
         private boolean isCandidateForMapping(Method methodCandidate) {
+            if ( methodCandidate.isPresenceCheck() ) {
+                return false;
+            }
             return isCreateMethodForMapping( methodCandidate ) || isUpdateMethodForMapping( methodCandidate );
         }
 
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_3565/Issue3565Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_3565/Issue3565Mapper.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._3565;
+
+import java.util.Optional;
+
+import org.mapstruct.Condition;
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+/**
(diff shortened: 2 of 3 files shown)


filiphr:
The changes in #3543 are going to cause this code to not compile. Therefore, once we merge this I'll need to update #3543 to have the changes from here

thunderhook (approved):

PR #3593, opened 2024-05-05
#3113 Add java LinkedHashSet, LinkedHashSet new factory methods for java >= 19

### LINKED ISSUE #3113: Use Java HashSet and HashMap new factory method with known capacity when on Java 19
### Use case

Instead of calculating the capacity in the generated code we should instead use

* [`HashSet.newHashSet(int)`](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/HashSet.html#newHashSet(int))
* [`HashMap.newHashMap(int)`](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/HashMap.html#newHashMap(int))

This should be done only when the Source level is >= 19.

### Generated Code

_No response_

### Possible workarounds

_No response_

### MapStruct Version

_No response_

Obolrom:
Fixes #3113 

- [x] Inject [VersionInformation](https://github.com/mapstruct/mapstruct/blob/main/processor/src/main/java/org/mapstruct/ap/internal/version/VersionInformation.java) to [IterableCreation](https://github.com/mapstruct/mapstruct/blob/main/processor/src/main/java/org/mapstruct/ap/internal/model/IterableCreation.java) class
- [x] Implement code generating using new factory methods for HashSet, HashMap, LinkedHashSet, LinkedHashSet
- [x] Separate tests for JDK 19+

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [17, 19, 21]
+        java: [17, 21]
     name: 'Linux JDK ${{ matrix.java }}'
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [17, 21]
+        java: [17, 19, 21]
     name: 'Linux JDK ${{ matrix.java }}'
     runs-on: ubuntu-latest
     steps:
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
@@ -96,6 +96,7 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
                 .forgedMethod( forgedMethod )
                 .returnTypeBuilder( builderType )
                 .mappingContext( ctx )
+                .versionInformation( versionInformation )
                 .build();
         }
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
@@ -16,7 +16,6 @@ import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.source.MappingMethodUtils;
 import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.util.Message;
-import org.mapstruct.ap.internal.version.VersionInformation;
 
 /**
  * @author Filip Hrisafov
@@ -26,7 +25,6 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
     protected B myself;
     protected MappingBuilderContext ctx;
     protected Method method;
-    protected VersionInformation versionInformation;
 
     AbstractBaseBuilder(Class<B> selfType) {
         myself = selfType.cast( this );
@@ -39,11 +37,6 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
 
     public B method(Method sourceMethod) {
         this.method = sourceMethod;
-        return myself;
-    }
-
-    public B versionInformation(VersionInformation versionInformation) {
-        this.versionInformation = versionInformation;
         return myself;
     }
 
@@ -96,7 +89,6 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
                 .forgedMethod( forgedMethod )
                 .returnTypeBuilder( builderType )
                 .mappingContext( ctx )
-                .versionInformation( versionInformation )
                 .build();
         }
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
@@ -16,6 +16,7 @@ import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.source.MappingMethodUtils;
 import org.mapstruct.ap.internal.model.source.Method;
 import org.mapstruct.ap.internal.util.Message;
+import org.mapstruct.ap.internal.version.VersionInformation;
 
 /**
  * @author Filip Hrisafov
@@ -25,6 +26,7 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
     protected B myself;
     protected MappingBuilderContext ctx;
     protected Method method;
+    protected VersionInformation versionInformation;
 
     AbstractBaseBuilder(Class<B> selfType) {
         myself = selfType.cast( this );
@@ -37,6 +39,11 @@ class AbstractBaseBuilder<B extends AbstractBaseBuilder<B>> {
 
     public B method(Method sourceMethod) {
         this.method = sourceMethod;
+        return myself;
+    }
+
+    public B versionInformation(VersionInformation versionInformation) {
+        this.versionInformation = versionInformation;
         return myself;
     }
 
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/ImplementationType.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/ImplementationType.java
@@ -22,7 +22,8 @@ public class ImplementationType {
         Type type,
         boolean initialCapacityConstructor,
         boolean loadFactorAdjustment,
-        String factoryMethodName) {
+        String factoryMethodName
+    ) {
         this.type = type;
         this.initialCapacityConstructor = initialCapacityConstructor;
         this.loadFactorAdjustment = loadFactorAdjustment;
@@ -37,8 +38,12 @@ public class ImplementationType {
         return new ImplementationType( type, true, false, null );
     }
 
-    public static ImplementationType withLoadFactorAdjustment(Type type, String factoryMethodName) {
-        return new ImplementationType( type, true, true, factoryMethodName );
+    public static ImplementationType withLoadFactorAdjustment(Type type) {
+        return new ImplementationType( type, true, true, null );
+    }
+
+    public static ImplementationType withFactoryMethod(Type type, String factoryMethodName) {
+        return new ImplementationType( type, true, false, factoryMethodName );
     }
 
     /**
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/ImplementationType.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/ImplementationType.java
@@ -16,23 +16,29 @@ public class ImplementationType {
     private final Type type;
     private final boolean initialCapacityConstructor;
     private final boolean loadFactorAdjustment;
+    private final String factoryMethodName;
 
-    private ImplementationType(Type type, boolean initialCapacityConstructor, boolean loadFactorAdjustment) {
+    private ImplementationType(
+        Type type,
+        boolean initialCapacityConstructor,
+        boolean loadFactorAdjustment,
+        String factoryMethodName) {
         this.type = type;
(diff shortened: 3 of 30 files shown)


Obolrom:
Basic implementation of the feature has been done. However, now 30 tests fail with errors similar to
Screenshot from 2024-05-07 22-30-49

@filiphr Could you please conduct a review and help me to understand how to fix those tests?

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
```diff
@@ -39,6 +41,11 @@ public B method(Method sourceMethod) {
         return myself;
     }
 
+    public B versionInformation(VersionInformation versionInformation) {
```
filiphr, line 44: I would actually add the `VersionInformation` to the `MappingBuilderContext`, no need to pass it along the builder.

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableCreation.ftl
```diff
@@ -11,6 +11,18 @@
...
+            LinkedHashSet.newLinkedHashSet( <@iterableSize /> )
+        <#elseif linkedHashMap>
+            LinkedHashMap.newLinkedHashMap( <@iterableSize /> )
+        <#elseif hashSet>
+            HashSet.newHashSet( <@iterableSize /> )
+        <#elseif hashMap>
+            HashMap.newHashMap( <@iterableSize /> )
+        <#else>
+            new <@includeModel object=resultType.implementationType/><#if ext.useSizeIfPossible?? && ext.useSizeIfPossible && canUseSize>( <@sizeForCreation /> )<#else>()</#if>
+        </#if>
```
filiphr, line 25: I would prefer if we can simplify this. I don't like that we have 4 if checks for the different implementations. 

I think that on the `ImplementationType` we can add additional information such as the `factoryMethodName` and that is defined then we can just do `<@includeModel object=resultType.implementationType raw=true/>.${resultType.implementationType.factoryMethodName}( <@sizeForCreation /> )`

Doing it like this only the `TypeFactory` needs to be adjusted with the `ImplementationType` based on the source version. The `IterableCreation.ftl` does not need to be aware about the source version.

filiphr (changes requested):
I've added some comments. We need a bit of cleanup before we get this on main.

Also for the failing tests, seems like we'll need to adjust our testing fixtures to be able to pick up different fails based on the java version we are testing with.

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableCreation.ftl
Obolrom: @filiphr, I have refactored that code. Please, check my update.

.github/workflows/main.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [17, 21]
+        java: [17, 19, 21]
```
filiphr, line 15: Why are we adding Java 19?

filiphr:
Thanks a lot @Obolrom. Looks way way cleaner now. I'll have a look for the fixtures to work on multiple java version

.github/workflows/main.yml
Obolrom: Don't we also need to run tests on version 19 to ensure the code is generated correctly?

filiphr: We are running the tests on Java 21, so that should cover this. Java 19 is no longer supported and we try to only run tests on LTS versions and the next upcoming version

jccampanero:
@filiphr @Obolrom Sorry for the late reply.

I have been working on the review too.

My main concern about the provided code is that, in my opinion, it actually doesn't deal with the actual problem. Please, let me trying explaining myself.

The problem the new factory methods trying addressing is the following.

Let's say you need to allocate `N` elements.

When you create an `ArrayList`, for instance, you do something like the following:

```
List<String> list = new ArrayList<>(N);
```

The JVM will ensure that enough room for the `N` elements will be allocated.

In contrast, if you do the same with a `HashMap`, for example:

```
Map<String,Object> map = new HashMap<>(N);
```

and you actual need `N` elements to be allocated, you are in trouble, because in fact the JVM will allocate N*<load factor>, by default, 0.75, elements instead: i.e., if you push `N` elements actually to that `HashMap` it will be internally rehashed.

Addressing this issue is the reason behind the new factory methods. Please, see the [actual code implementation](https://github.com/openjdk/jdk/blob/b68609f80d629846870c45d34046d8c47941003b/src/java.base/share/classes/java/util/HashMap.java#L2557-L2588).

In other words, I think the important point here is the actual initial capacity and not the load factor, as the implementation proposed suggests, as the implementation is focused on.

What do you think?

filiphr:
@jccampanero, sorry but I don't think that I understand your comment. The fact that `Map<String,Object> map = new HashMap<>(N);` does not load the right amount of data is something that we already know and that we handle in MapStruct.

The current implementation looks like:

```java
Map<String,Object> map = new HashMap<>( Math.max( (int) ( N / .75f ) + 1, 16 );
```

The goal of this PR is to generate code (on Java 19 or later) that looks like:

```java
Map<String,Object> map = HashMap.newHashMap( N );
```

With this PR, based on the java version of the code we would either use the load factor or the new factory method.

processor/src/main/java/org/mapstruct/ap/internal/model/common/ImplementationType.java
```diff
@@ -16,23 +16,29 @@ public class ImplementationType {
...
     }
 
     public static ImplementationType withInitialCapacity(Type type) {
-        return new ImplementationType( type, true, false );
+        return new ImplementationType( type, true, false, null );
     }
 
-    public static ImplementationType withLoadFactorAdjustment(Type type) {
-        return new ImplementationType( type, true, true );
+    public static ImplementationType withLoadFactorAdjustment(Type type, String factoryMethodName) {
```
jccampanero, line 40: Instead of using `withLoadFactorAdjustment`, I think we could use an explicit method to deal with factory methods: 

```java
    public static ImplementationType withFactoryMethod(Type type, String factoryMethodName) {
        return new ImplementationType( type, true, true, factoryMethodName );
    }
```

In the end, one concept may or not be related with the other.

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -118,16 +124,28 @@ public TypeFactory(ElementUtils elementUtils, TypeUtils typeUtils, FormattingMes
...
         implementationTypes.put( List.class.getName(), withInitialCapacity( getType( ArrayList.class ) ) );
 
-        implementationTypes.put( Set.class.getName(), withLoadFactorAdjustment( getType( LinkedHashSet.class ) ) );
+        implementationTypes.put(
+            Set.class.getName(),
+            withLoadFactorAdjustment(
+                getType( LinkedHashSet.class ),
+                isSourceVersionAtLeast19() ? LINKED_HASH_SET_FACTORY_METHOD_NAME : null
+            )
+        );
```
jccampanero, line 133: In case of using the proposed new method `withFactoryMethod`, this could be changed to:

```java
        Type linkedHashSetType = getType( LinkedHashSet.class );
        implementationTypes.put(
            Set.class.getName(),
            isSourceVersionAtLeast19() ?
                withLoadFactorAdjustment(linkedHashSetType) :
                withFactoryMethod(
                    linkedHashSetType, LINKED_HASH_SET_FACTORY_METHOD_NAME
                )
        );
```

In addition, it allows us to get rid of the last `null` argument.

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -118,16 +124,28 @@ public TypeFactory(ElementUtils elementUtils, TypeUtils typeUtils, FormattingMes
...
         implementationTypes.put( SortedSet.class.getName(), withDefaultConstructor( getType( TreeSet.class ) ) );
         implementationTypes.put( NavigableSet.class.getName(), withDefaultConstructor( getType( TreeSet.class ) ) );
 
-        implementationTypes.put( Map.class.getName(), withLoadFactorAdjustment( getType( LinkedHashMap.class ) ) );
+        implementationTypes.put(
```
jccampanero, line 137: Applies the same as before.

processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java
```diff
@@ -118,16 +124,28 @@ public TypeFactory(ElementUtils elementUtils, TypeUtils typeUtils, FormattingMes
...
+                getType( LinkedHashMap.class ),
+                isSourceVersionAtLeast19() ? LINKED_HASH_MAP_FACTORY_METHOD_NAME : null
+            )
+        );
         implementationTypes.put( SortedMap.class.getName(), withDefaultConstructor( getType( TreeMap.class ) ) );
         implementationTypes.put( NavigableMap.class.getName(), withDefaultConstructor( getType( TreeMap.class ) ) );
         implementationTypes.put(
             ConcurrentMap.class.getName(),
-            withLoadFactorAdjustment( getType( ConcurrentHashMap.class ) )
+            withLoadFactorAdjustment( getType( ConcurrentHashMap.class ), null )
```
jccampanero, line 148: Get rid of the last `null` argument.

jccampanero:
Hi @filiphr, thank you very much for your feedback.

I absolutely agree with you. I was merely trying to provide some context, but I understand the purpose of the PR. I apologize if I didn't explain myself clearly earlier.

First, I want to say that the provided implementation is quite good. Thank you, @Obolrom.

However, I believe it is very focused on the load factor when we should be concentrating more on the factory method itself. I have performed a code review which I hope clarifies my point.

Thank you very much for your patience, @filiphr, and great job, @Obolrom.

jccampanero (commented):
Great job @Obolrom.

Obolrom:
> I'll have a look for the fixtures to work on multiple java version
@filiphr any updates about supporting fixtures for multiple java versions?

filiphr:
@jccampanero I understood your comments now. @Obolrom I did some polishing from the feedback provided by @jccampanero. Let me know if you are OK with this.

I also gave it a go to have java 21 based fixtures. I'm not super happy with the solution and it doesn't work well with the current version of the eclipse plugin for some reason.

Obolrom:
@filiphr code looks a little bit cleaner now

Obolrom:
@filiphr, I did the fix for the Eclipse plugin.
@jccampanero, what do you think?

jccampanero (commented):
@Obolrom, I think it looks great now.

Regarding the tests, I would prefer if @Filiphr provides his opinion.

Great job, @Obolrom!

Obolrom:
@filiphr could you please conduct a code review? It seems we are close to merge this PR 😊

processor/src/main/resources/org/mapstruct/ap/internal/model/IterableCreation.ftl
Obolrom: It seems this requested changes have been resolved

Obolrom:
@filiphr, I have rebased the branch and fixed a test

thunderhook:
>  @thunderhook, could you please take a look at the PR?

I'm sorry, I'm on holiday at the moment. Maybe I can make it in about 2 weeks if Filip hasn't had a look yet.

Obolrom:
@filiphr , it seems the PR in status 'changes requested'. However, those changes are outdated

Obolrom (commented):
reviewed

filiphr:
>@filiphr , it seems the PR in status 'changes requested'. However, those changes are outdated.
Or have I missed something?

It is like that, because it still looks at my old review. I'll try to find some time to review this. Sorry that it is taking longer @Obolrom

Obolrom:
@filiphr @thunderhook 
I have updated the branch, it is up to date now.

filiphr:
Thanks @Obolrom. I'll try to review it it this week during my holidays.

filiphr:
I finally got the chance to merge this. Thanks @Obolrom. I did some final polishing in https://github.com/mapstruct/mapstruct/pull/3593/commits/e34abec139ff45c53996913d491bc01ee1668404

Obolrom:
@filiphr thank you a lot

PR #3595, opened 2024-05-06
[#1140] adds warning when the target has no target properties

zyberzebra:
Fixes: #1140 
TODO:

- [x] should no accessible properties also lead to the warning? 
   - [x] I guess so
- [x] remove reportErrorForUnmappedTargetPropertiesIfRequired, I guess?
   - realized unmappedTarget is a different case.
- [x] Any other possibilities or edge cases to test?


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -232,6 +232,14 @@ else if ( !method.isUpdateMethod() ) {
             // determine accessors
             Map<String, Accessor> accessors = resultTypeToMap.getPropertyWriteAccessors( cms );
             this.targetProperties = new LinkedHashSet<>( accessors.keySet() );
+            if ( targetProperties.isEmpty() ) {
```
zyberzebra, line 235: - [ ] Do we have a Validation Framework for stuff like that?

zyberzebra: I guess the validation framework is all those ifs in the build method haha

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -225,8 +229,10 @@ public enum Message {
     ANNOTATE_WITH_ENUM_CLASS_NOT_DEFINED( "enumClass needs to be defined when using enums." ),
     ANNOTATE_WITH_ENUMS_NOT_DEFINED( "enums needs to be defined when using enumClass." ),
     ANNOTATE_WITH_ANNOTATION_IS_NOT_REPEATABLE( "Annotation \"%s\" is not repeatable." ),
-    ANNOTATE_WITH_DUPLICATE( "Annotation \"%s\" is already present with the same elements configuration.", Diagnostic.Kind.WARNING ),
-    ;
+    ANNOTATE_WITH_DUPLICATE(
+        "Annotation \"%s\" is already present with the same elements configuration.",
+        Diagnostic.Kind.WARNING
+    );
```
filiphr, line 235: Please do not commit unrelated style changes. Can you please remove this change?

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -92,6 +92,10 @@ public enum Message {
     PROPERTYMAPPING_CANNOT_DETERMINE_SOURCE_PARAMETER_FROM_TARGET("No property named \"%s\" exists in source parameter(s). Please define the source explicitly."),
     PROPERTYMAPPING_NO_SUITABLE_COLLECTION_OR_MAP_CONSTRUCTOR( "%s does not have an accessible copy or no-args constructor." ),
     PROPERTYMAPPING_EXPRESSION_AND_CONDITION_QUALIFIED_BY_NAME_BOTH_DEFINED( "Expression and condition qualified by name are both defined in @Mapping, either define an expression or a condition qualified by name." ),
+    PROPERTYMAPPING_TARGET_HAS_NO_TARGET_PROPERTIES(
+        "Target \"%s\" has no target properties, targeted by this mapping method: \"%s\"",
+        Diagnostic.Kind.WARNING
+    ),
```
filiphr, line 98: Please put this on one line in order to be consistent with how it is for the other messages.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1140/AirplaneWithNoAccessors.java
```diff
@@ -0,0 +1,6 @@
+package org.mapstruct.ap.test.bugs._1140;
```
filiphr, line 1: I wouldn't call this a bug. I think some other test package would be a better fit.

processor/src/test/java/org/mapstruct/ap/test/bugs/_1140/JarMapperTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.mapstruct.ap.testutil.ProcessorTest;
+import org.mapstruct.ap.testutil.WithClasses;
+import org.mapstruct.ap.testutil.compilation.annotation.CompilationResult;
+import org.mapstruct.ap.testutil.compilation.annotation.Diagnostic;
+import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutcome;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+class JarMapperTest {
```
filiphr, line 12: The assertions in the test are irrelevant. What we want to test here is only the messages we have during compilation. So only `@ExpectedCompilationOutcome` for this

filiphr (commented):
Thanks for your work on this @zyberzebra. I've left some comments.

zyberzebra:
@filiphr I'm not 100% what the correct workflow is. Should I request reviews when I'm ready or not?

filiphr:
The workflow is me trying to get time and going through all the emails and updated PRs 😀. I usually go through PRs that have been updated as well

zyberzebra:
> The workflow is me trying to get time and going through all the emails and updated PRs 😀. I usually go through PRs that have been updated as well 

Thanks for the clarification ! I just scrolled through some other pull request here and saw that some request reviews and because it's my first time working open source I was unsure. ;)
At work I just ping my mates 😁

zyberzebra:
@filiphr Sorry for the ping. Just thought I give you a ping, because my PRs are probably way down on the pile now. 

With more time passing, it gets harder for me to remember what we fixed. Which is not a problem, put it might take longer to for me to respond if you come around for a review. :)

processor/src/main/java/org/mapstruct/ap/internal/util/Message.java
```diff
@@ -92,6 +92,7 @@ public enum Message {
     PROPERTYMAPPING_CANNOT_DETERMINE_SOURCE_PARAMETER_FROM_TARGET("No property named \"%s\" exists in source parameter(s). Please define the source explicitly."),
     PROPERTYMAPPING_NO_SUITABLE_COLLECTION_OR_MAP_CONSTRUCTOR( "%s does not have an accessible copy or no-args constructor." ),
     PROPERTYMAPPING_EXPRESSION_AND_CONDITION_QUALIFIED_BY_NAME_BOTH_DEFINED( "Expression and condition qualified by name are both defined in @Mapping, either define an expression or a condition qualified by name." ),
+    PROPERTYMAPPING_TARGET_HAS_NO_TARGET_PROPERTIES( "Target \"%s\" has no target properties, targeted by this mapping method: \"%s\"", Diagnostic.Kind.WARNING ),
```
filiphr, line 95: Do we really need the mapping method here? The line on which we report the warning is the method itself so it should not be needed to also put the method here.

processor/src/test/java/org/mapstruct/ap/test/emptytarget/JarMapperTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.mapstruct.ap.testutil.compilation.annotation.Diagnostic;
+import org.mapstruct.ap.testutil.compilation.annotation.ExpectedCompilationOutcome;
+
+
+class JarMapperTest {
+
+    @ProcessorTest
+    @ExpectedCompilationOutcome(value = CompilationResult.SUCCEEDED,
+        diagnostics = {
+            @Diagnostic(type = JarToJarMapper.class,
```
filiphr, line 20: We should also add the `line = 16` on the diagnostic as well

processor/src/test/java/org/mapstruct/ap/test/emptytarget/JarMapperTest.java
```diff
@@ -0,0 +1,39 @@
...
+        })
+    @WithClasses({ FilledJar.class, EmptyJar.class, JarToJarMapper.class })
+    void targetHasNoProperties() {
+    }
+
+
+    @ProcessorTest
+    @ExpectedCompilationOutcome(value = CompilationResult.SUCCEEDED,
+        diagnostics = {
+            @Diagnostic(type = JarToAirplaneMapper.class,
```
filiphr, line 32: We should also add the `line = 16` on the diagnostic as well

filiphr (changes requested):
Sorry for the waiting @zyberzebra. I've done another pass here. Have a look at my comments

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -232,6 +232,9 @@ else if ( !method.isUpdateMethod() ) {
...
```
filiphr, line 235: I missed this in my previous review. Looking at this if the target does not have setters but has constructor arguments then reporting it here before the constructor has been processed is incorrect. Shouldn't we move this after `this.targetProperties` is no longer updated (e.g. on line 257 it is populated)? See the adjusted tests.

filiphr (changes requested):
Good work @zyberzebra. I've done some small polishing and added a test case for the comment I left.

Have a look at it and let me know what you think

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
zyberzebra: You are right! 
I think we also have to come up with a solution for the newly added warning, which leads to a lot of existing Tests failing.

filiphr: Let me see what kind of tests are failing. Mapping into an empty object is anyways a bit strange. Perhaps we need to adjust those test cases.

PR #3600, opened 2024-05-09
Automate release with JReleaser

### LINKED ISSUE #3602: Automate releases using JReleaser
The idea is to automate the releases so that we are able to release faster with a more regular schedule

filiphr:
Add JReleaser for automating the release and add a step for automating the publishing of the website

This has been heavily inspired by the [Moditect](https://github.com/moditect/moditect/blob/main/.github/workflows/release.yml) and [JReleaser](https://github.com/jreleaser/jreleaser/blob/main/.github/workflows/release.yml) release procedures.

Not sure if @aalmiray, the creator of JReleaser, might be able to chime in and let us know if I'm doing something incorrectly here.  Not sure what the best approach would be to test this with an actual Maven release. 

I think doing the 1.6.0.Beta2 release would be a good first try for this.

Closes #3602

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -21,6 +21,15 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+        <project.repository>mapstruct/mapstruct</project.repository>
+        <local.repository.path>/tmp/repository</local.repository.path>
+        <java.version>1.8</java.version>
+        <maven.compiler.source>${java.version}</maven.compiler.source>
+        <maven.compiler.target>${java.version}</maven.compiler.target>
+        <!-- value comes from property git.commit.author.time -->
+        <project.build.outputTimestamp>${git.commit.author.time}</project.build.outputTimestamp>
+
         <org.mapstruct.gem.version>1.0.0.Alpha3</org.mapstruct.gem.version>
         <org.apache.maven.plugins.enforcer.version>3.4.1</org.apache.maven.plugins.enforcer.version>
         <org.apache.maven.plugins.surefire.version>3.2.2</org.apache.maven.plugins.surefire.version>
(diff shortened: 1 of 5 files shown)


aalmiray:
FWIW you can test most of the release configuration locally. You may run `mvn jreleaser:jreleaserConfig` to print out the resolved configuration. You may also run a release in dry run mode. 

More tips can be found at https://jreleaser.org/guide/latest/checklist.html

parent/pom.xml
```diff
@@ -823,18 +823,93 @@
...
+                                                </artifactOverrides>
+                                            </maven-central>
+                                        </nexus2>
+                                    </maven>
+                                </deploy>
+                                <release>
+                                    <github>
+                                        <tagName>{{projectVersion}}</tagName>
+                                        <releaseName>{{projectVersion}}</releaseName>
+                                        <overwrite>false</overwrite>
```
aalmiray, line 891: This is the default setting and thus may be omitted.

parent/pom.xml
```diff
@@ -823,18 +823,93 @@
...
+                                        <update>
```
aalmiray, line 892: This setting is needed when you would like to update an existing release by modifying: title, release notes, and/or release assets.

parent/pom.xml
```diff
@@ -823,18 +823,93 @@
...
+                                            <enabled>true</enabled>
+                                        </update>
+                                        <changelog>
+                                            <enabled>true</enabled>
+                                            <external>${maven.multiModuleProjectDirectory}/NEXT_RELEASE_CHANGELOG.md</external>
+                                        </changelog>
+                                    </github>
+                                </release>
+                                <files>
```
aalmiray, line 901: Alternatively, if these artifacts match a BINARY or JAVA_BINARY distribution then consider using the `<distributions>` section https://jreleaser.org/guide/latest/concepts/distributions/index.html

parent/pom.xml
filiphr: Good point, I think that I wanted to set the `overwrite` to `true`. However, I might have found a better approach by using the external makrdown.

parent/pom.xml
filiphr: Thanks for the pointer. I was reading up on the binaries, but I couldn't really understand if they are a right fit for this or not. This is non an executable binary, but a zip containing the sources, releases JARs, the Javadoc and the documentation

filiphr:
Thanks a lot for the comments @aalmiray. I was already trying it with the dry run locally and it looked like it is working correctly. 

Thanks a lot for your work on JReleaser, I think automating the release is going to make us release faster 😄

parent/pom.xml
aalmiray: Ah gotcha, then using the `files` section is the way to go.

Issue #3532, opened 2024-02-16
Method already defined when mapping a recursive data structure into a list of the same structure


codeStev:
### Expected behavior

I've got the following data structures I would like to map:

```java
class Source {

    private Node node;

    // getters/setters

}

class Target {

    private List<NodeDto> nodeDtoList;

    // getters/setters

}
```
The recursive structures "Node" and "NodeDto" look like this :

```java
class Node {

    //properties

    List<Node> childNodes;

    // getters/setters

}

class NodeDto {

    //properties

    List<NodeDto> childNodes;

    // getters/setters

}
```

The map method looks as follows: 
```java
@Mapping( target = "nodeList", source = "node.childNodes" )
abstract Target map( Source source);
```

The reason for mapping it like this is to omit the root element of the tree since it is not wanted in this case.

MapStruct generates the methods with the following signature correctly :

```java
NodeDto nodeToNodeDto( Node node )

List<NodeDto> nodeListToNodeDtoList( List<Node> )
```

I would expect MapStruct to use the already generated method to map the list and not to generate the same method twice.

### Actual behavior

MapStruct generates the method   `List<NodeDto> nodeListToNodeDtoList( List<Node> )` twice, provoking a compile error in the generated implementation.
If I don't specify the source the root element will get mapped, which is not wanted behaviour.

The Exception Message looks like this : 

`error: method nodeListToNodeDtoList(List<Node>) is already defined in class TargetMapperImpl`

### Steps to reproduce the problem

My configuration looks as follows: 

```java
@MapperConfig(
        unmappedTargetPolicy = ReportingPolicy.ERROR,
        unmappedSourcePolicy = ReportingPolicy.IGNORE,
        componentModel = MappingConstants.ComponentModel.SPRING,
        injectionStrategy = InjectionStrategy.CONSTRUCTOR,
        nullValueIterableMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT)
public interface CentralMapperConfig {

}
```

The problem should be reproducable with the snippets described above. I omitted the getters/setters and the Mapping class to keep it more readable.

### MapStruct Version

MapStruct 1.5.3

ktkjbs:
I tried to reproduce it with the code below, but it doesn't work. I don't see anything different from your code

this is my code

```Java
// it works!
  @Mapper(componentModel = "spring")
  interface TestMapper {

    @Mapping(source = "rootNode.childNodes", target = "childNodeDtos")
    Target map(Source source);

[...]

PR #3604, opened 2024-05-14
#3591 Potential fix for duplicate methods generated.

### LINKED ISSUE #3591: Compilation Error - Duplicate Methods generated
### Expected behavior
No duplicated methods.

leads to following error: `method beanListToBeanDTOList(java.util.List<Bean>) is already defined`

### Actual behavior

```java
public class BeanMapperImpl implements BeanMapper {
    @Override
    public BeanDTO map(Bean source, BeanDTO target) {
        if ( source == null ) { return target; }
        if ( target.getBeans() != null ) {
            List<BeanDTO> list = beanListToBeanDTOList( source.getBeans() );
            if ( list != null ) {
                target.getBeans().clear();
                target.getBeans().addAll( list );
            } else { target.setBeans( null ); }
        } else {
            List<BeanDTO> list = beanListToBeanDTOList( source.getBeans() );
            if ( list != null ) { target.setBeans( list ); }
        }
        return target;
    }
    protected List<BeanDTO> beanListToBeanDTOList(List<Bean> list) {
        if ( list == null ) { return null; }
        List<BeanDTO> list1 = new ArrayList<BeanDTO>( list.size() );
        for ( Bean bean : list ) { list1.add( beanToBeanDTO( bean ) ); }
        return list1;
    }
    protected BeanDTO beanToBeanDTO(Bean bean) {
        if ( bean == null ) { return null; }
        BeanDTO beanDTO = new BeanDTO();
        beanDTO.setBeans( beanListToBeanDTOList( bean.getBeans() ) );
        return beanDTO;
    }
    protected List<BeanDTO> beanListToBeanDTOList(List<Bean> list) {
        if ( list == null ) { return null; }
        List<BeanDTO> list1 = new ArrayList<BeanDTO>( list.size() );
        for ( Bean bean : list ) { list1.add( beanToBeanDTO( bean ) ); }
        return list1;
    }
}
``` 

SourceFiles to reproduce:
```java
public class Bean {
    private List<Bean> beans;
    public List<Bean> getBeans() { return beans; }
    public void setBeans(List<Bean> beans) { this.beans = beans; }
}
``` 
```java
public class BeanDTO {
    private List<BeanDTO> beans;
    public List<BeanDTO> getBeans() { return beans; }
    public void setBeans(List<BeanDTO> beans) { this.beans = beans; }
}
``` 
```java
@Mapper
public interface BeanMapper {
    BeanMapper INSTANCE = Mappers.getMapper(BeanMapper.class);
    @Mapping(source = "beans", target = "beans")
    BeanDTO map(Bean source, @MappingTarget BeanDTO target);
}
``` 

### MapStruct Version

1.5.5.Final

Hypnagokali:
Fixes #3591

I am not sure if my solution is an acceptable one, but it solves this bug for now.
**Problem**: There were two methods generated and both are written to the mapper class because the selectionParameters in the ContainerMappingMethods are not equal and therefore the methods are not equal and finally both are used.
- The defined mapping has an empty instance of SelectionParameters (returned from MappingOptions)
- The generated mapping has null for selectionParameters.

**Solution**: MappingOption now returns null for selectionParameters if no values have been set.

I am still confused why two methods of beanListToBeanDtoList are forged at all. I would like to take another look at this.


zyberzebra:
> I am not sure if my solution is an acceptable one, but it solves this bug for now. **Problem**: There were two methods generated and both are written to the mapper class because the selectionParameters in the ContainerMappingMethods are not equal and therefore the methods are not equal and finally both are used.
 Good Catch, the question is now what leads to this behaviour? And is your provided solution just a workaround/bandaid for the actual problem. But I tried your solution and it works. 
 I'm also gonna try and have look what causes this.
 
 Maybe it's something caused by the recursive nature here? 
 Bean#Beans#Bean#Beans#......
 And maybe it would generate even more duplicate methods, but it stops before that because of the first duplicate error

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
```diff
@@ -421,7 +421,7 @@ public FormattingParameters getFormattingParameters() {
     }
 
     public SelectionParameters getSelectionParameters() {
-        return selectionParameters;
+        return selectionParameters.isEmpty() ? null : selectionParameters;
```
zyberzebra, line 424: Is it "safe" to just return null here?

Hypnagokali:
> Maybe it's something caused by the recursive nature here? Bean#Beans#Bean#Beans#...... And maybe it would generate even more duplicate methods, but it stops before that because of the first duplicate error

Thanks for the feedback. Yes, the recursive structure of the bean must be the cause, but the MapperCreationProcess ends "successfully" and the duplication error only becomes visible during compilation. At some point, the mapper decides not to create any more methods although the equals method fails.

processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
Hypnagokali: This is no longer necessary, I will remove it with the next commit.

Hypnagokali:
I have added a test for StreamMapping and MapMapping.
It works, but actually both mapping methods are generated twice. However, finally only one method is added to `MappingBuilderContext#mappingsToGenerate`, because their equals methods do not depend on `selectionParameters`.

Should I fix this behaviour even though it does not lead to an error?
I would then implement a `createOrGetForgedAssignment` method in `AbstractBaseBuilder` that checks if there is a forged method under creation.

Hypnagokali:
Okay, everything should be done now. I hope I haven't missed anything.

I still have a question:
All workflow jobs fail with: "Some files do not have the expected license header", what does that mean?

thunderhook:
> Okay, everything should be done now. I hope I haven't missed anything.
> 
> I still have a question: All workflow jobs fail with: "Some files do not have the expected license header", what does that mean?

This is a check if every file contains the copyright header at the first lines:
```
/*
 * Copyright MapStruct Authors.
 *
 * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
 */
```

processor/src/test/java/org/mapstruct/ap/test/bugs/_3591/CompilationErrorDuplicateMethodsBugTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+        assertThat( dto.getBeanMap() ).isNotEmpty();
+        assertThat( dto.getBeanMap().get( "child" ).getValue() ).isEqualTo( "mapChild" );
+
+        List<ContainerBeanDto> fromStream = dto.getBeanStream().collect( Collectors.toList() );
+        assertThat( fromStream ).isNotEmpty();
+        assertThat( fromStream.get( 0 ).getValue() ).isEqualTo( "streamChild" );
+    }
+
+    static ContainerBean testContainerBean() {
```
zyberzebra, line 74: Should these helper methods really be own methods? They are only used 1 time in their respective tests.

processor/src/test/java/org/mapstruct/ap/test/bugs/_3591/CompilationErrorDuplicateMethodsBugTest.java
```diff
@@ -0,0 +1,97 @@
...
+        generatedSource.addComparisonToFixtureFor( ContainerBeanMapper.class );
+    }
+
+    @ProcessorTest
+    @WithClasses({
+        ContainerBean.class,
+        ContainerBeanDto.class,
+        ContainerBeanMapper.class,
+    })
+    void shouldMapNestedMapAndStream() {
```
zyberzebra, line 61: Inlined and with some assertJ features: 
```java
    void shouldMapNestedMapAndStream() {
        ContainerBean containerBean = new ContainerBean( "parent" );
        containerBean.setBeanMap( Map.of( "child", new ContainerBean( "mapChild" ) ) );

        ContainerBean streamChild = new ContainerBean( "streamChild" );
        containerBean.setBeanStream( Stream.of( streamChild ) );

        ContainerBeanDto dto = ContainerBeanMapper.INSTANCE.mapWithMapMapping( containerBean, new ContainerBeanDto() );

        assertThat( dto.getBeanMap() )
            .extractingByKey( "child" )
            .extracting( ContainerBeanDto::getValue )
            .isEqualTo( "mapChild" );

        assertThat( dto.getBeanStream() )
            .singleElement()
            .extracting( ContainerBeanDto::getValue )
            .isEqualTo( "streamChild" );
    }
``` 
Maybe you see something you like :)

zyberzebra:
It's working! Pretty cool! :) Lets see what the maintainers have to say

processor/src/main/java/org/mapstruct/ap/internal/model/AbstractBaseBuilder.java
```diff
@@ -121,7 +140,7 @@ Assignment createForgedAssignment(SourceRHS source, ForgedMethod methodRef, Mapp
         return createAssignment( source, methodRef );
     }
 
-    private Assignment createAssignment(SourceRHS source, ForgedMethod methodRef) {
+    Assignment createAssignment(SourceRHS source, ForgedMethod methodRef) {
```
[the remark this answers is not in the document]
Hypnagokali: You are right. Just a relic from my second draft. I have fixed that.

processor/src/test/java/org/mapstruct/ap/test/bugs/_3591/CompilationErrorDuplicateMethodsBugTest.java
Hypnagokali: That is much better and more readable. Thanks 👍

processor/src/main/java/org/mapstruct/ap/internal/model/ContainerMappingMethod.java
```diff
@@ -46,7 +46,14 @@ public abstract class ContainerMappingMethod extends NormalTypeMappingMethod {
             afterMappingReferences );
         this.elementAssignment = parameterAssignment;
         this.loopVariableName = loopVariableName;
-        this.selectionParameters = selectionParameters;
+
+        if ( selectionParameters == null ) {
+            this.selectionParameters = SelectionParameters.forSourceRHS( null );
```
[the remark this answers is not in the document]
Hypnagokali: You are right - it could be confusing. I have changed that to `SelectionParameters.emptyInstance()`

zyberzebra:
Is this Issue related? https://github.com/mapstruct/mapstruct/issues/3532

filiphr:
Nice work @Hypnagokali and @zyberzebra. Good catch with the `null` and empty `SelectionParameters`.

I pushed one small polish commit. I'll merge this once the build is done

filiphr:
And you are right @zyberzebra #3532 does look identical to this one. I'll close it out

PR #3615, opened 2024-05-28
#3609 Add test case for issue 3609

### LINKED ISSUE #3609: @SubclassMapping not working with @BeanMapping.ignoreUnmappedSourceProperties
### Expected behavior

`@BeanMapping.ignoreUnmappedSourceProperties` should not be ignored.

### Actual behavior

I want to use `@SubclassMapping` and `@BeanMapping.ignoreUnmappedSourceProperties` on the same mapping method. The `@BeanMapping.ignoreUnmappedSourceProperties` seems to be ignored. If I remove the  `@SubclassMapping` annotations the error (Unmapped source property: "id") disappears.

```java
class Vehicle {
    public int price;
}

class Car extends Vehicle {
    public int seats;
}

class Truck extends Vehicle {
    public int capacity;
}

class VehicleDto {
    public int id;
    public int price;
}

class CarDto extends VehicleDto {
    public int seats;
}

class TruckDto extends VehicleDto {
    public int capacity;
}

@MapperConfig(
)
class Config {
}
@Mapper(config = Config.class)
public abstract class VehicleMapper {

    @SubclassMapping(source = CarDto.class, target = Car.class)
    @SubclassMapping(source = TruckDto.class, target = Truck.class)
    @BeanMapping(ignoreUnmappedSourceProperties = { "id" })
    public abstract Vehicle toVehicle(VehicleDto vehicle);

    @InheritInverseConfiguration
    @Mapping(target = "id", ignore = true)
    public abstract VehicleDto toVehicleDto(Vehicle vehicle);
}

```

### MapStruct Version

1.5.5

Obolrom:
Fixes #3609 

- [x] Add test case to cover the bug
- [x] Fix bug

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
@@ -55,15 +55,26 @@ public class BeanMappingOptions extends DelegatingOptions {
         return options;
     }
 
-    public static BeanMappingOptions forForgedMethods(BeanMappingOptions beanMapping) {
-        BeanMappingOptions options = new BeanMappingOptions(
+    public static BeanMappingOptions forForgedMethods(MappingMethodOptions mappingMethodOptions) {
+
+        BeanMappingOptions beanMapping = mappingMethodOptions.getBeanMapping();
+
+        return new BeanMappingOptions(
             beanMapping.selectionParameters != null ?
                 SelectionParameters.withoutResultType( beanMapping.selectionParameters ) : null,
-            Collections.emptyList(),
+            takeOverIgnoreUnmappedSourceProperties( mappingMethodOptions ),
             beanMapping.beanMapping,
             beanMapping
         );
-        return options;
+    }
+
+    private static List<String> takeOverIgnoreUnmappedSourceProperties(MappingMethodOptions mappingMethodOptions) {
+
+        if (!mappingMethodOptions.getSubclassMappings().isEmpty()) {
+            return mappingMethodOptions.getBeanMapping().getIgnoreUnmappedSourceProperties();
+        }
+
+        return Collections.emptyList();
     }
 
     public static BeanMappingOptions empty(DelegatingOptions delegatingOptions) {
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
@@ -55,26 +55,26 @@ public class BeanMappingOptions extends DelegatingOptions {
         return options;
     }
 
-    public static BeanMappingOptions forForgedMethods(MappingMethodOptions mappingMethodOptions) {
+    public static BeanMappingOptions forForgedMethods(BeanMappingOptions beanMapping) {
+        BeanMappingOptions options = new BeanMappingOptions(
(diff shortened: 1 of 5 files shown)


filiphr:
@Obolrom do you already have some ideas about the bug fix or do you need me to have a look at it as well?

Obolrom:
@filiphr, I am trying to figure out how to fix the bug properly.
As I understand we should do something around these properties 
```java
private Map<String, Accessor> unprocessedConstructorProperties;
private Map<String, Accessor> unprocessedTargetProperties;
private Map<String, Accessor> unprocessedSourceProperties;
private Set<String> missingIgnoredSourceProperties;
```
in the `BeanMappingMethod` class.

I would appreciate any hint

filiphr:
@Obolrom, I believe that it is somewhere in the `BeanMappingMethos.Builder#createSubclassMapping`. Or more specifically in `forgeSubclassMapping`. I think that we are not passing all the information from the `BeanMapping` into the forged methods, but looking at the code it should work. Perhaps it is best if I look into it, no need to waste on time on this.

filiphr:
@Obolrom did my hint help you out a bit? If not, I'll take over and do the fix.

Obolrom:
@filiphr by some reason it doesn't work anyway. Could you please take a look?

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
```diff
@@ -55,15 +55,26 @@ public static BeanMappingOptions forInheritance(BeanMappingOptions beanMapping,
...
-        BeanMappingOptions options = new BeanMappingOptions(
+    public static BeanMappingOptions forForgedMethods(MappingMethodOptions mappingMethodOptions) {
+
+        BeanMappingOptions beanMapping = mappingMethodOptions.getBeanMapping();
+
+        return new BeanMappingOptions(
             beanMapping.selectionParameters != null ?
                 SelectionParameters.withoutResultType( beanMapping.selectionParameters ) : null,
-            Collections.emptyList(),
+            takeOverIgnoreUnmappedSourceProperties( mappingMethodOptions ),
```
thunderhook, line 65: The easy fix would've been to just take over the `beanMapping.getIgnoreUnmappedSourceProperties()`. However this led to an error in the `Issue2743Test` (`Issue2743Mapper`)...

processor/src/main/java/org/mapstruct/ap/internal/model/source/BeanMappingOptions.java
```diff
@@ -55,15 +55,26 @@ public static BeanMappingOptions forInheritance(BeanMappingOptions beanMapping,
...
             beanMapping.beanMapping,
             beanMapping
         );
-        return options;
+    }
+
+    private static List<String> takeOverIgnoreUnmappedSourceProperties(MappingMethodOptions mappingMethodOptions) {
+
+        if (!mappingMethodOptions.getSubclassMappings().isEmpty()) {
```
thunderhook, line 73: ...so that it is now only used when dealing with subclass mappings. Might argue for possibly refining that.

filiphr:
Thanks @Obolrom and @thunderhook for working on this.

@thunderhook I was not too happy with the fix. I changed it a bit. Have a look at how it looks like, I prefer when things are a bit more explicit than having something special in the generic method.

@Obolrom, I reduced the test case a bit to only cover what the actual bug is and reduce the complexity of it.

PR #3616, opened 2024-05-28
#2421 support for target this enum mapping.

### LINKED ISSUE #2421: MapStruct Exception: “enum” does not have an accessible constructor
When I upgrade Mapstruct from "1.3.1.Final" to "1.4.2.Final" , I face this error.
`CompanyType does not have an accessible constructor.`
Mapstruct cannot map enum to string in this version of Mapstruct and all of my mapping class face this error. and I need this version because of some features of this version.
please help me to solve this exception.

My code :
```java
@Entity
@Getter
@Setter
@Table(name = "CORE_COMPANY")
@NoArgsConstructor
@AllArgsConstructor
public class Company  {

    @Id
    @GeneratedValue(generator = "uuid2")
    @GenericGenerator(name = "uuid2", strategy = "uuid2")
    private String id;

    @NotNull
    @Size(min = 2, max = 50)
    @Column(name = "NAME", nullable = false)
    private String name;

    @NotNull
    @Column(name = "ADDRESS", nullable = false)
    private String address;

    @NotNull
    @Enumerated(EnumType.STRING)
    @Column(name = "COMPANY_TYPE", nullable = false)
    private CompanyType companyType;

    @Column(name = "IS_ACTIVE")
    private Boolean active;

}

@Getter
@AllArgsConstructor
public enum CompanyType {
    insurance("Insurance", "insurance"),
    inspection("Inspection", "inspectionem"),
    Shipping("Shipping", "naviportans");

    private final String title;
    private final String latinTitle;
}

@Getter
@Setter
@NoArgsConstructor
@AllArgsConstructor
public class CompanyDTO  {
    private String id;
    private String name;
    private String address;
    private String companyType;
    private String companyTypeLatinTitle;
    private Boolean active;
}


public interface CompanyMapper extends EntityMapper<CompanyDTO, Company> {
    CompanyMapper INSTANCE = Mappers.getMapper(CompanyMapper.class);

    @Mapping(target = "companyTypeLatinTitle", source = "companyType.latinTitle")
    @Mapping(target = "countryId", source = "country.id")
    @Mapping(target = "countryTitle", source = "country.title")
    CompanyDTO toDto(Company entitiy) ;

    @InheritInverseConfiguration
    Company toEntity(CompanyDTO dto) ;

}

```

Zegveld:
fixes #2421

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -235,7 +235,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
             this.unprocessedTargetProperties = new LinkedHashMap<>( accessors );
 
-            if ( !method.isUpdateMethod() && !hasFactoryMethod ) {
+            if ( !method.isUpdateMethod() && !hasFactoryMethod && !isCorrectlyDefinedTargetThisForEnumTarget() ) {
                 ConstructorAccessor constructorAccessor = getConstructorAccessor( resultTypeToMap );
                 if ( constructorAccessor != null ) {
 
@@ -310,6 +310,8 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                 // map parameters without a mapping
                 applyParameterNameBasedMapping();
 
+                // map this enum mapping
+                applyTargetThisMappingForEnumTarget();
             }
 
             // Process the unprocessed defined targets
@@ -514,7 +516,8 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         private boolean isAbstractReturnTypeAllowed() {
             return !method.getOptions().getSubclassMappings().isEmpty()
                 && ( method.getOptions().getBeanMapping().getSubclassExhaustiveStrategy().isAbstractReturnTypeAllowed()
-                    || isCorrectlySealed() );
+                    || isCorrectlySealed() )
+                || isCorrectlyDefinedTargetThisForEnumTarget();
         }
 
         private boolean isCorrectlySealed() {
@@ -756,7 +759,9 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
 
         private boolean isReturnTypeAbstractOrCanBeConstructed(Type returnType) {
             boolean error = true;
-            if ( !returnType.isAbstract() && !returnType.hasAccessibleConstructor() ) {
+            if ( !returnType.isAbstract()
+              && !returnType.hasAccessibleConstructor()
+              && !isCorrectlyDefinedTargetThisForEnumTarget() ) {
                 ctx
                    .getMessager()
                    .printMessage(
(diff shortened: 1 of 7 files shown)


thunderhook:
Thanks for your work on this @Zegveld!

I checked it out and played with it a bit, and yes, this solves the problem when an enum is used as a target via `target = "."`.
My mistake, I thought it was the same problem as described in #2421. It just has the same error message but a different cause.

The reproducer with an inherit inverse configuration provided by @filiphr is still not solved.
See https://github.com/mapstruct/mapstruct/commit/488d95121f94e04668855e43494c9469b2761f24 and just comment out this line:

* https://github.com/mapstruct/mapstruct/commit/488d95121f94e04668855e43494c9469b2761f24#diff-c07cc7f117120d70e073cd445da13c9fe570d3c0d7999504b0f6ccb71bd71b3eR21

So I have created another issue for your fix, which then solves #3437: #3617

I have relinked the PR with the new issue.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -1532,6 +1537,38 @@ private void applyTargetThisMapping() {
...
+         */
+        private void applyTargetThisMappingForEnumTarget() {
+            if ( !isCorrectlyDefinedTargetThisForEnumTarget() ) {
+                return;
+            }
+
+            for ( MappingReference targetThis : this.targetThisReferences ) {
+                PropertyMapping propertyMapping = new PropertyMappingBuilder().mappingContext( ctx )
+                    .sourceMethod( method )
+                    .sourcePropertyName( "test" )
```
thunderhook, line 1551: This seems to have no effect in this case, however this seems fishy - maybe just not set `sourcePropertyName` at all?

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -2115,6 +2152,10 @@ private boolean doesNotNeedPresenceCheckForSourceParameter(PropertyMapping mappi
         return mapping.getAssignment().isSourceReferenceParameter();
     }
 
+    public boolean shouldDirectlyReturnOnlyMappingResult() {
+        return this.returnTypeToConstruct.isEnumType() && this.propertyMappings.size() == 1;
```
thunderhook, line 2156: Maybe we could come up with more tests. I have no idea what scenarios could be tested, but after checking the coverage report we could provide a test where `this.propertyMappings.size()` is `!= 1`

Not quite sure though, my coverage report seems to be a bit buggy.

processor/src/test/java/org/mapstruct/ap/test/bugs/_2421/Issue2421Test.java
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._2421;
+
+import org.mapstruct.ap.testutil.ProcessorTest;
+import org.mapstruct.ap.testutil.WithClasses;
+
+public class Issue2421Test {
```
thunderhook, line 11: Sorry, we would need to change the issue number to 3617 - might aswell add @IssueKey("3617") here.

thunderhook (commented):
I don't have enough experience yet and would like @filiphr to take another look at it, I just have a few comments from my side.

filiphr:
I need to have a better look at this. However, I don't like the fact that this is done using the `BeanMappingMethod`. What might make more sense is to do this as part of the enum mapping and support `source` in `@EnumMapping`. This way people can still use `@ValueMapping` and basically it would be a shortcut to writing it by hand.

We currently support string to enum. The solution for this would be to convert the `EnumMapping#source` (or whatever we name that) to a method that just delegates to another method (which we already have and would be able to handle everything that we already have).

Zegveld:
> I need to have a better look at this. However, I don't like the fact that this is done using the `BeanMappingMethod`. What might make more sense is to do this as part of the enum mapping and support `source` in `@EnumMapping`. This way people can still use `@ValueMapping` and basically it would be a shortcut to writing it by hand.

You can actually use `@ValueMapping` with this solution. This is even present in the current test example.

> We currently support string to enum. The solution for this would be to convert the `EnumMapping#source` (or whatever we name that) to a method that just delegates to another method (which we already have and would be able to handle everything that we already have).

It would indeed be better to introduce a `source` property on the `EnumMapping` method, and perhaps throw an error on the current situation which refers to that route. It does feel like the better solution, then what I've currently implemented.

filiphr:
>You can actually use `@ValueMapping` with this solution. This is even present in the current test example.

Indeed, I missed that. 

>It would indeed be better to introduce a `source` property on the `EnumMapping` method, and perhaps throw an error on the current situation which refers to that route. It does feel like the better solution, then what I've currently implemented.

I think that we arleady have an error when using `@Mapping` with `EnumMapping`. It used to be supported, but I believe we removed it. Have a look at `Message.ENUMMAPPING_REMOVED`.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -514,7 +516,8 @@ private SubclassMapping createSubclassMapping(SubclassMappingOptions subclassMap
         private boolean isAbstractReturnTypeAllowed() {
             return !method.getOptions().getSubclassMappings().isEmpty()
                 && ( method.getOptions().getBeanMapping().getSubclassExhaustiveStrategy().isAbstractReturnTypeAllowed()
```
ronodhirSoumik, line 518: just to know-
```
( method.getOptions().getBeanMapping().getSubclassExhaustiveStrategy().isAbstractReturnTypeAllowed()
                    || isCorrectlySealed() 
                || isCorrectlyDefinedTargetThisForEnumTarget() ); ??
```

PR #3632, opened 2024-06-30
#3574 Respect only explicit mappings but fail on unmapped source fields

### LINKED ISSUE #3574: Respect only explicit mappings but fail on unmapped source fields
### Use case

Imagine a PATCH request with several JsonNullable<T> fields.
```java
class MyDto {
  private JsonNullable<String> firstname;
  private JsonNullable<String> surname
}
```

The mapper should overwrite present fields only:

```java
@Mapper(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.IGNORE,
        unmappedSourcePolicy = ReportingPolicy.ERROR)
interface MyPartialUpdateMapper {

  @Mapping(source = "firstname", target = "firstname", conditionExpression = "java(updatedContent.getFirstname().isPresent())")
  @BeanMapping(ignoreByDefault = true)
  void updateMyDomainObject(@MappingTarget MyDomainObject domain, MyDto updatedContent);

  default <T> T fromJsonNullable(JsonNullable<T> nullableValue) {
      if(nullableValue.isPresent()) {
          return nullableValue.get();
      }
      return null;
  }
}
```

The `@BeanMapping(ignoreByDefault = true)` is used to ensure that only those fields in the domain object that are set in the Patch-DTO are overwritten. Unfortunately, the `unmappedSourcePolicy = ReportingPolicy.ERROR` is ignored in this case.

As far as I can see, there is no way to enforce explicit mapping and cause the build to fail if one of the mappings is forgotten. This feature would make the mappings much better maintainable in such situations.

### Generated Code

_No response_

### Possible workarounds

_No response_

### MapStruct Version

1.5.3.Final

filiphr:
This reverts #2560, because we've decided that `@BeanMapping(ignoreByDefault = true)` should only be applied to target properties and not to source properties. Source properties are anyway ignored, the `BeanMapping#unmappedSourcePolicy` should be used to control what should happen with unmapped source policy

Fixes #3574

### DIFF
--- a/NEXT_RELEASE_CHANGELOG.md
+++ b/NEXT_RELEASE_CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### Enhancements
 
+* Breaking change: Respect only explicit mappings but fail on unmapped source fields (#3574) - 
+This reverts #2560, because we've decided that `@BeanMapping(ignoreByDefault = true)` should only be applied to target properties and not to source properties. 
+Source properties are anyway ignored, the `BeanMapping#unmappedSourcePolicy` should be used to control what should happen with unmapped source policy
+
 ### Bugs
 
 ### Documentation
--- a/NEXT_RELEASE_CHANGELOG.md
+++ b/NEXT_RELEASE_CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ### Enhancements
 
-* Breaking change: Respect only explicit mappings but fail on unmapped source fields (#3574) - 
+* Breaking change:g (#3574) - 
 This reverts #2560, because we've decided that `@BeanMapping(ignoreByDefault = true)` should only be applied to target properties and not to source properties. 
-Source properties are anyway ignored, the `BeanMapping#unmappedSourcePolicy` should be used to control what should happen with unmapped source policy
+Source properties are ignored anyway, the `BeanMapping#unmappedSourcePolicy` should be used to control what should happen with unmapped source policy
 
 ### Bugs
 
--- a/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
@@ -60,7 +60,7 @@ public interface SourceTargetMapper {
 ----
 ====
 
-The example demonstrates how the source properties `time` and `format` are composed into one target property `TimeAndFormat`. Please note that the fully qualified package name is specified because MapStruct does not take care of the import of the `TimeAndFormat` class (unless it's used otherwise explicitly in the `ErroneousSourceTargetMapperWithIgnoreByDefault`). This can be resolved by defining `imports` on the `@Mapper` annotation.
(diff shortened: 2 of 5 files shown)


documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
```diff
@@ -60,7 +60,7 @@ public interface SourceTargetMapper {
 ----
 ====
 
-The example demonstrates how the source properties `time` and `format` are composed into one target property `TimeAndFormat`. Please note that the fully qualified package name is specified because MapStruct does not take care of the import of the `TimeAndFormat` class (unless it's used otherwise explicitly in the `SourceTargetMapper`). This can be resolved by defining `imports` on the `@Mapper` annotation.
+The example demonstrates how the source properties `time` and `format` are composed into one target property `TimeAndFormat`. Please note that the fully qualified package name is specified because MapStruct does not take care of the import of the `TimeAndFormat` class (unless it's used otherwise explicitly in the `ErroneousSourceTargetMapperWithIgnoreByDefault`). This can be resolved by defining `imports` on the `@Mapper` annotation.
```
thunderhook, line 63: Attention, I think you changed it here unintentionally (due to an IntelliJ rename)

documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
```diff
@@ -108,7 +108,7 @@ public interface SourceTargetMapper {
 ----
 ====
 
-The example demonstrates how to use defaultExpression to set an `ID` field if the source field is null, this could be used to take the existing `sourceId` from the source object if it is set, or create a new `Id` if it isn't. Please note that the fully qualified package name is specified because MapStruct does not take care of the import of the `UUID` class (unless it’s used otherwise explicitly in the `SourceTargetMapper`). This can be resolved by defining imports on the @Mapper annotation (see <<expressions>>).
+The example demonstrates how to use defaultExpression to set an `ID` field if the source field is null, this could be used to take the existing `sourceId` from the source object if it is set, or create a new `Id` if it isn't. Please note that the fully qualified package name is specified because MapStruct does not take care of the import of the `UUID` class (unless it’s used otherwise explicitly in the `ErroneousSourceTargetMapperWithIgnoreByDefault`). This can be resolved by defining imports on the @Mapper annotation (see <<expressions>>).
```
thunderhook, line 111: Attention, I think you changed it here unintentionally (due to an IntelliJ rename)

NEXT_RELEASE_CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ### Enhancements
 
+* Breaking change: Respect only explicit mappings but fail on unmapped source fields (#3574) - 
+This reverts #2560, because we've decided that `@BeanMapping(ignoreByDefault = true)` should only be applied to target properties and not to source properties. 
+Source properties are anyway ignored, the `BeanMapping#unmappedSourcePolicy` should be used to control what should happen with unmapped source policy
```
thunderhook, line 7: nitpick: `anyway ignored` -> `ignored anyway`

thunderhook (changes requested):
Just a few nit picky notes, and of course Checkstyle makes an appearance 😉

filiphr:
Thanks for the feedback @thunderhook. I've applied the changes

thunderhook (approved):

PR #3642, opened 2024-07-15
disable ignoreEnhancedForColon in checkstyle

hduelme:
@thunderhook found in mapstruct/mapstruct-idea#200 that enhanced for loops are currently not check for whitespaces. I enable the check and also applied the formatting.


thunderhook:
Thank you for taking the initiative @hduelme !

PR #3675, opened 2024-08-17
#3652 Inverse Inheritance possible for ignore-mappings without source

### LINKED ISSUE #3652: Inverse Inheritance Strategy not working for ignored mappings only with target
### Expected behavior

Inverse Configuration is autmatically generated.

### Actual behavior

`@Mappings` have to be defined twice

### Setup
```java
@Mapper(config = FooBarConfig.class)
public interface FooBarMapper {

    Bar toBar(Foo foo);

    Foo toFoo(Bar bar);

}
@Data
class Foo { int secret; }
@Data
class Bar { int secret; }

@MapperConfig(mappingInheritanceStrategy = MappingInheritanceStrategy.AUTO_INHERIT_ALL_FROM_CONFIG)
interface FooBarConfig {

    @Mapping(target = "secret", ignore = true)
    Bar toBar(Foo foo);

    @InheritInverseConfiguration(name = "toBar")
    Foo toFoo(Bar bar);
}
``` 

### Generated

```java
public class FooBarMapperImpl implements FooBarMapper {

    @Override
    public Bar toBar(Foo foo) {
        if ( foo == null ) {
            return null;
        }

        Bar bar = new Bar();

        return bar;
    }

    @Override
    public Foo toFoo(Bar bar) {
        if ( bar == null ) {
            return null;
        }

        Foo foo = new Foo();

        foo.setSecret( bar.getSecret() );

        return foo;
    }
}
``` 
### Only way to get it working
```java
@MapperConfig(mappingInheritanceStrategy = MappingInheritanceStrategy.AUTO_INHERIT_FROM_CONFIG)
interface FooBarConfig {
    @Mapping(target = "secret", ignore = true)
    Bar toBar(Foo foo);
    @Mapping(target = "secret", ignore = true)
    Foo toFoo(Bar bar);
}
``` 

### MapStruct Version

1.5.5

Hypnagokali:
Fixes #3652

I have added a test case where the target has a property that does not exist on the source type.
The inverse inheritation still works, because this property is then simply ignored in `BeanMappingMethod#handleDefinedMapping` (line 1240):

```java
else if ( inheritContext.isReversed() ) {
    // When a configuration is reverse inherited and there are no read or write accessor
    // then we should ignore this mapping.
    // This most likely means that we were mapping the source parameter to the target.
    // If the error is due to something else it will be reported on the original mapping
    return false;
}
```

I have another question:
The javadoc of `MappingOptions#canInverse` says:
>mapping can only be inversed if the source was not a constant nor an expression **nor a nested property**

But there is no check for a nested property in this method and it seems to work: Some tests uses `@InheritInverseConfiguration` with nested properties:
e.g.: `ReversingNestedSourcePropertiesConstructorTest#shouldGenerateNestedWithConfigReverse`

Should I remove this statement from the javadoc comment?

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
@@ -479,7 +479,7 @@ public class MappingOptions extends DelegatingOptions {
     }
 
     /**
-     *  mapping can only be inversed if the source was not a constant nor an expression nor a nested property
+     *  Mapping can only be inversed if the source was not a constant nor an expression
      *
      * @return true when the above applies
      */
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
@@ -480,12 +480,11 @@ public class MappingOptions extends DelegatingOptions {
 
     /**
      *  mapping can only be inversed if the source was not a constant nor an expression nor a nested property
-     *  and the mapping is not a 'target-source-ignore' mapping
      *
      * @return true when the above applies
      */
     public boolean canInverse() {
-        return constant == null && javaExpression == null && !( isIgnored && sourceName == null );
+        return constant == null && javaExpression == null;
     }
 
     public MappingOptions copyForInverseInheritance(SourceMethod templateMethod,
--- a/NEXT_RELEASE_CHANGELOG.md
+++ b/NEXT_RELEASE_CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Bugs
 
+* Inverse Inheritance Strategy not working for ignored mappings only with target (#3652)
+
 ### Documentation
 
 ### Build
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_3652/Bar.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_3652/Bar.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package org.mapstruct.ap.test.bugs._3652;
+
+public class Bar {
+
+    private int secret;
+    private int doesNotExistInFoo;
+
+    public int getSecret() {
+        return secret;
+    }
+
+    public void setSecret(int secret) {
+        this.secret = secret;
+    }
+
+    public int getDoesNotExistInFoo() {
+        return doesNotExistInFoo;
+    }
+
+    public void setDoesNotExistInFoo(int doesNotExistInFoo) {
+        this.doesNotExistInFoo = doesNotExistInFoo;
+    }
+
+}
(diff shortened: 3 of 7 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/MappingOptions.java
```diff
@@ -480,12 +480,11 @@ public MappingControl getMappingControl(ElementUtils elementUtils) {
 
     /**
      *  mapping can only be inversed if the source was not a constant nor an expression nor a nested property
-     *  and the mapping is not a 'target-source-ignore' mapping
```
zyberzebra, line 483: I'm still curious why this was explicitly stated here.

filiphr: This looks like an ancient thing. The `MappingOptions` was called `Mapping` at some point. The rename was done in https://github.com/mapstruct/mapstruct/commit/7bee12138da5f601364987d053e4fd3eeb43cbcd.

The line `isIgnored && sourceName == null` was added first in https://github.com/mapstruct/mapstruct/commit/4d8bc29347e37bcf22d121829802e7f58603afbb. And looking at that commit it seems like the logic was not ported properly, because prior to that we had

```java
        // mapping can only be reversed if the source was not a constant nor an expression nor a nested property
        if ( constant != null || javaExpression != null ) {
            return null;
        }

        // should only ignore a property when 1) there is a sourceName defined or 2) there's a name match
        if ( isIgnored ) {
            if ( sourceName == null && !hasPropertyInReverseMethod( targetName, method ) ) {
                return null;
            }
        }
```

and after that we had

```java
        // mapping can only be reversed if the source was not a constant nor an expression nor a nested property
        // and the mapping is not a 'target-source-ignore' mapping
        if ( constant != null || javaExpression != null || ( isIgnored && sourceName == null ) ) {
            return null;
        }
```

I cannot say why this was the case back then

filiphr: I was curious too @zyberzebra. This looks like an ancient thing. The `MappingOptions` was called `Mapping` at some point. The rename was done in https://github.com/mapstruct/mapstruct/commit/7bee12138da5f601364987d053e4fd3eeb43cbcd.

The line `isIgnored && sourceName == null` was added first in https://github.com/mapstruct/mapstruct/commit/4d8bc29347e37bcf22d121829802e7f58603afbb. And looking at that commit it seems like the logic was not ported properly, because prior to that we had

```java
        // mapping can only be reversed if the source was not a constant nor an expression nor a nested property
        if ( constant != null || javaExpression != null ) {
            return null;
        }

        // should only ignore a property when 1) there is a sourceName defined or 2) there's a name match
        if ( isIgnored ) {
            if ( sourceName == null && !hasPropertyInReverseMethod( targetName, method ) ) {
                return null;
            }
        }
```

and after that we had

```java
        // mapping can only be reversed if the source was not a constant nor an expression nor a nested property
        // and the mapping is not a 'target-source-ignore' mapping
        if ( constant != null || javaExpression != null || ( isIgnored && sourceName == null ) ) {
            return null;
        }
```

Things were quite different back then. We are now handling the target later and differently to back then, so the fix done by @Hypnagokali looks like spot on.

filiphr:
Thanks a lot for the quick fix @Hypnagokali. Nice spotting this.

>Should I remove this statement from the javadoc comment? 

Good catch @Hypnagokali. I think that we can remove the _nor a nested property_ statement from the Javadoc. We are clearly not doing that there 😄

Hypnagokali:
All right, I have adjusted the comment :blush:
Thanks for reviewing @filiphr, @zyberzebra

filiphr:
Thanks @Hypnagokali

PR #3680, opened 2024-08-21
#3678 Prevent duplicate @BeforeMapping and @AfterMapping calls on classes that use the Builder pattern.

### LINKED ISSUE #3678: @AfterMapping methods are called twice when target uses Lombok @Builder annotation
### Expected behavior

Generated code calls `@AfterMapping` method only once. Here is the generated output for MapStruct 1.5.5:

```java
@Generated(
    date = "2024-08-19T16:21:29+0200",
    comments = "version: 1.5.5.Final, compiler: javac, environment: Java 21.0.2 (Oracle Corporation)"
)
class MappingExample$SimpleSourceDestinationMapperImpl extends MappingExample.SimpleSourceDestinationMapper {

    @Override
    MappingExample.SimpleDestination sourceToDestination(MappingExample.SimpleSource source) {
        if ( source == null ) {
            return null;
        }

        MappingExample.SimpleDestination.SimpleDestinationBuilder simpleDestination = MappingExample.SimpleDestination.builder();

        simpleDestination.name( source.getName() );
        simpleDestination.description( source.getDescription() );

        afterMapping( source );

        return simpleDestination.build();
    }
}
```

### Actual behavior

`@AfterMapping` method is called before and after using the build() method of the builder pattern.

```java
@Generated(
    date = "2024-08-19T16:24:57+0200",
    comments = "version: 1.6.0, compiler: javac, environment: Java 21.0.2 (Oracle Corporation)"
)
class MappingExample$SimpleSourceDestinationMapperImpl extends MappingExample.SimpleSourceDestinationMapper {

    @Override
    MappingExample.SimpleDestination sourceToDestination(MappingExample.SimpleSource source) {
        if ( source == null ) {
            return null;
        }

        MappingExample.SimpleDestination.SimpleDestinationBuilder simpleDestination = MappingExample.SimpleDestination.builder();

        simpleDestination.name( source.getName() );
        simpleDestination.description( source.getDescription() );

        afterMapping( source );

        SimpleDestination simpleDestinationResult = simpleDestination.build();

        afterMapping( source );

        return simpleDestinationResult;
    }
}
```

pom.xml
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>com.example</groupId>
    <artifactId>test-mapstruct-issue</artifactId>
    <version>1.0-SNAPSHOT</version>

    <properties>
        <maven.compiler.source>21</maven.compiler.source>
        <maven.compiler.target>21</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    </properties>
    <dependencies>
        <dependency>
            <version>1.6.0</version>
        </dependency>
        <dependency>
            <groupId>org.projectlombok</groupId>
            <artifactId>lombok</artifactId>
            <version>1.18.34</version>
        </dependency>
        <dependency>
            <groupId>org.projectlombok</groupId>
            <artifactId>lombok-mapstruct-binding</artifactId>
            <version>0.2.0</version>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <version>3.11.0</version>
                <configuration>
                    <annotationProcessorPaths>
                        <path>
                            <version>1.6.0</version>
                        </path>
                        <path>
                            <groupId>org.projectlombok</groupId>
                            <artifactId>lombok</artifactId>
                            <version>1.18.34</version>
                        </path>
                        <path>
                            <groupId>org.projectlombok</groupId>
                            <artifactId>lombok-mapstruct-binding</artifactId>
                            <version>0.2.0</version>
                        </path>
                    </annotationProcessorPaths>
                </configuration>
            </plugin>
        </plugins>
    </build>
</project>
```

Code example:
```java
import lombok.Builder;
import lombok.Data;
import org.mapstruct.AfterMapping;
import org.mapstruct.Mapper;

public class MappingExample {

  @Data
  public static class SimpleSource {
    private String name;
    private String description;
  }

  @Data
  @Builder
  public static final class SimpleDestination {
    private final String name;
    private final String description;

  }

  @Mapper
  abstract static class SimpleSourceDestinationMapper {

    abstract SimpleDestination sourceToDestination(SimpleSource source);

    @AfterMapping
    void afterMapping(SimpleSource simpleSource) {
      // do something
    }
  }
}
```

### MapStruct Version

MapStruct 1.6.0

thunderhook:
Fixes #3678


processor/src/test/java/org/mapstruct/ap/test/builder/lifecycle/BuilderLifecycleCallbacksTest.java
```diff
@@ -43,13 +48,15 @@ public void lifecycleMethodsShouldBeInvoked() {
         assertThat( context.getInvokedMethods() )
             .contains(
```
thunderhook, line 49: @filiphr 
At first I tried to validate the fix by using `containsExactly(...)`. But it soon became too complex and hard to understand due to the nested nature of this test.

So I added an additional fixture test that covers this.

Do you think this should be in a separate `Issue3678Mapper` or is it ok to test it here?

thunderhook: I removed the fixture, added the before/afterMapping only with source for the sake of completeness here.

But i provided a separate test that explicitly checks if before and after mapping methods are not called more than once.

processor/src/test/java/org/mapstruct/ap/test/bugs/_3678/Issue3678Mapper.java
```diff
@@ -0,0 +1,102 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.mapstruct.AfterMapping;
+import org.mapstruct.BeforeMapping;
+import org.mapstruct.Mapper;
+import org.mapstruct.ReportingPolicy;
+
+@Mapper(unmappedSourcePolicy = ReportingPolicy.ERROR)
```
filiphr, line 16: This is not needed for our tests as it'll anyways fails if there is a compiler warning

processor/src/test/java/org/mapstruct/ap/test/builder/lifecycle/BuilderLifecycleCallbacksTest.java
filiphr: If we have a dedicated test do we even need changes in this particular test here?

processor/src/test/java/org/mapstruct/ap/test/bugs/_3678/Issue3678Test.java
```diff
@@ -0,0 +1,34 @@
...
+    @ProcessorTest
+    void ignoreMappingsWithoutSourceShouldBeInvertible() {
+
+        Issue3678Mapper mapper = Mappers.getMapper( Issue3678Mapper.class );
+        Issue3678Mapper.SimpleSource source = new Issue3678Mapper.SimpleSource( "name", "description" );
+        Issue3678Mapper.SimpleDestination simpleDestination = mapper.sourceToDestination( source );
+
+        assertThat( mapper.getInvocations() )
+            .containsExactly( "beforeMapping", "afterMapping" )
+            .doesNotHaveDuplicates();
```
filiphr, line 31: `doesNotHaveDuplicates` is obsolete because we are using `containsExactly`.

filiphr (changes requested):
I'm not sure the fix is correct.

How will this behave for something like:

```java
@Mapper
public interface CustomMapper {

    Target map(SourceA sourceA, SourceB sourceB);

    @BeforeMapping
    default void beforeWithA(SourceA sourceA) {
        // ...
    }


    @BeforeMapping
    default void beforeWithB(SourceB sourceB) {
        // ...
    }

}
```

filiphr:
Looking at the issue I feel like the fix should be happening on a different level, perhaps when getting the lifecycle methods. Most likely the selection criteria needs to be changed

processor/src/test/java/org/mapstruct/ap/test/bugs/_3678/Issue3678Test.java
thunderhook: Yes, I just wanted to express the intent of this test more clearly. We can of course remove this.

thunderhook:
> I'm not sure the fix is correct.
> 
> How will this behave for something like:
> 
> ```java
> @Mapper
> public interface CustomMapper {
> 
>     Target map(SourceA sourceA, SourceB sourceB);
> 
>     @BeforeMapping
>     default void beforeWithA(SourceA sourceA) {
>         // ...
>     }
> 
> 
>     @BeforeMapping
>     default void beforeWithB(SourceB sourceB) {
>         // ...
>     }
> 
> }
> ```

Good catch. I will have a look at that again.

processor/src/test/java/org/mapstruct/ap/test/bugs/_3678/Issue3678Test.java
filiphr: I think `containsExactly` expresses the intent clearly. It says I'm expecting exactly this and nothing more. If there are duplicates it would clearly fail

processor/src/test/java/org/mapstruct/ap/test/builder/lifecycle/BuilderLifecycleCallbacksTest.java
thunderhook: Not necessariliy. It was however easier to debug the internals of the method resolution due to multiple lifecycle callback methods. I will remove it after implementing the fix.

processor/src/main/java/org/mapstruct/ap/internal/model/MethodReference.java
```diff
@@ -387,9 +387,6 @@ public boolean equals(Object obj) {
         if ( this == obj ) {
             return true;
         }
-        if ( !super.equals( obj ) ) {
-            return false;
-        }
```
thunderhook, line 392: This evaluated against `Object#equals` and therefore was always false. This might be a relic from legacy inheritance structures or never worked as intended.

processor/src/main/java/org/mapstruct/ap/internal/model/MethodReference.java
```diff
@@ -400,6 +397,9 @@ public boolean equals(Object obj) {
         if ( !Objects.equals( providingParameter, other.providingParameter ) ) {
             return false;
         }
+        if ( !Objects.equals( name, other.name ) ) {
+            return false;
+        }
```
thunderhook, line 402: Not 100% sure if the name (and already been there: declaringMapper + providingParameter) is enough to differ between multiple method references. All unit tests are green.

thunderhook:
So, i tried to adapt the method retrieval process. It became confusing very quickly and hard to differ between callback lifecycle methods that belong to the builder object and the finalized object.

The initial idea to remove the already invoked methods came from here:
* https://github.com/mapstruct/mapstruct/pull/3267#discussion_r1199799848

So, I had an idea how to improve this. Why not remove the methods via `removeAll(...)`? I had to generate equals/hashCode but now this looks more clean. There was a bug in the equals method anyway, I've fixed that (see the other comment)

@filiphr WDYT? Is this good enough?

Problems that arised during the attempt rewriting the selectors and criteria:
* Should there be a different `SelectionCriteria.Type.LIFECYCLE_CALLBACK_WHAT_SO_EVER`?
* The chaining from `BeanMappingMethod > LifecycleMethodResolver > SelectionContext` was somewhat cumbersome
* Should it be a blacklist like `collect everything except the methods from Set<...>`
    * If so, what model should be used for that blacklist? `MethodReference` or `SelectedMethod`?

processor/src/test/java/org/mapstruct/ap/test/bugs/_3678/Issue3678Mapper.java
```diff
@@ -0,0 +1,124 @@
...
+import org.mapstruct.BeforeMapping;
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+
+@Mapper
+public abstract class Issue3678Mapper {
+
+    @Mapping( target = "name", source = "sourceA.name")
+    @Mapping( target = "description", source = "sourceB.description")
+    abstract Target mapTwoSources(SourceA sourceA, SourceB sourceB);
```
thunderhook, line 21: Added a specific test with two sources to be sure.

filiphr:
Thanks @thunderhook. I'll need some time to think about this. I'm not too fond of the removals. I would prefer if we can select the right types when we need them. The `SelectionContext` already has `mappingTargetType` and `returnType`, so if we can somehow support multiple target types here we would be able to collect all in one go

thunderhook:
> Thanks @thunderhook. I'll need some time to think about this. I'm not too fond of the removals. I would prefer if we can select the right types when we need them. The `SelectionContext` already has `mappingTargetType` and `returnType`, so if we can somehow support multiple target types here we would be able to collect all in one go

Don't spend time on this yet, I think I get what you mean. I'll revisit this and hopefully come up with the solution you suggested. If there are any problems we can further discuss this here. 👍

thunderhook:
@filiphr Short heads up: Unfortunately, I won't be able to do this any time soon. What do you think about merging the fix with this variant and creating a refactoring issue? Or is it too dangerous for you that the PR could have side effects?

filiphr:
That's fine @thunderhook. I'll have a go at it. I'm not sure what kind of side effects this might have and I would rather spend the time on finding the right solution then rush into something

filiphr:
@thunderhook I found a different solution for this. Have a look at #3716. I'll close this PR in favour of that one

PR #3689, opened 2024-08-31
Fix typo in readme Maven plugin config

hduelme:
Currently the `target` tag for the `maven-compiler-plugin` is not closed correctly.


filiphr:
Thanks @hduelme. My bad when I am was upgrading it 😀

PR #3696, opened 2024-09-04
Use retainAll() for Set and Map instead of clear()

maxifridge:
Using retainAll() for Set and Map targets will minimize the amount of add/remove operations required for these types.
This limits side effects of removing preemptively all elements through clear() to non-Set Collections and Arrays.

A good example of why it is important (and the one that made me work on this issue) : when mapping to a Set property of a Hibernate-managed entity, calling clear() then addAll() effectively triggers requests for all deletions then all additions ; though because of the order of execution in Hibernate, only the newer elements are inserted, the others are deleted.


maxifridge:
This is still only a draft, it requires some cleaning/renaming, and certainly updating tests.

PR #3730, opened 2024-09-27
#3729 support for using inner class Builder without using static factory method

### LINKED ISSUE #3729: Detect Builder without a factory method
### Discussed in https://github.com/mapstruct/mapstruct/discussions/3724

<div type='discussions-op-text'>

<sup>Originally posted by **rmschots** September 25, 2024</sup>
Currently the `DefaultBuilderProvider` only detects builders if there is a static builder factory method available.
As I'm trying to use Wire (https://github.com/square/wire), and it only generates the inner builder class, I'd like to have mapstruct detect and use that builder class out of the box.

I brought us this issue at the Wire repository (https://github.com/square/wire/issues/3073), but the introduction of a static factory method for the builder got rejected.

I think it's either way more appropriate to add support in mapstruct out of the box.
I created an SPI that adds this support, but I prefer to have this baked into mapstruct.
I could open a PR if needed.

an example of target that cannot be used for mapping:
```
public class SomeTarget {
    private final String someField;

    private SomeTarget(String someField) {
        this.someField = someField;
    }

    public static class SomeTargetBuilder {
        private String someField;

        public SomeTargetBuilder someField(String someField) {
            this.someField = someField;
            return this;
        }

        public SomeTarget build() {
            return new SomeTarget(this.someField);
        }
    }
}
```
Trying to generate a mapper would currently produce the following error: `SomeTarget does not have an accessible constructor.`</div>

rmschots:
Fixes #3729

### DIFF
--- a/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-3-defining-a-mapper.asciidoc
@@ -421,8 +421,10 @@ If a Builder exists for a certain type, then that builder will be used for the m
 
 The default implementation of the `BuilderProvider` assumes the following:
 
-* The type has a parameterless public static builder creation method that returns a builder.
+* The type has either
+** a parameterless public static builder creation method that returns a builder.
 So for example `Person` has a public static method that returns `PersonBuilder`.
+** a public static inner class with the name having the suffix "Builder", and a public no-args constructor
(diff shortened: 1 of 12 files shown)


filiphr:
It took me a while, but I managed to merge this. Thanks @rmschots.

I merged it manually in https://github.com/mapstruct/mapstruct/commit/fce73aee6ab655f3896dee440fa890a1971a281f

PR #3733, opened 2024-10-01
ignoreUnmappedSourceProperties does not propagatevia  @InheritConfiguration

RichMacDonald:
Refer to https://github.com/mapstruct/mapstruct/discussions/3212#discussioncomment-8584497.

@BeanMapping(ignoreUnmappedSourceProperties = ...) does not propagate via @InheritConfiguration. So if a domain superclass object is mapped using ignoreUnmappedSourceProperties, any subclass mappings that use @InheritConfiguration have to redeclare the ignoreUnmappedSourceProperties.

Is this something that should always be true, or should it be an optional value, e.g.,  @InheritConfiguration(inheritIgnoreUnmappedSourceProperties=true)?

I am afraid I lack the understanding of the mapstruct implementation code to make the necessary code. However, I do have a failing test case that illustrates the concern.


filiphr:
Thanks for the tests @RichMacDonald. We'll look into it

PR #3736, opened 2024-10-03
resolve a mapper's constructor dependencies when creating via `Mappers.getMapper`

AlexElin:
At the moment when a mapper depends on another mappers which are passed via constructor, the call `Mappers.getMapper(...)` fails with an exception "java.lang.NoSuchMethodException: MapperName.<init>()"

This pull request fixes the issue by resolving such dependencies

### DIFF
--- a/core/src/main/java/org/mapstruct/factory/Mappers.java
+++ b/core/src/main/java/org/mapstruct/factory/Mappers.java
@@ -8,6 +8,7 @@ package org.mapstruct.factory;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.ServiceLoader;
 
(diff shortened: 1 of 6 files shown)


filiphr:
This is a bit tricky @AlexElin. What if one of the parameters for the constructor is not a mapper? What if there is a cyclic dependency?

I would rather leave the `Mappers` factory simple as is and if you want something more complex you can always use your own custom component model

filiphr:
@RafaelRfs that's not a bug in IntelliJ, it's rather a wrong usage of the `Mappers#getMapper` factory.

As I wrote in my earlier comment, I am not a fan of making `Mappers#getMapper` more complex and do dependency injection. The `Mappers` factory always looks for a class named `<MapperName>Impl`. However, the constructor parameters do not need to be mappers, they can be other classes.

You can always just use `new ProjectsMapperImpl(...)` in your tests.

janmaterne:
When using Spring, why calling the factory and not letting Spring inject the mapper?
```

@ExtendWith(SpringExtension.class)
class ProjectsMapperTest {

    @Autowired
    private ProjectsMapper projectsMapper;

    // no call to Mappers.getMapper()
```

Issue #3707, opened 2024-09-10
Mapstruct cannot find methods in abstract builder classes


dobrikov:
### Expected behavior

# abstract-builder-pattern-mapstruct

This Java project shows a use case of the abstract builder pattern adapted from the book "Effective Java" by Joshua Bloch.  
It aims to show missing functionality/bug of the mapstruct framework version 1.6.0. The use case:

We have an abstract class 'AbstractDto' with an abstract builder pattern for building the objects.

```Java
public abstract class AbstractDto {
  private final String name;
  private final Integer id;

  public abstract static class Builder<T> {
    private String name;
    private Integer id;

    protected abstract T self();

    public T name(String name) {
      this.name = name;
      return self();
    }

    public T id(Integer id) {
      this.id = id;
      return self();
    }

    public abstract AbstractDto build();
  }

  protected AbstractDto(Builder<?> builder) {
    this.name = builder.name;
    this.id = builder.id;
  }

  public String getName() {
    return name;
  }

  public int getId() {
    return id;
  }
}
```

Then we have two concrete implementations 'StringDto' and 'BigIntegerDto' of the abstract class 'AbstractDto':

```Java
public class StringDto extends AbstractDto {
  private final String value;
  public static class Builder extends AbstractDto.Builder<Builder> {

    private String value;

    @Override
    protected Builder self() {
      return this;
    }

    public Builder value(String value) {
      this.value = value;
      return self();
    }

    public Builder name(String value) {
      return super.name(value);
    }

    public Builder id(Integer value) {
      return super.id(value);
    }

    @Override
    public StringDto build() {
      return new StringDto(this);
    }
  }

  private StringDto(Builder builder) {
    super(builder);
    value = builder.value;
  }

  public String getValue() {
    return value;
  }

  public static Builder builder() {
    return new Builder();
  }
}
```

and 

```Java
public class BigIntegerDto extends AbstractDto {
  private final BigInteger value;

  public static class Builder extends AbstractDto.Builder<Builder> {

    private BigInteger value;

    @Override
    protected Builder self() {
      return this;
    }

    public Builder value(BigInteger value) {
      this.value = value;
      return self();
    }

    @Override
    public BigIntegerDto build() {
      return new BigIntegerDto(this);
    }
  }

  private BigIntegerDto(Builder builder) {
    super(builder);
    value = builder.value;
  }

  public BigInteger getValue() {
    return value;
  }

  public static Builder builder() {
    return new Builder();
  }
}
[...]

PR #3748, opened 2024-10-19
#3707 Fix abstract generic builder mappings

Obolrom:
Fixes bug #3707


filiphr (changes requested):
Thanks for the PR @Obolrom. However, I do not like the solution. The fact that the method returns a generic type var type does not mean that it should be treated as a potential fluent setter. 

In this phase we should be able to determine the generic return type to an actual type.

PR #3749, opened 2024-10-19
#3747 Fix empty if-block when NullValueMappingStrategy is RETURN_DEFAULT and target is immutable

### LINKED ISSUE #3747: Redundant if condition in Java record mapping with RETURN_DEFAULT strategy
### Expected behavior

The generated code should avoid unnecessary empty `if ` blocks. It should generate something like these:

```java
    @Override
    public ApiCategoryRecord mapRecord(Category category) {
        Long id = null;
        String title = null;
        if ( category != null ) {
            id = category.getId();
            title = category.getTitle();
        }

        ApiCategoryRecord apiCategoryRecord = new ApiCategoryRecord( id, title );

        return apiCategoryRecord;
    }
```

OR

```java
    @Override
    public ApiCategoryRecord mapRecord(Category category) {
        if ( category != null ) {
            return new ApiCategoryRecord( category.getId(), category.getTitle() );
        }
        else {
            return new ApiCategoryRecord( null, null );
        }
    }
```

### Actual behavior

MapStruct generates an empty and unnecessary `if` block after constructing the record:

```java
    @Override
    public ApiCategoryRecord mapRecord(Category category) {
        Long id = null;
        String title = null;
        if ( category != null ) {
            id = category.getId();
            title = category.getTitle();
        }
        ApiCategoryRecord apiCategoryRecord = new ApiCategoryRecord( id, title );
        if ( category != null ) {
        }
        return apiCategoryRecord;
    }
```

Literally this is not a bug but a minor inefficiency in the generated code.
It does not affect functionality, removing the redundant if condition would improve code clarity and doesn’t suggest anything went wrong during generation.

Use the following setup

```java
public class Category {
    private Long id;
    private String title;
    public Long getId() { return id; }
    public String getTitle() { return title; }
}

public record ApiCategoryRecord(Long id, String title) {}

@Mapper(nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT)
public interface ICategoryMapper {
    ApiCategoryRecord mapRecord(Category category);
}
```

### MapStruct Version

1.6.2

Hypnagokali:
Fixes #3747

**Problem:**
The BeanMappingMethod template always creates the if-block for the property mappings, even if the target only has constructor mappings. 

The unit tests are green but when I run "mvnw install" I get an exception from the maven shade plugin. The exception disappears when I set the version to 3.2.1

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -2048,6 +2048,10 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         return !constructorMappingsByParameter.isEmpty() || !constructorConstantMappings.isEmpty();
     }
 
+    public boolean hasPropertyMappingsByParameter() {
+        return !mappingsByParameter.isEmpty();
+    }
+
     public MethodReference getFinalizerMethod() {
         return finalizerMethod;
     }
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
@@ -99,30 +99,32 @@
 
     	</#if>
     </#list>
-    <#if (sourceParameters?size > 1)>
-        <#list sourceParametersNeedingPresenceCheck as sourceParam>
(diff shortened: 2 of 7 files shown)


filiphr:
Sorry I missed this @Hypnagokali and I already did #3758, so closing this in favor of that one. I also prefer the solution in my PR as it is slightly simpler then this PR.

PR #3765, opened 2024-11-12
Clarified documentation on @Named for usage on private methods

mipo256:


thunderhook:
Hi @mipo256!

Thanks for the improvement. I think it is good to have the information in the javadoc. Especially since I ran into the problem myself when I started using MapStruct.

We could also provide a compiler warning (or even an error) when a private method is annotated with `@Named` or maybe some other annotations like `@Condition`?
If we use a compile error, we could save ourselves the trouble to write this in the Javadoc for all annotations (for the sake of completeness).

@filiphr WDYT?

mipo256:
Hey @thunderhook !
I think it makes sense to add a compiler warming at the very least, I agree

filiphr:
Hey @mipo256 and @thunderhook. The javadoc here does not make much sense to me. 

If we are doing something then a compiler warning (or maybe even an error) is the way to go.

@mipo256 if you are interested in working on something like this then please open an issue and let us know if we can help somehow.

PR #3766, opened 2024-11-12
Fix typos in method and variable names.

hsjni0110:


thunderhook:
Hi @hsjni0110! Thanks!

PR #3784, opened 2024-12-04
Align README with v1.6.3 release

dlehammer:
Consistent with the [reference documentation](https://mapstruct.org/documentation/reference-guide/) "latest stable ..".

Screenshot from 2024-12-04 14-16-43


thunderhook:
Good catch! Thanks @dlehammer!

PR #3788, opened 2024-12-08
add ignored annotation issues 1958

xumk:
add new annotation for ignored more than one property

### DIFF
--- a/core/src/main/java/org/mapstruct/Ignored.java
+++ b/core/src/main/java/org/mapstruct/Ignored.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Configures the ignored of one bean attribute.
+ *
+ * <p>
+ * The name all attributes of for ignored is to be specified via {@link #value()}.
+ * </p>
+ *
+ * <p>
+ * <strong>Example 1:</strong> Implicitly mapping fields with the same name:
+ * </p>
+ *
+ * <pre><code class='java'>
+ * // We need ignored Human.name and Human.lastName
+ * // we can use &#64;Ignored with parameters  "name", "lastName" {@link #value()}
+ * &#64;Mapper
+ * public interface HumanMapper {
+ *    &#64;Ignored( { "name", "lastName" } )
+ *    HumanDto toHumanDto(Human human)
+ * }
+ * </code></pre>
+ * <pre><code class='java'>
+ * // generates:
+ * &#64;Override
+ * public HumanDto toHumanDto(Human human) {
+ *    humanDto.setFullName( human.getFullName() );
+ *    // ...
+ * }
+ * </code></pre>
+ *
+ * @author Ivashin Aleksey
+ * @see Javadoc
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})
+public @interface Ignored {
+
+    /**
+     * Whether the properties specified via {@link #value()} should be ignored by the generated mapping method.
+     * This can be useful when certain attributes should not be propagated from source to target or when properties in
+     * the target object are populated using a decorator and thus would be reported as unmapped target property by
+     * default.
+     *
+     * @return The target names of the configured property for ignored
+     */
+    String[] value();
+
+}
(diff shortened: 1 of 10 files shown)


diversit:
Would be nice if this would also support ignoring properties based on a prefix, like mentioned in [this comment](https://github.com/mapstruct/mapstruct/issues/1958#issuecomment-2054055873) related to generated 'withXX' methods.

xumk:
I'll try to see it on the weekend.

xumk:
@diversit  i'm add prefix for Ignored annotation for this case

core/src/main/java/org/mapstruct/Ignoreds.java
```diff
@@ -0,0 +1,54 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Configures the ignoreds of several bean attributes.
+ * <p>
+ * <strong>TIP: When using Java 8 or later, you can omit the @Ingoreds
+ * wrapper annotation and directly specify several @Ingoreds annotations on one method.</strong>
```
diversit, line 17: Typo. Specify several ‘@Ignored’ annotations. No ‘s’.

xumk: done

diversit:
Awesome!

filiphr:
>Would be nice if this would also support ignoring properties based on a prefix, like mentioned in https://github.com/mapstruct/mapstruct/issues/1958#issuecomment-2054055873 related to generated 'withXX' methods.

@diversit we do not want to support pattern ignores. We've said this few times already.

@xumk your idea with the prefix is OK I think, it basically transforms `customer.firstName`, `customer.lastName` to something more simpler. Unless, I am not seeing it properly

xumk:
@filiphr 
I assumed exactly this use
@Ignored( prefix = "animal", targets = { "publicAge", "size", "publicColor", "age", "color" } )
just add a dot prefix to the properties. It is more convenient than writing targets = { "animal.publicAge", "animal.size", "animal.publicColor"...etc
but nothing prevents you from writing it down according to the standard

core/src/main/java/org/mapstruct/Ignored.java
```diff
@@ -0,0 +1,71 @@
...
+ * <p>
+ * <strong>Example 1:</strong> Implicitly mapping fields with the same name:
+ * </p>
+ *
+ * <pre><code class='java'>
+ * // We need ignored Human.name and Human.lastName
+ * // we can use &#64;Ignored with parameters  "name", "lastName" {@link #targets()}
+ * &#64;Mapper
+ * public interface HumanMapper {
+ *    &#64;Ignored( { "name", "lastName" } )
```
filiphr, line 30: I just noticed this. This is not value it needs to be

```suggestion
 *    &#64;Ignored( targets = { "name", "lastName" } )
```

core/src/main/java/org/mapstruct/Ignored.java
```diff
@@ -0,0 +1,71 @@
...
+ * // generates:
+ * &#64;Override
+ * public HumanDto toHumanDto(Human human) {
+ *    humanDto.setFullName( human.getFullName() );
+ *    // ...
+ * }
+ * </code></pre>
+ *
+ * @author Ivashin Aleksey
+ * @see Javadoc
```
filiphr, line 44: This is not needed here, right?

processor/src/test/java/org/mapstruct/ap/test/ignored/ErroneousTargetHasNoWriteAccessorMapper.java
```diff
@@ -0,0 +1,25 @@
...
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.factory.Mappers;
+
+/**
+ *
+ * @author Sjaak Derksen
+ */
+@Mapper
+public interface ErroneousTargetHasNoWriteAccessorMapper {
```
filiphr, line 17: Why do we need this for this functionality?

filiphr (changes requested):
Thanks @xumk, sorry that it take some time, but I've been a bit busy.

I've left some comments.

Something additional. I'd like to have some tests for the following options:

```java
@Mapping(target = "name", ignore = true)
@Ignored(targets = "name")
```

and

```java
@Mapping(target = "name", source = "firstName")
@Ignored(targets = "name")
```

What is happening in those scenarios

xumk:
@filiphr 
```
@Mapping(target = "name", ignore = true)
@Ignored(targets = "name")
```
it test ok. property is ignored.

```
@Mapping(target = "name", source = "firstName")
@Ignored(targets =  "name")
```
it not ignored property. name have value firstName

i added two test

filiphr:
Thanks for checking that @xumk. My main point was the fact that a combination of `@Mapping` and `@Ignored` for the same property should not be allowed, we should have an error if such a thing happens. The same as if we are using the same target in `@Mapping`

xumk:
@filiphr ok, I will try to do it soon.

xumk:
@filiphr I added a check

PR #3803, opened 2025-01-19
Java EA GitHub Actions improvements

filiphr:
With this PR we are upgrading to Spring 6 and thus Java 17 for building MapStruct. However, we are still Java 8 compatible.


thunderhook:
Nice, LGTM!

PR #3815, opened 2025-02-11
chore(docs): Improved wording about @Condition usage

fml2:
The repeated "it is" and the comma makes the meaning slightly wrong IMO.


filiphr:
Thanks @fml2, sorry that it took a while

PR #3847, opened 2025-05-04
Issue #765: Mark String to number lossy conversions

SamilCan:
Fix Issue #765


filiphr:
Thanks for the PR @SamilCan. The PR looks OK. I have only one question regarding the hierarchy number. With this we are basically now treating `String` as a `BigDecimal`, right?

I'll need to think a bit about the potential impact of this for existing mappings. It's not as simple as the other ones I think.

SamilCan:
Hi @filiphr ,

I especially used the same hierarchy number for `String` because I know most of the time `BigDecimal` is the preferred type for the Money field and is represented as String in JSON serialization. So, I think we can see lots of `String <> Bigdecimal` conversions, which will be lossless in both directions.

Issue #3821, opened 2025-02-28
Add support for custom exception for subclass exhaustive strategy for @SubclassMapping mapping


s-jepsen:
### Use case

Introduce a new feature that can customize the IllegalArgumentException exception thrown, when a unexpected SubClassMapping is encountered.

Like the EnumMapping#unexpectedValueMappingException used to customize [EnumMapping exception](https://mapstruct.org/documentation/stable/api/org/mapstruct/EnumMapping.html#unexpectedValueMappingException())

Mapping example:
```java
@SubclassMapping(target = GroupSpecificationEntity.class, source = GroupSpecification.class)
@SubclassMapping(target = SiteSpecificationEntity.class, source = SiteSpecification.class)
LocationSpecificationEntity map(LocationSpecification source);
```

Generated code:
```java
@Override
    public Target map(Source source) {
        if ( source == null ) {
            return null;
        }

        if (source instanceof GroupSpecification) {
            return groupSpecificationToGroupSpecificationEntity( (GroupSpecification) source );
        }
        else if (source instanceof SiteSpecification) {
            return siteSpecificationToSiteSpecificationEntity( (SiteSpecification) source );
        }
        else {
            throw new IllegalArgumentException("Not all subclasses are supported for this mapping. Missing for " + source.getClass());
        }
    }
```

### MapStruct Version

Mapstruct 1.6.3

tangyang9464:
> ### Use case
> Introduce a new feature that can customize the IllegalArgumentException exception thrown, when a unexpected SubClassMapping is encountered.
> 
> Like the EnumMapping#unexpectedValueMappingException used to customize [EnumMapping exception](https://mapstruct.org/documentation/stable/api/org/mapstruct/EnumMapping.html#unexpectedValueMappingException())
> 

After analyzing the current implementation, I've identified a potential challenge with implementing custom exceptions for `@SubclassMapping` similar to `@EnumMapping#unexpectedValueMappingException`.

`@SubclassMapping` is designed to be used multiple times on the same method, this creates a conflict: if each `@SubclassMapping` could specify its own exception type, which one should be used in the generated code? The current implementation throws a single exception when an unexpected subclass is encountered, but it wouldn't be clear which exception type to use if multiple `@SubclassMapping` annotations specified different exceptions.

filiphr:
[...]

PR #3853, opened 2025-05-11
#3821 feat: add support for custom exception in subclass mapping

tangyang9464:
close [issue#3821]( https://github.com/mapstruct/mapstruct/issues/3821)
I implemented subclassExhaustiveException with reference to subclassExhaustiveStrategy, so this custom exception supports configuration on mapper, mapperConfig, and BeanMapping. 

plz review @filiphr

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -100,6 +100,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
     private final String finalizedResultName;
     private final List<LifecycleCallbackMethodReference> beforeMappingReferencesWithFinalizedReturnType;
     private final List<LifecycleCallbackMethodReference> afterMappingReferencesWithFinalizedReturnType;
+    private final Type subclassExhaustiveException;
 
     private final MappingReferences mappingReferences;
 
@@ -138,7 +139,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             return this;
         }
 
-        public Builder returnTypeBuilder( BuilderType returnTypeBuilder ) {
+        public Builder returnTypeBuilder(BuilderType returnTypeBuilder) {
             this.returnTypeBuilder = returnTypeBuilder;
             return this;
         }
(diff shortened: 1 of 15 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -378,6 +379,13 @@ else if ( !method.isUpdateMethod() ) {
 
             }
 
+            Type subclassExhaustiveExceptionType = ctx.getTypeFactory().getType( IllegalArgumentException.class );
+            TypeMirror subclassExhaustiveException =
+                beanMapping != null ? beanMapping.getSubclassExhaustiveException() : null;
+            if ( subclassExhaustiveException != null ) {
+                subclassExhaustiveExceptionType = ctx.getTypeFactory().getType( subclassExhaustiveException );
+            }
```
filiphr, line 387: This can be simplified:

```suggestion
            TypeMirror subclassExhaustiveException = method.getOptions()
                .getBeanMapping()
                .getSubclassExhaustiveException();
            Type subclassExhaustiveExceptionType = ctx.getTypeFactory().getType( subclassExhaustiveException );
```

filiphr (commented):
This looks as expected @tangyang9464, I've left one comment and can you please make sure that you remove the whitespace changes from the `BeanMappingMethod`, you don't need to reformat the entire file, only what you have changed.

tangyang9464:
For the first time, I applied the entire file formatting. I found that there were many changes. I gave up and then the CICD failed. So I executed the formatting again. I thought that all the code in the project was in line with the specifications. But it seems that some historical code was not like that.

 I will manually remove those changes that are not mine.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
tangyang9464: It can indeed be simplified here, but does `beanMapping `need to be judged `null`? Because I saw that other places in the code that use `beanMapping ` are judged null.

filiphr: I think you were looking in some of the lower level places where we are accessing the annotation. I see that on line 171 it is checked for `null`. However, what I proposed is already used on line 335, so I am not worried about that.

tangyang9464:
@filiphr plz review again

filiphr:
Thanks @tangyang9464

PR #3854, opened 2025-05-13
fix: default value handling for optional

### LINKED ISSUE #3852: Initialize `Optional` with `Optional.empty` instead of `null`
As written in https://github.com/mapstruct/mapstruct/issues/674#issuecomment-2769667346.

It would be better if we do not create `null` optional, but rather `Optional.empty` when needed. This could be done in `Type#getNull`.

MelleD:
This pull request introduces support for handling not nullable `Optional` types in the MapStruct framework and includes related test cases. The most significant changes involve adding a new method to check for `Optional` types, modifying the handling of null values, and introducing test classes to validate the new functionality. Now the default is `Optional.empty()` instead of `null`

Fix #3852

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
@@ -231,7 +231,7 @@ public class Type extends ModelElement implements Comparable<Type> {
 
     /**
      * Returns a String that could be used in generated code to reference to this {@link Type}.<br>
-     *  <p>
+     * <p>
      * The first time a name is referred-to it will be marked as to be imported. For instance
      * {@code LocalDateTime} can be one of {@code java.time.LocalDateTime} and {@code org.joda.LocalDateTime})
      * <p>
@@ -259,7 +259,7 @@ public class Type extends ModelElement implements Comparable<Type> {
             return name;
         }
 
-        if ( isTopLevelTypeToBeImported() && nameWithTopLevelTypeName != null) {
+        if ( isTopLevelTypeToBeImported() && nameWithTopLevelTypeName != null ) {
             return nameWithTopLevelTypeName;
         }
 
@@ -367,8 +367,12 @@ public class Type extends ModelElement implements Comparable<Type> {
         return componentType != null;
     }
 
+    public boolean isOptionalType() {
+        return !isPrimitive() || "java.util.Optional".equals( getFullyQualifiedName() );
+    }
+
     public boolean isTypeVar() {
-        return (typeMirror.getKind() == TypeKind.TYPEVAR);
+        return ( typeMirror.getKind() == TypeKind.TYPEVAR );
     }
 
     public boolean isIntersection() {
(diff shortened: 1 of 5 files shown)


filiphr (commented):
The solution looks OK. However, there are way too many whitespace changes in the `Type.java`, can you please revert all those changes and only commit changes that are related to the issue. Makes it easier to review and merge.

MelleD:
> The solution looks OK. However, there are way too many whitespace changes in the `Type.java`, can you please revert all those changes and only commit changes that are related to the issue. Makes it easier to review and merge.

But looks more that the file is broken and not formatted with you own code style.

I used this one and saved the file:
https://github.com/mapstruct/mapstruct/blob/main/etc/mapstruct.xml

filiphr:
>But looks more that the file is broken and not formatted with you own code style.

I need to update the formatter, I've noticed that there are some changes in newer IntelliJ versions, also it is good practice to only format new code, at least that's what I try to do. Thanks for removing the whitespaces

MelleD:
> > But looks more that the file is broken and not formatted with you own code style.
> 
> I need to update the formatter, I've noticed that there are some changes in newer IntelliJ versions, also it is good practice to only format new code, at least that's what I try to do. Thanks for removing the whitespaces

Mhm ok. For me it’s good practice in IntelliJ to use proper plugins like save actions and then have a clean format for the whole file and not only partly

MelleD:
@filiphr done

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -367,6 +368,10 @@ public boolean isArrayType() {
         return componentType != null;
     }
 
+    public boolean isOptionalType() {
+        return Optional.class.getName().equals( getFullyQualifiedName() );
+    }
```
filiphr, line 373: This is only used in the `getNull`, so it shouldn't be `public`. Additionally, we should also support `OptionalInt`, `OptionalDouble` and `OptionalLong`.

processor/src/test/java/org/mapstruct/ap/test/bugs/_3852/Issue3852Mapper.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._3852;
```
filiphr, line 6: This is not a bug. It's an enhancement. Perhaps you can just add to the tests in `org.mapstruct.ap.test.nestedproperties.simple` and / or put them in some other package

filiphr (changes requested):
thanks for the adjustments @MelleD. I've left few more comments, let me know if you disagree with some of them.

If you don't have time, I can wrap it up as well

processor/src/test/java/org/mapstruct/ap/test/bugs/_3852/Issue3852Mapper.java
MelleD: done

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
MelleD: done

filiphr (approved):

Issue #3849, opened 2025-05-07
Overloaded AfterMapping is called twice


gbrehmer:
### Expected behavior

If you have a class with hierarchy and multiple AfterMapping methods with the same name but for different level of the hierarchy, mapstruct should generate only one method call, because the overloaded method can not be reached

### Actual behavior

Mapstruct handles both methods individual and did not recognize that the overloaded method can not be reached. It results in a duplicated call to the same method

### Steps to reproduce the problem

```
public class Entity {
  long id;
  String name;
}

public interface ParentDTOInterface {
  long getId();
}

@Data
public class ParentDTO implements ParentDTOInterface {
  private long id;
}

@Data
public class ChildDTO extends ParentDTO {
  private String name;
}


public interface Mapper {

  ParentDTO mapParent(Entity entity);
  ChildDTO mapChild(Entity entity);
  
  @AfterMapping
  default void after(@MappingTarget ParentDTOInterface dto, Entity entity) {
  }
  
  @AfterMapping
  default void after(@MappingTarget ChildDTO childDTO, Entity entity) {
  }
}
```

generated child related code:

```
@Override
ChildDTO map(Entity entity) {
  ChildDTO childDTO = new ChildDTO();

  childDTO.setId(entity.getId());
  childDTO.setName(entity.getName());

  after(childDTO, entity);
  after(childDTO, entity); <---- this is the bug :)
}
```

generated parent related code:

```
@Override
ParentDTO map(Entity entity) {
  ParentDTO parentDTO = new ParentDTO();

  parentDTO.setId(entity.getId());

  after(parentDTO, entity);
}
```
### MapStruct Version

1.6.3

tangyang9464:
@gbrehmer I'm a little curious, why not merge one method?

these two methods are ambiguous, not a valid overload
```
@AfterMapping
void after(@MappingTarget ParentInterface childDTO, Entity entity) {
}

@AfterMapping
void after(@MappingTarget ChildDTO childDTO, Entity entity) {
}
```

gbrehmer:
first method is called for `ParentDTO` objects, second one for `ChildDTO`

Zegveld:
> Mapstruct handles both methods individual and did not recognize that the overloaded method can not be reached. It results in a duplicated call to the same method

The method can be reached, Mapstruct just does not call it correctly to reach it.

When I look at this from bug perspective I would say we need to change the second call, so that we can reach it.
[...]

PR #3856, opened 2025-05-13
fix: resolve duplicate invocation of overloaded lifecycle methods with inheritance

tangyang9464:
fix: [3849](https://github.com/mapstruct/mapstruct/issues/3849)

Newly added `LifecycleOverloadDeduplicateSelector`, which uses multiple features such as **class name** and **method name** as keys to group methods.

And then selects the best method within the group by comparing the inheritance distance of parameters, 
to achieve deduplication between similar overloaded methods. 

The parameters involved in the inheritance distance calculation are not limited to `@MappingTarget`, but all input parameters of a group of candidate methods are traversed and calculated in sequence until a parameter is found with the shortest distance, then the traversal is terminated. (Avoid the situation where the **@MappingTarget** parameter type is the same but other parameters are different).

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/LifecycleOverloadDeduplicateSelector.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/LifecycleOverloadDeduplicateSelector.java
@@ -6,6 +6,7 @@
 package org.mapstruct.ap.internal.model.source.selector;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.stream.Collectors;
@@ -32,9 +33,12 @@ import org.mapstruct.ap.internal.model.source.Method;
  * <p>
  * <b>Example (see Issue3849Test):</b>
  *
- * @AfterMapping default void afterMapping(Parent source, @MappingTarget ParentDto target) { ... }
- * @AfterMapping default void afterMapping(Parent source, @MappingTarget ChildDto target) { ... }
- * </pre>
+ * <pre>{@code
+ * @AfterMapping
+ * default void afterMapping(Parent source, @MappingTarget ParentDto target) { ... }
+ * @AfterMapping
+ * default void afterMapping(Parent source, @MappingTarget ChildDto target) { ... }
+ * }</pre>
  * When mapping a Child to a ChildDto,
  * only the method with ChildDto is selected, even though both methods match by signature
  * except for the target type's inheritance relationship.
@@ -43,19 +47,18 @@ public class LifecycleOverloadDeduplicateSelector implements MethodSelector {
     @Override
     public <T extends Method> List<SelectedMethod<T>> getMatchingMethods(List<SelectedMethod<T>> methods,
                                                                          SelectionContext context) {
-        if ( !context.getSelectionCriteria().isLifecycleCallbackRequired() ) {
+        if ( !context.getSelectionCriteria().isLifecycleCallbackRequired() || methods.size() <= 1 ) {
             return methods;
         }
-        List<List<SelectedMethod<T>>> methodSignatureGroups = new ArrayList<>(
+        Collection<List<SelectedMethod<T>>> methodSignatureGroups =
             methods.stream()
                 .collect( Collectors.groupingBy(
                     LifecycleOverloadDeduplicateSelector::buildSignatureKey,
                     LinkedHashMap::new,
                     Collectors.toList()
                 ) )
-                .values()
-        );
-        List<SelectedMethod<T>> deduplicatedMethods = new ArrayList<>();
+                .values();
+        List<SelectedMethod<T>> deduplicatedMethods = new ArrayList<>( methods.size() );
         for ( List<SelectedMethod<T>> signatureGroup : methodSignatureGroups ) {
             if ( signatureGroup.size() == 1 ) {
                 deduplicatedMethods.add( signatureGroup.get( 0 ) );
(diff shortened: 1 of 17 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/LifecycleOverloadDeduplicateSelector.java
```diff
@@ -0,0 +1,131 @@
...
+        key.add( method.getMethod().getDefiningType() );
+        key.add( method.getMethod().getName() );
+        for ( ParameterBinding binding : method.getParameterBindings() ) {
+            key.add( binding.getType() );
+            key.add( binding.getVariableName() );
+            key.add( binding.isMappingTarget() );
+            key.add( binding.isMappingContext() );
+            key.add( binding.isTargetType() );
+            key.add( binding.isSourcePropertyName() );
+            key.add( binding.isTargetPropertyName() );
```
filiphr, line 100: Why should these elements be part of the signature?

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/LifecycleOverloadDeduplicateSelector.java
```diff
@@ -0,0 +1,131 @@
...
+            return methods;
+        }
+        List<List<SelectedMethod<T>>> methodSignatureGroups = new ArrayList<>(
+            methods.stream()
+                .collect( Collectors.groupingBy(
+                    LifecycleOverloadDeduplicateSelector::buildSignatureKey,
+                    LinkedHashMap::new,
+                    Collectors.toList()
+                ) )
+                .values()
```
filiphr, line 56: Let's try to avoid creating a lot of objects. There is no need to create a new list for the values. You can just get the values (`Collection<List<SelectedMethod<T>>>`) and iterate over it

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/LifecycleOverloadDeduplicateSelector.java
```diff
@@ -0,0 +1,131 @@
...
+        );
+        List<SelectedMethod<T>> deduplicatedMethods = new ArrayList<>();
```
filiphr, line 58: Let's initialize the list with a size that is the `methods.size()`

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/LifecycleOverloadDeduplicateSelector.java
```diff
@@ -0,0 +1,131 @@
...
+ * </pre>
+ * When mapping a Child to a ChildDto,
+ * only the method with ChildDto is selected, even though both methods match by signature
+ * except for the target type's inheritance relationship.
+ */
+public class LifecycleOverloadDeduplicateSelector implements MethodSelector {
+    @Override
+    public <T extends Method> List<SelectedMethod<T>> getMatchingMethods(List<SelectedMethod<T>> methods,
+                                                                         SelectionContext context) {
+        if ( !context.getSelectionCriteria().isLifecycleCallbackRequired() ) {
```
filiphr, line 46: Let's add `methods.size() <= 1` as an additional short circuit

filiphr (changes requested):
Thanks @tangyang9464, the solution looks good to me. I've left some comments, which are mainly for performance optimization, as this can get called a lot, especially if you have a lot of lifecycle methods.

Additionally, I think that it'll be good if we can add some tests with generics. I am not sure whether everything would fine then.

Also, can we perhaps add some compiler flag so that people can disable this particular selector in case it causes them performance problems and / or something isn't working properly, just in case. I think we can pass the `Options` to the `MethodsSelectors` and then check whether we should include this new one or not. What do you think about that?

filiphr:
btw, the build is failing due to the javadoc

```
Error:  /home/runner/work/mapstruct/mapstruct/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/LifecycleOverloadDeduplicateSelector.java:35: error: unknown tag: AfterMapping
Error:   * @AfterMapping default void afterMapping(Parent source, @MappingTarget ParentDto target) { ... }
Error:     ^
Error:  /home/runner/work/mapstruct/mapstruct/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/LifecycleOverloadDeduplicateSelector.java:36: error: unknown tag: AfterMapping
Error:   * @AfterMapping default void afterMapping(Parent source, @MappingTarget ChildDto target) { ... }
Error:     ^
Error:  /home/runner/work/mapstruct/mapstruct/processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/LifecycleOverloadDeduplicateSelector.java:37: error: unexpected end tag: </pre>
Error:   * </pre>
```

tangyang9464:
I'll modify based on the comments and fix CICD later

processor/src/main/java/org/mapstruct/ap/internal/model/source/selector/LifecycleOverloadDeduplicateSelector.java
tangyang9464: This is not necessary.  The type and variable name are sufficient to determine the parameter signature.

filiphr: Cool thanks, Shall we remove those then? And just so that I understand properly, when you say variable name, it is the name of something in the mapping method, not the name of a parameter in the lifecycle method, right?

tangyang9464: > Cool thanks, Shall we remove those then? And just so that I understand properly, when you say variable name, it is the name of something in the mapping method, not the name of a parameter in the lifecycle method, right?
Of course.

> when you say variable name, it is the name of something in the mapping method, not the name of a parameter in the lifecycle method, right?
Yes.

tangyang9464:
@filiphr plz review again.  I did the following:
1. Added compile args to disable the selectors and add related tests
2. Added generic tests
3. Optimized the code to improve performance based on your comments

tangyang9464:
@filiphr any viewpoint？

filiphr:
Thanks for the changes @tangyang9464. During the week it's a bit more difficult for me to check these stuff out. I'll review it soon and merge it

PR #3864, opened 2025-05-22
fix: UpdateWrapper.ftl doesn't pass targetPropertyName when executing source presence check

tangyang9464:
fix: https://github.com/mapstruct/mapstruct/issues/3809

I passed in variables such as **targetPropertyName** and optimized the template format.


tangyang9464:
> Thanks @tangyang9464. I pushed a small polish, just to reduce the amount of files.

It's good. Thanks. @filiphr

Issue #3711, opened 2024-09-12
Support generic `@Context`


ilia1243:
### Use case

Consider the following example [mapstruct-jpa-child-parent](https://github.com/mapstruct/mapstruct-examples/tree/main/mapstruct-jpa-child-parent) but with generic JpaContext:

```java
@Data
class ParentDto {
    ChildDto child;
}

@Data
class ChildDto {
    ParentDto parent;
}

@Data
class ParentEntity {
    ChildEntity child;
}

public interface ParentRelation<T> {
    void setChildToParent(T parent);
}

@Data
class ChildEntity implements ParentRelation<ParentEntity> {
    ParentEntity childToParent;
}

class JpaContext<T> {
    T parentEntity;

    @BeforeMapping
    void setEntity(@MappingTarget T parentEntity) {
        this.parentEntity = parentEntity;
    }

    @AfterMapping
    void establishRelation(@MappingTarget ParentRelation<T> childEntity) {
        childEntity.setChildToParent(parentEntity);
    }
}

interface BaseMapper<T, E> {
    default E toEntity(T s) {
        return toEntity(s, new JpaContext<>());
    }

    E toEntity(T s, @Context JpaContext<E> ctx);
}

@Mapper
interface SourceTargetMapper extends BaseMapper<ParentDto, ParentEntity> {
}
```

In this case MapStruct does not invoke @BeforeMapping and @AfterMapping methods in SourceTargetMapperImpl, although it is allowed by Java assignment rules.

### Generated Code

Desired generated code:

```java
@Generated(
    value = "org.mapstruct.ap.MappingProcessor"
)
class SourceTargetMapperImpl implements SourceTargetMapper {

    @Override
    public ParentEntity toEntity(ParentDto s, Ctx ctx) {
        if ( s == null ) {
            return null;
        }

        ParentEntity parentEntity = new ParentEntity();

        ctx.setEntity( parentEntity );

        parentEntity.setChild( childDtoToChildEntity( s.getChild(), ctx ) );

        return parentEntity;
    }

    protected ChildEntity childDtoToChildEntity(ChildDto childDto, Ctx ctx) {
        if ( childDto == null ) {
            return null;
        }

        ChildEntity childEntity = new ChildEntity();

        ctx.establishRelation( childEntity );

        return childEntity;
    }
}
```

### Possible workarounds

[...]

PR #3867, opened 2025-05-23
fix: methods contained in @Context classes don't support generics

tangyang9464:
fix: https://github.com/mapstruct/mapstruct/issues/3711

The bug is caused by not handling generics properly when generating context methods

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
@@ -110,7 +110,12 @@ public class MethodRetrievalProcessor implements ModelElementProcessor<Void, Lis
         }
 
         List<SourceMethod> prototypeMethods = retrievePrototypeMethods( mapperTypeElement, mapperOptions );
-        return retrieveMethods( mapperTypeElement, mapperTypeElement, mapperOptions, prototypeMethods );
+        return retrieveMethods(
+            typeFactory.getType( mapperTypeElement.asType() ),
+            mapperTypeElement,
+            mapperOptions,
+            prototypeMethods
+        );
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
```diff
@@ -215,12 +221,19 @@ private TypeElement asTypeElement(DeclaredType type) {
...
                                    List<SourceMethod> prototypeMethods) {
-
-        ExecutableType methodType = typeFactory.getMethodType( (DeclaredType) usedMapper.asType(), method );
+        DeclaredType methodDeclaredType;
+        if ( useContext != null ) {
+            methodDeclaredType = (DeclaredType) useContext.getTypeMirror();
+        }
+        else {
+            methodDeclaredType = (DeclaredType) usedMapper.asType();
+        }
```
filiphr, line 235: I am not too fond of this approach. Why don't we change the first method of the `retrieveMethods ` and instead of passing `TypeElement` for the `usedMapper` we can pass our `Type`.

filiphr (changes requested):
Thanks for the fix @tangyang9464. I've left a comment, I am not too fond of the fix. I would rather do it slightly differently.

processor/src/main/java/org/mapstruct/ap/internal/processor/MethodRetrievalProcessor.java
tangyang9464: I thought about doing this, but I'm not sure if multiple types such as `Type `and `TypeElement `can be converted safely. I will try again.

tangyang9464:
> Why don't we change the first method of the retrieveMethods  and instead of passing TypeElement for the usedMapper we can pass our Type.

@filiphr done. plz review again.

filiphr (approved):

Issue #3836, opened 2025-03-24
SubclassMapping generates an else branch for a sealed interface when the target is not abstract


mecseid:
### Expected behavior

Generates the method implementation without any error, and creates the required if/else or switch/case statements.

### Actual behavior

Drops an exception with unmapped target properties.
If it is set to WARN, the generated code contains an else branch with default mapping.
```java
    @Override
    public TestMapping.To map(TestMapping.From from) {
        if ( from == null ) {
            return null;
        }

        if (from instanceof TestMapping.FromA) {
            return map( (TestMapping.FromA) from );
        }
        else if (from instanceof TestMapping.FromB) {
            return map( (TestMapping.FromB) from );
        }
        else {

            int to = 0;

            TestMapping.To to1 = new TestMapping.To( to );

            return to1;
        }
    }
```

### Steps to reproduce the problem

Create an interface with more than one permitted class, then create a mapper with SubclassMapping, like:
```java

import org.mapstruct.Mapper;
import org.mapstruct.Mapping;
import org.mapstruct.ReportingPolicy;
import org.mapstruct.SubclassMapping;

@Mapper(unmappedTargetPolicy = ReportingPolicy.WARN)
public interface TestMapping {

  sealed interface From permits FromA, FromB {}

  record FromA(int a) implements From {}

  record FromB(int b) implements From {}

  record To(int to) {}

  @Mapping(source = "a", target = "to")
  To map(FromA from);

  @Mapping(source = "b", target = "to")
  To map(FromB fromB);

  @SubclassMapping(source = FromA.class, target = To.class)
  @SubclassMapping(source = FromB.class, target = To.class)
  To map(From from);

}
```

### MapStruct Version

1.6.3

filiphr:
I get your point. You are right, the `else` here is obsolete as it is not reachable. Perhaps we can look into trying to use an enhanced `switch` here.

kssumin:
Hi! I'd like to work on this issue.

From what I understand, the problem is that `SubclassMapping` generates an unnecessary else branch when dealing with sealed interfaces, even though all permitted subclasses are mapped. This else branch is unreachable because sealed interfaces have a finite set of implementations.

My approach would be:
1. Detect if the source type is a sealed interface
2. Check if all permitted subclasses have corresponding mappings
3. Skip generating the else branch when all subclasses are covered

I believe this would align with the enhanced switch suggestion mentioned above. I'll start by writing tests to reproduce the issue and then work on the fix.

[...]

PR #3885, opened 2025-06-14
mapstruct#3836 sealed interface create impl class with switch

mildw428:
I've fixed it for bug number 3836.

I added sourceArgumentType to know the parameter type of the subclass.

Generates as a switch or if-else syntax according to isSealed().

If you have any issues, please advise.

https://github.com/mapstruct/mapstruct/issues/3836

### DIFF
--- a/integrationtest/src/test/java/org/mapstruct/itest/tests/MavenIntegrationTest.java
+++ b/integrationtest/src/test/java/org/mapstruct/itest/tests/MavenIntegrationTest.java
@@ -222,4 +222,11 @@ public class MavenIntegrationTest {
     void faultyAstModifyingProcessor() {
     }
 
+    /**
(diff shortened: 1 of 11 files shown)


mildw428:
In version 17, the switch statement needs a default return, resulting in an error. If anyone has any branching ideas based on the version, please let me know.

filiphr:
I'll need some time to go through this one @mildw428. I am not sure that we can always use the enhanced `switch`. We can only do that if the type is correctly sealed otherwise we need a fallback.

mildw428:
> I'll need some time to go through this one @mildw428. I am not sure that we can always use the enhanced `switch`. We can only do that if the type is correctly sealed otherwise we need a fallback.

Thank you for reviewing my code.

Issue #3858, opened 2025-05-16
@InheritConfiguration unexpected behavior


nikolaizhnikolov:
### Expected behavior

Hello, we've had a problem in our project with unexpected mapstruct annotation behaviour, which lead to problems. We caught it early and updated our code and testing. However, initially we were led to believe things are working as expected. This is a sample of the code we had:

```java
@Mapper(config = DefaultMapperConfig.class,
    collectionMappingStrategy = CollectionMappingStrategy.ACCESSOR_ONLY
)
public abstract class GroupMapper { 

    @Mapping(target = "name", source = "name")
    @Mapping(target = "parentGroupName", source = "parentGroup.name")
    // other mappings...
    public abstract GroupDto toDto(Group group);

    @InheritConfiguration(source = "toDto")
    @Mapping(target = "name", source = "groupNameDto.name")
    public abstract GroupDto toDto(Group group, GroupNameDto groupNameDto);
}
```

When trying to use this code this is what we expected:
- We can reuse the `toDto` configuration, because the full code has a lot of custom mappings, that we don't want to repeat.
- We will add new methods using other dtos to do partial updates of our groups, as they are large and complex, and a single update is hard to maintain.
- The existing mappings will be reused from the `Group group` parameter, and anything that comes from the `GroupNameDto groupNameDto` can be accessed using the full qualifier (so nothing will clash)
- If this is not possible a compilation error will disallow this usage

### Actual behavior

What actually happened:
- The mappings were inherited in an unexpected way
  - Direct mappings e.g. from the Group to the Dto were fine
  - More complex mappings e.g. accessing the parent Group (which is a different object of the same type nested inside the Group) were completely lost. Even when we tried modifying the original `toDto` method with the full path e.g. `group.parentGroup.name` we couldn't get it to work.
- In the documentation we couldn't find a specific mention of the intended behavior. But testing showed that things will work fine as long as there are no additional parameters. So for our case it wasn't possible to partially update our group with different dtos while reusing the original dto method.
- Our testing looked fine, because we didn't test ALL the fields for the new methods. Once it hit our testing environment we saw the inconsistencies and went back to the mapstruct implementation and saw its being generated wrong.

Our recommendation is to disallow compilation in the cases where @InheritConfiguration is used with incompatible parameters. Let me know your thoughts.

### Steps to reproduce the problem

```java
@MapperConfig(
    componentModel = "spring",
[...]

PR #3894, opened 2025-06-26
#3858 Compilation error if no methods are found to @InheritConfiguration from

nikolaizhnikolov:
Solution to #3585


filiphr:
Thanks for the PR @nikolaizhnikolov. I see that you didn't write any tests for this. Can you please do that?

PR #3906, opened 2025-07-27
[#3837] Add warning/error for redundant ignoreUnmappedSourceProperties entries

### LINKED ISSUE #3837: Report properties in `ignoreUnmappedSourceProperties` that actually do get mapped
It would be nice if we could have mappings that are more resistant to changes in the target object. There's no way to flat-out exclude source properties from mapping, the closest approximation would be `unmappedSourcePolicy = WARN` in combination with `ignoreUnmappedSourceProperties`. However, it's not robust, as you won't get notified if a property in `ignoreUnmappedSourceProperties` suddenly starts getting mapped, because a matching field was added in the target object. It would make sense if there was a warning or error, possibly depending on `unmappedSourcePolicy`.

codeswithritesh:
Enhancement: warn when mapped properties are listed in ignoreUnmappedSourceProperties (#3837)

- Enhances @BeanMapping handling by reporting a warning when a property is both mapped and listed in ignoreUnmappedSourceProperties.
- Respects effective unmappedSourcePolicy from @BeanMapping, @Mapper, or @MapperConfig.
- Avoids reporting warning when policy is set to IGNORE.
- Adds new unit tests for redundant, valid, mixed, and policy-specific behavior.
Fixes #3837


gnuemacscoin:
Looks good, though the message grammar is a little off. I would rephrase it as `"Source %s is mapped despite being listed in ignoreUnknownSourceProperties."` or `"Source %s is being mapped despite appearing in ignoreUnknownSourceProperties."`.

integrationtest/src/test/resources/fullFeatureTest/pom.xml
```diff
@@ -47,6 +47,7 @@
                         <exclude>**/*Test.java</exclude>
                         <exclude>**/testutil/**/*.java</exclude>
                         <exclude>**/spi/**/*.java</exclude>
+                        <exclude>**/ignoreunmapped/**/UserMapper*.java</exclude>
```
filiphr: Why is this added here? If you have some mappers that produce errors then you should use `Erroneous` in their name. That's already excluded using `<exclude>**/*Erroneous*.java</exclude>`

codeswithritesh:
Thanks for the suggestion, [gnuemacscoin](https://github.com/gnuemacscoin)! I have updated the validation messages as recommended.

Below are example warning/error messages generated by MapStruct using the format:
"`Source %s mapped despite being listed in ignoreUnmappedSourceProperties.`"

• For a **single** source property:
"`Source property 'email' is mapped despite being listed in ignoreUnmappedSourceProperties.`"

• For **multiple** source properties:
"`Source properties 'email', 'username' are mapped despite being listed in ignoreUnmappedSourceProperties.`"

Let me know if you'd prefer a different phrasing.

@filiphr When you have a moment, could you please take a look at this PR? Your feedback would be appreciated.

filiphr (changes requested):
The approach looks OK @codeswithritesh. I've left one comment that should be addressed before merging this, we should not need changes in the `fullFeaturesTest/pom.xml`. We should instead use `Erroneous` in the name of the mappers, that are producing compile errors, in our tests

integrationtest/src/test/resources/fullFeatureTest/pom.xml
codeswithritesh: Thank you for the review, @filiphr

I’ve addressed your feedback:

1) The mapper that was producing the compile error has been moved into the erroneous subpackage, which is already excluded using `<exclude>**/erroneous/**/*.java</exclude>` in fullFeaturesTest/pom.xml.

2) The changes to fullFeaturesTest/pom.xml have been reverted accordingly.

Please let me know if anything else needs to be adjusted.

filiphr: Thanks @codeswithritesh. We also have

```
<exclude>**/*Erroneous*.java</exclude>
```

which excludes files that contain `Erroneous` in their name. In any case. Your changes look OK. Let's see what the CI says.

codeswithritesh:
Thank you, @filiphr! 
Glad to hear the changes look good, and great to see the CI checks have passed.
Let me know if there’s anything else you'd like me to update. Otherwise, I’ll wait for you to proceed with the merge — looking forward to it! 🙂

filiphr:
Thanks @codeswithritesh, I've merged this

codeswithritesh:
Thank you @filiphr

Issue #1601, opened 2018-09-01
Compatibility with non-JavaBeans-style property accessors


danielsonjh:
The following docs from AutoValue and Immutables seem to indicate that both libraries do not recommend using the JavaBeans-style property accessors (`getProperty()`, `isBoolean()`, etc)
https://github.com/google/auto/blob/master/value/userguide/index.md
https://immutables.github.io/

Both libraries do support JavaBeans-style accessors to some degree however
https://github.com/google/auto/blob/master/value/userguide/howto.md#beans
https://immutables.github.io/style.html

In Mapstruct 1.3.0.Beta1 only JavaBeans-style properties are supported.
To support this new convention, Mapstruct will need an option to treat all methods as potential properties.
With no differentiation between standard methods and property getters, it seems that Mapstruct either needs:

1. Mapped classes to only have property getters as methods.
2. An annotation to allow Mapstruct to ignore certain methods when not using JavaBeans-style conventions.

Option 1 is up to the developer, and for Option 2, there is already an annotation that can do this, `@Mapping(target = "somemethod", ignore = true)`

filiphr:
@danielsonjh I think that someone else requested this as well, but I am not sure if it was within an issue.

As you noticed Option 2 is already in there. As for Option 1 it can be achieved by using a custom `AccessorNamingStrategy` and it is not that difficult to write for your own use. I don't see why such an opt in strategy won't be acceptable for us (if someone wants to contribute it). The only thing is that one needs to be really careful as then all methods would be considered as properties.

sjaakd:
and `Immutables` already works out-of-the box with fluent-accessors, right? #1608 reports something similar for Lombok..

filiphr:
Everything works out-of-the box with fluent setters. There are no fluent getters actually. As fluent implies that you get the same object back and for getters you get something else.

debug-desperado:
With immutables I think it makes the most sense to consider all methods as properties, and then have the user exclude derived, lazy, and check methods if they cause problems (@Value.Derived, @Value.Lazy, @Value.Check). Those won't be on the builders, but I'm not sure how MapStruct could use that information.

Looking briefly over the code, if ImmutablesAccessorNamingStrategy extended CustomAccessorNamingStrategy instead of DefaultAccessorNamingStrategy, wouldn't that have this effect?

filiphr:
[...]

PR #3914, opened 2025-08-10
#1601 Immutables fluent getters support

thmasker:
The test fails. I need to somehow differentiate fluent getters from other methods without parameters and with return type that are not getters. Any idea is appreciated, as I haven't been able to find a way so far...


thmasker:
Closes #1601

processor/src/main/java/org/mapstruct/ap/spi/ImmutablesAccessorNamingStrategy.java
```diff
@@ -19,6 +20,17 @@
...
+    @Override
+    public boolean isGetterMethod(ExecutableElement method) {
+        return super.isGetterMethod( method ) || isFluentGetter( method );
+    }
+
+    private boolean isFluentGetter(ExecutableElement method) {
+        return method.getParameters().isEmpty() &&
+            method.getReturnType().getKind() != TypeKind.VOID &&
+            !JAVA_JAVAX_PACKAGE.matcher( method.getEnclosingElement().asType().toString() ).matches();
+    }
```
filiphr, line 32: I don't think that we should just make this like this. We should look at the Immutables `@Styles`, see https://immutables.github.io/style.html and determine based on that.

Otherwise, this just applies to everything.

Additionally, I really do not understand how this even works, the `getElementName` is not overridden, so I don't think that this works properly

processor/src/test/java/org/mapstruct/ap/test/fluent/getters/Issue1601Test.java
```diff
@@ -0,0 +1,47 @@
...
+    @WithServiceImplementation(provides = BuilderProvider.class, value = Issue1601BuilderProvider.class),
+    @WithServiceImplementation(provides = AccessorNamingStrategy.class, value = ImmutablesAccessorNamingStrategy.class)
+})
+public class Issue1601Test {
+
+    @ProcessorTest
+    public void shouldIncludeBuildType() {
+        var item = ImmutableItemDto.builder().id( "test" ).build();
+
+        var target = ItemMapper.INSTANCE.map( item );
```
filiphr, line 41: Please do not use `var`. We don't have a check for this, but I am not a fan of it and we are not using it anywhere else in our codebase.

filiphr (changes requested):
Looking at the code I am not sure that this works @thmasker. I've approved the CI to run, so let's see. In any case, I think we need a better more granular support here

PR #3920, opened 2025-08-28
docs: reference gh discussions

jamezrin:


filiphr:
Good catch @jamezrin. Thanks

Issue #3940, opened 2025-10-01
Non-deterministic code generation (SourceTargetMapperImpl.java)


Fr4nk03:
### Summary

The processor test suite (mapstruct/processor/src/test/java/org/mapstruct/ap/test/conversion/java8time) contains tests that fail intermittently due to non-deterministic behavior in the code generation process. These issues were discovered using the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool. The root causes are related to iterating over collections with no guaranteed order, such as method lists from the compiler and `HashSet`s.

### Root Cause Analysis

**Unstable Field Order:** The `MapperRetrievalProcessor` iterates through each ExecutableElement from elementUtils.getAllEnclosedExecutableElements() to add method into List<SourceMethod>. However, getAllEnclosedExecutableElements is implemented using a HashSet which the order is not guaranteed to be stable. The processor then discovers and creates the required `DateTimeFormatter` fields in this unstable order, leading to a non-deterministic layout in the final generated file.

### Potential Fix
1. I will open a pull request for modification of the comparing logic in JavaFileAssert.java by sorting the fields in both files to make the comparison more robust.
2. A more fundamental and valuable alternative would be to make the code generation itself predictable. It can be fixed by sorting the list of methods right after they are retrieved from the compiler, before any processing happens.

### Actual behavior

**Unstable Field Order in Generated Mappers**

The order of helper fields (specifically `DateTimeFormatter` fields) in the generated mapper implementation classes is not consistent across builds.

* **Run 1 Output:**
    ```java
    private final DateTimeFormatter dateTimeFormatter_dd_MM_yyyy_HH_mm_12071769242 = ...
    private final DateTimeFormatter dateTimeFormatter_HH_mm_168697690 = ...
    ```
* **Internal Field Order (from debug print):**
    ```
    ----- Fields Order -----
    SupportingField@2e6109be
    SupportingField@12251427
    SupportingField@aa6b84f3
    SupportingField@3e07ade
    ```

* **Run 2 Output:**
    ```java
    private final DateTimeFormatter dateTimeFormatter_HH_mm_168697690 = ...
    private final DateTimeFormatter dateTimeFormatter_dd_MM_yyyy_HH_mm_12071769242 = ...
    ```
* **Internal Field Order (from debug print):**
    ```
    ----- Fields Order -----
    SupportingField@12251427
    SupportingField@2e6109be
    SupportingField@3e07ade
    SupportingField@aa6b84f3

[...]

PR #3941, opened 2025-10-01
#3940 Fix the non-determinism by modifying file assertion logic

Fr4nk03:
### Summary

This pull request addresses non-deterministic behavior in the processor module, such as `Java8CustomPatternDateTimeFormatterGeneratedTest`, which fail intermittently due to non-deterministic field ordering in the generated code. This behavior was identified using the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool.

### The Root Cause

The core issue is that the code generation process does not guarantee a stable order for helper fields (e.g., `DateTimeFormatter` instances). This is because the processor iterates over a list of methods, which are stored in HashSet initially, in an order that can vary between builds. As a result, the generated mapper implementation can have its fields declared in different sequences, causing tests that perform a direct string comparison against a "golden file" to fail.

### The Implemented Solution

This PR makes the test suite more robust by changing the assertion logic rather than the code generation logic itself.

The `JavaFileAssert` utility has been modified to handle non-deterministic field order. The new assertion logic works as follows:

1.  Parses both the actual and expected Java source files.
2.  Separates the member fields from the rest of the class implementation.
3.  Sorts both lists of fields alphabetically.
4.  Compares the sorted lists of fields for equality.
5.  Compares the remaining parts of the class implementation for equality.

This approach ensures that the test correctly verifies the generated code's content and structure while being resilient to changes in field ordering.

### Alternative Considered

An alternative approach would be to make the code generation itself deterministic by sorting the list of methods in `MethodRetrievalProcessor` before processing. This PR takes the approach of hardening the *test assertion* as it is a more localized change that directly addresses the test's flakiness without modifying the core processor's behavior.

Issue related: #3940

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/testutil/assertions/JavaFileAssert.java
+++ b/processor/src/test/java/org/mapstruct/ap/testutil/assertions/JavaFileAssert.java
@@ -5,19 +5,20 @@
  */
 package org.mapstruct.ap.testutil.assertions;
 
-import org.assertj.core.api.FileAssert;
-import org.assertj.core.error.ShouldHaveSameContent;
-import org.assertj.core.internal.Diff;
-import org.assertj.core.internal.Failures;
-import org.assertj.core.util.diff.Delta;
-
 import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
-import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


filiphr:
Thanks for the PR @Fr4nk03. I actually found the real root cause for the non-deterministic behavior. The fix is in PR #3987. Therefore, I am closing this one in favor of that one.

Issue #3943, opened 2025-10-14
Cannot create mapping method with single primitive argument


anenviousguest:
### Expected behavior

Given a mapping method which takes a single argument of primitive `long`, like below:

```java
import org.junit.jupiter.api.*;
import org.mapstruct.*;
import org.mapstruct.factory.Mappers;

import static org.assertj.core.api.Assertions.assertThat;

class PrimitiveMapperTest {

    @Test
    @DisplayName("should convert primitive type")
    void shouldConvertPrimitiveType() {
        var mapper = Mappers.getMapper(PrimitiveMapper.class);
        var target = mapper.map(123);
        assertThat(target.value()).isEqualTo(123);
    }

    protected record Target(long value) {}
}

@Mapper
abstract class PrimitiveMapper {
    @Mapping(target = "value", source = "value")
    public abstract PrimitiveMapperTest.Target map(long value);
}
```

An expectation is that the mapping method is generated and the test passes.

(Mapstruct is configured with `-Amapstruct.unmappedTargetPolicy=ERROR`)

### Actual behavior

Compilation fails with a message:

> `java: Can't generate mapping method with primitive parameter type.`

Funny enough, it starts working after adding a 2nd argument which could also be of a primitive (sic!) type. I could make the above example working like this:

```java
    ...
    public abstract PrimitiveMapperTest.Target map(long value, long unused);
```

### Steps to reproduce the problem
See above

### MapStruct Version

1.6.3

PR #3944, opened 2025-10-15
#3943: Allow generation of methods with a sole argument of primitive type

anenviousguest:
Took a stab at fixing #3943. 

Not quite sure what was the reason for explicitly forbidding such methods back then, given that it works perfectly fine - at least as far as the test below is concerned.


anenviousguest:
Hi @filiphr, any chance to get a feedback on this one? Thank you.

filiphr:
As you can see I've been quite busy with a lot of stuff. I kept postponing 1.7, so I wanted to do a first Beta release for it, before going deeper into other PRs. I'll get to this one.

filiphr:
One thing I would say is that if we are going to allow this, then we need to do it more proactively. It works if the target has a property named the same as the source value. However, if that's not the case, then we would treat the primitive as a bean. Therefore, I think that we need to do some adjustments in the `BeanMappingMethod#build`. 

i.e. Can you add a test case where the target does not have a property matching the source?

anenviousguest:
@filiphr it's been a while, apologies. I added a couple of extra tests, also the one which illustrates compilation failure with unmapped property - if that's what you've meant.

anenviousguest:
Hi @filiphr, any chance that this can be reviewed at some point? Thank you.

anenviousguest:
Hi @filiphr, any chance that this can be reviewed at some point? Thank you.

anenviousguest:
Hi @filiphr - I've just stumbled upon this issue in my other codebase  - wonder if there is any chance that this can be reviewed at some point? Thank you.

PR #3951, opened 2025-11-02
#3818: Fix target="." Doesn't Disambiguate Multiple Mappings

### LINKED ISSUE #3818: target="." doesn't disambiguate multiple mappings
### Expected behavior

using `@Mapping( target = ".", source = "record" )` should be equivalent as adding several `@Mapping( target = "x", source = "record.x" )` for all the properties on 'record' (unless there is already another explicit `@Mapping` with it).
Source: https://mapstruct.org/documentation/dev/reference/html/#mapping-nested-bean-properties-to-current-target

In particular, a code like this one:
```java
@Mapper
public interface Example {

  record Sample(
      String first,
      String second,
      String third
  ) {}

  @Mapping(target = ".", source = "main")
  @Mapping(target = "third", source = "secondary.third")
  Sample merge(Sample main, Sample secondary);
}
```

Should be valid and generate a mapping equivalent to `return new Sample(main.first, main.second, secondary.third)`.
From a user perspective, this mapping means: 'Use everything from main, except 'third' from secondary.

### Actual behavior

With the code above, mapstructs reports the following two errors:
```txt
Several possible source properties for target property "first".
Several possible source properties for target property "second".
```

It appears that the `@Mapping(target = ".", source = "main")` is not used for disambiguation.

Here are other examples with different outcomes:

<details><summary>Specify all fields: works</summary>
<p>
The following code:

```java
@Mapper
public interface Example {

  record Sample(
      String first,
      String second,
      String third
  ) {}
  
  @Mapping(target = "first", source = "main.first")
  @Mapping(target = "second", source = "main.second")
  @Mapping(target = "third", source = "secondary.third")
  Sample merge(Sample main, Sample secondary);
}
```
Correctly compiles are generates the expected mapping (equivalent to `return new Sample(main.first, main.second, secondary.third)`)

</p>
</details>

<details><summary>Secondary object without duplicated fields: works</summary>
<p>
The following code:

```java
@Mapper
public interface Example {

  record Sample(
      String first,
      String second,
      String third
  ) {}

  record DifferentSample(
      String third
  ) {}

  @Mapping(target = ".", source = "main")
  @Mapping(target = "third", source = "secondary.third")
  Sample merge(Sample main, DifferentSample secondary);
}
```
Correctly compiles are generates the expected mapping (equivalent to `return new Sample(main.first, main.second, secondary.third)`).
In this case there is no ambiguity.

</p>
</details> 

<details><summary> Double target=".": same outcome </summary>
<p>
The following two pieces of code:

```java
@Mapper
public interface Example {

  record Sample(
      String first,
      String second,
      String third
  ) {}

  @Mapping(target = ".", source = "main")
  @Mapping(target = ".", source = "secondary")
  @Mapping(target = "third", source = "secondary.third")
  Sample merge(Sample main, Sample secondary);
}
```

```java
@Mapper
public interface Example {

  record Sample(
      String first,
      String second,
      String third
  ) {}

  @Mapping(target = "third", source = "secondary.third")
  Sample merge(Sample main, Sample secondary);
}
```

Generates the same outcome as the 'expected behavior' example: `Several possible source properties for target property "first"/"second"`.

Based on this it appears as if something like `@Mapping(target = ".", source = "value")` is implicitly added to all mappings with a single parameter `value`, which means that explicitly adding it doesn't really do anything, instead of disambiguating.

</p>
</details>

---

Extra, if this were working as expected, suggestions like https://github.com/mapstruct/mapstruct/issues/3161 would not need any extra 'PrimaryMapping' annotation, as that would be fixed with a target="." mapping.
```java
@Mapping(target = "borrowed", source = "tblB.borrowed")
@Mapping(target = ".", source = "tblA")
EntityA toEntity(TblA tblA, TblB tblB)
```

### MapStruct Version

1.6.3

fgalaf:
Fixes #3818 

The implementation checks whether parameter has specifically defined child elements. In case it does, error message `Several possible source properties for target property "%s".` is not thrown.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_3818/Issue3818Mapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_3818/Issue3818Mapper.java
@@ -1,3 +1,9 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
 package org.mapstruct.ap.test.bugs._3818;
 
 import org.mapstruct.Mapper;
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_3818/Issue3818Test.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_3818/Issue3818Test.java
@@ -1,3 +1,9 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
 package org.mapstruct.ap.test.bugs._3818;
 
 import org.mapstruct.ap.testutil.IssueKey;


fgalaf:
Hi @filiphr, I saw that this job failed on step `Some files do not have the expected license header`.
So I added the comment which I believe is the mentioned license header. Now the pipeline should succeed.

PR #3952, opened 2025-11-05
change JavaLocalDateTimeToDateConversion zoneId UTC -> systemDefault

986510453:
modifying the JavaLocalDateTimeToDateConversion, change the zoneId from UTC to systemDefault


filiphr:
We cannot accept this change since it's a backwards incompatible one. If you would like to change the default that MapStruct generates it would need a bit more work so that this would be configurable.

PR #3968, opened 2026-01-09
Fix example #5 code comment.

ro-otgo:
Example 5: Mapping with default value should be using fullName instead of just name in order to map the property.


filiphr:
Thanks @ro-otgo

PR #3971, opened 2026-01-25
Add support for `Optional`

### LINKED ISSUE #674: Support Java 8 java.util.Optional type

filiphr:
This PR builds on top of PR #3183 and resolves #674.

The most notable things in the additional rework of #3183 are the following:

* Implement mapping as part of Bean Mapping
* Always assume that optionals are not `null`, i.e. we can always call `isPresent()` / `isEmpty()` on an `Optional`
* Add support for conversions e.g. `Optional<String>` -> `String`, `Optional<Long>` -> `String` etc.

### DIFF
--- a/NEXT_RELEASE_CHANGELOG.md
+++ b/NEXT_RELEASE_CHANGELOG.md
@@ -1,6 +1,11 @@
 ### Features
 
 * Support for Java 21 Sequenced Collections (#3240)
+* Native support for `java.util.Optional` mapping (#674) - MapStruct now fully supports Optional as both source and target types:
+  - `Optional` to `Optional` - Both source and target wrapped in `Optional`
+  - `Optional` to Non-`Optional` - Unwrapping `Optional` values
+  - Non-`Optional` to `Optional` - Wrapping values in `Optional`
+  - `Optional` properties in beans with automatic presence checks. Note, there is no null check done for `Optional` properties.
 * Improved support for Kotlin. Requires use of `org.jetbrains.kotlin:kotlin-metadata-jvm`.
   - Data Classes (#2281, #2577, #3031) - MapStruct now properly handles:
     - Single field data classes


documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
```diff
@@ -266,6 +267,18 @@ null check, regardless of the value of the `NullValuePropertyMappingStrategy`, t
...
 ====
 
+[NOTE]
+====
+When working with `java.util.Optional` types, `NullValuePropertyMappingStrategy` applies to empty Optionals in the same way it applies to null values.
+An empty `Optional` (i.e., `Optional.isEmpty()` returns `true`) is treated similarly to a `null` value for the purposes of this strategy. For example:
+
+* With `IGNORE` strategy: an empty `Optional` source property will not update the target property
+* With `SET_TO_NULL` strategy: an empty `Optional` source will set the target to `null` (for non-`Optional` targets) or `Optional.empty()` (for Optional targets)
+* With `SET_TO_DEFAULT` strategy: an empty Optional source will set the target to its default value
```
rala72: small nit-pick (to be consitent with the two lines above):
```suggestion
* With `SET_TO_DEFAULT` strategy: an empty `Optional` source will set the target to its default value
```

filiphr: Good point @rala72. Thanks

robotmrv:
regarding  
> Always assume that optionals are not null, i.e. we can always call isPresent() / isEmpty() on an Optional  

this looks like a breaking change  
see case with jackson + Optional deserialization https://github.com/mapstruct/mapstruct/pull/3183#issuecomment-1451548248   
In this case it is possible that Optional is `null` and probably could break existing users after update to a new version. Or at least this feature will be unusable for such case.

Is it possible to add config option for mapping like this  
| Source Type | Target Type	| Source Value	| Target Value |
| --------------|-----------------|----------------|---------------|
| Object	| Optional	| null	| null|
| Optional	|Object	| null	| null|
| Optional	| Optional	| null	| null|

i.e. `null` maps to `null` always for objects
?

filiphr:
>this looks like a breaking change

Breaking change compare to what? Currently we do not have support for optional at all.
Can you provide certain things that work in the moment, but won't keep working?

I personally really dislike using `Optional` as input parameter and / or a field, IntelliJ even constantly complains about that and in addition to that the `Optional` javadoc itself says

>`Optional` is primarily intended for use as a method return type where there is a clear need to represent "no result," and where using `null` is likely to cause errors. A variable whose type is `Optional` should never itself be `null`; it should always point to an `Optional` instance.

>see case with jackson + Optional deserialization

This is not because Jackson does that, it's more because Jackson will not invoke the setter if there is nothing in the json.

robotmrv:
breaking in terms that mapstruct usually checks source for null. But breaking was a wrong word maybe unexpected/surprising because null check is a boilerplate code and using mapstruct is the main use case for null safe boilerplate mapping code generation. 
And I am not sure how it would work with generic mappers for Optional like 
```java
@Mapper

    default <E> E unwrapFromOptional(Optional<E> optional) {
        return optional == null
            ? null
            : optional.orElse(null);
    }

    default <E> Optional<E> wrapToOptional(E value) {
        return value == null
            ? null
            : Optional.of(value);
    }
}
```

> I personally really dislike using Optional as input parameter and / or a field

I don't like it either, but ther are a lot of code that used it this way and at least part of the mapstruct users use is it this way and cannot change it due to their API contracts. 
 
> This is not because Jackson does that, it's more because Jackson will not invoke the setter if there is nothing in the json. 

This is just implementaion detail and the main point is that Optional is used as a container for patch requests. Because Optional is  part of the jdk and does not require additional dependencies and because jackson supports such case with Optional out of the box. 

An additional use case is cloning objects using mappers that map to the same type, and you want this to be symmetrical, but this is not possible with the `null -> Optional.empty()` asymmetric approach.

filiphr:
>I don't like it either, but ther are a lot of code that used it this way and at least part of the mapstruct users use is it this way and cannot change it due to their API contracts.

If MapStruct users are using this right now it means that they have custom methods where this is being done, this means that this will keep working like that. User defined methods always have a precedence over MapStruct internals.

>An additional use case is cloning objects using mappers that map to the same type, and you want this to be symmetrical, but this is not possible with the `null -> Optional.empty()` asymmetric approach.

That stays the same. Have a look at the `OptionalSameTypeMapperImpl` in the PR. I generates:

```java
constructorOptionalToOptional = source.getConstructorOptionalToOptional();
```

When the optional can be directly assigned, it is assigned without a presence check. 

>And I am not sure how it would work with generic mappers for Optional like

I added some tests like that. 

I am not super keen on adding something that only small number of users is going to use, especially if it can be achieved with custom methods. I would actually prefer to try to keep it simple and in the spirit of how optional is supposed to be used without any warnings from the IDE. If I treat `Optional` as `null` then IntelliJ gives me a warning about it, same when I use it as a method parameter.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/OptionalGetWrapper.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public Type getOptionalType() {
+        return optionalType;
+    }
+
+    @Override
+    public String toString() {
+        if ( optionalType.getFullyQualifiedName().equals( "java.util.Optional" ) ) {
+            return getAssignment() + ".get()";
```
rala72, line 31: Even since there is no difference in the result, I just wanted to mention that  
the (official) preferred alternative instead of `get()` is [`orElseThrow()`](https://devdocs.io/openjdk~21/java.base/java/util/optional#get()).

_(Not sure if this is the correct place for this comment; it just looked the most promising for me.)_

filiphr: `orElseThrow()` is there from Java 10 and we aim to support Java 8 code generation. It could be an improvement that we can do (like we are doing with the `isEmpty()` check), but I would do it separately from this PR. I've opened #3976 as a follow-up for this

sanmibuh:
Good job!!!!!

Issue #3949, opened 2025-10-28
Generated fill() with overloaded setters leads to compilation error.


morrica:
### Expected behavior

Compilation succeeds and fill method functions properly.

### Actual behavior

Generated method looks like the following and compilation fails with the error "java: reference to time is ambiguous".
```
    @Override
    public void fill(Event.Builder target, EventDto src) {
        if ( src == null ) {
            return;
        }

        target.transactionId( src.transactionId() );
        target.shopId( src.shopId() );
        if ( src.time() != null ) {
            target.time( ZonedDateTime.parse( src.time() ) );
        }
        else {
            target.time( null );
        }
        target.type( src.type() );
    }
```

### Steps to reproduce the problem

The target `Event.Builder` contains overloaded methods for setting time `time(Date)` and `time(ZonedDateTime)`. The generated `fill()` method checks the src `time` field for `null`. If it is `null` it directly calls `target.time(null)` leading to the compilation failure.

Instead, the generated `fill()` method should call `target.time((Date) null)` or something similar.

### MapStruct Version

MapStruct 1.6.3

thunderhook:
Hi @morrica!

Can you please provide the code of `Event`, `EventDto` and your mapper interface? It is easier to reproduce it that way. Thanks!

PR #3988, opened 2026-02-01
#3949 Support SET_TO_NULL for overloaded target methods, requiring a cast

hduelme:
Currently, invalid code is generated when `SET_TO_NULL` is used, as described in #3949.

This PR changes the generated code to include an explicit cast when more than two setters with the same name are present.

MapStruct does not fully support overloaded target methods yet (see #1135). To account for this limitation, the tests use a workaround: the preferred target type is determined via the getter with the same name to verify that the correct setter is invoked when passing `null`.


processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
```diff
@@ -916,6 +916,10 @@ public List<Element> getRecordComponents() {
         return recordComponents;
     }
 
+    public long settersWithName(String name) {
+        return getSetters().stream().filter( setter -> setter.getSimpleName().equals( name ) ).count();
+    }
```
filiphr: I would not expose this in the `Type`. This check should be in the `PropertyMapping`. We can actually implement this a bit better and stop the iteration if the count reaches 2. No need to iterate through all setters.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -2273,6 +2274,12 @@ public Set<Type> getImportTypes() {
                 // We need to add the target type imports for a constructor mapper since we define its parameters
                 types.addAll( propertyMapping.getTargetType().getImportTypes() );
             }
+            if (propertyMapping.getAssignment() instanceof SetterWrapper
+                    && ((SetterWrapper) propertyMapping.getAssignment()).isSetExplicitlyToNull()
+                    && ((SetterWrapper) propertyMapping.getAssignment()).isMustCastForNull()) {
+                // Setting to null with cast might require an import
+                types.add( propertyMapping.getTargetType() );
+            }
```
filiphr: This is not the responsibility of the `BeanMappingMethod`. The `getImportTypes()` of the `SetterWrapper` should do this instead. We should pass the `targetType` to the `SetterWrapper`, in a similar way like we are doing for the `UpdateWrapper`

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -471,7 +471,9 @@ private Assignment assignToPlainViaSetter(Type targetType, Assignment rhs) {
                     isFieldAssignment(),
                     includeSourceNullCheck,
                     includeSourceNullCheck && nvpms == SET_TO_NULL && !targetType.isPrimitive(),
-                    nvpms == SET_TO_DEFAULT );
+                    nvpms == SET_TO_DEFAULT,
+                    this.method.getResultType().settersWithName( this.targetWriteAccessor.getSimpleName() ) > 1
```
filiphr: We cannot implement this like this, because the method `resultType` is not what has the setters. If the `resultType` has a builder then the setters we are using are actually in the builder, not in the result type.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Type.java
hduelme: Moved to `PropertyMapping` and break if the second one is found.

processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
hduelme: Moved to `SetterWrapper` like you suggested

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
hduelme: Changed to look at the enclosing class.

hduelme:
@filiphr as you suggested, I moved the import logic to the `SetterWrapper`.
Now the containing class of the `targetWriteAccessor` is checked if more than one method with the name is present.

In addition I added the tests where  sub-mapping is generated.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -471,7 +473,10 @@ private Assignment assignToPlainViaSetter(Type targetType, Assignment rhs) {
                     isFieldAssignment(),
                     includeSourceNullCheck,
                     includeSourceNullCheck && nvpms == SET_TO_NULL && !targetType.isPrimitive(),
-                    nvpms == SET_TO_DEFAULT );
+                    nvpms == SET_TO_DEFAULT,
+                        hasTwoOrMoreSettersWithName(),
+                        targetType
```
filiphr: Style: the two new constructor arguments have extra indentation compared to the other arguments. Please align them consistently. Same issue at the second call site (line 556-557).

```suggestion
                    hasTwoOrMoreSettersWithName(),
                    targetType
```

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -547,12 +552,32 @@ else if ( result.getSourceType().isStreamType() ) {
...
+        private boolean hasTwoOrMoreSettersWithName() {
+            Element enclosingClass = this.targetWriteAccessor.getElement().getEnclosingElement();
+            if ( enclosingClass == null || !ElementKind.CLASS.equals( enclosingClass.getKind() ) ) {
+                return false;
+            }
+            String simpleWriteAccessorName = this.targetWriteAccessor.getSimpleName();
+            boolean firstMatchFound = false;
+            for ( Accessor setter : ctx.getTypeFactory().getType( enclosingClass.asType() ).getSetters() ) {
+                if ( setter.getSimpleName().equals( simpleWriteAccessorName ) ) {
+                    if (firstMatchFound) {
```
filiphr: Style: the project consistently uses spaces inside `if` parentheses: `if ( condition )`. Please adjust to match.

```suggestion
                    if ( firstMatchFound ) {
```

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -547,12 +552,32 @@ else if ( result.getSourceType().isStreamType() ) {
...
+                    hasTwoOrMoreSettersWithName(),
+                        targetType
                 );
             }
             return result;
         }
 
+        private boolean hasTwoOrMoreSettersWithName() {
+            Element enclosingClass = this.targetWriteAccessor.getElement().getEnclosingElement();
+            if ( enclosingClass == null || !ElementKind.CLASS.equals( enclosingClass.getKind() ) ) {
```
filiphr: This check only allows `ElementKind.CLASS`. If the target type is an interface with overloaded default setter methods, this returns `false` and the cast won't be emitted — leaving the same compilation error unfixed for that case.

Could you broaden this to also accept interfaces and add a test case for that scenario?

```java
ElementKind kind = enclosingClass.getKind();
if ( enclosingClass == null
    || ( !ElementKind.CLASS.equals( kind )
        && !ElementKind.INTERFACE.equals( kind ) ) ) {
    return false;
}
```

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapper.java
```diff
@@ -22,19 +24,25 @@ public class SetterWrapper extends AssignmentWrapper {
     private final boolean includeSourceNullCheck;
     private final boolean setExplicitlyToNull;
     private final boolean setExplicitlyToDefault;
+    private final boolean mustCastForNull;
+    private final Type targetImplementationType;
```
filiphr: Naming: in the MapStruct codebase, `targetImplementationType` conventionally refers to the concrete implementation type for abstract/collection types (e.g., `ArrayList` for `List` — see `UpdateWrapper`). Here it holds the setter's parameter type used for casting. Consider renaming to something like `castType` or `nullCastType` to avoid confusion.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapper.java
```diff
@@ -59,6 +69,15 @@ public List<Type> getThrownTypes() {
         return result;
     }
 
+    @Override
+    public Set<Type> getImportTypes() {
+        Set<Type> imported = new HashSet<>(super.getImportTypes());
```
filiphr: Style: other `getImportTypes()` overrides in the assignment package use spaces inside the constructor call — `new HashSet<>( super.getImportTypes() )`.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
hduelme: indentation fixed

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
hduelme: missing spaces added

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
hduelme: Good catch. I added the check and  the tests accordingly.

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapper.java
hduelme: missing spaces added

processor/src/main/java/org/mapstruct/ap/internal/model/assignment/SetterWrapper.java
hduelme: `targetImplementationType` renamed to `nullCastType`

hduelme:
@filiphr thanks for the review. You caught two edge cases I had overlooked.
- Line 21 in `UpdateWrapper.ftl` could generate code with a missing cast.
- Interfaces are valid targets and can have two methods with the same name.

For line 21, I added `updateParent` to the test.
For interfaces `Issue3949InterfaceMapper` iis now used.

Formatting and naming should be fixed now.

filiphr:
Thanks @hduelme

filiphr:
>You caught two edge cases I had overlooked.

Btw, I didn't caught them. Claude Code did it 😉

PR #3989, opened 2026-02-01
Remove unnecessary keySet() invocation

hduelme:
The size of the `keySet` is always equal to the size of the map, so the call to `keySet() `can be skipped.


filiphr:
Thanks @hduelme

PR #3991, opened 2026-02-03
improve performance of Type.describe() by removing regex matching

hduelme:
In `Type.describe(),` the `packageName` should be removed from the beginning of the `fullyQualifiedName`.
The current implementation uses a regex for this, which results in a regex compilation on every call. In addition, the dots in the regex are not escaped, which could theoretically lead to overmatching (even though this should never happen).

I changed the implementation to remove the `packageName` if the `fullyQualifiedName` actually starts with it, by using `substring`.


filiphr:
NIce, thanks @hduelme

PR #3995, opened 2026-02-11
fix self check in equals of Type

hduelme:
While investigating #3994, I noticed that in `Type.equals`, `hasExtendsBound()` and `hasSuperBound()` are compared to themselves, which always evaluates to `true`.

As far as I can tell, equals is currently only used for `Map` comparisons. Nevertheless, this behavior seems wrong.


filiphr:
Thanks @hduelme

PR #4002, opened 2026-02-26
use StandardCharsets.UTF_8 in tests

hduelme:
Since java 1.6 is no longer supported, `StandardCharsets` can be used instead of string lookup.


filiphr:
Thanks a lot @hduelme. Seems like there is some kind of a conflict here. Do you think you can resolve it?

hduelme:
@filiphr The conflict has been resolved with a rebase.

filiphr:
Thanks @hduelme

PR #4003, opened 2026-02-26
simplify fail in assertCheckstyleRules

hduelme:
If Checkstyle finds any error the method should fail. Using `fail()` directly is more obvious.


filiphr:
Thanks @hduelme

PR #4005, opened 2026-02-27
Fixed primitive early return if all ref types are null

### LINKED ISSUE #3992: MapStruct ignores primitive parameters when all non-primitive parameters are null due to generated null-check logic
### Expected behavior

MapStruct generates a null-check that considers only non-primitive method parameters. When all non-primitive arguments are null, the generated mapper method returns null early, even if primitive parameters contain valid values.

As a result, mappings for primitive (or boxed-with-default) arguments are skipped entirely.

Expected behavior

If primitive parameters (e.g. int, boolean) have valid values, MapStruct should still map them to the target object, even when all non-primitive parameters are null.

At minimum, primitive arguments should not be ignored solely because non-primitive arguments are null.

Actual behavior

MapStruct generates an early return null when all non-primitive parameters are null, causing primitive argument mappings to be ignored.

Example

Mapper method:

```
Target map(
    String sku,
    String name,
    Long categoryId,
    int size,
    int pageNumber,
    boolean draft
);

```

Generated logic (simplified):

```
if (sku == null && name == null && categoryId == null) {
    return null;
}

target.setSize(size);
target.setPageNumber(pageNumber);
target.setDraft(draft);

```

In this case:

- sku, name, and categoryId are null
- 
- size, pageNumber, and draft have valid values

However, the method returns null, and primitive fields are never mapped.

Impact

- This behavior makes it impossible to create partially populated target objects when:
- 
- all reference-type parameters are null
- 
- primitive parameters still carry meaningful values (e.g. pagination, flags)
- 
- This is especially problematic for search/filter DTOs.

Workaround

- Convert primitive parameters to wrapper types (Integer, Boolean)
- 
- Or manually implement mapping logic instead of relying on generated code

### Actual behavior

_No response_

I have told in description

### MapStruct Version

1.6.3

ronodhirSoumik:
fixes #3992 
Updated the check in the ftl file to only generate an early return-null block if every single parameter is subject to a null check. If there are parameters that don't need null checks (like primitives), keep going and map them.

### DIFF
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/BeanMappingMethod.ftl
@@ -26,7 +26,7 @@
 
     	</#if>
     </#list>
-    <#if !mapNullToDefault && !sourcePresenceChecks.empty>
+    <#if !mapNullToDefault && sourcePresenceChecks?has_content && sourcePresenceChecks?size == sourceParameters?size>
     if ( <#list sourcePresenceChecks as sourcePresenceCheck><@includeModel object=sourcePresenceCheck.negate() /><#if sourcePresenceCheck_has_next> && </#if></#list> ) {
         <#if returnType.name == "void">
             return;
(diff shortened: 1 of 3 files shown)


filiphr:
@ronodhirSoumik the CI is failing. Can you please fix it first and then I'll go ahead and review it

ronodhirSoumik:
> @ronodhirSoumik the CI is failing. Can you please fix it first and then I'll go ahead and review it

resolved

Issue #3937, opened 2025-09-25
Support for Builder like "record-builder"


smikesmike:
### Use case

When I am using records with `@RecordBuilder` (https://github.com/Randgalt/record-builder), then I receive warnings in my mapstruct mapper like:

```
Unmapped target properties: "with, withId, withName, withLogin, withEmail, withPassword, withRole, withCreatedAt, withRegistered, withEnabled".
```

It would be nice to have support for wildcards in mapping targets like:

```
@Mapping(target = "with*", ignore = true)
```

BR
Marc

rlei:
a side note: PR #4006 adds support for Immutables new `record` type support (`@Value.Builder`). `record-builder` generates builder the same way ("{Type}Builder") so the PR only needs trivial update to support `record-builder` too.

PR #4006, opened 2026-02-28
Add @Value.Builder record builder support to ImmutablesBuilderProvider

rlei:
## Problem

`ImmutablesBuilderProvider` supports `@Value.Immutable` on interfaces and abstract classes, where Immutables generates an `ImmutableXxx` implementation containing a nested `Builder`. However, since Immutables 2.11.0, Java records can be annotated with `@Value.Builder`, which causes Immutables to generate a top-level companion `{RecordName}Builder` class instead. The existing provider did not handle this case, causing MapStruct to fall back to calling the record constructor directly — bypassing the builder's mandatory-field enforcement.

## Solution

- Added `findBuilderInfoForImmutablesRecord()`: detects records annotated with `@Value.Builder` and looks up the companion `{RecordName}Builder` type.
- Added `asImmutableRecordBuilderElement()`: constructs the companion builder's qualified name following the `{RecordName}Builder` convention.
- Added `findBuilderInfoFromCompanionBuilder()`: constructs `BuilderInfo` directly using the companion class's public no-arg constructor (which Immutables always generates for `@Value.Builder` on records) and `findBuildMethods()` to locate `build()`.
- Added `immutablesRecordBuilderTest` integration test using Immutables 2.11.0, restricted to JDK compiler on Java 16+. Includes a `@Nullable` field and `NullValueCheckStrategy.ALWAYS` to demonstrate that the builder's mandatory-field check fires when MapStruct skips a null setter — behaviour that would be silently lost if MapStruct fell back to the record constructor.

## Notes

- `ElementKind.RECORD` was added in Java 16; the implementation uses `typeElement.getKind().name()` string comparison to remain compilable on older Java releases.
- `ImmutablesBuilderProvider` is automatically activated when `org.immutables.value.Value.Immutable` is on the classpath (see `AnnotationProcessorContext`); `@Value.Builder` is part of the same artifact so no additional activation is needed.


rlei:
Hi @filiphr could you please take a look at this one? 

With minor updates, this will also support the another `record` builder generator  (https://github.com/Randgalt/record-builder, see issue #3937).

rlei:
It seems the integration tests in the first commit was not considered in the coverage report.

@filiphr I've added unit tests to cover ImmutablesBuilderProviders, could you please check again?

processor/src/main/java/org/mapstruct/ap/spi/ImmutablesBuilderProvider.java
```diff
@@ -97,4 +176,29 @@ protected TypeElement asImmutableElement(TypeElement typeElement) {
...
     }
 
+    /**
+     * Constructs the qualified name of the Immutables-generated builder for a record type and returns the
+     * corresponding {@link TypeElement}. The builder class follows the {@code {RecordName}Builder} naming convention.
+     *
+     * @param typeElement the record type element
+     * @return the {@link TypeElement} for the generated builder, or {@code null} if not yet available
+     */
+    protected TypeElement asImmutableRecordBuilderElement(TypeElement typeElement) {
```
filiphr, line 172: The only difference between this and `asImmutableElement` is the prefix vs suffix. To reduce the duplication we could extract a shared method like:

```java
private TypeElement findRelatedTypeElement(TypeElement typeElement, String prefix, String suffix) {
    Element enclosingElement = typeElement.getEnclosingElement();
    StringBuilder qualifiedName = new StringBuilder(typeElement.getQualifiedName().length() + 17);
    if (enclosingElement.getKind() == ElementKind.PACKAGE) {
        qualifiedName.append(((PackageElement) enclosingElement).getQualifiedName().toString());
    } else {
        qualifiedName.append(((TypeElement) enclosingElement).getQualifiedName().toString());
    }
    if (qualifiedName.length() > 0) {
        qualifiedName.append(".");
    }
    qualifiedName.append(prefix).append(typeElement.getSimpleName()).append(suffix);
    return elementUtils.getTypeElement(qualifiedName);
}
```

Then:
* `asImmutableElement` -> `findRelatedTypeElement(typeElement, "Immutable", "")`
* `asImmutableRecordBuilderElement` -> `findRelatedTypeElement(typeElement, "", "Builder")`.

processor/src/test/java/org/immutables/value/Value.java
filiphr, line 1: If we are adding tests to the `processor` module then let's do this properly. We should not using other dependency packages. Instead we could add the immutables as a dependency or we could do something similar like we are doing in `org/mapstruct/ap/test/bugs/_1801`. I think adding it as a dependency is OK.

processor/src/test/java/org/mapstruct/ap/test/builder/immutables/record_/ImmutablesRecordBuilderTest.java
```diff
@@ -0,0 +1,60 @@
...
+@WithClasses({
+    Value.class,
+    GolfPlayer.class,
+    GolfPlayerBuilder.class,
+    GolfPlayerDto.class,
+    GolfPlayerMapper.class
+})
+class ImmutablesRecordBuilderTest {
+
+    @ProcessorTest(Compiler.JDK)
```
filiphr, line 35: Why only `Compiler.JDK`? Are there some problems with Eclipse?

filiphr (changes requested):
Thanks for adding the tests in the processor module @rlei. I was going to merge it without it as well, but it's better to have them closer. However, I think that we should not add the `@Value` like you did. I've left some comments inline. If we address those then I'll go ahead and merge this.

processor/src/main/java/org/mapstruct/ap/spi/ImmutablesBuilderProvider.java
rlei: Thank you @filiphr for the review! Will update the PR shortly.

EDIT: fixed

processor/src/test/java/org/mapstruct/ap/test/builder/immutables/record_/ImmutablesRecordBuilderTest.java
rlei: it's taken from test existing record related tests `processor/src/test/java/org/mapstruct/ap/test/record/jdk17/RecordsTest.java`. 

The Eclipse test runner currently hardcodes its source/target version to "11" when running on modern JDKs, so `record` is not supported (JDK 17+):
```java
  private static String getSourceVersion() {
      SourceVersion latest = SourceVersion.latest();
      if ( latest == SourceVersion.RELEASE_8 ) {
          return "1.8";
      }
      return "11";
  }
```

processor/src/test/java/org/immutables/value/Value.java
rlei: sure, let me add Immutables as a dep.

rlei:
Hi @filiphr sorry for the late update - was away for some time.

I've updated the PR and replied to your comments, could you please review it again?

PR #4007, opened 2026-03-03
Simplified boolean logic in ValueMappingMethod by removing inversion

hduelme:
Remove redundant boolean negation in enum mapping checks


filiphr:
Thanks @hduelme

PR #4011, opened 2026-03-04
upgrade Visitor6 to Visitor8

hduelme:
`*Visitor6` are deprecated. Since java 1.6 is no longer supported, they can be upgraded to java 1.8 `*Visitor8`


filiphr:
Thanks @hduelme

PR #4013, opened 2026-03-04
remove deprecated Number api usage from tests

hduelme:
Using the constructor of any number type is deprecated and results in a compilation warning.


filiphr:
Thanks @hduelme

PR #4014, opened 2026-03-11
#2436 Apply defaultValue/defaultExpression when a source parameter is null

### LINKED ISSUE #2436: With multiple source parameters, allow using defaultValue when a source parameter is null
When creating a mapper with multiple source parameters, it would be helpful if there was a way to specify a property default when one of the source parameters is null.  I've tested this against a local build of the latest 1.5.0-SNAPSHOT (since I was guessing #2023 might have addressed this issue), and the `defaultProperty` is ignored when the source is null.  It would be handy if `defaultProperty` or some other similar declarative way existed to specify a property default when the source object is null.

It occurs to me that a solution to this may also extend to standard a single-parameter mapper: it may be handy to be able to tell MapStruct to use a specific value rather than null/empty when the sole argument is null.  However, the need for this is more apparent in the multi-parameter case, as there may be legit reasons and mapping needs when some specific source arguments are null.

## Example

```java
import org.mapstruct.Mapper;
import org.mapstruct.Mapping;

public class DefaultValueIssue {
        MyMapper mapper = Mappers.getMapper(MyMapper.class);
        WorkaroundMapper workaroundMapper = Mappers.getMapper(WorkaroundMapper.class);

        System.out.println(mapper.map(null, new Source2()).getProperty1()); // null
        System.out.println(workaroundMapper.map(null, new Source2()).getProperty1()); // DEFAULT
        System.out.println(mapper.map(new Source1(null), new Source2()).getProperty1()); // DEFAULT
    }

    @Mapper
    public interface MyMapper {
        @Mapping(target = "property1", defaultValue = "DEFAULT")
        Target map(Source1 source1, Source2 source2);
    }

    @Mapper
    public interface WorkaroundMapper {
        @Mapping(target = "property1",
                expression = "java(source1 == null ? \"DEFAULT\" : source1.getProperty1())")
        Target map(Source1 source1, Source2 source2);
    }

    public static class Source1 {
        private String property1;

        public Source1() { }
        public Source1(String property1) { this.property1 = property1; }
        public String getProperty1() { return property1; }
        public void setProperty1(String property1) { this.property1 = property1; }
    }

    public static class Source2 {
        private String property2;

        public String getProperty2() { return property2; }
        public void setProperty2(String property2) { this.property2 = property2; }
    }

    public static class Target {
        private String property1;
        private String property2;

        public String getProperty1() { return property1; }
        public void setProperty1(String property1) { this.property1 = property1; }
        public String getProperty2() { return property2; }
        public void setProperty2(String property2) { this.property2 = property2; }
    }
}
```

dlwldnjs1009:
When a source parameter itself is `null` in a multi-source mapping, or in a single-source mapping with `RETURN_DEFAULT`, the generated code now applies `defaultValue` / `defaultExpression` instead of skipping those assignments.

The root cause was that the parameter-level presence checks in `BeanMappingMethod.ftl` had no fallback branch. If the source parameter was `null`, the whole block was skipped, including any configured defaults.

This change adds fallback handling for the parameter-level presence check branches in `BeanMappingMethod.ftl`, and adds `isBuiltInPresenceCheckByParameter()` in `BeanMappingMethod.java` so that custom `@SourceParameterCondition` methods keep their existing behavior.

The `_913` fixtures are updated as well. Only the `create` path changes here; `update` / `updateWithReturn` stay unchanged because they are guarded by `existingInstanceMapping`.

Tests cover:
- multi-source mappings with either source parameter `null`
- constructor mappings
- single-source `RETURN_DEFAULT`
- single-source constructor `RETURN_DEFAULT`
- `@SourceParameterCondition` true/false behavior
- the unchanged all-sources-null early return

Closes #2436

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -55,6 +55,8 @@ import org.mapstruct.ap.internal.model.common.Type;
 import org.mapstruct.ap.internal.model.common.TypeFactory;
 import org.mapstruct.ap.internal.model.dependency.GraphAnalyzer;
 import org.mapstruct.ap.internal.model.dependency.GraphAnalyzer.GraphAnalyzerBuilder;
+import org.mapstruct.ap.internal.model.presence.NullPresenceCheck;
+import org.mapstruct.ap.internal.model.presence.OptionalPresenceCheck;
 import org.mapstruct.ap.internal.model.source.BeanMappingOptions;
 import org.mapstruct.ap.internal.model.source.MappingOptions;
 import org.mapstruct.ap.internal.model.source.Method;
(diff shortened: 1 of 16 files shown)


dlwldnjs1009:
Rebased on the latest `main`.

Changes:
- Added regression coverage for empty `Optional` sources, `RETURN_DEFAULT` with `defaultExpression`, and multi-source `@MappingTarget` updates.
- Consolidated the four duplicated fallback blocks in `BeanMappingMethod.ftl` into a single macro.

One behavior to confirm, @filiphr: I kept the existing `@MappingTarget` update behavior. When a source parameter is `null`, the target's existing value is preserved rather than applying a parameter-level default (`updateWithNullSourceKeepsExistingTargetValue`).

Should parameter-level defaults apply to update methods as well, or should this behavior remain unchanged?

Tests: added `Issue2436Test`; existing `_913` generated-source fixture comparisons pass.

PR #4016, opened 2026-03-15
Enforce spaces inside parentheses for control flow statements via checkstyle

filiphr:


hduelme:
In #3994, Claude Code found a missing check for the import order. Maybe we could add this.

After merging, the new rules should also be applied to mapstruct-idea to prevent unexpected errors in future PRs.

filiphr:
Good point. That PR was a trigger for me to do this with Claude Code. I'll look into the import order as well

Issue #4018, opened 2026-03-20
Add built in conversion between URI and String


mk868:
### Use case

This comparison may be helpful, since `URI`s are often used in a similar way to `UUID`s or `URL`s.

PR #4019, opened 2026-03-20
Add URI to String built-in conversions

mk868:
Currently, I'm working on a project that requires mapping the domain's `URI` fields to the API's `String` fields.

This transformation is similar to the built-in mappings between:
* `UUID` <-> `String`
* `URL` <-> `String`

Fixes https://github.com/mapstruct/mapstruct/issues/4018

### DIFF
--- a/documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc
@@ -128,7 +128,7 @@ public interface CarMapper {
 * Between `String` and `StringBuilder`
 
 * Between `java.net.URI` and `String`.
-** When converting from a `String`, the value needs to be a valid https://en.wikipedia.org/wiki/URI[RFC 2396 URI] otherwise a `IllegalArgumentException` is thrown.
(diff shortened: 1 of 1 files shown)


documentation/src/main/asciidoc/chapter-5-data-type-conversions.asciidoc
```diff
@@ -127,6 +127,9 @@ public interface CarMapper {
 
 * Between `String` and `StringBuilder`
 
+* Between `java.net.URI` and `String`.
+** When converting from a `String`, the value needs to be a valid https://en.wikipedia.org/wiki/URI[RFC 2396 URI] otherwise a `IllegalArgumentException` is thrown.
```
filiphr: Grammar and maybe a different link for the URI spec
```suggestion
** When converting from a `String`, the value needs to be a valid https://datatracker.ietf.org/doc/html/rfc2396[RFC 2396 URI] otherwise an `IllegalArgumentException` is thrown.
```

filiphr (changes requested):
Looks good, only one small grammar change for the docs

PR #4020, opened 2026-03-21
refactor TypeFactory.getTypeParameters

hduelme:
Improves readability and avoids unnecessary overhead:
- Stores `typeArguments` in a local variable (since calling `getTypeArguments()` is not cached in eclipse compiler)
- Replace if/else with a ternary expression
- Avoids repeated `getType(...)` calls

No functional changes.


filiphr:
Nice one @hduelme

PR #4021, opened 2026-03-21
use multi-catch in generated code

hduelme:
Since MapStruct requires at least Java 8, multi‑catch can be used in the generated code.
I changed the code in `CommonMacros.ftl` to use multi‑catch when catching more than one exception, instead of generating multiple catch blocks.
To make this consistent, I replaced the duplicated catch code in `Java8FunctionWrapper.ftl `and `LocalVarWrapper.ftl` with the `handleExceptions` macro.


filiphr:
Nice improvement @hduelme. Thanks

PR #4022, opened 2026-03-23
Update maven compile plugin #3972

### LINKED ISSUE #3972: Upgrade maven-compiler-plugin to 3.15.0
The `maven-compiler-plugin` internally is using the `plexus.snapshots` which tries to get eclipse osgi metadata from https://oss.sonatype.org/content/repositories/plexus-snapshots. However, that repository no longer exists, which leads to 

```
[WARNING] Could not transfer metadata org.eclipse.platform:org.eclipse.osgi/maven-metadata.xml from/to plexus.snapshots (https://oss.sonatype.org/content/repositories/plexus-snapshots): status code: 503, reason phrase: Service Unavailable (503)
```

which takes 30 seconds and thus our integration tests are slower. 

For local development to our local `settings.xml` development we can add

```xml
    <mirrors>
        <mirror>
            <id>block-plexus-snapshots</id>
            <mirrorOf>plexus.snapshots</mirrorOf>
            <!-- Use invalid URL like http://localhost to fully block -->
            <url>http://localhost</url>  
        </mirror>
    </mirrors>
```

which would block the repo.

In our GitHub actions we could use https://github.com/s4u/maven-settings-action to disable.

Once the maven-compiler-plugin releases 3.14.2 we can remove the workarounds. See also https://github.com/apache/maven-compiler-plugin/pull/1010

hduelme:
Fixes #3972. [maven-compiler-plugin 3.15.0](https://github.com/apache/maven-compiler-plugin/releases/tag/maven-compiler-plugin-3.15.0) was released and removed the need for the workaround.

I reverted the workaround from #3973 and upgrade the plugin.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -413,7 +413,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-compiler-plugin</artifactId>
-                    <version>3.14.1</version>
+                    <version>3.15.0</version>
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>


filiphr:
Nice, thanks @hduelme

Issue #1018, opened 2016-12-24
Use Diamond operator in generated code


filiphr:
In stead of:

```java
List<Integer> list = new ArrayList<Integer>;
```
We can generate:
```java
List<Integer> list = new ArrayList<>;
```
if we are on Java 7 or higher. I think for this we need to pass whether we are creating a new instance to the `ModelIncludeDirective` which we can then somehow use it in the `Type.ftl`, or this can already be done with the `ext` property?

PR #4026, opened 2026-03-29
use diamond operator for collections and maps on SetterWrapperForCollectionsAndMapsWithNullCheck

hduelme:
In #3994 I found that MapStruct currently generates invalide copy constructors when trying to map from` Collection`/`Map` with `<? extends T>` or` <? super T>` bounds to itself.

I fixed the issue by using the diamond operator for `SetterWrapperForCollectionsAndMapsWithNullCheck` on the constructor. 

I found that the same issue is present in `ExistingInstanceSetterWrapperForCollectionsAndMaps.ftl`. I didn't include that  in this PR, as MapStruct tries to use `addAll(Collection<? extends E> c)`/`putAll(Map<? extends K, ? extends V> m)`. It is impossible to add a super bounded to to an extend bounded one, example
```
if ( extendTypes.getSimpleObjectsCollection() != null ) {
        Collection<? super SimpleObject> collection = types.getSimpleObjectsCollection();
        if ( collection != null ) {
            extendTypes.getSimpleObjectsCollection().clear();
            extendTypes.getSimpleObjectsCollection().addAll( collection ); // Error
        }
    }
    else {
        Collection<? super SimpleObject> collection = types.getSimpleObjectsCollection();
        if ( collection != null ) {
            extendTypes.setSimpleObjectsCollection( new ArrayList<SimpleObject>( collection ) );
        }
    }
}
```
I am not sure what the correct solution would be:

- Only override the values with a copy constructor (linke in the else branch)
- Reject the complete mapping as impossible


filiphr:
@hduelme I went down a rabbit hole while reviewing his one and I came up with #4045, which finally makes #1018 possible. I'll close this one in favor of the new PR. Please feel free to review the PR and let me know what you think

PR #4031, opened 2026-04-11
Use Optional#orElseThrow instead of Optional#get in generated code

### LINKED ISSUE #3976: Use `Optional#orElseThrow` instead of `Optional#get`
Follow up on #674. In some places instead of using `Optional#get` we could use `Optional#orElseThrow` if we are on Java 10 or later.

ayeshakhalid192007-dev:
## Summary

Replace all `Optional#get()` calls in MapStruct-generated mapper code with `Optional#orElseThrow()`.

## Problem

MapStruct generated code used `Optional#get()` to unwrap Optional values. This method is discouraged — it throws `NoSuchElementException` with no message and triggers IDE/linter warnings when called without a preceding `isPresent()` check.

## Changes

Seven source locations were updated:

| File | What changed |
|------|-------------|
| `TypeToOptionalConversion.java` | `<SOURCE>.get()` → `<SOURCE>.orElseThrow()` for non-primitive Optional conversions |
| `OptionalGetWrapper.java` | `+ ".get()"` → `+ ".orElseThrow()"` in `toString()` |
| `OptionalGetWrapper.ftl` | `${assignment}.get()` → `${assignment}.orElseThrow()` |
| `BeanMappingMethod.java` | `+ ".get()"` → `+ ".orElseThrow()"` in parameter binding |
| `BeanMappingMethod.ftl` | 4 occurrences of `.get()` → `.orElseThrow()` for source param reassignment |
| `NestedPropertyMappingMethod.java` | `+ ".get()"` → `+ ".orElseThrow()"` |
| `PropertyMapping.java` | `+ ".get()"` → `+ ".orElseThrow()"` |

8 test snapshot fixtures were updated to match the new generated output.

## Why this is safe

- `Optional#orElseThrow()` (no-arg) is semantically identical to `Optional#get()` — both throw `NoSuchElementException` when the value is absent.
- It has been available since Java 10. MapStruct now requires Java 21, so it is always present.
- All call sites already guard Optional access with presence checks (`isPresent()`, `isEmpty()`, `hasXxx()`), so behavior is unchanged.

## Testing

```
mvn test -pl processor
```

Result: **3467 tests, 0 failures, 0 errors.**

## Files changed

- 7 processor source files (`.java` + `.ftl`)
- 8 test snapshot fixtures

## Related issues

Fixes #3976

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/TypeToOptionalConversion.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/TypeToOptionalConversion.java
@@ -36,11 +36,10 @@ public class TypeToOptionalConversion extends SimpleConversion {
 
     @Override
     protected String getFromExpression(ConversionContext conversionContext) {
-        StringBuilder sb = new StringBuilder("<SOURCE>.get");
         Type optionalBaseType = conversionContext.getSourceType().getOptionalBaseType();
         if ( optionalBaseType.isPrimitive() ) {
-            sb.append( "As" ).append( Strings.capitalize( optionalBaseType.getName() ) );
+            return "<SOURCE>.getAs" + Strings.capitalize( optionalBaseType.getName() ) + "()";
         }
-        return sb.append( "()" ).toString();
+        return "<SOURCE>.orElseThrow()";
     }
 }
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -394,7 +394,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
                         ParameterBinding.fromParameter( parameter ) :
                         ParameterBinding.fromTypeAndName(
                             parameter.getType(),
-                            parameter.getOriginalName() + ".get()"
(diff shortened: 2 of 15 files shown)


filiphr:
>This method is discouraged — it throws `NoSuchElementException`

I think that all the places where we use `get()` are guarded with an `isEmpty()` check. Why is this discouraged?

>It has been available since Java 10. MapStruct now requires Java 21, so it is always present.

Where does it say that MapStruct requires Java 21? MapStruct requires Java 21 to build MapStruct itself. However, MapStruct still is Java 8 compatible in the generated code. With this change, that would no longer be the case.

nealeu:
I've just had a scan of the PR, and Filip's point is proven by the changes.  If you look at the tests, every one of them either calls `Optional.isPresent()` or `Optional.isEmpty()`, so no IDE should be complaining about the generated code.

ayeshakhalid192007-dev:
Thank you for the review, @filiphr and @nealeu.
                                                                                                                                                                                              
You're right on both points. I misread the Java version requirement — the Java 21 build requirement doesn't apply to the generated code, which still needs to be Java 8 compatible. That makes `orElseThrow()` a breaking change for users on Java 8.
I also see now that all call sites are already guarded with presence checks, so there are no actual IDE warnings in the generated output.                                                   
'll close this PR. If the minimum generated-code Java version is ever raised in the future, this change could be revisited then.

PR #4033, opened 2026-04-12
Add support for JSpecify nullness annotations (#1243)

### LINKED ISSUE #1243: A way to explicitly tell which source fields to null check
Currently, when we create an update mapping such as:
`User updateUserFromDto(@MappingTarget User user, UserDto userDto);`

In order to have null checks, we must configure the Mapper annotation (in the class level) to have nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS.

But this generates null checks everywhere, including the creator mappers.
It would be awesome if there was a property in @Mapping like "nullCheck=true/false", so that we can fully control that.

filiphr:
Long overdue, I finally got the time to dig into support for `NonNull` / `Nullable`. I've decided to only support JSpecify for now, since it seems the ecosystem has converged on that.

-----

Detect `@NonNull`, `@Nullable`, `@NullMarked` and `@NullUnmarked` from org.jspecify.annotations to control null check generation in mapping code.

Property-level null checks:
- Source `@NonNull`: skip null check (value guaranteed non-null)
- Target `@NonNull`: always add null check (must protect non-null target)
- All other cases: defer to existing `NullValueCheckStrategy`
- Safety guards (default values, unboxing, NVPMS) are preserved

Method-level source parameter:
- `@NonNull` parameter: skip the method-level null guard

`@NullMarked` / `@NullUnmarked` scope:
- In a `@NullMarked` class or package, unannotated types are effectively `@NonNull`
- `@NullUnmarked` reverts to unknown nullability within its scope
- `@Nullable` on a specific type overrides the scope
- Scope is resolved by walking the enclosing element chain (method -> class -> outer class -> package) and the result is cached on Type.isNullMarked()

Constructor parameters:
- Report a compile error when mapping a potentially nullable source to a @NonNull constructor parameter without a defaultValue, since a null check would leave the variable at null violating the contract
Fixes #1243

---

Would like to get the feedback from the community on this one. 

---

Some open things that could go in followup PRs:

- [ ] Tests with Kotlin nullable / non-null types

### DIFF
--- a/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
@@ -333,6 +333,10 @@ When the source parameter of a mapping method is annotated `@NonNull` (directly 
 
 If a property mapping would assign a potentially `null` source value to a `@NonNull` constructor parameter, MapStruct raises a *compilation error*. Neither inserting a null check (which would leave the variable at `null` and violate the contract) nor passing the value through is safe. Provide a `defaultValue` or `defaultExpression` on the `@Mapping` to satisfy the parameter when the source is absent.
 
+==== Disabling JSpecify support
+
+All JSpecify-driven inference can be turned off by setting the processor option `-Amapstruct.disableJSpecify=true`. With the option enabled, MapStruct ignores every JSpecify annotation (`@NonNull`, `@Nullable`, `@NullMarked`, `@NullUnmarked`) and the compilation error for nullable sources mapped to `@NonNull` constructor parameters is not raised. Null-check generation falls back to the pre-JSpecify behavior driven by `NullValueCheckStrategy`. Use this flag when you need to freeze MapStruct's null-handling behavior while JSpecify annotations exist elsewhere on the classpath.
+
 [[source-presence-check]]
 === Source presence checking
 Some frameworks generate bean properties that have a source presence checker. Often this is in the form of a method `hasXYZ`, `XYZ` being a property on the source bean in a bean mapping method. MapStruct will call this `hasXYZ` instead of performing a `null` check when it finds such `hasXYZ` method.
--- a/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
@@ -299,6 +299,40 @@ The option `nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS` will always 
 
 The strategy works in a hierarchical fashion. `@Mapping#nullValueCheckStrategy` will override `@BeanMapping#nullValueCheckStrategy`, `@BeanMapping#nullValueCheckStrategy` will override `@Mapper#nullValueCheckStrategy` and `@Mapper#nullValueCheckStrategy` will override `@MapperConfig#nullValueCheckStrategy`.
 
+[[jspecify-nullness-annotations]]
+=== JSpecify nullness annotations
+
+MapStruct recognises https://jspecify.dev/[JSpecify] nullness annotations on source and target types and uses them to refine when null checks are generated. Only the annotations from `org.jspecify.annotations` are interpreted; MapStruct does not introduce a compile-time dependency on JSpecify, so consumers opt in by adding the dependency themselves.
+
+The following annotations participate:
+
+* `@NonNull`: the annotated value is guaranteed not to be `null`.
+* `@Nullable`: the annotated value may be `null`.
+* `@NullMarked`: within this scope (package, class, or method), unannotated reference types are treated as `@NonNull`.
+* `@NullUnmarked`: reverts to unknown nullability within its scope (overrides a `@NullMarked` outer scope).
+
+==== Property-level rules
+
+For a single property mapping:
+
+* If the source is `@NonNull`, the null check is skipped, since the value is treated as guaranteed non-null.
+* If the target is `@NonNull` (and the source is not `@NonNull`), a null check is always added so the target's contract is not violated.
+* In all other cases, the existing `NullValueCheckStrategy` decides.
+
+The existing safety guards (`defaultValue` / `defaultExpression`, primitive unboxing, and the relevant `NullValuePropertyMappingStrategy` modes) continue to apply.
+
+==== Scope: `@NullMarked` and `@NullUnmarked`
+
+`@NullMarked` and `@NullUnmarked` are resolved by walking from the annotated element outward: method, then declaring class, then outer classes, then package. The closest annotation wins: a `@NullUnmarked` method inside a `@NullMarked` class reverts only that method, while a `@Nullable` annotation on an individual type always overrides the surrounding scope.
+
+==== Method-level source parameter
+
+When the source parameter of a mapping method is annotated `@NonNull` (directly or via a `@NullMarked` scope), MapStruct skips the method-level null guard, since the caller is contractually obliged to pass a non-null value.
+
+==== Constructor parameter constraint
+
+If a property mapping would assign a potentially `null` source value to a `@NonNull` constructor parameter, MapStruct raises a *compilation error*. Neither inserting a null check (which would leave the variable at `null` and violate the contract) nor passing the value through is safe. Provide a `defaultValue` or `defaultExpression` on the `@Mapping` to satisfy the parameter when the source is absent.
+
 [[source-presence-check]]
 === Source presence checking
 Some frameworks generate bean properties that have a source presence checker. Often this is in the form of a method `hasXYZ`, `XYZ` being a property on the source bean in a bean mapping method. MapStruct will call this `hasXYZ` instead of performing a `null` check when it finds such `hasXYZ` method.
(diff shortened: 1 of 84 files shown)


nealeu:
This looks great @filiphr.  Given the JDK 21 failure, are you going to add a Java 25 (26 too - i.e. last non-LTS) into the CI matrix?

filiphr:
>Given the JDK 21 failure, are you going to add a Java 25 (26 too - i.e. last non-LTS) into the CI matrix?

It's a silly error in the test code. Will be fixed. I'll add Java 25 and 26 separately into the CI matrix, it's not linked to this issue in particular.

nealeu:
I don't want to delay you merging what you have because it's already really valuable and will make a big difference for us.

However, I'm not clear if this does or doesn't cover type use on array elements.  Some permutations of copies between the arrays below should fail.

```java
@NullMarked
class JSpecifyArrays {

  public String @Nullable[] getNullableArrayOfNonNullStrings() { }

  public @Nullable String[] getNonNullArrayOfNullableStrings() { }

  public String[] getNonNullArrayOfNonNullStrings() { }
}
```

filiphr:
Thanks for looking into this @nealeu. I think that second 2 should be covered

```java
  public @Nullable String[] getNonNullArrayOfNullableStrings() { }

  public String[] getNonNullArrayOfNonNullStrings() { }
```

However, I think that the first one would be treated the same as the `getNonNullArrayOfNonNullStrings`.

I'll create another issue for this, as that looks a bit more involved.

vmckinney-cainc:
I would humbly ask that you also consider `@lombok.NonNull`, which is anecdotally very common as well (e.g. [272k]( https://github.com/search?q=%22import+lombok.NonNull%22&type=code) github results for `import lombok.NonNull` vs. [58k](https://github.com/search?q=%22import+org.jspecify.annotations.NonNull%22&type=code) for `import org.jspecify.annotations.NonNull`). Thank you so much for your hard work, I am a big fan of your project!

filiphr:
@vmckinney-cainc I would argue that you are doing an incorrect search for JSpecify. Perhaps a search for `import org.jspecify.annotations` with [260k](https://github.com/search?q=%22import+org.jspecify.annotations%22&type=code) results is more relevant. Especially considering the fact that maybe the most common use of JSpecify I would expect is the use of `@NullMarked` or `@NullUmarked` on `package-info.java` which would be applied to all classes in that package.

However, I'm not the biggest user of JSpecify yet, but what I've seen in the ecosystem is the `package-info.java` usage.

Having said that, I'm not sure that I would like to expand this support to all the other variations of `Nullable` / `NonNull`, especially due to the fact that it looks like JSpecify is becoming the industry standard for this.

rRgbMVg7tO:
First, thank you very much @filiphr for adding support for nullness annotations! This has been a long-expected feature that I’m happy to see.

I would like to support @vmckinney-cainc's feedback on considering additional nullness annotations (other than JSpecify, such as `lombok.NonNull` or `jakarta.annotation.Nonnull`), as this could improve the experience of working with third-party library code.

We have just tried out switching our codebase to the SNAPSHOT version of MapStruct built from this branch and seen new errors where source properties are annotated with non‑JSpecify `@NonNull` annotations (specifically jakarta.annotation.Nonnull) in third‑party code:

> ``Can't map potentially nullable source property "foobar" to `@NonNull` constructor parameter "foobar". Consider adding a defaultValue or defaultExpression.``

Because MapStruct currently only recognizes JSpecify annotations, those sources are treated as potentially nullable. For us, this results in many occurrences across mappers and would require adding defaultValue/defaultExpression for every affected property, even though the sources are annotated by their authors.

While I agree that JSpecify is becoming the industry standard, certain libraries may choose to not migrate to it from other annotation types. Adding support for non-JSpecify annotations would reduce the amount of errors an user of the MapStruct library would need to solve when transitioning to the new MapStruct release with improved nullness support.

Therefore, I would like to propose introducing an annotation processor option (e.g. `mapstruct.nullness.nonNullAnnotations`) to configure the list of considered annotation names. The default would remain JSpecify only, so the behavior is unchanged. For example: 
```text
-Amapstruct.nullness.nonNullAnnotations=org.jspecify.annotations.NonNull,jakarta.annotation.Nonnull,lombok.NonNull
```
This could improve the migration path for projects such as mine that depend on libraries unlikely to adopt JSpecify soon.

If this direction sounds reasonable, we could also discuss adding configurable support for variations of `@NullMarked` (e.g. JSR-305 `@ParametersAreNonnullByDefault`) and `@Nullable` annotations.

Thanks again for your time and for maintaining MapStruct.

pkernevez:
> First, thank you very much @filiphr for adding support for nullness annotations! This has been a long-expected feature that I’m happy to see.
> 
> I would like to support @vmckinney-cainc's feedback on considering additional nullness annotations (other than JSpecify, such as `lombok.NonNull` or `jakarta.annotation.Nonnull`), as this could improve the experience of working with third-party library code.

As lombok is a generation tool, isn't it supposed to generate the JSpecify annotation based on its annotations ?

vmckinney-cainc:
@filiphr

> I would argue that you are doing an incorrect search for JSpecify

Apologies!  Didn't mean to misrepresent the popularity, I'm just not familiar with JSpecify.

> JSpecify is becoming the industry standard

I can say at my company in my little niche that in our corporate git we have 1 single match for `jspecify`.  That being said, MapStruct only supporting JSpecify is another reason for us to consider it.  I understand giving bias towards open-source projects and don't want to challenge that or pressure you or anything.  It's not like closed-source is doing anything to help you :)

@pkernevez

> isn't [lombok] supposed to generate the JSpecify annotation[?]

With [`lombok.addNullAnnotations`](https://projectlombok.org/features/configuration) you can generate `jspecify` nullability annotations, but only for code which lombok generates.  This means that anywhere where we have `@lombok.NonNull` stays `@lombok.NonNull` unless we manually migrate.

vmckinney-cainc:
@rRgbMVg7tO

> Therefore, I would like to propose introducing an annotation processor option

Caveat with regards to Lombok: `@lombok.NonNull` has `CLASS` retention, as opposed to JSpecify and Jakarta's `RUNTIME`. I think there'd have to be some bytecode introspection or something to look towards Lombok. Ugly stuff.

pkernevez:
> > JSpecify is becoming the industry standard
> 
> I can say at my company in my little niche that in our corporate git we have 1 single match for `jspecify`. That being said, MapStruct only supporting JSpecify is another reason for us to consider it. I understand giving bias towards open-source projects and don't want to challenge that or pressure you or anything. It's not like closed-source is doing anything to help you :)

JSpecify is not just yet another nonNull annotation specification. It's a consortium with a lot of major Java actor https://jspecify.dev/about/
Regarding the adoption it's a quite new project (~1 year from its publication) and it's already adopted in open source like Spring, SpringBoot and Hibernate.
It was complex, ~4 years to create 4 annotations (and the behaviors).
Other annotations like JetBrains, lombok or Jakarta, will probably be deprecated.

filiphr:
Thanks @pkernevez for the clarification around JSpecify. 

JSpecify is indeed something that took a really long time to get to, and I think the reason has been that there have been all those big players with the different existing annotations. I think with JSpecify 1.0.0 out there, it would become the standard. Considering the fact that large open source projects like Spring and Hibernate have already adopted them, I think it's a matter of fact until other libraries follow suit. IntelliJ also supports them, Kotlin also supports them for their interoperability for reading Java code from Kotlin. They have [KT-47417](https://youtrack.jetbrains.com/issue/KT-47417) open for the other way arround, i.e. emit JSpecify annotation from Kotlin itself.

@vmckinney-cainc, I'm not sure which other dependencies you use, but I would be surprised if you do not have jspecify on the transitive path if you are on the latest versions of your dependencies.

@rRgbMVg7tO thanks for your comment.

>Because MapStruct currently only recognizes JSpecify annotations, those sources are treated as potentially nullable. For us, this results in many occurrences across mappers and would require adding defaultValue/defaultExpression for every affected property, even though the sources are annotated by their authors.

This was a trigger for me to actually add an opt-out for this using a compiler flag `-Amapstruct.disableJSpecify=true`. 

>While I agree that JSpecify is becoming the industry standard, certain libraries may choose to not migrate to it from other annotation types.

To be honest I would rather spend my time in helping those libraries get to JSpecify then spend the time on supporting the different flavors and quirks of the other nullability annoations.

>If this direction sounds reasonable, we could also discuss adding configurable support for variations of `@NullMarked` (e.g. JSR-305 `@ParametersAreNonnullByDefault`) and `@Nullable` annotations.

That's the pandora's box I do not want to open.


@vmckinney-cainc not sure what you mean with
>Caveat with regards to Lombok: `@lombok.NonNull` has `CLASS` retention, as opposed to JSpecify and Jakarta's `RUNTIME`. I think there'd have to be some bytecode introspection or something to look towards Lombok. Ugly stuff.

If you are thinking that it would be hard for us to read `CLASS` retention annotation, then no need to think about that. We run during the compilation phase, which means that we can read `SOURCE` (only for your own code in the same compilation module), `CLASS` and `RUNTIME` annotations.

---
Thanks everyone for your feedback. It is really valuable. However, my decision for this is final. I really do not want to spend time on supporting that the major Java actors have converged on. I would merge this PR in this stance and I would try to do a 1.7.0.Beta2 release soonish so that the community can already start trying it out.

vmckinney-cainc:
Thanks for all you do @filiphr!!  I am very happy to have this feature in any form :)

nealeu:
@filiphr Are you planning a beta.2 release before final 1.7.0?  I think it would be good for this to get a more widely visible release.  
[Edit: I just spotted/remembered your comment above, so this is a gentle nudge ;-) ]

PR #4039, opened 2026-04-19
Add JDK 25 and 26 to workflow matrix, drop integration testing on JDK 11

filiphr:
Upgrade Lombok to 1.18.38 since older versions fail with ExceptionInInitializerError on com.sun.tools.javac.code.TypeTag when compiling under JDK 25/26.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [21]
+        java: [21, 25, 26]
     name: 'Linux JDK ${{ matrix.java }}'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


nealeu:
Is the reason for dropping 11 but not 8 that 8 is still LTS but not 11?

filiphr:
Yes indeed. From what I could see the last Openjdk 11 release was last year, whereas the last 8 was this year in January. I also don't foresee that there would be problems on JDK 11 if something is running on JDK 8.

I addition to all of that, I believe that JDK 11 is not used that much. Most of the ecosystem has adopted 17 as a baseline. We are still 8 as a baseline.

nealeu:
Yeah. Great to cut build minutes where they don't deliver value :-)

PR #4040, opened 2026-04-21
Fix #1509: Discover existing mappers instead of providing 'Mapper.uses' class array

### LINKED ISSUE #1509: IMPROVEMENT: Discover existing mappers instead of providing "Mapper.uses" class array
Would it be possible to enable dynamic discovery of the mapper classes used in a generated code  according to the source and target class through the package annotation configuration like:
@Mapper (usesPackage = {"org.company.componenent.mappers"})

UtkarshSinghChauhan:
This PR fixes issue #1509 by introducing a new attribute 'usesPackage' in the 'Mapper' annotation, which allows users to specify a package where MapStruct can discover existing mappers. The library will then scan the specified package and its sub-packages for classes annotated with 'Mapper' and use them as mappers.

The fix includes a new test case to verify that MapStruct can discover and use existing mappers in the specified package.

Fixes #1509


nealeu:
@UtkarshSinghChauhan Can you check what you submitted.  It breaks things and doesn't fix anything.

UtkarshSinghChauhan:
So sorry i mixed up some things i am really srry

PR #4043, opened 2026-04-24
#4042 Specify OpenJDK 21 for Jitpack build

### LINKED ISSUE #4042: Jitpack builds failing due to JDK 8 not allowing checkstyle dep using more recent version
I notice we have a [class file version error](https://jitpack.io/com/github/mapstruct/mapstruct/1.7.0.Beta1/build.log) on Jitpack so we can't just try out main or PRs there.

I'll look into a PR to fix Jitpack in line with GH build config

nealeu:
Existing default of JDK 8 due to a maven plugin dependency having a post-Java 8 JAR version

Successful build of this branch: https://jitpack.io/com/github/nealeu/mapstruct/gh-4042-jitpack-jdk21-facedb548d-1/build.log

Fixes: #4042

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,0 +1,2 @@
+jdk:
+    - openjdk21


filiphr:
Thanks @nealeu

nealeu:
Build runs with this merged, but it looks like Jitpack are doing something weird as I get
```
Could not GET 'https://jitpack.io/com/github/mapstruct/mapstruct-processor/9856a425ba/mapstruct-processor-9856a425ba.pom'. 
Received status code 401 from server: Unauthorized
```
It would be good to know if this applies to everyone as it could be something I should look at further.

PR #4044, opened 2026-04-25
fix: make `URLToStringConversion` generate `URI.create(String).toURL()` instead of `new URL(String)`

### LINKED ISSUE #4041: `URL` conversion uses deprecated constructor `new URL(String)` in Java 20+
### Expected behavior

In Java 20 and later, the `URL(String)` constructor has been deprecated in favor of using `URI` to handle the parsing logic:

```java
if ( url != null ) {
    url1 = URI.create( url ).toURL();
}
```

### Actual behavior

Currently, MapStruct generates code using the deprecated constructor when mapping a `String` to a `java.net.URL`:

```java
if ( url != null ) {
    url1 = new URL( url );
}
```

1. Use Java 20 or higher.
2. Define a mapper that converts a `String` to a `java.net.URL` (or a record/bean containing a `URL`).
3. Observe the generated implementation.

Example:

```java
@Mapper
public interface ProductMapper {
    Product map(String url) throws MalformedURLException;

    record Product(URL url) {}
}
```

Generated implementation:

```java
@Generated(
    date = "2026-04-23T12:14:28+0300",
    comments = "version: 1.6.3, compiler: IncrementalProcessingEnvironment from gradle-language-java-9.4.1.jar, environment: Java 25.0.1 (Eclipse Adoptium)"
)
public class ProductMapperImpl implements ProductMapper {

    @Override
    public ProductMapper.Product map(String url) throws MalformedURLException {
        if ( url == null ) {
            return null;
        }

        URL url1 = null;

        if ( url != null ) {
            url1 = new URL( url );
        }

        ProductMapper.Product product = new ProductMapper.Product( url1 );

        return product;
    }
}
```

Compiling the mapper above will produce the following message in the console output:

```
Note: C:\Users\Yevhen\IdeaProjects\gradle-mapstruct-demo\build\generated\sources\annotationProcessor\java\main\io\github\yvasyliev\mapstruct\ProductMapperImpl.java uses or overrides a deprecated API.
```

### MapStruct Version

`1.6.3`

yvasyliev:
This pull request refactors the handling of URL conversions in the codebase, standardizing the approach by using `URI.create(...).toURL()` instead of direct `new URL(...)` construction. This change affects both the internal conversion logic and the related tests, aiming for improved consistency and reliability in URL handling.

**Refactoring of URL Conversion Logic:**

* The `ConversionUtils.url` method and related imports of `java.net.URL` were removed, and all references were updated to use `URI` and the `ConversionUtils.uri` method instead.
* In `URLToStringConversion.java`, the conversion now constructs URLs using `URI.create(<SOURCE>).toURL()` rather than `new URL(<SOURCE>)`. This also updates the import types from `URL` to `URI`.

**Test Adjustments for New Conversion Approach:**

* In `URLConversionTest.java`, all usages of `new URL(...)` were replaced with `URI.create(...).toURL()` to match the new conversion logic.
* Test cases for invalid URL strings were updated to use malformed URI schemes (e.g., `"xxxxxx://mapstruct.org/"`) instead of arbitrary invalid strings, ensuring the tests accurately reflect the new conversion path.

Closes #4041

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/ConversionUtils.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/ConversionUtils.java
@@ -8,7 +8,6 @@ package org.mapstruct.ap.internal.conversion;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.net.URI;
-import java.net.URL;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.text.DecimalFormat;
@@ -279,17 +278,6 @@ public final class ConversionUtils {
      */
     public static String uri(ConversionContext conversionContext) {
         return typeReferenceName( conversionContext, URI.class );
-    }
-
-    /**
-     * Name for {@link java.net.URL}.
-     *
-     * @param conversionContext Conversion context
-     *
-     * @return Name or fully-qualified name.
-     */
-    public static String url(ConversionContext conversionContext) {
-        return typeReferenceName( conversionContext, URL.class );
     }
 
     /**
--- a/processor/src/main/java/org/mapstruct/ap/internal/conversion/URLToStringConversion.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/conversion/URLToStringConversion.java
@@ -6,14 +6,14 @@
 package org.mapstruct.ap.internal.conversion;
 
 import java.net.MalformedURLException;
-import java.net.URL;
+import java.net.URI;
 import java.util.List;
 import java.util.Set;
 
 import org.mapstruct.ap.internal.model.common.ConversionContext;
 import org.mapstruct.ap.internal.model.common.Type;
 
-import static org.mapstruct.ap.internal.conversion.ConversionUtils.url;
+import static org.mapstruct.ap.internal.conversion.ConversionUtils.uri;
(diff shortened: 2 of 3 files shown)


hduelme:
@yvasyliev In general, I agree that `URI.create(...).toURL()` should be preferred. However, changing the current implementation has some implications. Instead of always throwing a `MalformedURLException`, it may now throw an`IllegalArgumentException`. Additionally, `URL` would no longer be imported automatically, which could break existing expression blocks.

We'll discuss internally whether to include this change in the upcoming release.
@filiphr, what do you think?

filiphr:
@hduelme I think that it is actually fine to do this. If using the constructor is not the recommended way of creating URLs then we should not stand in the way. We should only mention this in our release notes.

>However, changing the current implementation has some implications. Instead of always throwing a `MalformedURLException`, it may now throw an `IllegalArgumentException`

People will always have the option of writing their own mapping if they want to keep this behaviour

>Additionally, `URL` would no longer be imported automatically, which could break existing expression blocks.

That's completely fine @hduelme. People should not be relying on how we generate the code for their expressions. We also have `Mapper#imports` that people can use to import what they need always.

So I would say that we should go ahead and merge this

hduelme (approved):

PR #4045, opened 2026-05-01
#1018 Use diamond operator for `new` expressions in generated code

### LINKED ISSUE #1018: Use Diamond operator in generated code

filiphr:
Adds a `NewInstanceCreation` ModelElement that renders `new T<>` and contributes only the raw class to imports, so type parameters used solely in the new-expression are no longer orphaned (Issue #955). Also fixes #3994's follow-up wildcard cases where explicit type arguments produced uncompilable Java.

Fixes #1018.

@hduelme I went down a rabbit hole with Claude Code while reviewing your PR #4026, so I finally have #1018. Let me know what you think

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
@@ -46,6 +46,7 @@ import org.mapstruct.ap.internal.model.beanmapping.TargetReference;
 import org.mapstruct.ap.internal.model.common.Assignment;
 import org.mapstruct.ap.internal.model.common.BuilderType;
 import org.mapstruct.ap.internal.model.common.FormattingParameters;
+import org.mapstruct.ap.internal.model.common.NewInstanceCreation;
 import org.mapstruct.ap.internal.model.common.Parameter;
 import org.mapstruct.ap.internal.model.common.ParameterBinding;
 import org.mapstruct.ap.internal.model.common.PresenceCheck;
@@ -97,6 +98,7 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
     private final List<PropertyMapping> constructorConstantMappings;
     private final List<SubclassMapping> subclassMappings;
     private final Type returnTypeToConstruct;
+    private final NewInstanceCreation newInstance;
     private final BuilderType returnTypeBuilder;
     private final MethodReference finalizerMethod;
     private final String finalizedResultName;
@@ -2184,6 +2186,9 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
             }
         }
         this.returnTypeToConstruct = returnTypeToConstruct;
+        this.newInstance = ( returnTypeToConstruct != null && getFactoryMethod() == null )
+            ? NewInstanceCreation.forType( returnTypeToConstruct )
+            : null;
         this.subclassMappings = subclassMappings;
         this.sourceParametersReassignments = sourceParametersReassignments;
     }
@@ -2246,6 +2251,10 @@ public class BeanMappingMethod extends NormalTypeMappingMethod {
         return returnTypeToConstruct;
     }
 
+    public NewInstanceCreation getNewInstance() {
+        return newInstance;
+    }
+
     public boolean hasSubclassMappings() {
         return !subclassMappings.isEmpty();
     }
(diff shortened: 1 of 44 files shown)


processor/src/main/java/org/mapstruct/ap/internal/model/BeanMappingMethod.java
```diff
@@ -2279,7 +2288,12 @@ public Set<Type> getImportTypes() {
         }
 
         if ( returnTypeToConstruct != null  ) {
-            types.addAll( returnTypeToConstruct.getImportTypes() );
+            if ( newInstance != null ) {
+                types.addAll( newInstance.getImportTypes() );
+            }
+            else {
+                types.addAll( returnTypeToConstruct.getImportTypes() );
+            }
```
hduelme: Could be written with a ternary expression like 
` types.addAll( newInstance != null ? newInstance.getImportTypes() : returnTypeToConstruct.getImportTypes() );`
or even shorter
`types.addAll( (newInstance != null ? newInstance : returnTypeToConstruct).getImportTypes() );`

processor/src/main/java/org/mapstruct/ap/internal/model/common/NewInstanceCreation.java
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @param targetType the target type to be instantiated; must not be {@code null}
+     * @return a new model
+     */
+    public static NewInstanceCreation forType(Type targetType) {
+        Type effective = targetType.getImplementationType() != null
+            ? targetType.getImplementationType()
+            : targetType;
+        return new NewInstanceCreation( effective );
+    }
```
hduelme, line 45: A check could be added to verify whether an effective target can be constructed.

This isn’t a new issue—it existed before this change—but this might be a good place to catch or handle it.

hduelme (commented):
@filiphr this is a really nice solution. Handling new instance creation in one place is definitely the better approach.

I found one small potential improvement: Simplification by using a ternary expression.

It might also be worth adding a safeguard/check to ensure that an effective target can always be constructed. This isn’t introduced by this change (it existed before), but this could be a good place to catch or handle it more explicitly.
for example, attempting to instantiate an interface or a type without an accessible constructor:

```
import org.mapstruct.Mapper;
import org.mapstruct.NullValuePropertyMappingStrategy;

@Mapper
public interface EasyMap {

    class To {
        private Comparable Comparable;

        public Comparable getComparable() {
            return Comparable;
        }

        public void setComparable(Comparable comparable) {
            Comparable = comparable;
        }
    }

    class Fromm {
        private Comparable Comparable;

        public Comparable getComparable() {
            return Comparable;
        }

        public void setComparable(Comparable comparable) {
            Comparable = comparable;
        }
    }

    @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
    void update(@MappingTarget To to, Fromm from);
}
```
 generates:
```

@Generated(
    date = "2026-05-04T21:34:26+0200",
    comments = "version: 1.6.3, compiler: javac, environment: Java 21.0.5 (Eclipse Adoptium)"
)
public class EasyMapImpl implements EasyMap {

    @Override
    public void update(To to, Fromm from) {
        if ( from == null ) {
            return;
        }

        if ( from.getComparable() != null ) {
            to.setComparable( from.getComparable() );
        }
        else {
            to.setComparable( new Comparable() );
        }
    }
}

```

processor/src/main/java/org/mapstruct/ap/internal/model/common/NewInstanceCreation.java
filiphr: This is a nice catch. However, I do not believe that this is the right place for this, we need to do this way earlier in order to provide a proper error message, like the line of the `@Mapping`

filiphr:
Thanks for the review @hduelme, I would apply the simplification. Regarding the check you mentioned, yes it would be good to add that, but I am not sure that this new place is the right place to do it. I would expose this to be done way earlier in order to provide a proper error message like the line number.

PR #4050, opened 2026-05-09
add support for generics in Arrays and fix initialization of multidimensional arrays

hduelme:
While reviewing #4045 and looking for places where objects are initialized, I came across generics with arrays.

Bug fixes included in this PR:

1. Creation of arrays with generics
Turns out generics are not allowed on an array constructor. MapStruct generates:
`return new CustomerGenerics<String>[0];`
I centralized array initialization with size handling in  `constructArrayType` nd removed the generics there. The same issue existed when mapping Streams to arrays.

2. Missing generics on method creation
MapStruct currently drops generics on generated methods.
From:
`CustomerGenerics<String>[] array(CustomerGenerics<String>[] a);`
it generates:
` public CustomerGenerics[] array(CustomerGenerics[] a) {`
I fixed this by reading the generics from the first component type that is not an array. This also required implementing `withoutBounds()` and` replaceSuperBoundWith()` for arrays

3. Wrong creation of default arrays for multidimensional arrays
Currently MapStruct writes the size in the last brackets, but it needs to be in the first:
`return new byte[][0];`
4. Multidimensional arrays do not work correctly with generics or varargs
I fixed the implementation in `Type.ftl `: first the base type, then the generics, then the array dimensions, and finally the last array or varargs declaration.

I added tests for my changes to the existing test suite. It might make sense to move `Issue895Test` to the array package, since it directly tests whether multidimensional arrays work correctly.


filiphr:
@hduelme nice catch and nice fix. I did a small polish, have a look and let me know if you don't agree with it

hduelme:
@filiphr I agree, your polished version shows the difference between array types and none-array types much more clearly. The missing `s` was a copy-paste error on my part.

Issue #3961, opened 2025-12-04
Investigate using JUnit JUpter InvocationInterceptor for our testing infrastructure


filiphr:
JUnit Jupiter has an `InvocationInterceptor` that can be used for class path override. We can look into the Spring Boot [`ModifiedClassPathExtension`](https://github.com/spring-projects/spring-boot/blob/main/test-support/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/classpath/ModifiedClassPathExtension.java#L47) for a potential inspiration.

Also investigate the comments from https://github.com/junit-team/junit-framework/issues/201

filiphr:
Closing this since @hduelme already did an investigation of this and we have now a different approach in #4051

PR #4051, opened 2026-05-10
improve class loading for tests

hduelme:
I improved class loading for tests by replacing `LauncherDiscoveryListener` with `CompilerLauncherInterceptor` (introduced in junit-team/junit-framework#3091). This change reduces overhead during test discovery and execution and also enables proper lifecycle handling, including correct closing of `ClassLoader` instances.

I also attempted to implement #3961, but was unable to get it working. The InvocationInterceptor only allows interception at test execution time, which is too late in this case: by then, test classes have already been loaded. This leads to `IllegalAccessError` issues when multiple `ClassLoader` instances attempt to load the same package, which is not permitted.


filiphr (commented):
This looks really really good @hduelme. From what I can see it halves the time. From around 12 minutes to around 6 minutes for all the tests.

I need to find some time to have a quick look around how we want to store the private fields and to understand hot the `CompilerLauncherInterceptor` is instantiated.

If I look at the example from the docs:

https://docs.junit.org/6.0.3/advanced-topics/launcher-api.html#launcher-interceptors-custom

the custom classloader is instantiated in the constructor, and the original one is reset in the `intercept`. However, it could be that ours is needed later and you did all the heavy lifting checks, but I want to play with it a bit.

hduelme:
@filiphr 
My implementation took inspiration from an old PR of quarkus [34681](https://github.com/quarkusio/quarkus/pull/34681/files#diff-45317145bffbd2f36f772a1709921b8bc83d70aa8a73f642a0145db8d26b501e).

I did some experiments myself and restoring the ClassLoader to `originalClassLoader` in the finally block slows down the tests noticeably.
Ensuring that only a single instance of `ModifiableURLClassLoader` is created looks necessary for me. I moved the instantiation of the custom `ClassLoader` into the constructor.

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilerLauncherInterceptor.java
```diff
@@ -0,0 +1,54 @@
...
+        newClassLoader.withOriginOf( getClass() );
+    }
+
+    @Override
+    public <T> T intercept(Invocation<T> invocation) {
+        orginalClassLoader = Thread.currentThread().getContextClassLoader();
+        if ( !orginalClassLoader.getClass().isAssignableFrom( ModifiableURLClassLoader.class ) ) {
+            Thread.currentThread().setContextClassLoader( newClassLoader );
+        }
+        return invocation.proceed();
```
filiphr: This looks a bit weird. Looking at this we end up with the `originalClassLoader` becoming an instance of the `ModifiableURLClassLoader`.  Shouldn't we only set the `originalClassLoader` if it is not assignable from `ModifiableURLClassLoader`? i.e. if we actually change the context class loader?

filiphr (commented):
I just tried this out locally, it's almost 3x speedup locally, that's just amazing @hduelme. I've left a comment, I think that we are losing the correct `originalClassLoader`.

processor/src/test/java/org/mapstruct/ap/testutil/runner/CompilerLauncherInterceptor.java
hduelme: Good catch. That was not what should happen. I changed the implementation as you suggested.

hduelme:
@filiphr Good catch. That was not what should happen. That's should be fixed now

filiphr:
Amazing. Thanks @hduelme.

By the way, when you can, can you reach out to me by email? Your commits don't have an email I can use to reach out, that's why I am asking you to reach out to me 😀

PR #4052, opened 2026-05-11
registerAsParallelCapable for ModifiableURLClassLoader directly

hduelme:
Since Java Version 1.7 `ClassLoader.registerAsParallelCapable()` is alwa available.  Reflection is not needed anymore.


filiphr:
Nice one @hduelme. Thanks

PR #4053, opened 2026-05-11
remove reflection from isDefaultMethod

hduelme:
Since Java 8, `ExecutableElement.isDefault()` is always available. As MapStruct now requires Java 8 as the minimum supported version, the reflective check is no longer necessary.


filiphr:
Nice one @hduelme

PR #4061, opened 2026-06-03
#4060 Report compilation error for SET_TO_DEFAULT without accessible no-args constructor

### LINKED ISSUE #4060: Missing accessable Constrctor check for updating existing bean
### Expected behavior

MapStruct should report a compilation error during annotation processing when `SET_TO_DEFAULT` requires instantiation of a type that cannot be constructed through an accessible no-args constructor, rather than generating invalid code.
Maybe something like  https://github.com/mapstruct/mapstruct/blob/453602e3fa1de5e1f5717c3d9e59a23aa76aa7f5/processor/src/main/java/org/mapstruct/ap/internal/util/Message.java#L167

### Actual behavior

When updating an existing bean with `NullValuePropertyMappingStrategy.SET_TO_DEFAULT`, MapStruct generates code that attempts to instantiate target property types using a no-args constructor.

However, for types that do not have an accessible no-args constructor (e.g. `Comparable`, `LocalDa`te, `LocalDateTime`, `Instant`, `BigDecimal`, and `BigInteger`), MapStruct still generates instantiation code such as:

```
target.localDate = new LocalDate();
target.comparable = new Comparable<String>();
```

### Example
```
import java.math.BigDecimal;
import java.math.BigInteger;
import java.time.Instant;
import java.time.LocalDate;
import java.time.LocalDateTime;

import org.mapstruct.Mapper;
import org.mapstruct.NullValuePropertyMappingStrategy;

@Mapper
public interface UpdateMapper {

    class Target {
        public Comparable<String> comparable;
        public LocalDate localDate;
        public LocalDateTime localDateTime;
        public Instant instant;
        public BigDecimal bigDecimal;
        public BigInteger bigInteger;

    }

    class Source  {
        public Comparable<String> comparable;
        public LocalDate localDate;
        public LocalDateTime localDateTime;
        public Instant instant;
        public BigDecimal bigDecimal;
        public BigInteger bigInteger;
    }

    @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
    void update(@MappingTarget Target target, Source from);
}
```
generates

```

@Generated(
    date = "2026-06-02T22:45:28+0200",
    comments = "version: 1.6.3, compiler: javac, environment: Java 21.0.5 (Eclipse Adoptium)"
)
public class UpdateMapperImpl implements UpdateMapper {

    @Override
    public void update(Target target, Source from) {
        if ( from == null ) {
            return;
        }

        if ( from.comparable != null ) {
            target.comparable = from.comparable;
        }
        else {
            target.comparable = new Comparable<String>();
        }
        if ( from.localDate != null ) {
            target.localDate = from.localDate;
        }
        else {
            target.localDate = new LocalDate();
        }
        if ( from.localDateTime != null ) {
            target.localDateTime = from.localDateTime;
        }
        else {
            target.localDateTime = new LocalDateTime();
        }
        if ( from.instant != null ) {
            target.instant = from.instant;
        }
        else {
            target.instant = new Instant();
        }
        if ( from.bigDecimal != null ) {
            target.bigDecimal = from.bigDecimal;
        }
        else {
            target.bigDecimal = new BigDecimal();
        }
        if ( from.bigInteger != null ) {
            target.bigInteger = from.bigInteger;
        }
        else {
            target.bigInteger = new BigInteger();
        }
    }
}
```
None of the targets has an accessible constructor.

### MapStruct Version

MapStruct 1.6.3

seonwooj0810:
Fixes #4060

### Problem

When updating an existing bean with `NullValuePropertyMappingStrategy.SET_TO_DEFAULT`, MapStruct resets a `null` source property by creating a fresh default instance via the parameterless constructor (`new Target()`). For target property types that do **not** expose an accessible no-args constructor (e.g. `LocalDate`, `LocalDateTime`, `Instant`, `BigDecimal`, `BigInteger`, `Comparable`), this generated uncompilable code such as:

```java
target.setLocalDate( new LocalDate() );
```

As requested in the issue, MapStruct should report a compilation error during annotation processing rather than emitting code that fails to compile.

### Change

- `Type#hasAccessibleParameterlessConstructor()` — new helper that returns `true` only when `new Type()` would actually compile (an implicit default constructor, or a declared non-private no-args constructor). Unlike the existing `hasAccessibleConstructor()`, it returns `false` for types that only expose parameterized constructors (e.g. `BigDecimal`).
- `PropertyMapping#assignToPlainViaSetter(...)` — for the `SET_TO_DEFAULT` case, when the default instance would be built through a parameterless constructor (no factory/builder) and the target type lacks an accessible one, report `GENERAL_NO_SUITABLE_CONSTRUCTOR` (the message referenced in the issue). Types that are initialized differently — those with an implementation type (collections/maps), arrays, `Optional`, or a sensible default — are intentionally excluded.

### Test evidence

Added `org.mapstruct.ap.test.bugs._4060.Issue4060Test`:
- `setToDefaultWithoutParameterlessConstructorFails` — asserts the expected compilation errors for `LocalDate`, `BigDecimal` and `Comparable<String>`.
- `setToDefaultWithParameterlessConstructorResetsToNewInstance` — verifies the existing reset-to-new-instance behaviour still works for constructible types.

Local run (JDK toolchain, `processor` module):
- `Issue4060Test` — 4 tests, 0 failures.
- Related regression tests pass: `Issue1790Test`, `Issue3884Test`, `JSpecifySafetyGuardTest`, `NullValuePropertyMappingTest`, `NullValuePropertyMappingClearTest`.
- `checkstyle:check` on the `processor` module passes.

### Verification done
1. No in-flight PR (`gh pr list --search` for 4060 / SET_TO_DEFAULT constructor) and no self-claim comments on the issue.
3. Fix touches `.java` files only.
4. Reproduced the bug against `main` (`453602e`): without the change the erroneous mapper compiles and emits `new LocalDate()`; with the change the documented compilation error is raised.
5. No parent epic — issue is standalone and labelled `bug`, `hack.commit.push`.

### DIFF
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_4060/ErroneousSetToDefaultMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_4060/ErroneousSetToDefaultMapper.java
@@ -52,6 +52,37 @@ public interface ErroneousSetToDefaultMapper {
         }
     }
 
+    abstract class AbstractValue {
+    }
+
+    class WithAbstractValue {
+        private AbstractValue value;
+
+        public AbstractValue getValue() {
+            return value;
+        }
+
+        public void setValue(AbstractValue value) {
+            this.value = value;
+        }
+    }
+
+    enum EnumValue {
+        FIRST
+    }
+
+    class WithEnumValue {
+        private EnumValue value;
+
+        public EnumValue getValue() {
+            return value;
+        }
+
+        public void setValue(EnumValue value) {
+            this.value = value;
+        }
+    }
+
     @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
     void updateLocalDate(@MappingTarget WithLocalDate target, WithLocalDate source);
 
@@ -60,4 +91,10 @@ public interface ErroneousSetToDefaultMapper {
 
     @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
     void updateComparable(@MappingTarget WithComparable target, WithComparable source);
+
+    @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
+    void updateAbstractValue(@MappingTarget WithAbstractValue target, WithAbstractValue source);
+
+    @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
+    void updateEnumValue(@MappingTarget WithEnumValue target, WithEnumValue source);
 }
--- a/processor/src/test/java/org/mapstruct/ap/test/bugs/_4060/ErroneousSetToDefaultMapper.java
+++ b/processor/src/test/java/org/mapstruct/ap/test/bugs/_4060/ErroneousSetToDefaultMapper.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct.ap.test.bugs._4060;
+
+import java.math.BigDecimal;
+import java.time.LocalDate;
+
+import org.mapstruct.BeanMapping;
+import org.mapstruct.Mapper;
+import org.mapstruct.MappingTarget;
+import org.mapstruct.NullValuePropertyMappingStrategy;
+
+@Mapper
+public interface ErroneousSetToDefaultMapper {
+
+    class WithLocalDate {
+        private LocalDate value;
+
+        public LocalDate getValue() {
+            return value;
+        }
+
+        public void setValue(LocalDate value) {
+            this.value = value;
+        }
+    }
+
+    class WithBigDecimal {
+        private BigDecimal value;
+
+        public BigDecimal getValue() {
+            return value;
+        }
+
+        public void setValue(BigDecimal value) {
+            this.value = value;
+        }
+    }
+
+    class WithComparable {
+        private Comparable<String> value;
+
+        public Comparable<String> getValue() {
+            return value;
+        }
+
+        public void setValue(Comparable<String> value) {
+            this.value = value;
+        }
+    }
+
+    @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
+    void updateLocalDate(@MappingTarget WithLocalDate target, WithLocalDate source);
+
+    @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
+    void updateBigDecimal(@MappingTarget WithBigDecimal target, WithBigDecimal source);
+
+    @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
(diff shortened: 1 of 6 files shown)


processor/src/test/java/org/mapstruct/ap/test/bugs/_4060/Issue4060Test.java
```diff
@@ -0,0 +1,57 @@
...
+                line = 56,
+                message = "LocalDate does not have an accessible constructor."),
+            @Diagnostic(type = ErroneousSetToDefaultMapper.class,
+                kind = javax.tools.Diagnostic.Kind.ERROR,
+                line = 59,
+                message = "BigDecimal does not have an accessible constructor."),
+            @Diagnostic(type = ErroneousSetToDefaultMapper.class,
+                kind = javax.tools.Diagnostic.Kind.ERROR,
+                line = 62,
+                message = "Comparable<String> does not have an accessible constructor.")
```
hduelme: I know I originally suggested the message, but after reconsidering it, I think a message like:

> LocalDate does not have an accessible parameterless constructor. Either change the nullValuePropertyMappingStrategy or define a defaultValue or a defaultExpression.

would give a clearer indication of the actual problem and suggest possible ways to fix it.

processor/src/test/java/org/mapstruct/ap/test/bugs/_4060/ErroneousSetToDefaultMapper.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
+    void updateLocalDate(@MappingTarget WithLocalDate target, WithLocalDate source);
+
+    @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
+    void updateBigDecimal(@MappingTarget WithBigDecimal target, WithBigDecimal source);
+
+    @BeanMapping(nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.SET_TO_DEFAULT)
+    void updateComparable(@MappingTarget WithComparable target, WithComparable source);
+}
```
hduelme, line 100: You correctly added the check for abstract and enum types. Could you add the corresponding tests as well?

hduelme (changes requested):
@seonwooj0810 nice implementation.
I left two comments: one regarding the error message and one suggesting a small test improvement.

processor/src/test/java/org/mapstruct/ap/test/bugs/_4060/Issue4060Test.java
seonwooj0810: Good suggestion — done in f8e2c9f0. Since `GENERAL_NO_SUITABLE_CONSTRUCTOR` is shared with other call sites (e.g. `BeanMappingMethod`), I introduced a dedicated `PROPERTYMAPPING_NO_ACCESSIBLE_PARAMETERLESS_CONSTRUCTOR` message with the wording you suggested: "%s does not have an accessible parameterless constructor. Either change the nullValuePropertyMappingStrategy or define a defaultValue or a defaultExpression."

processor/src/test/java/org/mapstruct/ap/test/bugs/_4060/ErroneousSetToDefaultMapper.java
seonwooj0810: Added abstract class and enum target property cases to `ErroneousSetToDefaultMapper` in f8e2c9f0 — both now verify the new compilation error is reported.

hduelme (approved):

hduelme:
@seonwooj0810 Thanks for making the adjustments, and thanks again for the contribution!

PR #4064, opened 2026-06-05
Fix #4030 remove double getters in null checks

### LINKED ISSUE #4030: Getters can be computed twice when nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS
### Expected behavior

We may have getters with complex logic within, so we would want that getter to be computed only once but it can be computed twice, one for checking the nullability and other time for assigning the value. We may want to store the getter value in a variable, check for nullability and then assign the value.

### Actual behavior

_No response_

NA

### MapStruct Version

1.6.3

stefanos-kalantzis:
fixes #4030 

Note: There is an edge-case considered out of scope for this PR, where source is used multiple times for different mappings. Then the generated code has multiple blocks with a different extracted variable, with incremental names. For example in `DomainDtoWithNcvsAlwaysMapperImpl`:

   ```java
   List<String> strings = source.getStrings();
   if ( target.getStrings() != null ) {
       if ( source.hasStrings() ) {
           target.getStrings().clear();
           target.getStrings().addAll( strings );
       }
   }
   else {
       if ( source.hasStrings() ) {
           List<String> list = strings;
           target.setStrings( new LinkedHashSet<>( list ) );
       }
   }
   List<String> strings1 = source.getStrings();
   if ( target.getLongs() != null ) {
       if ( source.hasStrings() ) {
           target.getLongs().clear();
           target.getLongs().addAll( stringListToLongSet( strings1 ) );
       }
   }
   else {
       if ( source.hasStrings() ) {
           target.setLongs( stringListToLongSet( strings1 ) );
       }
   }
   ```

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
@@ -37,7 +37,7 @@ import static org.mapstruct.ap.internal.gem.NullValuePropertyMappingStrategyGem.
 
 /**
  * A builder that is used for creating an assignment to a collection.
- *
+ * <p>
  * The created assignments to the following null checks:
  * <ul>
  *     <li>source-null-check - For this the {@link SetterWrapperForCollectionsAndMapsWithNullCheck} is used when a
@@ -49,7 +49,7 @@ import static org.mapstruct.ap.internal.gem.NullValuePropertyMappingStrategyGem.
  *     <li>local-var-null-check - Done in {@link ExistingInstanceSetterWrapperForCollectionsAndMaps}, and
  *     {@link SetterWrapperForCollectionsAndMapsWithNullCheck}</li>
  * </ul>
- *
+ * <p>
  * A local-var-null-check is needed in the following cases:
  *
  * <ul>
@@ -106,7 +106,6 @@ public class CollectionAssignmentBuilder {
 
     /**
      * @param assignment the assignment that needs to be invoked
-     *
      * @return this builder for chaining
      */
     public CollectionAssignmentBuilder assignment(Assignment assignment) {
@@ -116,7 +115,6 @@ public class CollectionAssignmentBuilder {
 
     /**
      * @param sourceRHS the source right hand side for getting the property for mapping
-     *
      * @return this builder for chaining
      */
     public CollectionAssignmentBuilder rightHandSide(SourceRHS sourceRHS) {
@@ -124,12 +122,12 @@ public class CollectionAssignmentBuilder {
         return this;
     }
 
-    public CollectionAssignmentBuilder nullValueCheckStrategy( NullValueCheckStrategyGem nvcs ) {
+    public CollectionAssignmentBuilder nullValueCheckStrategy(NullValueCheckStrategyGem nvcs) {
         this.nvcs = nvcs;
         return this;
     }
 
-    public CollectionAssignmentBuilder nullValuePropertyMappingStrategy( NullValuePropertyMappingStrategyGem nvpms ) {
+    public CollectionAssignmentBuilder nullValuePropertyMappingStrategy(NullValuePropertyMappingStrategyGem nvpms) {
         this.nvpms = nvpms;
         return this;
     }
@@ -178,8 +176,12 @@ public class CollectionAssignmentBuilder {
                     ctx.getTypeFactory(),
                     targetAccessorType.isFieldAssignment()
                 );
+                if ( result.getSourceType().isPrimitive() ) {
+                    result.setSourceLocalVarName( null );
+                }
             }
-            else if ( method.isUpdateMethod() && nvpms == IGNORE ) {
+            else if ( method.isUpdateMethod() && nvpms == IGNORE
+                || setterWrapperNeedsSourceNullCheck( result ) && canBeMappedOrDirectlyAssigned( result ) ) {
 
                 result = new SetterWrapperForCollectionsAndMapsWithNullCheck(
                     result,
@@ -188,17 +190,7 @@ public class CollectionAssignmentBuilder {
                     ctx.getTypeFactory(),
                     targetAccessorType.isFieldAssignment()
                 );
-            }
-            else if ( setterWrapperNeedsSourceNullCheck( result )
-                && canBeMappedOrDirectlyAssigned( result ) ) {
-
-                result = new SetterWrapperForCollectionsAndMapsWithNullCheck(
-                    result,
-                    method.getThrownTypes(),
-                    targetType,
-                    ctx.getTypeFactory(),
-                    targetAccessorType.isFieldAssignment()
-                );
+                result.setSourceLocalVarName( null );
             }
             else if ( canBeMappedOrDirectlyAssigned( result ) ) {
                 //TODO init default value
@@ -210,6 +202,7 @@ public class CollectionAssignmentBuilder {
                     targetType,
                     targetAccessorType.isFieldAssignment()
                 );
+                result.setSourceLocalVarName( null );
             }
             else if ( hasNoArgsConstructor() ) {
                 result = new NewInstanceSetterWrapperForCollectionsAndMaps(
@@ -217,7 +210,9 @@ public class CollectionAssignmentBuilder {
                     method.getThrownTypes(),
                     targetType,
                     ctx.getTypeFactory(),
-                    targetAccessorType.isFieldAssignment() );
+                    targetAccessorType.isFieldAssignment()
+                );
+                result.setSourceLocalVarName( null );
             }
             else {
                 ctx.getMessager().printMessage(
@@ -244,6 +239,7 @@ public class CollectionAssignmentBuilder {
                 nvpms,
                 targetAccessorType.isFieldAssignment()
             );
+            result.setSourceLocalVarName( null );
         }
 
         return result;
@@ -251,8 +247,8 @@ public class CollectionAssignmentBuilder {
 
     private boolean canBeMappedOrDirectlyAssigned(Assignment result) {
         return result.getType() != AssignmentType.DIRECT
-                  || hasCopyConstructor()
-                  || targetType.isEnumSet();
+            || hasCopyConstructor()
+            || targetType.isEnumSet();
     }
 
     /**
@@ -294,8 +290,8 @@ public class CollectionAssignmentBuilder {
             }
             else {
                 Element sourceElement = targetType.getImplementationType() != null
-                                      ? targetType.getImplementationType().getTypeElement()
-                                      : targetType.getTypeElement();
+                    ? targetType.getImplementationType().getTypeElement()
+                    : targetType.getTypeElement();
                 if ( sourceElement != null ) {
                     for ( Element element : sourceElement.getEnclosedElements() ) {
                         if ( element.getKind() == ElementKind.CONSTRUCTOR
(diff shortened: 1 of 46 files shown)


processor/src/test/resources/fixtures/org/mapstruct/ap/test/conversion/java8time/OptionalSourceTargetMapperImpl.java
```diff
@@ -41,46 +41,60 @@ public Target sourceToTarget(OptionalSource source) {
 
         Target target = new Target();
 
-        if ( source.getZonedDateTime().isPresent() ) {
+        Optional<ZonedDateTime> zonedDateTime = source.getZonedDateTime();
+        if ( zonedDateTime.isPresent() ) {
             target.setZonedDateTime( dateTimeFormatter_dd_MM_yyyy_HH_mm_z_01894582668.format( source.getZonedDateTime().get() ) );
         }
```
filiphr, line 47: We are still using `source.getZonedDateTime().get()` in the optional conversion. I believe this should be `zonedDateTime.get()` instead of `source.getZonedDateTime()`

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -821,6 +830,14 @@ else if ( !sourceReference.isNested() ) {
             }
         }
 
+        private boolean presenceCheckAllowsLocalVar(PresenceCheck presenceCheck) {
```
filiphr, line 833: **🔴 Root cause of the unconditional-getter regression.** This guard only excludes `MethodReferencePresenceCheck`; every other presence-check type (notably `SuffixPresenceCheck`, i.e. the `hasX()`/`isSetX()` idiom) falls through to `return true`, so a local var is created and the getter is hoisted *above* the `if (source.hasX())` guard — executing it even when the check is false.

The principle here is right ("only hoist when the presence check consumes the source RHS value"), but it's enforced incompletely. Suggest returning `false` for presence checks that don't read the RHS — `SuffixPresenceCheck`, and please review `JavaExpressionPresenceCheck` plus `Negate`/`All`/`Any` wrappers around non-RHS checks. Local-var hoisting is safe/beneficial only for `NullPresenceCheck`, value-based `OptionalPresenceCheck`, and `MethodReferencePresenceCheck` bound to the source RHS.

Minor: this method encodes the subtlest decision in the PR — worth a short comment explaining the intent.

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -767,6 +773,9 @@ else if ( !sourceReference.isNested() ) {
                     sourceReference,
                     sourceRHS
                 ) );
+                if ( presenceCheckAllowsLocalVar( sourceRHS.getSourcePresenceCheckerReference() ) ) {
```
filiphr, line 776: This is where the local var gets set for simple (non-nested) properties. Because `presenceCheckAllowsLocalVar` returns `true` for `SuffixPresenceCheck` (see comment on the method below), a `hasX()`-guarded property ends up with an unconditionally-evaluated local var. The `_913` fixtures show the effect.

processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
```diff
@@ -178,19 +176,12 @@ else if ( method.isUpdateMethod() && !targetImmutable ) {
                     ctx.getTypeFactory(),
                     targetAccessorType.isFieldAssignment()
                 );
+                if ( result.getSourceType().isPrimitive() ) {
```
filiphr, line 179: In the mutable-collection update path (`ExistingInstanceSetterWrapperForCollectionsAndMaps`), the local var is only reset for **primitive** source types. So a `List<String>` fed by a `hasX()` presence check keeps the hoisted local var and is read unconditionally (this is the exact branch the `_913` update mappers hit). If the fix in `presenceCheckAllowsLocalVar` lands, this primitive-only reset may be redundant — but as it stands today it doesn't cover the regression.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/bugs/_913/DomainDtoWithPresenceCheckMapperImpl.java
```diff
@@ -59,64 +59,69 @@ public void update(DtoWithPresenceCheck source, Domain target) {
             return;
         }
 
+        List<String> strings = source.getStrings();
```
filiphr, line 62: Concrete evidence of the regression: `source.getStrings()` is now invoked here **unconditionally**, before the `if ( source.hasStrings() )` guards below. Previously it only ran inside those guards. For JAXB-style `isSetX()`/lazy/expensive getters this is an observable behavior change (and a counting-getter test would fail here).

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -660,6 +661,11 @@ private boolean hasDefaultValueOrDefaultExpression() {
         private Assignment assignToPlainViaAdder( Assignment rightHandSide) {
 
             Assignment result = rightHandSide;
+            if ( sourceReference == null || !sourceReference.isNested() ) {
+                // For non-nested properties, reset the local var name: the adder wrapper handles
```
filiphr, line 665: Minor: the comment says "For non-nested properties", but the guard also covers `sourceReference == null` (expression/constant mappings). Consider "For non-nested (or absent) source references". Optionally note *why* nested is excluded — nested references keep their local var to avoid re-invoking the forged accessor method.

stefanos-kalantzis:
hi @filiphr, I went through the changes again and I think it should be ok to merge.

filiphr (commented):
Thanks for tackling #4030, @stefanos-kalantzis.

I have one concern and one test gap (details in the inline comments):

**Behavioral change for custom presence-check methods (`hasX()` / `isSetX()`).** For a `SuffixPresenceCheck` (e.g. `source.hasStrings()`), the value getter `source.getStrings()` is now hoisted into an unconditional local variable, so it runs even when the presence check returns `false`. See the regenerated `_913/DomainDtoWithPresenceCheckMapperImpl`:

```java
// before: getStrings() only ran inside the guard
if ( source.hasStrings() ) {
    target.getStrings().addAll( source.getStrings() );
}
// after: getStrings() runs unconditionally
List<String> strings = source.getStrings();
if ( source.hasStrings() ) {
    target.getStrings().addAll( strings );
}
```

This is the JAXB `isSetFoo()` / `getFoo()` idiom - calling `getFoo()` unconditionally lazily instantiates the source collection, defeating the purpose of the presence check. It's the "expensive / side-effecting getter" case #4030 is meant to protect, and it yields no benefit here since the presence-check path never had a double call to begin with. Root cause: `presenceCheckAllowsLocalVar` only excludes `MethodReferencePresenceCheck`; `SuffixPresenceCheck` falls through to `return true`. Could we extend the guard to skip hoisting when the presence check doesn't read the source RHS value?

**No behavioral test.** The PR relies entirely on regenerated golden fixtures, which is not bad per se. However, I think it would be good if we can add some behaviour tests. e.g. for the scenario I mentioned before with the use of the presence checker perhaps?

processor/src/test/resources/fixtures/21/org/mapstruct/ap/test/bugs/_913/DomainDtoWithNcvsAlwaysMapperImpl.java
```diff
@@ -59,64 +59,69 @@ public void update(DtoWithPresenceCheck source, Domain target) {
...
                 target.getStrings().clear();
-                target.getStrings().addAll( source.getStrings() );
+                target.getStrings().addAll( strings );
             }
         }
         else {
             if ( source.hasStrings() ) {
-                List<String> list = source.getStrings();
+                List<String> list = strings;
                 target.setStrings( new LinkedHashSet<>( list ) );
```
hduelme, line 72: The assignment is now redundant and should be inlined.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/nestedbeans/UserDtoUpdateMapperSmartImpl.java
```diff
@@ -88,22 +92,24 @@ protected void carToCarDto(Car car, CarDto mappingTarget) {
...
-            List<WheelDto> list = wheelListToWheelDtoList( car.getWheels() );
+            List<WheelDto> list = wheelListToWheelDtoList( wheels );
             if ( list != null ) {
                 mappingTarget.getWheels().clear();
                 mappingTarget.getWheels().addAll( list );
             }
         }
         else {
-            List<WheelDto> list = wheelListToWheelDtoList( car.getWheels() );
+            List<WheelDto> list = wheelListToWheelDtoList( wheels );
```
hduelme, line 112: I don't think that the variable is needed, as `wheels` is only used in two separated branches.

processor/src/test/resources/fixtures/org/mapstruct/ap/test/optional/nested/OptionalNestedPresenceCheckFirstMapperImpl.java
```diff
@@ -23,8 +23,9 @@ public TargetAggregate map(Source song) {
 
         TargetAggregate targetAggregate = new TargetAggregate();
 
+        Optional<String> title = song.getTitle();
         if ( song.hasTitle() ) {
-            targetAggregate.setSongTitle( song.getTitle().get() );
+            targetAggregate.setSongTitle( title.get() );
```
hduelme, line 28: The variable is not needed as `song` as an check method (`hasTitle()`).

processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
```diff
@@ -178,19 +176,12 @@ else if ( method.isUpdateMethod() && !targetImmutable ) {
...
+                    result.setSourceLocalVarName( null );
+                }
```
hduelme, line 181: Why are primitives excluded?

processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
```diff
@@ -821,6 +830,14 @@ else if ( !sourceReference.isNested() ) {
...
         }
 
+        private boolean presenceCheckAllowsLocalVar(PresenceCheck presenceCheck) {
+            if ( presenceCheck instanceof MethodReferencePresenceCheck ) {
+                return ((MethodReferencePresenceCheck) presenceCheck).getMethodReference().getParameterBindings()
+                    .stream().anyMatch( ParameterBinding::isForSourceRhs );
+            }
+            return true;
+        }
+
```
hduelme, line 840: What is checked here?

hduelme (changes requested):
@stefanos-kalantzis Thanks for your contribution. The generated code looks much cleaner now.

I left a few comments regarding the generated code style. It looks like too many local variables are being generated now.

Additionally, I'm not fully convinced about setting `setSourceLocalVarName` to `null` later on. It feels a bit like we're using mutable state to undo a previous change, which makes the flow harder to follow. Could we avoid setting it in the first place, or model this differently?

hduelme:
Oops, it looks like @filiphr and I started reviewing at the same time. I didn't check for existing review comments before submitting mine, so there may be some overlap.

stefanos-kalantzis:
Sorry I haven't had the time to go through the comments yet.

@hduelme Regarding the `setSourceLocalVarName` to `null`, this was a pattern that was already there in the code. I just expanded on it. From my limited understanding, there are cases where this variable is needed so that the template does the correct thing, and cases where its not, and there doesn't seem to be a good default (its like 50/50 depending on the desired outcome). If you have better ideas please suggest, since this is the first time I'm getting familiar with the project :)

hduelme:
@stefanos-kalantzis you are right that pattern was already present. For example https://github.com/mapstruct/mapstruct/blob/f5253e82492ae693b9407017992edbc8de30765d/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java#L507

Maybe its better to not fix it in this PR.

Issue #4062, opened 2026-06-03
Use switch Expressions for JDK13 and newer


hduelme:
### Use case

Switch Expressions added in JDK13 give a better way of writing switches.
Currently value mappings uses normal switches only.

### Generated Code

Instead of generating:
```java
        switch ( b ) {
            case "A": a = ValueMapper.A.A;
            break;
            case "B": a = ValueMapper.A.B;
            break;
            default: throw new IllegalArgumentException( "Unexpected enum constant: " + b );
        }
```

the generated output could be 
```java
        ValueMapper.A a = switch (b) {
            case "A" -> A.A;
            case "B" -> A.B;
            default -> throw new IllegalArgumentException("Unexpected enum constant: " + b);
        };
```
For enums constants the default branch is not needed.

### MapStruct Version

1.6.3

hduelme:
Implemented in #4065

PR #4065, opened 2026-06-05
#4062 added the new switch expression usage for jdk greater than 13

bydrim:
As described in the [issue #4062](https://github.com/mapstruct/mapstruct/issues/4062) for jdk equal or greater than 13 the usage of new switch expression added. Related fixtures are updated. Also as discussed with Filip, for switch expressions with only **default** case, defaultTarget is directly assigned without switch expression, such as
```java
ExternalOrderType externalOrderType = ExternalOrderType.DEFAULT;
```

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/processor/DefaultVersionInformation.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/processor/DefaultVersionInformation.java
@@ -41,12 +41,13 @@ public class DefaultVersionInformation implements VersionInformation {
     private final String compiler;
     private final boolean sourceVersionAtLeast9;
     private final boolean sourceVersionAtLeast11;
+    private final boolean sourceVersionAtLeast14;
     private final boolean sourceVersionAtLeast19;
     private final boolean eclipseJDT;
     private final boolean javac;
 
     DefaultVersionInformation(String runtimeVersion, String runtimeVendor, String compiler,
-        SourceVersion sourceVersion) {
+                              SourceVersion sourceVersion) {
         this.runtimeVersion = runtimeVersion;
         this.runtimeVendor = runtimeVendor;
         this.compiler = compiler;
@@ -55,6 +56,7 @@ public class DefaultVersionInformation implements VersionInformation {
         // If the difference between the source version and RELEASE_6 is more that 2 than we are at least on 9
         this.sourceVersionAtLeast9 = sourceVersion.compareTo( SourceVersion.RELEASE_6 ) > 2;
         this.sourceVersionAtLeast11 = sourceVersion.compareTo( SourceVersion.RELEASE_6 ) > 4;
+        this.sourceVersionAtLeast14 = sourceVersion.compareTo( SourceVersion.RELEASE_6 ) > 7;
         this.sourceVersionAtLeast19 = sourceVersion.compareTo( SourceVersion.RELEASE_6 ) > 12;
     }
 
@@ -86,6 +88,11 @@ public class DefaultVersionInformation implements VersionInformation {
     @Override
     public boolean isSourceVersionAtLeast11() {
         return sourceVersionAtLeast11;
+    }
+
+    @Override
+    public boolean isSourceVersionAtLeast14() {
+        return sourceVersionAtLeast14;
     }
 
     @Override
--- a/processor/src/main/resources/org/mapstruct/ap/internal/model/ValueMappingMethod.ftl
+++ b/processor/src/main/resources/org/mapstruct/ap/internal/model/ValueMappingMethod.ftl
@@ -21,15 +21,35 @@
         <#if nullTarget.targetAsException>throw new <@includeModel object=unexpectedValueMappingException />( "Unexpected enum constant: " + ${sourceParameter.name} );<#else>return <@writeTarget target=nullTarget.target/>;</#if>
     }
 
-    <@includeModel object=resultType/> ${resultName};
+    <#if versionInformation.isSourceVersionAtLeast14()>
+        <#if valueMappings.empty>
+            <#if defaultTarget.targetAsException>
+                <@includeModel object=resultType/> ${resultName};
+                throw new <@includeModel object=unexpectedValueMappingException />( "Unexpected enum constant: " + ${sourceParameter.name} );
+            <#else>
+                <@includeModel object=resultType/> ${resultName} = <@writeTarget target=defaultTarget.target/>;
+            </#if>
+        <#else>
+            <@includeModel object=resultType/> ${resultName} = switch ( ${sourceParameter.name} ) {
+                <#list valueMappings as valueMapping>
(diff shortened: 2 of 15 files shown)


bydrim:
It seems that the code is not compliant with the formatting rules. Working on it.

bydrim:
I realized that Reformat Code on save was not active on my Intellij. I used git squash to keep the git logs clean, so these are the formatting changes: 
 - ValueMappingMethod.java line 567
 - DefaultVersionInformation.java line 7 (actually i didn't touch this line in my first commit but intellij formatted when i was reformatting the whole file)

bydrim:
The fail seems to be related to [codecov migrating their keybase accounts](https://github.com/codecov/codecov-action/issues/1956). Their advice is to use v7, and currently v5 is used in this project.

hduelme:
@bydrim yes the JDK 21 pipline failed because of the issue you referenced. We upgraded the plugin to v7.

> The fail seems to be related to [codecov migrating their keybase accounts](https://github.com/codecov/codecov-action/issues/1956). Their advice is to use v7, and currently v5 is used in this project.

processor/src/main/resources/org/mapstruct/ap/internal/model/ValueMappingMethod.ftl
```diff
@@ -21,15 +21,33 @@
...
+                throw new <@includeModel object=unexpectedValueMappingException />( "Unexpected enum constant: " + ${sourceParameter.name} );
+            <#else>
+                <@includeModel object=resultType/> ${resultName} = <@writeTarget target=defaultTarget.target/>;
+            </#if>
+        <#else>
+            <@includeModel object=resultType/> ${resultName} = switch ( ${sourceParameter.name} ) {
+                <#list valueMappings as valueMapping>
+                    case <@writeSource source=valueMapping.source/> -> <#if valueMapping.targetAsException > throw new <@includeModel object=unexpectedValueMappingException />( "Unexpected enum constant: " + ${sourceParameter.name} );<#else><@writeTarget target=valueMapping.target/>;</#if>
+                </#list>
+                default -> <#if defaultTarget.targetAsException >throw new <@includeModel object=unexpectedValueMappingException />( "Unexpected enum constant: " + ${sourceParameter.name} )<#else><@writeTarget target=defaultTarget.target/></#if>;
```
hduelme: The `default` branch is only necessary when not all values are explicitly covered by a `case`.

hduelme (changes requested):
@bydrim nice work.
I think we could stop generating the `default` branch if all enum values are explicitly covered by `case` statements. Am I overlooking something, or would that work here? Let me know what you think.

processor/src/main/java/org/mapstruct/ap/internal/processor/DefaultVersionInformation.java
```diff
@@ -88,6 +90,11 @@ public boolean isSourceVersionAtLeast11() {
         return sourceVersionAtLeast11;
     }
 
+    @Override
+    public boolean isSourceVersionAtLeast13() {
+        return sourceVersionAtLeast13;
+    }
+
```
stefanos-kalantzis, line 97: Weren't switch expressions released in Java 14? https://openjdk.org/jeps/361

Java 13 needs to have preview enabled to use them.

processor/src/main/resources/org/mapstruct/ap/internal/model/ValueMappingMethod.ftl
stefanos-kalantzis: expanding on @hduelme's comment,

There are 2 possibilities here:
1. if the switch covers all cases, the `default` branch should not be there  (it's unreachable/dead code).
2. if the switch does NOT cover all cases, then the compiler will fail with:
   ```
   'switch' expression does not cover all possible input values
   ```
   ([switch-expressions are exhaustive](https://openjdk.org/jeps/361#Exhaustiveness))

In my opinion, if the above compiler error occurs, its a valid outcome for mapstruct, indicating that the user needs to fix something. A default case would move this error from compile-time to runtime, which is the opposite direction from what the JEP was designed for.

If mapstruct should behave differently than that, then the user should define a custom mapping.

(I hope I'm not missing something...)

stefanos-kalantzis (changes requested):

processor/src/main/java/org/mapstruct/ap/internal/processor/DefaultVersionInformation.java
hduelme: @stefanos-kalantzis you are right in Java 13 it was only available with  `–enable-preview`.

bydrim: Oh, sorry that I didn't think of checking when exactly it was released. should i change it to version 14 then?

hduelme: @bydrim I think so, yes. It's available by default since Java 14.

Sorry, my analysis in the issue was incorrect.

processor/src/main/resources/org/mapstruct/ap/internal/model/ValueMappingMethod.ftl
hduelme: @stefanos-kalantzis currently mapstruct enforces that all constants are mapped. If not an error is given

> The following constants from the source enum have no corresponding constant in the target enum and must be be mapped via adding additional mappings: C.

This behavior can be alter with map `MappingConstants`. For example
`@ValueMapping(target = MappingConstants.THROW_EXCEPTION, source = MappingConstants.ANY_REMAINING)`
results in all unmapped values being mapped as `THROW_EXCEPTION`.

bydrim: I also agree that it is unnecessary to use ```default``` branch when all possible ```case```s are covered. To figure out if all the ```case```s are covered, i thought about checking ```defaultTarget.targetAsException == true```. However, if I understand correctly, it is possible to leave unmatched values and the default behavior of the library is to give warning instead of throwing error during compilation, unless configured otherwise by the user. Also it is possible to choose to throw exception for specified targets. Am i wrong and how can i check if all ```case```s are covered?

bydrim: I'm pinging in case you missed previous comment. How can i make sure that all the cases are covered?

hduelme: @bydrim I think the best way would be to move this into the `ValueMappingMethod` with an own method. For the following cases a default type should be requird.

- If Java-version is under 14
- If the SourceParameter type isn't an enum
- If the number of EnumConstants of the SourceParameter type didn't match the number of valueMappings.

In addition `getImportTypes` should be adjusted to this as well

hduelme (approved):

hduelme:
@bydrim Thanks for making the adjustments, and thanks for the contribution.

bydrim:
thank you @hduelme and Filip for the opportunity and support. I learned a lot and had fun. I hope to contribute more in the future. Also thank you @stefanos-kalantzis for correcting the version number.

PR #4069, opened 2026-06-09
#4037 Stop binding property-level @Condition to unrelated source parameter

### LINKED ISSUE #4037: @Condition applied on wrong source parameter while mapping
### Expected behavior

`@Condition` method should be correctly applied on correct source parameters when we are using `nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS`

### Actual behavior

When we have the mapping mapping with 2 source parameters, for eg:

```java
@Mapping(target = "browserId", source = "requestBean.deviceInfo.browserId")
public abstract AudienceProfileRequest map(RequestBean requestBean, String slotId);
```

If we are mapping some `Integer` source field from `RequestBean` to `AudienceProfileRequest` then the condition is applied on `slotId` which is not expected. Following is the snippet of the generated code:

```java
Integer browserId = requestBeanDeviceInfoBrowserId( requestBean );
if ( prechecksSupport.isNotBlank( slotId ) ) {
    audienceProfileRequest.browserId( browserId );
}
```

https://github.com/altaiezior/mapstruct-issues/tree/master/issue-4037

### MapStruct Version

MapStruct 1.5.5

seonwooj0810:
Fixes #4037.

## Why this happens

For a property-level `@Condition`, `PresenceCheckMethodResolver` asks `TypeSelector` to bind every parameter of the `@Condition` method to an available `ParameterBinding`. The bindings published for a property mapping include both the property's source value (`SourceRHS`) and every other source parameter of the mapping method.

`TypeSelector` does not know which binding is *the value being checked*. It enumerates permutations and keeps the ones whose types are assignable. When `SourceRHS` does not match the `@Condition` parameter's type but an unrelated sibling source parameter happens to match, that sibling is silently selected — even though it has nothing to do with the property being mapped.

In the reporter's example:

```java
@Mapping(target = "browserId", source = "requestBean.deviceInfo.browserId")
AudienceProfileRequest map(RequestBean requestBean, String slotId);

class PrechecksSupport {
    @Condition boolean isNotBlank(String value) { ... }
}
```

`browserId` is an `Integer`, so `SourceRHS` does not match `isNotBlank(String)`. The sibling `slotId` parameter does, so MapStruct generates

```java
Integer browserId = requestBeanDeviceInfoBrowserId( requestBean );
if ( prechecksSupport.isNotBlank( slotId ) ) {
    audienceProfileRequest.setBrowserId( browserId );
}
```

A blank `slotId` would then silently null out every `Integer` property of the result even when the actual source values are present.

## What changes

After `PresenceCheckMethodResolver.getPresenceCheck` collects the matching methods, discard any whose chosen binding consumes a sibling source parameter that is *not* the root of the property's source path. The root is `SourceRHS#getSourceParameterName()` — e.g. `requestBean` for `source = "requestBean.deviceInfo.browserId"`. All other binding kinds (the `SourceRHS` itself, `@MappingContext`, `@TargetType`, `@SourcePropertyName`, `@TargetPropertyName`, `@MappingTarget`) remain valid candidates.

This keeps every existing pattern working:

- `@Condition isNotBlank(String value)` on a `String` property — `value` binds to `SourceRHS`.
- `@Condition shouldMap(EmployeeDto source)` on a property of `employee` — `source` binds to the mapper parameter, which is the root of the property's source path.
- `@Condition isAmericanCitizen(EmployeeDto employerDto)` referenced via `conditionQualifiedByName` — same as above.
- `@Condition isPresent(EmployeeDto source, String value)` — `source` binds to the parameter root, `value` binds to `SourceRHS`.

When the unrelated-sibling case is the *only* match left, the resolver returns `null` and the property falls back to the default null check (or no check, depending on `nullValueCheckStrategy`).

## Tests

- `org.mapstruct.ap.test.bugs._4037.Issue4037Test` — multi-source-parameter mapper with `nullValueCheckStrategy = ALWAYS` and a `@Condition isNotBlank(String)`. Confirms the `Integer` properties no longer use `isNotBlank(slotId)` as their guard while the `String` property still gets the `@Condition`.
- The full `processor` module test suite runs green (3633 tests). The conditional/qualifier/expression and presence-check suites in particular continue to pass without changes.

## Verification done

- `./mvnw -pl processor -Dtest='Issue4037Test' test` — 4/4 pass.
- `./mvnw -pl processor -Dtest='*Conditional*Test,*Condition*Test,Issue4037Test' test` — 58/58 pass.
- `./mvnw -pl processor -Dtest='*NullCheck*Test,*NullValueCheckStrategy*Test,*Selection*Test,*Presence*Test,*PresenceCheck*Test,*Selector*Test' test` — 136/136 pass.
- `./mvnw -pl processor -Dtest='org.mapstruct.ap.test.bugs.**,*MultiSource*Test,*SourcePresence*Test' test` — 696/696 pass.
- `./mvnw -pl processor test` — 3633/3633 pass.
- `./mvnw -pl processor checkstyle:check` — 0 violations.
- Inspected the generated source for the reproducer (`Issue4037MapperImpl.java`) and confirmed the previously buggy `if ( prechecksSupport.isNotBlank( slotId ) )` guard is replaced with the default `if ( browserId != null )` / `if ( osId != null )` check, while the matching String property still emits `if ( prechecksSupport.isNotBlank( deviceName ) )`.


seonwooj0810:
Rebased onto current `main` to resolve the conflict. The only conflicting file was `NEXT_RELEASE_CHANGELOG.md` — its `### Bugs` section had been cleared upstream after the release, so the two stale entries my branch carried are gone and only the `#4037` line remains.

The fix in `PresenceCheckMethodResolver` and the `_4037` tests are unchanged; `Issue4037Test` passes locally (4 tests). Happy to reword the changelog entry if you'd prefer something different.

seonwooj0810:
Following up on this one: it's still conflict-free on current `main` with all checks green. It stops a property-level `@Condition` from being silently bound to an unrelated sibling source parameter (#4037). Happy to adjust if you'd prefer the fix to live in `TypeSelector` instead of `PresenceCheckMethodResolver`.

PR #4071, opened 2026-06-13
Define new enum ClassAccessibility

### LINKED ISSUE #2513: Generated mappers for Spring can be package-private
Working with Spring, Java > 1.8 and Mapstruct 1.4.2.Final.

The mappers generated for Spring ( `componentModel = "spring"`) are implementing an interface, annotated with `@Component`, so they don't need to be public, they can be package-private by default. 

They are meant to be injected by Spring in other classes/services, so the usual client code don't needs to do `new MyMapperImpl();`

Can you add a way to generate package-private implementations? 

In a project, it can be detected by running test using ArchUnit ( this test with fail due to MapStruct generated classes). 
```
ArchRuleDefinition.noClasses()
            .that()
            .haveSimpleNameEndingWith("Impl")
            .should().bePublic()
            .because("Implementations don't need to be public")
            .check(new ClassFileImporter().withImportOption(ImportOption.Predefined.DO_NOT_INCLUDE_TESTS).importPackages("my.nice.pacakge"));
```

Chessray:
Define new enum `ClassAccessibility` for controlling generated Mappers' declared accessibility.

Fixes #2513.

### DIFF
--- a/core/src/main/java/org/mapstruct/ClassAccessibility.java
+++ b/core/src/main/java/org/mapstruct/ClassAccessibility.java
@@ -6,7 +6,7 @@
 package org.mapstruct;
 
 /**
- * Indicates whether a generated Mapper implementation class should be declared {@code public} or not.
+ * Determines whether a generated Mapper implementation class will be declared {@code public} or not.
  *
  * @author Raimund Klein
  *
@@ -14,16 +14,15 @@ package org.mapstruct;
  */
 public enum ClassAccessibility {
     /**
-     * Indicates that the generated Mapper should have the same modifier ({@code public} or none) as the annotated
-     * class or interface.
+     * The generated Mapper will have the same modifier ({@code public} or none) as the annotated class or interface.
      */
     LIKE_ABSTRACTION,
     /**
-     * Indicates that the generated Mapoer should be declared {@code public}.
+     * The generated Mapper will be declared {@code public}.
      */
     PUBLIC,
     /**
-     * Indicates that the generated Mapper should have no visibility modifier ("package-private").
+     * The generated Mapper will have no visibility modifier ("package-private").
      */
     DEFAULT
 }
--- a/core/src/main/java/org/mapstruct/ClassAccessibility.java
+++ b/core/src/main/java/org/mapstruct/ClassAccessibility.java
@@ -16,7 +16,7 @@ public enum ClassAccessibility {
     /**
      * The generated Mapper will have the same modifier ({@code public} or none) as the annotated class or interface.
      */
-    LIKE_ABSTRACTION,
+    DEFAULT,
     /**
      * The generated Mapper will be declared {@code public}.
      */
@@ -24,5 +24,5 @@ public enum ClassAccessibility {
     /**
      * The generated Mapper will have no visibility modifier ("package-private").
      */
-    DEFAULT
+    PACKAGE_PRIVATE
 }
--- a/core/src/main/java/org/mapstruct/ClassAccessibility.java
+++ b/core/src/main/java/org/mapstruct/ClassAccessibility.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright MapStruct Authors.
+ *
+ * Licensed under the Apache License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package org.mapstruct;
+
+/**
+ * Indicates whether a generated Mapper implementation class should be declared {@code public} or not.
+ *
+ * @author Raimund Klein
+ *
+ * @since 1.7.0
+ */
+public enum ClassAccessibility {
+    /**
+     * Indicates that the generated Mapper should have the same modifier ({@code public} or none) as the annotated
+     * class or interface.
+     */
+    LIKE_ABSTRACTION,
+    /**
+     * Indicates that the generated Mapoer should be declared {@code public}.
+     */
+    PUBLIC,
+    /**
+     * Indicates that the generated Mapper should have no visibility modifier ("package-private").
+     */
+    DEFAULT
+}
--- a/processor/src/main/java/org/mapstruct/ap/internal/gem/ClassAccessibilityGem.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/gem/ClassAccessibilityGem.java
@@ -11,5 +11,5 @@ package org.mapstruct.ap.internal.gem;
  * @author Raimund Klein
  */
 public enum ClassAccessibilityGem {
-    LIKE_ABSTRACTION, PUBLIC, DEFAULT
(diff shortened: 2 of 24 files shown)


core/src/main/java/org/mapstruct/ClassAccessibility.java
```diff
@@ -0,0 +1,29 @@
...
+     */
+    LIKE_ABSTRACTION,
+    /**
+     * Indicates that the generated Mapoer should be declared {@code public}.
+     */
+    PUBLIC,
+    /**
+     * Indicates that the generated Mapper should have no visibility modifier ("package-private").
+     */
+    DEFAULT
```
hduelme: I’m not sure about the wording. _Indicates_ feels a bit weak, since this controls the actual behavior.
Something like `Generated mapper is declared {@code public}. ` would fit better.

core/src/main/java/org/mapstruct/ClassAccessibility.java
```diff
@@ -0,0 +1,29 @@
...
+ * @since 1.7.0
+ */
+public enum ClassAccessibility {
+    /**
+     * Indicates that the generated Mapper should have the same modifier ({@code public} or none) as the annotated
+     * class or interface.
+     */
+    LIKE_ABSTRACTION,
+    /**
+     * Indicates that the generated Mapoer should be declared {@code public}.
```
hduelme: typo Mapoer -> Mapper

processor/src/main/java/org/mapstruct/ap/internal/gem/ClassAccessibilityGem.java
```diff
@@ -0,0 +1,15 @@
...
+package org.mapstruct.ap.internal.gem;
+
+/**
+ * Gem for the enum {@link org.mapstruct.ClassAccessibility}
+ *
+ * @author Raimund Klein
+ */
+public enum ClassAccessibilityGem {
+    LIKE_ABSTRACTION, PUBLIC, DEFAULT
+}
```
hduelme, line 15: Gem-Enums currently have an mapping test in `org.mapstruct.ap.test.gem.EnumGemsTest`

hduelme (changes requested):
Hi @Chessray, left a few small comments,  implementation looks good overall. Could you also add the tests for `@MapperConfig`?

Chessray:
> Hi @Chessray, left a few small comments, implementation looks good overall. Could you also add the tests for `@MapperConfig`?

Thanks @hduelme, I hope I've got them all. Could you please check again?

hduelme (approved):
@Chessray thanks for the adjustments.

One last thought: currently, `DEFAULT` is used to represent package-private visibility. Would something like `PACKAGE_PRIVATE` be more intuitive and make the intent clearer? Let me know what you think.

Out of curiosity, are you planning to use this feature `mapstruct-spring-extensions`?

Chessray:
> @Chessray thanks for the adjustments.
> 
> One last thought: currently, `DEFAULT` is used to represent package-private visibility. Would something like `PACKAGE_PRIVATE` be more intuitive and make the intent clearer? Let me know what you think.

I had this exact mental discussion and ended up with `DEFAULT` because of the same terminology being used in [Accessibility.java](https://github.com/mapstruct/mapstruct/blob/9376d42561c503ceab5a18a14ca317a32f46d733/processor/src/main/java/org/mapstruct/ap/internal/model/common/Accessibility.java). I find `PACKAGE_PRIVATE` more descriptive myself. Maybe we can rename both places? Does @filiphr want to chime in here?

> 
> Out of curiosity, are you planning to use this feature `mapstruct-spring-extensions`?

I don't really see how. This influences the Mapper generation, not the Adapter.

filiphr:
Nice addition @Chessray. From the point of our API, I personally find the `LIKE_ABSTRACTION` a bit strange. What I would propose is to use `DEFAULT` for the mapstruct default (i.e. The current like abstraction) and use `PACKAGE_PROTECTED` for package protected.

hduelme:
> I had this exact mental discussion and ended up with `DEFAULT` because of the same terminology being used in [Accessibility.java](https://github.com/mapstruct/mapstruct/blob/9376d42561c503ceab5a18a14ca317a32f46d733/processor/src/main/java/org/mapstruct/ap/internal/model/common/Accessibility.java). I find `PACKAGE_PRIVATE` more descriptive myself

@Chessray Good point. I think we should rename [Accessibility.java](https://github.com/mapstruct/mapstruct/blob/9376d42561c503ceab5a18a14ca317a32f46d733/processor/src/main/java/org/mapstruct/ap/internal/model/common/Accessibility.java) too. `DEFAULT` doesn't really fit.

hduelme:
> I don't really see how. This influences the Mapper generation, not the Adapter.

Makes sense. Do you think we should adjust the examples in the extension? My thoughts were based on the original issue mentioning Spring.

Chessray:
Thanks guys, this should be it now. @hduelme I'll take a look at the Spring Extension examples separately. Maybe I'll only mention it in the documentation.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Accessibility.java
```diff
@@ -14,7 +14,7 @@
  * @author Andreas Gudian
  */
 public enum Accessibility {
-    PRIVATE( "private" ), DEFAULT( "" ), PROTECTED( "protected" ), PUBLIC( "public" );
+    PRIVATE( "private" ), PACKAGE_PRIVATE("" ), PROTECTED("protected" ), PUBLIC("public" );
```
hduelme: @Chessray now the enum constants are formatted inconsistent. No pace before the string an one after.

hduelme (changes requested):
@Chessray thanks for adjusting. The naming looks good to me now.
It seems like we currently don't enforce consistent enum formatting.

@filiphr I think the preferred formatting is to have spaces before and after enum constant parameters (e.g. `PUBLIC( "public" )`). Do you agree? Maybe we could add this rule to our Checkstyle configuration.

processor/src/main/java/org/mapstruct/ap/internal/model/common/Accessibility.java
Chessray: 😬  Must be something missing in the style setting. Anyway, corrected.

hduelme (approved):

processor/src/main/java/org/mapstruct/ap/internal/model/common/Accessibility.java
hduelme: I will have a look at the missing style settings. Thanks for adjusting.

hduelme:
@Chessray Nice one. Thanks

PR #4078, opened 2026-06-28
#4077 Add missing null check before reusing non-nullable methods

### LINKED ISSUE #4077: JSpecify: Missing null check before reusing non-nullable method
### Expected behavior

When a non-nullable mapper method (under `@NullMarked`) is reused for a `@Nullable` source property, MapStruct should emit a null guard before the call.

Given:
```java
@Mapper
@NullMarked
public interface Issue4077Mapper {
    Target map(Source source);
    Target.Nested mapNested(Source.Nested source); // non-null parameter
}

public class Source {
    private final @Nullable Nested nested;
    public @Nullable Nested getNested() { return nested; }
}
```

Expected:
```java
@Generated(
    date = "0000-00-00T00:00:00+0000",
    comments = "version: , compiler: javac, environment: Java 21"
)
public class Issue4077MapperImpl implements Issue4077Mapper {

    @Override

        Target.Nested nested = null;

        if ( source.getNested() != null ) {
            nested = mapNested( source.getNested() );
        }

        Target target = new Target( nested );

        return target;
    }

    @Override
    public Target.Nested mapNested(Source.Nested source) {

        String foo = null;
        Integer bar = null;

        foo = source.getFoo();
        bar = source.getBar();

        Target.Nested nested = new Target.Nested( foo, bar );

        return nested;
    }
}
```

### Actual behavior

The null check is missing, so `null` values will throw `NullPointerException`:

```
Changed content at line 22:
expecting:
  ["        if ( source.getNested() != null ) {",
   "            nested = mapNested( source.getNested() );",
   "        }"]
but was:
  ["        nested = mapNested( source.getNested() );"]
```

Can be reproduced in this branch: https://github.com/mapstruct/mapstruct/compare/main...RaniAgus:mapstruct:issues/4077

### MapStruct Version

1.7.0.Beta2

RaniAgus:
resolves #4077

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -610,7 +610,10 @@ public class PropertyMapping extends ModelElement {
             NullabilityResolver.Nullability targetNullability = resolver.getSetterNullability(
                 targetWriteAccessor.getElement(), this::targetDeclaringTypeIsNullMarked
             );
-            Boolean jspecifyDecision = resolver.requiresNullCheck( sourceNullability, targetNullability );
+            NullabilityResolver.Nullability setterParamNullability = getMethodParamNullability( rhs );
+            Boolean jspecifyDecision = resolver.requiresNullCheck(
+                sourceNullability, targetNullability, setterParamNullability
+            );
             if ( jspecifyDecision != null ) {
                 ctx.getMessager().note( 2,
                     jspecifyDecision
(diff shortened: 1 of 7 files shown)


aegliv:
Awesome work with mapstruct 1.7.0.Beta2. I encountered the same Bugs and desperately await the fixes. @filiphr Is there any chance of having a 1.7.0.Beta3 release withe the implemented JSpecify fixes soon™️ ? Is there a way we can help out? 🙏

filiphr:
Thanks for the kind words @aegliv and thanks for the PR and all reported issues @RaniAgus.

The goal is to do an RC release with all the reported jspecify issues fixed

hduelme:
@RaniAgus, thanks for your contribution! We’re currently working on a more general fix that addresses all of the reported issues. We’ll keep this PR open for reference while we complete that work.

PR #4080, opened 2026-06-28
Fix #4079 `@NullMarked` scope not detected for constructor/setter parameters

### LINKED ISSUE #4079: JSpecify: `@NullMarked` scope not detected for constructor/setter parameters
### Expected behavior

When a target class is annotated `@NullMarked`, unannotated constructor parameters should be treated as `@NonNull`. Mapping a `@Nullable` source to such a parameter without a default value should produce a compile error.

Given:
```java
@NullMarked
public class Target {
    private final Nested nested;
    public Target(Nested nested) { this.nested = nested; } // implicitly @NonNull
}

@NullMarked
public class Source {
    private final @Nullable Nested nested;
    public @Nullable Nested getNested() { return nested; }
}

@Mapper
@NullMarked
public interface Issue4079Mapper {
    Target map(Source source);
}
```

Expected: compilation fails with
```
error: Can't map potentially nullable source property "nested" to @NonNull constructor parameter "nested".
```

### Actual behavior

Compilation succeeds.

Can be reproduced in this branch: https://github.com/mapstruct/mapstruct/compare/main...RaniAgus:mapstruct:issues/4079

### MapStruct Version

1.7.0.Beta2

RaniAgus:
resolves #4079

### DIFF
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/PropertyMapping.java
@@ -681,7 +681,10 @@ public class PropertyMapping extends ModelElement {
                 return false;
             }
             Element declaring = targetElement.getEnclosingElement();
-            if ( !( declaring instanceof TypeElement ) ) {
+            while ( declaring != null && !( declaring instanceof TypeElement ) ) {
+                declaring = declaring.getEnclosingElement();
+            }
+            if ( declaring == null ) {
                 return false;
             }
             return ctx.getTypeFactory().getType( declaring.asType() ).isNullMarked();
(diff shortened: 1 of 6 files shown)


hduelme:
@RaniAgus, thanks for your first contribution! The implementation and the tests look good to me.

Could you also add the corresponding tests for setter parameters and `@NullUnmarked`?

RaniAgus:
Added tests to cover both affected write-accessor kinds (setter parameter and constructor parameter) and both relevant scope outcomes:
- enclosing target type is `@NullMarked`
- closer target type is `@NullUnmarked`, overriding an outer `@NullMarked`

hduelme:
@RaniAgus, thanks for adding the tests, and special thanks for all the bug reports. Really appreciate your help.

Issue #1685, opened 2019-01-13
NullValuePropertyMapping strategy not implemented fully.


sjaakd:
See also [stackoverflow](https://stackoverflow.com/questions/54037352/prevent-null-check-in-mapstruct-on-update) issue.

Null check occurs for update method (with an `@MappingTarget`) in case of:
* `NullValueStrategyCheck` sais so (e.g. always)
* In case of type conversion to get from source to target)
* In case of additional mapping methods called to get from source to target.
* In case of nested mappings (they will generate an mapping method).

Unfortunately the current strategy only works when from an update method, another update method is called. 

This issue will see that all the above scenarios are handled.

sjaakd:
done by merged PR

PR #4089, opened 2026-07-09
Avoid stale nested source local var in updates

### LINKED ISSUE #4088: JSpecify: broken code is generated
### Expected behavior

Correct code is generated

### Actual behavior

code that cannot be compiled is generated

I've created a repo to reproduce the problem https://github.com/chpasha/mapstruct_bug/tree/master/src/main/java/com/example/demo
for the update method of the mapper 
```@Mapper
public abstract class SwTerminMapper {
   @Mapping(target = "datum", source = "termin.termin")
    public abstract void updateFromRequest(UpdateSwTerminRequest request, @MappingTarget SwTermin termin);
    
}
```

following broken code is generated

``` 
@Override
    public void updateFromRequest(UpdateSwTerminRequest request, SwTermin termin) {

        if ( termin.getDatum() == null ) {
            termin.setDatum( new LocalDate() );
        }
        localDateTimeToLocalDate1( termin1, termin.getDatum() );
    }
```
here termin1 variable is undefined. if I remove jspecify (delete package-info.java)
correct code is generated like this
```
@Override
    public void updateFromRequest(UpdateSwTerminRequest request, SwTermin termin) {
        if ( request == null ) {
            return;
        }

        LocalDateTime termin1 = requestTerminTermin( request );
        if ( termin1 != null ) {
            if ( termin.getDatum() == null ) {
                termin.setDatum( new LocalDate() );
            }
            localDateTimeToLocalDate1( termin1, termin.getDatum() );
        }
        else {
            termin.setDatum( null );
        }
    }
```

It is important to notice, that LocalDate, LocalTime and LocalDateTime classes in my examples are custom, not from JDK (don't ask) - but they are crucial for replication of the bug, since they need conversion methods (which are correctly generated by mapstruct without intervention from my side)

### MapStruct Version

MapStruct 1.7.0.Beta2

sakshichitnis27:
## Summary

Fixes #4088.

This clears the nested source local variable name when an update mapping does not need a source null-check. Without that, MapStruct can generate code that references an undeclared variable.

I added a regression test for the JSpecify + Lombok + record update-mapping case.

## Testing

- `./mvnw -pl integrationtest verify -Dtest=MavenIntegrationTest#issue4088Test -DskipDistribution=true -Djacoco.skip=true`
- `./mvnw -pl processor -Dtest=JSpecify*Test -Dsurefire.failIfNoSpecifiedTests=false test -DskipDistribution=true -Djacoco.skip=true`
- `./mvnw -pl processor test -DskipDistribution=true -Djacoco.skip=true`


hduelme:
@sakshichitnis27, thanks for your first contribution!

The fix looks good overall. PR #1685 already introduced the "set to `null`" pattern, so this change is consistent with the existing approach.

One comment regarding the tests: integration tests aren't intended to cover individual issues. Instead, they execute the complete processor test suite (see `fullFeatureTest`).

For this issue, I'd suggest adding a bug test in the processor module instead. Since this is a code generation/style issue, I'd also recommend using:

```
@RegisterExtension
final GeneratedSource generatedSource = new GeneratedSource()
```

This allows you to verify that the generated source matches the expected output, which is a better fit for this kind of regression test.

sakshichitnis27:
@hduelme Thanks for the guidance. 
I moved the issue-specific regression test from integrationtest to the processor module and added GeneratedSource verification with an expected generated-source fixture. The production fix remains unchanged. The focused test and full processor test suite pass. Please take another look.

hduelme:
@sakshichitnis27 Thanks for making the adjustment.

Since you've decided to use records (which is fine with me), the tests should be moved into an inner package named `jdk17`, as records are not supported in Java 8.

Also, I noticed that the PR author and the commit author are different. Is that expected (for example, are you collaborating with someone or submitting a commit on their behalf)? I just wanted to confirm that the commit attribution is correct.

sakshichitnis27:
@hduelme 
Thanks for pointing this out. I moved the record-based issue 4088 regression test and its generated-source fixture into the _4088.jdk17 package so the Java 8 test build does not compile record sources. I also corrected the commit attribution; all commits are now authored by my GitHub account, sakshichitnis27. 
The production fix remains unchanged.

hduelme (approved):

hduelme:
@sakshichitnis27 Thanks again for making the adjustment and for your contribution

PR #4091, opened 2026-07-11
#3997 Fix InheritInverseConfiguration ignore for nested sources with parameter prefix

### LINKED ISSUE #3997: @InheritInverseConfiguration ignores "ignore = true" for nested source fields
### Expected behavior

When mapping from a DTO back to an entity using @InheritInverseConfiguration, any field marked with @Mapping(ignore = true) such as client, name, or address, should stay null. This should also hold when the source fields are nested and explicitly referenced with 'source.', for example source.client.id or source.address.city. The mapper should not create any objects or set values for these ignored fields. In MapStruct 1.3.0.Final, ignored fields were respected and left untouched, even with nested source paths.

### Actual behavior

In MapStruct 1.4.2.Final and later, using @InheritInverseConfiguration with nested source fields prefixed by 'source.' results in completely ignoring the ignore = true annotations. The generated mapper creates intermediate objects for these ignored fields and sets some of their values, even though the fields were explicitly marked as ignored. As a result, client and address are no longer null after mapping. One observation is that removing the 'source.' prefix, such as using 'client.id' instead of 'source.client.id', seems to solve the issue, but this may not reflect the intended behavior of the mapper.

Mapper:
@Mapper
public interface CommandeMapper {

    CommandeMapper INSTANCE = Mappers.getMapper(CommandeMapper.class);

    @Mapping(target = "clientId", source = "source.client.id")
    @Mapping(target = "nom", source = "source.name")
    @Mapping(target = "ville", source = "source.address.city")
    CommandeDto toDto(Commande source);

    @InheritInverseConfiguration
    @Mapping(target = "client", ignore = true)
    @Mapping(target = "name", ignore = true)
    @Mapping(target = "address", ignore = true)
    Commande toEntity(CommandeDto dto);
}

Mapper Implementation:

@Generated(
    date = "2026-02-17T09:06:33+0100",
    comments = "version: 1.7.0.Beta1, compiler: javac, environment: Java 21.0.1 (Eclipse Adoptium)"
)
public class CommandeMapperImpl implements CommandeMapper {

    @Override
    public CommandeDto toDto(Commande source) {
        if ( source == null ) {
            return null;
        }

        CommandeDto commandeDto = new CommandeDto();

        commandeDto.setClientId( sourceClientId( source ) );
        commandeDto.setNom( source.getName() );
        commandeDto.setVille( sourceAddressCity( source ) );
        commandeDto.setId( source.getId() );

        return commandeDto;
    }

    @Override
    public Commande toEntity(CommandeDto dto) {
        if ( dto == null ) {
            return null;
        }

        Commande commande = new Commande();

        commande.setClient( commandeDtoToClient( dto ) );
        commande.setAddress( commandeDtoToAddress( dto ) );
        commande.setId( dto.getId() );

        return commande;
    }

    private Long sourceClientId(Commande commande) {
        Client client = commande.getClient();
        if ( client == null ) {
            return null;
        }
        return client.getId();
    }

    private String sourceAddressCity(Commande commande) {
        Address address = commande.getAddress();
        if ( address == null ) {
            return null;
        }
        return address.getCity();
    }

    protected Client commandeDtoToClient(CommandeDto commandeDto) {
        if ( commandeDto == null ) {
            return null;
        }

        Client client = new Client();

        client.setId( commandeDto.getClientId() );

        return client;
    }

    protected Address commandeDtoToAddress(CommandeDto commandeDto) {
        if ( commandeDto == null ) {
            return null;
        }

        address.setCity( commandeDto.getVille() );

        return address;
    }
}

### MapStruct Version

MapStruct 1.4.2.Final and later

kamilkrzywanski:
## Summary

Fixes #3997

When a forward mapping uses a source path that includes the source parameter name (e.g. `source.client.id`), `@InheritInverseConfiguration` previously kept that prefix in the inverse target name (`source.client.id`). As a result, `@Mapping(target = "client", ignore = true)` on the inverse method did not match the nested inherited mapping, and MapStruct still created intermediate objects for ignored properties.

### Root cause

1. `copyForInverseInheritance` swapped source/target naively, so `source = "source.client.id"` became inverse `target = "source.client.id"`.
2. Code generation later strips the parameter prefix via `TargetReference`, so nested mapping still *worked*.
3. Ignore / redefinition matching runs earlier on the **raw** target name, where the first segment is the parameter name (`source`), not the property (`client`).

### Fix

When building the inverse target name, strip the source parameter name prefix (e.g. `source.client.id` → `client.id`), mirroring `SourceReference` / `TargetReference` resolution. If the first segment is also a property of the source type, the full path is kept (property path takes precedence).

### Tests

- Reproducer from the issue: nested sources with parameter prefix + inverse `ignore = true` → nested properties stay null
- Guard: same nested sources without ignores still reverse-map correctly

## Checklist

- [x] Provide test cases
- [x] One pull request per feature/bug
- [x] Meaningful history (single focused commit)
- [x] Changelog entry in `NEXT_RELEASE_CHANGELOG.md`


kdelay (commented):
I ran this locally on `75ad6b9` against `211b2be` (JDK 21.0.4, both javac and ecj through `@ProcessorTest`). The diagnosis and the fix hold for the reported shape. One thing about the guard clause looks worth resolving before it lands.

## Verification

Control first, to check the test actually defends the change: the four new test files on top of `main`, with `MappingOptions.java` reverted to `main` and nothing else changed.

```
Issue3997Test.inverseIgnoreShouldRespectNestedSourcesWithParameterPrefix:35
expected: null
 but was: org.mapstruct.ap.test.bugs._3997.Issue3997Mapper$Client@58168099
Tests run: 4, Failures: 2, Errors: 0, Skipped: 0
```

Two failures are the same assertion under the two compilers. On the PR head: `Tests run: 4, Failures: 0`. Full `processor` module on the PR head: `Tests run: 3638, Failures: 0, Errors: 0, Skipped: 0`.

I also checked that `first( templateMethod.getSourceParameters() )` cannot fail here: `SourceMethod#inverses` (`SourceMethod.java:352`) requires exactly one source parameter on both methods, so the template always has one.

## The guard uses a different rule than the two resolvers it mirrors

`stripSourceParameterPrefix` keeps the full path when the first segment is a read accessor of the source type. Neither resolver decides it that way.

* `SourceReference.BuilderFromMapping#buildFromSingleSourceParameters` (`SourceReference.java:152-188`) matches the **whole** path against the source accessors first and strips the parameter name only when the entire path fails to resolve. A first segment that happens to be a property is not enough to keep the path.
* `TargetReference` (`TargetReference.java:205-206`) strips when the first segment is **not among the target properties** and matches the parameter name.

So when the parameter name collides with a property that does not carry the rest of the path, the forward mapping resolves through the strip path while this PR keeps the prefix, and the fix does not apply. Reproducer: parameter named `client`, `Order` has both `getClient()` and `getId()`, `Client` has no `id`.

```java
@Mapper
public interface M {
    @Mapping(target = "dtoId", source = "client.id")
    OrderDto toDto(Order client);

    @InheritInverseConfiguration
    @Mapping(target = "client", ignore = true)
    Order toEntity(OrderDto dto);
}
```

The forward side resolves by stripping, so the inverse target should be `id`:

```
-- MapStruct: mapping property: client.getId() to: setDtoId(java.lang.Long).
```

Compiled with the processor built from this branch:

```
warning: Unmapped target property: "id".

    public Probe.Order toEntity(Probe.OrderDto dto) {
        if ( dto == null ) {
            return null;
        }

        Probe.Order order = new Probe.Order();

        return order;
    }
```

`dtoId -> id` is dropped: the inverse target name stays `client.id`, and `filterNestedTargetIgnores` then removes it under `@Mapping(target = "client", ignore = true)`. That is the same failure #3997 reports.

Two controls around it. The generated file is identical (apart from the `date` attribute) when compiled with a processor built from `main`, so this is not a regression introduced here, it is a case the fix declines to cover. And renaming the parameter to `order` (`source = "order.id"`, everything else unchanged) makes both `main` and this branch emit `order.setId( dto.getDtoId() )`, which isolates the guard as the deciding factor rather than anything else in the mapper.

Direction rather than a patch: have the guard ask the question `buildFromSingleSourceParameters` asks, i.e. whether the full path resolves against the source type's accessors, and strip only when it does not. That would also make the javadoc claim on the new method true of the code that runs.

Smaller notes:

* The `containsKey` branch is untested. Whichever rule you settle on, a mapper whose parameter name collides with a source property would pin it.
* The javadoc says the method mirrors `SourceReference` / `TargetReference`; going by the two line references above it currently matches neither.
* Agreed on the #4078 interaction noted in the description, nothing to add there.

kamilkrzywanski:
@kdelay thanks for the detailed review.

Updated the guard to match SourceReference: try the full path as property accessors first, strip the parameter name only when that fails. Added tests for both the strip and keep cases (including the client/id collision you described).

PR #4092, opened 2026-07-11
#3931 Fix @DeepClone to clone Date and Calendar instead of copying references

### LINKED ISSUE #3931: DeepClone does not clone java.util.Date fields but perform a copy reference
### Expected behavior

I have a mapper annotated DeepClone to perform a copy clone of some bean. I notice that the generated implementation on java.util.Date field it perform a simple `a.setDate(b.getDate())`. This could be ok for String or BigDecimal object that are immutable but old java Date object are not immutable that mean if I change `b.getDate().setTime(1l)` it will also update date in the bean a.

What I expect is that in case of DeepClone it produce the following code:
```
    @Override
    public BeanA clone(BeanA value) {
        if ( value == null ) {
            return null;
        }

        BeanA beanA = new BeanA();
        if (source.getDate() != null) {
            beanA.setDate(new Date(source.getDate().getTime())); // expected behaviour
        }
        
        return beanA;
    }
```

### Actual behavior

The generated code is
```
@Generated(
    date = "2025-09-16T10:12:55+0200",
    comments = "version: 1.6.3, compiler: Eclipse JDT (IDE) 3.41.0.v20250213-1140, environment: Java 21.0.6 (Eclipse Adoptium)"
)
public class ICloneMapperImpl implements ICloneMapper {

    @Override
    public BeanA clone(BeanA value) {
        if ( value == null ) {
            return null;
        }

        BeanA beanA = new BeanA();
        beanA.setDate(source.getDate());
        
        return beanA;
    }
}
```

Create the following classes
```
public class BeanA {
    private Date date;
    public Date getDate() { return date; }
    public void setDate(Date value) { date = value; }
}

@Mapper(mappingControl = DeepClone.class)
public interface ICloneMapper {
    BeanA clone(BeanA source);
}
```

### MapStruct Version

MapStruct 1.6.3

kamilkrzywanski:
## Summary

- Fixes [#3931](https://github.com/mapstruct/mapstruct/issues/3931): with `@Mapper(mappingControl = DeepClone.class)`, `java.util.Date` fields were assigned by reference (`setDate(source.getDate())`) instead of being cloned.
- Root cause: when `DIRECT` is disabled, types from the `java` package are still treated as always-direct. That is correct for immutable types like `String`, but wrong for mutable `Date` / `Calendar`.
- Processor-only fix in `MappingResolverImpl`: exclude `java.util.Date` and `java.util.Calendar` from the always-direct rule when `DIRECT` is off, and emit an inline clone:
  - `Date` → `new Date(source.getTime())`
  - `Calendar` → `((Calendar) source.clone())`
- Default mappers (with `DIRECT` enabled) are unchanged and still copy the reference.
- No public API changes.

## Generated code (after)

```java
if ( source.getDate() != null ) {
    beanA.setDate( new Date( source.getDate().getTime() ) );
}
if ( source.getCalendar() != null ) {
    beanA.setCalendar( ((Calendar) source.getCalendar().clone()) );
}
```

## Test plan

- [x] `Issue3931Test` — Date/Calendar independence under `@DeepClone`, null handling
- [x] `MappingControlTest` — existing deep-clone / mapping-control behaviour
- [x] `DateConversionTest` — normal Date conversion paths still pass


filiphr (changes requested):
Thanks for the PR @kamilkrzywanski. This might work for the use case. However, this is not the right solution for the problem at hand. There are other types that are mutable where we cannot do a direct assignment. 

One can argue that we should perhaps always create a clone of the objects that are mutable. Are you interested in working on a right fix that would fix in the architecture of MapStruct? If not, then we'll have to reject the PR and work on a fix for this separately

kamilkrzywanski:
@filiphr thanks. Agreed this isn't the right approach. I won't be able to take on the broader fix, so please close if you prefer to handle it yourselves.

Issue #2593, opened 2021-09-23
allow use of mapstruct in default package with nested classes


maxandersen:
- [x] Is this an issue (and hence not a question)? 

given this code:

```
///usr/bin/env jbang "$0" "$@" ; exit $?
// //DEPS <dependency1> <dependency2>
//DEPS org.mapstruct:mapstruct:1.4.2.Final
//DEPS org.mapstruct:mapstruct-processor:1.4.2.Final
// JAVAC_OPTIONS -Amapstruct.verbose=true -verbose -XprintRounds -XprintProcessorInfo -Xlint -J-verbose

import static java.lang.System.*;

import org.mapstruct.Mapper;
import org.mapstruct.Mapping;
import org.mapstruct.factory.Mappers;

public class bt {

    public static enum CarType {
        SEDAN,CAMPER,X4,TRUCK;
    }

    static public class Car {
     
        public Car(String string, int i, bt.CarType sedan) {
            this.make = string;
            this.numberOfSeats = i;
            this.type = sedan;
        }
        public String make;
        public int numberOfSeats;
        public CarType type;
     
        //constructor, getters, setters etc.
    }

   static public class CarDto {
    
       public String make;
       public int seatCount;
       public String type;
    
       //constructor, getters, setters etc.
   }


   @Mapper
   static public interface CarMapper {
    
       CarMapper INSTANCE = Mappers.getMapper( CarMapper.class ); 
    
       @Mapping(source = "numberOfSeats", target = "seatCount")
       CarDto carToCarDto(Car car); 
   }



    public static void main(String... args) {
        
            //given
            Car car = new Car( "Morris", 5, CarType.SEDAN );
         
            //when
            CarDto carDto = CarMapper.INSTANCE.carToCarDto( car );
        
            System.out.println(carDto.make);
            System.out.println(carDto.seatCount);
            System.out.println(carDto.type);

           
        }
    }
```

and running with `jbang bt.java` you get errors like:

```
/Users/max/.jbang/cache/jars/bt.java.40801430b10128fe194e3128fba5356928baa6560293bd3d350864d36c5e7b28.jar.tmp/bt$CarMapperImpl.java:3: error: package bt does not exist
import bt.CarDto;
```

mapstruct seem to think static public class is a package when it is inside default package.

Adding `package boo;` then it works.

filiphr:
Thanks for reporting this @maxandersen.

Will need to try it out. We've had some other problems with default packages. We will look into this.

[...]

Issue #1386, opened 2018-02-25
Generate imports only for top level classes


filiphr:
Instead of generating imports for nested classes we should only import the top class and use that when using the nested class.

For example for:

```java
public class Top {

    public static class Nested {
    }
}
```

When using `Nested` instead of importing it's FQN we should only import it's parent `Top` and use `Top.Nested` to access it

sjaakd:
👍

sahebpreet:
Hi @filiphr,

I would like to contribute on this issue. Can I pick this up?

From what I understood, we need to change the imports of nested classes throughout the codebase to import top classes instead of fully qualified names, wherever possible.

Is my understanding correct?

Thanks,
Saheb Preet Singh.

filiphr:
Hey @sahebpreet,

this is not about the imports in our codebase, but in the code that we generate. I tried doing this once and I don't think that it is as easy as I thought back then :). Have a look at https://github.com/filiphr/mapstruct/tree/top-level-imports-only.

Perhaps we should remove the label and completely close this issue.

If you are willing to give this a go feel free, otherwise you can always have a look at some of the other issues. You can also pick an issue that doesn't have the "good first issue" label ;)

sahebpreet:
Hi @filiphr ,

Thanks for your reply 😄.

I will probably look at some other issues 👍 .

Thanks.

PR #4101, opened 2026-07-25
#2435 Do not import types from the unnamed package

### LINKED ISSUE #2435: Compilation error when generating mappings for a nested class in the unnamed package
Bit of an edge case, but while testing out another issue I discovered that MapStruct will cause a compilation error when using a nested class within the [unnamed package](https://docs.oracle.com/javase/specs/jls/se8/html/jls-7.html#jls-7.4.2).  It works fine if with non-nested classes in the unnamed package, however.  The error is due to an invalid import statement of the nested package member.

## Source
```java
// This class is in the unnamed package
import org.mapstruct.Mapper;

public class UnnamedPackageNestedMapperIssue {
    @Mapper
    public interface NestedUnnamedPackageMapper {
        Target map(Source source);
    }

    public static class Source {
        private String property;

        public String getProperty() { return property; }
        public void setProperty(String property) { this.property = property; }
    }

    public static class Target {
        private String property;

        public String getProperty() { return property; }
        public void setProperty(String property) { this.property = property; }
    }
}
```

## MapStruct generated code

```java
import UnnamedPackageNestedMapperIssue.NestedUnnamedPackageMapper;
import UnnamedPackageNestedMapperIssue.Source;
import UnnamedPackageNestedMapperIssue.Target;

@Generated(
    date = "2021-05-07T12:56:29-0500",
    comments = "version: 1.4.2.Final, compiler: IncrementalProcessingEnvironment from gradle-language-java-6.8.jar, environment: Java 11.0.9.1 (AdoptOpenJDK)"
)
public class UnnamedPackageNestedMapperIssue$NestedUnnamedPackageMapperImpl implements NestedUnnamedPackageMapper {

    @Override
        if ( source == null ) {
            return null;
        }

        Target target = new Target();

        target.setProperty( source.getProperty() );

        return target;
    }
}
```

## Gradle build output
```none
/Path/To/build/generated/sources/annotationProcessor/java/main/Path/To/build/generated/sources/annotationProcessor/java/main/UnnamedPackageNestedMapperIssue$NestedUnnamedPackageMapperImpl.java:2: error: package UnnamedPackageNestedMapperIssue does not exist
import UnnamedPackageNestedMapperIssue.NestedUnnamedPackageMapper;
                                      ^
/Path/To/build/generated/sources/annotationProcessor/java/main/UnnamedPackageNestedMapperIssue$NestedUnnamedPackageMapperImpl.java:3: error: package UnnamedPackageNestedMapperIssue does not exist
import UnnamedPackageNestedMapperIssue.Source;
                                      ^
/Path/To/build/generated/sources/annotationProcessor/java/main/UnnamedPackageNestedMapperIssue$NestedUnnamedPackageMapperImpl.java:4: error: package UnnamedPackageNestedMapperIssue does not exist
import UnnamedPackageNestedMapperIssue.Target;
                                      ^
/Path/To/build/generated/sources/annotationProcessor/java/main/UnnamedPackageNestedMapperIssue$NestedUnnamedPackageMapperImpl.java:12: error: cannot find symbol
public class UnnamedPackageNestedMapperIssue$NestedUnnamedPackageMapperImpl implements NestedUnnamedPackageMapper {
                                                                                       ^
  symbol: class NestedUnnamedPackageMapper
/Path/To/build/generated/sources/annotationProcessor/java/main/UnnamedPackageNestedMapperIssue$NestedUnnamedPackageMapperImpl.java:15: error: cannot find symbol
                      ^
  symbol:   class Source
  location: class UnnamedPackageNestedMapperIssue$NestedUnnamedPackageMapperImpl
/Path/To/build/generated/sources/annotationProcessor/java/main/UnnamedPackageNestedMapperIssue$NestedUnnamedPackageMapperImpl.java:15: error: cannot find symbol
           ^
  symbol:   class Target
  location: class UnnamedPackageNestedMapperIssue$NestedUnnamedPackageMapperImpl
6 errors
```

The compilation failure is due to the following import statements:

```none
import UnnamedPackageNestedMapperIssue.NestedUnnamedPackageMapper;
import UnnamedPackageNestedMapperIssue.Source;
import UnnamedPackageNestedMapperIssue.Target;
```

arimu1:
## Summary

Fixes compilation failures when generating mappers for **nested types in the unnamed package**.

Previously MapStruct could emit invalid imports such as:

```java
import UnnamedPackageNestedMapperIssue.Source;
import UnnamedPackageNestedMapperIssue.Target;
```

Java does not allow importing types from the unnamed package ([JLS §7.5](https://docs.oracle.com/javase/specs/jls/se8/html/jls-7.html#jls-7.5)), so those statements fail with `package UnnamedPackageNestedMapperIssue does not exist`.

### Change

In `GeneratedType.needsImportDeclaration`, skip import generation when the type’s package name is `null` or empty. Nested (and top-level) types from the unnamed package are then referenced without import statements, e.g. `Outer.Nested`.

Top-level-only imports (#1386 / #2616) already mitigated much of this; this change makes the JLS rule explicit and guards the edge case.

### Tests

- New processor regression test matching the issue setup (`Issue2435Test` / `UnnamedPackageNestedMapperIssue` in the unnamed package)
- Asserts generated impl does not contain `import UnnamedPackageNestedMapperIssue…`
- Asserts mapping works at runtime
- Existing `defaultPackage` integration test documented as covering #2435 as well as #2593

Fixes #2435

## Checklist

- [x] Test cases
- [x] Code formatter (MapStruct style)
- [ ] Reference documentation (not required for this bugfix)

### DIFF
--- a/integrationtest/src/test/java/org/mapstruct/itest/tests/MavenIntegrationTest.java
+++ b/integrationtest/src/test/java/org/mapstruct/itest/tests/MavenIntegrationTest.java
@@ -170,7 +170,7 @@ public class MavenIntegrationTest {
     void simpleTest() {
     }
 
-    // for issue #2593
+    // for issues #2593 and #2435 (nested types in the unnamed package)
     @ProcessorTest(baseDir = "defaultPackage")
     void defaultPackageTest() {
     }
--- a/integrationtest/src/test/resources/defaultPackage/main/java/DefaultPackageObject.java
+++ b/integrationtest/src/test/resources/defaultPackage/main/java/DefaultPackageObject.java
@@ -9,6 +9,8 @@ import org.mapstruct.Mapping;
 import org.mapstruct.factory.Mappers;
 
 /**
+ * Nested types and mappers in the unnamed package (issues #2593, #2435).
+ *
  * @author Filip Hrisafov
  */
 public class DefaultPackageObject {
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/GeneratedType.java
@@ -284,23 +284,29 @@ public abstract class GeneratedType extends ModelElement {
             return false;
         }
 
-        if ( typeToAdd.getPackageName() != null ) {
-            if ( typeToAdd.getPackageName().equals( JAVA_LANG_PACKAGE ) ) {
-                // only the types in the java.lang package are implicitly imported, the packages under java.lang
-                // like java.lang.management are not.
-                return false;
-            }
+        String typePackageName = typeToAdd.getPackageName();
 
-            if ( typeToAdd.getPackageName().equals( packageName ) ) {
-                if ( typeToAdd.getTypeElement() != null ) {
-                    if ( !typeToAdd.getTypeElement().getNestingKind().isNested() ) {
-                        return false;
-                    }
+        // Types in the unnamed package cannot be imported (JLS §7.5).
+        // Nested types there used to produce invalid imports like `import Outer.Nested`.
+        if ( typePackageName == null || typePackageName.isEmpty() ) {
+            return false;
+        }
+
+        if ( typePackageName.equals( JAVA_LANG_PACKAGE ) ) {
+            // only the types in the java.lang package are implicitly imported, the packages under java.lang
+            // like java.lang.management are not.
+            return false;
+        }
+
+        if ( typePackageName.equals( packageName ) ) {
+            if ( typeToAdd.getTypeElement() != null ) {
+                if ( !typeToAdd.getTypeElement().getNestingKind().isNested() ) {
+                    return false;
                 }
-                else if ( typeToAdd.getComponentType() != null ) {
-                    if ( !typeToAdd.getComponentType().getTypeElement().getNestingKind().isNested() ) {
-                        return false;
-                    }
+            }
+            else if ( typeToAdd.getComponentType() != null ) {
+                if ( !typeToAdd.getComponentType().getTypeElement().getNestingKind().isNested() ) {
+                    return false;
                 }
             }
         }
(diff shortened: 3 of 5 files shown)


filiphr (commented):
Thanks for the PR @arimu1. How does this work if the mapper is in one package and it uses something from the unnamed package (is this even possible)?

arimu1:
Thanks @filiphr — short answer: it is not possible, so the branch is unreachable. I checked it rather than assuming.

A type in the unnamed package is not accessible from a named package at all. JLS §7.5 says *"A type in an unnamed package cannot be imported"*, and simple-name resolution (§6.5.5.1) in a compilation unit of package `p` only sees types declared in `p`, single-/on-demand-imported types, and `java.lang`. There is simply no syntax that names an unnamed-package type from inside a named package.

Verified with `javac` 21 — `Outer` in the unnamed package, consumer in package `p`:

```java
// p/Consumer.java
package p;
public class Consumer {
    public Outer.Nested nested;
}
```
```
p/Consumer.java:4: error: package Outer does not exist
    public Outer.Nested nested;
                ^
```

and with an explicit import:

```java
// p/ConsumerWithImport.java
package p;
import Outer.Nested;
public class ConsumerWithImport {
    public Nested nested;
}
```
```
p/ConsumerWithImport.java:3: error: package Outer does not exist
import Outer.Nested;
            ^
```

So if any type in the model has an empty package name, the generated mapper is necessarily in the unnamed package too — which is exactly the case the issue reports and the case the new `Issue2435Test` covers.

And if it somehow did arise (a hand-built model, say), returning `false` is still the only sound answer: emitting `import Outer.Nested;` from a named package is a compile error either way. Before this change we emitted an illegal import; now we emit none. There is no input for which the old behaviour compiled and the new one does not.

Happy to add a comment on `needsImportDeclaration` recording the JLS reference if you would like that pinned in the source.

filiphr:
@arimu1 looking at the issue (#2435) it seems like this has been fixed already by #1386 and we already had a test case for this.

@arimu1 since this is clearly AI generated PR, I would advise you next time to properly validate whether a PR is even needed. My AI review surfaced the fact that this was already fixed and that the test passes even if we remove the production changes from this PR.

Anyways, I am closing this without merging it, as it is not needed

Issue #4086, opened 2026-07-04
JSpecify: False positive error for nullable source with nullable-param non-null-return method


RaniAgus:
### Expected behavior

When a `@Nullable` source property is mapped to a `@NonNull` constructor parameter via a reused mapping method that accepts `@Nullable` (explicitly handles null) and returns a non-null value, the mapping should compile successfully. The reused method's parameter and return type annotations prove the mapping is safe - null is handled and the result is guaranteed non-null.

Given:
```java
@Mapper
public interface Issue4086Mapper {
    @Mapping(target = "payload", source = "value", qualifiedByName = "mapValue")
    Target map(Source source);

    @Named("mapValue")
    default Target.Nested mapValue(@Nullable String value) {
        return value == null ? new Target.Nested( "default" ) : new Target.Nested( value );
    }
}

public class Source {
    private String value;
    public @Nullable String getValue() { return value; }
}

public class Target {
    private final Nested payload;
    public Target(@NonNull Nested payload) { this.payload = payload; }
}
```

Expected: compilation succeeds with:
```java
public Target map(Source source) {
    Target.Nested payload = mapValue( source.getValue() );
    Target target = new Target( payload );
    return target;
}
```

### Actual behavior

Compilation fails with:
```
error: Can't map potentially nullable source property "value" to @NonNull
constructor parameter "payload". Consider adding a defaultValue or defaultExpression.
```

The current check only inspects the source getter chain nullability, which reports `NULLABLE`. It does not consider that the assignment was resolved through a reused method whose parameter is `@Nullable` (no null guard needed before the call) and whose return type is `@NonNull` (the result is guaranteed non-null).

### Steps to reproduce the problem

A `@Nullable` source property mapped to a `@NonNull` constructor parameter
via `qualifiedByName` referencing a method with `@Nullable` parameter and non-null return. Can be reproduced in this branch: https://github.com/mapstruct/mapstruct/compare/main...RaniAgus:mapstruct:issues/4086

### MapStruct Version

1.7.0.Beta2

r0h1tb:
I had a go at this and got as far as pinning the root cause, so posting it in case it's useful to whoever picks it up.

The error is raised in `PropertyMapping.java` (`PropertyMappingBuilder.build()`, ~L286–303):

```java
if ( assignment != null
    && targetWriteAccessorType == AccessorType.PARAMETER
    && !hasDefaultValueOrDefaultExpression() ) {
[...]

Issue #4081, opened 2026-06-28
JSpecify: Missing compile error when reused method returns `@Nullable` to `@NonNull` constructor


RaniAgus:
### Expected behavior

Missing compile error when reused method returns `@Nullable` but constructor expects `@NonNull` parameter

Given:
```java
@Mapper
@NullMarked
public interface ErroneousIssue4081Mapper {
    Target map(Source source);

    default Target.@Nullable Payload mapPayload(String payload) {
        if ( payload.isBlank() ) {
            return null;
        }
        return new Target.Payload( payload );
    }
}

public class Source {
    private final String payload;

    public @NonNull String getPayload() { return this.payload; }
}

public class Target {
    private final Payload payload;
    public Target(@NonNull Payload payload) { this.payload = payload; }

    public class Payload {
        private final String value;
        public Payload(@NonNull String value) { this.value = value; }
    }
}
```

Expected: compilation fails with
```
error: Can't map potentially nullable source property "payload" to @NonNull constructor parameter "payload".
```

### Actual behavior

Compilation succeeds silently:
```java
@Generated
public class ErroneousIssue4081MapperImpl implements ErroneousIssue4081Mapper {

    @Override
    public Target map(Source source) {

        Target.Payload payload = null;

        payload = mapValue( source.getPayload() );

        Target target = new Target( payload );

        return target;
    }
}
```

### Steps to reproduce the problem

A `@NullMarked` mapper where a reused method has an explicit `@Nullable` return type, and the target constructor parameter is `@NonNull`.

Can be reproduced in this branch: https://github.com/mapstruct/mapstruct/compare/main...RaniAgus:mapstruct:issues/4081

### MapStruct Version

1.7.0.Beta2

PR #4110, opened 2026-08-10
Honor JSpecify nullability of reused mapping methods

### LINKED ISSUE #4106: Be able to generate the nullable mapping from the NotNull mapping
### Use case

We have a lot of technical mapping like date, enum (domain <-> Dto), etc.
I would like to write only the not null version add have mapstruct generating the boiling code to deal with the Nullable version when needed.

Exa

### Generated Code

Mapper definition:
```java
import org.jspecify.annotations.NullMarked;
import org.jspecify.annotations.Nullable;
import org.mapstruct.AnnotateWith;
import org.mapstruct.Mapper;

import java.time.OffsetDateTime;
import java.time.ZonedDateTime;

import static tech.lightframe.shift.shared.TimeZoneConstants.UTC_ZONE_ID;

@Mapper
@AnnotateWith(NullMarked.class)
@NullMarked
public interface SimpleMapper {

    record SourceEntity(ZonedDateTime createdAt, @Nullable ZonedDateTime closedAt) {
    }

    record TargetDto(OffsetDateTime createdAt, @Nullable OffsetDateTime closedAt) {
    }

    // Only define the mapping once for non null
    default OffsetDateTime toOffsetTime(ZonedDateTime zt) {
        return zt.withZoneSameInstant(UTC_ZONE_ID).toOffsetDateTime();
    }

    /// The mapping method should be able to map it
    TargetDto toDto(SourceEntity sourceEntity);
}

```

What is generated now:
```java
@NullMarked
@Component
public class SimpleMapperImpl implements SimpleMapper {

    @Override
    public SimpleMapper.TargetDto toDto(SimpleMapper.SourceEntity sourceEntity) {

        OffsetDateTime createdAt = null;
        OffsetDateTime closedAt = null;

        createdAt = toOffsetTime( sourceEntity.createdAt() );
        closedAt = toOffsetTime( sourceEntity.closedAt() ); <= Invalid code, it silently pass a null value to nonnull method

        SimpleMapper.TargetDto targetDto = new SimpleMapper.TargetDto( createdAt, closedAt );

        return targetDto;
    }
}
```

Expected:
```java
@NullMarked
@Component
public class SimpleMapperImpl implements SimpleMapper {

    @Override
    public SimpleMapper.TargetDto toDto(SimpleMapper.SourceEntity sourceEntity) {

        OffsetDateTime createdAt = null;
        OffsetDateTime closedAt = null;

        createdAt = toOffsetTime( sourceEntity.createdAt() );
        if (sourceEntity.closedAt() != null){
            closedAt = toOffsetTime( sourceEntity.closedAt() );
        }
        SimpleMapper.TargetDto targetDto = new SimpleMapper.TargetDto( createdAt, closedAt );

        return targetDto;
    }
}
```

It's not possible to define 2 mapping methods like that:
```
    // Only define the mapping once for non null
    default OffsetDateTime toOffsetTime(ZonedDateTime zt) {
        return zt.withZoneSameInstant(UTC_ZONE_ID).toOffsetDateTime();
    }

    default @Nullable OffsetDateTime toOffsetTimeNull(@Nullable ZonedDateTime zt) {
        if (zt == null) {
            return null;
        }
        return zt.withZoneSameInstant(UTC_ZONE_ID).toOffsetDateTime();
    }
```
With this configuration MapStruct fails with the error:
```
Ambiguous mapping methods found for mapping property "ZonedDateTime createdAt" to OffsetDateTime: OffsetDateTime toOffsetTime(ZonedDateTime zt), OffsetDateTime toOffsetTimeNull(ZonedDateTime zt). See https://mapstruct.org/faq/#ambiguous for more info.
```
### Possible workarounds

Use a qulifier with 2 different mapping.

### MapStruct Version

1.7.0.Beta2

kush2439p:
## Summary

Fixes #4106 by carrying the JSpecify input and output contracts of reused mapping methods through generated assignments.

MapStruct previously based null guards primarily on source/target nullability and null-value-check strategy. It did not retain whether a selected helper method required a non-null argument or guaranteed a non-null result, which could generate unsafe calls or reject safe constructor mappings.

## Implementation

- Add internal assignment contracts for the innermost reused-method input and outermost reused-method result.
- Resolve explicit and scoped JSpecify nullability from selected mapping methods, including methods from `uses` types.
- Propagate contracts through method chains and conversion wrappers.
- Generate a source guard when a nullable value is passed to a non-null helper parameter.
- Invoke nullable-input helpers directly when their result is not explicitly `@Nullable`, preserving legacy behavior for unannotated return types.
- Validate constructor parameters against the effective assignment result.
- Preserve legacy behavior for built-ins, forged methods, assignments without method contracts, and `-Amapstruct.disableJSpecify=true`.
- Add a verbose diagnostic for guards introduced by reused-method input contracts.
- Document null propagation and clarify that JSpecify annotations do not participate in overload selection.

This also covers the related scenarios described in #4077, #4081, and #4086 without adding public API.

## Verification

- Exact #4086 reproducer: compiled and executed successfully; generated code invokes `mapValue(source.getValue())` directly.
- Explicit-`@Nullable` return negative case: still produces the expected non-null-constructor diagnostic.
- Full `JSpecify*Test` suite: 111 tests passed.
- Focused reused-method, disabled-option, and verbose-diagnostic tests: 13 passed.
- Constructor contract regression tests: 2 passed.
- Full processor Surefire results: 3,561 tests, 0 failures, 0 errors.
- Processor Checkstyle: passed with 0 violations.
- Targeted `MavenIntegrationTest#fullFeatureTest`: passed.

The full reactor `clean install -DskipDistribution=true` exceeded the local 15-minute command limit after the processor suite completed; the affected integration path and Checkstyle were rerun successfully afterward.

### DIFF
--- a/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
+++ b/documentation/src/main/asciidoc/chapter-10-advanced-mapping-options.asciidoc
@@ -334,6 +334,12 @@ This rule applies uniformly to bean, iterable, map, and stream mapping methods.
 
 When the return type of a mapping method is `@NonNull` (directly or via a `@NullMarked` scope), MapStruct forces `NullValueMappingStrategy.RETURN_DEFAULT` semantics. The generated method returns a default-constructed target (bean methods), an empty collection (`Iterable` / array mappings), an empty map (`Map` mappings), or `Stream.empty()` (stream mappings) rather than `null`, so the return contract is never violated. This rule applies regardless of the explicit `NullValueMappingStrategy` setting.
 
+==== Reused mapping methods
+
+When a property assignment reuses a mapping method, MapStruct considers both sides of that method's JSpecify contract. A nullable source is guarded before calling a method whose first source parameter is `@NonNull`. A method whose first source parameter is `@Nullable` and whose return type is `@NonNull` is called directly, even when the source and target properties are nullable or non-null. For a chain of methods, the innermost input contract and outermost result contract are used.
+
+JSpecify annotations refine null-check generation only; they do not participate in mapping-method overload selection. If annotations make several methods otherwise applicable, use qualifiers to select the intended method.
+
 ==== Constructor parameter constraint
 
 If a property mapping would assign a potentially `null` source value to a `@NonNull` constructor parameter, MapStruct raises a *compilation error*. Neither inserting a null check (which would leave the variable at `null` and violate the contract) nor passing the value through is safe. Provide a `defaultValue` or `defaultExpression` on the `@Mapping` to satisfy the parameter when the source is absent.
--- a/processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
+++ b/processor/src/main/java/org/mapstruct/ap/internal/model/CollectionAssignmentBuilder.java
@@ -75,6 +75,7 @@ public class CollectionAssignmentBuilder {
     private NullValueCheckStrategyGem nvcs;
     private NullValuePropertyMappingStrategyGem nvpms;
     private NullabilityResolver.Nullability sourceJSpecifyNullability = NullabilityResolver.Nullability.UNKNOWN;
+    private NullabilityResolver.Nullability targetJSpecifyNullability = NullabilityResolver.Nullability.UNKNOWN;
 
     public CollectionAssignmentBuilder mappingBuilderContext(MappingBuilderContext ctx) {
         this.ctx = ctx;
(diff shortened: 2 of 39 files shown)


aegliv:
Issue #4086 is only partially fixed by PR #4110. The reused method's result nullability is honoured when it is provable - either an explicit `@NonNull` return type or a `@NullMarked` scope - but this reproducer (the verbatim snippet from the issue) leaves the return type unannotated and outside any `@NullMarked` scope, so it stays `UNKNOWN` and  `PROPERTYMAPPING_NULLABLE_SOURCE_TO_NON_NULL_CONSTRUCTOR_PARAM` is still raised.

@kush2439p plz verify using the verbatim snippet from #4086

kush2439p:
Verified and fixed using the verbatim #4086 shape.

The new `Issue4086Mapper` leaves `mapValue`'s return type unannotated and outside any `@NullMarked` scope. The generated implementation now invokes it directly:

```java
payload = mapValue( source.getValue() );
Target target = new Target( payload );
```

The compatibility rule is intentionally narrow: a helper that accepts `@Nullable` input is invoked directly when its result is not explicitly `@Nullable`; an explicit `@Nullable` return still produces `PROPERTYMAPPING_NULLABLE_SOURCE_TO_NON_NULL_CONSTRUCTOR_PARAM`.

Verification:
- exact #4086 mapper compiled and executed successfully
- explicit-`@Nullable` negative case still produced the expected diagnostic
- processor Checkstyle passed with 0 violations

The PR is ready for review.

PR #4114, opened 2026-08-18
#3385 Warn when mapper-declared lifecycle methods are never used

### LINKED ISSUE #3385: Unused @AfterMapping does not trigger warning
### Expected behavior

When a developer specifies @AfterMapping on a method (https://mapstruct.org/documentation/stable/reference/html/#customizing-mappings-with-before-and-after), and that method is **not consumed** by Mapstruct during code generation, then Mapstruct should emit a warning.

Rationale: The developer has declared an explicit intent that the method be _used_ but may have made an implementation error. Not warning about this allows for subtle, hard-to-detect bugs (in particular if test coverage is not good).

### Actual behavior

Mapstruct does not emit a warning that - in the reproducing case - "nowarning" is not consumed by Mapstruct.

```
import org.mapstruct.AfterMapping;
import org.mapstruct.Mapper;

@Mapper
public interface X {

    B map(A a);

    @AfterMapping
    default void good(A a) {
        //
    }

    @AfterMapping
    default void nowarning(String break_it, A a) {
        //
    }
}
```

### MapStruct Version

1.5.5.Final

bkks1004:
Unused `@BeforeMapping` / `@AfterMapping` methods currently produce no diagnostic.
If a developer adds an extra parameter by mistake, the method is silently ignored.

For the mapper from the issue:

```java
@Mapper
public interface X {

    B map(A a);

    @AfterMapping
    default void good(A a) {
    }

    @AfterMapping
    default void nowarning(String break_it, A a) {
    }
}
```

`good` is invoked, but `nowarning` is not, and MapStruct does not warn.

## Change

Warn only for lifecycle methods that are **declared directly on the processed mapper** and never type-match any generated mapping method.

Out of scope on purpose, so existing overload / inheritance patterns stay quiet:

- methods inherited from a super type (`BaseMapper`)
- methods coming from `Mapper#uses`
- overloads that match by type but are dropped in favor of a more specific one

`CarMapper.neverMatched(...)` is the remaining true unused case and now expects the warning.

## Verification

Run from the repository root with JDK 21.

- `Issue3385Test` covers unused `@AfterMapping` and `@BeforeMapping`
- `CallbackMethodTest`, `CallbackMethodTypeMatchingTest`, `Issue3849Test` stay green
- `./mvnw -pl processor test -Dtest=Issue3385Test,CallbackMethodTest,CallbackMethodTypeMatchingTest,Issue3849Test`

Fixes #3385


bkks1004:
Hi @filiphr,

When you have a chance, could you please take a look at this PR?

I believe it’s ready for review, and I’d especially appreciate feedback on whether the implementation approach fits MapStruct’s existing design and test conventions.

Thanks for your time!

PR #4115, opened 2026-08-19
#4111 Fix Optional<Map>/Optional<Collection> source properties calling Optional#get() unguarded

### LINKED ISSUE #4111: Optional<Map> / Optional<Collection> source property generates unguarded Optional#get, throwing NoSuchElementException when empty
### Expected behavior

When a source property is `Optional<T>`, the generated mapper should read the value only when it is present — as it already does for scalar types:

```java
if ( source.getName().isPresent() ) {
    name = source.getName().get();
}
```

### Actual behavior

For `Optional<Map<K, V>>` and `Optional<List<E>>` source properties the presence check is omitted and `Optional#get` is called unconditionally, so mapping a source whose `Optional` is empty throws `NoSuchElementException`.

Generated mapper (note the scalar property in the same method *is* guarded):

```java
@Override
public Target toTarget(Source source) {
    if ( source == null ) {
        return null;
    }

    String name = null;
    Map<String, String> attributes = null;
    List<String> tags = null;

    if ( source.getName().isPresent() ) {
        name = source.getName().get();
    }
    attributes = source.getAttributes().get();   // no presence check
    tags = source.getTags().get();               // no presence check

    Target target = new Target( name, attributes, tags );

    return target;
}
```

Runtime, mapping a `Source` whose three properties are all `Optional.empty()`:

```
Exception in thread "main" java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.get(Optional.java:143)
	at repro.SourceMapperImpl.toTarget(SourceMapperImpl.java:27)
	at Demo.main(Demo.java:9)
```

The inconsistency between the scalar and the map/collection properties in the same generated method suggests the presence check is lost on the branch that handles map and collection types.

This makes `Optional`-returning getters unusable in practice for map and collection properties, since an absent value is exactly the case `Optional` exists to express.

Related to #3976, but that one concerns `get()` vs `orElseThrow()`; here the presence check is missing altogether.

`Source.java`:

```java
package repro;

import java.util.List;
import java.util.Map;
import java.util.Optional;

public class Source
{
    private final Optional<String> name;
    private final Optional<Map<String, String>> attributes;
    private final Optional<List<String>> tags;

    public Source( Optional<String> name, Optional<Map<String, String>> attributes, Optional<List<String>> tags )
    {
        this.name = name;
        this.attributes = attributes;
        this.tags = tags;
    }

    public Optional<String> getName()
    {
        return name;
    }

    public Optional<Map<String, String>> getAttributes()
    {
        return attributes;
    }

    public Optional<List<String>> getTags()
    {
        return tags;
    }
}
```

`Target.java`:

```java
package repro;

import java.util.List;
import java.util.Map;

public record Target( String name, Map<String, String> attributes, List<String> tags )
{ }
```

`SourceMapper.java`:

```java
package repro;

import org.mapstruct.Mapper;

@Mapper
public interface SourceMapper
{
    Target toTarget( Source source );
}
```

`Demo.java`:

```java
import java.util.Optional;
import repro.*;

public class Demo
{
    public static void main( String[] args )
    {
        Source source = new Source( Optional.empty(), Optional.empty(), Optional.empty() );
        System.out.println( new SourceMapperImpl().toTarget( source ) );
    }
}
```

Compiled with plain `javac` (no build tool, no other annotation processors):

```
javac -cp mapstruct-1.7.0.Beta2.jar \
      -processorpath mapstruct-processor-1.7.0.Beta2.jar:mapstruct-1.7.0.Beta2.jar \
      -s gen -d out src/repro/*.java
```

Also reproduces with `nullValueMapMappingStrategy = RETURN_NULL` and with `nullValueCheckStrategy = ALWAYS`.

### MapStruct Version

1.7.0.Beta2, Java 25.0.4 (Eclipse Adoptium), javac.

seonwooj0810:
Fixes #4111

## Root cause

For a source property typed `Optional<T>`, scalar properties are guarded with an `isPresent()` check before the generated code calls `Optional#get()` (via `handleSourceReferenceNullCheck` in `CommonMacros.ftl`). `Optional<Map<K, V>>` / `Optional<Collection<E>>` source properties don't go through that macro though — `CollectionAssignmentBuilder.setterWrapperNeedsSourceNullCheck()` only requested a null-check wrapper when there's a source presence checker, `NullValueCheckStrategy.ALWAYS`, or a `DIRECT` assignment, so an Optional-wrapped collection/map fell through to the bare `SetterWrapperForCollectionsAndMaps`, which calls the assignment (`source.getAttributes().get()`) unconditionally. Mapping a source whose `Optional` is empty throws `NoSuchElementException` instead of leaving the target property unset, exactly as described in the issue.

## Fix

- `CollectionAssignmentBuilder.setterWrapperNeedsSourceNullCheck()` now also returns `true` when the source type is an `Optional`, so the `SetterWrapperForCollectionsAndMapsWithNullCheck` wrapper (and its `handleLocalVarNullCheck` macro) is used instead of the bare setter wrapper.
- `handleLocalVarNullCheck` in `CommonMacros.ftl` gained an `<#elseif sourceType.optionalType>` branch that guards the local-var assignment with `Optional#isPresent()` before calling `Optional#get()`, mirroring the existing scalar behavior in `handleSourceReferenceNullCheck`.

## Tests

Added `processor/src/test/java/org/mapstruct/ap/test/bugs/_4111/` with a mapper that has `Optional<String>`, `Optional<Map<String, String>>` and `Optional<List<String>>` source properties mapped to plain target properties:

- `emptyOptionalMapAndCollectionShouldNotThrow` — reproduces the reported bug: fails with `NoSuchElementException` before the fix, target properties are left `null` after the fix.
- `presentOptionalMapAndCollectionShouldBeMapped` — confirms present Optionals still map through correctly.

Verification done:
1. No in-flight PR/branch for #4111 (checked via `gh pr list --search` and `closedByPullRequestsReferences`/`linkedBranches` GraphQL query — both empty).
2. No active claim on the issue (unassigned, single non-maintainer comment tracing the same root cause, no PR referencing it).
3. Code-focused fix (`.java` + `.ftl` template), not docs/config.
4. Confirmed the bug reproduces on current `main` (`211b2be`) by running the new test against the pre-fix code — both new-test iterations failed with `NoSuchElementException` at the exact reported call site; after the fix, `./mvnw -pl processor -am test -Dtest=Issue4111Test` passes (4/4, JDK + Eclipse compiler).
5. No related open epic/tracking issue found for this specific divergence.
6. Ran the full `optional.*` and `collection.*` test packages (`OptionalSimpleTest`, `OptionalDifferentTypesTest`, `OptionalNestedTest`, `OptionalNullCheckAlwaysTest`, `CollectionMappingTest`, `MapMappingTest`, `AdderTest`, etc. — ~24 test classes) after the fix: all green, no regressions. `./mvnw -pl processor checkstyle:check` passes.


seonwooj0810:
Gentle ping — whenever there's bandwidth, could this get a look? It's scoped to `Optional<Map>`/`Optional<Collection>` source properties (#4111), which currently throw `NoSuchElementException` on an empty `Optional` instead of leaving the target unset. CI is green and it merges cleanly; happy to adjust the approach if you'd prefer the guard somewhere else.

PR #4124, opened 2026-09-09
#4123 Null check Optional source properties with NullValueCheckStrategy.ALWAYS

### LINKED ISSUE #4123: Optional source property that is itself null throws NullPointerException on the generated isPresent() check; NullValueCheckStrategy.ALWAYS is ignored
### Expected behavior

When a source property is `Optional<T>`, the generated presence check should tolerate the `Optional` reference itself being `null`, e.g.:

```java
if ( source.getName() != null && source.getName().isPresent() ) {
    name = source.getName().get();
}
```

or, at minimum, `nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS` should produce that guard, since that is the documented knob for "always null-check the source before using it".

### Actual behavior

The generated code calls `isPresent()` on the raw getter result, with no null check:

```java
@Override
public Target toTarget(Source source) {
    if ( source == null ) {
        return null;
    }

    String name = null;

    if ( source.getName().isPresent() ) {
        name = source.getName().get();
    }

    Target target = new Target( name );

    return target;
}
```

If the `Optional` property is `null` (which is a plain reference like any other, and is what many object builders and deserializers leave it as), this throws:

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.util.Optional.isPresent()" because the return value of "repro.Source.getName()" is null
	at repro.SourceMapperImpl.toTarget(SourceMapperImpl.java:20)
	at Demo.main(Demo.java:8)
```

The 1.7.0.Beta1 release notes say this is intentional: *"Note that we are not doing any null checks for the optional properties. Instead, we do a check if the optional is present or not and map it."* The problem is that this makes the generated mapper unsafe for any source object that can hold a `null` `Optional`, and there is no configuration to opt out of it.

**`NullValueCheckStrategy.ALWAYS` has no effect here.** With the identical model and

```java
@Mapper(nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS)
public interface SourceMapperAlways
{
    Target toTarget( Source source );
}
```

the generated body is byte-for-byte the same as with the default strategy — still a bare `source.getName().isPresent()`. That seems like the core issue: the setting that exists precisely to force source null checks is silently ignored for `Optional` properties.

This is not a hypothetical input. Two common cases where a `null` `Optional` source field arises:

1. **Generated builders.** Lombok's `@Builder` leaves an unset `Optional` component as `null`, not `Optional.empty()` (verified below with Lombok 1.18.48).
2. **Jackson deserialization**, where `null` (field absent from the JSON) and `Optional.empty()` (field explicitly `null` in the JSON) are deliberately distinguished — this exact scenario was raised in https://github.com/mapstruct/mapstruct/issues/674#issuecomment-4291090849 but there is no tracking issue for it, hence this report.

This is distinct from #4111, which is about the presence check being omitted entirely for `Optional<Map>` / `Optional<Collection>`; here the presence check is generated, but it is itself unguarded.

**Workaround** (works, but requires one `@Condition` method per mapper/type):

```java
@Condition
default boolean isSet( Optional<String> value )
{
    return value != null && value.isPresent();
}
```

which produces `if ( isSet( source.getName() ) )`.

Minimal case, no Lombok, plain `javac`, no other annotation processors.

`Source.java`:

```java
package repro;

import java.util.Optional;

public class Source
{
    private Optional<String> name;

    public Optional<String> getName()
    {
        return name;
    }

    public void setName( Optional<String> name )
    {
        this.name = name;
    }
}
```

`Target.java`:

```java
package repro;

public record Target( String name )
{ }
```

`SourceMapper.java`:

```java
package repro;

import org.mapstruct.Mapper;

@Mapper
public interface SourceMapper
{
    Target toTarget( Source source );
}
```

`Demo.java`:

```java
import repro.*;

public class Demo
{
    public static void main( String[] args )
    {
        Source source = new Source();
        System.out.println( new SourceMapperImpl().toTarget( source ) );
    }
}
```

```
javac -cp mapstruct-1.7.0.Beta2.jar \
      -processorpath mapstruct-processor-1.7.0.Beta2.jar:mapstruct-1.7.0.Beta2.jar \
      -s gen -d out src/repro/*.java src/Demo.java
```

Adding a second mapper with `nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS` over the same types generates an identical method body.

**Lombok variant**, showing that this is reachable without ever writing `null` by hand:

```java
package repro;

import java.util.Optional;
import lombok.Builder;

@Builder
public record Source( String id, Optional<String> name )
{ }
```

```java
Source source = Source.builder().id( "1" ).build();
System.out.println( "name field is: " + source.name() );   // => name field is: null
new SourceMapperImpl().toTarget( source );                  // => NPE
```

```
name field is: null
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.util.Optional.isPresent()" because the return value of "repro.Source.name()" is null
	at repro.SourceMapperImpl.toTarget(SourceMapperImpl.java:22)
	at Demo.main(Demo.java:9)
```

### MapStruct Version

1.7.0.Beta2, Java 25.0.2 (Eclipse Adoptium), javac. Lombok 1.18.48 for the second variant.

velo:
Fixes #4123

### Problem

For an `Optional` source property MapStruct generates a bare presence check:

```java
if ( source.getName().isPresent() ) {
    target.setName( source.getName().get() );
}
```

If the `Optional`-typed property is itself `null`, that throws a `NullPointerException`. A `null` `Optional` field is easy to reach in practice: Lombok's `@Builder` leaves an unset `Optional` component as `null`, and Jackson leaves it `null` when the JSON property is absent.

The release notes state the missing `null` check is intentional, and this PR keeps that default. The actual defect is that `nullValueCheckStrategy = NullValueCheckStrategy.ALWAYS` has no effect on `Optional` source properties: the generated method body is identical with and without it, which the committed fixture `OptionalNullCheckAlwaysMapperImpl.java` shows. `ALWAYS` is documented as "always include a null check when source is non primitive", so a source that can be `null` is exactly what it is for.

The reason is that an `Optional` source already forces `includeSourceNullCheck` on through the type conversion, so the `ALWAYS` branch in `setterWrapperNeedsSourceNullCheck` never changes the outcome, and the template renders the check as `.isPresent()`.

### Fix

`PropertyMapping` now builds an explicit presence check for an `Optional` source property when `NullValueCheckStrategy.ALWAYS` is configured, composing the existing `NullPresenceCheck` and `OptionalPresenceCheck`:

```java
if ( source.getName() != null && source.getName().isPresent() ) {
    target.setName( source.getName().get() );
}
```

### Scope

* Only applies with `nullValueCheckStrategy = ALWAYS`. The default `ON_IMPLICIT_CONVERSION` is unchanged, so no existing user sees a different mapper.
* Only applies to a direct (non-nested) source property, where the getter can safely be evaluated twice. Nested paths already go through a forged method and a local variable.
* A JSpecify `@NonNull` source still skips the guard, consistent with JSpecify taking precedence over `NullValueCheckStrategy` elsewhere.
* Not covered: an `Optional` source *parameter* (`Target map(Optional<Source> src)`) is still checked with a bare `isPresent()`, and #4111 (missing presence check for `Optional<Map>` / `Optional<Collection>`) is unrelated.

### Tests

Extended `org.mapstruct.ap.test.optional.nullcheckalways`:

* `optionalToOptionalWhenNull` / `optionalToNonOptionalWhenNull` — a `null` `Optional` property no longer throws and the target is left untouched.
* Updated `OptionalNullCheckAlwaysMapperImpl` fixture to the guarded form.
* New `OptionalDefaultNullCheckMapper` plus fixture, pinning the unchanged default-strategy output.

The existing empty/present cases and the other `optional` fixtures (default strategy) are unchanged.

`mvn install` on JDK 21: `MapStruct Processor` module green, 3649 tests, 0 failures. The `integrationtest` module was not run locally (it spawns nested Maven builds that need network access).

Documentation updated in `chapter-3` (Optional section) and `chapter-10` (`NullValueCheckStrategy`).


ijioio:
Please take patch semantics into account. I've described the problem in more detail [here](https://github.com/mapstruct/mapstruct/issues/4123#issuecomment-5690978045)
.

velo:
@ijioio your comments here and on #4123 were deleted, quoting from the notification emails:

> **[ijioio on #4124](https://github.com/mapstruct/mapstruct/pull/4124#issuecomment-5691001545):** Please take patch semantics into account. I've described the problem in more detail [here](https://github.com/mapstruct/mapstruct/issues/4128).

> **[ijioio on #4123](https://github.com/mapstruct/mapstruct/issues/4123#issuecomment-5690978045):**
>
> Do not add a check like this:
>
> ```java
> if (source.getName() != null && source.getName().isPresent()) {
>     target.setName(source.getName().get());
> }
> ```
>
> This breaks patch semantics.
>
> A patch needs to distinguish between three states:
>
> - `null` - ignore the field and keep its existing value
> - `Optional.empty()` - clear the field by setting it to `null`
> - `Optional.of(value)` - update the field with the wrapped value
>
> Therefore, the correct behavior is to restore the original semantics of `org.mapstruct.NullValuePropertyMappingStrategy.IGNORE`. This is also explicitly stated in its Javadoc:
>
> > If a source bean property equals null the target bean property will be ignored and retain its existing value.
>
> In other words, `null` must remain distinguishable from an empty Optional. The mapping logic should preserve that distinction rather than treating both cases as "do nothing."

---

Agreed on `Optional` → `Optional`: gating a same-type copy on `isPresent()` drops `Optional.empty()`. That is already on `main` under `ALWAYS` and this PR froze it into a fixture — I will change it to a `!= null` check only, keeping `!= null && isPresent()` only where the value is unwrapped.

The three-state patch behaviour itself is `NullValuePropertyMappingStrategy` on update methods, not `NullValueCheckStrategy`, so I would leave that to #4128. Fine with me if you would rather item 1 land there too.

ijioio:
@velo Yes, I've deleted my previous comment here cause I've confused `ALWAYS` strategy with `IGNORE` strategy, sorry for that. 

I'm still not sure whether the `Optional` unwrapping code can both preserve the three-state semantics and perform the unwrapping correctly. This is the approach I'm currently using for patch mappings (for MapStruct `1.6.3`):

```java
public class Source {
    private Optional<String> foo;
    public Optional<String> getFoo() { return foo; }
    public void setFoo(Optional<String> foo) { this.foo = foo; }
}

public class Target {
    private String foo;
    public String getFoo() { return foo; }
    public void setFoo(String foo) { this.foo = foo; }
}

@Mapper
    public static final OptionalMapper INSTANCE = Mappers.getMapper(OptionalMapper.class);
    public default <T> T extract(Optional<T> optional) {
        return optional.orElse(null);
    }
}

@Mapper(uses = OptionalMapper.class, nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.IGNORE)
public interface PatchMapper {
    public static final PatchMapper INSTANCE = Mappers.getMapper(PatchMapper.class);
    public void patch(@MappingTarget Target target, Source source);
}
```

So, `NullValuePropertyMappingStrategy.IGNORE` is what handles the three-state semantics, while `OptionalMapper` is responsible for unwrapping the `Optional` value. I hope this behavior can be preserved

PR #4129, opened 2026-09-17
#4034 Cache Type instances in TypeFactory.getType(...) to avoid redundant work

### LINKED ISSUE #4034: Cache `Type` instances in `TypeFactory.getType(...)` to avoid redundant work
## Context

`TypeFactory.getType(TypeMirror)` and `getType(TypeElement)` (processor/src/main/java/org/mapstruct/ap/internal/model/common/TypeFactory.java:209-238) allocate a new `Type` each call and re-run the full type analysis: `Types#isSubtypeErased` against `Iterable`/`Collection`/`Map`/`Stream`, declared-kind introspection, component-type computation, etc.

There is no interning. Callers that ask for the same `TypeMirror` repeatedly pay the full cost every time.

## Why this matters now

While reviewing #4033 (JSpecify nullness), two places were flagged that call `getType(...)` per property and then walk the enclosing-element chain for `@NullMarked`/`@NullUnmarked`:

- `PropertyMapping.java` — `getSourceJSpecifyNullability()` / `targetDeclaringTypeIsNullMarked()` per property mapping
- `PresenceCheckMethodResolver.java:98-108` — rebuilds `Type` for the mapper type on every presence-check resolution

`Type#isNullMarked` memoizes the walk per `Type` instance, but since each `getType(...)` returns a fresh instance, the memoization does not survive across calls. The effect is O(mappings × enclosingDepth) element-chain walks on large mappers, and it is not unique to JSpecify — any code path that asks for the same type twice pays similar overhead today.

## Proposal

Intern `Type` by a stable key:
- For `TypeElement`-keyed lookups: `IdentityHashMap<TypeElement, Type>` is safe within a single processing round.
- For `TypeMirror`-keyed lookups: `TypeMirror` does not have useful equality, so intern via the resolved `TypeElement` (when `DeclaredType`) or a fingerprint for primitive/array/wildcard types. Fall back to no-caching where a stable key is not available.

Edges to think about:
- `alwaysImport` / `isLiteral` flags are currently passed through — the cache key needs to include these, or the memoized `Type` needs to be immutable w.r.t. them.
- Incremental builds (javac + Gradle/Eclipse): `TypeElement` identity is stable within a round, but not across rounds. Scope the cache to the `TypeFactory` lifetime (one per processing round today) and this is fine.
- `TypeHierarchyErroneousException` must still be thrown for unresolvable mirrors — don't cache the erroneous state in a way that swallows it later.

## Acceptance

- `getType(someElement)` called N times returns the same instance (or an equivalent instance from a cache), and measurable reduction in `isSubtypeErased` calls on a representative mapper (e.g. something in `integrationtest` with many properties).
- No behavior change for `alwaysImport` / literal variants.
- `Type#isNullMarked` (and any future per-`Type` memoization) becomes effective across repeated lookups.

## Related

- PR #4033 (JSpecify nullness support) added `Type#isNullMarked` with per-instance caching that the current `getType` contract makes ineffective; the Javadoc there was corrected in a follow-up but the underlying perf concern remains.

AnkitaAdvitot:
Fixes #4034.

### Background & Motivation
In MapStruct's annotation processor, `TypeFactory.getType(...)` is invoked frequently across rounds and model element processors (mapping methods, source/target property resolution, accessor discovery, enclosing element inspection). Previously, `TypeFactory` instantiated a fresh `Type` object on nearly every invocation:
- This caused repetitive enclosing-element traversals, member lookups, and `typeUtils.isSubtypeErased(...)` checks.
- Per-`Type` memoized caches (such as `isNullMarked()`, accessors, and method lists) were repeatedly re-allocated or re-queried instead of shared.

### Solution
1. **Round-Scoped Type Element Cache:**
   - Added `typeElementCache` (`IdentityHashMap<TypeElement, Type>`) in `TypeFactory` to intern prototypical types derived directly from non-generic `TypeElement` instances.
2. **Structural Type Mirror Cache:**
   - Added `typeCache` (`HashMap<TypeCacheKey, Type>`) using an immutable structural fingerprint (`TypeFingerprint`) that safely distinguishes:
     - Declared types (interned by qualified name / canonical `TypeElement`, type argument fingerprints, and enclosing type fingerprint).
     - Array types (by component type fingerprint).
     - Wildcard types (by extends / super bound fingerprints).
     - Type variables (by name and upper bound fingerprints).
     - Primitive and void types (by `TypeKind`).
   - The cache key strictly distinguishes contextual configuration flags: `isLiteral` and `alwaysImport`.
3. **Safety & Erroneous Types:**
   - Pre-validation check `canBeProcessed(mirror)` ensures erroneous/unresolvable types are never cached or suppressed, preserving `TypeHierarchyErroneousException` semantics.
4. **Testing:**
   - Added `Issue4034Test` and `Issue4034Mapper` verifying that identical `Type` instances are returned for repeated lookups (`getType(TypeElement)`, `getType(TypeMirror)`, generic `List<String>`), differing type arguments (`List<String>` vs `List<Integer>`) produce distinct instances, and flag variants (`isLiteral`, `alwaysImport`) remain distinct.
   - All tests pass and checkstyle audit passes with 0 violations.


hduelme:
@AnkitaAdvitot thanks for your contribution.
Can you explain your test a bit more. Currently I don't understand why you combined `@ProcessorTest` and `@Test` in one test class.

AnkitaAdvitot:
Hi @hduelme,

Thanks for pointing that out!

Originally, I combined them in `Issue4034Test` to cover verification for #4034 from two angles:
1. `@ProcessorTest` (`shouldCompileMapperWithCachedTypes`) verified that end-to-end mapper compilation and runtime execution worked without regressions across both compilers (JDK & Eclipse).
2. `@Test` (`shouldInternTypeInstancesAcrossLookups`) directly asserted `TypeFactory`'s caching/interning semantics (`isSameAs`, parameterized types, literals, etc.), which couldn't be asserted inside `@ProcessorTest` since it runs post-compilation without access to internal compiler components like `TypeFactory`.

I agree that combining them in one test class with `@WithClasses` is confusing and doesn't follow MapStruct conventions.

I have separated them:
- `Issue4034Test` is now purely a `@ProcessorTest` bug test for `Issue4034Mapper`.
- The `TypeFactory` caching unit test has been moved into its own dedicated unit test class: `org.mapstruct.ap.internal.model.common.TypeFactoryTest`.

I've pushed the updated commit to this branch.

